PR #7, opened 2015-02-03
Console fixes

terrajobst:


KrzysztofCwalina:
One thing I wanted to do with the demo was to show that it's possible to pass command line args to the app. Maybe you could change the sample such that optionally people can pass in the size of the square?

But I think it's ok to check in as-is, if you don't have time to do it now.

akoeplinger:
Where should I file the feature request for drawing a penguin on Linux and an apple on OSX :see_no_evil: :laughing:

KrzysztofCwalina:
Maybe that should be the command line arg? :-)

stephentoub:
If you go ahead and check this in, I'll add Tux :)

terrajobst:
Go for it! It would be awesome to detect the OS and draw the appropriate ASCII art :-)

KrzysztofCwalina:
Looks great

PR #9, opened 2015-02-03
Add per-platform ANSI art to HelloWorld

stephentoub:
Just by looking at the code you can likely guess what gets drawn for Linux ;)

I've left Mac as a todo for someone else who'd like to contribute something beautiful.

Right now, the console app just takes the target platform as a command-line argument.  We can update it later to fall back to a default based on Environment.


KrzysztofCwalina:
looks good

PR #13, opened 2015-02-08
Add missing nuget.config

terrajobst:
Some of the packages require CoreFX MyGet feed. In order to make restoring packages work out-of-the-box we need to add a nuget.config file that points to the correct feed.

@KrzysztofCwalina


KrzysztofCwalina:
Looks good!

PR #77, opened 2015-03-31
Fixed a bug exposed in the recent tools update

KrzysztofCwalina:
For some reason the previous C# compiler was not complaining about default ctors in structs.
Also, it apparently was running such ctors.


davidfowl:
This was a recent change to the C# compiler

PR #83, opened 2015-04-28
HelloWorld: Pimp up the FreeBSD daemon to have colors.

josteink:
Basically pay full respects to the original console screen-saver.

Word on the street is that there's this "build"-conference going on soon and some people (like @richlander ) mentioned it would be nice to have this committed before then :)


jkotas:
Beautiful

richlander:
Wow. That's high praise from @jkotas. 

@josteink Print the t-shirt now!

josteink:
If you look at the [original source](https://github.com/oza/FreeBSD-8.0-dyntick/blob/master/sys/dev/syscons/daemon/daemon_saver.c#L57), the original artist is uncredited.

We don't know who he or she is, but I refuse to steal the honours from this *fantastic* ASCII artist.

PR #90, opened 2015-06-02
Polling watcher

KrzysztofCwalina:
It's often preferable to poll for file system changes instead of using FileSystemWatcher. There are many samples on the internet showing how to do such polling. All the ones I looked at allocate memory on every polling internal. This library only allocates when actual changes are detected. 

It is an early prototype and so does not support several important features: watching subfolders, discriminating between different change types, watching UNC paths, and some others.


src/System.IO.FileSystem.Watcher.Polling/src/System/IO/PollingWatcher.cs
```diff
@@ -0,0 +1,141 @@
...
+                var changes = ComputeChangesAndUpdateState(_directory);
+                if (!changes.IsEmpty)
+                {
+                    handler();
+                }
+            }
+
+            _timer.Change(_pollingIntervalInMilliseconds, Timeout.Infinite);
+        }
+        private FileChangeList ComputeChangesAndUpdateState(string directory)
```
davidfowl: I need to see how this scales for large directories. The mono file watcher fell over pretty quickly in that case. CPU usage was through the roof.

KrzysztofCwalina: How large are the directories you need to watch? What do you think is a reasonable polling interval? If you let me know, I can run some quick tests to see how this implementation behaves.

KrzysztofCwalina: I just ran quick and dirty test on a folder with:
a) 4000 files. One polling cycle takes ~40ms (~150k ticks).
b) 500 files. ~7000 ticks (1 ms)
c) 100 files -> ~700 ticks (approaching 0ms)

The hot code is in the Equal(string, WIN32_FIND_DATAW) routine. It's responsible for ~50% of the samples. I can optimize the routine and in addition, the routine is probably called too much because of hashatable bucket conflicts, which I can also improve by making it sparser.

KrzysztofCwalina: After some changes to the code I got 4000 files folder down to 20ms per polling cycle

KrzysztofCwalina: Yeah, it's all about the hashcode computation and distribution in the hashtable. I am pretty sure I can get it down another 50% by playing with it. though, I am not sure if it's even needed (let me know).

akoeplinger: Just for reference, my corefx dir has ~13.000 files after building and we're still missing 40% of the code in that repo :smile: you might want to use that as a testbed as well.

davidfowl: @KrzysztofCwalina try a folder with `node_modules` in it. Those have super deep hierarchies that might push the limits. Those numbers look promising though.

KrzysztofCwalina:
OK, I am going to merge this prototype and keep working on it. Right now it does not support watching subfolders, so I cannot actually try it on node_modules, but node_modules has ~2500 files and so I think I can make it work in under 10ms.

PR #92, opened 2015-06-03
Changed to use FindFirstFileExW instead of FindFirstFile

KrzysztofCwalina:
This improves performance approximatelly by 2x

Now, on my home computer, the node_modules folder (and its subfolders) can be processed in ~35ms


janvorli:
@KrzysztofCwalina is this supposed to run on Unix too? If that's the case, we would need to add the `FindFirstFileExW` to PAL.

stephentoub:
We wouldn't want to add it to the libcoreclr PAL, though; if/when we want to make this library support Unix, we'd want to introduce a mini-PAL layer into the library itself, as we've done elsewhere in corefx.

janvorli:
@stephentoub Makes sense

PR #106, opened 2015-06-18
Add dnulist tool

richlander:
This is a tool that is intended to solve the problem of using NuGet package artifacts w/o needing to understand the format / binding semantics. This is discussed here: https://github.com/dotnet/coreclr/issues/995.

The tool overlaps a fair bit with dnu. Hopefully, this functionality is added to dnu. Currently, the tool has the following basic functions:

- List targets.
- List packages by target.
- Copy packages to a directory, by target.

@davidfowl, @ericstj, @blackdwarf, @akoeplinger, @KrzysztofCwalina

### DIFF
--- a/tools/dnulist/program.cs
+++ b/tools/dnulist/program.cs
@@ -1,0 +1,241 @@
+using System;
+using System.IO;
+using Newtonsoft.Json;
+using System.Collections.Generic;
+
+public class Program
+{
+    private static void Main (string[] args)
+    {
+        // project.lock.json reader
+        // See PrintHelp for arg info
+        
+        const string Lockfile = "project.lock.json";
+        Project project;
+        
+        Console.WriteLine("project.json.lock file reader");
+        
+        using (var reader = new StreamReader(Lockfile))
+        {
+            project = JsonConvert.DeserializeObject<Project>(reader.ReadToEnd());
+        }
+        
+        if (args == null || args.Length == 0)
+        {
+            PrintTargets(project);
+            Console.WriteLine();
+            Console.WriteLine("Use --help argument to print help.");
+        }
+        else
+        {
+            var commands = GetArguments(args);
+            const string Target = "--target";
+            const string Copy = "--copy";
+            const string Verbose = "--verbose";
+            const string Help = "--help";
+            
+            if (commands == null)
+            {
+                Console.WriteLine("Bad argument!");
+                Console.WriteLine();
+                PrintHelp();
+            }
+            else if (commands.ContainsKey("--targets"))
+            {
+                PrintTargets(project);
+            }
+            else if (commands.ContainsKey(Target) && commands.ContainsKey(Copy) && !string.IsNullOrEmpty(commands[Copy]))
+            {
+                bool verbose = commands.ContainsKey(Verbose);
+                CopyTarget(project, commands[Target], commands[Copy], verbose);
+            }
+            else if (commands.ContainsKey(Target))
+            {
+                bool verbose = commands.ContainsKey(Verbose);
+                PrintTarget(project,commands[Target], verbose);
+            }
+            else if (commands.ContainsKey(Help))
+            {
+                Console.WriteLine();
+                PrintHelp();
+            }
+        }
+    }
+    
+    private static void PrintTargets(Project project)
+    {
+        Console.WriteLine ("This project has {0} targets.", project.Targets.Count);
+        Console.WriteLine();
+        
+        foreach (var t in project.Targets.Keys)
+        {
+            Console.WriteLine(t);
+        }
+    }
+    
+    private static void PrintTarget(Project project, string target, bool verbose = false)
+    {
+        Dictionary<string,Package> packages;
+        var targetFound = project.Targets.TryGetValue(target, out packages);
+        
+        if (targetFound && packages.Count > 0)
+        {
+            Console.WriteLine("Package dependencies for target: {0}",target);
+            
+            foreach (var package in packages.Keys)
+            {
+                Console.WriteLine(package);
+                if (verbose)
+                {
+                    var p = packages[package];
+               
+                    if (p.Compile != null)
+                    {
+                        foreach (var key in p.Compile.Keys)
+                        {
+                            Console.WriteLine("**Compile: {0}", key);
+                        }
+                    }
+                    
+                    if (p.Runtime != null)
+                    {
+                        foreach (var key in p.Runtime.Keys)
+                        {
+                            Console.WriteLine("**Runtime: {0}", key);
+                        }
+                    }
+                    
+                    if (p.Dependencies != null)
+                    {
+                        foreach (var key in p.Dependencies.Keys)
+                        {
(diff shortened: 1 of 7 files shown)


tools/dnulist/program.cs
```diff
@@ -0,0 +1,241 @@
...
+        {
+            Console.WriteLine("{0} is not a target for this project.", target);
+        }
+    }
+    
+    private static void PrintHelp()
+    {
+        Console.WriteLine("Primary arguments:");
+        Console.WriteLine("--target - prints targets");
+        Console.WriteLine("--target [target] - selects targets to print (the default) or otherwise act upon");
```
ericstj: I think you need both framework and runtime selection instead of all in one.  Compile time assets should only be selected from "RID"-less targets.  Runtime assets should be selected from the target with the RID.

tools/dnulist/project.lock.json
```diff
@@ -0,0 +1,408 @@
...
+        "System.Runtime.Handles.4.0.0-beta-22906.nupkg",
+        "System.Runtime.Handles.4.0.0-beta-22906.nupkg.sha512",
+        "System.Runtime.Handles.nuspec",
+        "lib/aspnetcore50/System.Runtime.Handles.dll",
+        "lib/contract/System.Runtime.Handles.dll",
+        "lib/net45/System.Runtime.Handles.dll",
+        "lib/portable-wpa81+wp80+win80+net45+aspnetcore50/System.Runtime.Handles.dll"
+      ]
+    },
+    "System.Text.Encoding/4.0.10-beta-22231": {
```
ericstj: Are you sure you had https://www.myget.org/F/dotnet-core/ enabled when generating this?  These are some very old packages.

tools/dnulist/program.cs
```diff
@@ -0,0 +1,241 @@
...
+        }
+        else
+        {
+            Console.WriteLine("{0} is not a target for this project.", target);
+        }
+    }    
+    
+    private static void CopyTarget(Project project, string target, string path, bool verbose = false)
+    {
+        const string packageRoot = "/Users/richlander/.dnx/packages";
```
ericstj: Did you mean to  use %USERPROFILE% here?  Also you probably need an argument for overriding this.

tools/dnulist/program.cs
```diff
@@ -0,0 +1,241 @@
...
+    {
+        // project.lock.json reader
+        // See PrintHelp for arg info
+        
+        const string Lockfile = "project.lock.json";
+        Project project;
+        
+        Console.WriteLine("project.json.lock file reader");
+        
+        using (var reader = new StreamReader(Lockfile))
```
ericstj: So you'll crash without any error if the lock file is invalid?

tools/dnulist/program.cs
```diff
@@ -0,0 +1,241 @@
...
+using System.Collections.Generic;
+
+public class Program
+{
+    private static void Main (string[] args)
+    {
+        // project.lock.json reader
+        // See PrintHelp for arg info
+        
+        const string Lockfile = "project.lock.json";
```
ericstj: Should be a parameter with a default value of this.

tools/dnulist/program.cs
```diff
@@ -0,0 +1,241 @@
...
+            }
+            else if (args[count].StartsWith(ArgSyntax) && !args[count+1].StartsWith(ArgSyntax))
+            {
+                commands.Add(args[count],args[count+1]);
+                count++;
+            }
+            else if (args[count].StartsWith(ArgSyntax))
+            {
+                commands.Add(args[count],string.Empty);
+            }
```
ericstj: This will ignore an unqualified argument in the middle of the list.  EG: --target abc def --copy.  def would be ignored.

tools/dnulist/program.cs
```diff
@@ -0,0 +1,241 @@
...
+
+        while (count < args.Length)
+        {   
+            if (count +1 == args.Length && args[count].StartsWith(ArgSyntax))
+            {
+                commands.Add(args[count],string.Empty);
+            }
+            else if (count +1 == args.Length && !args[count].StartsWith(ArgSyntax))
+            {
+                // bad argument
```
ericstj: Perhaps you could use the unqualified syntax to represent the file to operate on?  I think in that case you could turn this into  if (!args[count].StartsWith(ArgSyntax))

richlander:
Awesome feedback @ericstj. I'll look into all of that.

I haven't done the next step of actually using the tool the compile code and run it (the real scenario). I think that I'll find that having the tool generate the CSC command line would be quite convenient. I was thinking of adding that, too.

tools/dnulist/program.cs
richlander: I haven't checked. Is that supported on OS X and Linux, too? I was looking for an API to get that info, but didn't find one. Seems like a common scenario. Any plans on that?

ericstj: Roots are very platform specific.  We do have cross-plat environment variables.  It does look like DNX sets DNX_HOME env variable.  Perhaps this is a reasonable cross-plat way to locate a default package cache.  /cc @davidfowl @anurse

jasonwilliams200OK: > Is that supported on OS X and Linux, too

Related discussion about [`System.Environment.SpecialFolder`](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Environment.cs#L1551-L1751) on Gitter chat CoreCLR channel: https://gitter.im/dotnet/coreclr?at=5580010ac030cae67ed3ccef

tools/dnulist/project.lock.json
richlander: I did have a wrong setup. How's this look? I believe I'm now pulling down much later packages. If that doesn't look good, how do I enable pull that feed via dnu?

"System.Resources.ResourceManager/4.0.0-beta-23011": {
        "dependencies": {
          "System.Runtime": "4.0.0-beta-23011",
          "System.Reflection": "4.0.0-beta-23011",
          "System.Globalization": "4.0.0-beta-23011"
        },
        "compile": {
          "ref/dotnet/System.Resources.ResourceManager.dll": {}
        },
        "runtime": {
          "lib/DNXCore50/System.Resources.ResourceManager.dll": {}
        }
      },

tools/dnulist/program.cs
richlander: None of these environment variables seem to be defined under 1.0.0-beta6-12100.

Having DNX_HOME would be nice, but that shouldn't be the primary answer for this.

tools/dnulist/program.cs
richlander: Nice. Will fix that. ;)

tools/dnulist/program.cs
ericstj: The environment variable is defined by DNVM, not DNX.  If not the environment variable then force folks to pass it in.  You shouldn't be making an assumption about where nuget/DNX decides to store it's packages.  The location is configurable and not part of any "public contract".

KrzysztofCwalina:
@richlander, what should we do with this PR? It's been 2 months.

PR #107, opened 2015-06-23
Update to HelloWorld.cs

ryaneliseislalom:
Updated corefxlab/demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs to include a function for OS detection while still allowing args[0] to be used for overriding and showing various OS logos. No break to existing behavior.

Also note, that proper OS detection would use Diagnostics.Environment, which is not yet implemented in CoreFX. This has been noted in the detection method comments.

### DIFF
--- a/HelloWorld.cs
+++ b/HelloWorld.cs
@@ -1,0 +1,345 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using System.Diagnostics;
+
+internal class Program
+{
+    private static void Main(string[] args)
+    {
+        string osDetect = DetectOS();
+
+        if (args.Length >= 1)
+        {
+            osDetect = args[0];
+        }
+
+        if (osDetect == "linux")
+        {
+            DrawLinux();
+        }
+        else if (osDetect == "freebsd")
+        {
+            DrawFreeBSD();
+        }
+        else if (osDetect == "mac")
+        {
+            DrawMac();
+        }
+        else if (osDetect == "windows")
+        {
+            DrawWindows();
+        }
+        else
+        {
+            DrawWindows();
+        }
+
+        Console.WriteLine();
+        Console.WriteLine("Press ENTER to continue . . . ");
+        Console.ReadLine();
+    }
+
+    private static void DrawWindows()
+    {
+        Console.WriteLine("Hello, Windows...");
+
+        const int squareSize = 20;
+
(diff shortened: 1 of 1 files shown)


HelloWorld.cs
```diff
@@ -0,0 +1,345 @@
...
+                    Console.BackgroundColor = colors[row * 2 + col];
+                    Console.ForegroundColor = colors[row * 2 + col];
+                    for (int j = 0; j < squareSize; j++) Console.Write('@');
+                    Console.ResetColor();
+                }
+            }
+        }
+        Console.WriteLine();
+    }
+
```
KrzysztofCwalina: Why all these show as new? Did you change anything in the Draw* methods?

ryaneliseislalom: I didn’t see a place to create a feature branch, but did see an option to submit a file in the fork dialog. Please advise if change is needed. Summary is addition of new function, and minor modification of conditionals in Main. No other changes.

HelloWorld.cs
```diff
@@ -0,0 +1,345 @@
...
+            }
+
+        }
+
+        Console.ResetColor();
+        Console.WriteLine();
+    }
+
+    //HACK: Needed until CoreFX implements Environment.OSVersion
+    private static string DetectOS()
```
ryaneliseislalom: This method is new

HelloWorld.cs
```diff
@@ -0,0 +1,345 @@
...
+
+using System;
+using System.Collections;
+using System.Diagnostics;
+
+internal class Program
+{
+    private static void Main(string[] args)
+    {
+        string osDetect = DetectOS();
```
ryaneliseislalom: This detection behavior is new

HelloWorld.cs
```diff
@@ -0,0 +1,345 @@
...
+        {
+            string shellValue = proc.StartInfo.Environment["SHELL"].ToLowerInvariant();
+
+            if (shellValue.EndsWith("/tcsh") || shellValue.EndsWith("/sh"))
+            {
+                detectionResult = "freebsd";
+            } else {
+                detectionResult = "linux";
+            }
+        } else if (proc.StartInfo.Environment.ContainsKey("COMSPEC"))
```
bartonjs: The bracing on the else/else-ifs should be
```
if (condition)
{
    //body;
}
else if (condition2)
{
    //body2;
}
else
{
    //body3;
}
```

https://github.com/dotnet/corefx/blob/master/Documentation/coding-style.md (rule 1)

PR #108, opened 2015-06-24
Added OS detection pending CoreFX implementing Diagnostics.Environment

ryaneliseislalom:
Updated corefxlab/demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs to include a function for OS detection while still allowing args[0] to be used for overriding and showing various OS logos. No break to existing behavior.

Also note, that proper OS detection would use Diagnostics.Environment, which is not yet implemented in CoreFX. This has been noted in the detection method comments.

### DIFF
--- a/demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs
+++ b/demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs
@@ -2,22 +2,35 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Collections;
+using System.Diagnostics;
 
 internal class Program
 {
     private static void Main(string[] args)
     {
-        if (args.Length == 1 && args[0] == "linux")
+        string osDetect = DetectOS();
+
+        if (args.Length >= 1)
+        {
+            osDetect = args[0];
+        }
+
+        if (osDetect == "linux")
         {
             DrawLinux();
         }
-        else if (args.Length == 1 && args[0] == "freebsd")
+        else if (osDetect == "freebsd")
         {
             DrawFreeBSD();
         }
-        else if (args.Length == 1 && args[0] == "mac")
+        else if (osDetect == "mac")
         {
             DrawMac();
+        }
+        else if (osDetect == "windows")
+        {
+            DrawWindows();
         }
         else
         {
@@ -25,7 +38,7 @@ internal class Program
         }
 
         Console.WriteLine();
-        Console.WriteLine("Press ENTER to exit ...");
+        Console.WriteLine("Press ENTER to continue . . . ");
         Console.ReadLine();
     }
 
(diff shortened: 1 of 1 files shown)


demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs
```diff
@@ -2,30 +2,43 @@
...
     {
-        if (args.Length == 1 && args[0] == "linux")
+        string osDetect = DetectOS();
+
+        if (args.Length >= 1)
+        {
+            osDetect = args[0];
+        }
+
+        if (osDetect == "linux")
```
stephentoub: I'd suggest changing this to just be a switch statement now, especially with the duplicated DrawWindows at the end, e.g.
```C#
switch (osDetect)
{
    case "linux": 
        DrawLinux(); 
        break;
    case "freebsd": 
        DrawFreeBSD();
        break;
    case "mac":
        DrawMac();
        break;
    default:
        DrawWindows();
        break;
}
```

demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs
```diff
@@ -301,4 +314,36 @@ @@@@@@@@@@@         @@@@@@@@@@@@@
...
     }
+
+    //HACK: Needed until CoreFX implements Environment.OSVersion
+    private static string DetectOS()
+    {
+        Process proc = new Process();
+
+        string detectionResult = String.Empty;
+
+        if (proc.StartInfo.Environment.ContainsKey("TERM_PROGRAM"))
```
stephentoub: Any reason not to use Environment.GetEnvironmentVariable rather than using Process?

demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs
```diff
@@ -301,4 +314,36 @@ @@@@@@@@@@@         @@@@@@@@@@@@@
         Console.ResetColor();
         Console.WriteLine();
     }
+
+    //HACK: Needed until CoreFX implements Environment.OSVersion
```
stephentoub: I realize this is a temporary hack, but it still might be better to have something slightly less hacky, e.g. P/Invoke to uname... if the P/Invoke fails to find the entrypoint, assume you're on Windows, and if it succeeds, use the system name returned to determine what OS you're on.  We've been doing something similar in the corefx tests with https://github.com/dotnet/corefx/blob/master/src/Common/src/Interop/Interop.PlatformDetection.cs#L45 (this will also switch to the new System.Runtime.InteropServices.RuntimeInformation once it's available via nuget).

demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs
KrzysztofCwalina: The problem with this is that the actual OS detection APIs are a query, not something that returns a value that can be switched on. And so once this sample moves to the official OS detection APIs, it will have to be rewritten back to if/else statements.

stephentoub: One more reason I'd have preferred if we had more of a uname-like mechanism. ;)

http://linux.die.net/man/2/uname

stephentoub: (If the structure is left the way it is, we should at least remove the duplication at the end around Windows.)

demos/CoreClrConsoleApplications/HelloWorld/HelloWorld.cs
```diff
@@ -301,4 +314,36 @@ @@@@@@@@@@@         @@@@@@@@@@@@@
...
+
+        if (proc.StartInfo.Environment.ContainsKey("TERM_PROGRAM"))
+        {
+            detectionResult = "mac";
+        }
+        else if (proc.StartInfo.Environment.ContainsKey("SHELL"))
+        {
+            string shellValue = proc.StartInfo.Environment["SHELL"].ToLowerInvariant();
+
+            if (shellValue.EndsWith("/tcsh") || shellValue.EndsWith("/sh"))
```
bartonjs: I wouldn't question an assertion that bash is the most popular shell on Linux, but I do question an assertion that no one on Linux uses tcsh or sh.  (Or bash executed as /bin/sh).

Really I think I'm just in @stephentoub's camp here: for best (consistency in) results copy the PlatformDetection logic from CoreFx (pending it being officially queryable API)

KrzysztofCwalina:
@ryaneliseislalom could the code be updated to use the OS detection APIs?

ryaneliseislalom:
@KrzysztofCwalina which APIs do you recommend? From what I can tell, System.Environment.OSVersion is not implemented yet.

KrzysztofCwalina:
@ryaneliseislalom, sorry for a long delay. I was on vacation. Please use https://github.com/dotnet/corefx/tree/master/src/System.Runtime.InteropServices.RuntimeInformation

KrzysztofCwalina:
I will close this for now. Please reactivate if you still want to merge it.

PR #117, opened 2015-07-07
Graphics library david

dalbanhi:
Would appreciate any feedback


src/System.Drawing.Graphics/tests/UnitTests.cs
```diff
@@ -9,12 +9,189 @@
 public partial class GraphicsUnitTests
 {
     [Fact]
-    public void Test()
```
axxu: don't get rid of my test plz :'(

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
...
+        {
+            get { return _width; }
+        }
+        public int HeightInPixels
+        {
+            get { return _height; }
+        }
+        public PixelFormat PixelFormat
+        {
+            get { return PixelFormat.RGBA; }
```
whoisj: There's likely history here, but why is the format fixed at RGBA? A comment here would likely go a long way.

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        public byte[] Data
+        {
+            get { return _data; }
```
whoisj: Returning the raw data can likely lead to a race condition as image manipulation is often not done on the UI thread while image rendering is. Is that a concern for this class?

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
...
+        }
+        public int BytesPerPixel
+        {
+            get { return 0; }
```
whoisj: Fixing `BytesPerPixel` at `0` seems... concerning?

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+
+namespace System.Drawing.Graphics
+{
+    public enum PixelFormat
```
whoisj: Is there a reason that we're assuming only 32-bit image formats and only these three formats?

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
dalbanhi: This is just a placeholder for now, until functionality to actually return the appropriate PixelFormat is implemented, will add that comment for sure

whoisj: Have we started enumerating the supported types? If so, there could be issues as some of the types are 96-bit floating color types etc. There's a lot of assumptions in this class (which can be OK) but there are no comments or documentation addressing them.

dalbanhi:
Will add comments to explain placeholders until further functionality is implemented

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
...
+            /* Constructors */
+        private Image(int width, int height)
+        {
+            throw new NotImplementedException();
+        }
+        private Image(string filepath)
+        {
+            throw new NotImplementedException();
+        }
+        private Image(object stream)
```
whoisj: Should this be of type `object` or of type `Stream`?

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
dsplaisted: > Returning the raw data can likely lead to a race condition as image manipulation is often not done on the UI thread while image rendering is. Is that a concern for this class?

I don't think so.  Here we are prototyping a library focused on server-side image manipulation.  See https://github.com/dotnet/corefx/issues/2020 for some of the context.

We do intend for the library to still work in client scenarios, and in that context you'd need to copy an image so you can render one copy and manipulate another.

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
...
+using System.Collections.Generic;
+using System.Text;
+using System.Threading.Tasks;
+
+
+namespace System.Drawing.Graphics
+{
+    public enum PixelFormat
+    {
+        ARGB,
```
KrzysztofCwalina: These should follow our casing conventions. See https://msdn.microsoft.com/en-us/library/ms229043(v=vs.110).aspx

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
...
+            /* Factory Methods */
+        public static Image Create(int width, int height)
+        {
+            return new Image(width, height);
+        }
+        public static Image Load(string filePath)
+        {
+            return new Image(filePath);
+        }
+        public static Image Load(object stream)
```
KrzysztofCwalina: Why is it typed as object?

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
+﻿using System;
```
KrzysztofCwalina: You should add the licensing comment we use in this repo. See the DLLImports file I changed yesterday.

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -0,0 +1,75 @@
...
+        private byte[] _data;
+        private int _width;
+        private int _height;
+
+            /* Properties */
+        public int WidthInPixels
+        {
+            get { return _width; }
+        }
+        public int HeightInPixels
```
KrzysztofCwalina: I wonder if simply Height and Width would not be better.

src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
dalbanhi: Do you mean specifically that ARGB should be Argb? The enum PixelFormat is in the Pascal Casing Convention

whoisj: Argb
Rgba
Cmyk

KrzysztofCwalina: Correct.

PR #118, opened 2015-07-07
Add System.Numerics.Matrices

whoisj:
Adds System.Numerics.Matrices

Based on a text-template which can support any sized matrix

Created up to 4x4 matrices initially (up for discussion)

Adds tests for each matrix type

Tests are text template generated, therefore easy to add to each matrix type

Adds a .gitignore to the System.Numerics.Matrices directory to ignore TT temp files

### DIFF
--- a/src/System.Numerics.Matrices/src/System/Numerics/IMatrix.cs
+++ b/src/System.Numerics.Matrices/src/System/Numerics/IMatrix.cs
@@ -1,0 +1,9 @@
+﻿namespace System.Numerics.Matrices
+{
+    interface IMatrix
+    {
+        double this[int col, int row] { get; set; }
+        int Columns { get; }
+        int Rows { get; }
+    }
+}
--- a/src/System.Numerics.Matrices/src/System/Numerics/Matrix4x4.cs
+++ b/src/System.Numerics.Matrices/src/System/Numerics/Matrix4x4.cs
@@ -1,0 +1,435 @@
+// Copyright (c) Microsoft. All rights reserved. 
+// Licensed under the MIT license. See LICENSE file in the project root for full license information. 
+
+using System.Runtime.InteropServices;
+
+namespace System.Numerics.Matrices
+{
+    /// <summary>
+    /// Represents a matrix of double precision floating-point values defined by its number of columns and rows
+    /// </summary>
+    [StructLayout(LayoutKind.Sequential)]
+    public struct Matrix4x4: IEquatable<Matrix4x4>, IMatrix
+    {
+        public const int ColumnCount = 4;
+        public const int RowCount = 4;
+
+        static Matrix4x4()
+        {
+            Zero = new Matrix4x4(0);
+            Identity = new Matrix4x4(1, 0, 0, 0, 
+                                     0, 1, 0, 0, 
+                                     0, 0, 1, 0, 
+                                     0, 0, 0, 1);
+        }
+
+        
+        /// <summary>
+        /// Constant Matrix4x4 with value intialized to the identity of a 4 x 4 matrix
+        /// </summary>
+        public static readonly Matrix4x4 Identity;
+        /// <summary>
+        /// Gets the smallest value used to determine equality
+        /// </summary>
+        public double Epsilon { get { return MatrixHelper.Epsilon; } }
+
+        /// <summary>
+        /// Constant Matrix4x4 with all values initialized to zero
+        /// </summary>
+        public static readonly Matrix4x4 Zero;
+
+        /// <summary>
+        /// Initializes a Matrix4x4 with all of it values specifically set
+        /// </summary>
+        /// <param name="m11">The column 1, row 1 value</param>
+        /// <param name="m21">The column 2, row 1 value</param>
+        /// <param name="m31">The column 3, row 1 value</param>
+        /// <param name="m41">The column 4, row 1 value</param>
+        /// <param name="m12">The column 1, row 2 value</param>
+        /// <param name="m22">The column 2, row 2 value</param>
+        /// <param name="m32">The column 3, row 2 value</param>
+        /// <param name="m42">The column 4, row 2 value</param>
+        /// <param name="m13">The column 1, row 3 value</param>
+        /// <param name="m23">The column 2, row 3 value</param>
+        /// <param name="m33">The column 3, row 3 value</param>
+        /// <param name="m43">The column 4, row 3 value</param>
+        /// <param name="m14">The column 1, row 4 value</param>
+        /// <param name="m24">The column 2, row 4 value</param>
+        /// <param name="m34">The column 3, row 4 value</param>
+        /// <param name="m44">The column 4, row 4 value</param>
+        public Matrix4x4(double m11, double m21, double m31, double m41, 
+                         double m12, double m22, double m32, double m42, 
+                         double m13, double m23, double m33, double m43, 
+                         double m14, double m24, double m34, double m44)
+        {
+            M11 = m11; M21 = m21; M31 = m31; M41 = m41; 
+            M12 = m12; M22 = m22; M32 = m32; M42 = m42; 
+            M13 = m13; M23 = m23; M33 = m33; M43 = m43; 
+            M14 = m14; M24 = m24; M34 = m34; M44 = m44; 
+        }
+
+        /// <summary>
+        /// Initialized a Matrix4x4 with all values set to the same value
+        /// </summary>
+        /// <param name="value">The value to set all values to</param>
+        public Matrix4x4(double value)
+        {
+            M11 = M21 = M31 = M41 = 
+            M12 = M22 = M32 = M42 = 
+            M13 = M23 = M33 = M43 = 
+            M14 = M24 = M34 = M44 = value;
+        }
+
+        public double M11;
+        public double M21;
+        public double M31;
+        public double M41;
+        public double M12;
+        public double M22;
+        public double M32;
+        public double M42;
+        public double M13;
+        public double M23;
+        public double M33;
+        public double M43;
+        public double M14;
+        public double M24;
+        public double M34;
+        public double M44;
+
+        public unsafe double this[int col, int row]
+        {
+            get
+            {
+                if (col < 0 || col >= ColumnCount)
+                    throw new ArgumentOutOfRangeException(nameof(col), $"Expected greater than or equal to 0 and less than {ColumnCount}, found {col}.");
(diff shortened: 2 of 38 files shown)


src/NuGet.Config
```diff
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="utf-8"?>
+﻿<?xml version="1.0" encoding="utf-8"?>
```
whoisj: Many `git push -f` later and I still do not know what I've done to edit this line. :confounded:

src/System.Numerics.Matrices/src/System/Numerics/Matrix5x5.cs
```diff
@@ -0,0 +1,543 @@
...
+
+        public unsafe override int GetHashCode()
+        {
+            fixed (Matrix5x5* p = &this)
+            {
+                int* x = (int*)p;
+                unchecked
+                {
+                    return 0xFFE1
+                         + 7 * ((((x[00] ^ x[01]) << 0) + ((x[02] ^ x[03]) << 1) + ((x[04] ^ x[05]) << 2) + ((x[06] ^ x[07]) << 3) + ((x[08] ^ x[09]) << 4)) << 0)
```
whoisj: This logic could be hokum, somebody with knowledge of hashing should take a closer look. This kind of logic has always worked finely for me in the past, but it very well could be over designed and/or a terrible algo.

src/System.Numerics.Matrices/src/System/Numerics/Matrix5x5.cs
```diff
@@ -0,0 +1,543 @@
...
+                         + 7 * ((((x[10] ^ x[11]) << 0) + ((x[12] ^ x[13]) << 1) + ((x[14] ^ x[15]) << 2) + ((x[16] ^ x[17]) << 3) + ((x[18] ^ x[19]) << 4)) << 2)
+                         + 7 * ((((x[15] ^ x[16]) << 0) + ((x[17] ^ x[18]) << 1) + ((x[19] ^ x[20]) << 2) + ((x[21] ^ x[22]) << 3) + ((x[23] ^ x[24]) << 4)) << 3)
+                         + 7 * ((((x[20] ^ x[21]) << 0) + ((x[22] ^ x[23]) << 1) + ((x[24] ^ x[25]) << 2) + ((x[26] ^ x[27]) << 3) + ((x[28] ^ x[29]) << 4)) << 4);
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            return "Matrix5x5: "
```
whoisj: I originally did this with `String.Format(...)` and having the concat strings was "free" as they should have been concatenated at compile time. With interpolated strings, does it mean a bunch of strings will get alloc'd then concatenated on each `ToString` invocation?

src/System.Numerics.Matrices/src/System/Numerics/MatrixTemplate.tt
```diff
@@ -0,0 +1,611 @@
...
+        public unsafe double this[int col, int row]
+        {
+            get
+            {
+                if (col < 0 || col >= ColumnCount)
+                    throw new ArgumentOutOfRangeException(nameof(col), $"Expected greater than or equal to 0 and less than {ColumnCount}, found {col}.");
+                if (row < 0 || row >= RowCount)
+                    throw new ArgumentOutOfRangeException(nameof(row), $"Expected greater than or equal to 0 and less than {RowCount}, found {row}.");
+
+                fixed (<#= structName #>* p = &this)
```
whoisj: Anyone know if this alignment, which works well on x86, will work on ARM and other platforms?

src/NuGet.Config
akoeplinger: `git diff` on the commandline is helpful for this:

image


Looks like you changed the BOM.

whoisj: @akoeplinger Thanks for that. I'm using Git for Windows, and git diff HEAD~..HEAD -- src/NuGet.Config wasn't showing the BOM difference. :unamused:

src/System.Numerics.Matrices/src/System/Numerics/Matrix4x4.cs
```diff
@@ -0,0 +1,435 @@
...
+                         + 7 * ((((x[12] ^ x[13]) << 0) + ((x[14] ^ x[15]) << 1) + ((x[16] ^ x[17]) << 2) + ((x[18] ^ x[19]) << 3)) << 3);
+                }
+            }
+        }
+
+        public override string ToString()
+        {
+            return "Matrix4x4: "
+                 + $"{{|{M11:00}|{M21:00}|{M31:00}|{M41:00}|}}"
+                 + $"{{|{M12:00}|{M22:00}|{M32:00}|{M42:00}|}}"
```
whoisj: I originally did this with String.Format(...) and having the concat strings was "free" as they should have been concatenated at compile time. With interpolated strings, does it mean a bunch of strings will get alloc'd then concatenated on each ToString invocation?

src/System.Numerics.Matrices/src/System/Numerics/Matrix4x4.cs
```diff
@@ -0,0 +1,435 @@
...
+        public unsafe override int GetHashCode()
+        {
+            fixed (Matrix4x4* p = &this)
+            {
+                int* x = (int*)p;
+                unchecked
+                {
+                    return 0xFFE1
+                         + 7 * ((((x[00] ^ x[01]) << 0) + ((x[02] ^ x[03]) << 1) + ((x[04] ^ x[05]) << 2) + ((x[06] ^ x[07]) << 3)) << 0)
+                         + 7 * ((((x[04] ^ x[05]) << 0) + ((x[06] ^ x[07]) << 1) + ((x[08] ^ x[09]) << 2) + ((x[10] ^ x[11]) << 3)) << 1)
```
whoisj: This logic could be hokum, somebody with knowledge of hashing should take a closer look. This kind of logic has always worked finely for me in the past, but it very well could be over designed and/or a terrible algo.

src/System.Numerics.Matrices/src/System/Numerics/Matrix4x4.cs
```diff
@@ -0,0 +1,435 @@
...
+            set
+            {
+                if (col < 0 || col >= ColumnCount)
+                    throw new ArgumentOutOfRangeException(nameof(col), $"Expected greater than or equal to 0 and less than {ColumnCount}, found {col}.");
+                if (row < 0 || row >= RowCount)
+                    throw new ArgumentOutOfRangeException(nameof(row), $"Expected greater than or equal to 0 and less than {RowCount}, found {row}.");
+
+                fixed (Matrix4x4* p = &this)
+                {
+                    double* d = (double*)p;
```
whoisj: Anyone know if this alignment, which works well on x86, will work on ARM and other platforms?

src/System.Numerics.Matrices/src/System/Numerics/Matrix4x4.cs
```diff
@@ -0,0 +1,435 @@
...
+                    d[row * ColumnCount + col] = value;
+                }
+            }
```
whoisj: Anyone know if this alignment, which works well on x86, will work on ARM and other platforms?

whoisj:
/CC @KrzysztofCwalina how is the style/formatting. I can address those issues now and then worry about technical issues as they come up (they always come up). :smirk:

KrzysztofCwalina:
@whoisj, could you give me some context for this PR? Why is it being added to corfxlab? What is there some writeup describing the feature?

JohanLarsson:
Mutable struct with equality feels scary. Why not immutable?

whoisj:
@KrzysztofCwalina it is here because you asked for it to be here (https://github.com/dotnet/corefx/pull/1846).

There's no write up, though I suppose given enough time I could find the time to do a write up. I mostly added this because in several previous project we've had need for matrices and always ended up rolling our own. The .NET library doesn't supply any kind of matrices and there's not really a great reason not to.

@JohanLarsson I don't understand how this is scary? If you touch the struct you'll get a copy anyways, so any edits would be on your local copy - unless it was passed by `ref` in which case it is expected that you might edit it. All other structs are editable.

KrzysztofCwalina:
@whoisj, ah, thanks for pointing me to the thread. It's been so long. The thread will provide good context.

src/System.Numerics.Matrices/src/System/Numerics/IMatrix.cs
```diff
@@ -0,0 +1,9 @@
+﻿namespace System.Numerics.Matrices
+{
+    interface IMatrix
+    {
+        double this[int col, int row] { get; set; }
```
KrzysztofCwalina: Why double? Would it make sense to make it IMatrix<T>?

whoisj: In my experience (mostly in the world of graphics) `double` is the most useful (though I could easily argue for `float` as well). Since there is no `<T> where T is intrinsic` or equivalent, I cannot  make these generic. Therefore `double` was the closest thing as it solves the widest range of issues.

 * supports positive and negative values
 * supports a very wide range of values with decent precision (thanks to 64-bitness)
 * supports infinity
 * supports NaN
 * has built in CPU and GPU support
 * etc

kcwalina: Yeah, I realize double is a reasonable choice given how our type system and CLR work today. I would be interested in exploring what would we need to do in the type system and the CLR to enable scenarios like such generic matrix type. 
We explored many solutions here in the past, but never pulled the trigger. Adding @CarolEidt who worked in this space int the past.

whoisj: There are a few good ideas in the Rosyln project bouncing around. My favorite is the ability to specify operate compliance on a type. Additionally, for optimizations there would be way to enforce that a type was not only a struct but a blittlable struct for optimized `unsafe` work on it.

Example:
```
public struct Matrix2x2<T>
    where T: struct, unsafe,
    operator +(T, T) => T,
    operator -(T, T) => T,
    operator *(T, T) => T,
    operator /(T, T) => T
{
    // ...
}
```

KrzysztofCwalina: Anyway, I will merge the PR. The goal of corefx lab projects IMO is to start with what's possible to day, force discussions on limitations, and then possibly try to fix solutions for these limitations.

PR #119, opened 2015-07-07
Graphics library

dalbanhi:
@axxu @t-roblo and others would appreciate feedback on this preliminary structure for main class


axxu:
:100:

PR #120, opened 2015-07-07
Added Image Extensions Class, refactored most tests from before repo …

dalbanhi:
Added Image Extensions Class (and factory methods)

Refactored existing tests to XUnit format (TODO: add more tests) from before repo switch

@axxu @t-roblo would appreciate a quick review


t-roblo:
NICE

PR #122, opened 2015-07-08
Graphics library rj

t-roblo:
Created a Wrapper "gdImageFile" for writing image to a file-path.


src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -76,6 +76,11 @@ private Image(Stream stream)
             throw new NotImplementedException();
         }
 
+            /* Manipulations */
+        public Image Write(string im, string dst)
```
KrzysztofCwalina: these parameter names don't follow our naming conventions. They need to not use abbreviations.

KrzysztofCwalina: Why does it take two strings? I thought we would have extension methods over Image type to manipulate images.

PR #126, opened 2015-07-09
Started working on Image

axxu:


src/System.Drawing.Graphics/src/System/Drawing/Graphics/Image.cs
```diff
@@ -19,13 +19,16 @@ public enum PixelFormat
...
-            /* Private Fields */
+        /* Private Fields */
+        ////do we even need any of this anymore??
         private byte[] _data = null;
         private int _width = 0;
         private int _height = 0;
         private PixelFormat _pixelFormat = PixelFormat.ARGB;
         private int _bytesPerPixel = 0;
 
+        private DLLImports.gdImageStruct _gdImgStruct;
```
KrzysztofCwalina: What's the relationship between _data and _gdImgStruct? We don't need both, do we? 
Also, _gdImgStruct is unsafe, and so we should not make it public, and so transitively the extension methods (loaders and image transformation routines) won't be able to use it.

PR #127, opened 2015-07-10
Implemented Resize and Create

dalbanhi:
Implemented Resize
   Tests work for empty images
Implemented Creating an empty image
 Tests pass for correct parameters

Would appreciate a code review, @axxu @t-roblo @KrzysztofCwalina especially in unsafe portions in Resize


axxu:
I need your build to fix my problems

PR #128, opened 2015-07-10
Add System.Time assembly

mj1856:
Per prior discussion (see dotnet/corefx#2116 & dotnet/corefx#700), this PR migrates the prototype project from [mj1856/corefx-dateandtime](https://github.com/mj1856/corefx-dateandtime) into corefxlab.

### DIFF
--- a/src/System.Time/src/System/Date.cs
+++ b/src/System.Time/src/System/Date.cs
@@ -1,0 +1,1725 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Diagnostics.Contracts;
+using System.Globalization;
+using System.Text.RegularExpressions;
+using System.Xml;
+using System.Xml.Schema;
+using System.Xml.Serialization;
+
+namespace System
+{
+    /// <summary>
+    /// Represents a whole date, having a year, month and day component.
+    /// All values are in the proleptic Gregorian (ISO 8601) calendar system unless otherwise specified.
+    /// </summary>
+    [DebuggerDisplay("{ToString()}")]
+    [XmlSchemaProvider("GetSchema")]
+    public struct Date : IEquatable<Date>, IComparable<Date>, IComparable, IFormattable, IXmlSerializable
+    {
+        private const int MinDayNumber = 0;
+        private const int MaxDayNumber = 3652058;
+
+        /// <summary>
+        /// Represents the smallest possible value of <see cref="Date"/>. This field is read-only.
+        /// </summary>
+        public static readonly Date MinValue = new Date(MinDayNumber);
+
+        /// <summary>
+        /// Represents the largest possible value of <see cref="Date"/>. This field is read-only.
+        /// </summary>
+        public static readonly Date MaxValue = new Date(MaxDayNumber);
+
+        // Number of days in a non-leap year
+        private const int DaysPerYear = 365;
+
+        // Number of days in 4 years
+        private const int DaysPer4Years = DaysPerYear * 4 + 1;       // 1461
+
+        // Number of days in 100 years
+        private const int DaysPer100Years = DaysPer4Years * 25 - 1;  // 36524
+
+        // Number of days in 400 years
+        private const int DaysPer400Years = DaysPer100Years * 4 + 1; // 146097
+
+        // The following arrays contain the starting day-of-year number of each month, for regular and leap years
+        private static readonly int[] DaysToMonth365 = { 0, 31, 59, 90, 120, 151, 181, 212, 243, 273, 304, 334, 365 };
+        private static readonly int[] DaysToMonth366 = { 0, 31, 60, 91, 121, 152, 182, 213, 244, 274, 305, 335, 366 };
+
+        // internal enum
+        private const int DatePartYear = 0;
+        private const int DatePartDayOfYear = 1;
+        private const int DatePartMonth = 2;
+        private const int DatePartDay = 3;
+
+        // Number of whole days since 0001-01-01 (which is day 0)
+        // NOTE: This is the only field in this structure.
+        private readonly int _dayNumber;
+
+        /// <summary>
+        /// Initializes a new instance of a <see cref="Date"/> structure to a specified number of days.
+        /// </summary>
+        /// <param name="dayNumber">The number of days since January 1, 0001 in the proleptic Gregorian calendar.</param>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// <paramref name="dayNumber"/> is out of the range supported by the <see cref="Date"/> object.
+        /// </exception>
+        public Date(int dayNumber)
+        {
+            if (dayNumber < MinDayNumber || dayNumber > MaxDayNumber)
+            {
+                throw new ArgumentOutOfRangeException("dayNumber");
+            }
+
+            Contract.EndContractBlock();
+
+            _dayNumber = dayNumber;
+        }
+
+        /// <summary>
+        /// Initializes a new instance of a <see cref="Date"/> structure to a specified year, month, and day.
+        /// </summary>
+        /// <param name="year">The year (1 through 9999).</param>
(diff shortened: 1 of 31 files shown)


whoisj:
Very much a step in the right direction. Thanks. Still a little disappointing that there's no IDate or ITimeOfDay interfaces for those who would like to implement their own versions for whatever reason, yet remain compatible.

mj1856:
@whoisj - I don't think interfaces would help in this particular case.  They might even make things worse, as one couldn't depend on any particular set of behavior without knowing about the concrete implementation.

Using DDD terminology, interfaces work best on *services*, not on *value objects*

Consider that `IComparable` is an interface which says "you can compare me to another instance".  Same for `IEquatable`, `ISerializable`, etc.  It would be a lot less meaningful to have a `IDate`, `ITime`, etc.  It would be like if `int`, `long`, `double`, etc. all had an `INumber` interface.

We could consider adding some interfaces where they make sense - such as `IClock` for getting the current time, etc.  But that's a topic for a separate issue.  For now, let's focus on getting this one merged.  There are already changes and additions I know need to be made before it is complete.

whoisj:
@mj1856 I suppose, though I see the current `DateTime` and new `Date` and `Time` as mixed simple-types and services. The simple-type here is the "ticks since 0". All of the cultural translation on top is a service. Just my PoV.

src/System.Time/src/System/Date.cs
```diff
@@ -0,0 +1,1629 @@
...
+
+using System.Diagnostics;
+using System.Diagnostics.Contracts;
+using System.Globalization;
+using System.Text.RegularExpressions;
+using System.Xml;
+using System.Xml.Schema;
+using System.Xml.Serialization;
+
+namespace System
```
kcwalina: We can decide later on this (i.e. don't block the initial commit), but I wonder if System.Time would not be a better namespace. One thing we want to avoid is confusing existing developers who are happy with DateTimeOffset.

src/System.Time/src/System/Date.cs
```diff
@@ -0,0 +1,1629 @@
...
+        /// upon whether <paramref name="year"/> is a leap year.
+        /// </returns>
+        /// <param name="year">The year.</param>
+        /// <param name="month">The month (a number ranging from 1 to 12).</param>
+        /// <exception cref="ArgumentOutOfRangeException">
+        /// <paramref name="month"/> is less than 1 or greater than 12.
+        /// <para>-or-</para>
+        /// <paramref name="year"/> is less than 1 or greater than 9999.
+        /// </exception>
+        public static int DaysInMonth(int year, int month)
```
kcwalina: Is this a property of Day or rather of a calendar?

src/System.Time/src/System/Date.cs
```diff
@@ -0,0 +1,1629 @@
...
+        ///   <term>Zero</term>
+        ///   <term><paramref name="d1"/> is the same as <paramref name="d2"/>.</term>
+        /// </item>
+        /// <item>
+        ///   <term>Greater than zero</term>
+        ///   <term><paramref name="d1"/> is later than <paramref name="d2"/>.</term>
+        /// </item>
+        /// </list>
+        /// </returns>
+        public static int Compare(Date d1, Date d2)
```
kcwalina: We call parameters of comparison operations "left" and "right"

src/System.Time/src/System/Date.cs
```diff
@@ -0,0 +1,1629 @@
...
+        public static bool operator !=(Date d1, Date d2)
+        {
+            return !d1.Equals(d2);
+        }
+
+        /// <summary>
+        /// Determines whether one specified <see cref="Date"/> is later than another specified <see cref="Date"/>.
+        /// </summary>
+        /// <param name="d1">The first object to compare.</param>
+        /// <param name="d2">The second object to c
```
kcwalina: this should use curly braces.

kcwalina: this should use curly braces.

src/System.Time/src/System/DateTimeEx.cs
```diff
@@ -0,0 +1,160 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System
+{
+    /// <summary>
+    /// Extension methods for <see cref="DateTime"/>.
+    /// </summary>
+    public static class DateTimeEx
```
kcwalina: "Ex" is not a suffix we use in .NET APIs.

src/System.Time/src/System/Date.cs
mj1856: These types complement the existing types.  They don't replace them.  IMHO, it wouldn't make sense to find them in a new namespace unless there were a complete set of new types there.  We should discuss further.

src/System.Time/src/System/Date.cs
mj1856: Yes, and exists as `System.Globalization.Calendar.GetDaysInMonth`.  However there is also `System.DateTime.DaysInMonth`, and since other properties and methods of `Date` are Gregorian-centric (such as `IsLeapYear`, then I think it's ok to have this one here as well.

src/System.Time/src/System/Date.cs
mj1856: Acknowledged.  I was matching conventions of `System.DateTime` but I will update.  Should those be changed as well?

src/System.Time/src/System/DateTimeEx.cs
mj1856: What should I call a class then that provides extension methods to the types not in this assembly?  Is `DateTimeExtensions` acceptable?

KrzysztofCwalina: Yeah, that sounds good.

src/System.Time/src/System/Date.cs
KrzysztofCwalina: Now that I read you see these types as complementary to the old APIs, I think we could go either way. I slightly prefer new names, but it's your call.

KrzysztofCwalina:
BTW, matt, I am leaving for vacation starting Friday. it would be great to merge the change before then.

mj1856:
@KrzysztofCwalina - done.  LMK if you want it rebased.

KrzysztofCwalina:
OK. Looks good. Please squash and I will merge.
BTW, at some point we should close on the issue whether this library can be independent of DateTime and possibly DateTimeOffset, or it's complementary. I honestly think DateTime has so many problems that I would love to try to do something that can be "standalone" and for the transition period add a set of bridge APIs that will let people move between the APIs. 
I imagine such standalone library would be UTC based a its core, and things like wall clock time, days, calendars, etc. would built on top.

mj1856:
Squashed.

mj1856:
WRT the last point, striking a good balance is the difficult part.  If we introduce entirely new APIs, then how do we make it easy to migrate to those without going through the existing ones?  This is the problem that Noda Time has today.  All of the historical integration points (such as in `System.Data`) rely on hard-coded use of `DateTime`.  For example, one can't get an `Instant` or a `LocalDate` directly from a database without hitting `DateTime` first.

The bridge API approach may be worth exploring further.  The interface-based approach may also be worthwhile, if we can define the interfaces in such a way that are meaningful, but only if we can get dependencies to use the interfaces instead of the concrete types.  That may be considered a breaking change for some.

Also, keep in mind that having UTC at the core only works well for tracking instantaneous points in time, not calendrical measures.  That's the mistake that `java.Util.Date` and JavaScript's `Date` class have made.   Specifically consider that the two types being added here are `Date` and `TimeOfDay` - neither of which make sense to be UTC based.

We should continue to discuss, though I'm not sure this format is the most productive for such a high-level discussion.

whoisj:
> For example, one can't get an Instant or a LocalDate directly from a database without hitting DateTime first.

I disagree. In most enterprise setting I've worked in where we stored time, we usually stored it as either a string (terrible) or as an integer (number of seconds since X). In both cases, the `DateTime` type is useless without custom conversion logic.

I do mostly agree with the rest of the statement. :smile:

PR #135, opened 2015-07-14
Dalbanhi master

dalbanhi:


KrzysztofCwalina:
BTW, you should squash commits before merging.

PR #140, opened 2015-07-15
Checked some tests, added TrueColor internal field ,transparency can …

dalbanhi:
…be changed TODO: Stream I/O, add test images and exception strings to resources


t-roblo:
Nice!

PR #142, opened 2015-07-17
Alpha Blending and Refactored Classes

axxu:
Refactored classes according to Krzysztof's original design for more extensibility.
Figured out how to effectively use alpha blending and the alpha saving flag.


dalbanhi:
Nice

PR #143, opened 2015-07-17
Changed bitmap back to image

axxu:
Because there is a file type called bitmap and this will get confusing later on


t-roblo:
Thank You!

PR #148, opened 2015-07-23
Starting Linux

axxu:
- added Linux binaries for LibGD
- created a temporary second copy of the imports using the new binaries (will replace with partial classes later)
- the Create method's basic tests are passing.
- added directive if/elses to Image to separate Windows and Linux versions (Windows temporary doesn't work because I hardcoded a directive in)


dalbanhi:
Great, ready to start

PR #154, opened 2015-07-29
FIxed Resizing PNG bug and added more tests

dalbanhi:
TODO: Organize tests, change csproj files to choose OS, embed resources to not have magic numbers in tests


dalbanhi:
Nice

PR #156, opened 2015-08-03
Initial commit of the System.Reflection.Metadata.Cil

SantiFdezM:
This introduces a library to disassemble a .NET assembly


src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+        #region Public APIs
+
+        /// <summary>
+        /// Method that given a token defines if it is a type reference token.
+        /// </summary>
+        /// <param name="token">token to solve</param>
+        /// <returns>true if is a type reference false if not</returns>
+        public static bool IsTypeReference(int token)
+        {
+            return (token >> 24) == 0x01;
```
stephentoub: The ```token >> 24``` shifting occurs a fair amount.  Should that be factored out into some private method that gives this an appropriate name?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+        }
+
+        /// <summary>
+        /// Method that given a token defines if it is a type definition token.
+        /// </summary>
+        /// <param name="token">token to solve</param>
+        /// <returns>true if is a type definition false if not</returns>
+        public static bool IsTypeDefinition(int token)
+        {
+            return (token >> 24) == 0x02;
```
stephentoub: Are these 0x02, 0x70, etc. values that should be in named constants somewhere?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+        /// Method that given a token defines if it is a standalone signature token.
+        /// </summary>
+        /// <param name="token">token to solve</param>
+        /// <returns>true if is a standalone signature false if not</returns>
+        public static bool IsStandaloneSignature(int token)
+        {
+            return (token >> 24) == 0x11;
+        }
+
+        public static string GetByteArrayString(byte[] array)
```
stephentoub: Is the specific format being generated here what's actually required? If any such format could work, callers could just use BitConverter.ToString(array) rather than this method being exposed.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+            for (int i = 0; i < array.Length; i++)
+            {
+                if (i == 0)
+                {
+                    sb.Append("( ");
+                }
+                sb.Append(array[i].ToString("X2"));
+                sb.Append(" ");
+            }
+            sb.Append(")");
```
stephentoub: If the array length is 0, the string will end up being just ")".  I assume that's not desired?

stephentoub: Assuming you do need to keep this method, there are additional optimizations that can be applied.  For example, you know there won't be more than 2 + 3*array.Length characters in the resulting string, so you could initialize the StringBuilder with that size.  You could also just use a char[] created with the right size and fill in each character, rather than needing a StringBuilder.  Plus, that'll help avoid all of the intermediate garbage strings getting generated for each ToString call.  There's a very similar example of that here: https://github.com/dotnet/corefx/blob/6504501399c6f9ea4d48296058e7e9a57e41ca21/src/System.Security.Cryptography.X509Certificates/src/Internal/Cryptography/Helpers.cs#L18-L28

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+        }
+
+        internal static ILLocal[] DecodeLocalSignature(MethodBodyBlock methodBody, MetadataReader mdReader, ILTypeProvider provider)
+        {
+            if (methodBody.LocalSignature.IsNil)
+            {
+                return new ILLocal[0];
+            }
+            var localTypes = SignatureDecoder.DecodeLocalSignature(methodBody.LocalSignature, provider);
+            ILLocal[] locals = new ILLocal[localTypes.Count()];
```
stephentoub: ```localTypes.Length``` instead of ```localTypes.Count()``` ?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+            }
+        }
+
+        internal static ILLocal[] DecodeLocalSignature(MethodBodyBlock methodBody, MetadataReader mdReader, ILTypeProvider provider)
+        {
+            if (methodBody.LocalSignature.IsNil)
+            {
+                return new ILLocal[0];
+            }
+            var localTypes = SignatureDecoder.DecodeLocalSignature(methodBody.LocalSignature, provider);
```
stephentoub: We should try to avoid ```var``` when the right-side of the assignment doesn't make it very clear what the type is.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+                ilOffset += instruction.Size;
+                yield return instruction;
+            }
+        }
+
+        internal static ILLocal[] DecodeLocalSignature(MethodBodyBlock methodBody, MetadataReader mdReader, ILTypeProvider provider)
+        {
+            if (methodBody.LocalSignature.IsNil)
+            {
+                return new ILLocal[0];
```
stephentoub: ```Array.Empty<ILLocal>()``` ?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+            }
+            var localTypes = SignatureDecoder.DecodeLocalSignature(methodBody.LocalSignature, provider);
+            ILLocal[] locals = new ILLocal[localTypes.Count()];
+            for (int i = 0; i < localTypes.Length; i++)
+            {
+                string name = "V_" + i;
+                locals[i] = new ILLocal(name, localTypes[i].ToString());
```
stephentoub: Looks like we're throwing away some information by converting the type to a string.  We don't want ILLocal to store the string name and the strongly-typed type?  Or we can't for some reason?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+            return locals;
+        }
+
+        internal static ILParameter[] DecodeParameters(MethodSignature<ILType> signature, ParameterHandleCollection parameters, ref Readers readers)
+        {
+            var types = signature.ParameterTypes;
+            int requiredCount = Math.Min(signature.RequiredParameterCount, types.Length);
+            if (requiredCount == 0)
+            {
+                return new ILParameter[0];
```
stephentoub: ```Array.Empty<ILParameter>()``` ?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+            }
+            ILParameter[] result = new ILParameter[requiredCount];
+            for (int i = 0; i < requiredCount; i++)
+            {
+                var parameter = readers.MdReader.GetParameter(parameters.ElementAt(i));
+                result[i] = ILParameter.Create(parameter, ref readers, types[i].ToString());
```
stephentoub: I'll stop asking about it, but same question regarding why we're passing a string version of the type rather than the type itself.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+            {
+                var handle = MetadataTokens.StandaloneSignatureHandle(intOperand);
+                var standaloneSignature = mdReader.GetStandaloneSignature(handle);
+                var signature = SignatureDecoder.DecodeMethodSignature(standaloneSignature.Signature, provider);
+                return string.Format("{0}{1}", GetMethodReturnType(signature), provider.GetParameterList(signature));
+            }
+            throw new ArgumentException("Get signature invalid token");
+        }
+
+        private static string GetVariableName(OpCode opCode,int token, ILMethodDefinition _methodDefinition)
```
stephentoub: Nit: space before ```int token```

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+                }
+                return _methodDefinition.GetParameter(token).Name;
+            }
+            return _methodDefinition.GetLocal(token).Name;
+            
+        }
+
+        private static string GetInlineTokenType(MetadataReader mdReader, int intOperand, ILTypeProvider provider)
+        {
+            if(IsMethodDefinition(intOperand) || IsMethodSpecification(intOperand) || IsMemberReference(intOperand))
```
stephentoub: Nit: space before the opening paren (missing a few times below as well)

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+
+        private static ILInstruction CreateSwitchInstruction(ref BlobReader ilReader, int expectedSize, int ilOffset, OpCode opCode)
+        {
+            var caseNumber = ilReader.ReadUInt32();
+            int[] jumps = new int[caseNumber];
+            for(int i = 0; i < caseNumber; i++)
+            {
+                jumps[i] = ilReader.ReadInt32();
+            }
+            int size = 4 + expectedSize;
```
stephentoub: It'd be good to put values like the ```4```s here into named constants.  As a reader not deeply familiar with this space, I'm not sure what these values mean and why they're correct.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+                var typeHandle = MetadataTokens.TypeSpecificationHandle(intOperand);
+                return SignatureDecoder.DecodeType(typeHandle, provider).ToString();
+            }
+            var defHandle = MetadataTokens.TypeDefinitionHandle(intOperand);
+            return SignatureDecoder.DecodeType(defHandle, provider).ToString();
+        }
+
+        private static ILInstruction CreateSwitchInstruction(ref BlobReader ilReader, int expectedSize, int ilOffset, OpCode opCode)
+        {
+            var caseNumber = ilReader.ReadUInt32();
```
stephentoub: Nit: fwiw, even though the right-hand side does say UInt32, uint is only one more character than var, and as such var isn't saving much and I find the code easier to understand with uint.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+        private static string GetArgumentString(MetadataReader mdReader, int intOperand, out bool isPrintable)
+        {
+            if (IsUserString(intOperand))
+            {
+                UserStringHandle usrStr = MetadataTokens.UserStringHandle(intOperand);
+                var str = mdReader.GetUserString(usrStr);
+                str = ProcessAndNromalizeString(str, out isPrintable);
+                return str;
+            }
+            throw new NotImplementedException("Argument String");
```
stephentoub: Is this something that's still TODO or we don't plan to implement this?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+        }
+
+        private static string ProcessAndNromalizeString(string str, out bool isPrintable)
```
stephentoub: Nit: ```Nromalize``` => ```Normalize```

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+                        sb.Append(_byte.ToString("X2"));
+                        sb.Append(" ");
+                    }
+
+                    isPrintable = false;
+                    break;
+                }
+            }
+
+            sb.Append(")");
```
stephentoub: It looks like this could be moved to after the subsequent isPrintable check, or into the loop above before isPrintable is set to false.  And if you do the latter, the creation of the StringBuilder could also move into the loop where it's filled in.  Then you could structure it like:
```C#
foreach (char c in str)
{
    UnicodeCategory category = ...;
    if (category == ...)
    {
        isPrintable = false;
        return SameLogicAsDiscussedEarlierForConvertingCharsToHexString(str);
    }
}
isPrintable = true;
return str;
```

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+
+        private static string GetMethodReturnType(MethodSignature<ILType> signature)
+        {
+            StringBuilder sb = new StringBuilder();
+            if (signature.Header.IsInstance)
+            {
+                sb.Append("instance ");
+            }
+            sb.Append(signature.ReturnType.ToString());
+            return sb.ToString();
```
stephentoub: A StringBuilder isn't really needed here.  You could avoid the allocation with:
```C#
return signature.Header.IsInstance ?
    "instance" + signature.ReturnType.ToString() :
    signature.ReturnType.ToString();
```

justinvp:
Are we really calling this "ILDasm"? I realize there's some history with ildasm.exe, but wouldn't it be better to spell it out, e.g. something like "ILDisassembler"?

 - `System.Reflection.Metadata.ILDasm`

vs. some alternatives:

 - `System.Reflection.Metadata.ILDisassembler`
 - `System.Reflection.Metadata.Disassembler`
 - `System.Reflection.Metadata.IL`
 - `System.Reflection.Metadata.IL.Disassembler`
 - `System.Reflection.Metadata.IntermediateLanguage`
 - `System.Reflection.Metadata.IntermediateLanguageDisassembler`
 - `System.Reflection.Metadata.IntermediateLanguage.Disassembler`

I can think of variants of `Disassembler` as well, e.g.
 - `Disassembly`
 - `Disassemble`

cc @terrajobst @KrzysztofCwalina

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,582 @@
...
+using System.Globalization;
+using System.Linq;
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Decoding;
+using System.Reflection.Metadata.Ecma335;
+using System.Text;
+
+namespace ILDasmLibrary.Decoder
```
justinvp: Shouldn't the namespace match the assembly, e.g. `System.Reflection.Metatdata.ILDasm.Decoder`? This applies throughout. (As an aside, I left an overall comment about the use of "ILDasm").

akoeplinger:
@justinvp IIRC, the namespace was also discussed at the [API review](https://github.com/dotnet/apireviews/tree/master/2015-07-08-ildasm-and-drawing#ildasm), about 12minutes into the video.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
SantiFdezM: This would make it really complex, would have to resolve tokens, but it is possible. Going to file an issue on this to fix it later.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
SantiFdezM: Same as on ILLocal type.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
SantiFdezM: This method should only be called with UserString token types, so I'm going to change that to InvalidArgumentException.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
SantiFdezM: Name will be System.Reflection.Metadata.CIL
Decided today in a Design Review that will be uploaded later.

And I'm updating all Namespaces to match that name.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
SantiFdezM: Need to update to VS RTM, will fix it in a little while.

SizeOf/SizeOf/Program.cs
```diff
@@ -0,0 +1,18 @@
+﻿using System;
```
KrzysztofCwalina: The files should have the standard license comments. See other files in this repo.

SizeOf/SizeOf/SizeOf.csproj
```diff
@@ -0,0 +1,59 @@
...
+    <OutputPath>bin\Release\</OutputPath>
+    <DefineConstants>TRACE</DefineConstants>
+    <ErrorReport>prompt</ErrorReport>
+    <WarningLevel>4</WarningLevel>
+  </PropertyGroup>
+  <ItemGroup>
+    <Reference Include="System" />
+    <Reference Include="System.Core" />
+    <Reference Include="System.Xml.Linq" />
+    <Reference Include="System.Data.DataSetExtensions" />
```
KrzysztofCwalina: Could you remove references to unused assemblies?

src/System.Reflection.Metadata.ILDasm/src/Decoder/CILInstructionSize.cs
```diff
@@ -0,0 +1,12 @@
+﻿namespace System.Reflection.Metadata.ILDasm.Decoder
+{
+    public enum CILInstructionSize
```
KrzysztofCwalina: CIL should follow acronym casing guidelines, i.e. "Cil"

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+using System.Reflection;
+using System.Reflection.Emit;
+using System.Reflection.Metadata;
+using System.Reflection.Metadata.Decoding;
+using System.Reflection.Metadata.Ecma335;
+using System.Text;
+
+namespace System.Reflection.Metadata.ILDasm.Decoder
+{
+    public struct ILDecoder
```
KrzysztofCwalina: Shoudl it be called CilDecoder?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+        /// Method that given a token defines if it is a standalone signature token.
+        /// </summary>
+        /// <param name="token">token to solve</param>
+        /// <returns>true if is a standalone signature false if not</returns>
+        public static bool IsStandaloneSignature(int token)
+        {
+            return GetTokenType(token) == CILTokenType.Signature;
+        }
+
+        public static string GetByteArrayString(byte[] array)
```
KrzysztofCwalina: I would call these methods "CreateXxx" not "GetXxx"

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
KrzysztofCwalina: Should there be a Create overload that does not require calling ToString on the third parameter?

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+                        break;
+                    default:
+                        break;
+                }
+                ilOffset += instruction.Size;
+                yield return instruction;
+            }
+        }
+
+        internal static ILLocal[] DecodeLocalSignature(MethodBodyBlock methodBody, MetadataReader mdReader, ILTypeProvider provider)
```
KrzysztofCwalina: we don't use prefixes like "md".

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+            {
+                var handle = MetadataTokens.StandaloneSignatureHandle(intOperand);
+                var standaloneSignature = mdReader.GetStandaloneSignature(handle);
+                var signature = SignatureDecoder.DecodeMethodSignature(standaloneSignature.Signature, provider);
+                return string.Format("{0}{1}", GetMethodReturnType(signature), provider.GetParameterList(signature));
+            }
+            throw new ArgumentException("Get signature invalid token");
+        }
+
+        private static string GetVariableName(OpCode opCode, int token, ILMethodDefinition _methodDefinition)
```
KrzysztofCwalina: we don't use underscores in parameter names.

src/System.Reflection.Metadata.ILDasm/src/ILConstant.cs
```diff
@@ -0,0 +1,152 @@
...
+        }
+
+        public string GetValueString()
+        {
+            BlobReader reader = _readers.MdReader.GetBlobReader(_constant.Value);
+            object value = Value;
+            switch (TypeCode)
+            {
+                case ConstantTypeCode.Byte:
+                    return string.Format("uint8(0x{0})", ((byte)value).ToString("X2"));
```
KrzysztofCwalina: You should do Format("uint8(0x{0:X2})", (byte)value);
This will do less allocations.

src/System.Reflection.Metadata.ILDasm/src/ILEventDefinition.cs
```diff
@@ -0,0 +1,174 @@
...
+        {
+            visitor.Visit(this);
+        }
+
+        private IEnumerable<ILCustomAttribute> GetCustomAttributes()
+        {
+            foreach(var handle in _eventDefinition.GetCustomAttributes())
+            {
+                var attribute = _readers.MdReader.GetCustomAttribute(handle);
+                yield return new ILCustomAttribute(attribute, ref _readers);
```
KrzysztofCwalina: We don't use yield in framework code. It's inefficient and sometimes unpredictable.

src/System.Reflection.Metadata.ILDasm/src/ILExceptionRegion.cs
```diff
@@ -0,0 +1,113 @@
...
+            EndOffset = endOffset;
+        }
+
+        /// <summary>
+        /// This method is used to sort regions from outter to inner (smaller offsets first) so that the regions are dumped
+        /// with the desired format on msil in case we've got nested regions.
+        /// </summary>
+        /// <param name="span2">Region to compare to</param>
+        /// <returns>This method returns -1 if the first is smaller, 0 if they are equal, 1 if the first is greater.</returns>
+        public int CompareTo(ILExceptionRegion span2)
```
KrzysztofCwalina: We call comparison parameters "other" (if one) or "left" and "right" if two.

src/System.Reflection.Metadata.ILDasm/src/Visitor/IVisitable.cs
```diff
@@ -0,0 +1,13 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace ILDasmLibrary.Visitor
+{
+    public interface IVisitable
```
KrzysztofCwalina: I really don't like the IVisitor and IVisitable names. They are too generic.

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+        public static string GetVersionString(Version version)
+        {
+            int build = version.Build;
+            int revision = version.Revision;
+            if (build == -1) build = 0;
+            if (revision == -1) revision = 0;
+            return String.Format("{0}:{1}:{2}:{3}", version.Major.ToString(), version.Minor.ToString(), build.ToString(), revision.ToString());
+        }
+
+        public static string DecodeSignatureParamerTypes(MethodSignature<ILType> Signature)
```
justinvp: The `Signature` parameter should be `signature` (lowercase).

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+        {
+            StringBuilder sb = new StringBuilder();
+            sb.Append("(");
```
justinvp: `"("` -> `'('`

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+            var types = Signature.ParameterTypes;
+            for (int i = 0; i < types.Length; i++)
+            {
+                if (i > 0)
+                {
+                    sb.Append(",");
```
justinvp: `","` -> `','`

src/System.Reflection.Metadata.ILDasm/src/Decoder/ILDecoder.cs
```diff
@@ -0,0 +1,583 @@
...
+                }
+                sb.Append(types[i].ToString());
+            }
+            sb.Append(")");
```
justinvp: `")"` -> `')'`

AlexGhiondea:
LGTM

PR #157, opened 2015-08-05
Refactored to match performance benchmarks, ready to start stress tes…

axxu:
…ting performance.

:dancer:


dalbanhi:
I think the Brackets might not be lined up in ImageExtensions class, we should double check that

PR #163, opened 2015-08-06
Changed Matirices project to not use C# 6.0 features

KrzysztofCwalina:
This is to fix a build break after updating to a new engineering system.
@whoisj, let me know if you are happy with the changes. At some point we might allow C# 6.0 features and we can change the templates back.


whoisj:
Looks like we did the same thing. LGTM. I'll let you choose who's PR gets submitted.

You were good enough to do a `StringBuilder` where I was just allowing concat + allocation. :+1:

PR #181, opened 2015-08-13
Downgrade immutable collections to version 36

AlexGhiondea:
This fixes the build break where the immutable collection package version could not be resolved.


tarekgh:
LGTM

PR #188, opened 2015-08-19
Added image filters

axxu:
Added image filters for our stretch goal
-iterate through image for each pixel and manipulate using matrix multiplication
-added some default filter matrices (greyscale, sepia, negative)

standing issue: is math for changing alpha value backwards or just counterintuitive?


dalbanhi:
LGTM

PR #189, opened 2015-08-19
Added perf Images to get rid of hard coded file paths 

dalbanhi:
(No need for interop in testing project and refactored tests to skip perf tests unless necessary)


axxu:
nice good job!
:clap: :clap: :clap:

Issue #172, opened 2015-08-08
Dispose method on CilAssembly needs to be updated on System.Reflection.Metadata.Cil


SantiFdezM:
The dispose method on CilAssembly needs to be updated to keep track of disposed state in order to avoid disposing the readers twice and get an exception.

See:

https://github.com/dotnet/corefxlab/blob/master/src/System.Reflection.Metadata.Cil/src/CilAssembly.cs#L300-L303

cc @AlexGhiondea

PR #190, opened 2015-08-20
Fix issue #172 in System.Reflection.Metadata.Cil

SantiFdezM:
This fixes dotnet/corefxlab#172 so that CilAssembly can't be disposed twice in order to avoid exceptions.

This was done keeping track of the current dispose status.

cc @AlexGhiondea


KrzysztofCwalina:
looks good.

PR #211, opened 2015-09-02
Adding support for building dlls

ahsonkhan:
@KrzysztofCwalina can you please review?

### DIFF
--- a/demos/dotnet/dotnet/ProjectProperties.cs
+++ b/demos/dotnet/dotnet/ProjectProperties.cs
@@ -9,6 +9,7 @@ namespace Microsoft.Core
     {
         public string AssemblyName;
         public string OutputDirectory;
+        public string OutputType;
 
         public readonly List<string> Sources = new List<string>();
         public readonly List<string> References = new List<string>();
@@ -18,7 +19,7 @@ namespace Microsoft.Core
 
         public string OutputAssemblyPath
         {
-            get { return Path.Combine(OutputDirectory, AssemblyName + ".exe"); }
+            get { return Path.Combine(OutputDirectory, AssemblyName + OutputType); }
         }
 
         internal string ProjectDirectory;
--- a/demos/dotnet/dotnet/ProjectProperties.cs
+++ b/demos/dotnet/dotnet/ProjectProperties.cs
@@ -19,7 +19,12 @@ namespace Microsoft.Core
 
         public string OutputAssemblyPath
         {
-            get { return Path.Combine(OutputDirectory, AssemblyName + OutputType); }
+            get { return Path.Combine(OutputDirectory, AssemblyNameAndExtension); }
+        }
+
+        public string AssemblyNameAndExtension
+        {
+            get { return AssemblyName + OutputType; }
         }
 
         internal string ProjectDirectory;
--- a/demos/dotnet/dotnet/dotnet.cs
+++ b/demos/dotnet/dotnet/dotnet.cs
@@ -43,6 +43,10 @@ static class Program
                 Build(args, Log);
                 break;
 
+            case "/lib":
+                Build(args, Log, true);
+                break;
+
             case "/clean":
                 Clean(args, Log);
                 break;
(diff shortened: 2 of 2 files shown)


demos/dotnet/dotnet/dotnet.cs
```diff
@@ -43,6 +43,10 @@ static void Main(string[] args)
                 Build(args, Log);
                 break;
 
+            case "/lib":
```
KrzysztofCwalina: this should be /target:library. /lib csc.exe option is to link additional libraries.

demos/dotnet/dotnet/ProjectProperties.cs
```diff
@@ -9,6 +9,7 @@ class ProjectProperties
     {
         public string AssemblyName;
         public string OutputDirectory;
+        public string OutputType;
```
KrzysztofCwalina: Should it be called "OutputExtension"? Your call.

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -59,6 +63,7 @@ static void PrintUsage()
     {
         string appName = Environment.GetCommandLineArgs()[0];
         Console.WriteLine("{0} [/log] - compiles sources in current direcotry. optionally logs diagnostics info.", appName);
+        Console.WriteLine("{0} /lib - compiles sources in current directory ito dll.", appName);
```
KrzysztofCwalina: "into" is misspelled.

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -78,9 +83,9 @@ static void Clean(string[] args, Log log)
         Directory.Delete(properties.PackagesDirectory, true);
     }
 
-    static void Build(string[] args, Log log)
+    static void Build(string[] args, Log log, bool buildLib = false)
```
KrzysztofCwalina: The parameter should be called "buildLibrary" or "buildDll". We try to avoid abbreviations in identifiers.

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -106,11 +111,14 @@ static void Build(string[] args, Log log)
         NugetAction.DownloadNugetAction(properties);
         NugetAction.RestorePackagesAction(properties, Log);
 
-        if (CscAction.Execute(properties, Log))
+        if (CscAction.Execute(properties, Log, buildLib))
```
KrzysztofCwalina: I think all compilation options should be passed using properties argument.

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -106,11 +111,14 @@ static void Build(string[] args, Log log)
...
         {
-            ConvertToCoreConsoleAction(properties);
+            if (!buildLib)
+            {
+                ConvertToCoreConsoleAction(properties);
+            }
             OutputRuntimeDependenciesAction(properties);
-            Console.WriteLine("bin\\{0}.exe created", properties.AssemblyName);
+            Console.WriteLine(buildLib ? "bin\\{0}.dll created" : "bin\\{0}.exe created", properties.AssemblyName);
```
KrzysztofCwalina: I would do:
("{0} created", properties.AssemblyNameAndExtension);
... and add the property to ProjectProperties

demos/dotnet/dotnet/ProjectProperties.cs
ahsonkhan: I used the Common MSBuild Project Properties as an example (https://msdn.microsoft.com/en-us/library/vstudio/bb629394(v=vs.110).aspx) which had OutputType to 'Specifies the file format of the output file.'

KrzysztofCwalina: ok. makes sense.

ahsonkhan:
@KrzysztofCwalina I have addressed the comments. Can you please review?

KrzysztofCwalina:
The changes look good. Please squash the changes and merge. Thanks!

PR #226, opened 2015-09-04
Adding support for building sources in multiple directories and using  a basic '.dotnetproj' file.

ahsonkhan:


ahsonkhan:
@KrzysztofCwalina can you please review?

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -27,7 +28,10 @@ static void Main(string[] args)
             Log.IsEnabled = true;
         }
 
-        switch (args[0])
+        var commandArgument = args[0];
```
KrzysztofCwalina: At some point we need to rewire this command parsing logic. The current implementation does not really scale past one or two options.

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -43,8 +47,26 @@ static void Main(string[] args)
...
+                else
+                {
+                    Console.WriteLine("Please specify the {0} compiler option.", compilerOptions[0]);
+                }
+                break;
+
+            case "/target":
+                if (compilerOptions.Length > 1)
+                {
+                    Build(args, Log, compilerOptions[1] == "library");
```
KrzysztofCwalina: Hmm, could you validate that it's actually exe if we build an exe?

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -64,6 +86,7 @@ static void PrintUsage()
         string appName = Environment.GetCommandLineArgs()[0];
         Console.WriteLine("{0} [/log] - compiles sources in current direcotry. optionally logs diagnostics info.", appName);
         Console.WriteLine("{0} /target:library - compiles sources in current directory into dll.", appName);
+        Console.WriteLine("{0} /recurse:<wildcard> - compiles sources in current directory and subdirectories according to the wildcard specifications.", appName);
```
KrzysztofCwalina: Hopefully packages directory will never have cs files :-) And more seriously, I think we need to filter out subdirectories created by the tool: packages, bin, tools.

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -157,8 +180,24 @@ public static ProjectProperties InitializeProperties(string[] args, Log log, boo
         properties.OutputType = buildDll ? ".dll" : ".exe";
         FindCompiler(properties);
 
+        var projectFiles = Directory.GetFiles(properties.ProjectDirectory, "*.dotnetproj");
+        bool projectFileExists = projectFiles.Length > 0;
         // Sources
-        properties.Sources.AddRange(Directory.GetFiles(properties.ProjectDirectory, "*.cs"));
+        if (projectFileExists)
+        {
+            foreach(var projectFile in projectFiles)
```
KrzysztofCwalina: Why do you want to enable more than one project file?

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -157,8 +180,24 @@ public static ProjectProperties InitializeProperties(string[] args, Log log, boo
...
+            {
+                properties.Sources.AddRange(ParseProjectFile(properties.ProjectDirectory, projectFile));
```
KrzysztofCwalina: I would rather pass properties to the parse method. We will have more things in the project file besides sources (very soon).

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -157,8 +180,24 @@ public static ProjectProperties InitializeProperties(string[] args, Log log, boo
...
+            }
+        }
+        properties.Sources.AddRange(Directory.GetFiles(properties.ProjectDirectory, "*.cs", (args.Length > 1 && args[1] == "*.cs") ? SearchOption.AllDirectories : SearchOption.TopDirectoryOnly));
+        if (args.Length > 1 && args[1] != "*.cs")
```
KrzysztofCwalina: Why is this indexing just to the second element of the array?

demos/dotnet/dotnet/dotnet.cs
ahsonkhan: Agreed. Let's discuss the behavior of the command parsing logic to make it more flexible and scalable.

demos/dotnet/dotnet/dotnet.cs
ahsonkhan: Let's say the command argument is generally of the form "/*:*". Once we split the argument and the specification (split over ':'), we have two elements only - the command option and the specification. Once we grab the specification (i.e. whether it is '*.cs' or a subdirectory name), we can get the appropriate source files.

KrzysztofCwalina:
Looks good. Please merge.

PR #231, opened 2015-09-09
Adding support for /debug and /optimize compiler options.

ahsonkhan:


demos/dotnet/dotnet/dotnet.cs
```diff
@@ -240,6 +264,29 @@ public static ProjectProperties InitializeProperties(string[] args, Log log, boo
...
+        if (Array.Exists(args, element => element == "/optimize"))
+        {
+            properties.CscOptions.Add("/optimize");
+        }
+
+        var platformOption = Array.Find(args, element => element.StartsWith("/platform"));
+
+        if (platformOption != null)
+        {
+            if (!(platformOption == "/platform:anycpu32bitpreferred" && buildDll)) properties.CscOptions.Add(platformOption);
```
KrzysztofCwalina: The other options this PR add are simple options that need to be just passed to the compiler. But the architecture option is not so simple. In particular, the routine that copies package files to the output folder needs to take architecture into the account. 

BTW, feel free to remove either fix this issue, or remove /platfrom from this PR and do it as a separate work item.

KrzysztofCwalina:
As we talked before, we need a bit more robust command line parsing, but we agreed to do it as a separate work item. 

Otherwise looks good. Please go ahead and merge.

Issue #246, opened 2015-09-14
Add support for commonly used compiler options: /platform


ahsonkhan:
The routine that copies package files to the output folder needs to take architecture into the account.

ahsonkhan:
This is done.

PR #249, opened 2015-09-16
Added support for /platform and [SourceFiles] to be specified in the …

ahsonkhan:
…command arguments

Added option for /platform and user can specify source files to compile.
Fixed some issues with /recurse option so that functionality matches
csc.exe.

### DIFF
--- a/demos/dotnet/dotnet/dotnet.cs
+++ b/demos/dotnet/dotnet/dotnet.cs
@@ -179,9 +179,17 @@ static class Program
         log.IsEnabled = false;
         var properties = ProjectPropertiesHelpers.InitializeProperties(args, log);
         log.IsEnabled = previous;
-        if (Directory.Exists(properties.ToolsDirectory)) Directory.Delete(properties.ToolsDirectory, true);
-        if (Directory.Exists(properties.OutputDirectory)) Directory.Delete(properties.OutputDirectory, true);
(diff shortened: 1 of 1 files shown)


ahsonkhan:
Please review. Related to issue #246

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -293,20 +312,39 @@ public static ProjectProperties InitializeProperties(string[] args, Log log, boo
             properties.AssemblyName = Path.GetFileNameWithoutExtension(properties.Sources[0]);
         }
 
+        string result = string.Empty;
+
+        var osName = GetOSName(Environment.OSVersion);
```
KrzysztofCwalina: Why do you read the OS from environment? How will people compile for Win7 on Win 10, for example? Also, some of the runtime packages don't have folders for all OSes. And so the tool will fail. I thought we decided to skip OS specific builds till we have packages that support Linux and/or Mac.

ahsonkhan: This osName is not used anywhere atm. The tool works as discussed and as expected since OS is not taken into account (hardcoded 'win7'). I agree that it doesn't make sense to read the OS from the user's environment. I will remove the method and this function call.

ahsonkhan:
@KrzysztofCwalina please review.

KrzysztofCwalina:
looks good

Issue #242, opened 2015-09-11
Self host dotnet.exe


KrzysztofCwalina:
dotnet.exe should be able to compile itself, and so be a .NET Core app.

KrzysztofCwalina:
BTW, we should keep it as a .NET Framework project too. At this point it's easier to develop and debug with VS support. Later on (when we have solid support for our console apps), we can stop building it as a .NET Framework app.

KrzysztofCwalina:
This is done

Issue #218, opened 2015-09-03
dotnet.exe should be a CoreCLR app


KrzysztofCwalina:

KrzysztofCwalina:
This tool is superseded by dotnet-cli

PR #255, opened 2015-09-18
Making dotnet.exe a .Net Core app so that it can compile itself.

ahsonkhan:
Changing some API uses to only the ones available via NuGet and adding
the extra packages, references, and dependencies so that dotnet can
compile itself.


ahsonkhan:
About issue #242 /issue #218

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -712,13 +735,6 @@ public static bool RestorePackagesAction(ProjectProperties properties, Log log)
             Console.WriteLine(e);
             return false;
         }
-        finally
-        {
-            if (process != null)
-            {
-                process.Close();
```
weshaggard: You should switch this Close call to a Dispose call.

ahsonkhan: I removed the Close call already. Is disposal necessary?

demos/dotnet/dotnet/dotnet.cs
```diff
@@ -651,7 +671,10 @@ static class NugetAction
 {
     static void CreateNugetConfig(ProjectProperties properties)
     {
-        using (var file = new StreamWriter(Path.Combine(properties.ToolsDirectory, "nuget.config"), false))
+        string fileName = Path.Combine(properties.ToolsDirectory, "nuget.config");
+        FileStream fs = null;
+        fs = new FileStream(fileName, FileMode.Create);
```
KrzysztofCwalina: Why the above are two separate lines of code?

KrzysztofCwalina:
looks good

demos/dotnet/dotnet/dotnet.cs
weshaggard: Yes it is a good idea to call Dispose. You should just add a using statement around the Process usage.

demos/dotnet/dotnet/dotnet.cs
ahsonkhan: Agreed. Merged into single line.

demos/dotnet/dotnet/dotnet.cs
ahsonkhan: Good idea. Done.

PR #261, opened 2015-09-23
Removed the support for packages.txt and work towards making dotnet into a CoreCLR app.

ahsonkhan:
Removed support for packages.txt file (use packages listed in .dotnetproj file).
Added a new .csproj file which can be used to build against coreclr.
Removed the /edit command.

### DIFF
--- a/demos/dotnet/dotnet/project.json
+++ b/demos/dotnet/dotnet/project.json
@@ -1,8 +1,7 @@
 ﻿{
   "dependencies": {
     "Microsoft.NetCore.Console": "1.0.0-beta-23321",
-    "Newtonsoft.Json": "7.0.1",
-    "System.IO.Compression": "4.0.1-beta-23321",
+    "System.IO.Compression": "4.0.1-beta-23321"
   },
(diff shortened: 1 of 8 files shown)


demos/dotnet/dotnet/project.json
```diff
@@ -0,0 +1,12 @@
+﻿{
+  "dependencies": {
```
KrzysztofCwalina: Why project.json and dotnet.dotnetproj contain different packages?

ahsonkhan: dotnet.dotnetproj contains all the required packages to self-host (along with the explicit references.txt and dependencies.txt files) using nugget as of now. The project.json file is used by dnu to restore the packages so that dotnet can be compiled into a coreclr app using msbuild and the dotnet-core.csproj file.

KrzysztofCwalina: Could you then add a readme note to the folder explaining what all these projects are for? 
Otherwise looks good.

PR #262, opened 2015-09-24
Substantial code refactoring.

ahsonkhan:
Splitting classes into separate files.
Remove unnecessary or redundant code.
General code clean-up.
Creating a 'Settings' class to hold all the command line argument options that are passed into dotnet.


demos/dotnet/dotnet/CscAction.cs
```diff
@@ -0,0 +1,103 @@
...
+using System.IO;
+using System.Text;
+
+namespace dotnet
+{
+    internal static class CscAction
+    {
+        public static bool Execute(ProjectProperties properties, Log log)
+        {
+            Console.WriteLine("compiling");
```
piotrpMSFT: Why is this not the log?
It's not really compiling, right? "Invoking Compiler"?

demos/dotnet/dotnet/CscAction.cs
```diff
@@ -0,0 +1,103 @@
...
+            var processSettings = new ProcessStartInfo
+            {
+                FileName = properties.CscPath,
+                Arguments = properties.FormatCscArguments()
+            };
+
+            log.WriteLine("Executing {0}", processSettings.FileName);
+            log.WriteLine("Csc Arguments: {0}", processSettings.Arguments);
+
+            processSettings.CreateNoWindow = true;
```
piotrpMSFT: Why are these properties not set above when you create processSettings? Seems like the initializer can own this concern all-up...

demos/dotnet/dotnet/CscAction.cs
```diff
@@ -0,0 +1,103 @@
...
+            processSettings.UseShellExecute = false;
+
+            Process cscProcess;
+            try
+            {
+                cscProcess = Process.Start(processSettings);
+            }
+            catch (Win32Exception)
+            {
+                Console.WriteLine("ERROR: csc.exe needs to be on the path.");
```
piotrpMSFT: Log.WriteLine?

piotrpMSFT: that message seems to conflict with properties.CscPath. Why is there a path on properties but it also needs to be on "the path" which I assume means the PATH env var?

demos/dotnet/dotnet/CscAction.cs
```diff
@@ -0,0 +1,103 @@
...
+            if (cscProcess == null) return false;
+            var output = cscProcess.StandardOutput.ReadToEnd();
+            log.WriteLine(output);
+
+            cscProcess.WaitForExit();
+
+            return !output.Contains("error CS");
+        }
+
+        private static string FormatReferenceOption(this ProjectProperties project)
```
piotrpMSFT: This doesn't seem like the right place for these ProjectProperties formatters. ProjectPropertyCscExtensions or something...?

demos/dotnet/dotnet/CscAction.cs
```diff
@@ -0,0 +1,103 @@
...
+                {
+                    first = false;
+                }
+                builder.Append(Path.Combine(project.PackagesDirectory, reference));
+            }
+            builder.Append(" ");
+            return builder.ToString();
+        }
+
+        private static string FormatSourcesOption(this ProjectProperties project)
```
piotrpMSFT: same as above...

demos/dotnet/dotnet/CscAction.cs
```diff
@@ -0,0 +1,103 @@
...
+            foreach (var source in project.Sources)
+            {
+                builder.Append(" ");
+                builder.Append(source);
+            }
+            return builder.ToString();
+        }
+
+        private static string FormatCscOptions(this ProjectProperties project)
+        {
```
piotrpMSFT: And here...

demos/dotnet/dotnet/CscAction.cs
```diff
@@ -0,0 +1,103 @@
...
+
+            builder.Append(" ");
+            builder.Append("/out:");
+            builder.Append(project.OutputAssemblyPath);
+            builder.Append(" ");
+            return builder.ToString();
+        }
+
+        private static string FormatCscArguments(this ProjectProperties project)
+        {
```
piotrpMSFT: And here...

demos/dotnet/dotnet/Helpers.cs
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.IO;
+
+namespace dotnet
+{
+    internal static class Helpers
```
piotrpMSFT: These are System.IO.File extensions? FileSystemHelpers? Helpers seems like a dumping-ground sort of name...

demos/dotnet/dotnet/Log.cs
```diff
@@ -4,39 +4,42 @@
 using System;
 using System.Collections.Generic;
 
-class Log
```
piotrpMSFT: Can we put this behind an interface? I don't imagine this is the only logger folks might want...

demos/dotnet/dotnet/CscAction.cs
ahsonkhan: Already fixed for next check-in.
I will change the string.

demos/dotnet/dotnet/CscAction.cs
ahsonkhan: Agreed.

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+
+using System;
+using System.Diagnostics;
+using System.IO;
+using System.Net.Http;
+using System.Text;
+
+namespace dotnet
+{
+    internal static class NugetAction
```
piotrpMSFT: These class-level Action concepts seem like they'd be better named Agents or similar. They are not actions themselves, instead being containers/groupings of related actions. This is especially visible in the method of this Type called "DownloadNugetAction" leading to "NugetAction.DownloadNugetAction()". 

I think I'd rather see "NugetAgent.DownloadNuget()". Thoughts?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        public static bool GetNugetAndRestore(ProjectProperties properties, Log log)
+        {
+            var nugetFile = Path.Combine(properties.ToolsDirectory, "nuget.exe");
```
piotrpMSFT: This feels like it should follow your Csc extension method pattern. Something like properties.GetNugetExePath()...

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+            var success = true;
+            if (!File.Exists(nugetFile))
+            {
+                success = DownloadNugetAction(properties, nugetFile);
+            }
+
+            var projectJsonFile = Path.Combine(properties.ProjectDirectory, "project.json");
```
piotrpMSFT: properties.GetProjectJsonPath()?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+using System.IO;
+using System.Net.Http;
+using System.Text;
+
+namespace dotnet
+{
+    internal static class NugetAction
+    {
+        public static bool GetNugetAndRestore(ProjectProperties properties, Log log)
+        {
```
piotrpMSFT: This method seems like it's really this:

    {
        EnsureNugetExe();
        RestoreProjectDependencies(properties);
    }

Is that the right way to factor it?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+            var success = true;
+            if (!File.Exists(nugetFile))
+            {
+                success = DownloadNugetAction(properties, nugetFile);
+            }
+
+            var projectJsonFile = Path.Combine(properties.ProjectDirectory, "project.json");
+            if (!File.Exists(projectJsonFile))
+            {
+                projectJsonFile = Path.Combine(properties.ToolsDirectory, "project.json");
```
piotrpMSFT: Why would project.json be in the ToolsDirectory?

Seems you want a GetProjectJsonOrDefault()?

demos/dotnet/dotnet/CscAction.cs
ahsonkhan: If the csc.exe is not found in the properties.CscPath, the CscPath property is default set to 'csc.exe'. If csc.exe is not found anywhere up the path (from the current directory), this error is reported. Where we get csc.exe needs to be fixed as right now it basically just looks for it anywhere within .NET framework and hopes one exists. Once that is fixed, I think, that is when we can address proper error handling. Thoughts?

demos/dotnet/dotnet/NugetAction.cs
ahsonkhan: Agreed, that makes sense to me.

demos/dotnet/dotnet/NugetAction.cs
ahsonkhan: The project.json in the ToolsDirectory is the auto-generated one.

demos/dotnet/dotnet/CscAction.cs
piotrpMSFT: Sounds complicated. Can we encapsulate the CSC path finding stuff somewhere?

demos/dotnet/dotnet/NugetAction.cs
piotrpMSFT: Is that auto-generated one constant? Or does it change?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+            {
+                Console.WriteLine("Failed to get nuget or restore packages.");
+                return false;
+            }
+            return true;
+        }
+
+        public static bool DownloadNugetAction(ProjectProperties properties, string nugetFile)
+        {
+            CreateDefaultProjectJson(properties);
```
piotrpMSFT: Why are these two methods here? I'd not expect that from "Download Nuget"

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+            CreateNugetConfig(properties);
+
+            var client = new HttpClient();
+            var requestUri = new Uri(@"http://dist.nuget.org/win-x86-commandline/v3.1.0-beta/nuget.exe",
```
piotrpMSFT: Probably later, but this should be configurable.

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                UriKind.Absolute);
+
+            var sourceStreamTask = client.GetStreamAsync(requestUri);
```
piotrpMSFT: Is WebClient not available to us? DownloadFile seems cleaner...

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+            try
+            {
+                sourceStreamTask.Wait();
+            }
+            catch (AggregateException exception)
+            {
+                foreach (var ex in exception.InnerExceptions)
+                {
+                    Console.WriteLine(ex.Message);
```
piotrpMSFT: log.Writeline?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                UriKind.Absolute);
+
+            var sourceStreamTask = client.GetStreamAsync(requestUri);
+            try
+            {
+                sourceStreamTask.Wait();
+            }
+            catch (AggregateException exception)
+            {
+                foreach (var ex in exception.InnerExceptions)
```
piotrpMSFT: Should we dump all exceptions? Or just say that download has failed?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                    var read = sourceStream.Read(buffer, 0, buffer.Length);
+                    if (read < 1) break;
+                    destinationStream.Write(buffer, 0, read);
+                }
+            }
+
+            return true;
+        }
+
+        private static void CreateDefaultProjectJson(ProjectProperties properties)
```
piotrpMSFT: Why is the NugetAgent creating the Project.JSON?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+        {
+            var fileName = Path.Combine(properties.ToolsDirectory, "project.json");
```
piotrpMSFT: properties.GetProjectJsonPath()

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                        file.WriteLine();
+                    }
+                }
+                file.WriteLine(@"    },");
+                file.WriteLine(@"    ""frameworks"": {");
+                file.WriteLine(@"        ""dnxcore50"": { }");
+                file.WriteLine(@"    }");
+                file.WriteLine(@"}");
+
+                //"runtimes": {
```
piotrpMSFT: delete these ;)

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                //"win7-x86": { },
+                //"win7-x64": { }
+                //},
+            }
+        }
+
+        private static void CreateNugetConfig(ProjectProperties properties)
```
piotrpMSFT: If this were nuget.exe, is it responsible for Create Nuget Config? Or is that someone else's job?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                file.WriteLine(@"    </packageSources>");
+                file.WriteLine(@"</configuration>");
+            }
+        }
+
+        public static bool RestorePackagesAction(ProjectProperties properties, Log log, string nugetFile, string jsonFile)
+        {
+            Console.WriteLine("restoring packages");
+
+            if (!File.Exists(nugetFile))
```
piotrpMSFT: We already have a method for ensuring NuGet exists, right?

demos/dotnet/dotnet/NugetAction.cs
ahsonkhan: Sure, that is the intent.

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                file.WriteLine(@"        <add key = ""nuget.org"" value = ""https://www.nuget.org/api/v2/""/>");
+
+                file.WriteLine(@"    </packageSources>");
+                file.WriteLine(@"</configuration>");
+            }
+        }
+
+        public static bool RestorePackagesAction(ProjectProperties properties, Log log, string nugetFile, string jsonFile)
+        {
+            Console.WriteLine("restoring packages");
```
piotrpMSFT: log.WriteLine()?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+
+            if (!File.Exists(nugetFile))
+            {
+                Console.WriteLine("Could not find file {0}.", nugetFile);
```
piotrpMSFT: log.WriteLine()?

demos/dotnet/dotnet/NugetAction.cs
ahsonkhan: I don't understand what you mean by that.

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                return false;
+            }
+
+            if (!File.Exists(jsonFile))
+            {
+                Console.WriteLine("Could not find file {0}.", jsonFile);
```
piotrpMSFT: log.WriteLine()?

demos/dotnet/dotnet/NugetAction.cs
```diff
@@ -0,0 +1,188 @@
...
+                return false;
+            }
+
```
piotrpMSFT: For all of the above, it seems we'd want to throw an Argument Exception, no?

demos/dotnet/dotnet/NugetAction.cs
ahsonkhan: To implement it the way you are suggesting would require a bit more work. Can we add that as a separate issue?

demos/dotnet/dotnet/OtherActions.cs
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.IO;
+using System.Text;
+
+namespace dotnet
+{
+    internal static class OtherActions
```
piotrpMSFT: This seems like another dumping-ground class. This class represents the ProjectAgent and the BuildAgent... no?

demos/dotnet/dotnet/ProjectProperties.cs
```diff
@@ -17,15 +18,9 @@ class ProjectProperties
         public readonly List<string> CscOptions = new List<string>();
```
piotrpMSFT: Looks like this class is specialized to CSC... why not put the CSC extension methods in here, along with beefing up the name?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Xml;
+
+namespace dotnet
+{
+    internal static class ProjectPropertiesHelpers
+    {
+        public static ProjectProperties InitializeProperties(Settings settings, Log log)
```
piotrpMSFT: Why isn't this a constructor?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+        {
+            // General Properites
+            var properties = new ProjectProperties();
+            var currentDirectory = Directory.GetCurrentDirectory();
+
+            var buildDll = settings.Target == "library";
+
+            properties.ProjectDirectory = Path.Combine(currentDirectory);
```
piotrpMSFT: initializer?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            properties.PackagesDirectory = Path.Combine(properties.ProjectDirectory, "packages");
+            properties.OutputDirectory = Path.Combine(properties.ProjectDirectory, "bin");
+            properties.ToolsDirectory = Path.Combine(properties.ProjectDirectory, "tools");
+            properties.AssemblyName = Path.GetFileName(properties.ProjectDirectory);
+            properties.OutputType = buildDll ? ".dll" : ".exe";
+            FindCompiler(properties);
+
+            AddToListWithoutDuplicates(properties.Sources, ParseProjectFile(properties, settings.ProjectFile, "Compile"));
```
piotrpMSFT: Why not use a Set?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            var currentDirectory = Directory.GetCurrentDirectory();
+
+            var buildDll = settings.Target == "library";
+
+            properties.ProjectDirectory = Path.Combine(currentDirectory);
+            properties.PackagesDirectory = Path.Combine(properties.ProjectDirectory, "packages");
+            properties.OutputDirectory = Path.Combine(properties.ProjectDirectory, "bin");
+            properties.ToolsDirectory = Path.Combine(properties.ProjectDirectory, "tools");
+            properties.AssemblyName = Path.GetFileName(properties.ProjectDirectory);
+            properties.OutputType = buildDll ? ".dll" : ".exe";
```
piotrpMSFT: Is this guaranteed to be one of the two? do we not want an enum?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            FindCompiler(properties);
+
+            AddToListWithoutDuplicates(properties.Sources, ParseProjectFile(properties, settings.ProjectFile, "Compile"));
+            
+            foreach (var file in settings.SourceFiles.Where(f => !f.StartsWith(properties.PackagesDirectory) && !f.StartsWith(properties.OutputDirectory) &&
+                                                              !f.StartsWith(properties.ToolsDirectory)))
+            {
```
piotrpMSFT: this condition is pretty gnarly... what is it meant to represent?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+                AddToListWithoutDuplicates(properties.Sources, file);
+            }
+
+            if (!string.IsNullOrWhiteSpace(settings.Recurse) && settings.Recurse != "*.cs")
```
piotrpMSFT: This deserves a name as well... not sure what's going on!

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            {
+                AddToListWithoutDuplicates(properties.Sources,
+                    Directory.GetFiles(properties.ProjectDirectory, settings.Recurse, SearchOption.AllDirectories));
+            }
+
+            if (properties.Sources.Count == 1)
```
piotrpMSFT: else...?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            {
+                properties.AssemblyName = Path.GetFileNameWithoutExtension(properties.Sources[0]);
+            }
+
+            var platformOptionSpecicifcation = GetPlatformOption(settings.Platform);
+
+            // The anycpu32bitpreferred setting is valid only for executable (.EXE) files
```
piotrpMSFT: here down, I think most places where you have a comment we should break out that block of code into a method.

piotrpMSFT: We should also build a list of strings, then add the preamble to that list of strings. It's more work to understand the code like this and there is more opportunity for error.

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            AddToListWithoutDuplicates(properties.Packages, ParseProjectFile(properties, settings.ProjectFile, "Package"));
+
+            LogProperties(log, properties, "Adjusted Properties Log:", buildDll);
+
+            return properties;
+        }
+
+        public static string GetPlatformOption(string optionSpecification)
+        {
+            if (optionSpecification == "anycpu32bitpreferred" || optionSpecification == "x86")
```
piotrpMSFT: switch(optionSpecification):?

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            {
+                return "x86";
+            }
+            return "x64";
+        }
+
+        public static string GetRuntimeCoreClr(string platform)
+        {
+            // platform - x86, x64, arm
+            return "\"Microsoft.NETCore.Runtime.CoreCLR-" + platform.ToLower() + "\": \"1.0.0\"";
```
piotrpMSFT: from here down, consider string.Format()

Better yet, consider C# 6 string interpolation: https://msdn.microsoft.com/en-us/magazine/dn879355.aspx

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+            log.WriteLine("output path          {0}", project.OutputAssemblyPath);
+            log.WriteList(project.Sources, "SOURCES");
+            log.WriteList(project.Packages, "PACKAGES");
+            log.WriteList(project.References, "REFERENCES");
+            log.WriteList(project.Dependencies, "DEPENDENCIES");
+            log.WriteList(project.CscOptions, "CSCOPTIONS");
+            log.WriteLine("-------------------------------------------------");
+        }
+
+        private static void Adjust(ProjectProperties properties, string adjustmentFilePath, ICollection<string> list)
```
piotrpMSFT: name is unclear...

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+                    list.Add(Path.Combine(properties.PackagesDirectory, adjustment));
+                }
+            }
+        }
+
+        private static List<string> ParseProjectFile(ProjectProperties properties, string projectFile,
+            string elementName)
+        {
+            var attributes = GetAttributes(elementName);
+            if (!File.Exists(projectFile) || attributes == null)
```
piotrpMSFT: initialize attributeValues before this block. Then just return it if the condition is met.

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+                    list.Remove(Path.Combine(properties.PackagesDirectory, adjustment));
+                }
+                else
+                {
+                    list.Add(Path.Combine(properties.PackagesDirectory, adjustment));
+                }
+            }
+        }
+
+        private static List<string> ParseProjectFile(ProjectProperties properties, string projectFile,
```
piotrpMSFT: Think about the intent of this method. It seems to be building a list of Xml attributes for nodes with a particular name... that's not what I thinjk of when I read "ParseProjectFile"...

piotrpMSFT: Ok, it does some other stuff with the values depending on the node type, but you understand the concern I'm sure.

demos/dotnet/dotnet/ProjectPropertiesHelpers.cs
```diff
@@ -0,0 +1,335 @@
...
+                default:
+                    values = new List<string>();
+                    break;
+            }
+            return values;
+        }
+
+        private static List<string> GetAttributes(string elementName)
+        {
+            var compileAttributes = new List<string> {"Include"};
```
piotrpMSFT: Can these be const?

ahsonkhan:
I will create a work item to address some of the comments made here as they are important but would take time. I will continue and merge the changes however since additional work depends on some of these changes.

PR #270, opened 2015-09-25
Added nuget feed to myget.org

KrzysztofCwalina:


weitzhandler:
Would love to see this package on NuGet, so we don't have to subscribe to more feeds.

PR #276, opened 2015-09-28
Optimized allocations when disposing UVStream

KrzysztofCwalina:
The change elimizates Action<int> and display class allocations.

### DIFF
--- a/demos/LibuvWithNonAllocatingFormatters/src/System/Net/Libuv/Handle.cs
+++ b/demos/LibuvWithNonAllocatingFormatters/src/System/Net/Libuv/Handle.cs
@@ -50,7 +50,7 @@ namespace System.Net.Libuv
             GC.SuppressFinalize(this);
         }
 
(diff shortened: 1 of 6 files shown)


demos/LibuvWithNonAllocatingFormatters/src/System/Net/Libuv/Handle.cs
```diff
@@ -50,7 +50,7 @@ public void Dispose()
             GC.SuppressFinalize(this);
         }
 
-        protected virtual void Dispose(bool disposing)
+        public virtual void Dispose(bool disposing)
```
weshaggard: Why are you violating your own dispose pattern, by making this public? :)

KrzysztofCwalina: Ah, good catch. It's a left over from an experiment I did and then dropped. I will fix that with the next commit.

PR #296, opened 2015-10-07
Adding yet another Slice<T>

KrzysztofCwalina:
There are several competing implementation we are considering for slices.
This one is what we call "three field" implementation.

### DIFF
--- a/src/System.Buffers/src/System/ByteSpan.cs
+++ b/src/System.Buffers/src/System/ByteSpan.cs
@@ -19,6 +19,11 @@ namespace System {
             _length = length;
         }
 
+        public static ByteSpan Empty
+        {
+            get { return new ByteSpan(); }
+        }
+
         public byte this[int index]
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -40,6 +45,16 @@ namespace System {
         {
             Precondition.Require(valueLength <= Length);
             BufferInternal.MemoryCopy(value, _data, _length, valueLength);
+        }
+
+        [CLSCompliant(false)]
+        public bool TrySet(byte* value, int valueLength)
+        {
+            if(valueLength <= Length) {
+                return false;
+            }
+            BufferInternal.MemoryCopy(value, _data, _length, valueLength);
+            return true;
         }
 
         [CLSCompliant(false)]
@@ -76,6 +91,9 @@ namespace System {
 
         public byte[] CreateArray()
         {
+            if(_length == 0) {
+                return new byte[0];
+            }
             byte[] array = new byte[_length];
             Marshal.Copy((IntPtr)UnsafeBuffer, array, 0, _length);
             return array;
--- a/src/System.Slices/src/System/Span.cs
+++ b/src/System.Slices/src/System/Span.cs
@@ -1,0 +1,293 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System
+{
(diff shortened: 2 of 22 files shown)


KrzysztofCwalina:
@joeduffy

src/System.Slices/src/System/Span.cs
```diff
@@ -0,0 +1,293 @@
...
+
+namespace System
+{
+    /// <summary>
+    /// Span is a uniform API for dealing with arrays and subarrays, strings
+    /// and substrings, and unmanaged memory buffers.  It adds minimal overhead
+    /// to regular accesses and is a struct so that creation and subslicing do
+    /// not require additional allocations.  It is type- and memory-safe.
+    /// </summary>
+    public struct Span<T> : IEnumerable<T>
```
sharwell: :question: Why not implement `IList<T>`?

src/System.Slices/src/System/Span.cs
```diff
@@ -0,0 +1,293 @@
...
+        /// <param name="array">The target array.</param>
+        /// <param name="start">The index at which to begin the span.</param>
+        /// <param name="end">The index at which to end the span (exclusive).</param>
+        /// <exception cref="System.ArgumentException">
+        /// Thrown if the 'array' parameter is null.
+        /// </exception>
+        /// <exception cref="System.ArgumentOutOfRangeException">
+        /// Thrown when the specified start or end index is not in range (&lt;0 or &gt;&eq;length).
+        /// </exception>
+        public Span(T[] array, int start, int end)
```
sharwell: :exclamation: .NET APIs representing a span are specified by start and count. The constructor needs to follow this pattern. If you also want to provide a way to construct a span from start and end indexes, you can provide a static `Span<T>.FromBounds` helper method.

src/System.Slices/src/System/Span.cs
```diff
@@ -0,0 +1,293 @@
...
+            }
+            else {
+                m_object = null;
+                m_offset = UIntPtr.Zero;
+                Length = 0;
+            }
+        }
+
+        /// <summary>
+        /// Creates a new span over the target unmanaged buffer.  Clearly this
```
sharwell: :thought_balloon: I'm not a fan of the way this is worded. I would prefer something like the following:

> Initializes a new instance of the Span&lt;T&gt; struct. This method does not perform bounds checking for the unmanaged memory buffer. The caller should ensure that ptr and length always represent a valid range of memory which can be used as a buffer.

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: Maybe at some point we would implement it.

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: Yeah, good point. I will fix it.

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: BTW, I did not write the original comment (@joeduffy did). But I have to say I like Joe's better. It explains what constitutes a valid buffer (right size for Ts). Your version just says that it needs to be valid.

src/System.Buffers/src/System/ByteSpan.cs
```diff
@@ -76,6 +91,9 @@ public ByteSpan Slice(int index, int count)
 
         public byte[] CreateArray()
         {
+            if(_length == 0) {
+                return new byte[0];
```
davidfowl: Array.Empty?

KrzysztofCwalina: Does not exist in System.Runtime 4.0.0

PR #306, opened 2015-10-08
Utf8String prototype

krwq:


src/System.Buffers/tests/BasicUnitTests.cs
```diff
@@ -40,5 +40,57 @@ public void ByteSpanEmptyCreateArrayTest()
             var array = empty.CreateArray();
             Assert.Equal(0, array.Length);
         }
+        
+        [Fact]
+        public unsafe void ByteSpanEqualsTestsTwoDifferentBuffersSameValues()
```
krwq: Test name is confusing. It was supposed to be: two different instances of buffer with the same values

KrzysztofCwalina:
@ellism, would you have time to take a look too?

src/System.Buffers/src/System/Buffers/BufferInternal.cs
```diff
@@ -192,5 +192,121 @@ public static unsafe void MemoryCopy(void* source, void* destination, int destin
...
+        public static unsafe bool MemoryEqual(void* lhs, int lhsLength, void* rhs, int rhsLength)
+        {
+            // 1. Fast paths
+            if (lhsLength != rhsLength)
+                return false;
+            if (lhs == rhs)
+                return true;
+            if (lhsLength == 0)
+                return true;
+            if (lhsLength == 4)
```
KrzysztofCwalina: why these are not in the switch/case?

src/System.Buffers/src/System/Buffers/BufferInternal.cs
```diff
@@ -192,5 +192,121 @@ public static unsafe void MemoryCopy(void* source, void* destination, int destin
...
+            
+            // 2. We split bytes into two groups of: (n/4)*4 and n%4 bytes
+            
+            // TODO: should we play with alignment here?
+            UInt32* a = (UInt32*)lhs;
+            UInt32* b = (UInt32*)rhs;
+            
+            // I think those are one processor instruction, so keeping it together
+            int count = lhsLength / sizeof(UInt32);
+            int remainder = lhsLength % sizeof(UInt32);
```
KrzysztofCwalina: Yeah, I checked our JITs some time ago and they generate two idiv instructions. @CarolEidt was thinking about a way to add an intrinsic for this or pattern optimizations.

src/System.Buffers/src/System/Buffers/BufferInternal.cs
krwq: I don't remember my reasoning behind that, will change it

src/System.Buffers/src/System/Buffers/BufferInternal.cs
```diff
@@ -192,5 +192,118 @@ public static unsafe void MemoryCopy(void* source, void* destination, int destin
...
+                case 1: return *(byte*)lhs == *(byte*)rhs;
+                case 2: return *(UInt16*)lhs == *(UInt16*)rhs;
+                case 3: return *(UInt16*)lhs == *(UInt16*)rhs &&  *((byte*)lhs + 2) == *((byte*)rhs + 2);
+                case 4: return *(UInt32*)lhs == *(UInt32*)rhs;
+                case 8: return *(UInt64*)lhs == *(UInt64*)rhs;
+            }
+            
+            // 2. We split bytes into two groups of: (n/4)*4 and n%4 bytes
+            
+            // TODO: should we play with alignment here?
```
KrzysztofCwalina: Yeah, at some point it would be great to do it.

src/System.Buffers/src/System/ByteSpan.cs
```diff
@@ -74,7 +75,8 @@ public int Length
 
         public ByteSpan Slice(int index)
         {
-            Precondition.Require(index < Length);
+            if (index >= Length)
+                return new ByteSpan(_data, 0);
```
KrzysztofCwalina: Shouldn't index > Length be an error?

src/System.Buffers/tests/BasicUnitTests.cs
```diff
@@ -40,5 +40,57 @@ public void ByteSpanEmptyCreateArrayTest()
...
+            byte[] buffer1 = new byte[128];
+            byte[] buffer2 = new byte[128];
+            
+            for (int i = 0; i < 128; i++)
+            {
+                buffer1[i] = (byte)(129 - i);
+                buffer2[i] = (byte)(129 - i);
+            }
+            
+            fixed (byte* buffer1pinned = buffer1) fixed (byte* buffer2pinned = buffer2)
```
KrzysztofCwalina: Could you make it two lines?

src/System.Buffers/tests/BasicUnitTests.cs
```diff
@@ -40,5 +40,57 @@ public void ByteSpanEmptyCreateArrayTest()
...
+                {
+                    Assert.True(b1.Slice(i).Equals(b2.Slice(i)));
+                }
+            }
+        }
+        
+        [Fact]
+        public unsafe void ByteSpanEqualsTestsTwoDifferentBuffersOneValueDifferent()
+        {
+            byte[] buffer1 = new byte[128];
```
KrzysztofCwalina: would it make sense to create a variable for 128?

src/System.Text.Utf8/src/System/Text/SpecConstants.cs
```diff
@@ -0,0 +1,13 @@
+﻿namespace System.Text
+{
+    internal static class SpecConstants
```
KrzysztofCwalina: Should this type be called UnicodeConstants?

src/System.Text.Utf8/src/System/Text/UnicodeCodePoint.cs
```diff
@@ -0,0 +1,39 @@
...
+            return codePoint.Value < 17 * (1 << 16);
+        }
+
+        public bool IsSurrogate()
+        {
+            return Value >= SpecConstants.SurrogateRangeStart && Value <= SpecConstants.SurrogateRangeEnd;
+        }
+
+        [CLSCompliant(false)]
+        public static explicit operator uint(UnicodeCodePoint cp) { return cp.Value; }
```
KrzysztofCwalina: parameter name does not follow naming conventions

src/System.Text.Utf8/src/System/Text/Utf8/Utf8EncodedCodePoint.cs
```diff
@@ -0,0 +1,68 @@
...
+namespace System.Text.Utf8
+{
+    // TODO: Remove this class, make people use char.IsSurrogate and char.ToUtf32 to convert to code points and use Utf8Encoder to write bytes
+    [StructLayout(LayoutKind.Explicit)]
+    public struct Utf8EncodedCodePoint
+    {
+        public unsafe Utf8EncodedCodePoint(char character) : this()
+        {
+            if (char.IsSurrogate(character))
+                throw new ArgumentOutOfRangeException("character", "Surrogate characters are not allowed");
```
KrzysztofCwalina: Because of this error position, I would not make this API public. People will us it to encode/decode chars and it will work most of the time, and then they will crash when they go beyond BMP.

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
```diff
@@ -1,150 +1,161 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
KrzysztofCwalina: Why is the copyright notice removed?

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
```diff
@@ -1,150 +1,161 @@
...
+using System.Text.Utf16;
+using System.Threading.Tasks;
 
 namespace System.Text.Utf8
 {
-    public struct Utf8String : IEquatable<Utf8String>, IComparable<Utf8String>
+    public struct Utf8String : IEnumerable<UnicodeCodePoint>, IEquatable<Utf8String>// TODO:, IComparable<Utf8String> 
+                               // TODO: fix it when we get Span<byte> runtime support
     {
-        static readonly Utf8String s_empty = new Utf8String(new byte[0]);
```
KrzysztofCwalina: Why is empty removed?

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
```diff
@@ -1,150 +1,161 @@
...
+        private ByteSpan _buffer;
 
-        byte[] _array;
-        unsafe byte* _buffer;
-        int _length;
+        private byte[] _bytes;
+        private int _index;
+        private int _length;
```
KrzysztofCwalina: Please add a TODO to move to real Span<byte> once we get it working.

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
KrzysztofCwalina: Why is empty removed?

src/System.Text.Utf8/src/System/Text/Utf8/Utf8StringEnumerator.cs
```diff
@@ -0,0 +1,148 @@
+﻿using System.Collections;
+using System.Collections.Generic;
+
+namespace System.Text.Utf8
+{
+    // TODO: This object is heavier than string itself... Once we got ByteSpan runtime support, change it.
+    public struct Utf8StringEnumerator : IEnumerator<UnicodeCodePoint>
```
KrzysztofCwalina: This type should be nested in Utf8String and should be called Enumerator. This is the pattern we use. See List<T>

src/System.Text.Utf8/src/System/Text/Utf8/Utf8StringEnumerator.cs
```diff
@@ -0,0 +1,148 @@
...
+                    fixed (byte* pinnedBytes = _bytes)
+                    {
+                        ByteSpan buffer = new ByteSpan(pinnedBytes + _index, _lastIndex - _index);
+
+                        succeeded = Utf8Encoder.TryDecodeCodePoint(buffer, out ret, out _currentLenCache);
+                    }
+                }
+                else
+                {
+                    if (_buffer.UnsafeBuffer == _startBuffer.UnsafeBuffer)
```
KrzysztofCwalina: I would always use braces for control statements. Too many bugs result form not having them.

src/System.Text.Utf8/src/System/Text/Utf8/Utf8StringEnumerator.cs
```diff
@@ -0,0 +1,148 @@
...
+                        throw new InvalidOperationException();
+                    succeeded = Utf8Encoder.TryDecodeCodePoint(_buffer, out ret, out _currentLenCache);
+                }
+
+                if (!succeeded || _currentLenCache == 0)
+                    throw new InvalidOperationException();
```
KrzysztofCwalina: message?

KrzysztofCwalina:
Nice PR in general!

src/System.Buffers/tests/BasicUnitTests.cs
mellinoe: You could also pin both arrays in the same statement:

````CSharp
fixed (byte* buffer1pinned = buffer1, buffer2pinned = buffer2)
````

src/System.Buffers/src/System/ByteSpan.cs
krwq: I think this way it is easier to use this class. I.e. when you want to iterate you can do:
while (byteSpan.Length != 0)
with that it would throw when it gets to 0. I think that was the main reason I added this changed it in the first place.

src/System.Buffers/tests/BasicUnitTests.cs
krwq: Nice one, I actually didn't know you can write it in the same statement. Will change it to one of those.

src/System.Buffers/tests/BasicUnitTests.cs
krwq: Will change

src/System.Text.Utf8/src/System/Text/SpecConstants.cs
krwq: Will change

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
krwq: Will revert and add header to the rest of the files

src/System.Text.Utf8/src/System/Text/Utf8/Utf8EncodedCodePoint.cs
krwq: agreed

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
krwq: I started this class from a scratch while learning more about Unicode and then unified to your old version. Will revert

src/System.Text.Utf8/src/System/Text/Utf8/Utf8StringEnumerator.cs
krwq: I agree about the bugs resulting from that but also makes it slightly less readable since the text becomes way vertically wider. I personally like doing that for checking initial conditions and rarely do it in other places. Will change it

src/System.Buffers/src/System/Buffers/BufferInternal.cs
krwq: I was actually thinking about doing that but then thought about the case where one buffer is aligned and the other one isn't and wasn't sure how to handle. I think we should rather move to some known and well perf tested implementation (maybe there is something like that in one of the C++ standard libraries)

src/System.Buffers/src/System/ByteSpan.cs
KrzysztofCwalina: But you could accomplish that by allowing slicing to zero length, but not "past" that.

src/System.Buffers/tests/BasicUnitTests.cs
krwq: pending

src/System.Buffers/src/System/ByteSpan.cs
krwq: Correct but I have another example. Let's say your input is byte array with invalid utf8 character near the end. I.e. First byte of the codepoint has encoded it's width of 4 but it is the last character of the span, if I just try to advance the pointer it would be 0 which I can easily check, if you fail fast I can't do anything about it and have to check (currentpos + charlen >= arrlen)

src/System.Buffers/tests/BasicUnitTests.cs
krwq: pending (2 lines feels more readable)

src/System.Buffers/tests/BasicUnitTests.cs
krwq: pending

src/System.Text.Utf8/src/System/Text/SpecConstants.cs
krwq: pending

src/System.Text.Utf8/src/System/Text/UnicodeCodePoint.cs
krwq: pending

src/System.Text.Utf8/src/System/Text/Utf8/Utf8StringEnumerator.cs
krwq: pending

src/System.Text.Utf8/src/System/Text/Utf8/Utf8EncodedCodePoint.cs
krwq: I added TODO to validate these constructors. I will probably end up removing this class at one point.

src/System.Buffers/src/System/ByteSpan.cs
KrzysztofCwalina: It's clearly an error then. We should either throw an exception or use Try pattern, if perf is an issue (I would just throw for now).

src/System.Text.Utf8/src/System/Text/Utf8/Utf8StringEnumerator.cs
krwq: pending

src/System.Text.Utf8/src/System/Text/Utf8/Utf8StringEnumerator.cs
krwq: pending

KrzysztofCwalina:
OK, looks good. We can discuss the Slice error behavior independently.

krwq:
Thanks!

PR #312, opened 2015-10-10
Add more methods to Utf8String

krwq:


src/System.Text.Utf8/src/System/Text/Utf8/Utf8CodeUnit.cs
```diff
@@ -0,0 +1,45 @@
+﻿namespace System.Text.Utf8
+{
+    public struct Utf8CodeUnit : IEquatable<Utf8CodeUnit>
```
KrzysztofCwalina: Why do we need this to be public? Why do we need a type at all? UTF8 code unit is byte, isn't it?

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
```diff
@@ -101,6 +102,33 @@ IEnumerator IEnumerable.GetEnumerator()
...
+            {
+                return (Utf8CodeUnit)_bytes[_index + i];
+            }
+            else
+            {
+                return (Utf8CodeUnit)_buffer[i];
+            }
+        }
+
+        public Utf8CodeUnit this[int i]
```
KrzysztofCwalina: Why do we need the indexer at all? If we need it, can we return byte?

src/System.Text.Utf8/src/System/Text/Utf8/Utf8CodeUnit.cs
krwq: UTF-16's code unit is UInt16 and we called it char, I think it is more consistent to have some wrapper type (not pushing on the name). You could as well operate directly on byte[] and never create Utf8String and only extension methods to byte[]

KrzysztofCwalina: I think public types need to add enough value to justify their existence. I don't think Utf8CodeUnit adds enough value. And we are not consistent with System.String anyway. Utf8String is actually a sub-string (slice of string), it's not immutable, and amny other differences.

krwq: Here's the thing: Whenever we release Utf8String we will need a way of providing how to assign a single character to Utf8String. If you decide to use byte, you won't be able to implement assignment as byte is not a character, same for char, as char is not UTF8 character and thus can be multibyte. There is no standard type representing it. If you decide to use Utf8CodeUnit you can easily add compiler support for ascii characters or "surrogate" utf8 string characters. Byte seems inappropriate in this context.

krwq: Talked offline, we will discuss it during API Review, for now leaving as is

demos/LowAllocationWebServer/HttpRequestParser.cs
```diff
@@ -5,49 +5,49 @@ namespace System.Net.Http.Buffered
...
-        {
-            throw new InvalidOperationException("you don't want it in a hashtable, do you?");
-        }
-
-        public override string ToString()
-        {
-            return Encoding.UTF8.GetString(_bytes.CreateArray());
-        }
-    }
+    //public struct Utf8Span
```
krwq: that wasn't intended in this PR. will revert Server changes

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.CodePointEnumerator.cs
```diff
@@ -0,0 +1,164 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
krwq: This is the old (unchanged except for the name) Enumerator implementation. Current Enumerator is new.

KrzysztofCwalina:
Please squash the commits and feel free to merge.

PR #322, opened 2015-10-14
Add substring from to char

krwq:


src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
```diff
@@ -337,16 +337,117 @@ public Utf8String Substring(int index, int length)
...
+            return -1;
+        }
+
+        // TODO: Should this be public or should focused on scenarios?
+        private int IndexOf(UnicodeCodePoint codePoint)
         {
             throw new NotImplementedException();
         }
 
+        public bool TrySubstringFrom(Utf8String substring, out Utf8String result)
```
KrzysztofCwalina: The first parameter should not be called substring.

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
```diff
@@ -337,16 +337,117 @@ public Utf8String Substring(int index, int length)
...
+            {
+                result = default(Utf8String);
+                return false;
+            }
+
+            result = Substring(idx);
+            return true;
+        }
+
+        public bool TrySubstringFrom(Utf8CodeUnit codeUnit, out Utf8String result)
```
KrzysztofCwalina: It's also not clear and possibly needs to be controllable where the delimiter ends up being: in the result or in the rest.

krwq: From sounds inclusive to me. Let's discuss naming on the API review

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
krwq: Well, it is a substring or the method returns false, what name do you propose?

KrzysztofCwalina: To me the result is the substring. The substring parameter is more like a prefix the method will be searching for. But please feel to commit this change, just make a note so we talk about this name at the API review.

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
KrzysztofCwalina: Yeah, let's discuss separately.

KrzysztofCwalina:
Squash the commits, and it's good to go.

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
krwq: Renaming to "value" as in string.IndexOf

PR #323, opened 2015-10-14
Adding JsonReader and InvariantParser.TryParse uint from Utf8String

ahsonkhan:
Initial implementation.
Please provide feedback.

@krwq @KrzysztofCwalina 

There is a TODO to use SubstringFrom and SubstringTo once it is completed.

### DIFF
--- a/src/System.Text.Formatting/src/System/Text/Parsing/InvariantParser.cs
+++ b/src/System.Text.Formatting/src/System/Text/Parsing/InvariantParser.cs
@@ -100,6 +100,45 @@ namespace System.Text.Parsing
             return true;
         }
 
+        [CLSCompliant(false)]
+        public static bool TryParse(Utf8String text, out ulong value, out int bytesConsumed)
+        {
+            Precondition.Require(text.Length > 0);
+
+            value = 0;
+            bytesConsumed = 0;
+
+            for (int byteIndex = 0; byteIndex < text.Length; byteIndex++)
+            {
+                byte nextByte = (byte)text[byteIndex];
+                if (nextByte < '0' || nextByte > '9')
+                {
+                    if (bytesConsumed == 0)
+                    {
+                        value = default(ulong);
+                        return false;
+                    }
+                    else
+                    {
+                        return true;
+                    }
+                }
+                ulong candidate = value * 10;
+                candidate += (ulong)nextByte - '0';
+                if (candidate >= value)
+                {
+                    value = candidate;
+                }
+                else
+                {
+                    return true;
+                }
+                bytesConsumed++;
+            }
+
+            return true;
+        }
+
         internal static bool TryParse(string text, int index, int count, out uint value, out int charsConsumed)
         {
             Precondition.Require(count > 0);
--- a/src/System.Text.Formatting/src/System/Text/Parsing/InvariantParser.cs
+++ b/src/System.Text.Formatting/src/System/Text/Parsing/InvariantParser.cs
@@ -3,6 +3,7 @@
 
 using System.Diagnostics;
 using System.Text.Formatting;
+using System.Text.Utf8;
 
 namespace System.Text.Parsing
 {
@@ -55,6 +56,45 @@ namespace System.Text.Parsing
                     }
                     bytesConsumed++;
                 }
+            }
+
+            return true;
+        }
+
+        [CLSCompliant(false)]
+        public static bool TryParse(Utf8String text, out uint value, out int bytesConsumed)
+        {
+            Precondition.Require(text.Length > 0);
+
+            value = 0;
+            bytesConsumed = 0;
+
+            for (int byteIndex = 0; byteIndex < text.Length; byteIndex++)
+            {
+                byte nextByte = (byte)text[byteIndex];
+                if (nextByte < '0' || nextByte > '9')
+                {
+                    if (bytesConsumed == 0)
+                    {
+                        value = default(uint);
+                        return false;
+                    }
+                    else
+                    {
+                        return true;
+                    }
+                }
+                uint candidate = value * 10;
+                candidate += (uint)nextByte - '0';
+                if (candidate >= value)
+                {
+                    value = candidate;
+                }
+                else
+                {
+                    return true;
+                }
+                bytesConsumed++;
             }
 
             return true;
--- a/src/System.Text.Json/src/System/Text/Json/JsonReader.cs
+++ b/src/System.Text.Json/src/System/Text/Json/JsonReader.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Collections.Generic;
 using System.Text.Parsing;
 using System.Text.Utf8;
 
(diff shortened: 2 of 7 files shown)


src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -0,0 +1,181 @@
...
+        public Utf8String ReadProperty()
+        {
+            SkipAll();
+            // TODO: Use _str.SubstringFrom(new Utf8String("\"")).Substring(1).SubstringTo(new Utf8String("\""));
+            ReadToByte(QuoteString, true);
+            return ReadToByte(QuoteString, false);
+        }
+
+        [CLSCompliant(false)]
+        public uint ReadPropertyValueInt()
```
krwq: ReadPropertyValueUInt?

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -0,0 +1,88 @@
...
+            new Utf8String("425-000-1213")
+        };
+
+        private readonly Utf8String _expectedAddress = new Utf8String("1 Microsoft Way, Redmond, 98052");
+
+        private static ReadObject Read(JsonReader json)
+        {
+            var output = new ReadObject();
+
+            json.ReadObjectStart();
```
krwq: I was expecting usability like:
while (json.Read())
{
  switch (json.TokenType)
  {
    case JsonTokenType.StringProperty:
    {
      Console.WriteLine("{0}", json.ReadPropertyValueAsString());
      break;
    }
  }
}

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -0,0 +1,181 @@
...
+
+        private static readonly List<Utf8CodeUnit> EmptyString = new List<Utf8CodeUnit>
+        {
+            new Utf8CodeUnit((byte) ' '),
+            new Utf8CodeUnit((byte) '\n'),
+            new Utf8CodeUnit((byte) '\r'),
+            new Utf8CodeUnit((byte) '\t')
+        };
+
+        private static readonly Utf8CodeUnit QuoteString = new Utf8CodeUnit((byte) '"');
```
KrzysztofCwalina: @krwq, it might be worth to have better APIs for this. It seems like casting chars to bytes might lead to bugs.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -0,0 +1,181 @@
...
+using System.Text.Utf8;
+
+namespace System.Text.Json
+{
+    public struct JsonReader : IDisposable
+    {
+        private readonly Utf8String _str;
+        private int _index;
+
+        private static readonly List<Utf8CodeUnit> EmptyString = new List<Utf8CodeUnit>
```
KrzysztofCwalina: This should probably be an array, not List<T>. Array is more memory efficient and you don't need to resize it.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -0,0 +1,181 @@
...
+        }
+
+        public void Dispose()
+        {
+        }
+
+        public Utf8String ReadObjectStart()
+        {
+            SkipAll();
+            return ReadToByte(CurlyOpenString, true);
```
KrzysztofCwalina: This seems to advance to object start. If that's the case, I don't think we want this behavior. I think we should skip only witespace, but i f there is no "{" after witespace, this method should fail.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
KrzysztofCwalina: Why only unsigned? How do I read signed numbers? Also, why only 32-bit integers?

src/System.Text.Json/tests/JsonReaderTests.cs
KrzysztofCwalina: This is a very good point. We need the ability to read without knowing the order and number of objects, arrays, members in the JSON payload.

src/System.Text.Formatting/src/System/Text/Parsing/InvariantParser.cs
```diff
@@ -60,6 +61,45 @@ public static bool TryParse(ByteSpan text, FormattingData.Encoding encoding, out
             return true;
         }
 
+        [CLSCompliant(false)]
+        public static bool TryParse(Utf8String text, out uint value, out int bytesConsumed)
```
KrzysztofCwalina: These general purpose parsing APIs should be at some point moved to System.Text.Formatting/Parsing. They are more general purpose than JSON parsing.

ahsonkhan: This is in System.Text.Formatting\src\System\Text\Parsing already.

KrzysztofCwalina: Ah, I just noticed that.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
ahsonkhan: Ok. I was thinking if user had comments or some special strings before the start of the json, we would want to skip that until the actual start of the object. The assumption is that we have valid json.

src/System.Text.Json/tests/JsonReaderTests.cs
ahsonkhan: I assumed the user knew the json schema that he was trying to read. Should I enable the usability of the json reader as defined by @krwq?

KrzysztofCwalina: I think we can assume that the schema is known, but the schema does not specify the order or properties, the number of items in arrays, and many other things. I think the sample @krwq outlined solves all these problems nicely.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
ahsonkhan: Does this cover all use case:
* ReadPropertyValueAsUInt (uses InvariantParser.TryParse(Utf8String, out uint, out int))
* ReadPropertyValueAsInt (reads possible "-" and calls ReadPropertyValueAsUInt)
* ReadPropertyValueAsLongUInt  (uses InvariantParser.TryParse(Utf8String, out ulong, out int))
* ReadPropertyValueAsLongInt (reads possible "-" and calls ReadPropertyValueAsLongUInt)

Along with ReadMemberValueAs* which just call ReadPropertyValueAs* (for arrays).

src/System.Text.Json/tests/JsonReaderTests.cs
ahsonkhan: OK, I will target the above defined usage.

ahsonkhan:
Please review and provide feedback. I have tried to enable the usage as per the comments from @krwq,
There is still work left but I wanted to move the conversation further.

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -0,0 +1,160 @@
...
+                    case JsonReader.JsonTokenType.PropertyValueAsString:
+                        value = json.ReadPropertyAsString();
+                        break;
+                    case JsonReader.JsonTokenType.PropertyValueAsInt:
+                        value = json.ReadPropertyValueAsInt();
+                        break;
+                    default:
+                        property = new Utf8String("");
+                        value = null;
+                        break;
```
KrzysztofCwalina: I would simplify some of these more, but for now, let's check in and then talk about additional changes.

PR #324, opened 2015-10-14
made some APIs safe

KrzysztofCwalina:


krwq:
Looks good although I don't quite understand how is adding unsafe less broadly different from having that in the signature (it's not making the code any more safe, is it?)

KrzysztofCwalina:
I just wanted to isolate unsafe code to the smallest scope possible.

PR #329, opened 2015-10-14
Make Slices single module assembly

krwq:


krwq:
TODO: remove checked in .exe

krwq:
TODO: Fix PDBs

KrzysztofCwalina:
Cool. Looks good. As the next commit, could you add an xunit test project? Also, conver the existing tests to xunit. For the performance tests, see how we measure perf in System.Text.Formatting.Tests project.

PR #334, opened 2015-10-16
Expanding the supported schema of the Json Reader

ahsonkhan:
Adjusted the usability based on feedback.
Added more states to enable parsing of 'null', boolean, etc.


src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -70,120 +89,218 @@ public bool Read()
...
+            MoveToNextTokenType();
+            return true;
+        }
+
+        public bool ReadTrue()
+        {
+            if (!Read())
+                throw new IndexOutOfRangeException("Json length is " + _str.Length + " and reading index " + _index +
+                                                   ".");
+            var utf8Bytes = new byte[4];
```
KrzysztofCwalina: Why do you allocate the array?

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -70,120 +89,218 @@ public bool Read()
...
+            for (var i = 0; i < utf8Bytes.Length; i++)
             {
+                utf8Bytes[i] = (byte) _str[_index];
                 _index++;
             }
+            var utf8BytesString = new Utf8String(utf8Bytes);
```
KrzysztofCwalina: why do you create this local? It's not used anywhere, is it?

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
ahsonkhan: This should be resolved after some of the new Utf8String APIs are in place, such as:
bool IsSubstringAt(index, Utf8String)

KrzysztofCwalina: @krwq, don't you already have APIs to do this?

ahsonkhan: Yes, you are right. The API I listed above is not yet available, but I can still use Substring(index, length) to achieve the same behavior without allocating.

Thanks.

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -8,153 +7,252 @@ namespace System.Text.Json.Tests
...
+                        };
+                        if (jsonPairs != null) jsonPairs.Add(pair);
                         break;
-                    case JsonReader.JsonTokenType.PropertyValueAsString:
-                        value = json.ReadPropertyAsString();
+                    case JsonReader.JsonTokenType.Value:
                         break;
-                    case JsonReader.JsonTokenType.PropertyValueAsInt:
-                        value = json.ReadPropertyValueAsInt();
+                    case JsonReader.JsonTokenType.Finish:
```
krwq: Why not just return false from Read()?

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -8,153 +7,252 @@ namespace System.Text.Json.Tests
...
+        }
+
+        private static Array ReadArray(ref JsonReader jsonReader)
+        {
+            Array jsonArray = null;
+            List<Value> jsonValues = null;
+
+            while (jsonReader.Read())
+            {
+                switch (jsonReader.TokenType)
```
krwq: I really think that Whitespace node should exist - it is easy filter them out but hard to add them when they are removed

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -8,153 +7,252 @@ namespace System.Text.Json.Tests
...
+                        break;
+                    case JsonReader.JsonTokenType.ArrayEnd:
+                        if (jsonArray != null)
+                        {
+                            var jsonElements = new Elements {Values = jsonValues};
+                            jsonArray.Elements = new List<Elements> {jsonElements};
+                            return jsonArray;
+                        }
+                        break;
+                    case JsonReader.JsonTokenType.Pair:
```
krwq: Pair?

ahsonkhan: A pair is a 'member' of an object and is of the form string:value.

http://json.org/
> An object is an unordered set of name/value pairs. An object begins with { (left brace) and ends with } (right brace). Each name is followed by : (colon) and the name/value pairs are separated by , (comma).

src/System.Text.Json/tests/JsonReaderTests.cs
ahsonkhan: Is this necessary? I am assuming the usage of Whitespace would be to maintain the input formatting in some way? What should the ReadWhitespace API return?

src/System.Text.Json/tests/JsonReaderTests.cs
ahsonkhan: You are right. I am already returning false from Read() and we never hit this case anyway.

KrzysztofCwalina:
@ahsonkhan, it looks good to me. Let's merge and then do a quick API review (if you feel the APIs are reasonably ready)

ahsonkhan:
I did not add have the Whitespace node. What are your thoughts on it? I don't think it is necessary to add it.

From @krwq 
> I really think that Whitespace node should exist - it is easy to filter them out but hard to add them when they are removed

From @ahsonkhan
> Is this necessary? I am assuming the usage of Whitespace would be to maintain the input formatting in some way? What should the ReadWhitespace API return?

PR #335, opened 2015-10-16
Initial implementation of System.Threading.Tasks.Channels

stephentoub:
Prototype of a new System.Threading.Tasks.Channels library, which provides data structures for handing off data between producers and consumers.  The included [README.md](https://github.com/stephentoub/corefxlab/blob/initial_channels/src/System.Threading.Tasks.Channels/README.md) provides an overview of the library and its exposed functionality.

An initial set of functional tests are included, with ~95% line and branch coverage.

Further work is desired to expose more operations over channels, provide additional built-in channel implementations, and optimize the existing channel and operator implementations further.


stephentoub:
cc @KrzysztofCwalina, @ljw1004, @Petermarcu, @MadsTorgersen, @bleroy

StephenCleary:
I think this line in the README:

    ```WaitToRead```/```WriteAsync```

would be less confusing as:

    ```WaitToReadAsync```/```WaitToWriteAsync```

Also, I would really like to see a separate `Deque` collection proposal. .NET has long needed one, as illustrated by the `SimpleQueue` included here and other places. Ditto for IAsyncEnumerator - it feels as though that would be a type with wider application than Channels.

stephentoub:
> I think this line in the README

Yes, that's just a typo, I'll fix it.

> Ditto for IAsyncEnumerator

Yes, I included it here simply as a convenient place to put it.  If it were to be included in the framework, it would certainly be put somewhere more general, especially with language support depending on it.

> as illustrated by the SimpleQueue included here and other places

Didn't quite understand this... SimpleQueue isn't a deque, it's just a regular queue without some of the (small) overheads of ```Queue<T>```.

StephenCleary:
> Didn't quite understand this... SimpleQueue isn't a deque, it's just a regular queue without some of the (small) overheads of Queue<T>.

You're right, of course. Not sure why I was thinking it was a deque. :)

src/System.Threading.Tasks.Channels/src/System/Threading/Tasks/Channels/Channel.EnumerableChannel.cs
```diff
@@ -0,0 +1,108 @@
...
+{
+    public static partial class Channel
+    {
+        /// <summary>Provides a channel around an enumerable.</summary>
+        private sealed class EnumerableChannel<T> : IReadableChannel<T>
+        {
+            /// <summary>A task completed when the channel is done.</summary>
+            private readonly TaskCompletionSource<VoidResult> _completion = new TaskCompletionSource<VoidResult>(TaskCreationOptions.RunContinuationsAsynchronously);
+            /// <summary>The wrapped enumerator.</summary>
+            private readonly IEnumerator<T> _source;
```
MrJul: This enumerator is never disposed. Shouldn't it be disposed when the channel is complete?

stephentoub: Thanks. Yup, it should. I'll fix that.

IanYates:
Just a quick note to say the link to the readme in the first post is now broken - it should be to https://github.com/stephentoub/corefxlab/tree/master/src/System.Threading.Tasks.Channels (and possibly having README.md tack on there if you wish) instead of https://github.com/stephentoub/corefxlab/blob/initial_channels/src/System.Threading.Tasks.Channels/README.md

stephentoub:
Thanks, @IanYates.  I've now fixed the link.

i3arnon:
I think a sentence or two on what's `ValueTask<T>` and what is it for would be nice as It's part of the public API but isn't a common concept (as far as I know).

stephentoub:
> I think a sentence or two on what's ValueTask<T>

Sure, I can add that.

>  Am I missing something?

Probably not.  I likely just neglected to remove the ```unsafe``` annotation after some experiments with different implementations.  I'll take a look and remove them assuming it's possible.

PR #344, opened 2015-10-20
Fix build break in System.Slices

krwq:


KrzysztofCwalina:
At some point we should try to make the assembly CLS compliant. For now, it's fine as-is, i.e. looks good.

krwq:
It's only test assembly, System.Slices is still CLS Compliant.

PR #347, opened 2015-10-20
Making Spans similar

krwq:
Getting closer with similar public surface. Making sure that scenarios covered by the others tests are working (Span<T> tests work with ByteSpan and reverse)


KrzysztofCwalina:
looks good.

PR #351, opened 2015-10-22
Adding JsonReader perf tests and using a Resource file for sample json strings

ahsonkhan:


src/System.Text.Json/tests/PerfSmokeTests.cs
```diff
@@ -0,0 +1,261 @@
...
+        private const int NumberOfSamples = 3;
+        private const int ExpectedMemoryBenchMark = 3000000;
+
+        private static readonly bool OutputResults = false;
+
+        [Fact]
+        public void ReadBasicJson()
+        {
+            Output("====== TEST ReadBasicJson ======");
+            RunTest(TestJson.BasicJson, 100 + NumberOfIterations, ExpectedMemoryBenchMark*MemoryToleranceFactor);
```
KrzysztofCwalina: When are the resources read? I think you need to ensure that everything is loaded into memory (byte[]) before starting the times. Otherwise the test might have lots of noise associated with resource access.

src/System.Text.Json/tests/PerfSmokeTests.cs
```diff
@@ -0,0 +1,261 @@
...
+        private static Json ReadJson(string jsonString)
+        {
+            var json = new Json();
+            if (string.IsNullOrEmpty(jsonString))
+            {
+                return json;
+            }
+
+            var jsonReader = new JsonReader(jsonString);
+            var jsonObjectMain = new Object();
```
KrzysztofCwalina: Why are you doing all these allocations? (object, members, etc). I would imagine for a micro-benchmark we would want to simply read the JSON data and verify that what we read is expected. I don't think you want to actually create deserialized objects.

ahsonkhan: Yes agreed. I realized that after I started doing comparison tests with Json.NET. Will bring focus to this performance test to what we care about.

src/System.Text.Json/tests/PerfSmokeTests.cs
ahsonkhan: I did see a 1.5x difference in the test results between reading the resource into a string/UTF8string and then starting the test vs reading the resource string directly. I will address this.

Should we only support byte[] or Utf8Strings or can we allow  users to pass in a string?

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -293,15 +289,15 @@ private static void ReadJsonHelper(JsonReader jsonReader, ICollection<Members> j
 
         private static Value GetValue(ref JsonReader jsonReader)
         {
-            var value = new Value {Type = (Value.ValueType) jsonReader.GetValueType()};
-            var obj = jsonReader.ReadValue();
+            var value = new Value {Type = (Value.ValueType) jsonReader.GetJsonValueType()};
+            var obj = jsonReader.GetValue();
```
KrzysztofCwalina: I thought you will remove this method?

ahsonkhan: This is not part of the performance test. JsonReaderTests checks for correctness and hence I need to deserialize to the Json Object and use Json.ToString() to compare input json to output json.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -103,69 +72,68 @@ public Utf8String ReadString()
...
             }
 
-            if (nextByte == QuoteString)
+            if (nextByte == '"')
             {
-                return ValueType.String;
+                return JsonValueType.String;
             }
 
             throw new ArgumentOutOfRangeException();
```
KrzysztofCwalina: BTW, this is a strange exception to throw form a method that takes no arguments

src/System.Text.Json/tests/JsonReaderTests.cs
KrzysztofCwalina: Ah, ok. I would just feel much better if you removed this method, but it's ok to do it in a separate commit.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
ahsonkhan: Agreed.
Although, I was considering changing the API to take in an argument:
public Utf8String GetValue(JsonValueType type)

PR #353, opened 2015-10-22
Adding initial, prototype implementation for a value type BufferPool

sokket:
Adding initial, prototype implementation for a value type BufferPool.

/cc @KrzysztofCwalina

### DIFF
--- a/src/System.Buffers/src/System/Buffers/BufferBucket.cs
+++ b/src/System.Buffers/src/System/Buffers/BufferBucket.cs
@@ -1,0 +1,82 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Diagnostics;
+using System.Threading;
+
+namespace System.Buffers
+{
+    internal sealed class BufferBucket<T> where T : struct
+    {
+        private volatile int _index;
+        private T[][] _data;
+        private int _bufferSize;
+        private SpinLock _lock;
+
+        internal BufferBucket(int bufferSize, int numberOfBuffers)
+        {
+            _index = 0;
+            _bufferSize = bufferSize;
+            _lock = new SpinLock();
+            _data = new T[numberOfBuffers][];
+            for (int i = 0; i < numberOfBuffers; i++)
+            {
+                _data[i] = new T[bufferSize];
+            }
+        }
+
+        internal T[] Rent()
+        {
+            T[] buffer;
+
+            // Use a SpinLock since it is super lightweight
+            // and our lock is very short lived
+            bool taken = false;
+            _lock.Enter(ref taken);
+            Debug.Assert(taken);
+
+            // Be quick and do an unsynchronized read to see if all
+            // of our buffers have been used
+            if (_index >= _data.Length)
+            {
+                // Exit the lock as soon as possible
+                _lock.Exit();
+                buffer = new T[_bufferSize];
+            }
+            else
+            {
+                buffer = _data[_index];
+                _data[_index] = null;
+                _index++;
+
+                // Exit the lock as soon as possible
+                _lock.Exit();
+            }
+
+            return buffer;
+        }
+
+        internal void Return(ref T[] buffer)
+        {
+            // Use a SpinLock since it is super lightweight
+            // and our lock is very short lived
+            bool taken = false;
+            _lock.Enter(ref taken);
+            Debug.Assert(taken);
+
+            // If we have space to put the buffer back, do it. If we don't
+            // then there was a buffer alloc'd that was returned instead so
+            // we can just drop this buffer
+            if (_index == 0)
+                buffer = null;
+            else
+            {
+                _index--;
+                _data[_index] = buffer;
+            }
+
+            _lock.Exit();
+        }
+    }
+}
--- a/src/System.Buffers/src/System.Buffers.csproj
+++ b/src/System.Buffers/src/System.Buffers.csproj
@@ -21,8 +21,10 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="System\Buffers\ArrayBufferPool.cs" />
+    <Compile Include="System\Buffers\BufferBucket.cs" />
     <Compile Include="System\Buffers\BufferInternal.cs" />
     <Compile Include="System\Buffers\BufferPool.cs" />
+    <Compile Include="System\Buffers\NativeBufferPool.cs" />
     <Compile Include="System\Buffers\Pack.cs" />
     <Compile Include="System\ByteSpan.cs" />
     <Compile Include="System\Diagnostics\Precondition.cs" />
--- a/src/System.Buffers/src/System/Buffers/BufferPool.cs
+++ b/src/System.Buffers/src/System/Buffers/BufferPool.cs
@@ -1,89 +1,88 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading;
 
 namespace System.Buffers
 {
-    public unsafe class NativeBufferPool : IDisposable
+    public sealed class BufferPool<T> where T : struct
     {
-        byte* _memory;
-        int[] _freeList;
-        int _nextFree = 0;
-        readonly int _totalBytes;
-        readonly int _bufferSizeInBytes;
+        private const int NUMBER_OF_BUFFERS_IN_BUCKET = 50;
+        private int _maxBufferSize;
+        private BufferBucket<T>[] _buckets;
 
-        public NativeBufferPool(int bufferSizeInBytes, int numberOfBuffers)
+        // 2MB taken as the default since the average HTTP page is 1.9MB
+        // per http://httparchive.org/interesting.php, as of October 2015
+        public BufferPool(int maxBufferSizeInBytes = 2048000)
         {
-            _bufferSizeInBytes = bufferSizeInBytes;
-            _freeList = new int[numberOfBuffers];
-            _totalBytes = bufferSizeInBytes * numberOfBuffers;
-            _memory = (byte*)Marshal.AllocHGlobal(_totalBytes).ToPointer();
-        }
+            int maxBuckets = SelectBucketIndex(maxBufferSizeInBytes);
 
-        public ByteSpan Rent()
-        {
-            var freeIndex = Reserve();
-            if (freeIndex == -1) {
-                throw new NotSupportedException("buffer resizing not supported.");
-            }
-            var start = _bufferSizeInBytes * freeIndex;
-            return new ByteSpan(_memory + start, _bufferSizeInBytes);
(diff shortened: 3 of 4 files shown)


src/System.Buffers/src/System/Buffers/BufferData.cs
```diff
@@ -0,0 +1,20 @@
...
+
+using System;
+using System.Runtime.InteropServices;
+
+namespace System.Buffers
+{
+    public struct BufferData
+    {
+        internal byte[] _data;
+        internal bool _allocd;
```
VSadov: I do not think you need to track _allocd. You are using this to not accept back to the pool buffers that did not come from the pool, but I think as long as there is a space in the pool and the buffer is not larger than _maxBufferSize you could just take it. They are all just byte arrays.

src/System.Buffers/src/System/Buffers/BufferBucket.cs
```diff
@@ -0,0 +1,76 @@
...
+            }
+            else
+            {
+                // Use a SpinLock since it is super lightweight
+                // and our lock is very short lived
+                bool taken = false;
+                _lock.Enter(ref taken);
+                Debug.Assert(taken);
+
+                bd._data = _data[_index];
```
VSadov: Looks like a race. 
_index may have changed since we did  ``` if (_index + 1>= _data.Length)``` above

src/System.Buffers/src/System/Buffers/BufferBucket.cs
```diff
@@ -0,0 +1,76 @@
...
+
+        internal BufferBucket(int bufferSize, int numberOfBuffers)
+        {
+            _index = 0;
+            _bufferSize = bufferSize;
+            _lock = new SpinLock();
+            _data = new byte[numberOfBuffers][];
+            for (int i = 0; i < numberOfBuffers; i++)
+            {
+                _data[i] = new byte[bufferSize];
```
VSadov: Seems a bit generous to allocate all buffers upfront. We may need just one or two.
Perhaps these could be lazy allocated in Rent when actually needed?

src/System.Buffers/src/System/Buffers/BufferData.cs
sokket: That's a good idea, but what about when all the buffers are returned? Then we would have extra buffers with no space to put them...re-allocing the array for more space would be bad and adding special code to handle that case inside the spin lock seems like it would hurt perf.

Thoughts?

src/System.Buffers/src/System/Buffers/BufferBucket.cs
sokket: I went back and forth on this one and went with the upfront allocation so that all calls to Rent would (assuming we had free buffers of the requested size) be constant-time calls; if we had the buffers be lazy allocated, you would lose efficiency since you would have to go and allocate the buffers on first response. My thought was that it would be better to push the cost up-front and have constant-time rent calls, especially since the Pool itself is an instance-object...so they wouldn't create a pool with certain buffer sizes unless the caller knew they needed one.

src/System.Buffers/src/System/Buffers/BufferBucket.cs
sokket: Good catch, thanks...fixing now

src/System.Buffers/src/System/Buffers/BufferData.cs
VSadov: I'd just drop extra buffers on the floor. 
You are already dropping buffers when pool is oversubscribed. That is unavoidable since you have a fixed size pool but potentially unbounded demand for buffers (however unlikely).
I only suggest to drop under different conditions - not when a buffer was not from the pool originally, but when pool does not have space to accept it back.

sokket: Cool, good idea...changed to work that way instead

src/System.Buffers/src/System/Buffers/BufferBucket.cs
```diff
@@ -0,0 +1,82 @@
...
+
+            // Use a SpinLock since it is super lightweight
+            // and our lock is very short lived
+            bool taken = false;
+            _lock.Enter(ref taken);
+            Debug.Assert(taken);
+
+            // Be quick and do an unsynchronized read to see if all
+            // of our buffers have been used
+            if (_index + 1 >= _data.Length)
```
bartonjs: Why `+ 1` and `>=`, subtract one from both sides and get `if (_index > _data.Length)`, though since it's about to be a deref maybe you still want `>=`...

src/System.Buffers/src/System/Buffers/IBufferPool.cs
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace System.Buffers
+{
+    public interface IBufferPool<T>
```
KrzysztofCwalina: What's the reason for this abstraction? And if we need an abstraction, why not abstract class?

src/System.Buffers/src/System/Buffers/ByteBufferPool.cs
```diff
@@ -0,0 +1,86 @@
...
+            {
+#if DEBUG
+                System.Diagnostics.Debugger.Break();
+#endif
+                buffer = new byte[size];
+            }
+            else
+            {
+                buffer = _buckets[SelectBucketIndex(size)].Rent();
+                if (clearBuffer) Array.Clear(buffer, 0, buffer.Length);
```
KrzysztofCwalina: I wonder if Return should not have clear either or only. I think clear on return is more useful for code that does not want to leak secrets.

src/System.Buffers/src/System/Buffers/ByteBufferPool.cs
```diff
@@ -0,0 +1,86 @@
...
+        public ByteBufferPool(int maxBufferSizeInBytes = 2048000)
+        {
+            int maxBuckets = SelectBucketIndex(maxBufferSizeInBytes);
+
+            _maxBufferSize = maxBufferSizeInBytes;
+            _buckets = new BufferBucket[maxBuckets + 1];
+
+            for (int i = 0; i < _buckets.Length; i++)
+            {
+                _buckets[i] = new BufferBucket(GetMaxSizeForBucket(i), NUMBER_OF_BUFFERS_IN_BUCKET);
```
KrzysztofCwalina: This is a pretty large allocation. Could we make it lazy?

KrzysztofCwalina:
Also,

KrzysztofCwalina:
closed by mistake. sorry.

src/System.Buffers/src/System/Buffers/IBufferPool.cs
davidfowl: Because interfaces are cool

src/System.Buffers/src/System/Buffers/BufferBucket.cs
```diff
@@ -0,0 +1,82 @@
...
+            // and our lock is very short lived
+            bool taken = false;
+            _lock.Enter(ref taken);
+            Debug.Assert(taken);
+
+            // If we have space to put the buffer back, do it. If we don't
+            // then there was a buffer alloc'd that was returned instead so
+            // we can just drop this buffer
+            if (_index == 0)
+                buffer = null;
```
bartonjs: Why bother assigning the input parameter to null?

src/System.Buffers/src/System/Buffers/BufferBucket.cs
sokket: Good point :)

src/System.Buffers/src/System/Buffers/IBufferPool.cs
sokket: I originally made the abstraction for the ASP.NET case where they look to need byte[] and char[] buffers, but I think I can refactor this PR a little bit more to make it flexible for both without the abstraction

src/System.Buffers/src/System/Buffers/ByteBufferPool.cs
sokket: Good idea, I'll add it there. I think it's useful to have in both Rent and Return cases...for instance, if someone rents a buffer, the contract doesn't guarantee that the buffer is empty so if they are going to use the buffer to receive data then they probably want it empty; however, if they are going to fill it then they don't care if it is empty or not. Having a clear on both Rent and Return looks to give the maximum amount of flexibility with very minimal overhead

src/System.Buffers/src/System/Buffers/ByteBufferPool.cs
sokket: I went back and forth with the lazy initialization and I went with this way for a couple reasons:

1) Since the buffer pool is an instance class, if someone creates one then they know that they will need the buffers and it isn't initialized for them behind the scenes
2) In step with above, since we know someone needs one since it's created, I optimized for constant time (with the disclaimer of all buffers being full and needing to alloc one) Rent calls; the caller doesn't need to worry about if this Rent call will kick a Lazy initialization or not since the initialization cost is paid during initialization.
3) This is more a guess than the other two, but my guess on the usage pattern of a Buffer Pool is that it will probably be initialized during the program initialization with other book keeping (such as binding and listening to sockets or hitting the disk for I/O) so if the program needs a buffer pool then it's better (IMHO) to upfront the cost so you can have fast access when you really need perf later on doing important work.

Thoughts?

src/System.Buffers/src/System/Buffers/BufferBucket.cs
sokket: Ah I should make this a ref parameter so that the caller can't Return a buffer and then still have a reference and keep using it after returning

sokket:
Changed the Buffer Pool to be generic to any value type so that it can work with Byte[] or Char[], meaning it should help with ASP.NET. Also addressing feedback

sokket:
Any more feedback @bartonjs, @KrzysztofCwalina, @VSadov?

VSadov:
@sokket   Honestly, I am not sure I like the lock. Locks have a habit of not bothering you until there is a contention in a real-world scenario causing dramatic perf drops because of convoying.

I'd prefer a lockless approach as something that scales more predictably. Perhaps similar strategy to https://github.com/dotnet/roslyn/blob/master/src/Compilers/Core/SharedCollections/ObjectPool%601.cs

However, lock might be ok for initial implementation.

sokket:
@VSadov it should be super lightweight since it isn't a full lock, just a SpinWait. Since the code inside the spin is super tiny, it seemed like the simplest approach for the initial implementation until we get the full requirements of how we want this new data structure to look and act.

src/System.Buffers/src/System/Buffers/BufferBucket.cs
KrzysztofCwalina: I had that as a ref parameter before. It does not work really well. It's annoying and it does not allow returning buffers that are properties, e.g. BufferFormatter.Buffer. In the end I changed this API back to not ref.

KrzysztofCwalina:
I think it's good for the initial commit, i.e. good to go with me.

sokket:
Cool, thanks! Merging

PR #357, opened 2015-10-23
Use faster Buffer.BlockCopy

jkotas:
See https://github.com/dotnet/corefx/pull/3615 on why this is better.

### DIFF
--- a/src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
+++ b/src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
@@ -494,7 +494,7 @@ namespace System.Text.Utf8
             }
             else
             {
-                _bytes.CopyTo(buffer, 0);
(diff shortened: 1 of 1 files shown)


jkotas:
@KrzysztofCwalina PTLA

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
```diff
@@ -494,7 +494,7 @@ public void CopyTo(byte[] buffer)
             }
             else
             {
-                _bytes.CopyTo(buffer, 0);
```
stephentoub: Is there any further benefit to Buffer.BlockCopy, since these are byte arrays?

stephentoub:
LGTM

src/System.Text.Utf8/src/System/Text/Utf8/Utf8String.cs
jkotas: Thanks - changed it to Buffer.BlockCopy

PR #370, opened 2015-10-27
Initial open-sourcing commit of the MultiValueDictionary

ianhays:
The MultiValueDictionary is an experimental collection that was first released the summer of 2014 as a Nuget package. This commit releases the source code of the MultiValueDictionary tied into .NET Core with the intention of it being considered as an addition to System.Collections.Generic in CoreFX.

- All code is open sourced from the original package with only changes made to get it functioning. As such, some things will need to be changed to better match the style of CoreFX (e.g. the tests).
- I based the csproj, project.json, folder structure, and strings.resx heavily on that of System.Collections.Generic in CoreFX so that they may be easily merged in the future. This is why Strings.resx has some strings that aren't used by the MVD
- Addition of performance tests required an update of BuildTools from 70 to 107.
- The MVD released earlier had tie-in functions to ILookup like ConvertToLookup. I removed those with the intention of adding them as extension functions in ILookup instead of being in MultiValueDictionary. The reason for this is that System.Collections.Generic shouldn't reference System.Linq but System.Linq can reference System.Collections.Generic.
- The design decisions around the MVD were made at point where .NET was very different than it is today. Some things might be worth reconsidering now that .NET Core exists.

### DIFF
--- a/src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
+++ b/src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
@@ -1,0 +1,1102 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Collections.Generic
+{
+    /// <summary>
+    /// A MultiValueDictionary can be viewed as a <see cref="IDictionary" /> that allows multiple 
+    /// values for any given unique key. While the MultiValueDictionary API is 
+    /// mostly the same as that of a regular <see cref="IDictionary" />, there is a distinction
+    /// in that getting the value for a key returns a <see cref="IReadOnlyCollection{TValue}" /> of values
+    /// rather than a single value associated with that key. Additionally, 
+    /// there is functionality to allow adding or removing more than a single
+    /// value at once. 
+    /// 
+    /// The MultiValueDictionary can also be viewed as a IReadOnlyDictionary&lt;TKey,IReadOnlyCollection&lt;TValue&gt;t&gt;
+    /// where the <see cref="IReadOnlyCollection{TValue}" /> is abstracted from the view of the programmer.
+    /// 
+    /// For a read-only MultiValueDictionary, see <see cref="ILookup{TKey, TValue}" />.
+    /// </summary>
+    /// <typeparam name="TKey">The type of the key.</typeparam>
+    /// <typeparam name="TValue">The type of the value.</typeparam>
+    public class MultiValueDictionary<TKey, TValue> :
+        IReadOnlyDictionary<TKey, IReadOnlyCollection<TValue>>
+    {
+        #region Variables
+        /*======================================================================
+        ** Variables
+        ======================================================================*/
+
+        /// <summary>
+        /// The private dictionary that this class effectively wraps around
+        /// </summary>
+        private Dictionary<TKey, InnerCollectionView> dictionary;
+
+        /// <summary>
+        /// The function to construct a new <see cref="ICollection{TValue}"/>
+        /// </summary>
+        /// <returns></returns>
+        private Func<ICollection<TValue>> NewCollectionFactory = () => new List<TValue>();
+
+        /// <summary>
+        /// The current version of this MultiValueDictionary used to determine MultiValueDictionary modification
+        /// during enumeration
+        /// </summary>
+        private int version;
+
+        #endregion
+
+        #region Constructors
+        /*======================================================================
+        ** Constructors
+        ======================================================================*/
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MultiValueDictionary{TKey, TValue}" /> 
+        /// class that is empty, has the default initial capacity, and uses the default
+        /// <see cref="IEqualityComparer{TKey}" /> for <typeparamref name="TKey"/>.
+        /// </summary>
+        public MultiValueDictionary()
+        {
+            dictionary = new Dictionary<TKey, InnerCollectionView>();
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MultiValueDictionary{TKey, TValue}" /> class that is 
+        /// empty, has the specified initial capacity, and uses the default <see cref="IEqualityComparer{TKey}"/>
+        /// for <typeparamref name="TKey"/>.
+        /// </summary>
+        /// <param name="capacity">Initial number of keys that the <see cref="MultiValueDictionary{TKey, TValue}" /> will allocate space for</param>
+        /// <exception cref="ArgumentOutOfRangeException">capacity must be >= 0</exception>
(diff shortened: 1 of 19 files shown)


src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
```diff
@@ -0,0 +1,1102 @@
...
+        /// <summary>
+        /// The private dictionary that this class effectively wraps around
+        /// </summary>
+        private Dictionary<TKey, InnerCollectionView> dictionary;
+
+        /// <summary>
+        /// The function to construct a new <see cref="ICollection{TValue}"/>
+        /// </summary>
+        /// <returns></returns>
+        private Func<ICollection<TValue>> NewCollectionFactory = () => new List<TValue>();
```
KrzysztofCwalina: Should this be static?

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
```diff
@@ -0,0 +1,1102 @@
...
+        }
+
+        /// <summary>
+        /// Initializes a new instance of the <see cref="MultiValueDictionary{TKey, TValue}" /> class 
+        /// that is empty, has the default initial capacity, and uses the 
+        /// specified <see cref="IEqualityComparer{TKey}" />.
+        /// </summary>
+        /// <param name="comparer">Specified comparer to use for the <typeparamref name="TKey"/>s</param>
+        /// <remarks>If <paramref name="comparer"/> is set to null, then the default <see cref="IEqualityComparer" /> for <typeparamref name="TKey"/> is used.</remarks>
+        public MultiValueDictionary(IEqualityComparer<TKey> comparer)
```
KrzysztofCwalina: Could it take EqualityComparer<T> instead? It seems like doing interface dispatch calls in lookup could be a big penalty.

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
```diff
@@ -0,0 +1,1102 @@
...
+
+        /// <summary>
+        /// An inner class that functions as a view of an ICollection within a MultiValueDictionary
+        /// </summary>
+        private class InnerCollectionView :
+            ICollection<TValue>,
+            IReadOnlyCollection<TValue>
+        {
+            private TKey key;
+            private ICollection<TValue> collection;
```
KrzysztofCwalina: Why not use the concrete type here? Right now you doing two interface dispatch calls per access to the IROCollection

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
ianhays: The purpose of the NewCollectionFactory is to allow an instance of a MultiValueDictionary to be created that has a custom specified inner collection. I set it here to new List<TValue>() as the default, but I don't think it should be static. It would make sense that you may want a MultiValueDictionary of Lists as well as a MultiValueDictionary of HashSets in the same program. The static Factories starting on line 357 are an example of when this would be used.

 As an aside, if I remember correctly we also discussed removing the public NewCollectionFactory constructors in favor of a bool "AllowRepeatedValues" that would act as a simple function to choose between HashSet<TValue> and List<TValue> as the internal structure. It was decided on giving the user more control through the current pattern though.

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
ianhays: We could, but I'm not sure that we would want to. Wouldn't that limit the functionality of the method since it would constrain the user to making sure their comparer inherited from EqualityComparer when we only really need it to inherit from IEqualityComparer?

I modeled this code after the constructors for [Dictionary<TKey, TValue>](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Collections/Generic/Dictionary.cs) so whatever we decide would be similarly relevant there.

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
ianhays: This kinda goes back to your first comment about the NewCollectionFactory. It's all designed around that functionality. The reason I didn't use the concrete type here is because we don't actually know the type. All we know is that the NewCollectionFactory will return an ICollection<TValue>. The code that creates InnerCollectionViews (lines 615 and 644) calls our NewCollectionFactory Func to create a new backing collection for the InnerCollectionView.

These ICollections could be removed if we went the path of the "AllowRepeatedValues" instead of custom specification of the inner collections, but in the current state I think they need to be interface declarations.

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
KrzysztofCwalina: OK. Thanks for the explanation. I think it's fine for this initial checkin. But we should discuss at the API review if we think this feature (ability to change the type of the collection) is really needed /used. This feature has a non-trivial cost per instance of the dictionary and I am almost sure most of the instances use the default collection type.

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
KrzysztofCwalina: Why would anybody want to implement IEqualityComparer instead of inheriting from EqualityComparer?

KrzysztofCwalina:
BTW, please feel free to merge in, just open an issue to discuss the two design points at the API review.

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
ianhays: Hmm, I guess if they didn't know they were supposed to implement EqualityComparer? I mainly used IEqualityComparer for consistency with Dictionary<TKey, TValue>, but there's no other good reason why it couldn't be EqualityComparer instead.

src/System.Collections.Generic.MultiValueDictionary/src/System/Collections/Generic/MultiValueDictionary.cs
ianhays: I agree. It's a pretty specific feature for such a general collection and I it may just add unnecessary confusion to what should be a simple decision of "do I want to allow repeated TValues for a given TKey".

ianhays:
Thanks @KrzysztofCwalina will do

PR #377, opened 2015-10-28
Removed Span<T> and ReadOnlySpan<T> form System.Buffers

KrzysztofCwalina:
We will now be using the Span<T> defined in System.Slices
I converted all usages of ReadOnlySpan<T> to Span<T>. Once we add ReadOnlySpan<T>
to System.Slices, I will convert the formatting code back. I left some TODOs there.

Also, this change resulted in a 10-20% regression in formatting perf. I am pretty sure it's
due to unoptimized Span<T>.Set methods (extensions). There is TODO in the code (see TryCopyTo) to optimize it.


KrzysztofCwalina:
@krwq, @sokket, could you take a look?

src/System.Text.Formatting.Globalization/src/project.lock.json
```diff
@@ -1,5 +1,5 @@
 {
-  "locked": true,
+  "locked": false,
```
sokket: I thought we wanted to keep this true...is that not correct?

sokket: And below

sokket:
One question but otherwise LGTM

src/System.Text.Formatting.Globalization/src/project.lock.json
KrzysztofCwalina: I just changed the project.json file to not use wildcards, and so I think nuget changes locked to false as there is nothing to ever upgrade to.

PR #382, opened 2015-10-28
Update DNX and lock files

terrajobst:
@KrzysztofCwalina, this fixes this error:

```
error CS0731: The type forwarder for type 'System.Diagnostics.CodeAnalysis.SuppressMessageAttribute' in assembly 'mscorlib' causes a cycle
```

### DIFF
--- a/dir.props
+++ b/dir.props
@@ -11,7 +11,7 @@
   <!-- Build Tools Versions -->
   <PropertyGroup>
     <BuildToolsVersion>1.0.25-prerelease-00107</BuildToolsVersion>
-    <DnxVersion>1.0.0-beta5-12101</DnxVersion>
+    <DnxVersion>1.0.0-rc1-15838</DnxVersion>
     <DnxPackageName Condition="'$(DnxPackageName)' == '' and '$(OsEnvironment)'!='Unix'">dnx-coreclr-win-x86.$(DnxVersion)</DnxPackageName>
     <DnxPackageName Condition="'$(DnxPackageName)' == '' and '$(OsEnvironment)'=='Unix'">dnx-mono.$(DnxVersion)</DnxPackageName>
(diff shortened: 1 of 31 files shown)


KrzysztofCwalina:
Thanks for doing the tools update. I am not great at reading computer generated files, but a cursory scan of the changes looks good. :-)
Ship it!

davidfowl:
Lock files shouldn't be checked in.....

terrajobst:
@davidfowl Ha! :-) I just asked @ericstj what we should do and he said I should check them in and lock them.

So what's the recommendation?

davidfowl:
Never check them in. We're going to be removing the lock: true soon as well. Going to have to prepare core fx for that change

ericstj:
Right now we have to do this if we are using *'s in the project.json.  Otherwise you build and it may fail depending on what's in the feed (as you found with the un-locked files).
David is suggesting we stop using *s and don't check in the lock files.  I think that can be done later.  Ideally after we have some sort of tooling in DNX/Nuget to accomplish the equivalent scenario.

ericstj:
LGTM

terrajobst:
OK, then I'll simply follow what CoreFX does today. Thanks @davidfowl and @ericstj!

davidfowl:
Sure but labs doesn't float versions.... :)

ericstj:
Yeah it does: https://github.com/dotnet/corefxlab/blob/master/src/System.CommandLine/src/project.json

weshaggard:
I agree we must keep the lock files locked and checked when using * versions. 

@davidfowl once you guys remove the support for locked lock files are there plans to provide some mechanisms to update large sets of project.json files to newer versions? On top of that I assume you will no longer use lock in the name and provide a way to generate them outside of the source tree.

PR #393, opened 2015-10-30
Changing GetValue() API to GetValue(JsonValueType) + other optimizations

ahsonkhan:
- fixed perf test, assert allocation is 0.
- removed extra bound checks / exception throws.


src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -59,7 +62,7 @@ public void Dispose()
 
         public bool Read()
         {
-            var canRead = _index < _str.Length;
```
KrzysztofCwalina: Why this change? Is Utf8String.Length expensive? If yes, can we fix it in Utf8String?

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
```diff
@@ -107,17 +110,16 @@ public JsonValueType GetJsonValueType()
...
-                return JsonValueType.String;
+                return JsonValueType.Number;
             }
 
-            throw new ArgumentOutOfRangeException();
+            throw new FormatException("Invalid json, tried to read char '" + nextByte + "'.");
         }
 
-        public Utf8String GetValue()
+        public Utf8String GetValue(JsonValueType type)
```
KrzysztofCwalina: I know we discussed it, but now that I look at this, I think this API turned out to be very confusing. The user will have no idea why they need to pass the type given they get a string back.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
krwq: True, after I merge my changes with Span<byte> it should be no-op

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
ahsonkhan: I agree. The usability seems to over rule the small perf gain. I will revert back to GetValue() and explicitly call GetJsonValueType inside.

src/System.Text.Json/src/System/Text/Json/JsonReader.cs
ahsonkhan: I am going to revert back this change. Utf8String.Length was uncharacteristically slower because the test was using debug builds. Using release builds, there isn't much difference between Utf8String.Length and me saving the property myself.

KrzysztofCwalina:
looks good. Let's discuss the other issues separatelly

PR #395, opened 2015-10-30
Add support for dotnet.exe to change csc.exe

krwq:
This is step 1 to make dotnet.exe work with Utf8String literals


KrzysztofCwalina:
looks good

PR #398, opened 2015-10-31
Changing API back to GetValue()

ahsonkhan:


ahsonkhan:
Please note:
Using UTF8String.Length is 10% slower than keeping a _length property that is initialized during JsonReader construction. This is something that needs to be addressed.

> Raw data:
Using Utf8String.Length = 3.46ms/iteration
Using my own class member = 3.18ms/iteration

KrzysztofCwalina:
Why is that? Isn't Utf8String.Length get inlined?

ahsonkhan:
The IL is very similar and is doing exactly as expected. So, I decided to repeat the testing a few times and observed larger fluctuation in the results. The 5-10% difference can be attributed to this. There is no difference in performance between using Utf8String.Length or my own class member, statistically speaking.

Test Data:
>Using Utf8String.Length:
3.4
3.19
3.36
Avg: 3.32
Std dev: 0.11

>Using my own class member:
3.3
3.11
3.42
Avg: 3.28
Std dev: 0.16

KrzysztofCwalina:
Looks good!

PR #400, opened 2015-10-31
Slice/Span performance improvements

adamsitnik:
The optimizations I have introduced:
1. Making Length a readonly field instead of property. It gives a lot for the "for loop"
2. Simpler range check by casting index and length to unsigned integers to eliminate index >= 0 check
3. Removing Dispose from struct Enumerator to avoid compiler from putting foreach in try+finally block
4. Introducing new EnumeratorObject class that is created when the collection is being enumerated via the IEnumerable interface. Since IEnumerable.GetEnumerator() returns IEnumerator (interface) it would be costly if this was struct because invoking interface method on structures is expensive
5. Added inlining of indexer getter, which makes the EnumeratorObject inline the indexer method.
6. Added GetItemWithoutBoundariesCheck method, which is used in safe EnumeratorObject, where Improved MoveNext ensures that it is impossible to move outside boundaries. It eliminates single range check (stadard indexer)
7. Equals based on for loop which reduces number of method calls (enumerating via Enumerator calls to MoveNext and Current, looping just to indexer) and added generic Comparer to avoid boxing (but I had to add dependency to System.Collections to use EqualityComparer<T>.Default)

I have tested all these changes a long time ago when I added them to the original Slice repo of Joe Duffy

### DIFF
--- a/src/System.Slices/src/System/Span.cs
+++ b/src/System.Slices/src/System/Span.cs
@@ -1,9 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.Collections;
 using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace System
 {
@@ -13,14 +12,14 @@ namespace System
     /// to regular accesses and is a struct so that creation and subslicing do
     /// not require additional allocations.  It is type- and memory-safe.
     /// </summary>
-    public struct Span<T> : IEnumerable<T>, IEquatable<Span<T>>
+    public partial struct Span<T> : IEnumerable<T>, IEquatable<Span<T>>
     {
         /// <summary>A managed array/string; or null for native ptrs.</summary>
         readonly object _object;
         /// <summary>An byte-offset into the array/string; or a native ptr.</summary>
         readonly UIntPtr _offset;
         /// <summary>Fetches the number of elements this Span contains.</summary>
-        public int Length { get; private set; }
+        public readonly int Length;
 
         /// <summary>
         /// Creates a new span over the entirety of the target array.
@@ -162,6 +161,7 @@ namespace System
         /// </exception>
         public T this[int index]
         {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get
             {
                 Contract.RequiresInRange(index, Length);
@@ -299,27 +299,17 @@ namespace System
                 return false;
             }
 
-            Enumerator thisIt = this.GetEnumerator();
-            Enumerator otherIt = other.GetEnumerator();
-            while (true)
+            var nonBoxingComparer = EqualityComparer<T>.Default;
+            for (int i = 0; i < Length; i++)
             {
-                bool hasNext = thisIt.MoveNext();
-                if (hasNext != otherIt.MoveNext())
-                {
-                    return false;
-                }
-
-                if (!hasNext)
-                {
-                    return true;
-                }
-
-                // TODO: Fix it so it doesn't box (memcmp)
-                if (!thisIt.Current.Equals(otherIt.Current))
+                if (!nonBoxingComparer.Equals(
+                    this.GetItemWithoutBoundariesCheck(i), other.GetItemWithoutBoundariesCheck(i)))
                 {
                     return false;
                 }
             }
+
+            return true;
         }
 
         public override bool Equals(object obj)
@@ -333,63 +323,15 @@ namespace System
         }
 
         /// <summary>
-        /// Returns an enumerator over the span's entire contents.
+        /// Returns item from given index without the boundaries check
+        /// use only in places where moving outside the boundaries is impossible
+        /// gain: performance: no boundaries check (single operation) 
         /// </summary>
-        public Enumerator GetEnumerator()
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private T GetItemWithoutBoundariesCheck(int index)
         {
-            return new Enumerator(this);
-        }
-
-        IEnumerator IEnumerable.GetEnumerator()
-        {
-            return GetEnumerator();
-        }
-
-        IEnumerator<T> IEnumerable<T>.GetEnumerator()
-        {
-            return GetEnumerator();
-        }
-
-        /// <summary>
-        /// A struct-based enumerator, to make fast enumerations possible.
-        /// This isn't designed for direct use, instead see GetEnumerator.
-        /// </summary>
-        public struct Enumerator : IEnumerator<T>
-        {
-            Span<T> _slice;    // The slice being enumerated.
-            int _position; // The current position.
-
-            public Enumerator(Span<T> slice)
-            {
-                _slice = slice;
-                _position = -1;
-            }
-
-            public T Current
-            {
-                get { return _slice[_position]; }
-            }
-
-            object IEnumerator.Current
-            {
-                get { return Current; }
-            }
-
-            public void Dispose()
-            {
-                _slice = default(Span<T>);
-                _position = -1;
-            }
-
-            public bool MoveNext()
-            {
-                return ++_position < _slice.Length;
-            }
-
-            public void Reset()
-            {
-                _position = -1;
-            }
+            return PtrUtils.Get<T>(
+                    _object, _offset + (index * PtrUtils.SizeOf<T>()));
         }
     }
 }
--- a/src/System.Slices/src/System/SpanEnumerators.cs
+++ b/src/System.Slices/src/System/SpanEnumerators.cs
@@ -1,0 +1,106 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System
+{
+    public partial struct Span<T>
+    {
+        public Enumerator GetEnumerator()
+        {
+            return new Enumerator(this);
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return new EnumeratorObject(this);
+        }
+
+        IEnumerator<T> IEnumerable<T>.GetEnumerator()
+        {
+            return new EnumeratorObject(this);
+        }
+
+        /// <summary>
+        /// A struct-based enumerator, to make fast enumerations possible.
+        /// This isn't designed for direct use, instead see GetEnumerator.
+        /// </summary>
+        public struct Enumerator
+        {
+            Span<T> _span;    // The slice being enumerated.
+            int _position; // The current position.
+
+            internal Enumerator(Span<T> span)
+            {
+                _span = span;
+                _position = -1;
+            }
+
+            public T Current
+            {
+                get { return _span[_position]; }
+            }
+
+            public bool MoveNext()
+            {
+                return ++_position < _span.Length;
+            }
+
+            public void Reset()
+            {
+                _position = -1;
+            }
+        }
+
+        /// <summary>
+        /// enumerator that implements <see cref="IEnumerator{T}"/> pattern (including <see cref="IDisposable"/>).
+        /// it is used by LINQ and foreach when Slice is accessed via <see cref="IEnumerable{T}"/>
+        /// it is reference type to avoid boxing when calling interface methods on stuctures
+        /// </summary>
+        private class EnumeratorObject : IEnumerator<T>
+        {
+            Span<T> _span;    // The slice being enumerated.
+            int _position; // The current position.
+
+            public EnumeratorObject(Span<T> span)
+            {
+                _span = span;
+                _position = -1;
+            }
+
+            public T Current
+            {
+                get { return _span.GetItemWithoutBoundariesCheck(_position); }
+            }
+
+            object IEnumerator.Current
+            {
+                get { return Current; }
+            }
+
+            public void Dispose()
+            {
+                _span = default(Span<T>);
+                _position = -1;
+            }
+
+            public bool MoveNext()
+            {
+                int nextItemIndex = _position + 1;
+                if (nextItemIndex < _span.Length)
+                {
+                    _position = nextItemIndex;
+                    return true;
+                }
+                return false;
+            }
+
+            public void Reset()
+            {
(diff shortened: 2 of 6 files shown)


KrzysztofCwalina:
Dzieki!
.

KrzysztofCwalina:
oops. closed by mistake.

VSadov:
I have been reviewing these changes in the original Slice repo and was waiting for reappearance here.
Good stuff !!
LGTM

src/System.Slices/src/System/SpanEnumerators.cs
```diff
@@ -0,0 +1,106 @@
...
+        {
+            return new EnumeratorObject(this);
+        }
+
+        /// <summary>
+        /// A struct-based enumerator, to make fast enumerations possible.
+        /// This isn't designed for direct use, instead see GetEnumerator.
+        /// </summary>
+        public struct Enumerator
+        {
```
VSadov: The nice part here is that foreach -pattern iterator does not implement IDisposable, so C# codegen for foreach will be much better.
Otherwise compiler would need to wrap the whole deal in try/finally to guarantee that Dispose is called. And that in turn would impact quality of JITed code as well.

src/System.Slices/src/System/Contract.cs
```diff
@@ -31,17 +31,26 @@ public static void Requires(bool condition)
...
         public static void RequiresInRange(int start, int length)
         {
-            if (!(start >= 0 && start < length))
+            // explicit unchecked block is used in order to avoid OverflowException when casting 
+            // negative int to unsigned int when the code is compiled with /checked option
+            unchecked
             {
-                throw new ArgumentOutOfRangeException();
+                // use unsigned int to reduce range check by one ( >= 0 )
+                if ((uint)start >= (uint)length)
```
whoisj: Does this do what you think it does? What is the value of `u` in the following? `int i = -1; uint u = (uint)i;` I'd assume `u == Uint32.MaxValue`.

VSadov: Seems like the helper assumes nonnegative length. It makes sense to use it in checks like index range checking  (since we know the span length is nonnegative by construction). In fact this is exactly the same technique as used by JIT for array index operations.
Perhaps it is worth adding Debug.Assert(length >= 0) ?

whoisj: Asserting the value is the least we should be doing here, but asserts are generally compiled out and this is a `public` entry-point. Seems dangerous to me. I'm not sure you can get away from a two part test scenario. `length >= 0 && (uint)start < (uint)length`.

There are many days when I wish CLS didn't care about signed-ness.

VSadov: I think Contract class is not public API.

For a public entry point Assert would not be sufficient indeed. In such case it would be worth having a separate private helper specifically for cases where we know for a fact that length is nonnegative. It is a very common case.

adamsitnik: I did some calculations. The minimal value that can be created out of casting signed negative integer to unsigned integer is 2,147,483,648. .NET has limit for single dimension array size which is 2,147,483,591 for bytes array and 2,146,435,071  for array of other types.

So the minimal value that can be generated is always bigger that the .NET's array size limit so the single comparision with Length should be always safe. 

What do you guys think?

whoisj: @adamsitnik is that a guaranteed limit forever on all platforms and version of the framework? If not, you're trading a large amount of risk for a very minimal performance gain.

adamsitnik: @whoisj The gain is small, but the indexer is probably going to be the "hottest path" here.

IMHO the size limit is so old that probably a lot of system parts rely on it and it is not going to change in the future. If it changes, the whole idea of Slice is to have small pieces of huge vector, so the Length should be much smaller than Max, because it makes no sense otherwise. And even if somebody abuses that in the future he or she is going to get some item from the collection instead of Exception so we keep the Memory Safety. 

List<T> also relies on that: http://referencesource.microsoft.com/#mscorlib/system/collections/generic/list.cs,0c65bec4d3fac21e

I think that we can keep it, but if you wish I can revert this change easily.

whoisj: I'm not the final call here - I'm just helping highlight potential issues :smile: 

Thanks for doing the homework - and I agree, every cycle matters on the hot path.

Can we get unit tests to prove this is the correct decision?

adamsitnik: @whoisj I am glad that I could help and learn a lot at the same time. Just added the tests. Do they convince you?

src/System.Slices/src/System/Span.cs
```diff
@@ -333,63 +323,15 @@ public override bool Equals(object obj)
...
-        /// <summary>
-        /// A struct-based enumerator, to make fast enumerations possible.
-        /// This isn't designed for direct use, instead see GetEnumerator.
+        /// Returns item from given index without the boundaries check
+        /// use only in places where moving outside the boundaries is impossible
+        /// gain: performance: no boundaries check (single operation) 
         /// </summary>
-        public struct Enumerator : IEnumerator<T>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private T GetItemWithoutBoundariesCheck(int index)
```
krwq: Should this be public + unsafe? We don't want people to do any hacks to achieve the same thing

src/System.Slices/src/System/SpanEnumerators.cs
krwq: This is really nice! Does compiler do that even if you leave the dispose empty and not get the results through IEnumerator?

mburbea: I think you don't need two separate enumerators. Can you not just implement the `IDisposable.Dispose` as an explicit implementation? Wouldn't that cause roslyn not to generate the call when it uses pattern matching?

VSadov: Once compiler sees that enumerator implements IDisposable it will be required to emit try/finally and call Dispose in there. Compiler cannot make assumptions about Dispose having no sideeffects.

VSadov: The other subtle thing here is that IEnumerator has stronger requirements for MoveNext. Implementation must continue returning "false" from MoveNext after iteration is over, no matter how many times called. (think of the overflowing the index). 
Foreach pattern does not impose such requirements. Since compiler will not call MoveNext redundantly, it is possible to have simpler and more efficient implementation.

If performance is super important, and it is in this case since we are dealing with fundamental scenario competing with array indexing, it would be reasonable to have separate enumerators.

adamsitnik: I have written some benchmarks to show why having two dedicated enumerators is better for performance. The code can be found here: https://github.com/adamsitnik/DemoCode/blob/master/Performance/Benchmarks/Collections/EnumeratorsBenchmarks.cs

The results (scroll down to summary): https://github.com/adamsitnik/DemoCode/blob/master/Performance/Benchmarks/results/EnumeratorsBenchmarks.txt

Summary:
1. Calling interface method on struct enumerators is very expensive. Looping with interface method calls compared to direct calls is x3 times slower on x86 and more than 10x slower on x64.
2.  On x86 looping with direct method calls to class enumerator is 15% faster than to direct with struct. But on the other hand on x64 the looping with direct calls to struct enumerator is more than 3x times faster than class.

That is why we have two dedicated enumerators here.

src/System.Slices/src/System/Contract.cs
whoisj: If the maintainers are fine, I'm fine. :wink:

KrzysztofCwalina: @adamsitnik, do you have any performance comparisons of the two alternatives (with and without >0 check)?

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: Let's wait with exposing this publicly. Unsafe on a method with no pointers in signature can be called in safe context, and the method when used incorrectly could violate memory safety.

src/System.Slices/src/System/SpanEnumerators.cs
KrzysztofCwalina: I was looking at the tests and the results and I don't seem to see a difference. I am looking at SingleStructEnumeratorCalledDirectly and DedicatedStructEnumeratorCalledDirectly  and they seem to be roughly equal in performance (similar ns/op)

KrzysztofCwalina: OK, I think I see it. You are comparing SingleStructEnumeratorCalledViaInterface vs DedicatedClassEnumeratorCalledViaInterface.

src/System.Slices/src/System/Contract.cs
```diff
@@ -31,17 +31,26 @@ public static void Requires(bool condition)
...
+                {
+                    throw new ArgumentOutOfRangeException();
```
KrzysztofCwalina: Could you pass the arg to the ctor?

src/System.Slices/src/System/Contract.cs
KrzysztofCwalina: BTW, from practical perspective, I would add the check for >0 and do a commit. Then create a separate PR to remove the check if we see the benefit/risk being large.

src/System.Slices/src/System/Span.cs
```diff
@@ -13,14 +12,14 @@ namespace System
...
-    public struct Span<T> : IEnumerable<T>, IEquatable<Span<T>>
+    public partial struct Span<T> : IEnumerable<T>, IEquatable<Span<T>>
     {
         /// <summary>A managed array/string; or null for native ptrs.</summary>
         readonly object _object;
         /// <summary>An byte-offset into the array/string; or a native ptr.</summary>
         readonly UIntPtr _offset;
         /// <summary>Fetches the number of elements this Span contains.</summary>
-        public int Length { get; private set; }
+        public readonly int Length;
```
KrzysztofCwalina: Can you get the same benefit by making the filed readonly and then wrapping in inlinable Length property?

src/System.Slices/src/System/Span.cs
```diff
@@ -299,27 +299,17 @@ public bool Equals(Span<T> other)
                 return false;
             }
 
-            Enumerator thisIt = this.GetEnumerator();
-            Enumerator otherIt = other.GetEnumerator();
-            while (true)
+            var nonBoxingComparer = EqualityComparer<T>.Default;
```
KrzysztofCwalina: @krwq (cc: @adamsitnik), this is not blocking this commit, but could we optimize this even more by simply comparing memory? Walking every single byte in a Span<byte> (probably the main use of Span<T>) seems expensive in comparison.

KrzysztofCwalina:
Nice changes. As I said in the relevant comment, I would just do a full range check and then do a separate PR later. And so besides this and a couple of small nits, it looks very good. 
Thanks!

src/System.Slices/src/System/SpanEnumerators.cs
VSadov: The issue is mostly that once C# sees IDisposable (which unavoidably comes with IEnumerator), it will have to wrap the whole thing into "using". 
It impacts performance of foreach enough that it makes sense to have a separate light-weight struct enumerator just for foreach pattern.

Then, once you have two iterators, the one that implements IEnumerator is better be a class because you have to allocate anyways (new or box) and interface dispatch on boxed structs is noticeable slower.

We came to exactly same conclusions when implementing ImmutableArray.Enumerator
https://github.com/dotnet/corefx/blob/master/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray%601%2BEnumerator.cs

VSadov: In ideal world, JIT would allocate a class iterator on the stack when possible (it seems possible in typical foreach), then devirtualize call to Dispose and notice that Dispose is actually empty and thus erase it as well as redundant try/finally.

In such case we could just have one enumerator that is a class. This is not the case today. :-(

src/System.Slices/src/System/Contract.cs
whoisj: > BTW, from practical perspective, I would add the check for >0 and do a commit. Then create a separate PR to remove the check if we see the benefit/risk being large.

:+1:

src/System.Slices/src/System/Span.cs
adamsitnik: @KrzysztofCwalina @krwq I think that it is unusual to have collection's Equal and GetHashCode method comparing and being dependent on the content. Maybe we should implement some SequenceEqual extension methods? One with the memory comparision for structs and other with IEqutable<T> constraint?

krwq: It feels like it should exist on the span. I feel like this type without any useful methods will be quite useless. I'm not convinced about implementing all the interfaces although memory comparison, enumerators etc. should be here IMO

src/System.Slices/src/System/Span.cs
adamsitnik: @KrzysztofCwalina  Unfortunately not, link to [benchmark](https://github.com/adamsitnik/DemoCode/blob/master/Performance/Benchmarks/Collections/LengthForLoopBenchmarks.cs) and [results](https://github.com/adamsitnik/DemoCode/blob/master/Performance/Benchmarks/results/LengthForLoopBenchmarks.txt)

src/System.Slices/src/System/Contract.cs
adamsitnik: I initially wanted to do this, but the problem is that this is an internal heper method, that might be called from public methods with different parameter names (start & index).

src/System.Slices/src/System/Contract.cs
adamsitnik: reverted, will be back in separate PR ;)

adamsitnik:
@KrzysztofCwalina I have squashed the commits, revert the problematic changes and fixed the compilation error that I had introduced before.

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: cc @krzysztofcwalina, it would be great to investigate why. But ok, let's use as a field first.

PR #415, opened 2015-11-05
Add sample for porting to .NET Core

richlander:
/cc @Petermarcu, @weshaggard, @davidfowl, @martinwoodward

### DIFF
--- a/samples/CoreApp.sln
+++ b/samples/CoreApp.sln
@@ -1,0 +1,46 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 14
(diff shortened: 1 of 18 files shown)


Petermarcu:
Looks good!

KrzysztofCwalina:
Rich, could you move this to the "demos" folder or rename the demos folder to "samples". After this commit, we now have two folders "demos" and "samples" that are for the same things. We should have one such folder.

KrzysztofCwalina:
Also, it would be great if either the commit message or a readme file explained what this sample illustrate, i.e. most other samples in the demo folder as CoreCLR apps.

davidfowl:
Can we add an xproj example?

PR #417, opened 2015-11-06
Fix for JsonReader where root is array and not object

ahsonkhan:


src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -230,11 +238,23 @@ private static Json ReadJson(string jsonString)
             var jsonReader = new JsonReader(jsonString);
             var jsonObjectMain = new Object();
             var jsonMembersMain = new List<Members>();
+            var jsonArrayMain = new Array();
+            var jsonElementsMain = new List<Elements>();
+
+            if (jsonString.Trim().Substring(0, 1) == "[")
```
krwq: 1. Trim is not that cheap as you think - it does need to iterate so it might be useful to store this partial result somewhere.
2. To compare I'd use .StartsWith(Utf8CodeUnit) overload - this one is pretty bad due to utf16 utf8 mixing (although it's not that slow)

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -287,6 +307,54 @@ private static void ReadJsonHelper(JsonReader jsonReader, ICollection<Members> j
             }
         }
 
+        private static void ReadJsonHelper(JsonReader jsonReader, ICollection<Elements> jsonElementsMain)
```
krwq: Please give it more meaningful name

krwq:
If json can start with an array or an object can it also be just a string or a number? Is this officially documented in the json spec? (if not is it what other parsers do?)

KrzysztofCwalina:
Aside from the few comets @krwq made, the change looks good!

ahsonkhan:
@krwq , the json can only start with an array or an object. It cannot be just a string or a number. I do believe this is documented in the json spec.

src/System.Text.Json/tests/JsonReaderTests.cs
ahsonkhan: This is only in the JsonReaderTest code and not in the actual JsonReader. I am not sure if we the performance matters to this scale there.
Also, will StartsWith(Utf8CodeUnit) skip white space?
For instance, will "   x123".StartsWith((Utf8CodeUnit)(byte)'x') return true?

ahsonkhan:
(previous commit - Adding runtime support to build for ubuntu/linux) Committed to the wrong branch/PR, and hence I reverted that commit.

krwq:
Please remove the two bottom commits and rebase the first one, I'll help if needed.

PR #427, opened 2015-11-06
Fixing groovy script to allow Debug and Release builds on CI

sokket:
Fixing groovy script to allow Debug and Release builds on CI, removing the Daily run from the Groovy script since it will be handled by VSO, and changing the packaging script to work with the new myget feed

/cc @mmitche


sokket:
test ci please

PR #433, opened 2015-11-09
shorten utf8string perf tests to around 1s per test

krwq:
This is running for too long now to actually be useful when writing code


KrzysztofCwalina:
looks good

PR #434, opened 2015-11-10
Removing old, unused buffers pools, adding a Native pool and renaming current pool

sokket:
Removing old, unused buffers pools. Also renaming the BufferPool class to be ManagedBufferPool, which returns arrays, and adding a NativeBufferPool which returns Span<T>. Both buffer pools use the same Rent-Return-MaxSize logic.

/cc @KrzysztofCwalina


src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
```diff
@@ -1,91 +1,73 @@
...
 
+using System;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading;
 
 namespace System.Buffers
 {
-    public unsafe class NativeBufferPool : IDisposable
+    public unsafe sealed class NativeBufferPool<T> where T : struct
```
sokket: Oops, forgot the dispose method to free the spans.

src/System.Slices/src/System/Span.cs
```diff
@@ -131,6 +131,12 @@ internal Span(object obj, UIntPtr offset, int length)
             Length = length;
         }
 
+        public void Clear()
```
sokket: Now that I think about it, this change isn't necessary since I can just set the callers reference to default(T) to clear it. Will revert

PR #435, opened 2015-11-10
Adding Json.Net and System.Text.Json comparison solution in expriments folder

ahsonkhan:
Added comparison tests for writer as well.


KrzysztofCwalina:
Could you add some information on the results? Maybe add it to the commit message. Also, some info on caveats, next steps, etc.

ahsonkhan:
Where should I put the results info along with the other info that was requested? It would be too long for a commit message.

I can create a GitHub readme/about page in the experiments directory which details the results. Would that work?

KrzysztofCwalina:
What about adding a readme explaining the project with some info on perf. Similar to the readme in https://github.com/dotnet/corefxlab/tree/master/src/System.CommandLine

KrzysztofCwalina:
looks good.

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonNetComparison.cs
```diff
@@ -0,0 +1,229 @@
...
+            TimingResultsJsonReader.Clear();
+        }
+
+        private static void RunTestJsonNet(string str, bool output)
+        {
+            var utf8Str = new Utf8String(str);
+            Timer.Restart();
+            for (var i = 0; i < NumberOfIterations; i++)
+            {
+                JsonNetReaderHelper(new StringReader(utf8Str.ToString()), output);
```
JamesNK: You're including the overhead of converting Utf8String to a string with Json.NET. Just pass it str.

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonNetComparison.cs
```diff
@@ -0,0 +1,229 @@
...
+                        if (output) Console.WriteLine(tokenType);
+                        break;
+                    case JsonReader.JsonTokenType.Property:
+                        var name = reader.GetName();
+                        if (output) Console.WriteLine(name);
+                        var x = reader.GetValue();
+                        if (output) Console.WriteLine(x);
+                        break;
+                    case JsonReader.JsonTokenType.Value:
+                        x = reader.GetValue();
```
JamesNK: If I understand the API correctly, GetValue returns a Utf8String. To be equivalent with what JsonTextReader is doing shouldn't these be converted to string/int/decimal/etc?

JamesNK:
Had a quick look and added a couple of comments. There are other things JsonTextReader is doing that JsonReader isn't like validating that JSON is valid, tracking line number and path in JSON for error messages. Also as far as I can tell JsonReader doesn't stream reading text. Won't loading a 400kb text file into memory every time it loads JSON have a pretty big effect on memory usage?

I released 8.0.1-beta2 onto NuGet a couple of days ago. It adds a BufferPool property to JsonTextReader/JsonTextWriter. Hook that up to System.Buffers and that should sort out JsonTextReader's memory usage.

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonNetComparison.cs
ahsonkhan: For testing and getting results, I removed that utf8string conversion.

```C#
        private static void RunTestJsonNet(string str, bool output)
        {
            //var utf8Str = new Utf8String(str);
            Timer.Restart();
            for (var i = 0; i < NumberOfIterations; i++)
            {
                JsonNetReaderHelper(new StringReader(str), output);
            }
            TimingResultsJsonNet.Add(Timer.ElapsedMilliseconds);
        }
```

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonNetComparison.cs
KrzysztofCwalina: I think conversions to int/decimal/etc make sense. Converting to System.String does not. The whole point of this JSON API is to be able to nativelly work with UTF8 buffers.

KrzysztofCwalina:
Yeah, I think it would make sense to add some more validation and better error reporting. 

I am not sure about streaming. There is a trade off here, but the assumption we are making is that it's cheaper to hold the data in pooled memory than to stream. Of course the assumption might be wrong for some scenarios.

@ahsonkhan, I think it would be great to add some info to these comparison tests explaining the differences in these two JSON APIs, i.e. explain where the perf differences come from.

JamesNK:
Unless you're expecting users to use Utf8String throughout their app then you'll be converting to string at some point.

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonNetComparison.cs
JamesNK: It returns an object but that object has already been converted into an end user value, e.g. string, int, long, DateTime, etc, that is set in the JObject DOM or the deserializing object. Casting != converting.

KrzysztofCwalina:
looks good.

Issue #356, opened 2015-10-23
Add a way to rent ByteSpan from an ambient BufferPool


KrzysztofCwalina:
I would like to use a buffer pool in Qotd service (demo app). But I need to be able to get ByteSpan from it. I could use NativeBufferPool, but it does not have Shared property, and so it's lots of busy setup code. Could we either add Shared property to NativeBufferPool or allow renting ByteSpan from ArrayBufferPool, or combine these two pools.

sokket:
Fixed with #436

PR #436, opened 2015-11-10
Adding a lazy-loading Shared buffer to both the Managed and Native pools

sokket:
Adding a lazy-loading Shared buffer to both the Managed and Native buffer pools to reduce the boilerplate needed to use the pools if desired.

/cc @KrzysztofCwalina 

#356

### DIFF
--- a/src/System.Buffers/src/System/Buffers/NativeBufferBucket.cs
+++ b/src/System.Buffers/src/System/Buffers/NativeBufferBucket.cs
@@ -9,13 +9,14 @@ using System.Threading;
 
 namespace System.Buffers
 {
-    internal unsafe sealed class NativeBufferBucket<T> where T : struct
+    internal unsafe sealed class NativeBufferBucket<T> : IDisposable where T : struct
     {
         private volatile int _index;
         private IntPtr _buffer;
         private Span<T>?[] _slices;
         private int _elementsInBuffer;
         private SpinLock _lock;
+        private bool _disposed;
 
         internal NativeBufferBucket(int elementsInBuffer, int numberOfBuffers)
         {
@@ -35,11 +36,29 @@ namespace System.Buffers
 
         ~NativeBufferBucket()
         {
+            Dispose(false);
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        private void Dispose(bool disposing)
+        {
             Marshal.FreeHGlobal(_buffer);
+            if (disposing)
+            {
+                _disposed = true; // don't touch in the finalizer
+            }
         }
 
         internal Span<T> Rent()
         {
+            if (_disposed)
+                throw new ObjectDisposedException("NativeBufferBucket");
+
             Span<T> buffer;
 
             // Use a lightweight spinlock for our super-short lock
@@ -67,6 +86,9 @@ namespace System.Buffers
 
         internal void Return(ref Span<T> buffer)
         {
+            if (_disposed)
+                throw new ObjectDisposedException("NativeBufferBucket");
+
             // Use a lightweight spinlock for our super short lock
             bool taken = false;
             _lock.Enter(ref taken);
--- a/src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
+++ b/src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
@@ -8,11 +8,31 @@ using System.Threading;
 
 namespace System.Buffers
 {
-    public unsafe sealed class NativeBufferPool<T> where T : struct
+    public unsafe sealed class NativeBufferPool<T> : IDisposable where T : struct
     {
         private const int NUMBER_OF_BUFFERS_IN_BUCKET = 50;
         private int _maxElements;
         private NativeBufferBucket<T>[] _buckets;
+        private bool _disposed;
+
+        private static NativeBufferPool<byte> _sharedInstance = null;
+
+        public static NativeBufferPool<byte> SharedByteBufferPool
+        {
+            get
+            {
+                if (Volatile.Read(ref _sharedInstance) == null)
+                {
+                    // 2MB taken as the default since the average HTTP page is 1.9MB
+                    // per http://httparchive.org/interesting.php, as of October 2015
+                    NativeBufferPool<byte> newPool = new NativeBufferPool<byte>(2048000);
+                    if (Interlocked.CompareExchange(ref _sharedInstance, newPool, null) != null)
+                        newPool.Dispose();
+                }
+
+                return _sharedInstance;
+            }
+        }
 
         public NativeBufferPool(int maxElementsInBuffer)
         {
@@ -21,8 +41,34 @@ namespace System.Buffers
             _buckets = new NativeBufferBucket<T>[maxBuckets + 1];
         }
 
+        ~NativeBufferPool()
+        {
+            Dispose(false);
+        }
+
+        public void Dispose()
+        {
+            Dispose(true);
+            GC.SuppressFinalize(this);
+        }
+
+        private void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                foreach (NativeBufferBucket<T> bucket in _buckets)
+                    if (bucket != null)
+                        bucket.Dispose();
+
+                _disposed = true;
+            }
+        }
+
         public Span<T> RentBuffer(int numberOfElements, bool clearBuffer = false)
         {
+            if (_disposed)
+                throw new ObjectDisposedException("NativeBufferPool");
+
             Span<T> buffer;
             if (numberOfElements > _maxElements)
             {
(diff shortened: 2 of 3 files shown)


src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
```diff
@@ -14,6 +14,23 @@ public unsafe sealed class NativeBufferPool<T> where T : struct
...
+
+        public static NativeBufferPool<byte> SharedByteBufferPool
+        {
+            get
+            {
+                if (Volatile.Read(ref _sharedInstance) == null)
+                {
+                    // 2MB taken as the default since the average HTTP page is 1.9MB
+                    // per http://httparchive.org/interesting.php, as of October 2015
+                    Interlocked.CompareExchange(ref _sharedInstance, new NativeBufferPool<byte>(2048000), null);
```
KrzysztofCwalina: If it's not successfully exchanged, don't you have to dispose the newely created pool?

sokket: The pool isn't IDisposable so it doesn't need to have a Dispose call; also wouldn't it get GC'd if it was not exchanged since no variable would have a reference to it?

KrzysztofCwalina: How do I delete the native memory, if the pool is not Disposable? It would get GCed, but then native memory would not be deleted except during finalization.

sokket: Gotcha, thanks for the (offline) chat about finalizers vs Disposable. Switching the Native pool over to use the IDisposable pattern

KrzysztofCwalina:
Though, this is not how you should implement Dispose. It's a bit convoluted, but here is the guidance: https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.110).aspx

whoisj:
@KrzysztofCwalina personally, I find that pattern terrible and thread-unsafe.

I've started using `Interlocked.Exchange` for these operations with a member variable tracking state. Usually something like

```
public void Dispose()
{
    IntPtr localHandle = IntPtr.Zero;
    if ((localHandle = Interlocked.Exchange(ref _nativeBuffer, IntPtr.Zero)) != IntPtr.Zero)
    {
        Marshal.FreeHGlobal(localHandle);
    }
}
```

Additionally, there's no guarantee that `Dispose()` will ever be called because it is doubtful this construct will be placed into a `using` statement. Therefore, the finalizer should call `Dispose` like

```
~NativeBufferBucket()
{
    Dispose();
}
```

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
```diff
@@ -21,6 +40,13 @@ public NativeBufferPool(int maxElementsInBuffer)
             _buckets = new NativeBufferBucket<T>[maxBuckets + 1];
         }
 
+        public void Dispose()
+        {
+            foreach (NativeBufferBucket<T> bucket in _buckets)
```
whoisj: what happens if `_buckets` is or becomes `null`? The implied contract is that `IDispose.Dispose()` should never throw.

Try 
```
var local = null;
if ((local = Interlocked.Exchange(ref _buckets, null)) != null)
{
    foreach (NativeBufferBucket<T> bucket in local)
        // same as before
}
```

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
```diff
@@ -21,6 +40,13 @@ public NativeBufferPool(int maxElementsInBuffer)
             _buckets = new NativeBufferBucket<T>[maxBuckets + 1];
         }
 
+        public void Dispose()
```
whoisj: Now that the object supports disposal, what happens if one of the member methods gets called?

I do not see any disposed state checking occurring there? We're likely to NRE in the library class. Better to detect the disposal and throw an `ObjectDisposedException` so the violator understands their error.

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
```diff
@@ -8,12 +8,31 @@
...
+    public unsafe sealed class NativeBufferPool<T> : IDisposable where T : struct
     {
         private const int NUMBER_OF_BUFFERS_IN_BUCKET = 50;
         private int _maxElements;
         private NativeBufferBucket<T>[] _buckets;
 
+        private static NativeBufferPool<byte> _sharedInstance = null;
+
+        public static NativeBufferPool<byte> SharedByteBufferPool
+        {
```
whoisj: how does this work?

alloc new pool if shared reference is null, assign pool to shared reference, **dispose pool**, return reference?

What am I missing?

Are we attempting to guard against a race condition causing multiple pools to be allocated? If so, then add an actual guard. Either use a mutex or set a value that an allocation is happening. I prefer a mutex because the second caller will block and always successfully return.

sokket: This algorithm works as follows:
1) Check if the shared instance is null; if it has a value, simply return it
2) If it is null, allocate a new one
3) Do an Interlocked.CompareExchange on the shared instance; if there was another thread that beat us to the CompareExchange (meaning by the time we got here, another thread already did the same thing but quicker) then dispose of the instance we created since it won't be used
4) Return the shared instance, which is guaranteed to be non-null now

A mutex is a very heavy hammer to solve this problem; Interlocked and Volatile are much lighter weight to achieve the same result, meaning we have a much smaller perf hit. This class is also trying to be as lock-free as possible

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
sokket: Buckets will never be null since it is constructed in the ctor and is never set to null anywhere, even in the dispose call...therefore, this will never throw a null reference exception

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
sokket: Good idea, added

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
whoisj: I'll argue that allocating 2MB of unnecessary memory is a heavier hammer.

Allocations are not free nor cheap. I firmly believe the mutex (use `Monitor` here for performance) is cheaper.

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
whoisj: In my example, I set `_buckets` to null. This guarantees that the native allocations are disposed *once* and it is lock free.

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
sokket: The 2MB buffer allocation *only* happens if the Shared pool gets a request for a 2MB buffer; 2MB is the *max* bucket size it will allocate. The initial construction of a Pool does not allocate any bucket or memory, therefore using any locking mechanism will be much heavier and go against the lock-free pattern we are trying to achieve

src/System.Buffers/src/System/Buffers/NativeBufferPool.cs
sokket: The only reason I can think of where the Interlock would be needed is during a double-dispose problem, in which case the caller has a logic error in their calling pattern anyway. The Dispose function should be called, at most, twice; once via the explicit Dispose call from the caller (by a ```using``` block or an explicit call) and once via the Finalizer, which will pass in False to the ```Dispose``` function, meaning nothing will happen. Having an interlocked operation and setting to null shouldn't be necessary

KrzysztofCwalina:
@whoisj, making Dispose thread safe might make sense. In general, .net APIs are not guaranteed to be thread safe (with the exception of statics), but it's ok to opt into thread safety. 

As to the finalized calling Dispose, it is not what we want APIs to do. There are many reasons for it, the main being that you should not be accessing managed objects from finalizers as they might be already finalized. Finalizer code path is very tricky to implement and should be super simple, ideally just a pinvoke call to release resource/handle.

src/System.Buffers/src/System/Buffers/NativeBufferBucket.cs
```diff
@@ -33,13 +34,26 @@ internal NativeBufferBucket(int elementsInBuffer, int numberOfBuffers)
             }
         }
 
-        ~NativeBufferBucket()
```
KrzysztofCwalina: I don't think the finalizer should be removed. I think we need both a finalizer and Dispose. The finalizer is an insurance in case dispose is not called.

sokket: So the Finalizer should check if we need to dispose and call Dispose if necessary?

KrzysztofCwalina: Finalizer should call Dispose(false). Dispose(bool) should have all cleanup logic. The "false" branch, should do minimum work to release the resources and should not call other managed object. It can call some statics. Ideally it should just make simple pinvoke calls (or equivalent) to release the resources.

sokket: Got it...updated to add the finalizers + call to Dispose(false) and made the NativeBufferBucket always call FreeHGlobal

KrzysztofCwalina:
looks good.

sokket:
Thanks! Once the CI passes I'll merge

sokket:
Thanks folks!

PR #437, opened 2015-11-11
update compiler to experimental version (krwq/roslyn)

krwq:
Please also take a look at what kind of compiler changes are being added:
https://github.com/krwq/roslyn/commit/8cb4528d5fc4354fd99350a793c8e26833ab5d1d


KrzysztofCwalina:
looks good.

sokket:
LGTM

mellinoe:
Wasn't this feature already cut in the past? Are we sure it's going to eventually get into the official spec like this? It would be a bit awkward if it got cut again but we were already using it here :smile:

PR #438, opened 2015-11-11
Automating the MyGet publishing with incrementing versions

sokket:
Rev'ing the nuspec version number for all packages

/cc @KrzysztofCwalina 
/fyi @davidfowl


nuget/System.Buffers.nuspec
```diff
@@ -2,7 +2,7 @@
 <package>
   <metadata>
     <id>System.Buffers</id>
-    <version>0.1.0-d103015-1</version>
+    <version>0.1.0-d103015-2</version>
```
KrzysztofCwalina: BTW, the d103015 is meant to represent October 30th, 2015. And so if we publish packages today, it should be d111115-1

KrzysztofCwalina:
Is the plan to manually update these every day?

nuget/System.Buffers.nuspec
sokket: Got it, thanks...I'll update the version numbers

sokket:
@KrzysztofCwalina to get around the manual update problem, I changed to use a powershell script that will format the date correctly and pass it to nuget via the arguments list; this will rev the version everyday automatically without manual intervention. I also removed the dependency version since we aren't version dependent (yet) and this should be fixed when we switch to autogenerated nuspec files, which can go in once Livar's changes go in

package.ps1
```diff
@@ -0,0 +1,19 @@
+﻿Push-Location $PSScriptRoot
+[Environment]::CurrentDirectory = $PWD
+
+$version = "0.1.0-d" + (Get-Date -Format FileDate) + "-2"
```
KrzysztofCwalina: Why is "-2" hardcoded here?

sokket: Oops, that should be -1 like we had before

KrzysztofCwalina:
looks good. Thanks.

sokket:
Thanks! Merged

nuget/System.Text.Utf8.nuspec
```diff
@@ -14,7 +14,7 @@
     <copyright> Microsoft Corporation.  All rights reserved.</copyright>
     <tags>.NET Core UTF8 corefxlab</tags>
     <dependencies>
-      <dependency id="System.Slices" version="0.1.0-d103015-1" />
+      <dependency id="System.Slices" />
```
davidfowl: This is a no no. *Always* specify a version for dependencies. You can use the $version$ token here as well.

sokket: Ah thanks, fixed via #444

PR #442, opened 2015-11-12
Update sample

weshaggard:
This updates the NetCoreSample to allow targeting ubuntu, osx, and windows. The default is to set to windows still but people can pick another target in the configuration manager.

### DIFF
--- a/samples/NetCoreSample/CoreApp/CoreApp.csproj
+++ b/samples/NetCoreSample/CoreApp/CoreApp.csproj
@@ -17,33 +17,11 @@
     <AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
     <NuGetTargetMoniker>DNXCore,Version=v5.0</NuGetTargetMoniker>
     <BaseNuGetRuntimeIdentifier>win7</BaseNuGetRuntimeIdentifier>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'">
-    <DebugSymbols>true</DebugSymbols>
-    <OutputPath>bin\x86\Debug\</OutputPath>
-    <DefineConstants>DEBUG;TRACE</DefineConstants>
-    <NoStdLib>true</NoStdLib>
-    <DebugType>full</DebugType>
-    <PlatformTarget>x86</PlatformTarget>
-    <ErrorReport>prompt</ErrorReport>
-    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
-    <Prefer32Bit>true</Prefer32Bit>
-    <UseVSHostingProcess>false</UseVSHostingProcess>
-  </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Release|x86'">
-    <OutputPath>bin\x86\Release\</OutputPath>
-    <DefineConstants>TRACE</DefineConstants>
-    <Optimize>true</Optimize>
-    <NoStdLib>true</NoStdLib>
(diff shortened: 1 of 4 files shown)


samples/NetCoreSample/CoreApp/CoreApp.csproj
```diff
@@ -17,33 +17,61 @@
     <AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
     <NuGetTargetMoniker>DNXCore,Version=v5.0</NuGetTargetMoniker>
     <BaseNuGetRuntimeIdentifier>win7</BaseNuGetRuntimeIdentifier>
+    <HostExe>CoreConsole.exe</HostExe>
   </PropertyGroup>
-  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x86'">
+  <PropertyGroup Condition="'$(Configuration)|$(Platform)' == 'Debug|x64'">
```
mellinoe: "Windows_Debug" ?

samples/NetCoreSample/CoreApp/CoreApp.csproj
```diff
@@ -17,33 +17,61 @@
...
+    <OutputPath>bin\x64.Debug.Win</OutputPath>
     <DefineConstants>DEBUG;TRACE</DefineConstants>
     <NoStdLib>true</NoStdLib>
     <DebugType>full</DebugType>
-    <PlatformTarget>x86</PlatformTarget>
+    <PlatformTarget>x64</PlatformTarget>
+    <UseVSHostingProcess>false</UseVSHostingProcess>
+    <ErrorReport>prompt</ErrorReport>
+    <CodeAnalysisRuleSet>MinimumRecommendedRules.ruleset</CodeAnalysisRuleSet>
+    <Prefer32Bit>true</Prefer32Bit>
```
mellinoe: Probably don't want that :smile:

samples/NetCoreSample/CoreApp/CoreApp.csproj
```diff
@@ -17,33 +17,61 @@
     <AddAdditionalExplicitAssemblyReferences>false</AddAdditionalExplicitAssemblyReferences>
     <NuGetTargetMoniker>DNXCore,Version=v5.0</NuGetTargetMoniker>
     <BaseNuGetRuntimeIdentifier>win7</BaseNuGetRuntimeIdentifier>
```
mellinoe: Should probably remove this one up here

mellinoe: or move it into the windows one, rather

weshaggard: By default I wanted to keep these the default windows, which is why I also just kept Debug|Release targeting windows as well without renaming them. This change was more about showing what is possible but I might do further clean-up of it later.

samples/NetCoreSample/CoreApp/CoreApp.csproj
weshaggard: I didn't explicitly add those it was just the default. In the case of using an explicit host this bit doesn't actually matter and will be ignored anyway. It only comes into play with the global system-wide mscoree host. At some other point I will likely streamline and clean up this project.

samples/NetCoreSample/CoreApp/project.json
```diff
@@ -1,10 +1,11 @@
 ﻿{
   "dependencies": {
-    "Microsoft.NETCore.Console": "1.0.0-beta-23504"
+    "Microsoft.NETCore.Console": "1.0.0-beta-23511"
   },
   "runtimes": {
```
davidfowl: This sucks hard :smile:

Issue #410, opened 2015-11-04
System.Drawing.Graphics tests all fail


sokket:
The Unit Tests in System.Drawing.Graphics are failing

tarekgh:
interesting, we have been awhile didn't touch anything there and the tests used to pass. anyway, I'll try to take a look later

tarekgh:
Fixed it.

PR #453, opened 2015-11-15
Fix the test failure in System.Drawing.Graphics

tarekgh:
The test used to use the native libgd.dll which is compiled for x86
while the test was running as x64 which couldn't load libgd.dll

the fix here is adding the native libgd.dll compiled for x64 and make the
test project detects which architecture is building for and then copy
the right DLL


tarekgh:
linking to the issue #410

PR #455, opened 2015-11-17
Slice: fast SequenceEqual & StructuralEqual

adamsitnik:
I have implemented fast SequenceEqual & StructuralEqual methods for Span and changed the behaviour of Equals and GetHashCode.

SequenceEqual compares all elements of Slice, it has generic constraint of IEqutable<T> to make fast comparison possible. Reason: [benchmarks](https://github.com/adamsitnik/DemoCode/blob/master/Performance/Benchmarks/Collections/SequentialEqualityBenchmarks.cs) and [results](https://github.com/adamsitnik/DemoCode/blob/master/Performance/Benchmarks/results/SequentialEqualityBenchmarks.txt) It does not call the memcmp to avoid any custom IEqutable implementations problems.

There are also new dedicated SequenceEqual methods for most common primitive types (byte, char, long etc). When such Span is small, simple for loop iteration is being done. If it is huge, then memcmp is called. Reason: pinning is expensive, as you can see in previous benchmark.

StructuralEqual simply casts two Spans of different types  to Span of same type and calls SequenceEqual. If size in bytes is multiplication of 8, then it casts to Span< long >, if % 4 then to Span< int >, if % 2 then Span< short >. Otherwise to Span< byte >. The reason is that comparing N long elements is faster than comparing N*8 byte elements (in a loop, optimization for small, but common slices)

The changed behaviour of Equals and GetHashCode is now the same as of any other .NET collection: simple ReferenceEquals. 

Moreover there is no ReferenceEquals check in SequenceEqual/StructuralEqual. I believe that Spans of bytes are going to be the most common ones, and if they are small, then the cost of such a check is relatively expensive. I also think that comparison of Spans referencing same memory is going to be very uncommon so I chose the certain gain in most common scenario with possible huge loss in most pessimistic case.  

Gain: the bigger the slice, the bigger the gain. 
For Spans of bytes: 
1 - 8 elements = no gain
16 elements = 1.11 times faster
64 elements = 1.42 times faster
512 elements = 2.7 times faster
4096 elements = 11 times faster
1048576 elements = 45 times faster


KrzysztofCwalina:
@krwq, could you review?

src/System.Slices/src/System/Span.cs
```diff
@@ -283,45 +288,30 @@ public bool ReferenceEquals(Span<T> other)
...
-            // TODO: Write something better
-            uint hash = unchecked((uint)Length);
-            foreach (T el in this)
+            var hashCode = Length;
+            unchecked
             {
-                hash = (hash >> 7) | (hash << 25);
-                hash ^= unchecked((uint)el.GetHashCode());
-            }
+                if (Object != null)
```
VSadov: unchecked  seems no longer needed

src/System.Slices/src/System/SpanExtensions.cs
```diff
@@ -162,6 +165,69 @@ public static Span<char> Slice(this string str, int start, int length)
...
+            {
+                return false;
+            }
+
+            var bytesCount = first.Length * PtrUtils.SizeOf<T>();
+            // pinvoke calls are expensive (mostly due to memory pinning) so we call memcmp only for big slices
+            if (bytesCount >= 512)
+            {
+                return MemoryUtils.MemCmp(first, second);
+            }
```
VSadov: is this correct if T overrides Equals?

src/System.Slices/src/System/SpanExtensions.cs
```diff
@@ -162,6 +165,69 @@ public static Span<char> Slice(this string str, int start, int length)
...
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        /// <summary>
+        /// Determines whether two spans are structurally equal by comparing the elements by using memcmp
+        /// </summary>
```
VSadov: is this a byte-wise comparison?

src/System.Slices/src/System/PtrUtils.cs
```diff
@@ -93,5 +93,20 @@ ldc.i4 0
...
+
+        /// <summary>
+        /// computes the address of object reference plus an offset in bytes
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        [ILSub(@"
+            .maxstack 2
+            .locals ([0] uint8& addr)
+            ldarg.0     // load the object
+            stloc.0     // convert the object pointer to a byref
```
adamsitnik: You are right! My CIL skills needs more sharpening like this!

src/System.Slices/src/System/Span.cs
adamsitnik: Sure! Thanks again for very insightful code review

src/System.Slices/src/System/SpanExtensions.cs
adamsitnik: @VSadov  Upss, it is not. Do you think that adding additional generic constraint <[Primitive]T> as used and described in Cast method would solve the issue?

VSadov: @adamsitnik As I understand [Primitive] guarantees that the type does not contain references. It still can be a struct that implements Equals in a nontrivial way.
I think it is unavoidable for SequenceEqual to do element-wise comparison. With easy-outs for cases of different length or complete match of both length and the object.

VSadov: If it is believed to be common enough it might be easier to introduce specialized/optimized overloads like ``` bool SequenceEqual(this Span<int> first, Span<int> second) ``` that could compare in optimized memcmp style.

adamsitnik:
@VSadov I have updated the code. Do you think it is ok now?

src/System.Slices/src/System/SpanEqualsExtensions.cs
```diff
@@ -0,0 +1,138 @@
...
+        /// </summary>
+        /// <param name="first">A span of doubles to compare to second.</param>
+        /// <param name="second">A span of doubles T to compare to first.</param>
+        public static bool SequenceEqual(this Span<double> first, Span<double> second)
+        {
+            if (first.Length != second.Length) { return false; }
+
+            // pinvoke calls are expensive (mostly due to memory pinning) so we call memcmp only for big slices
+            if (first.Length >= 256) { return MemoryUtils.MemCmp(first, second); }
+
```
VSadov: there are doubles like 0.0 and -0.0 which are equal, but not binary-equal. I do not know which way to compare them here, but there seems to be inconsistency. There could also be weirdness around NaN.
How is that handled in existing SequenceEqual implementations?

src/System.Slices/src/System/SpanExtensions.cs
```diff
@@ -162,6 +165,62 @@ public static Span<char> Slice(this string str, int start, int length)
...
+            return true;
+        }
+
+        /// <summary>
+        /// Determines whether two spans are structurally (byte-wise) equal by comparing the elements by using memcmp
+        /// </summary>
+        /// <param name="first">A span, of type T to compare to second.</param>
+        /// <param name="second">A span, of type U to compare to first.</param>
+        public static bool StructuralEqual<[Primitive]T, [Primitive]U>(this Span<T> first, Span<U> second)
+            where T : struct
```
VSadov: Not sure if this is the most intuitive name. I'd probably call it BlockEquals  (by vague similarity with BlockCopy)

src/System.Slices/src/System/MemoryUtils.cs
```diff
@@ -0,0 +1,43 @@
...
+
+using System.Runtime.InteropServices;
+
+namespace System
+{
+    static class MemoryUtils
+    {
+        [DllImport("msvcrt.dll", CallingConvention = CallingConvention.Cdecl)]
+        private static extern unsafe int memcmp(void* first, void* second, long sizeInBytes);
+
```
VSadov: Not sure if it worth it dealing with PInvokes here and msvcrt could be not portable enough.
I think there are ways to implement pretty decent memcmp entirely managed. There are some examples around BCL. Like String.EqualsHelper.

src/System.Slices/src/System/SpanEqualsExtensions.cs
adamsitnik: You are right, I should have not compare floating points in bitwise way. I have removed the SequenceEqual< double > extension method and added tests for that case.

src/System.Slices/src/System/MemoryUtils.cs
adamsitnik: I did some tests and the original memcmp is still the fastest way to go. So I just added new MemoryUtils.Other.cs file with platform independent MemCmp implementation based on the idea from String.EqualsHelper.

src/System.Slices/src/System/SpanExtensions.cs
adamsitnik: I am nor sure. Perhaps BitwiseEquals would be good? Or BlockEquals as you suggested?

src/System.Slices/src/System/MemoryUtils.Other.cs
```diff
@@ -0,0 +1,60 @@
...
+                    byte* firstPointer = (byte*)PtrUtils.ComputeAddress(first.Object, first.Offset).ToPointer();
+                    byte* secondPointer = (byte*)PtrUtils.ComputeAddress(second.Object, second.Offset).ToPointer();
+
+                    long bytesCount = first.Length * PtrUtils.SizeOf<T>();
+
+                    while (bytesCount >= 20)
+                    {
+                        if (*(int*)firstPointer != *(int*)secondPointer) break;
+                        if (*(int*)(firstPointer + 4) != *(int*)(secondPointer + 4)) break;
+                        if (*(int*)(firstPointer + 8) != *(int*)(secondPointer + 8)) break;
```
VSadov: Could it be advantageous to use void* here instead of int. Those also can be compared, but could be faster on 64bit.
```C#
            if (*(void**)firstPointer != *(void**)secondPointer) break;
```
you will need to scale the stepping by IntPtr.Size

src/System.Slices/src/System/MemoryUtils.Other.cs
```diff
@@ -0,0 +1,60 @@
...
+                var firstPinnedHandle = GCHandle.Alloc(first.Object, GCHandleType.Pinned);
+                var secondPinnedHandle = GCHandle.Alloc(second.Object, GCHandleType.Pinned);
+
+                try
+                {
+                    byte* firstPointer = (byte*)PtrUtils.ComputeAddress(first.Object, first.Offset).ToPointer();
+                    byte* secondPointer = (byte*)PtrUtils.ComputeAddress(second.Object, second.Offset).ToPointer();
+
+                    long bytesCount = first.Length * PtrUtils.SizeOf<T>();
+
```
VSadov: perhaps it is better to avoid long here. We can compute an upper limit for one of the ptrs and compare against that.

VSadov: You should get something like
```C#
            int step = sizeof(void*) * 5;
            byte* firstPointerLimit = (byte*)(first.Length * PtrUtils.SizeOf<T>() - step);

            while (firstPointer < firstPointerLimit)
            {
                if (*((void**)firstPointer + 0) != *((void**)secondPointer + 0)) return false;
                if (*((void**)firstPointer + 1) != *((void**)secondPointer + 1)) return false;
                if (*((void**)firstPointer + 2) != *((void**)secondPointer + 2)) return false;
                if (*((void**)firstPointer + 3) != *((void**)secondPointer + 3)) return false;
                if (*((void**)firstPointer + 4) != *((void**)secondPointer + 4)) return false;

                firstPointer += step;
                secondPointer += step;
            }

            // check the tail bytes here

            return true;
```
It might work faster. I did not test this code though :-)

adamsitnik: @VSadov  I took your sample code and all your hints and worked a little on. The result is platform independent MemCmp, which for x64 it is as fast as memcmp of msvcrt.dll, for x86 it is up to two times faster!!

In the next few hours my machines should finish executing full benchmarks, I will update them then.

src/System.Slices/src/System/Span.cs
```diff
@@ -283,45 +288,25 @@ public bool ReferenceEquals(Span<T> other)
 
         public override int GetHashCode()
         {
-            // TODO: Write something better
-            uint hash = unchecked((uint)Length);
-            foreach (T el in this)
+            var hashCode = Length ^ Offset.GetHashCode();
```
krwq: I don't think this is a good hashing function although if someone is using span as a key then they already have a problem :)

src/System.Slices/src/System/Span.cs
```diff
@@ -283,45 +288,25 @@ public bool ReferenceEquals(Span<T> other)
...
-            {
-                if (!nonBoxingComparer.Equals(
-                    this.GetItemWithoutBoundariesCheck(i), other.GetItemWithoutBoundariesCheck(i)))
-                {
-                    return false;
-                }
-            }
-
-            return true;
+            return ReferenceEquals(other);
```
krwq: In this case I think we should just remove reference equals

src/System.Slices/src/System/SpanEqualsExtensions.cs
```diff
@@ -0,0 +1,117 @@
...
+namespace System
+{
+    public static partial class SpanExtensions
+    {
+        /// <summary>
+        /// Determines whether two spans are equal (byte-wise) by comparing the elements by using memcmp
+        /// </summary>
+        /// <param name="first">A span of bytes to compare to second.</param>
+        /// <param name="second">A span of bytes T to compare to first.</param>
+        public static bool SequenceEqual(this Span<byte> first, Span<byte> second)
```
krwq: I feel like we are lacking some better templates in our language, feels like this should be fixed in the compiler and not in here

src/System.Slices/src/System/MemoryUtils.cs
```diff
@@ -0,0 +1,75 @@
...
+        /// platform independent fast memory comparison
+        /// for x64 it is as fast as memcmp of msvcrt.dll, for x86 it is up to two times faster!!
+        /// </summary>
+        internal static bool MemCmp<[Primitive]T>(Span<T> first, Span<T> second)
+            where T : struct
+        {
+            unsafe
+            {
+                // prevent GC from moving memory
+                var firstPinnedHandle = GCHandle.Alloc(first.Object, GCHandleType.Pinned);
```
krwq: is pinning worth for short spans? I'm not sure how expensive is doing it (could be close to nop too although worth checking)

src/System.Slices/src/System/MemoryUtils.cs
```diff
@@ -0,0 +1,75 @@
...
+                    byte* secondPointer = (byte*)PtrUtils.ComputeAddress(second.Object, second.Offset).ToPointer();
+
+                    int step = sizeof(void*) * 5;
+                    int totalBytesCount = first.Length * PtrUtils.SizeOf<T>();
+                    byte* firstPointerLimit = firstPointer + (totalBytesCount - step);
+
+                    if (totalBytesCount > step)
+                    {
+                        while (firstPointer < firstPointerLimit)
+                        {   // IMPORTANT: in order to get HUGE benefits of loop unrolling on x86 we use break instead of return
```
krwq: that sounds like compiler bug

src/System.Slices/src/System/MemoryUtils.cs
adamsitnik: It is not, but the method is never used for small spans https://github.com/adamsitnik/corefxlab/blob/master/src/System.Slices/src/System/SpanExtensionsTemplate.tt#L41

src/System.Slices/src/System/MemoryUtils.cs
adamsitnik: Do you mean the usage of break instead of return or the whole method itself ? ;)

src/System.Slices/src/System/Span.cs
```diff
@@ -335,7 +320,7 @@ public override bool Equals(object obj)
         /// gain: performance: no boundaries check (single operation) 
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private T GetItemWithoutBoundariesCheck(int index)
+        internal T GetItemWithoutBoundariesCheck(int index)
```
krwq: If we need it internally to implement extension method or anything then sounds like it should it be public as the public surface is insufficient to implement stuff (since this is incredibly unsafe method I would add unsafe in the name). cc: @KrzysztofCwalina

src/System.Slices/src/System/SpanExtensions.cs
krwq: I like BlockEquals

krwq:
Overall looks good to me :+1: Thank you!

src/System.Slices/src/System/Span.cs
adamsitnik: @krwq  I did some research: none of the existing .NET collections implements Equals as it's content equality check.  [ArraySegment](http://referencesource.microsoft.com/#mscorlib/system/arraysegment.cs,f7a911466af4caf1) [ImmutableArray](https://github.com/dotnet/corefx/blob/5f8f64437cf36f3baf0438b72a3987d3a527308f/src/System.Collections.Immutable/src/System/Collections/Immutable/ImmutableArray%601.cs#L1045)
Only the struct-based collections/wrappers do implement IEqutable interface at all. Mostly to avoid boxing when used in any scenartio that uses the default comparers (like List.Contains etc).

Enumerable offers quite popular (but very slow!) SequenceEquals extension method that allows to compare content of two IEnumerables. That is why I believe that it would be intuitive for users to call this method for content equality check. 

What do you think on such solution:
1. Keep the existing ReferenceEquals method. Add both == != operators that would call it.
2. Implement IEqutable interface in explicit way. There is no boxing, people don't use it by accident.
3. Keep the SequenceEquals extensions methods to get the best performance for IEqutable value types (most common scenario?).
4. Rename the proposed StructuralEquals to BlockEquals to get fast and both type and memory safe comparison for slices of different primitive types.

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: @ellismg, could you comment on the hashcode?

Since we implement IEquatable, we should override GetHashCode. But I have to say using spans as hash-table keys is something we should not encourage, especially that we might need to make Span<T> a stackonly type.

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: Yeah, I think we should remove (or hide those that cannot be removed) all O(n) comparison instance methods from Span<T>. I think it would be reasonable for these methods to be either static, or on a separate type.

src/System.Slices/src/System/Span.cs
KrzysztofCwalina: Hmm, I really don't like this method being internal. It's just so easy to misuse it. What are the implications of not making it internal (and leaving it private)?

src/System.Slices/src/System/Span.cs
ellismg: At a minimum this seems like a not great hashing function because we don't mix in the native pointer for spans over un-managed data, so all spans backed by native data with the same offset and length will have the same hashcode.

Maybe start from either the raw pointer (either truncated down to an int (this is what IntPtr.GetHashCode does) or XOR the high and low 32 bits together) or Object.GetHashCode() and then mix in the length and offset via multiplying the value by a prime (e.g. 33) and the adding in each piece of data.  This would be similar to how `string.GetHashCode()` works.

src/System.Slices/src/System/Span.cs
ellismg: I would keep this as reference equality testing, as you proposed.  If you want to have an easy way to do structural equality you could consider explicitly implementing `IStructuralEquatable`.

src/System.Slices/src/System/Span.cs
ellismg: Historically we've used the word "Dangerous" for these sorts of methods.  However, I don't think we should expose something like this as it makes it very easy to break type safety if `PtrUtils.Get<T>` doesn't do bounds checking.

adamsitnik: I just wanted to get the best possible performance (boundaries check is 20% time of  the getter). Previously this method was private and used only for Enumerator, but I wanted to use it in exension methods, so I made it internal.

Is it possible to make RyuJit do boundaries check elimination for Slice as it does for Array?

On the other hand, the method name itself gives clear declaration of what it does, so nobody should misuse it.

src/System.Slices/src/System/MemoryUtils.cs
krwq: Nice!

src/System.Slices/src/System/Span.cs
krwq: I think this is a valid scenario. If you name something dangerous then it is dangerous to use it and I wouldn't worry about bounds checks. What I see is that @adamsitnik can't implement extension on top of public surface on top of Span<T> because this is inefficient which sounds like a missing functionality on Span. If we can't implement it in our own library then no one can. Considering that span is acting as base for any data operations I call this should be public. Let's resolve this separately though. I think we should get rid of all "internal" in our libraries as they seem to be as pretty as "friend" in C++

krwq:
Thank you! This is really good PR :+1:

src/System.Slices/src/System/MemoryUtils.cs
krwq: I think the unrolling should be done by compiler :)

adamsitnik:
@krwq Dzięki!

GSPP:
The hash code is being built using `* 33` but this is probably a mistake because `33` is not prime. Did you mean `31`? This is in `Span.cs`.

GSPP:
Regarding expensiveness of pinning: Pinning through a GCHandle is indeed slow, but pinning through `fixed` is fast (nearly zero cost). What could be done here to improve performance is to code-gen one `fixed` comparison loop for each supported primitive type. Call those functions based on `switch (typeof(T).GetTypeCode())`.

adamsitnik:
@GSPP thanks for the info, I have just created a PR with [bug fix](https://github.com/dotnet/corefxlab/pull/473)

The hint with reducing the pinning price is really good! I am going to take a look at this later on.

PR #461, opened 2015-11-17
Add nuspec file for System.Time library

tarekgh:
This will allow building a nuget package for System.Time library

### DIFF
--- a/nuget/System.Time.nuspec
+++ b/nuget/System.Time.nuspec
@@ -1,0 +1,27 @@
+<?xml version="1.0"?>
+<package>
+  <metadata>
+    <id>System.Time</id>
+    <version>$version$</version>
+    <authors>Microsoft</authors>
+    <owners>Microsoft</owners>
+    <licenseUrl>http://go.microsoft.com/fwlink/?LinkId=329770</licenseUrl>
+    <iconUrl>http://go.microsoft.com/fwlink/?LinkID=288859</iconUrl>
+    <projectUrl>https://github.com/dotnet/corefxlab</projectUrl>
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Date, TimeOfDay</description>
+    <releaseNotes>Initial package</releaseNotes>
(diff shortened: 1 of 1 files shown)


nuget/System.Time.nuspec
```diff
@@ -0,0 +1,21 @@
...
+    <requireLicenseAcceptance>true</requireLicenseAcceptance>
+    <description>Date, TimeOfDay</description>
+    <releaseNotes>Initial package</releaseNotes>
+    <copyright> Microsoft Corporation.  All rights reserved.</copyright>
+    <tags>Date and Time helper classes</tags>
+  </metadata>
+  <files>
+    <file src="..\bin\Windows_NT.AnyCPU.Release\System.Time\System.Time.dll" target="lib/portable-net45+win8+wpa81+aspnetcore50"/>
+  </files>
+</package>
```
KrzysztofCwalina: I think you need to add the dependencies here. See how it's done in the System.Text.Formatting (which BTW is just an example)

tarekgh: the dependencies mentioned in System.Text.Formatting are only what is in corefxlab libraries which we don't have any of such dependency in System.Time. 
Looking at other nuspec System.CommandLine.nuspec and I am seeing no dependencies are mentioned there. 
I'll look at that to know if I need to specify the dependencies or not

KrzysztofCwalina: These are the dependecies I think you need:

    "System.Diagnostics.Contracts": "4.0.0-beta-*",
    "System.Diagnostics.Debug": "4.0.10-beta-*",
    "System.Globalization": "4.0.10",
    "System.Text.RegularExpressions": "4.0.10",
    "System.Xml.ReaderWriter": "4.0.10"

tarekgh:
@KrzysztofCwalina  @socket  could you please have another quick lock?

sokket:
LGTM

KrzysztofCwalina:
looks good

PR #473, opened 2015-11-30
fix for Span: using prime number in hashing function

adamsitnik:
In my previous PR I have changed the GetHashCode method. As @GSPP noticed [here](https://github.com/dotnet/corefxlab/pull/455#issuecomment-160470516) I have used the wrong number, 33 instead of 31. 33  is not a prime number, hence the fix.


KrzysztofCwalina:
looks good.

PR #495, opened 2015-12-20
Fix dependency on date and time format of test environment

alexanderkozlenko:
Tests should rely on invariant culture, because they are dependent on test environment customizations and can fail.


mj1856:
Agreed.  Thanks!

PR #496, opened 2015-12-20
Remove debugger break instruction

alexanderkozlenko:
The instruction makes debugging of projects, which relate on System.Buffers, quite inconvenient.


KrzysztofCwalina:
looks good. thanks!

PR #497, opened 2015-12-20
Fix dependency on time format of test environment

alexanderkozlenko:
Tests should rely on invariant culture or its derivative, because they are dependent on test environment customizations and can fail.


KrzysztofCwalina:
looks good. thanks!

PR #501, opened 2015-12-24
Add minor improvements for System.Time types

alexanderkozlenko:
1. Remove useless DebuggerDisplay attribute and dependency on System.Diagnostics.Debug package.
2. Use nameof expression to reference method for XmlSchemaProvider atribute.

### DIFF
--- a/src/System.Time/src/System/Date.cs
+++ b/src/System.Time/src/System/Date.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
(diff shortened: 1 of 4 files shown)


mj1856:
The `DebuggerDisplay` attribute was added to workaround the bug in VS2015 described [here](http://stackoverflow.com/a/31569625/634824).  I haven't followed through with the status of that bug since then.  Is there an update?  Thanks.

Also, `nameof` is fine, but if you have any better idea of how to ensure that `Date` and `TimeOfDay` are serialized as `xs:date` and `xs:time` respectively, I'd be happy to hear it.  I dislike how `GetSchema` has to be public for it to work.  It seems like there should be a simpler way.

PR #504, opened 2015-12-25
Improve extension methods names for DateTime type in System.Time

alexanderkozlenko:
Extension methods names for System.DateTime type, which returns System.Date and System.TimeOfDay types should start with "Get" (or, probably, "To") prefix to highlight that this is an extension method and not a native property. Moreover, Date() extension method name is very similar to Date property name.

### DIFF
--- a/src/System.Time/src/System/DateTimeExtensions.cs
+++ b/src/System.Time/src/System/DateTimeExtensions.cs
@@ -14,7 +14,7 @@ namespace System
         /// </summary>
         /// <param name="dateTime">The <see cref="DateTime"/> instance.</param>
         /// <returns>The <see cref="Date"/> value.</returns>
-        public static Date Date(this DateTime dateTime)
+        public static Date GetDate(this DateTime dateTime)
         {
             return new Date((int) (dateTime.Ticks / TimeSpan.TicksPerDay));
         }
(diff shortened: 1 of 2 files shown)


mj1856:
The original intent in the naming was for these to be discoverable near their non-extension counterparts which are the `Date` and `TimeOfDay` properties.  So:

```csharp
DateTime dt = // ...

DateTime d1 = dateTime.Date;  // existing property returning a DateTime
Date d2 = dt.Date; // existing property with implicit conversion to Date
Date d3 = dt.Date(); // extension method returning Date directly

TimeSpan t1 = dt.TimeOfDay; // existing property returning a TimeOfDay
TimeOfDay t2 = dt.TimeOfDay; // existing property with implicit conversion to TimeOfDay
TimeOfDay t3 = dt.TimeOfDay(); // extension method returning TimeOfDay directly
```

I'm okay with switching these to `GetDate` and `GetTimeOfDay` if others think that would be more understandable.  I'd like to hear some opinions.

The `To` prefix doesn't work for me though.  Both `To` and `As` imply some preservation of data just exposed differently, where in this case part of the `DateTime` is stripped away.

alexanderkozlenko:

alexanderkozlenko:
@dotnet-bot test Windows Debug

KrzysztofCwalina:
looks good to me. 

Having said that, we don't specially name extension methods, as opposed to "native" (member) methods. I think the Get prefix is desired because these are methods not properties, not because they are extensions.

mj1856:
Thanks!

Issue #510, opened 2015-12-26
Slices: Reduce size of code generated for Cotract methods


omariom:
PR: https://github.com/dotnet/corefxlab/pull/511

Two changes reduce size of machine code generated in **each** call site of Span's indexer by 27 bytes and eliminates one comparison and branch:
1. The trick used by array bounds checking
2. Wrapping creation of exception instance in a separate non-inlinable method.

Here is an example:
```C#
[MethodImpl(MethodImplOptions.NoInlining)]
public static long TestSpan(Span<long> span, int index)
{
    return span[index];
}
```
Before:
```asm
00007FF81D260C70  push        rsi  
00007FF81D260C71  sub         rsp,20h  
00007FF81D260C75  mov         eax,dword ptr [rcx+10h]  
00007FF81D260C78  test        edx,edx  
00007FF81D260C7A  jl          00007FF81D260C9A  
00007FF81D260C7C  cmp         edx,eax  
00007FF81D260C7E  jge         00007FF81D260C9A  
00007FF81D260C80  mov         rax,qword ptr [rcx]  
00007FF81D260C83  mov         rcx,qword ptr [rcx+8]  
00007FF81D260C87  shl         edx,3  
00007FF81D260C8A  movsxd      rdx,edx  
00007FF81D260C8D  add         rcx,rdx  
00007FF81D260C90  mov         rax,qword ptr [rax+rcx]  
00007FF81D260C94  add         rsp,20h  
00007FF81D260C98  pop         rsi  
00007FF81D260C99  ret  
00007FF81D260C9A  mov         rcx,7FF85F62C658h  
00007FF81D260CA4  call        00007FF87C8D4BB0  
00007FF81D260CA9  mov         rsi,rax  
00007FF81D260CAC  mov         rcx,rsi  
00007FF81D260CAF  call        00007FF85FC4DB90  
00007FF81D260CB4  mov         rcx,rsi  
00007FF81D260CB7  call        00007FF87CA035F0  
00007FF81D260CBC  int         3 
```
After the change:
```asm
00007FF81D280C70  sub         rsp,28h  
00007FF81D280C74  mov         eax,dword ptr [rcx+10h]  
00007FF81D280C77  cmp         edx,eax  
00007FF81D280C79  jae         00007FF81D280C94  
00007FF81D280C7B  mov         rax,qword ptr [rcx]  
00007FF81D280C7E  mov         rcx,qword ptr [rcx+8]  
[...]

PR #511, opened 2015-12-26
Slices: Reduce size of code generated for Contract methods.

omariom:
Issue: https://github.com/dotnet/corefxlab/issues/510 

Two changes reduce size of machine code generated in **each** call site of Span's indexer by 27 bytes and eliminates one comparison and branch:
1. The trick used by array bounds checking
2. Wrapping creation of exception instance in a separate non-inlinable method.

Here is an example:
```C#
public static long TestSpan(Span<long> span, int index)
{
    return span[index];
}
```
Before:
```asm
00007FF81D260C70  push        rsi  
00007FF81D260C71  sub         rsp,20h  
00007FF81D260C75  mov         eax,dword ptr [rcx+10h]  
00007FF81D260C78  test        edx,edx  
00007FF81D260C7A  jl          00007FF81D260C9A  
00007FF81D260C7C  cmp         edx,eax  
00007FF81D260C7E  jge         00007FF81D260C9A  
00007FF81D260C80  mov         rax,qword ptr [rcx]  
00007FF81D260C83  mov         rcx,qword ptr [rcx+8]  
00007FF81D260C87  shl         edx,3  
00007FF81D260C8A  movsxd      rdx,edx  
00007FF81D260C8D  add         rcx,rdx  
00007FF81D260C90  mov         rax,qword ptr [rax+rcx]  
00007FF81D260C94  add         rsp,20h  
00007FF81D260C98  pop         rsi  
00007FF81D260C99  ret  
00007FF81D260C9A  mov         rcx,7FF85F62C658h  
00007FF81D260CA4  call        00007FF87C8D4BB0  
00007FF81D260CA9  mov         rsi,rax  
00007FF81D260CAC  mov         rcx,rsi  
00007FF81D260CAF  call        00007FF85FC4DB90  
00007FF81D260CB4  mov         rcx,rsi  
00007FF81D260CB7  call        00007FF87CA035F0  
00007FF81D260CBC  int         3 
```
After the change:
```asm
00007FF81D280C70  sub         rsp,28h  
00007FF81D280C74  mov         eax,dword ptr [rcx+10h]  
00007FF81D280C77  cmp         edx,eax  
00007FF81D280C79  jae         00007FF81D280C94  
00007FF81D280C7B  mov         rax,qword ptr [rcx]  
00007FF81D280C7E  mov         rcx,qword ptr [rcx+8]  
00007FF81D280C82  shl         edx,3  
00007FF81D280C85  movsxd      rdx,edx  
00007FF81D280C88  add         rdx,rcx  
00007FF81D280C8B  mov         rax,qword ptr [rax+rdx]  
00007FF81D280C8F  add         rsp,28h  
00007FF81D280C93  ret  
00007FF81D280C94  call        00007FF81D280780  
00007FF81D280C99  mov         rcx,rax  
00007FF81D280C9C  call        00007FF87CA035F0  
00007FF81D280CA1  int         3  
```

### DIFF
--- a/src/System.Slices/src/System/Contract.cs
+++ b/src/System.Slices/src/System/Contract.cs
@@ -11,7 +11,7 @@ namespace System
         {
             if (!condition)
             {
-                throw new ArgumentException();
+                throw NewArgumentException();
             }
         }
 
@@ -19,35 +19,38 @@ namespace System
         {
             if (n < 0)
             {
-                throw new ArgumentOutOfRangeException();
+                throw NewArgumentOutOfRangeException();
             }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static void RequiresInRange(int start, int length)
         {
-            if (!(start >= 0 && start < length))
+            if ((uint)start >= (uint)length)
             {
-                throw new ArgumentOutOfRangeException();
+                throw NewArgumentOutOfRangeException();
             }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public static void RequiresInInclusiveRange(int start, int length)
         {
-            if (!(start >= 0 && start <= length))
+            if ((uint)start > (uint)length)
             {
-                throw new ArgumentOutOfRangeException();
+                throw NewArgumentOutOfRangeException();
             }
         }
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void RequiresInInclusiveRange(int start, int end, int length)
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static Exception NewArgumentException()
         {
-            if (!(start >= 0 && start <= end && end >= 0 && end <= length))
-            {
-                throw new ArgumentOutOfRangeException();
-            }
+            return new ArgumentException();
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static Exception NewArgumentOutOfRangeException()
+        {
+            return new ArgumentOutOfRangeException();
         }
     }
 }


alexanderkozlenko:
@dotnet-bot test Windows Debug

alexanderkozlenko:

alexanderkozlenko:
@dotnet-bot test Windows Debug

omariom:
@mikedn 

I am afraid it will reduce ability of JIT to do some optimizations.

I tried and here is the code generated:
```ASM
00007FF81D270C60  push        rdi  
00007FF81D270C61  push        rsi  
00007FF81D270C62  sub         rsp,28h  
00007FF81D270C66  mov         rsi,rcx  
00007FF81D270C69  mov         edi,edx  
00007FF81D270C6B  mov         eax,dword ptr [rsi+10h]  
00007FF81D270C6E  cmp         edi,eax  
00007FF81D270C70  jb          00007FF81D270C77  
00007FF81D270C72  call        00007FF81D270780  
00007FF81D270C77  mov         rax,qword ptr [rsi]  
00007FF81D270C7A  mov         rdx,qword ptr [rsi+8]  
00007FF81D270C7E  mov         ecx,edi  
00007FF81D270C80  shl         ecx,3  
00007FF81D270C83  movsxd      rcx,ecx  
00007FF81D270C86  add         rdx,rcx  
00007FF81D270C89  mov         rax,qword ptr [rax+rdx]  
00007FF81D270C8D  add         rsp,28h  
00007FF81D270C91  pop         rsi  
00007FF81D270C92  pop         rdi  
00007FF81D270C93  ret  
```

mikedn:
Right, this is one case where the lack of "no return" information negatively impacts code generation.

omariom:
@mikedn 
Ideally the part getting element value would be smth like this, right?

```ASM
mov         rax,qword ptr [rcx]  
mov         rcx,qword ptr [rcx+8]  
movsxd      rdx,edx 
mov         rax,qword ptr [rax+ rcx + rdx * 3]  
```
Is it possible to somehow make JIT generate it? )

mikedn:
> mov rax,qword ptr [rax+rcx+rdx*8]  

The memory operand isn't valid, only 2 registers can be used: `base_register + index_register * scale_constant + offset_constant`. `scale_constant` can only be 1, 2, 4 or 8.

Also, the index and return type of `SizeOf<T>` are `int` and that prevents the use of `scale_constant` to multiply with 8.  If `scale_constant` is used the operation is performed as `long` and that changes the meaning.

It looks to me that the use of the `int` in this case is actually incorrect because that multiplication can overflow and it does so after the range check. The consequence of an overflow in this code would probably be pretty serious. On 64 bit architectures the `index` should be sign extended to 64 bit before multiplication. The return of `SizeOf<T>` should be 64 bit as well to avoid an unnecessary extension to 64 bit.

omariom:
@mikedn
> The memory operand isn't valid, only 2 registers can be used

What a pity! )

Thanks for the detailed explanation.

VSadov:
The shorter bound checking sequence looks good.

The part with exceptions thrown in a separate method - I am not sure. Typically JIT places throwing code at the end of the method, assuming it is rarely touched. If that assumption is correct, the extra size does not matter that much. In such case the size of stack frame could matter more and in the new implementation it is slightly bigger (``` sub         rsp,28h```). I wonder if the trick with exception actually helps performance.

Another observation - since these exception objects have no specific data, could they just be readonly static instances in Contract?

mikedn:
>  Typically JIT places throwing code at the end of the method, assuming it is rarely touched.

Yes but since Span is generic it's useful to keep the code size small because the same code will be generated for all unshared generic instantiation.

> In such case the size of stack frame could matter more and in the new implementation it is slightly bigger ( sub rsp,28h ). 

Yes but there's no recursion in this code so it's unlikely that the frame size will have any measurable impact.

> since these exception objects have no specific data, could they just be readonly static instances in Contract?

Maybe but static field initialization might cause more problems than caching exception objects.

omariom:
> Another observation - since these exception objects have no specific data, could they just be readonly static instances in Contract?

That was my first thought ) By the size of the code gets bigger.

mikedn:
> Yes but since Span is generic it's useful to keep the code size small because the same code will be generated for all unshared generic instantiation.

Actually the generic aspect doesn't matter that much since this particular method is inlined. But the fact that it is inlined makes it even more important to keep the code size down.

mikedn:
> It looks to me that the use of the  int  in this case is actually incorrect because that multiplication can overflow and it does so after the range check.

This is a real bug, I created #516 for it. It only happens if `gcAllowVeryLargeObjects` is enabled, that's quite rare but tolerating memory corruption isn't an option.

omariom:
@mikedn 
Wow! I thought you meant my ''proposal" to x86 instruction set )

KrzysztofCwalina:
looks good.

omariom:
@mikedn 
>Right, this is one case where the lack of "no return" information negatively impacts code generation.

Will you create an issue in CoreCLR for that?

mikedn:
@omariom Will you create an issue in CoreCLR for that?

I briefly mentioned this issue in https://github.com/dotnet/coreclr/issues/1270. Not sure if a separate issue is needed, anyway there are lots of JIT code quality issue and I have no idea when they'll be addressed...

PR #512, opened 2015-12-26
Add escaping support for System.Text.Formatting

### LINKED ISSUE #61: Composite format should support escaping "{"
The following test will fail:

            var formatter = new StringFormatter();
            formatter.Format("{{{0}", 1);
            Assert.Equal("{1", formatter.ToString());

alexanderkozlenko:
Resolves #61

### DIFF
--- a/src/System.Text.Formatting/tests/CompositeFormattingTests.cs
+++ b/src/System.Text.Formatting/tests/CompositeFormattingTests.cs
@@ -50,7 +50,7 @@ namespace System.Text.Formatting.Tests
         }
 
         [Fact]
-        public void CompositeFormattingEscapingMissingStartBracket()
+        public void CompositeFormattingEscapingMissingEndBracket()
         {
             var pool = new ManagedBufferPool<byte>(1024);
             var formatter = new StringFormatter(pool);
@@ -59,7 +59,7 @@ namespace System.Text.Formatting.Tests
         }
 
         [Fact]
-        public void CompositeFormattingEscapingMissingEndBracket()
+        public void CompositeFormattingEscapingMissingStartBracket()
         {
             var pool = new ManagedBufferPool<byte>(1024);
             var formatter = new StringFormatter(pool);
--- a/src/System.Text.Formatting/tests/CompositeFormattingTests.cs
+++ b/src/System.Text.Formatting/tests/CompositeFormattingTests.cs
@@ -37,5 +37,34 @@ namespace System.Text.Formatting.Tests
 
             Assert.Equal("HelloaFF3", formatter.ToString());
         }
+
+        [Fact]
+        public void CompositeFormattingEscaping()
+        {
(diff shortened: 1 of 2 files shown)


src/System.Text.Formatting/tests/CompositeFormattingTests.cs
```diff
@@ -37,5 +37,34 @@ public void CompositeFormattingFormatStrings()
...
+
+        [Fact]
+        public void CompositeFormattingEscaping()
+        {
+            var format = "}}a {0} b {0} c {{{0}}} d {{e}} {{";
+            var pool = new ManagedBufferPool<byte>(1024);
+            var formatter = new StringFormatter(pool);
+            formatter.Format(format, 1);
+
+            Assert.Equal(string.Format(format, 1), formatter.ToString());
```
KrzysztofCwalina: Should it explicitly use the invariant culture?

alexanderkozlenko: It seems there is no way to specify culture for StringFormatter, thus StringFormatter and string.Format will use a derivative of invariant culture, which is defined for current thread in test class constructor.

KrzysztofCwalina: string.Format uses the "current culture", not invariant culture. I think the test should use: string.Format(CultureInfo.InvariantCulture, format, 1)

src/System.Text.Formatting/tests/CompositeFormattingTests.cs
```diff
@@ -37,5 +37,34 @@ public void CompositeFormattingFormatStrings()
...
+            var format = "}}a {0} b {0} c {{{0}}} d {{e}} {{";
+            var pool = new ManagedBufferPool<byte>(1024);
+            var formatter = new StringFormatter(pool);
+            formatter.Format(format, 1);
+
+            Assert.Equal(string.Format(CultureInfo.InvariantCulture, format, 1), formatter.ToString());
+        }
+
+        [Fact]
+        public void CompositeFormattingEscapingMissingStartBracket()
```
KrzysztofCwalina: Should this test be called ...MissingEndBracket?

src/System.Text.Formatting/tests/CompositeFormattingTests.cs
```diff
@@ -37,5 +37,34 @@ public void CompositeFormattingFormatStrings()
...
+        {
+            var pool = new ManagedBufferPool<byte>(1024);
+            var formatter = new StringFormatter(pool);
+
+            Assert.Throws<Exception>(() => formatter.Format("{{0}", 1));
+        }
+
+        [Fact]
+        public void CompositeFormattingEscapingMissingEndBracket()
```
KrzysztofCwalina: Should this test be called ...MissingStartBracket?

KrzysztofCwalina:
looks great! Thanks.

PR #513, opened 2015-12-27
Move strings to resources and add missing XML comment in PollingWatcher

alexanderkozlenko:


KrzysztofCwalina:
looks good. Thanks for doing all these small fixes!

PR #527, opened 2016-01-01
Slices: Make Span's Object and Offset fields internal

omariom:
They are readonly and of immutable types anyway.


KrzysztofCwalina:
What's the value of making them internal?

omariom:
Span currently has two internal properties Object and Offset. I think it makes little sense to have props wrapping read only fields of immutable types.

KrzysztofCwalina:
Ah, I missed the fact that we already have internal properties. Looks good.

PR #534, opened 2016-01-06
Slices: Implement IndexOf for Span<T>

omariom:
It is an implementation of IndexOf method with range check eliminated.
For Span of chars it serves as ordinal IndexOf.

```C#
public static int TestSpanIndexOf(Span<long> span, longvalue)
{
    return span.IndexOf(value);
}
```
The generated code for different element types.

**long**:
```ASM
; Assembly listing for method System.SpanExtensions:IndexOf(struct,long):int
; Lcl frame size = 0

       4C63C0               movsxd   r8d, eax
       4D85C0               test     r8, r8
       741C                 je       SHORT G_M9848_IG06
       4C8B09               mov      r9, gword ptr [rcx]
       488B4108             mov      rax, qword ptr [rcx+8]
       4C03C8               add      r9, rax
       33C0                 xor      rax, rax
G_M9848_IG03:
       493914C1             cmp      qword ptr [r9+8*rax], rdx
       7501                 jne      SHORT G_M9848_IG05
       C3                   ret      
G_M9848_IG05:
       48FFC0               inc      rax
       493BC0               cmp      rax, r8
       7CF1                 jl       SHORT G_M9848_IG03
G_M9848_IG06:
       B8FFFFFFFF           mov      eax, -1
       C3                   ret      

; Total bytes of code 44, prolog size 0 for method System.SpanExtensions:IndexOf(struct,long):int
```
**char**
```ASM
; Assembly listing for method System.SpanExtensions:IndexOf(struct,char):int
; Lcl frame size = 0
       4C63C0               movsxd   r8d, eax
       4D85C0               test     r8, r8
       7425                 je       SHORT G_M64671_IG06
       4C8B09               mov      r9, gword ptr [rcx]
       488B4108             mov      rax, qword ptr [rcx+8]
       4C03C8               add      r9, rax
       33C0                 xor      rax, rax
       81E2FFFF0000         and      edx, 0xFFFF
G_M64671_IG03:
       410FB70C41           movzx    rcx, word  ptr [r9+2*rax]
       3BCA                 cmp      ecx, edx
       7501                 jne      SHORT G_M64671_IG05
       C3                   ret      
G_M64671_IG05:
       48FFC0               inc      rax
       493BC0               cmp      rax, r8
       7CEE                 jl       SHORT G_M64671_IG03
G_M64671_IG06:
       B8FFFFFFFF           mov      eax, -1
       C3                   ret      
; Total bytes of code 53, prolog size 0 for method System.SpanExtensions:IndexOf(struct,char):int
```

**Guid**

```ASM
; Assembly listing for method System.SpanExtensions:IndexOf(struct,struct):int
; Lcl frame size = 56
       57                   push     rdi
       56                   push     rsi
       55                   push     rbp
       53                   push     rbx
       4883EC38             sub      rsp, 56
       488BF2               mov      rsi, rdx

       8B5110               mov      edx, dword ptr [rcx+16]
       4863FA               movsxd   edi, edx
       4885FF               test     rdi, rdi
       7441                 je       SHORT G_M17753_IG06
       488B19               mov      rbx, gword ptr [rcx]
       488B4908             mov      rcx, qword ptr [rcx+8]
       4803D9               add      rbx, rcx
       33ED                 xor      rbp, rbp
G_M17753_IG03:
       488BCD               mov      rcx, rbp
       48C1E104             shl      rcx, 4
       4803CB               add      rcx, rbx
       F30F6F06             movdqu   xmm0, qword ptr [rsi]
       F30F7F442428         movdqu   qword ptr [rsp+28H], xmm0
       488D542428           lea      rdx, bword ptr [rsp+28H]
       E8400E435E           call     System.Guid:Equals(struct):bool:this
       85C0                 test     eax, eax
       740B                 je       SHORT G_M17753_IG05
       8BC5                 mov      eax, ebp
       4883C438             add      rsp, 56
       5B                   pop      rbx
       5D                   pop      rbp
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      
G_M17753_IG05:
       48FFC5               inc      rbp
       483BEF               cmp      rbp, rdi
       7CCB                 jl       SHORT G_M17753_IG03
G_M17753_IG06:
       B8FFFFFFFF           mov      eax, -1
       4883C438             add      rsp, 56
       5B                   pop      rbx
       5D                   pop      rbp
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      

; Total bytes of code 101, prolog size 8 for method System.SpanExtensions:IndexOf(struct,struct):int
```

For comparison this is what we get for for **a  simple iteration over an array**:
```C#
        public static int TestSpanIndexOf(long[] array, long value)
        {
            for (int i = 0; i < array.Length; i++)
            {
                if (array[i] == value)
                    return i;
            }

            return -1;
        }
```

```ASM
; Lcl frame size = 0
       33C0                 xor      eax, eax
       448B4108             mov      r8d, dword ptr [rcx+8]
       4585C0               test     r8d, r8d
       7E13                 jle      SHORT G_M39245_IG06
G_M39245_IG03:
       4C63C8               movsxd   r9d, eax
       4A3954C910           cmp      qword ptr [rcx+8*r9+16], rdx
       7502                 jne      SHORT G_M39245_IG05
       C3                   ret      
G_M39245_IG05:
       FFC0                 inc      eax
       443BC0               cmp      r8d, eax
       7FEE                 jg       SHORT G_M39245_IG03
G_M39245_IG06:
       B8FFFFFFFF           mov      eax, -1
       C3                   ret      

; Total bytes of code 35, prolog size 0 for method CoreClrTest.Program:TestSpanIndexOf(ref,long):int
```

Benchmarked it with [BenchmarkDotNet](https://github.com/PerfDotNet/BenchmarkDotNet)
The code is [here](https://gist.github.com/omariom/e2806c35507550ce3153) - searching for the last symbol in a string of  223 chars length.

BenchmarkDotNet=v0.8.0.0
Type=Benchmarks  Mode=Throughput  Platform=HostPlatform  Jit=HostJit  .NET=HostFramework  toolchain=Classic  Runtime=Clr  Warmup=5  Target=10  

           Method |              op/s |
----------------- |--------------- |
 CharArrayIndexOf | 5,782,149.16 |
     SliceIndexOf  | 5,694,112.95 |
    StringIndexOf  | 5,521,419.05 | 

Actually on short distance and when it is in a local var Span is faster than string and char[] though not by much.

### DIFF
--- a/src/System.Slices/System/SpanExtensions.cs
+++ b/src/System.Slices/System/SpanExtensions.cs
@@ -317,6 +317,66 @@ namespace System
             return SequenceEqual(Cast<T, byte>(first), Cast<U, byte>(second));
         }
 
+        /// <summary>Searches for the specified value and returns the index of the first occurrence within the entire <see cref="T:System.Span" />.</summary>
+        /// <returns>The zero-based index of the first occurrence of <paramref name="value" /> within the entire <paramref name="slice" />, if found; otherwise, –1.</returns>
+        /// <param name="slice">The <see cref="T:System.Span" /> to search.</param>
+        /// <param name="value">The value to locate in <paramref name="slice" />.</param>
+        /// <typeparam name="T">The type of the elements of the slice.</typeparam>
+        [ILSub(@"   
+            .maxstack 3
+            .locals([0] uint8 & addr,
+                    [1] native uint i,
+                    [2] native uint length)
+            ldarg.0
+            ldfld      int32 valuetype System.ReadOnlySpan`1<!!T>::Length
+            dup
+            stloc.2
+            brfalse.s  EMPTY_SPAN
+ 
+            ldarg.0
+            ldfld      object valuetype System.ReadOnlySpan`1<!!T>::Object
+            stloc.0     
+            ldloc.0     
+            ldarg.0
+            ldfld      native uint valuetype System.ReadOnlySpan`1<!!T>::Offset
+
+            add         
+            stloc.0 
+
+            ldc.i4.0    
+            stloc.1
+            
+        LOOP_START:
+            ldloc.0
+            ldloc.1     
+            sizeof !!T  
+            conv.u  
+            mul         
+            add         
+            ldarg.1
+            constrained. !!T
+            callvirt   instance bool class [System.Runtime]System.IEquatable`1<!!T>::Equals(!0)
+            brfalse.s   NOT_EQUAL
+            ldloc.1     
+            ret 
+     
+        NOT_EQUAL:  
+            ldloc.1     
+            ldc.i4.1    
+            add         
+            stloc.1    
+            ldloc.1     
+            ldloc.2     
+            blt.s       LOOP_START
+        EMPTY_SPAN:
+            ldc.i4.m1 
+            ret")]
+        public static int IndexOf<T>(this ReadOnlySpan<T> slice, T value)
+           where T : struct, IEquatable<T>
+        {
+            return 0;
+        }
+
         // Helper methods similar to System.ArrayExtension:
 
         // String helper methods, offering methods like String on Slice<char>:
@@ -350,11 +410,6 @@ namespace System
             }
 
             return true;
-        }
-
-        public static int IndexOf(this ReadOnlySpan<char> str, char value)
-        {
-            throw new NotImplementedException();
         }
 
         public static int IndexOf(this ReadOnlySpan<char> str, string value)
--- a/tests/System.Slices.Tests/IndexOfTests.cs
+++ b/tests/System.Slices.Tests/IndexOfTests.cs
@@ -1,0 +1,106 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using Xunit;
+
+namespace System.Slices.Tests
+{
+    public class IndexOfTests
+    {
+        [Fact]
+        public void FindInUniqueValues()
+        {
+            var range = Enumerable.Range(0, 10).ToArray();
+
+            UniqueValuesSuccess(range);
+            UniqueValuesSuccess(range.Select(i => Guid.NewGuid()).ToArray());
+        }
+
+        [Fact]
+        public void FindInNonUniqueValues()
+        {
+            var range = Enumerable.Range(0, 10).ToArray();
+
+            NonUniqueValuesSuccess(range.Concat(range).ToArray());
+
+            var guids = range.Select(i => Guid.NewGuid()).ToArray();
+            NonUniqueValuesSuccess(guids.Concat(guids).ToArray());
+        }
+
+        [Fact]
+        public void DontFindNonExisting()
+        {
+            var range = Enumerable.Range(0, 10).ToArray();
+
+            NonExistingValuesFailure(
+                range.Take(5).ToArray(),
+                range.Skip(5).ToArray());
+
+            var guids = range.Select(i => Guid.NewGuid()).ToArray();
+            NonExistingValuesFailure(
+                guids.Take(5).ToArray(),
+                guids.Skip(5).ToArray());
+        }
+
+        private void UniqueValuesSuccess<T>(T[] values)
+            where T : struct, IEquatable<T>
+        {
+            Debug.Assert(values.Distinct().Count() == values.Length);
+
+            ReadOnlySpan<T> slice = values.Slice();
+
+            for (int i = 0; i < values.Length; i++)
+            {
+                T value = values[i];
+                int indexInArray = Array.IndexOf(values, value);
+
+                int index = slice.IndexOf(value);
+
+                Assert.True(index >= 0);
+                Assert.Equal(indexInArray, index);
+                
+            }
+        }
+
+        private void NonUniqueValuesSuccess<T>(T[] values)
+            where T : struct, IEquatable<T>
+        {
+            Debug.Assert(values.Distinct().Count() < values.Length);
+
+            ReadOnlySpan<T> slice = values.Slice();
+
+            for (int i = 0; i < values.Length; i++)
+            {
+                T value = values[i];
+                int indexInArray = Array.IndexOf(values, value);
+
+                int index = slice.IndexOf(value);
+
+                Assert.True(index >= 0);
+                Assert.Equal(indexInArray, index);
+            }
+        }
+
+        private void NonExistingValuesFailure<T>(T[] values, T[] nonExistingValues)
+            where T : struct, IEquatable<T>
+        {
+            Debug.Assert(!values.Intersect(nonExistingValues).Any());
+
+            ReadOnlySpan<T> slice = values.Slice();
+
+            for (int i = 0; i < nonExistingValues.Length; i++)
+            {
+                T value = nonExistingValues[i];
+                int indexInArray = Array.IndexOf(values, value);
+
+                int index = slice.IndexOf(value);
+
+                Assert.True(index == -1);
+                Assert.Equal(indexInArray, index);
+            }
+        }
+    }
+}


omariom:
I had to change ILSub to make it preserve ExtensionAttribute on rewritten methods.

omariom:
@mikedn  have a look please.

KrzysztofCwalina:
@krwq, @joeduffy

KrzysztofCwalina:
I will take a look at the details of the change later today, but the benchmark results and the codegen looks great! Thanks!

KrzysztofCwalina:
@vancem, @jkotas

VSadov:
LGTM
clever use of managed refs to avoid pinning

src/System.Slices/tools/ILSub/ILSub/ILSub.csproj
```diff
@@ -43,11 +43,13 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="Program.cs" />
-    <Compile Include="Properties\AssemblyInfo.cs" />
```
KrzysztofCwalina: Should the file be deleted as well?

src/System.Slices/tools/ILSub/ILSub/ILSub.csproj
```diff
@@ -43,11 +43,13 @@
...
   </ItemGroup>
   <ItemGroup>
     <None Include="App.config" />
   </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Properties\" />
+  </ItemGroup>
```
KrzysztofCwalina: And this removed?

src/System.Slices/tools/ILSub/ILSub/Program.cs
```diff
@@ -52,9 +52,7 @@ static Match FindILSubClass(string s)
             string[] tokens = new string[]
             {
                 @".class private auto ansi beforefieldinit System.ILSub",
-                "}",
-                "}",
-                "\n"
+                "} // end of class System.ILSub",
```
KrzysztofCwalina: @krwq, can you look into these changes?

src/System.Slices/tools/ILSub/ILSub/ILSub.csproj
omariom: Oh.. the file was absent and I removed to compile to tool. I planned to return the ref back and forgot about that.

src/System.Slices/tools/ILSub/ILSub/ILSub.csproj
omariom: It is left after removing non existent AssembyInfo.cs

mikedn:
:+1:

src/System.Slices/src/System/SpanExtensions.cs
```diff
@@ -226,6 +226,66 @@ public static bool SequenceEqual<T>(this Span<T> first, Span<T> second)
...
+            blt.s       LOOP_START
+        EMPTY_SPAN:
+            ldc.i4.m1 
+            ret")]
+        public static int IndexOf<T>(this Span<T> slice, T value)
+           where T : struct, IEquatable<T>
+        {
+            return 0;
+        }
+
```
KrzysztofCwalina: Do we have tests for this?

KrzysztofCwalina:
I like the change. I think we should check it in. Having said that, I really hope that we can do range check elimination in the backend compilers (i.e. specialize Span<T> just like we do T[]). We want the code gen to be the same, if people open code such loops over slices.

src/System.Slices/src/System/SpanExtensions.cs
omariom: I will add them.

src/System.Slices/tools/ILSub/ILSub/Program.cs
krwq: Tool changes look good, not sure about the IL part yet. If you checked that already it's good with me. I'll try to analyze it but it might take a while as I'm fairly newbie to IL

src/System.Slices/tools/ILSub/ILSub/Program.cs
```diff
@@ -275,16 +273,19 @@ static string RewriteIL(string input)
...
                 {
                     sb.Append(input, i, input.Length - i);
                     return sb.ToString();
                 }
 
                 sb.Append(input, i, methodBody.Index - i);
-                byte[] bytes = GetBytes(input, constructorBytes);
-                sb.Append(Encoding.UTF8.GetString(bytes, 4, bytes.Length - 6));
+
+                var ilSb = new StringBuilder(input, ilText.Index, ilText.EndIndex - ilText.Index, ilText.EndIndex - ilText.Index);
```
krwq: If GetBytes/IsHex/HexToByte is no longer used I'd remove it. I didn't know about the caverbal

src/System.Slices/src/System/SpanExtensions.cs
```diff
@@ -226,6 +226,66 @@ public static bool SequenceEqual<T>(this Span<T> first, Span<T> second)
...
+            brfalse.s   NOT_EQUAL
+            ldloc.1     
+            ret 
+     
+        NOT_EQUAL:  
+            ldloc.1     
+            ldc.i4.1    
+            add         
+            stloc.1    
+            ldloc.1
```
krwq: are those stloc+ldloc a nop? (I'm newbie to IL but they seem to be redundant)

krwq:
Thank you for this PR :+1: Both IL and tool changes look good to me.
I have not measured the perf but the code seems good. Except for inlining/SIMD for some specific types I don't see any way to further optimize it.

The only concern I have is the exe commited with the tool change. When I added initial version of it I made a mistake of pushing the exe as it was the easiest thing to do :( While I personally believe that this exe is fine I can see how this could go wrong (i.e. if your machine was infected) and as a precaution I believe that this exe should be produced as part of the build or one of the owners will need to rebuild it and merge this commit manually - both options are fine (no need for removing commit). Sorry about that.

omariom:
@krwq Thanks

src/System.Slices/src/System/SpanExtensions.cs
mikedn: They're needed, the stack based nature of IL tends to lead to such sequences. The index needs to be used in the next loop iteration so it is stored in a local. But storing pops the value from the stack and that value also needs to be used in the loop exit test that follows so you end up reloading what you just stored.

An alternative is 
```
dup 
stloc.1
``` 
but it shouldn't make any difference.

KrzysztofCwalina:
@krwq, yeah, please fix the issue with the exe, and then merge.
Thanks!

src/System.Slices/src/System/SpanExtensions.cs
VSadov: dup would be more interesting if we could avoid an IL temp, but here we already have a local anyways, so stloc.1/ldloc.1 is more preferable - just to make logical data flow simpler.
Actual storing/loading would depend on how JIT implements the local. If it is enregistered, which is likely, there will not be any storing or loading at all.

omariom: @VSadov Using dup vs locals can reduce the  frame size? In general, not in this case.

VSadov: The frame size will heavily depend on what JIT does. 
Overal, IL is just a serialization format for abstract code. Dup may make IL more compact and that is often the only advantage when dealing with optimizing JIT.

omariom:
@KrzysztofCwalina
:scream: I've just realized that IndexOf constraints (struct, IEquatable) may be too restrictive

omariom:
Added tests for IndexOf method.

I played with removing the constraints, but the approach may be a bit controversial so I  decided  to leave it  for now as is. It can be changed later.

krwq:
@omariom, does your last commit contain only merge from master? Can I remove it and rebase/merge?

omariom:
@krwq 
Yes. I should've probably rebase it. Not get accustomed fully to  Git  yet )

KrzysztofCwalina:
@dotnet-bot test this please

omariom:
Squashed the commits.

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@omariom, sorry that it is taking so long to merge this in. We are all very busy trying to ship .NET Core and then we broke corfxlab build for a couple of weeks. But now things are stable again. I am interested in merging this one in, but you would have to make it build and pass tests again.

omariom:
> but you would have to make it build and pass tests again.

No problemo )

omariom:
@KrzysztofCwalina

I can't see the details on why the build failed. Only [404](http://dotnet-ci.cloudapp.net/job/dotnet_corefxlab/job/Debug_prtest/321/).
 What could be the cause?

KrzysztofCwalina:
@omariom, maybe the test run is to old. Let me restart it.

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

omariom:
@KrzysztofCwalina 
The changed files include `src/System.Slices/tools/ILSub/ILSub/Program.cs`. It must be compiled to exe and added to the  change. 
How can it be done?

KrzysztofCwalina:
Could you point me at the updated program.cs. I will build the exe and check it in.

omariom:
https://github.com/omariom/corefxlab/blob/Slices-implement-simple-IndexOf/src/System.Slices/tools/ILSub/ILSub/Program.cs

KrzysztofCwalina:
The FindILSubInstance and FindILSubText methods are very fragile. For example, the ildasm I have on my machine does not generate cast to string, and so the match does not occur. Also, it does not generate ')}\r\n" at the end. When I change the code to just match ")" it matches comments. If I add newline, sometimes it fails as a newline happens in comments, etc.

@krwq, this is one hacky parser :-)

omariom:
@KrzysztofCwalina 

Did you use [caverbal](https://github.com/dotnet/corefxlab/pull/534/files#diff-47b38b592f6e7bc063917e03ae70a334) flag? 
I forgot to mention it.

KrzysztofCwalina:
Ah, that fixed it. I opened a PR #646

omariom:
@KrzysztofCwalina 
Should I rebase this PR  on it?

KrzysztofCwalina:
Yes, please rebase. I just merged the ILSub.exe change.

omariom:
@dotnet-bot test this please

PR #535, opened 2016-01-07
added native dependency to libuv package nuspec

KrzysztofCwalina:
@ericstj

### DIFF
--- a/nuget/System.Net.Libuv.nuspec
+++ b/nuget/System.Net.Libuv.nuspec
@@ -20,6 +20,7 @@
   </metadata>
   <files>
     <file src="..\bin\Windows_NT.AnyCPU.Release\System.Net.Libuv\System.Net.Libuv.dll" target="lib/portable-net45+win8+wpa81+aspnetcore50"/>
+    <file src="..\bin\Windows_NT.AnyCPU.Release\System.Net.Libuv\libuv.dll" target="runtimes/win7-x64/native"/>
   </files>
 </package>
 


KrzysztofCwalina:
CC: @livarcocc, @sokket

nuget/System.Net.Libuv.nuspec
```diff
@@ -20,6 +20,7 @@
   </metadata>
   <files>
     <file src="..\bin\Windows_NT.AnyCPU.Release\System.Net.Libuv\System.Net.Libuv.dll" target="lib/portable-net45+win8+wpa81+aspnetcore50"/>
+    <file src="..\bin\Windows_NT.AnyCPU.Release\System.Net.Libuv\libuv.dll" target="runtimes/win7-x64/native"/>
```
sokket: What about the OSX and Linux native binaries?

KrzysztofCwalina: Once we have these binaries, we will add them to the nuspec

ericstj: If you have multiple runtimes ideally you should split those into other packages so that folks on one platform don't have to download implementations that they'll never use.

davidfowl: Not worth it. We bundle libuv into kestrel and it's tiny (tho we leave our Linux)

davidfowl: Why not just copy what we do

KrzysztofCwalina: @davidfowl, do you use nuspec files?

KrzysztofCwalina: @ericstj, but does the change look good? i.e. should I commit?

ericstj: I think @davidfowl  just meant to mimic the structure of their package, which is a fine suggestion.  Have a look at their kestrel package.  If size is small you can ignore my split suggestion (which in retrospect is probably a bad idea since NuGet only follows runtime.json in the compile graph).

Here is the package @davidfowl was referring to: https://www.nuget.org/api/v2/package/Microsoft.AspNet.Server.Kestrel/1.0.0-rc1-final

ericstj: @KrzysztofCwalina it's ok as a partial fix but it'll only work for windows x64.  If it isn't much trouble I'd include the other platforms as well.

KrzysztofCwalina: OK. I will commit as-is and work on x64 and other platfroms separatelly. I just wanted this quick fix as somehow the package regressed.

KrzysztofCwalina: I added issue #538

Issue #537, opened 2016-01-07
Slices: Overflows


omariom:
There are possible overflows when addresses or number of items calculated  in a few places in Slices.
One is [described  here](https://github.com/dotnet/corefxlab/issues/516).
We should fix them.

omariom:
Overflow in Span.Cast https://github.com/dotnet/corefxlab/pull/608

ahsonkhan:
@omariom, is this still an issue?

omariom:
Closing..

PR #540, opened 2016-01-09
Slices: Fix oveflows in Span<T>.Slice methods

omariom:
Issue #537  
Both overloads of Span.Slice method allow to read at negative offsets.
Besides that they lack AggressiveInlining attribute what makes them slow.

I used the following benchmark code:
```C#
    public class BenchmarkSlice
    {
        Span<char> slice;

        public BenchmarkSlice()
        {
            string str = new string('c', 256);
            slice = str.Slice();
        }

        [Benchmark]
        public void Slice()
        {
            var copy = slice;
            
            do
            {
                copy = copy.Slice(1);
            }
            while (copy.Length != 0);
        }

        [Benchmark]
        public void SliceWithLength()
        {
            var copy = slice;
            do
            {
                copy = copy.Slice(1, copy.Length - 1);
            }
            while (copy.Length != 0);
        }
    }
```

**Original results (before the change)**

BenchmarkDotNet=v0.8.0.0
Type=BenchmarkSlice  Mode=Throughput  Platform=HostPlatform  Jit=HostJit  .NET=HostFramework  toolchain=Classic  Runtime=Clr  Warmup=5  Target=10

          Method |   AvrTime |    StdDev |       op/s |
---------------- |---------- |---------- |----------- |
           Slice | 1.6296 us | 0.1791 us | 620,547.63 |
 SliceWithLength | 1.8450 us | 0.1283 us | 544,803.54 |

**The results just after adding AggressiveInlining:**

          Method |     AvrTime |     StdDev |         op/s |
---------------- |------------ |----------- |------------- |
           Slice | 269.2951 ns |  1.3123 ns | 3,713,482.96 |
 SliceWithLength | 273.1879 ns | 17.6509 ns | 3,671,898.16 |

**The end result (after the fix):**

          Method |     AvrTime |    StdDev |         op/s |
---------------- |------------ |---------- |------------- |
           Slice | 192.6039 ns | 2.5312 ns | 5,192,833.50 |
 SliceWithLength | 357.4231 ns | 1.4362 ns | 2,797,847.44 |

### DIFF
--- a/src/System.Slices/src/System/Contract.cs
+++ b/src/System.Slices/src/System/Contract.cs
@@ -41,6 +41,18 @@ namespace System
             }
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static void RequiresInInclusiveRange(int start, int length, int existingLength)
+        {
+            if ((uint)start > (uint)existingLength
+                || length < 0
+                || (uint)(start + length) > (uint)existingLength)
+            {
+                throw NewArgumentOutOfRangeException();
+            }
+        }
+
+        
         [MethodImpl(MethodImplOptions.NoInlining)]
         private static Exception NewArgumentException()
         {
--- a/src/System.Slices/src/System/Contract.cs
+++ b/src/System.Slices/src/System/Contract.cs
@@ -24,29 +24,29 @@ namespace System
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void RequiresInRange(int start, int length)
+        public static void RequiresInRange(int start, uint length)
         {
-            if ((uint)start >= (uint)length)
+            if ((uint)start >= length)
             {
                 throw NewArgumentOutOfRangeException();
             }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void RequiresInInclusiveRange(int start, int length)
+        public static void RequiresInInclusiveRange(int start, uint length)
         {
-            if ((uint)start > (uint)length)
+            if ((uint)start > length)
             {
                 throw NewArgumentOutOfRangeException();
             }
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static void RequiresInInclusiveRange(int start, int length, int existingLength)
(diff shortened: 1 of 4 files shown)


src/System.Slices/src/System/Contract.cs
```diff
@@ -41,6 +41,18 @@ public static void RequiresInInclusiveRange(int start, int length)
...
+        {
+            if ((uint)start > (uint)existingLength
+                || length < 0
+                || (uint)(start + length) > (uint)existingLength)
+            {
+                throw NewArgumentOutOfRangeException();
+            }
+        }
+
+
```
KrzysztofCwalina: RequiresInInclusiveRange(0, 1, int.MinValue) passes the check. Is that what we want?

omariom: The last param  expected to be Span.Length which can't be negative.

omariom: I will add tests for the methods.

KrzysztofCwalina: In your case, it's length, but I worry that people will reuse the very convenient method (especially once we make the contract class public, which I want to do), and they will use it with something that is not guaranteed to be positive.

omariom: The first override suffers from the same issue. 
I don't know what to do. Additional comparisons will slow things down a bit.

KrzysztofCwalina: Maybe this API should take uint for length? This way the cast will be in the consumer side and so  the problem will be visible to the consumer.

omariom: Can you give an example?

KrzysztofCwalina: @omariom, could you clarify what example you'd like to see?

omariom: @KrzysztofCwalina 
Now I see what you mean. The last last param should be uint.
ok

KrzysztofCwalina: yes.

krwq:
LGTM :+1:

PR #545, opened 2016-01-11
Formatting.Globalization: memory: less allocations + stream Dispose

adamsitnik:
1. Wrapped stream in using statement just to make sure the resources are released properly
2. Creation of new Utf8String without usage of CreateArray method, which was unnecessary allocating memory

@KrzysztofCwalina I believe you are the right person to approve this PR ;)


KrzysztofCwalina:
looks good. thanks.

PR #568, opened 2016-01-15
Json dom APIs

ahsonkhan:
Please note that there were some changes in the last 2 weeks that might get overwritten with some of the changes here or cause build failures in System.Text.Json (I believe they were made to allow for build with dotnet cli). 

Those changes include:
- Removal of resources files (which I added back, can be removed if it is required to build).
- Removal of csproj and addition of xproj files ("All projects are relying on nuget packages for all dependencies but test to source.").
- Changes to project.json and project.json.lock files along with System.Text.Json.sln file. I did not make any changes to these files to respect new changes.


KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@ahsonkhan, could you please sign the CLA?

ahsonkhan:
@KrzysztofCwalina, how do I do that? Are you talking about: "Pre-Merge signing tool — Sign-off"? When I try to login to codesignoff, it states: 
"Sorry, but we’re having trouble signing you in.
We received a bad request."

joshfree:
@ahsonkhan you can submit the CLA @ https://cla2.dotnetfoundation.org

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
...
+                        GetValue();
+                        numValues++;
+                        numPairs++;
+                        break;
+                    default:
+                        throw new ArgumentOutOfRangeException();
+                }
+            }
+
+            /*for(int i = _end+1; i < _dbIndex; i+=RowSize)
```
KrzysztofCwalina: Should these be removed?

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonNetComparison.cs
```diff
@@ -5,15 +5,17 @@
 using Newtonsoft.Json;
```
KrzysztofCwalina: Why is this file in "experiments" subfolder?

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
+﻿using System;
+
+namespace Json.Net.Tests
+{
+    public struct JsonParser : IDisposable
```
KrzysztofCwalina: Remind me why you have this type here? i.e. why do you define the parser in tests project?

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+{
+    public struct JsonParseObject
+    {
+        private byte[] _buffer;
+        private int _start;
+        private int _end;
+
+        private const int RowSize = 9;
+
+        public JsonParseObject(byte[] buffer, int start, int end)
```
KrzysztofCwalina: I think the first parameter has to be UTF8 encoded text? If that's correct, we should rename the parameter to utf8JsonText or something like that.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+            _end = end;
+        }
+
+        public bool HasValue()
+        {
+            var typeCode = _buffer[_start + 8];
+
+            if (typeCode == (byte)JsonParser.JsonValueType.Object || typeCode == (byte)JsonParser.JsonValueType.Array)
+            {
+                int length = BitConverter.ToInt32(_buffer, _start + 4);
```
KrzysztofCwalina: Should we worry about the buffer being too small to read _start + 4 + 4 bytes?

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+    {
+        private byte[] _buffer;
+        private int _start;
+        private int _end;
+
+        private const int RowSize = 9;
+
+        public JsonParseObject(byte[] buffer, int start, int end)
+        {
+            _buffer = buffer;
```
KrzysztofCwalina: The ctor should do some validation that start and end are consistent, i.e. start < end. 

Also, in most framework APis we use index and count, not start and end. Why don't we do the same here?

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+                int location = BitConverter.ToInt32(_buffer, _start);
+                if (_buffer[location - 1] == '"' && _buffer[location + 4] == '"')
+                {
+                    return true;
+                }
+                return (_buffer[location] != 'n' || _buffer[location + 1] != 'u' || _buffer[location + 2] != 'l' || _buffer[location + 3] != 'l');
+            }
+        }
+
+        public JsonParseObject this[string index]
```
KrzysztofCwalina: Is "index" the right name for the parameter? On a related note, it's strange that KeyNotFound is thrown even though there is no key in the picture.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+                var typeCode = _buffer[_start + 8];
+
+                if (length == 0)
+                {
+                    throw new KeyNotFoundException();
+                }
+
+                if (typeCode != (byte)JsonParser.JsonValueType.Object)
+                {
+                    throw new NullReferenceException();
```
KrzysztofCwalina: Framework code should not throw this exception. Only runtimes throw it.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+                        i += RowSize;
+                    }
+                }
+
+                throw new KeyNotFoundException();
+
+            }
+        }
+
+        private bool isEqual(string str, byte[] buffer, int location, int length)
```
KrzysztofCwalina: isEqual does not follow our naming/casing conventions. I would call it simply Equals. 
I think it would also be useful to make it clearer that the location and length refer to the array. Possibly by describing it in a comment. Lastly, we call such parameters "index" and "count", not "location" and "length"

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                if (str[i] != buffer[location + i])
+                {
+                    return false;
+                }
+            }
+            return true;
+        }
+
+        public JsonParseObject this[int index]
```
KrzysztofCwalina: I assume this is applicable to arrays only? If yes, should we have a separate type for arrays? e.g. JsonParsedArray?

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+            }
+
+            int length = BitConverter.ToInt32(json._buffer, json._start + 4);
+            if (length < 4 || length > 5)
+            {
+                throw new InvalidCastException();
+            }
+
+            int location = BitConverter.ToInt32(json._buffer, json._start);
+            bool isTrue = json._buffer[location] == 't' && json._buffer[location + 1] == 'r' && json._buffer[location + 2] == 'u' && json._buffer[location + 3] == 'e';
```
KrzysztofCwalina: if isTrue == true, you should return immediatelly instead of checking if it's "false".

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+            {
+                return true;
+            }
+            else if (isFalse)
+            {
+                return false;
+            }
+            else
+            {
+                throw new InvalidCastException();
```
KrzysztofCwalina: Boolean value can be null too, correct? If yes, then this (and other similar casts) need to be able to handle/return null values.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+            }
+        }
+
+        public static explicit operator int (JsonParseObject json)
```
KrzysztofCwalina: This should really be in System.Text.Formatting package in the parser namespace (see InvariantParser class). BTW, I am ok with you punting these large changes for later, as long as you file issues for them.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
```diff
@@ -0,0 +1,350 @@
...
+                }
+                int digit = nextByte - '0';
+                result = result * 10 + digit;
+                count++;
+            }
+
+            return isNegative ? result * -1 : result;
+        }
+
+        public static explicit operator double (JsonParseObject json)
```
KrzysztofCwalina: Same with this one.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Text.Json
+{
+    public struct JsonParser : IDisposable
```
KrzysztofCwalina: Could you add comments explaining what each of the types is for. I now forgot, is this the reader that keeps track of buffer locations?

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
...
+                _index++;
+                nextByte = _buffer[_index];
+            }
+
+            _dbIndex = _end + 1;
+
+            _jsonStartIsObject = _buffer[_index] == '{';
+        }
+
+        public void Dispose()
```
KrzysztofCwalina: Why is it disposable if Dispose is empty?

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
...
+        }
+
+        public JsonParseObject Parse()
+        {
+            int numValues = 0;
+            int numPairs = 0;
+            int numObj = 0;
+            int numArr = 0;
+
+            int topOfStackObj = _buffer.Length - 1;
```
KrzysztofCwalina: I think we need some good comments here describing the architecture (the stack, the database, etc)

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
...
+            _dbIndex = _end + 1;
+
+            _jsonStartIsObject = _buffer[_index] == '{';
+        }
+
+        public void Dispose()
+        {
+        }
+
+        public JsonParseObject Parse()
```
KrzysztofCwalina: I wonder it the ctor should not be private and the Parse method would simply take the buffer as a parameter.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
...
+                        GetValue();
+                        numValues++;
+                        numPairs++;
+                        break;
+                    default:
+                        throw new ArgumentOutOfRangeException();
+                }
+            }
+
+            /*for(int i = _end+1; i < _dbIndex; i+=RowSize)
```
KrzysztofCwalina: Should the commented code be removed?

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
```diff
@@ -0,0 +1,565 @@
...
+            var nextByte = _buffer[_index];
+
+            while (isWhiteSpace(nextByte))
+            {
+                _index++;
+                nextByte = _buffer[_index];
+            }
+        }
+
+        private static bool isWhiteSpace(byte nextByte)
```
KrzysztofCwalina: casing

src/System.Text.Json/tests/JsonReaderTests.cs
```diff
@@ -1,5 +1,6 @@
 ﻿using System.Collections.Generic;
```
KrzysztofCwalina: I think we need more tests, especially for invalid JSON, buffers too short, etc. Again, feel free to punt some or all of this and just file a TODO issues. I don't want to take your time away from school :-) and I want this commit to go in.

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonParser.cs
ahsonkhan: Yes, the commented out block can be removed. It was only used for debugging..

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonNetComparison.cs
ahsonkhan: That is what was initially decided, but it can be moved to the demos folder.

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonParser.cs
ahsonkhan: It was there to help with performance analysis when running the profiler so you get more details. It is essentially a copy of what is found in src.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: Would that also apply to JsonParser?

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: We are already accessing buffer at index _start + 8 to grab the typecode. If a 'row' exists, it will definitely contain ROWSIZE bytes (9) or else the DB wasn't created properly.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: The validation can be added. We can change it from start/end to index and count.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: Agreed, this should be "key".

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: OK. So in the scenario when user searches for a key string within a Json Array, what should we do? Only integer indexing makes sense in that case.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: ok

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: JsonParseObject is not tied to Json Objects or Json Arrays. It is just a misnamed type indicating that it is the object that contains the db created after the json has been parsed. Can you suggest a better name? Hence, it is still applicable to Json Arrays.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: Yes, agreed. Although, then I should also check that length is actually 4. Currently, if json buffer had truex, it would still pass isTrue == true (since the first 4 characters match).

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: What would "null" casted to boolean return here?

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: Are you suggesting I extract some method that takes buffer and location and returns an int and add that to System.Text.Formatting?

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: OK.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
ahsonkhan: JsonParser actually reads/parses through the json string and builds the DB in the buffer. Is this the type of information you were looking for? I will add the issues, sure.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
ahsonkhan: I followed the template from JsonWriter which had IDisposable. This can be removed.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
ahsonkhan: OK. I had a lot of the explanation/comments within the slides.  I can add them as inline comments.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
ahsonkhan: ctor? It isn't private. Are you suggesting we make this private and have the user pass in the buffer? JsonParser ctor is public and so is Parse().

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
ahsonkhan: Yes. It was used for debugging.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
ahsonkhan: Got it. Thanks.

src/System.Text.Json/tests/JsonReaderTests.cs
ahsonkhan: There was limited focus on invalid JSON because I was focused on correctness and under the assumption that the JSON is valid. I think we need to specify how to handle some of the error cases that you mentioned and thoroughly tests those scenarios. Regarding the case when the buffer is too short, do we allocate a larger buffer and copy over the data?

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonParser.cs
KrzysztofCwalina: How are we going to keep the sources in sync? I would try to find a way to remove this file.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
KrzysztofCwalina: I think so.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
KrzysztofCwalina: I am not sure I understand why we cannot always return JsonParsedObject, but if we cannot, then throw InvalidOperationException.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
KrzysztofCwalina: But it seems to me like not all operations on JsonParsedObject work, e.g. the integer based indexer only works for arrays. And so I thought maybe you need two types: JsonParsedObject and JsonParsedArray.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
KrzysztofCwalina: I think this case would have to be to Nullable<bool>. Or maybe we add two casts: to bool and to Nullable<bool>.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
KrzysztofCwalina: Yes.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
KrzysztofCwalina: Yes.

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
KrzysztofCwalina: yeah, make the ctor private and the usage be:
JsonParsedObject parsed = JsonParser.Parse(buffer, ...);

src/System.Text.Json/tests/JsonReaderTests.cs
KrzysztofCwalina: For now let's just throw. But still there should be a test validating that it throws.

src/System.Text.Json/experiments/Json.Net.Test/Json.Net.Tests/JsonParser.cs
ahsonkhan: It essentially can be removed. I only included it into the project whenever I wanted to deep dive into perf. They are not used in JsonNetComparison. If the files are excluded from the project, then the DLL ends up being used (as expected).

using JsonReader = System.Text.Json.JsonReader;
using JsonWriter = System.Text.Json.JsonWriter;
using JsonParser = System.Text.Json.JsonParser;

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
ahsonkhan: I don't understand. I think the name JsonParseObject is causing some confusion. Integer indexer  only makes sense for arrays just like string indexer makes sense for objects. This class is only there to encapsulate the DB information for appropriate index and casts to use. The user should know whether the JsonParseObject is pointing to an array or an object and index appropriately. If they don't, we throw. How does having separate types resolve that issue? The user can still try to index a JsonParsedArray object with a string, no?

src/System.Text.Json/src/System/Text/Json/JsonParser.cs
ahsonkhan: JsonParser.Parse() relies on some of the JsonParser members. Some of those members have values set after processing the buffer passed into the JsonParser ctor such as _end and _dbIndex.

src/System.Text.Json/src/System/Text/Json/JsonParseObject.cs
KrzysztofCwalina: JsonParsedArray would not have an indexer that takes a string. It would only have an indexer that takes an int.

Having said that, I think we should merge this PR as-is and then possibly change the design later. So, treat this comment as simply something to think about.

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

PR #569, opened 2016-01-15
Fixed build breaks in formatting

KrzysztofCwalina:
Trying to fix build breaks after the recent move to the new build system and nuget sources.


KrzysztofCwalina:
I am going to just shove this in. The JSON project is still failing, but this change is a prerequisite to fix the JSON project.

PR #574, opened 2016-01-17
Slices: Faster SequenceEqual<T> implemented in IL

omariom:
It is similar to IndexOf impl. The same benefits - shorter code, tighter loop.

Assembly for the long and char instantiations.
```ASM
; Assembly listing for method System.SpanExtensions:SequenceEqual(struct,struct):bool
; Lcl frame size = 0
       3B4210               cmp      eax, dword ptr [rdx+16]
       7539                 jne      SHORT G_M39316_IG06
       4863C0               movsxd   eax, eax
       4885C0               test     rax, rax
       7428                 je       SHORT G_M39316_IG04
       4C8B01               mov      r8, gword ptr [rcx]
       488B4908             mov      rcx, qword ptr [rcx+8]
       4C03C1               add      r8, rcx
       488B0A               mov      rcx, gword ptr [rdx]
       488B5208             mov      rdx, qword ptr [rdx+8]
       4803CA               add      rcx, rdx
       33D2                 xor      rdx, rdx
G_M39316_IG03:
       4C8B0CD1             mov      r9, qword ptr [rcx+8*rdx]
       4D390CD0             cmp      qword ptr [r8+8*rdx], r9
       750E                 jne      SHORT G_M39316_IG06
       48FFC2               inc      rdx
       483BD0               cmp      rdx, rax
       7CEE                 jl       SHORT G_M39316_IG03
G_M39316_IG04:
       B801000000           mov      eax, 1
       C3                   ret      
G_M39316_IG06:
       33C0                 xor      eax, eax
       C3                   ret      

; Total bytes of code 68, prolog size 0 for method System.SpanExtensions:SequenceEqual(struct,struct):bool
; ============================================================

; Assembly listing for method System.SpanExtensions:SequenceEqual(struct,struct):bool
; Lcl frame size = 0
       3B4210               cmp      eax, dword ptr [rdx+16]
       753E                 jne      SHORT G_M17275_IG06
       4863C0               movsxd   eax, eax
       4885C0               test     rax, rax
       742D                 je       SHORT G_M17275_IG04
       4C8B01               mov      r8, gword ptr [rcx]
       488B4908             mov      rcx, qword ptr [rcx+8]
       4C03C1               add      r8, rcx
       488B0A               mov      rcx, gword ptr [rdx]
       488B5208             mov      rdx, qword ptr [rdx+8]
       4803CA               add      rcx, rdx
       33D2                 xor      rdx, rdx
G_M17275_IG03:
       440FB70C51           movzx    r9, word  ptr [rcx+2*rdx]
       450FB71450           movzx    r10, word  ptr [r8+2*rdx]
       453BD1               cmp      r10d, r9d
       750E                 jne      SHORT G_M17275_IG06
       48FFC2               inc      rdx
       483BD0               cmp      rdx, rax
       7CE9                 jl       SHORT G_M17275_IG03
G_M17275_IG04:
       B801000000           mov      eax, 1
       C3                   ret      
G_M17275_IG06:
       33C0                 xor      eax, eax
       C3                   ret      

; Total bytes of code 73, prolog size 0 for method System.SpanExtensions:SequenceEqual(struct,struct):bool
```

The results of the [benchmark](https://gist.github.com/omariom/5ac39f97747f96cfe941) for lengths of 1, 2, 4, 8, 16, 32, 64, 128, 512, 1024 and types **byte**, **char** and **int**. For comparison I added string.Equals and looping through two char arrays.
Meaning of the labels for chars:
**char** - calls to SequenceEquals specialized for Span[char]
**char array** - loop through char arrays
**char gen** - current generic SequenceEquals[T]
**char gen 2** - current SequenceEquals[T] but with Span params copied to local vars (JIT enregisters them)
**char gen new** - the new SequenceEquals[T] implemented in IL

 I ran the benchmark on my ultrabook  - Windows 10 and NET 4.6 (BenchmarkDotNet doesn't yet support .NET Core).

BenchmarkDotNet-Dev=v0.8.1.0+
Type=BenchmarkSeqEqual  Mode=Throughput  Jit=HostJit  .NET=HostFramework  toolchain=Classic  Runtime=Clr  Warmup=2  Target=3 

seqequals

### DIFF
--- a/src/System.Slices/System/MemoryUtils.cs
+++ b/src/System.Slices/System/MemoryUtils.cs
@@ -14,6 +14,11 @@ namespace System
         internal static bool MemCmp<[Primitive]T>(Span<T> first, Span<T> second)
             where T : struct
         {
+            if (first.Length != second.Length)
+            {
+                return false;
+            }
+
             unsafe
             {
                 // prevent GC from moving memory
@@ -79,6 +84,11 @@ namespace System
         internal static bool MemCmp<[Primitive]T>(ReadOnlySpan<T> first, ReadOnlySpan<T> second)
             where T : struct
         {
+            if (first.Length != second.Length)
+            {
+                return false;
+            }
+
             unsafe
             {
                 // prevent GC from moving memory
--- a/src/System.Slices/System/SpanEqualsExtensions.cs
+++ b/src/System.Slices/System/SpanEqualsExtensions.cs
@@ -15,39 +15,21 @@ namespace System
         /// <param name="second">A span of bytes T to compare to first.</param>
         public static bool SequenceEqual(this Span<byte> first, Span<byte> second)
         {
-            if (first.Length != second.Length) { return false; }
-
-            // pinvoke calls are expensive (mostly due to memory pinning) so we call memcmp only for big slices
-            if (first.Length >= 512) { return MemoryUtils.MemCmp(first, second); }
-
-            // for small slices we just use simple loop
-            for (int i = 0; i < first.Length; i++) {
-                if (first.GetItemWithoutBoundariesCheck(i) != second.GetItemWithoutBoundariesCheck(i)) { 
-                    return false; 
-                }
-            }
-            return true;
+            return first.Length >= 320
+                ? MemoryUtils.MemCmp(first, second)
+                : SequenceEqual<byte>(first, second);
         }
 
-		/// <summary>
+        /// <summary>
         /// Determines whether two read-only spans are equal (byte-wise) by comparing the elements by using memcmp
         /// </summary>
         /// <param name="first">A span of bytes to compare to second.</param>
         /// <param name="second">A span of bytes T to compare to first.</param>
         public static bool SequenceEqual(this ReadOnlySpan<byte> first, ReadOnlySpan<byte> second)
         {
-            if (first.Length != second.Length) { return false; }
-
-            // pinvoke calls are expensive (mostly due to memory pinning) so we call memcmp only for big slices
-            if (first.Length >= 512) { return MemoryUtils.MemCmp(first, second); }
-
-            // for small slices we just use simple loop
-            for (int i = 0; i < first.Length; i++) {
-                if (first.GetItemWithoutBoundariesCheck(i) != second.GetItemWithoutBoundariesCheck(i)) { 
-                    return false; 
-                }
-            }
-            return true;
+            return first.Length >= 320
+                ? MemoryUtils.MemCmp(first, second)
+                : SequenceEqual<byte>(first, second);
         }
 
         /// <summary>
@@ -57,39 +39,21 @@ namespace System
         /// <param name="second">A span of characters T to compare to first.</param>
         public static bool SequenceEqual(this Span<char> first, Span<char> second)
         {
-            if (first.Length != second.Length) { return false; }
-
-            // pinvoke calls are expensive (mostly due to memory pinning) so we call memcmp only for big slices
-            if (first.Length >= 512) { return MemoryUtils.MemCmp(first, second); }
-
-            // for small slices we just use simple loop
-            for (int i = 0; i < first.Length; i++) {
-                if (first.GetItemWithoutBoundariesCheck(i) != second.GetItemWithoutBoundariesCheck(i)) { 
-                    return false; 
-                }
-            }
-            return true;
+            return first.Length >= 512
+                ? MemoryUtils.MemCmp(first, second)
+                : SequenceEqual<char>(first, second);
         }
 
-		/// <summary>
+        /// <summary>
         /// Determines whether two read-only spans are equal (byte-wise) by comparing the elements by using memcmp
         /// </summary>
         /// <param name="first">A span of characters to compare to second.</param>
         /// <param name="second">A span of characters T to compare to first.</param>
         public static bool SequenceEqual(this ReadOnlySpan<char> first, ReadOnlySpan<char> second)
         {
-            if (first.Length != second.Length) { return false; }
-
-            // pinvoke calls are expensive (mostly due to memory pinning) so we call memcmp only for big slices
-            if (first.Length >= 512) { return MemoryUtils.MemCmp(first, second); }
-
-            // for small slices we just use simple loop
-            for (int i = 0; i < first.Length; i++) {
-                if (first.GetItemWithoutBoundariesCheck(i) != second.GetItemWithoutBoundariesCheck(i)) { 
-                    return false; 
-                }
-            }
-            return true;
+            return first.Length >= 512
+                ? MemoryUtils.MemCmp(first, second)
+                : SequenceEqual<char>(first, second);
         }
 
         /// <summary>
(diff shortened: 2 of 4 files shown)


omariom:
@mikedn @VSadov  do you see anything suspicious?

omariom:
string.Equals beats them all, but that's because [it unrolls the loop and reads in longs](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/String.cs#L362).

mikedn:
Code looks fine but I wonder where the whole IL thing goes. Initially IL was only when strictly needed but now it looks like half of this library may end up being rewritten in IL. It's perhaps unfortunate given that this is just an workaround for C# language and JIT compiler limitations.

omariom:
@mikedn 
I am concerned too. But for while, until proper JIT support, IL is Ok. 
People will be comparing Slices' perf with arrays and strings and a bad name can stay longer than it  will deserve to. 
And I think better shape increases its chance to get into CoreFx even without favors from JIT.

omariom:
>  I wonder where the whole IL thing goes
 
To make things worse I'm thinking about implementing a non-pinning version of [MemoryUtils.MemCmp](https://github.com/dotnet/corefxlab/blob/master/src/System.Slices/src/System/MemoryUtils.cs#L14) unrolling the loop into 4 non-aligned long reads :)

Yeah, baby Slice, we will beat the shit out of string.Equals and teach it to respect small things that can't yet enjoy language and JIT support :rage3:

If it is too much of IL.. well, it will have been a good exercise for me )

omariom:
I will try to replace a loop in the  specialized overloads (byte, char, int etc.) of   SequenceEqual with a call to  the new generic one and see if it will make them faster.

VSadov:
Is the benefit mostly from omitting the range check?
I wonder it it makes sense to implement overloads for primitives like  byte,char,int etc by just deferring to optimized byte-wise compare (do we have such thing already? like BlockCompare?) .

For general T, it may be a diminishing return thing to create and maintain custom IL implementation. Those scenarios may be not as common and are likely be dominated by things other than a range check.

omariom:
> Is the benefit mostly from omitting the range check?
No,  all the overloads already use  GetItemWithoutBoundariesCheck. I think it is number of instructions.

> I wonder it it makes sense to implement overloads for primitives like byte,char,int etc by just deferring to optimized byte-wise compare (do we have such thing already? like BlockCompare?) .

Those overloads already do that if the length >= 512 (256 for `int` and `long`). [It is pinning](https://github.com/dotnet/corefxlab/blob/86ea15d941e0599ef5c992f228c4cced318fee3e/src/System.Slices/src/System/MemoryUtils.cs#L14) and on shorter distances slower than the loop.

Some improvements can be gained by copying span parameters to locals vars, using the version of the indexer  fixed in another PR  and passing params  to MemCmp by ref.

I will try to squeeze everything from IL and non IL approaches and benchmark both.

Speaking of the burden of maintaining IL, I started thinking that Slices is a fairly low level thing,  and as there are parts of CLR  for perf sake written and maintained in Assembly, it might be ok for Slices to be implemented in  IL for the same reasons.

mikedn:
> No, all the overloads already use GetItemWithoutBoundariesCheck. I think it is the number of instructions.

For reference here's the loop assembly code generated for the current implementation (slightly adjusted for clarity):

```
00007FFCF02F0E79  mov         rax,qword ptr [rcx]  
00007FFCF02F0E7C  mov         r9,qword ptr [rcx+8]  
00007FFCF02F0E80  add         rax,r9  
00007FFCF02F0E83  mov         r9d,r8d  
00007FFCF02F0E86  mov         rax,qword ptr [rax+r9*8]  
00007FFCF02F0E8A  mov         r9,qword ptr [rdx]  
00007FFCF02F0E8D  mov         r10,qword ptr [rdx+8]  
00007FFCF02F0E91  add         r9,r10  
00007FFCF02F0E94  mov         r10d,r8d  
00007FFCF02F0E97  cmp         qword ptr [r9+r10*8],rax  
00007FFCF02F0E9B  jne          00007FFCF02F0EA0  
00007FFCF02F0EA0  inc         r8d  
00007FFCF02F0EA3  cmp         r8d,dword ptr [rcx+10h]  
00007FFCF02F0EA7  jl          00007FFCF02F0E79
```

There are portions that are loop invariant but the JIT isn't that smart to hoist them so the resulting code is quite large.

mikedn:
The following generates slightly better assembly code:

```C#
var fobj = first._object;
var fofs = first._offset;
var sobj = second._object;
var sofs = second._offset;
for (int i = 0; i < first.Length; i++) {
    if (PtrUtils.Get<long>(fobj, fofs, i) != PtrUtils.Get<long>(sobj, sofs, i)) { 
        return false; 
    }
}
```
The introduced local variables are enregistered so a lot of memory loads are saved. But the object + offset addition from `PtrUtils.Get` is still not hoisted. I think that's actually expected since a managed pointer is involved and that requires GC tracking.

omariom:
@mikedn 
The generated code is the same if we just copy the params to locals and use them instead. 

I don't get why JIT can't enregister struct params which are passed by value.

mikedn:
Are you saying that my version and a version that copies params to locals generate the same code? Could be, I haven't checked. But I suppose that the object + offset problem is still present, right?

omariom:
> Are you saying that my version and a version that copies params to locals generate the same code? 

Yes, JIT it handles it  well.

> But I suppose that the object + offset problem is still present, right?

Yes, it is still there. And the size of the method is about 140,  the frame same is about 40. The IL version is about 70 and zero frame.

Anyway, I will battle test both. Thought I think the difference will be bigger on weaker CPUs  like Atoms.

mikedn:
> Yes, JIT it handles it well.

Makes sense. Value types are actually passed by ref and that prevents various optimizations. But when value types stored in local variables the JIT can sometimes replace them with individual scalar variables and then optimization start working.

omariom:
> Value types are actually passed by ref and that prevents various optimizations.

At the CLR level it is guaranteed that the caller passes refs to copies which are invisible to anyone except the callee, isn't  it? Then why JIT can't enregister them? Or I miss something?

mikedn:
> At the CLR level it is guaranteed that the caller passes refs to copies which are invisible to anyone except the callee, isn't it?

Well, yes, in the current implementation a `ref` to a temporary copy is passed. That's required to preserve the by value semantics of the code.

> Then why JIT can't enregister them?

Eh, the usual JIT limitations, warts and all. It has a history of problems with value types. Funny thing that if span was a reference type then this would have likely worked.

omariom:
@mikedn 
 I will create an issue for that.

Interesting that [readonly locals and params](https://github.com/dotnet/roslyn/issues/115) will let the caller to avoid making copies which are currently created  [even when the calls are inlined](https://github.com/dotnet/coreclr/issues/1133).

mikedn:
There are better ways to avoid making copies that do not require language support. It's really up to the JIT to make this scenario better, it's not the job of the language to work around JIT limitations.

mburbea:
Can we use SIMD registers for this? With something like casting a pointer to a `Vector<long>` (type doesn't matter), you could then compare 16 / 32 bytes at a time. For the remaining amount of buffer you can fall back to the current test.

omariom:
@mburbea  
> Can we use SIMD registers for this?

I know little about Vector and how to use it :confused:

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

omariom:
@KrzysztofCwalina 
This PR is not done yet. I am going to route non generic `SequenceEquals'` to IL version so that they benefit from the change too.

omariom:
This is the [microbenchmarking results](https://onedrive.live.com/redir?resid=4A71F4F45CC000CE!327329&authkey=!AIg0z8rHMxzcw5M&ithint=file%2cxlsx)  for slices of bytes.

seqeq

omariom:
Now it is ready.

KrzysztofCwalina:
@dotnet-bot test this please

PR #588, opened 2016-01-29
Utf8: Some optimizations of usage of Spans

omariom:
@KrzysztofCwalina

### DIFF
--- a/src/System.Text.Utf8/src/System/Text/Utf16/Utf16LittleEndianEncoder.cs
+++ b/src/System.Text.Utf8/src/System/Text/Utf16/Utf16LittleEndianEncoder.cs
@@ -17,7 +17,7 @@ namespace System.Text.Utf16
                 return false;
             }
 
-            uint codePointValue = (uint)buffer[0] | ((uint)buffer[1] << 8);
+            uint codePointValue = buffer.Read<ushort>();
             encodedBytes = 2;
             if (UnicodeCodePoint.IsSurrogate((UnicodeCodePoint)codePointValue))
             {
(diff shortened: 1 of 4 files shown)


KrzysztofCwalina:
Our ultimate plan is to teach the GC about interior pointers from Span<T> and stop using the Object field in the span indexer. Do you believe this change will still be net positive when the GC work happens?

omariom:
> Our ultimate plan is to teach the GC about interior pointers from Span and stop using the Object field in the span indexer. 

You mean an interior pointer will replace  Object and Offset fields? 

>  Do you believe this change will still be net positive when the GC work happens? 

The change is mainly trying to save on the range check and the calculation each indexer access requires. So yes, I think   the change can shave off a few cycles. At least, until JIT support comes.

KrzysztofCwalina:
I hope we will end up with Span<T> with 2 fields: UIntPtr and length:int. The GC will update the pointer when it moves the array that the pointer points to (if it points to an array not to native memory). 

I also hope we will teach the optimizers to do range check elimination on spans. 

In general, the goal is for span to be equivalent to array in terms of the indexer performance. 

Having said that, we will probably not do this change in previous versions of our runtimes. And also, we might not have all of these optimizations initially, even in the new versions of the runtimes. And so, I am totally fine with this commit. I just wanted to decide if we should file a tracking issue to revert some of these when the optimizations are in place.

omariom:
> I hope we will end up with Span with 2 fields: UIntPtr and length:int.

That' great news :+1:

KrzysztofCwalina:
@dotnet-bot test this please

PR #636, opened 2016-03-10
Update exit from help output and app name

russellhadley:
After printing the help screen System.CommandLine was calling fastfail which tears down the process with an exception.  This caused a popup and looked like an exceptional exit.  Switched the code to use Environment.Exit(), which is now available.  At the same time I also include the code to call Environment.GetCommandLineArgs() to allow the help code to find the app name.

### DIFF
--- a/src/System.CommandLine/System/CommandLine/ArgumentSyntax.cs
+++ b/src/System.CommandLine/System/CommandLine/ArgumentSyntax.cs
@@ -51,13 +51,7 @@ namespace System.CommandLine
             {
                 var helpText = GetHelpText();
                 Console.Error.Write(helpText);
-
-                // TODO: This should use Environment.Exit(0) but this API isn't available yet.
-#if NET_FX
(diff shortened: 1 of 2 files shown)


russellhadley:
@KrzysztofCwalina @terrajobst PTAL

KrzysztofCwalina:
Any tests need to be updated to test the new behavior?

russellhadley:
@KrzysztofCwalina there are already tests that check this functionality I just removed the #ifdef so that the functionality would be available for dnxcore.  We should be able to run these tests other ways  (actually I'd like to see a build I can run on OSX or Linux for instance) but changing the automation is a bit beyond me.  Who is doing the CI for this stuff?

KrzysztofCwalina:
cc: @terrajobst, @piotrpMSFT 

Looks good to me.

terrajobst:
Thanks for fixing this!

PR #638, opened 2016-03-14
Added Span<T>.TryCopyTo(T[])

KrzysztofCwalina:


src/System.Slices/System/Span.cs
```diff
@@ -195,6 +195,26 @@ public bool TryCopyTo(Span<T> dest)
             return true;
         }
 
+        /// <summary>
+        /// Copies the contents of this span into an array.  The destination
+        /// must be at least as big as the source, and may be bigger.
+        /// </summary>
```
VSadov: Do we promise any deterministic behavior if the span happens to be a sub-range over the destination array?

KrzysztofCwalina: Yeah, good issue. We should make a decision here and either error or work properly. I filed #639.

PR #643, opened 2016-03-16
Added Multispan<T>

KrzysztofCwalina:
Very useful when dealing with I/O traffic coming in batches.

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/Multispan.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/Multispan.cs
@@ -1,0 +1,216 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Buffers
+{
+    /// <summary>
+    /// A collection of slices backed by pooled arrays.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    /// <remarks>
+    /// Slices should not be stored on the heap, and soon we will make it impossible to do.
+    /// Yet if asynchronous operations must store their context on the heap. 
+    /// This type creates and illusion of a collection of slices that can be stored on the heap.
+    /// Note that this type is a mutable value type. You need to be careful when using it. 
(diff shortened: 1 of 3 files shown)


KrzysztofCwalina:
@sokket

src/System.Buffers.Experimental/System/Buffers/Multispan.cs
```diff
@@ -0,0 +1,213 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Buffers
+{
+    /// <summary>
+    /// A collection of pooled arrays.
```
sokket: This is confusing...what scenario does this allow that the regular pool does not? It looks like (and this comment specifies) that this is just a collection of different size, pooled ```T```s.

omariom:
How is it intended  to be used?

KrzysztofCwalina:
@omariom, see the doc comment at the type level.
It might be not the greatest doc, but I think it will point you in the right direction. 

The main usage is in async i/o. Calling code allocates new segment before each i/o call (ReceiveAsync, for example); the i/o operation fills in the next segment with data that was received in current chunk, and then you repeat till i/o is completed. 

The same cannot be achieved with a collection of spans, because spans should not be on the heap. It also cannot be achieved with a collection of arrays, because user code could store the pooled arrays for later (and what's worse return them to the pool multiple times). 

Soon I should be able to check in sample showing how it all comes together in an iocp service.

Maybe Multispan is not the greatest name, but we had a type like that in an incubation project I worked on, and so I reused it.

omariom:
What if one wants to use unmanaged chunks of memory for IO? 
 ArrayPool allocates from managed heap. Can it be abstracted?

KrzysztofCwalina:
Yeah, we should think about ArrayPool like abstraction that gives off Span<T>s instead of T[]s. Once we have such abstraction, we should add Multispan ctor that takes such abstract pool and uses it instead of using ArrayPool<T>.Shared.

PR #650, opened 2016-04-11
Added Bytes property to Utf8String

KrzysztofCwalina:
Now that we have ReadOnlySpan<T>, there is no reason to expose the bytes,
and it allows to avoid copies in many scenarios.


KrzysztofCwalina:
@dotnet-bot test this please

PR #652, opened 2016-04-13
Make System.CommandLine not depend on generic virtual methods

MichalStrehovsky:
Generic virtual methods are pretty expensive for native compilation
scenarios (such as .NET Native). System.CommandLine should be a low
level component and not depend on expensive features so that it can be
used for small command line apps.

Generic virtual methods were used indirectly through LINQ Where and
Select operators.

### DIFF
--- a/src/System.CommandLine/System/CommandLine/Enumerable.cs
+++ b/src/System.CommandLine/System/CommandLine/Enumerable.cs
@@ -1,0 +1,86 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.CommandLine
+{
+    // Low level replacement of LINQ Enumerable class. In particular, this implementation
+    // doesn't use generic virtual methods.
+    // System.CommandLine needs to be usable for small .NET Native apps that don't carry the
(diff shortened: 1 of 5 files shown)


MichalStrehovsky:
@piotrpMSFT PTAL

Cc @jkotas

jkotas:
@KrzysztofCwalina @terrajobst Could you please take look?

src/System.CommandLine/System/CommandLine/Enumerable.cs
```diff
@@ -0,0 +1,84 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.CommandLine
+{
+    // Low level replacement of LINQ Enumerable class. In particular, this implementation
+    // doesn't use generic virtual methods.
```
KrzysztofCwalina: Could you add a note to the comment explaining why we don't want GVMs in this library?
Thanks.

KrzysztofCwalina:
looks good.
I just added one small comment about a comment.

src/System.CommandLine/System/CommandLine/Enumerable.cs
MichalStrehovsky: Comment added.

mellinoe:
@dotnet-bot Test this, please

PR #656, opened 2016-04-25
Revamp build for latest VS/CLI tooling, Update dependencies

mellinoe:
This has many changes in it, some of which touch most parts of the repo:

* Delete everything related to MSBuild.
* Delete all project.lock.json files.
* Copy the `install-dotnet.ps1` script from dotnet/cli. This is used to initialize/install dotnet if it is not present in the repo. The version is pinned to a specific version of the toolset; it does not automatically roll forward.
* Add a very simple `build.ps1` script which can be used to build and test the whole repo. This uses the CLI installed by the above, and is locked to a specific version. All this script does is enumerate the files under "src", `dotnet build` them, enumerate the files under "tests", and `dotnet test` them. There is very little logic in here. I've skipped tests with trait category=outerloop because they take way too long.
* Delete unneeded things under "tools". Only the signing keys remain.
* Auto-generate all .xproj files using VS, and modify them slightly to work around a bug in the tooling related to project references. I would have preferred to not check these in, but the repo is not buildable in VS unless the workaround is present in the xproj files.
* Update some dependencies in project.json files as necessary to work with the latest version of CLI.

Some things that are still incomplete:

* System.Reflection.Metadata.Cil does not compile successfully against the newest version of the package. There were some breaking changes in the pre-release API that need to be accounted for, but I'm not sure how to do that. I've "disabled" the source and test projects for now.
* System.Drawing.Graphics builds but does not test correctly. I'm not sure how to specify the native dependency such that `dotnet test` will pull it in and deploy it appropriately. Something needs to be done for non-Windows platforms as well.
* Some projects do not successfully build or test outside Windows. This is to be expected, the repo has lots of Windows-specific things in it. Some COULD work, but aren't set up to do so. This isn't a regression.
* There is no "full-build" option outside of Windows (i.e. `build.sh`). This isn't a regression.

@KrzysztofCwalina , @joshfree , @piotrpMSFT


KrzysztofCwalina:
Shoudl the commits be squashed or you want them separate?

KrzysztofCwalina:
@mellinoe, have you tried to open and build this solution in VS?

mellinoe:
I can squash them when we're ready to merge, yeah. I'd like to address at least the Windows-related bullets from the bottom list, first, but hopefully those shouldn't take too long.

And yes, this builds in VS now, but it requires the new VS tooling and a newish version of CLI to do so. There do seem to be some issues with test projects sometimes not being detected properly, but I'll try to fix that as well.

KrzysztofCwalina:
looks good to me

src/System.Reflection.Metadata.Cil/Decoder/CilTypeProvider.cs
```diff
@@ -101,17 +101,11 @@ public CilType GetGenericTypeParameter(int index)
...
-            unmodifiedType.Append(" ");
-
-            foreach(var modifier in customModifiers)
-            {
-                unmodifiedType.Append(modifier.IsRequired ? "modreq(" : "modopt(");
-                unmodifiedType.Append(modifier.Type.ToString());
-                unmodifiedType.Append(")");
-            }
-
+            unmodifiedType.Append(isRequired ? " modreq(" : "modopt(");
```
nguerrera: I made a mistake here. There should be a space before "modopt(" too.

mellinoe:
Rebased, and switched netci.groovy to use build.ps1. We'll see if it works or if I messed up the syntax in some way.

mellinoe:
Ah, I remember that the CI changes won't actually occur until after we've checked this in. @KrzysztofCwalina Any objections to merging this in and doing a follow-up to make sure that CI is working correctly?

mellinoe:
Merging now, will monitor CI to make sure things are working.

PR #657, opened 2016-04-26
Update README build instructions

mellinoe:
Add some more specific instructions for building the projects in this repository.

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,9 @@ This project is a part of the [.NET Foundation].
 
 ## Building and Testing
 
-To find out how you can build and test .NET Core, see the [Developer Guide].
+To build the projects in this repo, you have a few options:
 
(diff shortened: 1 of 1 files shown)


mellinoe:
This should kick off a CI run.

mellinoe:
Hmm, something seems to be wrong with any of the tests using System.Slices. I think that the re-writing step isn't working properly, but I can't repro the issue locally. All of the test failures seem to indicate that the postcompile script didn't complete correctly. Investigating now and trying to get a repro...

omariom:
@mellinoe 
Look at this issue https://github.com/dotnet/corefxlab/issues/620

mellinoe:
Thanks for the hint, @omariom . It looks like those tools only work when run from a VS Developer Command Prompt... In the meantime I may try to get the CI jobs to run that command before running the build script. They were doing it previously.

mellinoe:
Going to open a different PR, this description isn't accurate now.

PR #658, opened 2016-04-27
Add a wrapper script for the CI to use

mellinoe:
ILDasm and ILAsm must be invoked from a VS Developer Command Prompt, or they will silently fail. CI will call `cibuild.cmd` in order to set the appropriate environment before executing the build. Locally, you need to use a developer command prompt to build System.Slices or any of the projects relying on it.


mellinoe:
@KrzysztofCwalina This should fix the path enumeration issue you were hitting. Note that you probably won't be able to run `scripts/build.ps1` directly from a powershell window because you need to be running from a VS developer command prompt, because the tools we have in here require it (ILDasm/ILAsm). Using build.cmd from the root directory will work, though, because it ensures the correct environment is set.

PR #660, opened 2016-04-27
Changed UnsafePointer to be able to use it with arrays.

KrzysztofCwalina:
Ultimatelly, I think it will need to pin, but this is to unblock some experiments
I am doing now.

### DIFF
--- a/src/System.Slices/System/Span.cs
+++ b/src/System.Slices/System/Span.cs
@@ -141,7 +141,7 @@ namespace System
         public unsafe void* UnsafePointer
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
-            get { return Offset.ToPointer(); }
+            get { return PtrUtils.ComputeAddress(Object, Offset).ToPointer(); }
         }
 
         /// <summary>


adamsitnik:
Hello @KrzysztofCwalina ! 

I am not sure if this can help you, but I had faced the same problem before when I was implementing memcmp for Slices. I wanted to pin Slices in single way to support all scenarios (arrays/unmanaged memory/stackalloc) and I [ended up](https://github.com/dotnet/corefxlab/blob/master/src/System.Slices/System/MemoryUtils.cs#L20) with the usage of `GCHandle.Alloc(slice.Object, GCHandleType.Pinned);` I know that it is the slow way to pin things, but that was my best idea then.

I don't know how pinning is implemented in .NET, but what I see is that you try to pin a pointer, which is not pointing to the object itself, but to some part of it. Are you sure that this will work?

KrzysztofCwalina:
No, this change will not work if the array is moved. I just wanted to add it so I can continue some experiments (even though they will not be 100% reliable). But I did talk to some people on the runtime team and once we start to productize spans, we can make it actually pin.

adamsitnik:
@KrzysztofCwalina Thanks for the explanation!

Btw if you would like to measure some of your experiments performance I encourage you to use [BenchmarkDotNet](https://github.com/PerfDotNet/BenchmarkDotNet). I am the person responsible for .NET Core support for BenchmarkDotNet and I would be very happy to help you if you have any problems using it.

KrzysztofCwalina:
@adamsitnik, I will take a look. Thanks!

PR #661, opened 2016-04-27
Update package.ps1 for new build

mellinoe:
Made a slightly more flexible script than before. We can create packages for different configurations (Debug/Release), and we can also create packages without publishing to nuget/myget. If the ApiKey parameter is passed, then publishing happens.

Nuget.exe is now downloaded as part of the script if publishing is going to happen.

@KrzysztofCwalina


mellinoe:
The version suffix should be correct now, as well.

mellinoe:
I'm going to merge this so I can update the official build definition so we can try to get new packages tonight.

PR #671, opened 2016-05-19
Added Slice overloads taking UInt32

KrzysztofCwalina:
Often, I am finding that it's more convenient to use uints for buffers, but then slice methods require cast. These overloads solve the problem.

The casts (from uint to int) are safe because the contract checks above them.

### DIFF
--- a/src/System.Slices/System/Contract.cs
+++ b/src/System.Slices/System/Contract.cs
@@ -32,11 +32,29 @@ namespace System
                 throw NewArgumentOutOfRangeException();
             }
         }
+        
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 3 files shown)


svick:
AFAIK, `uint` is never used for indexes or lengths in BCL. Why start now? What makes `uint` more convenient?

KrzysztofCwalina:
Most of the BCL is designed optimizing for what we call "usability", i.e. easy to get started with. I am pretty sure that with the stack only nature of Span<T> it won't be the easiest type to use. I think it's sweet spot will be in low level programming, in particular development of higher level framework libraries. When doing low level framework development, you are less concerned with usability, and more with correctness and perfromance. Because of the correctness goals, I often use unsigned integers to represent values that should not (or cannot) be negative.

PR #672, opened 2016-05-19
Fixed build warnings

KrzysztofCwalina:
The repo no builds with no warnings. Let's keep it as such.


mellinoe:
LGTM

mellinoe:
BTW: There's an issue with the build script that doesn't check the exit code of `dotnet` commands. You hit a test failure, but it's something I've seen locally every so often. The "smoke" tests in System.Text.Json occasionally fail. I'm going to merge this and I have a change ready that will correctly detect failures.

PR #673, opened 2016-05-19
Rename files in System.CommandLine to remove `

hughbe:
As suggested by @justinvp in dotnet/corefx#8646 to allow nuget symbols to work with this package


mellinoe:
Looks good. Thanks for the fix!

Issue #614, opened 2016-02-16
Slices: Array covariance strikes back


omariom:
@KrzysztofCwalina  @mikedn @VSadov 
```C#
var persons = new [] { new Person() };
((object[])persons).Slice()[0] = "ouch!";
Person first = persons[0];
first.Test();

public class Person
{
    private int age;

    public void Test()
    {
        Console.WriteLine(this);
        Console.WriteLine(GetType());
        Console.WriteLine(age);
    }
}
```

Outputs:
```
ouch!
System.String
5
```

We probably need to check in ctors that element type of the array is exactly `T`. That implies using Reflection.

mikedn:
Ouch.

>  That implies using Reflection.

Seems to me that `requires(array.GetType() == typeof(T[]))` should be enough and relatively efficient.

omariom:
@mike :+1: 
though for arrays the check is not as good for plain objects.
And the check is not eliminated for arrays of value types. That's sad because I was going to make ctors lighter.

```C#
[MethodImpl(MethodImplOptions.NoInlining)]
private static bool TestArrayElementType<T>(T[] array) => array.GetType() == typeof(T[]);

TestArrayElementType(new int[0]);
TestArrayElementType(new object[0]);
```

```ASM
; Lcl frame size = 32
       56                   push     rsi
       4883EC20             sub      rsp, 32
       E85658615F           call     System.Object:GetType():ref:this
       488BF0               mov      rsi, rax
       48B9925CCA8FFB7F0000 mov      rcx, 0x7FFB8FCA5C92
       E8C4B7CC5E           call     CORINFO_HELP_TYPEHANDLE_TO_RUNTIMETYPE
       483BF0               cmp      rsi, rax
       0F94C0               sete     al
       0FB6C0               movzx    rax, al
       4883C420             add      rsp, 32
       5E                   pop      rsi
       C3                   ret      

; Total bytes of code 43, prolog size 5
================================================================
; Lcl frame size = 40
       57                   push     rdi
       56                   push     rsi
       4883EC28             sub      rsp, 40
       48894C2420           mov      qword ptr [rsp+20H], rcx
       488BF1               mov      rsi, rcx
       488BCA               mov      rcx, rdx
       E80A58615F           call     System.Object:GetType():ref:this
       488BF8               mov      rdi, rax
       488B4E38             mov      rcx, qword ptr [rsi+56]
       488B4908             mov      rcx, qword ptr [rcx+8]
[...]

PR #678, opened 2016-05-20
Slices: Don't allow covariant Spans

omariom:
Fixes https://github.com/dotnet/corefxlab/issues/614

Luckily the check added is a noop for value types.

### DIFF
--- a/tests/System.Slices.Tests/BasicUnitTests.cs
+++ b/tests/System.Slices.Tests/BasicUnitTests.cs
@@ -301,5 +301,26 @@ namespace System.Slices.Tests
             Assert.Equal(original, array.Array);
             Assert.Equal(0, array.Count);
         }
+
+        [Fact]
+        public void CovariantSlicesNotSupported1()
+        {
+            object[] array = new string[10];
+            Assert.Throws<ArgumentException>(() => { var slice = new Span<object>(array); });
+        }
+
+        [Fact]
+        public void CovariantSlicesNotSupported2()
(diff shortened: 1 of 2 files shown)


tests/System.Slices.Tests/BasicUnitTests.cs
```diff
@@ -301,5 +301,26 @@ public void GetArray()
...
+
+        [Fact]
+        public void CovariantSlicesNotSupported2()
+        {
+            object[] array = new string[10];
+            Assert.Throws<ArgumentException>(() => { var slice = array.Slice(0); });
+        }
+
+        [Fact]
+        public void CovariantSlicesNotSupporte3()
```
danmosemsft: typo

omariom: Fixed. 
Thanks.

KrzysztofCwalina:
Span and ReadOnlySpan Slice methods use the ctor. But we don't need to redo the check when we slice a slice. Would it make sense to have and use a private ctor that skips the check?

omariom:
[That ctor](https://github.com/omariom/corefxlab/blob/e65fec428e709a9bc10a16f7e51d26735610f7c9/src/System.Slices/System/Span.cs#L131) is different.

KrzysztofCwalina:
Ah, you are right. Looks good.

KrzysztofCwalina:
@dotnet-bot test this please

PR #689, opened 2016-06-14
UTF-8 ulong parser method now supports byte arrays

jonfortescue:
Overloaded UTF-8 ulong parser method to support byte arrays rather than the (currently unsafe) Utf8String class.

### DIFF
--- a/src/System.Text.Formatting/System/Text/Parsing/InvariantParser_ulong.cs
+++ b/src/System.Text.Formatting/System/Text/Parsing/InvariantParser_ulong.cs
@@ -46,5 +46,49 @@ namespace System.Text.Parsing
 
             return true;
         }
+        /// <summary>
+        /// Parses an unsigned 64-bit integer from a location within a UTF-8 byte array buffer.
+        /// </summary>
+        /// <param name="text">The UTF-8 buffer, passed as a byte array.</param>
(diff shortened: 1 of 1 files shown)


src/System.Text.Formatting/System/Text/Parsing/InvariantParser_ulong.cs
```diff
@@ -46,5 +46,41 @@ public static bool TryParse(Utf8String text, out ulong value, out int bytesConsu
 
             return true;
         }
+		public static bool TryParse(byte[] text, out ulong value, out int bytesConsumed)
+        {
```
KrzysztofCwalina: There needs to be an int index parameter as well, otherwise you can only parse integers at the beginning of the array, which is not very useful.

jonfortescue: Ah, so the byte array just represents some buffer passed to us then, not necessarily the specific string we want to parse -- do I have that right? Will add that in then.

KrzysztofCwalina:
looks good! Please sign the CLA and I will merge this one in.
Thanks!

KrzysztofCwalina:
Also, you should squash the commits.

PR #700, opened 2016-06-17
Parsing for all integer types + tests.

jonfortescue:
Squashed commits from #698. See that PR for full info.

### DIFF
--- a/src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
+++ b/src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
@@ -1,51 +1,11 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
-
-using System.Diagnostics;
+﻿using System.Diagnostics;
 using System.Text.Formatting;
 using System.Text.Utf8;
 
 namespace System.Text.Parsing
 {
     public static partial class InvariantParser
-    {      
-        public static bool TryParse(Utf8String utf8Text, out ulong value, out int bytesConsumed)
-        {
-            Precondition.Require(utf8Text.Length > 0);
-
-            value = 0;
-            bytesConsumed = 0;
-
-            for (int byteIndex = 0; byteIndex < utf8Text.Length; byteIndex++)
-            {
-                byte nextByte = (byte)utf8Text[byteIndex];
-                if (nextByte < '0' || nextByte > '9')
-                {
-                    if (bytesConsumed == 0)
-                    {
-                        value = default(ulong);
-                        return false;
-                    }
-                    else
-                    {
-                        return true;
-                    }
-                }
-                ulong candidate = value * 10;
-                candidate += (ulong)nextByte - '0';
-                if (candidate >= value)
-                {
-                    value = candidate;
-                }
-                else
-                {
-                    return true;
-                }
-                bytesConsumed++;
-            }
-
-            return true;
-        }
+    {
         /// <summary>
         /// Parses an unsigned 64-bit integer from a location within a UTF-8 byte array buffer.
         /// </summary>
@@ -54,7 +14,7 @@ namespace System.Text.Parsing
         /// <param name="value">The parsed value.</param>
         /// <param name="bytesConsumed">The length (in bytes) of the unparsed value within the UTF-8 buffer.</param>
         /// <returns>True if parsing is successful; false otherwise.</returns>
-		public static bool TryParse(byte[] utf8Text, int index, out ulong value, out int bytesConsumed)
+		public static bool TryParse(byte[] utf8Text, int index, out byte value, out int bytesConsumed)
         {
             Precondition.Require(utf8Text.Length > 0);
 
@@ -68,7 +28,7 @@ namespace System.Text.Parsing
                 {
                     if (bytesConsumed == 0) // check to see if we've processed any digits at all
                     {
-                        value = default(ulong); // if we haven't, set value to 0 and return false
+                        value = default(byte); // if we haven't, set value to 0 and return false
                         return false;
                     }
                     else
@@ -76,8 +36,8 @@ namespace System.Text.Parsing
                         return true; // otherwise return true
                     }
                 }
-                ulong candidate = value * 10; // left shift the value
-                candidate += (ulong)nextByte - '0'; // parse the current digit to a ulong and add it to the temporary value
+                byte candidate = (byte)(value * 10); // left shift the value
+                candidate += (byte)(nextByte - '0'); // parse the current digit to a byte and add it to the temporary value
                 if (candidate >= value) // if it was a digit 0-9, this should be true
                 {
                     value = candidate;
@@ -91,7 +51,7 @@ namespace System.Text.Parsing
 
             return true;
         }
-        unsafe public static bool TryParse(byte *utf8Text, int index, int length, out ulong value, out int bytesConsumed)
+        unsafe public static bool TryParse(byte* utf8Text, int index, int length, out byte value, out int bytesConsumed)
         {
             value = 0;
             bytesConsumed = 0;
@@ -103,7 +63,7 @@ namespace System.Text.Parsing
                 {
                     if (bytesConsumed == 0)
                     {
-                        value = default(ulong);
+                        value = default(byte);
                         return false;
                     }
                     else
@@ -111,8 +71,8 @@ namespace System.Text.Parsing
                         return true;
                     }
                 }
-                ulong candidate = value * 10;
-                candidate += (ulong)nextByte - '0';
+                byte candidate = (byte)(value * 10);
+                candidate += (byte)(nextByte - '0');
                 if (candidate >= value)
                 {
                     value = candidate;
--- a/src/System.Text.Formatting/System/Text/Parsing/InvariantParser_int.cs
+++ b/src/System.Text.Formatting/System/Text/Parsing/InvariantParser_int.cs
@@ -1,0 +1,146 @@
+﻿using System.Diagnostics;
+using System.Text.Formatting;
+using System.Text.Utf8;
+
+namespace System.Text.Parsing
+{
+    public static partial class InvariantParser
+    {
+        /// <summary>
+        /// Parses an unsigned 64-bit integer from a location within a UTF-8 byte array buffer.
+        /// </summary>
+        /// <param name="utf8Text">The UTF-8 buffer, passed as a byte array.</param>
+        /// <param name="index">The index location of the value to be parsed within the buffer.</param>
+        /// <param name="value">The parsed value.</param>
+        /// <param name="bytesConsumed">The length (in bytes) of the unparsed value within the UTF-8 buffer.</param>
+        /// <returns>True if parsing is successful; false otherwise.</returns>
+		public static bool TryParse(byte[] utf8Text, int index, out int value, out int bytesConsumed)
+        {
+            Precondition.Require(utf8Text.Length > 0);
+
+            value = 0;
+            bytesConsumed = 0;
+            bool negative = false;
+
+            if (utf8Text[index] == '-')
+            {
+                negative = true;
+                index++;
+                bytesConsumed++;
+            }
+
+            for (int byteIndex = index; byteIndex < utf8Text.Length; byteIndex++) // loop through the byte array
+            {
+                byte nextByte = utf8Text[byteIndex];
+                if (nextByte < '0' || nextByte > '9') // if the next character is not a digit
+                {
+                    if (bytesConsumed == 1 && negative) // if the first character happened to be a '-', we reset the byte counter so logic proceeds as normal.
+                    {
+                        bytesConsumed = 0;
+                    }
+                    if (bytesConsumed == 0) // check to see if we've processed any digits at all
+                    {
+                        value = default(int); // if we haven't, set value to 0 and return false
+                        return false;
+                    }
+                    else
+                    {
+                        if (negative) // We check if the value is negative at the very end to save on comp time
+                        {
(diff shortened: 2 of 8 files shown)


src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
```diff
@@ -0,0 +1,86 @@
...
+    {
+        /// <summary>
+        /// Parses an unsigned 64-bit integer from a location within a UTF-8 byte array buffer.
+        /// </summary>
+        /// <param name="text">The UTF-8 buffer, passed as a byte array.</param>
+        /// <param name="startIndex">The index location of the value to be parsed within the buffer.</param>
+        /// <param name="value">The parsed value.</param>
+        /// <param name="bytesConsumed">The length (in bytes) of the unparsed value within the UTF-8 buffer.</param>
+        /// <returns>True if parsing is successful; false otherwise.</returns>
+		public static bool TryParse(byte[] text, int startIndex, out byte value, out int bytesConsumed)
```
KrzysztofCwalina: Feel free to fix it in a subsequent commit, but I think the parameters should be renamed:
1. The first one should be called utf8Text
2. startIndex should be called simply "index". This is a convention we use in many other places in the Framework.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_int.cs
```diff
@@ -0,0 +1,135 @@
...
+                    if (bytesConsumed == 0) // check to see if we've processed any digits at all
+                    {
+                        value = default(int); // if we haven't, set value to 0 and return false
+                        return false;
+                    }
+                    else
+                    {
+                        if (negative) // We check if the value is negative at the very end to save on comp time
+                        {
+                            value *= -1;
```
KrzysztofCwalina: why not value = -value; ?

jonfortescue: That requires a boolean check every single iteration. This one required a single check. I can change it, if you'd like!

jonfortescue: Whoops, misread what you said. I can change it to that for sure. Is there a semantic difference between the two in C#?

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_int.cs
```diff
@@ -0,0 +1,135 @@
...
+                negative = true;
+                index++;
+                bytesConsumed++;
+            }
+
+            for (int byteIndex = index; ; byteIndex++)
+            {
+                if (utf8Text[byteIndex] < '0' || utf8Text[byteIndex] > '9')
+                {
+                    if (bytesConsumed == 0)
```
KrzysztofCwalina: But if you read '-' sign before, bytesConsumed won't be zero, and so it seems like "-A" will parse as zero, but it should fail to parse.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_long.cs
```diff
@@ -0,0 +1,135 @@
...
+            if (utf8Text[index] == '-')
+            {
+                negative = true;
+                index++;
+                bytesConsumed++;
+            }
+
+            for (int byteIndex = index; byteIndex < utf8Text.Length; byteIndex++) // loop through the byte array
+            {
+                if (utf8Text[byteIndex] < '0' || utf8Text[byteIndex] > '9') // if the next character is not a digit
```
KrzysztofCwalina: It might be better to read the next byte into a local, especially if we ever port this code to the out of band (i.e. corfxlab) Span<T>, which might not be as optimized as arrays are.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_int.cs
jonfortescue: Ah, good point. I will add logic for that.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_long.cs
```diff
@@ -0,0 +1,135 @@
...
+                    else
+                    {
+                        if (negative) // We check if the value is negative at the very end to save on comp time
+                        {
+                            value *= -1;
+                        }
+                        return true; // otherwise return true
+                    }
+                }
+                long candidate = (value * 10); // left shift the value
```
KrzysztofCwalina: Do we need long local here? I think it might be expensive on some architectures (32-bit architectures) to do these long operations and back and forth casts. Feel free to file an investigation issue for this and investigate later.

KrzysztofCwalina: And in general, as part of the perf work you will be doing later, you should look at the assembly generated and compare it to some good baseline (e.g. some C libraries). If there are differences, we should either fix them here in C# or in the back end compilers (JIT/AOT).

KrzysztofCwalina:
@dotnet-bot test this please

tests/System.Text.Formatting.Tests/Parsing/ParserTests.cs
```diff
@@ -0,0 +1,389 @@
...
+using System.Text.Parsing;
+using Xunit;
+
+namespace System.Text.Formatting.Tests
+{
+    public class ParserTests
+    {
+        #region ulong
+        [Theory]
+        [InlineData(new byte[] { 0x31, 0x37, 0x32, 0x38 } /* "1728" */, true, 0, 1728, 4)]
```
KrzysztofCwalina: Feel free to do it as a separate commit, but it would be great to add some negative tests, i.e. tests that get false returned or exceptions thrown from TryParse methods.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
```diff
@@ -0,0 +1,88 @@
...
+        {
+            Precondition.Require(utf8Text.Length > 0);
+
+            value = 0;
+            bytesConsumed = 0;
+
+            for (int byteIndex = index; byteIndex < utf8Text.Length; byteIndex++) // loop through the byte array
+            {
+                byte nextByte = utf8Text[byteIndex];
+                if (nextByte < '0' || nextByte > '9') // if the next character is not a digit
```
whoisj: is the `nextByte > '9'` safe? Shouldn't you make sure that you're not mid-sequence before this step? Or am I missing something?

tests/System.Text.Formatting.Tests/Parsing/ParserTests.cs
```diff
@@ -0,0 +1,393 @@
...
+        [InlineData(new byte[] { 0xCF, 0xA0, 0x20, 0xDF, 0xB7, 0xDE, 0x96, 0xDD, 0xAE, 0x31, 0x37, 0x35, 0x31, 0x31, 0x31, 0x30 }, // stuff + "1751110"
+            true, 9, 1751110, 7)]
+        [InlineData(new byte[] { 0x39, 0x38, 0x37, 0x66, 0x61, 0x64, 0x66, 0x61, 0x6B, 0x6A, 0x6C, 0x66, 0x68, 0x6B }, // "987" + trailing text
```
whoisj: Can the data be strings passed through `System.Text.Encoding.UTF8`? If it can, that would a lot easier for most people to understand.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
jonfortescue: Not totally sure what you're getting at. Can you clarify what you mean by checking that we're not mid-sequence?

tests/System.Text.Formatting.Tests/Parsing/ParserTests.cs
KrzysztofCwalina: :+1:

jonfortescue: The purpose of these methods, as I understand it, is to read in raw byte buffers. These are the primitive parsers, so we're specifically avoiding any explicit typing.

tests/System.Text.Formatting.Tests/Parsing/ParserTests.cs
jonfortescue: I've got one two in there now, but I'll add it to my queue!

tests/System.Text.Formatting.Tests/Parsing/ParserTests.cs
KrzysztofCwalina: I think @whoisj just means to make it more readable, i.e. 
  [InlineData(Encoding.UTF8.GetBytes("1751110"), true, 9, 1751110, 7)]

or something like that

jonfortescue: @KrzysztofCwalina @whoisj Ah, yes, I understand. Will do that now!

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@jonfortescue, I fixed the nuget feed issue. But there seem to be more failures unrelated to it.

jonfortescue:
@dotnet-bot test this please

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_uint.cs
```diff
@@ -9,14 +9,14 @@ namespace System.Text.Parsing
 {
     public static partial class InvariantParser
     {
-        public static bool TryParse(string text, int index, int count, out uint value, out int charactersConsumed)
+        public static bool TryParse(string utf8Text, int index, int count, out uint value, out int charactersConsumed)
```
KrzysztofCwalina: BTW, this should not be called utf8Text. System.String is always UTF16.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+                byte candidate = (byte)(value * 10); // left shift the value
+                candidate += (byte)(nextByte - '0'); // parse the current digit to a byte and add it to the temporary value
+                if (candidate >= value) // if it was a digit 0-9, this should be true
+                {
+                    value = candidate;
+                }
+                else // this should never happen, but just in case
+                {
+                    return true;
```
KrzysztofCwalina: If it should not happen, I think you should simply have an assert here.

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
```diff
@@ -0,0 +1,88 @@
...
+                }
+                bytesConsumed++; // increment the number of bytes consumed, then loop
+            }
+
+            return true;
+        }
+        unsafe public static bool TryParse(byte* utf8Text, int index, out byte value, out int bytesConsumed)
```
KrzysztofCwalina: For pointer overloads, I think we need to always take a length parameter and stop the loop.

jonfortescue: Cool, yeah. I wasn't sure how restrictive you wanted to be. I'll add that in right away.

tests/System.Text.Formatting.Tests/Parsing/ParserTests.cs
```diff
@@ -0,0 +1,340 @@
...
+        [InlineData("987abcdefg", true, 0, 987, 3)]
+        [InlineData("-32768", true, 0, -32768, 6)]
+        [InlineData("Letthem-32684eatcake", true, 7, -32684, 6)]
+        [InlineData("I am 1", false, 0, 0, 0)]
+        public unsafe void ParseUtf8ByteStarToShort(string text, bool expectSuccess, int index, short expectedValue, int expectedBytesConsumed)
+        {
+            short parsedValue;
+            int bytesConsumed;
+
+            GCHandle arrayPinned = GCHandle.Alloc(UtfEncode(text), GCHandleType.Pinned);
```
KrzysztofCwalina: You need to "unpin" (free) if you pin using GCHandles. Though since you only need this pinned in this lexical scope, you should use "fixed" statement to pin, so you don't have to worry about freeing.

tests/System.Text.Formatting.Tests/Parsing/ParserTests.cs
```diff
@@ -0,0 +1,340 @@
...
+            byte* arrayPointer = (byte*)arrayPinned.AddrOfPinnedObject();
+
+            bool result = InvariantParser.TryParse(arrayPointer, index, 50, out parsedValue, out bytesConsumed);
```
KrzysztofCwalina: 50 should not be hardcoded

src/System.Text.Formatting/System/Text/Parsing/InvariantParser_byte.cs
whoisj: Utf-8 characters are 1 to 4 bytes long. If the value of a byte is greater than `0x7F` then you know that the byte cannot stand alone and you are in fact mid-sequence. In your case, you can likely just fail out of the test and ignore the byte.

I must have read this code while quite tired, as the `nextByte < '0'` will guard you against any issues cause by UTF-8 encoding.

jonfortescue:
@KrzysztofCwalina fixed the main bugs and squashed the commits. Everything's working on my machine, so we should be ready to merge assuming the server builds succeed.

KrzysztofCwalina:
Looks good.

PR #701, opened 2016-06-17
Changed Span<byte> copy routines to use cpblk

KrzysztofCwalina:
@raymcc, for span of bytes, the copy operations will now use cpblk, i.e. 

var source = new ReadOnlySpan<byte>(nativePointer);
var destination = new Span<byte>(managedArray);
source.TryCopyTo(destination);

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -11,6 +11,13 @@
     <add key="repositoryPath" value="..\packages" />
   </config>
   <packageSources>
+    <add key="dotnet.myget.org dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
+    <add key="dotnet.myget.org dotnet-buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/" />
+    <add key="dotnet.myget.org dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/" />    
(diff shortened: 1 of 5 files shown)


omariom:
`ComputeAddress` doesn't pin underlying arrays. Is it safe to use  it then to access  the data?

Probably it is if it is inlined, I am just not sure.

KrzysztofCwalina:
good point, @raymcc you need it just for native memory, correct? If yes, I will for now special case it to when Object==null and later we can add PtrUtils.Copy(Span,Span) and pin them

omariom:
I will start working on `PtrUtils.Copy(Span,Span) ` if no-one minds )

KrzysztofCwalina:
@omariom, sounds great!

KrzysztofCwalina:
@omariom, BTW, I think we have more problems with CompputeAddress in this repo. I tried to make all of them unsafe (by adding dummy void* parameters), but possibly I missed some. Also, i am now not sure if simply making them unsafe is the best approach.

I filed an issue to review #703.

raymcc:
On our side, we just need it for native memory.

Today we are inefficiently thunking to Win32.CopyMemory for large blocks, as there doesn’t seem to be a way to trick the compiler into emitting an inline cpblk.

KrzysztofCwalina:
@dotnet-bot test this please

PR #716, opened 2016-06-23
Redesigned how response buffers are being formatted

KrzysztofCwalina:
This change allows to write response body first and then write headers. Very useful to be able to write the content length header.

After this commit, the low allocating server is very close to zero allocations for simple JSON post and JSON response.

I don't really like the SharedData APIs/abstraction. I do like the concept. I will think about fixing the abstraction in a subsequent PR.

### DIFF
--- a/demos/LowAllocationWebServer/SampleServer.cs
+++ b/demos/LowAllocationWebServer/SampleServer.cs
@@ -3,7 +3,6 @@
 
 using Microsoft.Net.Http;
 using System;
-using System.Buffers;
 using System.Diagnostics;
 using System.Text.Formatting;
 using System.Text.Http;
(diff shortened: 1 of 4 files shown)


demos/LowAllocationWebServer/SampleServer.cs
```diff
@@ -29,93 +28,81 @@ public SampleRestServer(Log log, ushort port, byte address1, byte address2, byte
...
-        ArrayPool<byte>.Shared.Return(buffer);
+
+        var headers = response.Headers;
+        headers.AppendHttpStatusLine(HttpVersion.V1_1, 200, new Utf8String("OK"));
+        headers.Append("Content-Length : ");
+        headers.Append(response.Body.CommitedBytes);
+        headers.AppendHttpNewLine();
+        headers.Append("Content-Type : text/plain; charset=UTF-8");
+        headers.AppendHttpNewLine();
+        headers.Append("Server : .NET Core Sample Serve");
```
ilmax: typo? .NET Core Sample Server?

KrzysztofCwalina: thanks! will fix

PR #723, opened 2016-06-23
Added parsing support for explicitly positive signed integers.

jonfortescue:
This is just a small fix&emdash;if a signed integer is explicitly positive (i.e. +46) we should still be able to parse it.


KrzysztofCwalina:
looks good.

Issue #443, opened 2015-11-12
Non-allocating parsing APIs


JamesNK:
One thing I'd find very useful are fast methods for parsing characters to int, double, decimal, with false returned if the number overflows or contains invalid characters.

I know there are some here for uint and ulong - https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Formatting/src/System/Text/Parsing/InvariantParser.cs. I'd like it to be expanded for signed integers, doubles and decimals.

KrzysztofCwalina:
@JamesNK, we definitely want to do these, but it's going to take some time.

KrzysztofCwalina:
I think it's now covered in https://github.com/dotnet/corefx/blob/master/src/System.Memory/ref/System.Memory.cs#L335. 

I will close the issue. @JamesNK, if you still need some datatype that is not covered, please open an issue in corfx repo. thanks.

PR #730, opened 2016-06-24
Added support for handling integer overflow

jonfortescue:
Primitive parsing APIs will now return false if overflow occurs (per #443). Also added tests for these cases, which should neatly fulfill #704 for now.

### DIFF
--- a/src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
@@ -35,17 +35,20 @@ namespace System.Text
                         return true; // otherwise return true
                     }
                 }
-                byte candidate = (byte)(value * 10); // left shift the value
-                candidate += (byte)(nextByte - '0'); // parse the current digit to a byte and add it to the temporary value
-                if (candidate >= value) // if it was a digit 0-9, this should be true
+                try
                 {
+                    byte candidate = checked((byte)(value * 10 + nextByte - '0')); // left shift the value and add the nextByte.
+                    Debug.Assert(candidate >= value);
+
                     value = candidate;
+                    bytesConsumed++; // increment the number of bytes consumed, then loop
                 }
-                else // this should never happen, but just in case
+                catch (OverflowException e) // catch any overflow
                 {
-                    return true;
+                    value = 0;
+                    bytesConsumed = 0;
+                    return false;
                 }
-                bytesConsumed++; // increment the number of bytes consumed, then loop
             }
 
             return true;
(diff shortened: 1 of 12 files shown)


src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -76,6 +78,12 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out by
                 {
                     value = candidate;
                 }
+                else // overflow
```
KrzysztofCwalina: Is this a reliable check for overflow? i.e are you sure that (value * 10 + digit) % 255 cannot be greater than value, which would mean the check is incorrect for at least parsing byte values.

jonfortescue:
Ah, you're right. I'll have to rethink this next week.

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -35,19 +35,23 @@ public static bool TryParse(byte[] utf8Text, int index, out byte value, out int
...
                     }
                 }
-                byte candidate = (byte)(value * 10); // left shift the value
-                candidate += (byte)(nextByte - '0'); // parse the current digit to a byte and add it to the temporary value
-                if (candidate >= value) // if it was a digit 0-9, this should be true
+                try
                 {
-                    value = candidate;
+                    byte candidate = (byte)(value * 10); // left shift the value
+                    candidate = checked((byte)(candidate + nextByte - '0')); // parse the current digit to a byte and add it to the temporary value
```
KrzysztofCwalina: Shouldn't both of the lines (above) be checked?

jonfortescue: True. I'm is there any real reason we have it split into two lines? I'll just collapse it if not.

KrzysztofCwalina: No reason. Go ahead and collapse.

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -35,17 +35,22 @@ public static bool TryParse(byte[] utf8Text, int index, out byte value, out int
...
                 }
-                byte candidate = (byte)(value * 10); // left shift the value
-                candidate += (byte)(nextByte - '0'); // parse the current digit to a byte and add it to the temporary value
-                if (candidate >= value) // if it was a digit 0-9, this should be true
+                try
                 {
-                    value = candidate;
+                    byte candidate = checked((byte)(value * 10 + nextByte - '0')); // left shift the value and add the nextByte.
+
+                    if (candidate >= value) // if it was a digit 0-9, this should be true
```
KrzysztofCwalina: Can it be every smaller?

jonfortescue: Not unless something completely unexpected occurs.

Line 26's if statement makes it so 0 <= `nextByte - '0'` <= 9. If overflow occurs, checked will take care of that. If value is 0 and `nextByte - '0'` is 0, then candidate == value; otherwise, candidate > value.

KrzysztofCwalina: Then why do you have the if check? Maybe Debug.Assert would be better?

jonfortescue: Sure, we can switch it to a Debug.Assert.

KrzysztofCwalina:
Looks good.
Thanks!

PR #732, opened 2016-07-02
Slices: Made copying methods use cpblk without pinning

### LINKED ISSUE #699: Span<T> copy methods should use cpblk instruction

omariom:
Fixes #699

`cpblk` can work with managed pointers. This PR exploits that.

[The generated code ](https://gist.github.com/omariom/543db8c8502ca42492e057e02460d685)for Spans of `byte`, ` char`, `DateTime` and `string`.

For the predefined list of primitives it uses `cpblk`, for types not in the list and ref types it loops. The decision is made at JIT time.

@KrzysztofCwalina @adamsitnik

### DIFF
--- a/src/System.Slices/System/ReadOnlySpan.cs
+++ b/src/System.Slices/System/ReadOnlySpan.cs
@@ -192,22 +192,28 @@ namespace System
         /// <param name="destination">The span to copy items into.</param>
         public bool TryCopyTo(Span<T> destination)
         {
-            if (Length > destination.Length) {
+            // There are some benefits of making local copies. See https://github.com/dotnet/coreclr/issues/5556
+            var dest = destination;
+            var src = this;
+
+            if (src.Length > dest.Length)
                 return false;
+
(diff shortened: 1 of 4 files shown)


src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -336,18 +333,5 @@ public override bool Equals(object obj)
...
-        /// <summary>
-        /// Returns item from given index without the boundaries check
-        /// use only in places where moving outside the boundaries is impossible
-        /// gain: performance: no boundaries check (single operation) 
-        /// </summary>
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal T GetItemWithoutBoundariesCheck(int index)
-        {
-            return PtrUtils.Get<T>(Object, Offset, (UIntPtr)index);
-        }
```
omariom: I removed  `GetItemWithoutBoundariesCheck` -  it is not inlined for reference types.

KrzysztofCwalina:
Nice. Thanks!

PR #736, opened 2016-07-02
Don't Slice ReadOnlySpan to Span

### LINKED ISSUE #735: Don't Slice ReadOnlySpan to Span
It breaks the immutability guarantees. 

Otherwise methods will still have to defensive copy the array rather than return a static array wrapped in a ReadOnlySpan for fixed items like lookup tables etc otherwise they can be changed by slicing.

benaadams:
It breaks the immutability guarantees. 

Otherwise methods will still have to defensive copy the array rather than return a static array wrapped in a ReadOnlySpan for fixed items like lookup tables etc as they could be changed by slicing.

Resolves  #735

@jkotas @KrzysztofCwalina


KrzysztofCwalina:
Yeah, makes total sense. Good catch.
Thanks!

PR #738, opened 2016-07-05
Implicit conversion to Span from Array

### LINKED ISSUE #734: Implicit conversion to Span from Array?
Extending the `Span`->`ReadOnlySpan` implicit conversion theme. 

For Span
```csharp
public static implicit operator Span<T>(T[] array)
{
    return new Span<T>(array);
}
```
For ReadOnlySpan
```csharp
public static implicit operator ReadOnlySpan<T>(T[] array)
{
    return new ReadOnlySpan<T>(array);
}
```
This would allow new apis to only need a single method which would convey intent rather than also requiring array overloads which are ambiguous in intention as the method name is generally reversed in action with the parameter.

e.g. `Stream.Write(ReadOnlySpan<byte> array)` and `Stream.Read(Span<byte> array)`

```csharp
SpanAwareStream s;
byte[] a = new byte[255];

s.Write(a); // conversion to ReadOnlySpan
s.Read(a); // conversion to Span
```

Would it have unintended side-effects?

benaadams:
Extending the `Span`->`ReadOnlySpan` implicit conversion theme. 

This would allow new apis to only need a single method which would convey intent rather than also requiring array overloads which are ambiguous in intention as the method name is generally reversed in action with the parameter.

e.g. `Stream.Write(ReadOnlySpan<byte> array)` and `Stream.Read(Span<byte> array)`

```csharp
SpanAwareStream s;
byte[] a = new byte[255];

s.Write(a); // conversion to ReadOnlySpan
s.Read(a); // conversion to Span
```
Resolves #734

@KrzysztofCwalina @jkotas @jaredpar

### DIFF
--- a/src/System.Slices/System/ReadOnlySpan.cs
+++ b/src/System.Slices/System/ReadOnlySpan.cs
@@ -133,6 +133,11 @@ namespace System
             Length = length;
         }
 
+        public static implicit operator ReadOnlySpan<T>(T[] array)
+        {
+            return new ReadOnlySpan<T>(array);
+        }
+
         public static implicit operator ReadOnlySpan<T>(Span<T> slice)
         {
             return new ReadOnlySpan<T>(slice.Object, slice.Offset, slice.Length);
@@ -325,6 +330,11 @@ namespace System
             return ReferenceEquals(other);
         }
 
+        public bool Equals(Span<T> other)
+        {
+            return other.StructuralEquals(Object, Offset, Length);
+        }
+
         public override bool Equals(object obj)
         {
             if (obj is ReadOnlySpan<T>)
@@ -333,5 +343,8 @@ namespace System
             }
             return false;
         }
+
+        public static bool operator ==(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => span1.Equals(span2);
+        public static bool operator !=(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => !span1.Equals(span2);
     }
 }
--- a/src/System.Slices/System/ReadOnlySpan.cs
+++ b/src/System.Slices/System/ReadOnlySpan.cs
@@ -12,7 +12,7 @@ namespace System
     /// </summary>
     [DebuggerTypeProxy(typeof(SpanDebuggerView<>))]
     [DebuggerDisplay("Length = {Length}")]
-    public partial struct ReadOnlySpan<T> : IEnumerable<T>, IEquatable<ReadOnlySpan<T>>
+    public partial struct ReadOnlySpan<T> : IEnumerable<T>, IEquatable<Span<T>>, IEquatable<ReadOnlySpan<T>>, IEquatable<T[]>
     {
         /// <summary>A managed array/string; or null for native ptrs.</summary>
         internal readonly object Object;
@@ -344,6 +344,7 @@ namespace System
             return false;
         }
 
+        public bool Equals(T[] other) => Equals(new ReadOnlySpan<T>(other));
         public static bool operator ==(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => span1.Equals(span2);
         public static bool operator !=(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => !span1.Equals(span2);
     }
--- a/src/System.Slices/System/Span.cs
+++ b/src/System.Slices/System/Span.cs
@@ -16,7 +16,7 @@ namespace System
     /// </summary>
     [DebuggerTypeProxy(typeof(SpanDebuggerView<>))]
     [DebuggerDisplay("Length = {Length}")]
-    public partial struct Span<T> : IEnumerable<T>, IEquatable<Span<T>>
+    public partial struct Span<T> : IEnumerable<T>, IEquatable<Span<T>>, IEquatable<ReadOnlySpan<T>>, IEquatable<T[]>
     {
         /// <summary>A managed array/string; or null for native ptrs.</summary>
         internal readonly object Object;
@@ -373,6 +373,7 @@ namespace System
             return ReferenceEquals(other);
         }
 
+
         public override bool Equals(object obj)
         {
             if (obj is Span<T>)
@@ -383,6 +384,7 @@ namespace System
         }
 
         public bool Equals(ReadOnlySpan<T> other) => StructuralEquals(other.Object, other.Offset, other.Length);
+        public bool Equals(T[] other) => Equals(new Span<T>(other));
         public static bool operator ==(Span<T> span1, Span<T> span2) => span1.Equals(span2);
         public static bool operator !=(Span<T> span1, Span<T> span2) => !span1.Equals(span2);
 
--- a/src/System.Slices/System/Span.cs
+++ b/src/System.Slices/System/Span.cs
@@ -139,6 +139,11 @@ namespace System
             Length = length;
         }
 
+        public static implicit operator Span<T>(T[] array)
+        {
+            return new Span<T>(array);
+        }
+
         public static Span<T> Empty { get { return default(Span<T>); } }
 
         public bool IsEmpty
@@ -339,6 +344,12 @@ namespace System
                 Offset == other.Offset && Length == other.Length;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        internal bool StructuralEquals(object obj, UIntPtr offset, int length)
+        {
+            return Object == obj && Offset == offset && Length == length;
+        }
+
         public override int GetHashCode()
         {
             unchecked
@@ -370,6 +381,10 @@ namespace System
             }
             return false;
         }
+
+        public bool Equals(ReadOnlySpan<T> other) => StructuralEquals(other.Object, other.Offset, other.Length);
+        public static bool operator ==(Span<T> span1, Span<T> span2) => span1.Equals(span2);
+        public static bool operator !=(Span<T> span1, Span<T> span2) => !span1.Equals(span2);
 
         public ReadOnlySpan<T>.Enumerator GetEnumerator()
         {
--- a/tests/System.Slices.Tests/EqualityTests.cs
+++ b/tests/System.Slices.Tests/EqualityTests.cs
@@ -34,11 +34,141 @@ namespace System.Slices.Tests
             SpansReferencingSameMemoryAreEqualInEveryAspect(ref span, ref structCopy);
         }
 
+        [Theory]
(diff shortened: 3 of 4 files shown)


benaadams:
PtrUtils seem to be broken?

omariom:
There is `TryCopyTo(T[] destination)` in Span and ROSpan. It is not needed with your additions.

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -133,6 +133,11 @@ internal ReadOnlySpan(object obj, UIntPtr offset, int length)
             Length = length;
         }
 
+        public static implicit operator ReadOnlySpan<T>(T[] array)
+        {
+            return new ReadOnlySpan<T>(array);
+        }
+
```
KrzysztofCwalina: Might be good to add tests for these.

benaadams:
@omariom this change is for other direction array->span

But the tests fail because:
`GetArray`->`TryGetArray`->`PtrUtils.SizeOf<T>` is returning zero
`SetArray`->`Set`->`TryCopyTo`->`PtrUtils.CopyBlock` doesn't copy (perhaps because sizeof is returning zero)
but the other path doesn't seem to copy either as `PtrUtils.Get<T>` returns zero :cry:

jaredpar:
👍

omariom:
@benaadams 
> GetArray->TryGetArray->PtrUtils.SizeOf<T> is returning zero

It happens when the postbuild step rewriting IL is not run or fails to rewrite IL bodied members.

omariom:
@benaadams Try to close and reopen the PR. It sometimes helped in the past )

omariom:
@benaadams  @KrzysztofCwalina 

Found the cause. With your change C# [here ](https://github.com/benaadams/corefxlab/blob/implicit-convert-array/tests/System.Slices.Tests/UsageScenarioTests.cs#L43) prefers to convert the array to a `Span<byte>` and now chooses the overload  `Equals(Span<T> other)` instead of `Equals(object obj)`.

benaadams:
Hmm, maybe should add an `Equals(T[] array)` overload?

jaredpar:
@benaadams 

> Hmm, maybe should add an Equals(T[] array) overload?

I'd think about the other direction.  Is there a reason why we don't want this particular line to return `true`?  

``` cs
Assert.False(span.Equals(array));
```

This particular change is further removing the differences between `Span<T>` and `T[]`.  Allowing them to be `Equals` here would seem to be another step in that direction.  

There should be an explicit test though to verify equality on an actual `T[]` remains `false`:

``` cs
Assert.False(span.Equals((object)array));
```

omariom:
At there is `string` as `ReadOnlySpan<char>`.

benaadams:
>  Is there a reason why we don't want this particular line to return true?

Lol, I thought it was returning false and was meant to return true :)

benaadams:
@jaredpar question that would imply a readonlyspan == span if its data is the same? Which I'd go for...

jaredpar:
@benaadams I'd agree.  Again with the caveat that `ReadOnlySpan<T>.Equals((object)Span<T>))` is still `false`.

benaadams:
Needs tests but something like that?

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -338,5 +338,19 @@ public override bool Equals(object obj)
...
         }
+
+        public static bool operator ==(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => span1.Equals(span2);
+        public static bool operator !=(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => !span1.Equals(span2);
+        public bool Equals(Span<T> other) => other.ReferenceEquals(Object, Offset, Length);
+
+        public static bool operator ==(ReadOnlySpan<T> span, T[] array)
+        {
+            return span.Object == array && 
+                span.Length == array.Length && span.Offset == new UIntPtr((uint)SpanHelpers<T>.OffsetToArrayData);
```
jaredpar: Why not use `ReferenceEquals(object, UIntPtr, int)` here? Will keep core equality logic in one place.

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -338,5 +338,19 @@ public override bool Equals(object obj)
             }
             return false;
         }
+
+        public static bool operator ==(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => span1.Equals(span2);
+        public static bool operator !=(ReadOnlySpan<T> span1, ReadOnlySpan<T> span2) => !span1.Equals(span2);
+        public bool Equals(Span<T> other) => other.ReferenceEquals(Object, Offset, Length);
```
jaredpar: I'd consider calling it something other than `ReferenceEquals`.  That is really easy to confuse with `object.ReferenceEquals` and this is not actually doing reference equality checks under the hood.  It's doing a combination of reference and value equality.

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -338,5 +338,19 @@ public override bool Equals(object obj)
...
+
+        public static bool operator ==(ReadOnlySpan<T> span, T[] array)
```
omariom: array can be null

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -338,5 +338,19 @@ public override bool Equals(object obj)
...
+        {
+            return span.Object == array && 
+                span.Length == array.Length && span.Offset == new UIntPtr((uint)SpanHelpers<T>.OffsetToArrayData);
+        }
+
+        public static bool operator !=(ReadOnlySpan<T> span, T[] array) => !(span == array);
+        public static bool operator ==(T[] array, ReadOnlySpan<T> span) => span == array;
+        public static bool operator !=(T[] array, ReadOnlySpan<T> span) => span != array;
```
jaredpar: These last four operator definitions should be unnecessary.  Given that you already have an implicit conversion from `T[]` to `ReadOnlySpan<T>` then any comparison `T[] == ReadOnySpan<T>` should bind to the standard `ReadOnySpan<T> == ReadOnlySpan<T>` operator.

src/System.Slices/System/Span.cs
```diff
@@ -376,6 +382,24 @@ public override bool Equals(object obj)
...
+
+        public static bool operator ==(Span<T> span, T[] array)
+        {
+            return span.Object == array && 
+                span.Length == array.Length && span.Offset == new UIntPtr((uint)SpanHelpers<T>.OffsetToArrayData);
+        }
+
+        public static bool operator !=(Span<T> span, T[] array) => !(span == array);
+        public static bool operator ==(T[] array, Span<T> span) => span == array;
+        public static bool operator !=(T[] array, Span<T> span) => span != array;
```
jaredpar: Same comment about these last four being unnecessary.

src/System.Slices/System/ReadOnlySpan.cs
benaadams: removed the operator will throw a null ref via ctor now

src/System.Slices/System/ReadOnlySpan.cs
benaadams: Changed to `StructuralEquals`

src/System.Slices/System/ReadOnlySpan.cs
benaadams: Removed

src/System.Slices/System/Span.cs
benaadams: Removed

src/System.Slices/System/ReadOnlySpan.cs
benaadams: Added tests

benaadams:
May have added too many tests...

jaredpar:
> May have added too many tests... 

There is no such thing.  :smile:

👍

khellang:
Is `IEquatable<T>` still a thing? Is it worth adding it, since the two types now implement the contract?

benaadams:
@khellang added the `IEquatable` impl

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -333,5 +343,9 @@ public override bool Equals(object obj)
             }
             return false;
         }
+
+        public bool Equals(T[] other) => Equals(new ReadOnlySpan<T>(other));
```
khellang: Are these necessary when there are implicit conversions in place? Not sure on overload resolution off the top of my head :stuck_out_tongue:

benaadams: IEquatable gets upset

omariom:
@benaadams @khellang 
If `Span` is going to be a `ref like` type it won't be possible to cast it to an interface.

benaadams:
@omariom not sure I follow? Already has IEnumerable and IEquatable for other types, just not for T[] and the other span type.

src/System.Slices/System/ReadOnlySpan.cs
khellang: Yeah, I guess my question was if `IEquatable<ReadOnlySpan<T>>` needed to be implemented...

omariom:
@benaadams 
Implementing interface means value can be boxed which won't be allowed for spans.  Spans can only live on stack. 
This is how I understood the plans.

/cc @KrzysztofCwalina

jkotas:
Right, the current implementation in corefxlab is experimental - the ultimate implementation in the runtime won't be same.

KrzysztofCwalina:
@jkotas, can byref-based span implement interfaces, as long as they are used just for constraints, not for casting. If yes, then CLR Span would implement the interface and then polymorphic code would have to do:

Compare\<TEquatable, T\>(TEquatable left, TEquatable right) where TEquatable : IEquatable\<T\>
{ ... }

jkotas:
byref-based span has the same limitation as byrefs and other byref-like types - it cannot be used as generic argument.

KrzysztofCwalina:
Why don't we allow it in contravariant position?

jkotas:
contravariant position does not prevent it from being stored as a field, etc.

KrzysztofCwalina:
Ah, makes sense. We would need scoped/borrowed constraint, which won't happen anytime soon.

davidfowl:
@KrzysztofCwalina what's up with this PR?

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
@benaadams, thanks for this nice addition!

PR #741, opened 2016-07-21
Complete and working xunit perf tests.

jonfortescue:


NuGet.Config
```diff
@@ -13,7 +13,7 @@
   <packageSources>
     <add key="dotnet.myget.org dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
     <add key="dotnet.myget.org dotnet-buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/" />
-    <add key="dotnet.myget.org dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/" />
```
mellinoe: Undo

global.json
```diff
@@ -1,3 +1,3 @@
 {
-  "projects": [ "src" ]
+  "projects": [ "src", "tests" ]
```
mellinoe: I think we can undo this as well, it was only for the BenchmarkDotNet stuff, right?

jonfortescue: ah, that's right! I'll undo both.

mellinoe:
LGTM :+1:

tests/System.Text.Primitives.Tests/PrimitivesPerfTests.cs
```diff
@@ -0,0 +1,875 @@
...
+        [InlineData("255")] // Max value
+        [InlineData("0")] // Min value
+        private static void BaselineByteParse(string text)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                byte value;
+                using (iteration.StartMeasurement())
+                {
+                    Byte.TryParse(text, NumberStyles.None, CultureInfo.InvariantCulture, out value);
```
KrzysztofCwalina: This work is too inexpensive. I worry that the cost of the framework calls is a big par of it. Would it make sense to parse this multiple times per iteration?

tests/System.Text.Primitives.Tests/PrimitivesPerfTests.cs
```diff
@@ -0,0 +1,875 @@
...
+        [InlineData("0128")] // leading zero
+        [InlineData("00000000128")] // many leading zeroes
+        [InlineData("-128")] // negative value handling for unsigned types
+        [InlineData("255")] // Max value
+        [InlineData("0")] // Min value
+        private static void ByteArrayToByte(string text)
+        {
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                byte[] utf8ByteArray = UtfEncode(text);
```
KrzysztofCwalina: You should move encoding out of the loop.

KrzysztofCwalina:
I added some comments in the sources. In addition:
1. Please add instructions on how to run these tests.
2. The main value of the new parasers is that: a) they can parse UTF8, b) the can parse byte[] as opposed to strings, c) they can parse unknown lenght subsections of byte[] buffers, as opposed to having to have exact string representing the parsed type. It would be great to add perf comparisons for the scenarios (above) where the new parsing routines would shine.

jonfortescue:
Wait, for some reason my sln file was changed. Fixing now.

KrzysztofCwalina:
looks good!

PR #743, opened 2016-08-01
Added new perf tests to show off a few more scenarios.

jonfortescue:
New perf tests added to show off server scenario perf increases. Updated tests to remove setup from foreach loop.

### DIFF
--- a/tests/System.Text.Primitives.Tests/PrimitivesPerfTests.cs
+++ b/tests/System.Text.Primitives.Tests/PrimitivesPerfTests.cs
@@ -58,10 +58,10 @@ namespace System.Text.Primitives.Tests
         [InlineData("0")] // Min value
         private static void BaselineArrayByteParse(string text)
         {
+            byte[] utf8ByteArray = UtfEncode(text);
             foreach (var iteration in Benchmark.Iterations)
             {
                 byte value;
-                byte[] utf8ByteArray = UtfEncode(text);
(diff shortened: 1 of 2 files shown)


tests/System.Text.Primitives.Tests/PrimitivesPerfTests.cs
```diff
@@ -142,21 +142,72 @@ private unsafe static void ByteStarUnmanagedToByte(string text)
         {
 			int length = text.Length;
 			byte[] utf8ByteArray = UtfEncode(text);
```
KrzysztofCwalina: these two lines don't seemed to be indented properly.

tests/System.Text.Primitives.Tests/PrimitivesPerfTests.cs
```diff
@@ -709,21 +927,78 @@ private unsafe static void ByteStarUnmanagedToSbyte(string text)
...
+            byte* unmanagedBytePtr;
```
KrzysztofCwalina: the 2 lines above are not indented properly.

KrzysztofCwalina:
looks good.

jonfortescue:
Not sure what was up with the formatting exactly, but I just did a global ctrl-K,D on the document, so everything should be formatted properly now.

KrzysztofCwalina:
looks good!

PR #744, opened 2016-08-02
Perf Improvement for InvariantParser

jonfortescue:
Significant perf gains by removing dependency on exception handling. Also added new tests to test edge cases.

### DIFF
--- a/src/System.Text.Primitives/System/Text/InvariantParser_int.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_int.cs
@@ -38,8 +38,8 @@ namespace System.Text
 
             for (int byteIndex = index; byteIndex < utf8Text.Length; byteIndex++) // loop through the byte array
             {
-                byte nextByte = utf8Text[byteIndex];
-                if (nextByte < '0' || nextByte > '9') // if the next character is not a digit
+                byte nextByteVal = (byte)(utf8Text[byteIndex] - '0');
+                if (nextByteVal > 9) // if the next character is not a digit
                 {
                     if (bytesConsumed == 1 && signed) // if the first character happened to be a '-' or a '+', we reset the byte counter so logic proceeds as normal.
                     {
@@ -52,38 +52,28 @@ namespace System.Text
                     }
                     else
                     {
-                        if (negative) // We check if the value is negative at the very end to save on comp time
+                        if (negative && value != Int32.MinValue) // We check if the value is negative at the very end to save on comp time
                         {
                             value = -value;
                         }
                         return true; // otherwise return true
                     }
                 }
-                //int candidate = 0; // we declare candidate out of the try-block so that it is still available to the catch (only necessary for signed types)
-                try
+                else if (value > Int32.MaxValue / 10) // overflow
                 {
-                    int candidate = checked(value * 10 + (nextByte - '0')); // parse the current digit to an int and add it to the left-shifted value
-                    Debug.Assert(candidate >= value);
-
-                    value = candidate;
-                    bytesConsumed++; // increment the number of bytes consumed, then loop
+                    value = 0;
+                    bytesConsumed = 0;
+                    return false;
                 }
-                catch (OverflowException e)
+                else if (value > 0 && (Int32.MaxValue - value * 10) + (negative ? 1 : 0) < nextByteVal) // overflow (+1 allows for overflow for min value)
                 {
-                    int candidate = value * 10 + (nextByte - '0');
-                    if (negative && candidate == int.MinValue) // overflow is expected when calculating the minimum value.
-                    {
-                        bytesConsumed++;
-                        value = candidate;
-                        return true;
-                    }
-                    else
-                    {
-                        value = 0;
-                        bytesConsumed = 0;
-                        return false;
-                    }
+                    value = 0;
+                    bytesConsumed = 0;
+                    return false;
                 }
+                int candidate = value * 10 + (nextByteVal); // parse the current digit to an int and add it to the left-shifted value
+                value = candidate;
+                bytesConsumed++; // increment the number of bytes consumed, then loop
             }
 
             if (negative) // We check if the value is negative at the very end to save on comp time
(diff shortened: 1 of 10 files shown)


jonfortescue:
To give you an idea of what these perf gains look like, here's a few charts:

0_byte
3_ulong
6_int

joshfree:
cc @mellinoe @Priya91

src/System.Text.Primitives/System/Text/InvariantParser_ulong.cs
```diff
@@ -30,20 +30,23 @@ public static bool TryParse(Utf8String utf8Text, out ulong value, out int bytesC
...
+                    return false;
                 }
-                catch (OverflowException e)
+                else if (UInt64.MaxValue - value * 10 < (ulong)(nextByte - '0')) // overflow
                 {
                     value = 0;
                     bytesConsumed = 0;
                     return false;
                 }
+                ulong candidate = checked(value * 10 + nextByte - '0');
```
mellinoe: Why is this one checked / Assert'd?

mellinoe: I think the assertion is fine btw, you could even consider adding it back to the other type variants.

src/System.Text.Primitives/System/Text/InvariantParser_ulong.cs
```diff
@@ -30,20 +30,23 @@ public static bool TryParse(Utf8String utf8Text, out ulong value, out int bytesC
                         return true;
                     }
                 }
-                try
+                else if (value > UInt64.MaxValue / 10) // overflow
                 {
-                    ulong candidate = checked(value * 10 + nextByte - '0');
```
mellinoe: You might want to pull out this `nextByte - '0'` expression into a local for clarity. You could probably re-use it above in this check as well: `if (nextByte < '0' || nextByte > '9')`
i.e.

```C#
var nextByteVal = nextByte - '0';
if (nextByteVal < 9) // Values below '0' will wrap from the subtraction.
   ...
```

src/System.Text.Primitives/System/Text/InvariantParser_ulong.cs
jonfortescue: Ah, good catch. We don't use the Utf8String class yet, so I hadn't made the change fully here.

I removed the asserts from the signed classes because we want to allow overflow in the case of the min value; but I may add it back if it proves useful in some cases.

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
```diff
@@ -98,10 +88,18 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out in
             value = 0;
             bytesConsumed = 0;
             bool negative = false;
+            bool signed = false;
 
             if (utf8Text[index] == '-')
             {
                 negative = true;
+                signed = true;
```
mellinoe: Can't you remove this `signed` flag and just not increment bytesConsumed in these two cases?

src/System.Text.Primitives/System/Text/InvariantParser_ulong.cs
jonfortescue: that's a fabulous idea and we should get minor perf bonuses from doing less checks. Doing this now

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
jonfortescue: Have to increment bytesConsumed so that we can accurately report how far into the buffer we read.

mellinoe:
Regarding precondition checks:

```CSharp
public static bool TryParse(byte[] utf8Text, int index, out byte value, out int bytesConsumed)
         {
             Precondition.Require(utf8Text.Length > 0);
             ...
```
Don't we need to validate that the initial index given is valid? I think you will just get an `IndexOutOfRangeException` if you pass, for example, `TryParse(new byte[0], index:500, out val, out consumed)`. Also, it seems like we could just return `false` if the array passed in is empty. Won't this Precondition method throw an exception? I don't think we want a "Try" method to throw.

Changing the preconditions like this might affect perf very slightly.

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
mellinoe: But don't we reset the byteCounter in the next step if `signed` is true?

jonfortescue: We only do that if the next character is not a digit. e.g. if the user submits a buffer with "-sdf", we reset the byte counter to zero and return false.

mellinoe: Actually, I might need some clarification on this part:

```CSharp
if (bytesConsumed == 1 && signed) // if the first character happened to be a '-', we reset the byte counter so logic proceeds as normal.
{
    bytesConsumed = 0;
}
```

It seems like `signed` should only ever be true if the very first character was '+' or '-'. Otherwise, we would have just aborted parsing. Maybe I'm not understanding this part.

src/System.Text.Primitives/System/Text/InvariantParser_ulong.cs
mellinoe: We might see slight improvements. I think ideally the JIT would recognize this common subexpression and optimize it out. But it probably can't coordinate that with the initial range check I mentioned above, so there's probably a bit of a benefit.

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
jonfortescue: I think you might be missing the logic for determining the where the buffer ends.

    if (bytesConsumed == 0) // check to see if we've processed any digits at all
    {
           value = default(byte); // if we haven't, set value to 0 and return false
           return false;
    }

If no bytes were consumed, we return `false` (the user passed an invalid buffer). However, if one or more bytes were consumed, we assume that we were passed a valid number and return `true`.

Because of this logic, if the user passes a buffer of the form "-asfsd", the code will return `true` without the check you mentioned. The check you mentioned specifically looks and says "if we've consumed exactly one byte and that byte was a sign, then we just were directed to a + or a - sign in the middle of an invalid buffer (e.g. '-asdf'). No actual numbers were found, so we should return false."

mellinoe: That makes sense. Ignore this one!

src/System.Text.Primitives/System/Text/InvariantParser_ulong.cs
jonfortescue: Added a fix for this!

KrzysztofCwalina:
@jonfortescue, aside from the comments people already made, it looks good to me. Please feel free to merge after you respond to the comments.

jonfortescue:
@mellinoe Agreed with reworking preconditions. Will do in a future PR.

davidfowl:
👏

PR #747, opened 2016-08-03
Removed potential exception throwing preconditions.

jonfortescue:
Preconditions raise exceptions when they are invalid. Since we want to avoid exceptions in our code (as it's a TryParse method), I have replaced these with simple if statements that return false if error conditions are met.

### DIFF
--- a/src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
@@ -15,7 +15,13 @@ namespace System.Text
         /// <returns>True if parsing is successful; false otherwise.</returns>
 		public static bool TryParse(byte[] utf8Text, int index, out byte value, out int bytesConsumed)
         {
-            Precondition.Require(utf8Text.Length > 0);
+            // Precondition replacement
+            if (utf8Text.Length < 1 || index < 0 || index >= utf8Text.Length)
+            {
+                value = 0;
+                bytesConsumed = 0;
+                return false;
+            }
 
             value = 0;
             bytesConsumed = 0;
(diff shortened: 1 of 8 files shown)


KrzysztofCwalina:
I would not do this change. The previously removed exceptions were actually being thrown during normal parsing, i.e. the caller could not do any checks before calling the method to avoid the exceptions. These exceptions are bugs in calling code and should not happen in normal programs.

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -15,7 +15,13 @@ public static partial class InvariantParser
         /// <returns>True if parsing is successful; false otherwise.</returns>
 		public static bool TryParse(byte[] utf8Text, int index, out byte value, out int bytesConsumed)
         {
-            Precondition.Require(utf8Text.Length > 0);
+            // Precondition replacement
+            if (utf8Text.Length < 1 || index < 0 || index > utf8Text.Length)
```
mellinoe: `index` needs to be strictly less than utf8Text.Length, since you can only access up to `Length - 1` in an array.

jonfortescue: good catch. Thanks

jonfortescue:
@KrzysztofCwalina Sounds good. Will revert to using Precondition, but with the additional preconditions now.

mellinoe:
@KrzysztofCwalina Does it make sense to be consistent with `int.TryParse`, et al? They don't throw an exception if you pass in `null` for the string to be parsed, they just return false; This seems to be analogous, although we have some additional parameters here.

jonfortescue:
@mellinoe @KrzysztofCwalina That is sorta what I was thinking. I will commit the second version now and we can decide if we want to roll back to this version.

KrzysztofCwalina:
I don't feel super strongly about it; you guys should make the final call. Maybe being consistent with int.TryParse makes sense.

jonfortescue:
@mellinoe I'm leaning towards remove the Preconditions and just return false. Let me know what you think and then we can sign off and merge this.

mellinoe:
@jonfortescue I'm fine with that, feel free to merge if there's nothing else to address here.

jonfortescue:
@mellinoe just did the rebase, looks like you need to resign off on it. Sorry about that.

PR #755, opened 2016-08-05
Added prototype float parsing method.

jonfortescue:
This is the prototype of the float parsing method. As discussed, this is a "cheat" method for now; we will double back to enhance perf after completing higher priority tasks.

### DIFF
--- a/src/System.Text.Primitives/System/Text/InvariantParser_float.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_float.cs
@@ -1,0 +1,91 @@
+﻿using System.Diagnostics;
+using System.Text.Utf8;
+
+namespace System.Text
+{
+    public static partial class InvariantParser
+    {
+        public static bool TryParse(byte[] utf8Text, int index, out float value, out int bytesConsumed)
+        {
+            // Precondition replacement
+            if (utf8Text.Length < 1 || index < 0 || index >= utf8Text.Length)
+            {
+                value = 0;
+                bytesConsumed = 0;
+                return false;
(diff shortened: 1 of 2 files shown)


src/System.Text.Primitives/System/Text/InvariantParser_float.cs
```diff
@@ -0,0 +1,91 @@
...
+            if (utf8Text.Length < 1 || index < 0 || index >= utf8Text.Length)
+            {
+                value = 0;
+                bytesConsumed = 0;
+                return false;
+            }
+
+            value = 0f;
+            bytesConsumed = 0;
+            string floatString = "";
```
mellinoe: Not that it will really matter since this is a temporary method, but we should probably use a StringBuilder here.

mellinoe:
I don't think this implementation will catch a lot of valid values, like ones in scientific notation, "infinity" strings, multiple -/+ (in scientific notation), and potentially more. https://msdn.microsoft.com/en-us/library/2thct5cb(v=vs.110).aspx#Anchor_2 has some examples that aren't handled here.

Unfortunately, that will probably complicate the parsing quite a lot. I would start with adding the examples I linked as unit test cases and get them to work.

jonfortescue:
@mellinoe Agreed. This was my v1 and was just made to pass the test cases provided. Will be setting it up to catch everything else soon.

mellinoe:
Okay, sounds good. It looks good for the current cases, and the tests look good as well. Feel free to merge it now and handle the extra cases in a new PR.

PR #756, opened 2016-08-05
Float and Double temporary parsers implemented with unit tests.

jonfortescue:


mellinoe:
I didn't look through the implementation in-depth, but as long as all of the included test cases work, then this looks good to me. The implementation is temporary anyways, so as long as it's functional then I'm okay with merging.

PR #758, opened 2016-08-09
Added boolean support to InvariantParser.

jonfortescue:
Added boolean support to the InvariantParser. These methods are _not_ prototype methods; I started implementing the prototypes when I realized that booleans were simple enough that I could quickly and easily implement the parsers fully.

I also added unit & perf tests for these methods. Here is a graph of perf gains.

8_bool

### DIFF
--- a/src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
@@ -1,0 +1,199 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Text.Utf8;
+
+namespace System.Text
+{
+    public static partial class InvariantParser
+    {
+        #region Helper Methods
+        private static bool isTrue(byte[] utf8Text, int index)
+        {
+            if (utf8Text.Length - index < 4)
+                return false;
+
+            byte firstChar = utf8Text[index];
+            if (firstChar != 't' && firstChar != 'T')
+                return false;
+
+            byte secondChar = utf8Text[index + 1];
+            if (secondChar != 'r' && secondChar != 'R')
+                return false;
+
+            byte thirdChar = utf8Text[index + 2];
+            if (thirdChar != 'u' && thirdChar != 'U')
+                return false;
+
+            byte fourthChar = utf8Text[index + 3];
+            if (fourthChar != 'e' && fourthChar != 'E')
+                return false;
+
+            return true;
+        }
+        private unsafe static bool isTrue(byte* utf8Text, int index, int length)
+        {
+            if (length < 4)
+                return false;
+
+            byte firstChar = utf8Text[index];
+            if (firstChar != 't' && firstChar != 'T')
+                return false;
+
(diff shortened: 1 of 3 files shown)


src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -0,0 +1,213 @@
...
+        {
+            bytesConsumed = 0;
+            value = default(bool);
+
+            if (length < 1 || index < 0)
+            {
+                return false;
+            }
+
+            byte firstByte = utf8Text[index];
```
omariom: `byteAtIndex`  at line 181 does the same check?

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -0,0 +1,213 @@
...
+                return false;
+            }
+        }
+
+        public static unsafe bool TryParse(byte* utf8Text, int index, int length, out bool value, out int bytesConsumed)
+        {
+            bytesConsumed = 0;
+            value = default(bool);
+
+            if (length < 1 || index < 0)
```
omariom: `|| index >= length`?

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
jonfortescue: oh, thank you. I had done some refactoring and apparently ended up with duplicate code I didn't notice.

jonfortescue: fixed this in the latest push.

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -0,0 +1,199 @@
...
+            return true;
+        }
+        #endregion
+
+        public static bool TryParse(byte[] utf8Text, int index, out bool value, out int bytesConsumed)
+        {
+            bytesConsumed = 0;
+            value = default(bool);
+
+            if (utf8Text.Length < 1 || index < 0 || index >= utf8Text.Length)
```
omariom: there is nice trick `(uint)index >= utf8Text.Length`

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -0,0 +1,199 @@
...
+
+            byte secondChar = utf8Text[index + 1];
+            if (secondChar != 'r' && secondChar != 'R')
+                return false;
+
+            byte thirdChar = utf8Text[index + 2];
+            if (thirdChar != 'u' && thirdChar != 'U')
+                return false;
+
+            byte fourthChar = utf8Text[index + 3];
```
omariom: I remember that JIT  removes bounds check when array is accessed in  the opposite order.
Like ```var fourthChar = array[3]; var firstChar = array[0];```. 
Interesting, will it work here?

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
jonfortescue: oh yeah. Nice idea. Will implement.

PR #760, opened 2016-08-10
Reduced a precondition check.

jonfortescue:
Reduced a precondition check per comment https://github.com/dotnet/corefxlab/pull/758/files#r74155156. Perf analysis shows no statistically significant gains or losses, but it is stylistically pleasing.

### DIFF
--- a/src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
@@ -117,7 +117,7 @@ namespace System.Text
             bytesConsumed = 0;
             value = default(bool);
 
-            if (utf8Text.Length < 1 || index < 0 || index >= utf8Text.Length)
+            if (utf8Text.Length < 1 ||  (uint)index >= utf8Text.Length)
             {
                 return false;
             }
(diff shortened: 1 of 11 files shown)


svick:
> but it is stylistically pleasing

I don't understand how is it pleasing, I find `index < 0 || index >= utf8Text.Length` much clearer than `(uint)index >= utf8Text.Length`.

To me, the decrease in readability would be worth it if there was a performance improvement, but you claim there isn't one.

jonfortescue:
@svick 

Yeah, totally fair. I shouldn't say that there was "no" performance increase, just that the statistical tests I did were inconclusive. I'm going to shelve this change until I can do better testing.

joshfree:
Closing this PR for now

omariom:
@jonfortescue 
I made a mistake in the comment. It should be ```(uint)i1 >= (uint)i2```
Result is the same but the code generated is a bit better.
Btw, it is the same trick used for array bounds check.

```C#
private static bool LessThanZeroOrGreater(int i1, int i2)
{
    return (uint)i1 >= i2;
}
```

```ASM
;(uint)i1 >= i2
mov         eax,ecx  
movsxd      rdx,edx  
cmp         rax,rdx  
setge       al  
movzx       eax,al  
ret 

;(uint)i1 >= (uint)i2
cmp         ecx,edx
setae       al  
movzx       eax,al 
ret  
```

PR #761, opened 2016-08-12
Added Uri and Decimal support to InvariantParser.

jonfortescue:


src/System.Text.Primitives/System/Text/InvariantParser_decimal.cs
```diff
@@ -0,0 +1,187 @@
...
+                        else
+                        {
+                            bytesConsumed = 0;
+                            return false;
+                        }
+                    }
+                }
+                else
+                {
+                    if (!digitLast)
```
joshfree: do you actually need digitLast in this method?

tests/System.Text.Primitives.Tests/ParserTests.cs
```diff
@@ -566,5 +611,59 @@ public unsafe void ParseUtf8ByteStarToBool(string text, bool expectSuccess, int
...
+        #region uri
+        [Theory]
+        [InlineData("http://www.example.com/", true, 0, "http://www.example.com/", 23)]
+        [InlineData("https://www.example.org/important.php?true=true", true, 0, "https://www.example.org/important.php?true=true", 47)]
+        [InlineData("http://msw", true, 0, "http://msw", 10)]
+        [InlineData("ftp://example.org", true, 0, "ftp://example.org", 17)]
+        [InlineData("file:///C:\\Users", true, 0, "file:///C:\\Users", 16)]
+        [InlineData("C:\\Users", true, 0, "C:\\Users", 8)]
+        [InlineData("http://&5", false, 0, "", 0)]
+        [InlineData("(*)$#*(", false, 0, "", 0)]
```
joshfree: could you add more negative test cases here?

src/System.Text.Primitives/System/Text/InvariantParser_decimal.cs
```diff
@@ -0,0 +1,187 @@
...
```
joshfree: same comment

tests/System.Text.Primitives.Tests/ParserTests.cs
jonfortescue: indeed. Will do.

src/System.Text.Primitives/System/Text/InvariantParser_decimal.cs
jonfortescue: Ah, no, you're right -- that's an unnecessary carry-over from float/double. Changing now.

PR #762, opened 2016-08-16
Refactored code to T4 templates, added (little-endian) UTF-16 Invariant Support

jonfortescue:
Also hardcoded UTF-16 Invariant support.


joshfree:
@tarekgh could you take a quick look at @jonfortescue 's PR?

joshfree:
cc @mellinoe

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -29,42 +30,105 @@ public static bool TryParse(byte[] utf8Text, int index, out byte value, out int
...
                 }
-                else if (value > Byte.MaxValue / 10)
-                {
-                    value = 0;
-                    bytesConsumed = 0;
-                    return false;
-                }
-                else if (value > 0 && Byte.MaxValue - value * 10 < nextByteVal) // overflow
+            }
+            else if (cultureAndEncodingInfo.IsInvariantUtf16)
```
tarekgh: >else if (cultureAndEncodingInfo.IsInvariantUtf16) [](start = 12, length = 49)

I assume the data in the byte array is encoded as UTF16 and not UTF8. if this true, we can implement the parsing in simpler way by using unsafe char* pointing to the byte array and consume the buffer as UTF16 char at  time. maybe I am missing something here so I'll try to talk to you offline.

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -122,39 +123,46 @@ public static bool TryParse(byte[] utf8Text, int index, out bool value, out int
...
-                return true;
-            }
-            else if (isFalse(utf8Text, index))
-            {
-                bytesConsumed = 5;
-                value = false;
-                return true;
-            }
-            else
+            if (cultureAndEncodingInfo.IsInvariantUtf8)
```
tarekgh: >IsInvariantUtf8 [](start = 39, length = 15)

are we going to support invariant UTF16 with the bool type too?

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -122,39 +123,46 @@ public static bool TryParse(byte[] utf8Text, int index, out bool value, out int
...
+                    value = true;
+                    return true;
+                }
+                else if (firstByte == '0')
+                {
+                    bytesConsumed = 1;
+                    value = false;
+                    return true;
+                }
+                else if (isTrue(utf8Text, index))
```
mellinoe: I know it's not part of the review, but these method names should be capitalized.

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -77,40 +141,109 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out by
...
+                            return false;
+                        }
+                        else
+                        {
+                            return true; // otherwise return true
+                        }
+                    }
+                    else
+                    {
+                        byte nextByteVal = (byte)(nextByte - '0');
```
tarekgh: >byte nextByteVal = (byte)(nextByte - '0'); [](start = 24, length = 42)

I believe there is something wrong in the logic when encountering characters less than the ordinal value of '0' something like 2F

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
mellinoe: I was going to make the same comment. Unless I'm misunderstanding something, we should be able to loop through 16-bit chunks instead of looping over individual bytes, and it would probably eliminate some of this confusing logic.

tarekgh:
in general I would suggest to refactor the code to remove all duplicated code and create helper methods that centralize the logic we are doing. also we need to write the UTF16 parsing work on 2 bytes at a time for code clarity and performance too. thanks @jonfortescue for your work.

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
jonfortescue: Yup, agreed. Fixing this in the newest commit.

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
jonfortescue: Yes, but in a future PR.

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
jonfortescue: Good catch, will fix.

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -77,40 +119,87 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out by
...
                 {
-                    value = 0;
-                    bytesConsumed = 0;
-                    return false;
-                }
-
-                byte candidate = (byte)(value * 10 + nextByteVal); // left shift the value and add the nextByte.
+                    byte nextByte = utf8Text[byteIndex];
+                    byte byteAfterNext = utf8Text[byteIndex + 1];
+                    byte nextByteVal = (byte)(nextByte - '0');
```
tarekgh: >byte nextByteVal = (byte)(nextByte - '0'); [](start = 20, length = 42)

nit:
you can write it as 

byte nextByteVal = (byte)(utf8Text[byteIndex]- '0');

and completely get rid of the line 

byte nextByte = utf8Text[byteIndex];

jonfortescue: Ah, no you're totally right -- nextByte was only there for when I was doing the check to see if it was = 0. Will fix right away. :+1:

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -77,40 +119,87 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out by
...
+                    if (nextByteVal > 9 || byteAfterNext != 0) // if the second byte isn't zero, this isn't an ASCII-equivalent code unit and we can quit here
```
tarekgh: >if (nextByteVal > 9  [](start = 20, length = 20)

could you please put comment that nextByteVal will be greater than 9 if (nextByte - '0') is negative value

jonfortescue: Yes! Will do.

jonfortescue: Actually @tarekgh -- I'm refactoring all this code right now in a separate branch. I will make these two changes in that branch as it will just be less time consuming overall.

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -77,40 +119,87 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out by
...
+                    {
+                        if (bytesConsumed == 0) // check to see if we've processed any digits at all
+                        {
+                            value = default(byte); // if we haven't, set value to 0 and return false
```
tarekgh: >value = default(byte); // if we haven't, set value to 0 and return false [](start = 28, length = 72)

you are really don't need this line. if there is no bytesConsumed then you didn't touch value which is already initialized to 0 in the beginning of the method

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
tarekgh: can we have just one PR have the whole change? it will be easier than doing it twice.

---
In reply to: [75390658](https://github.com/dotnet/corefxlab/pull/762#discussion_r75390658) [](ancestors = 75390658)

jonfortescue: Sure.

tarekgh: thank you.

---
In reply to: [75391092](https://github.com/dotnet/corefxlab/pull/762#discussion_r75391092) [](ancestors = 75391092)

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -77,40 +119,87 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out by
...
+                            return false;
+                        }
+                        else
+                        {
+                            return true; // otherwise return true
+                        }
+                    }
+                    else if (value > Byte.MaxValue / 10)
```
tarekgh: this condition and the next condition can be combined together for the overflow check. 

another idea you may use is do the following:

int newValue = value * 10 + nextByteVal;

if (newValue > Byte.MaxValue) // overflow
{
}

value = (byte) newValue;

this will make the code a lot cleaner and faster. but I'll leave it to you to decide about it :-)

src/System.Text.Primitives/System/Text/InvariantParser_byte.cs
```diff
@@ -77,40 +119,87 @@ unsafe public static bool TryParse(byte* utf8Text, int index, int length, out by
...
+                    else if (value > 0 && Byte.MaxValue - value * 10 < nextByteVal) // overflow
+                    {
+                        value = 0;
+                        bytesConsumed = 0;
+                        return false;
+                    }
 
-                value = candidate;
-                bytesConsumed++;
+                    byte candidate = (byte)(value * 10 + nextByteVal); // left shift the value and add the nextByte.
```
tarekgh: >byte candidate [](start = 20, length = 14)

you don't need to define a new variable here. you can just assign it to value

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
```diff
@@ -31,68 +32,139 @@ public static bool TryParse(byte[] utf8Text, int index, out int value, out int b
...
+
+                if (negative) // We check if the value is negative at the very end to save on comp time
+                {
+                    value = -value;
+                }
+                return true;
+            }
+            else if (cultureAndEncodingInfo.IsInvariantUtf16)
+            {
+                if (utf8Text[index] == '-')
```
tarekgh: you need to check the second byte to be zero.

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
```diff
@@ -31,68 +32,139 @@ public static bool TryParse(byte[] utf8Text, int index, out int value, out int b
...
+                        }
+                        if (bytesConsumed == 0) // check to see if we've processed any digits at all
+                        {
+                            value = default(int); // if we haven't, set value to 0 and return false
+                            return false;
+                        }
+                        else
                         {
-                            value = -value;
+                            if (negative && value != Int32.MinValue) // We check if the value is negative at the very end to save on comp time
```
tarekgh: >value != Int32.MinValue [](start = 44, length = 23)

is it possible the value can be Int32.MaxValue when negative is true? I think this case should be covered in the overflow check. right?

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
```diff
@@ -31,68 +32,139 @@ public static bool TryParse(byte[] utf8Text, int index, out int value, out int b
...
                         }
-                        return true; // otherwise return true
                     }
+                    else if (value > Int32.MaxValue / 10) // overflow
+                    {
+                        value = 0;
+                        bytesConsumed = 0;
+                        return false;
+                    }
+                    else if (value > 0 && (Int32.MaxValue - value * 10) + (negative ? 1 : 0) < nextByteVal) // overflow (+1 allows for overflow for min value)
```
tarekgh: please look at the code 
[http://index/#mscorlib/system/number.cs,502](http://index/#mscorlib/system/number.cs,502)

which doing the overflow check in simpler way especially when you have negative sign.

tarekgh: by the way, the file number.cs has the parsing for all number types which can give you the idea how to do the same.

---
In reply to: [75397007](https://github.com/dotnet/corefxlab/pull/762#discussion_r75397007) [](ancestors = 75397007)

src/System.Text.Primitives/System/Text/InvariantParser_int.cs
jonfortescue: MinValue, not MaxValue. We allow overflow to occur in the case of the MinValue since abs(MinValue) > MaxValue.

So, if `negative` is true and `value == [class].MinValue`, then we know that the value is already negative, and thus we shouldn't swap the sign (although swapping the sign will still result in the MinValue).

tarekgh: you are right. we can simplify the implementation by just parsing the number without checking at all for the sign and after we finish parsing we can set the sign and check. this is like the implementation in http://index/#mscorlib/system/number.cs,552

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+                {
+                    value = (<#= primitives[i] #>)-value;
+                }
+                return true;
+            }
+            else if (cultureAndEncodingInfo.IsInvariantUtf16)
+            {
+                if (utf8Text[index] == '-')
+                {
+                    negative = true;
```
tarekgh: you need to check for the second byte. it has to be zero.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+															   // if nextByteVal > 9, we know it is not a digit because any value less than '0' will overflow
+															   // to greater than 9 since byte is an unsigned type.
+                    {
+                        if (bytesConsumed == 2 && signed) // if the first character happened to be a '-' or a '+', we reset the byte counter so logic proceeds as normal
+                        {
+                            bytesConsumed = 0;
+                        }
+                        if (bytesConsumed == 0) // check to see if we've processed any digits at all
+                        {
+                            value = default(int); // if we haven't, set value to 0 and return false
```
tarekgh: >value = default(int); [](start = 28, length = 21)

this already initialized to zero in the beginning of  method. you don't need to reset i there.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+                            return false;
+                        }
+                        else
+                        {
+                            if (negative && value != <#= classes[i] #>.MinValue) // We check if the value is negative at the very end to save on comp time
```
tarekgh: maybe you can do the same trick done in the file [http://index/#mscorlib/system/number.cs,518](http://index/#mscorlib/system/number.cs,518)

            if (number.sign) {
                n = -n;
                if (n > 0) {
                    return false;
                }
            }
so you don't have to check for value != <#= classes[i] #>.MinValue

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+                            return true; // otherwise return true
+                        }
+                    }
+                    else if (value > <#= classes[i] #>.MaxValue / 10) // overflow
+                    {
+                        value = 0;
+                        bytesConsumed = 0;
+                        return false;
+                    }
+                    else if (value > 0 && (<#= classes[i] #>.MaxValue - value * 10) + (negative ? 1 : 0) < nextByteVal) // overflow (+1 allows for overflow for min value)
```
tarekgh: I believe you don't really need to have this whole check. look at [http://index/#mscorlib/system/number.cs,510](http://index/#mscorlib/system/number.cs,510). the check you have (value > <#= classes[i] #>.MaxValue / 10) should be enough in your case.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+                    {
+                        value = 0;
+                        bytesConsumed = 0;
+                        return false;
+                    }
+                    <#= primitives[i] #> candidate = (<#= primitives[i] #>)(value * 10 + nextByteVal); // parse the current digit to a <#= primitives[i] #> and add it to the left-shifted value
```
tarekgh: >candidate [](start = 41, length = 9)

you don't need to defile a new variable here. just assign the RHS to value.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+                    value = candidate;
+                    bytesConsumed += 2; // increment the number of bytes consumed, then loop
+                }
+
+                if (negative && value != <#= classes[i] #>.MinValue) // We check if the value is negative at the very end to save on comp time
```
tarekgh: same as previous comment. you can do it as we did it in the number.cs file I pointed at before.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+                    index += 2;
+                    bytesConsumed += 2;
+                }
+
+                for (int byteIndex = index; byteIndex < length + index - 1; byteIndex += 2) // loop through the byte array two bytes at a time for UTF-16
+                {
+                    byte byteAfterNext = utf8Text[byteIndex + 1];
+                    byte nextByteVal = (byte)(utf8Text[byteIndex] - '0');
+                    if (nextByteVal > 9 || byteAfterNext != 0) // if the second byte isn't zero, this isn't an ASCII-equivalent code unit and we can quit here
+															   // if nextByteVal > 9, we know it is not a digit because any value less than '0' will overflow
```
tarekgh: thank you for adding the comment.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+                }
+                if (negative && value != <#= classes[i] #>.MinValue) // We check if the value is negative at the very end to save on comp time
+                {
+                    value = (<#= primitives[i] #>)-value;
+                }
+                return true;
+            }
+            else if (cultureAndEncodingInfo.IsInvariantUtf16)
+            {
+                if (utf8Text[index] == '-')
```
tarekgh: previous comments I added will apply to this method too.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
+﻿<#@ template hostspecific="false" language="C#" #>
+<#@ output extension=".cs" #>
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
tarekgh: I assume this template file is used to generate the cs files. could you please add a comment here telling don't modify this file manually and instead modify the template file .tt

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
```diff
@@ -0,0 +1,321 @@
...
+
+using System.Diagnostics;
```
tarekgh: do we manually generate the cs files from .tt or this will be done automatically through the build? may be adding some comment for how we generate the cs files from the tt files would be great.

tarekgh:
@jonfortescue the current changes looks a lot better and cleaner. thanks for doing that.I just added some more minor comments.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
mellinoe: We will need to manually generate the .cs files and keep them updated when changes to the .tt file are made. I'm not sure it's easy or possible to make that kind of thing happen automatically with xproj's.  For the .tt files in corefx I have just used VS to regenerate the .cs files (they are generated when the .tt file is saved).

tests/System.Text.Primitives.Tests/ParserTests.cs
```diff
@@ -4,472 +4,983 @@
 
 namespace System.Text.Primitives.Tests
 {
-    public class ParserTests
+	public class ParserTests
     {
-        private byte[] UtfEncode(string s)
+		private byte[] UtfEncode(string s)
```
mellinoe: I think you should just either have both `Uft8Encode` and a `Uft16Encode` methods, or just a single `UtfEncode` method with a bool switch. I'd prefer the former (two explicitly-named methods).

tests/System.Text.Primitives.Tests/ParserTests.cs
```diff
@@ -4,472 +4,983 @@
...
+            if (utf16)
+                return Encoding.Unicode.GetBytes(s);
+            else
+                return UtfEncode(s);
+        }
+
+		#region byte
+
+		[Theory]
+        [InlineData("55", true, 0, 55, 2)]
```
mellinoe: If all of this `InlineData` is going to be the same for multiple methods, I would recommend putting it into a member function `X` and then using [MemberData(nameof(X))] instead of duplicating it into each method's attributes. It looks like you can share it for all 4 Utf8/Utf16 array/pointer versions of these tests.

mellinoe: I see that the last parameter will differ for Utf8/Utf16 cases (by a factor of two), but I wonder if you couldn't somehow still re-use the test data for both cases.

In the end it doesn't matter a whole lot because this is a generated file, but it would help reading it a bit :smile: .

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
jonfortescue: I will make this change for now, but I'm opening an issue to investigate the perf of this change.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
jonfortescue: Actually, I think that's a bug. Or at least, our code is doing the equivalent of my check somewhere else. If you don't have this check in there, it will fail on values like "256" for byte or "32679" for short -- essentially, values where the current value is <= MaxValue / 10, but the upcoming digit will cause the overflow.

jonfortescue: However, on review I figured out that I can reduce this by using by checking for >=8 in signed cases and >=6 in unsigned cases.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
jonfortescue: Will do.

tests/System.Text.Primitives.Tests/ParserTests.cs
jonfortescue: Yeah, I might do a second pass at refactoring the code later on. For now, though, having the template files is enough, I think.

src/System.Text.Primitives/System/Text/InvariantParser_signed.tt
tarekgh: it is not a bug in the framework code number.cs. in the code we still check for the overflow differently 

[http://index/#mscorlib/system/number.cs,612](http://index/#mscorlib/system/number.cs,612)

                    UInt64 newN = n + (UInt64)(*p++ - '0');
                    // Detect an overflow here...
                    if (newN < n) {
                        return false;
                    }


---
In reply to: [75535728](https://github.com/dotnet/corefxlab/pull/762#discussion_r75535728) [](ancestors = 75535728)

tarekgh: I see we'll have to check for overflow anyway. I prefer if you can do the check similar to what I pasted before 

                    UInt64 newN = n + (UInt64)(*p++ - '0');
                    // Detect an overflow here...
                    if (newN < n) {
                        return false;
                    }

the good thing here is catching the overflow and not really care specifying the limits values (8 or 6). note that we parse the negative numbers as positives and we apply the sign at the end of the parsing. 

---
In reply to: [75536675](https://github.com/dotnet/corefxlab/pull/762#discussion_r75536675) [](ancestors = 75536675)

jonfortescue: Ah yeah, I see where they do the check. I'm going to add this to my perf investigations for a future commit.

PR #768, opened 2016-08-21
Modified the FormattingData class to allow for parsing. Full support for cultural specific parsing.

jonfortescue:
This is a big PR, so apologies in advance.

Overview:
* Added a new data structure to `FormattingData`: a parsing [trie](https://en.wikipedia.org/wiki/Trie) (or prefix tree). This structure serves as the inverse of the formatting map that is provided to `FormattingData`.
* As of this commit, this structure is allocated when the `FormattingData` object is created, which means that there is an expensive upfront cost. However, since the trie is navigated using binary search, lookup is extremely cheap, which means we trade off that upfront cost for very performant parsing.
    * This means it is optimized for the webserver scenario (initialize my encoding, then parse a bunch of requests)
    * However, there are obviously other scenarios that would benefit from slower parsing in favor of not having any (or as much of) an upfront cost. These optimizations will come in future pull requests. (Currently, some basic flags exist for a future "lazy construction" option.)
* In order to parse, the `FormattingData` structure acts as a state machine. You pass it bytes, and it tells you to continue passing it bytes until it successfully parses a code unit, at which point it returns the value of that code unit.
* The structure has the potential to allow special parsing rules in the future for encodings that require them. However, as of now the parsing structure has a few basic assumptions:
    * Any culture/encoding is based on a decimal 0-9 digit system
    * An culture/encoding can be basically understood as a sequence of bytes read from left to right
    * Code units in every culture/encoding are unambiguous
* If any of these assumptions are violated by a culture/encoding, special casing and a state flag would need to be introduced
* I've added tests for an Invariant UTF-8 lookalike and a simple Thai-based UTF-8 encoding. All tests are passing, which shows just how versatile this structure is.
* No perf testing has been done yet. Will investigate perf on Monday.

### DIFF
--- a/src/System.Text.Primitives/System/Text/FormattingData.cs
+++ b/src/System.Text.Primitives/System/Text/FormattingData.cs
@@ -12,7 +12,8 @@ namespace System.Text
     {
         private static FormattingData s_invariantUtf16;
         private static FormattingData s_invariantUtf8;
-        private byte[][] _digitsAndSymbols; // this could be flatten into a single array
+        private byte[][] _digitsAndSymbols; // this could be flattened into a single array
+        private TrieNode[] _parsingTrie;
         private Encoding _encoding;
 
         public enum Encoding
@@ -21,10 +22,84 @@ namespace System.Text
             Utf8 = 1,
         }
 
+        // The parsing trie is structured as an array, which means that there are two types of
+        // "nodes" for representational purposes
+        //
+        // The first node type (the parent node) uses the valueOrNumChildren to represent the number of children
+        // underneath it. The index is unused for this type of node, except when it's used for
+        // sequential node mapping (see below). If valueOrNumChildren is zero for this type of node, the index
+        // is used and represents an index into _digitsAndSymbols.
+        //
+        // The second node types immediately follow the first (the childe nodes). They are composed of a value
+        // (valueOrNumChildren), which is walked via binary search, and an index, which points to another
+        // node contained in the array.
+        //
+        // We use the int index here to encode max-min info for sequential leaves
+        // It's very common for digits to be encoded sequentially, so we save time by mapping here
+        // The index is formatted as such: 0xAABBCCDD, where AA = the min value,
+        // BB = the index of the min value relative to the current node (1-indexed),
+        // CC = the max value, and DD = the max value's index in the same coord-system as BB.
+        public struct TrieNode
+        {
+            public byte valueOrNumChildren;
+            public int index;
+        }
+
+        // TODO: make these private once bin file generator is used
+        public FormattingData(byte[][] digitsAndSymbols, TrieNode[] parsingTrie, Encoding encoding)
+        {
+            _digitsAndSymbols = digitsAndSymbols;
+            _encoding = encoding;
+            _parsingTrie = parsingTrie;
+        }
+
         public FormattingData(byte[][] digitsAndSymbols, Encoding encoding)
         {
             _digitsAndSymbols = digitsAndSymbols;
             _encoding = encoding;
+            _parsingTrie = null;
+        }
+
+        // This binary search implementation returns an int representing either:
+        //      - the index of the item searched for (if the value is positive)
+        //      - the index of the location where the item should be placed to maintain a sorted list (if the value is negative)
+        private int BinarySearch(int nodeIndex, int level, byte value)
+        {
+            int maxMinLimits = _parsingTrie[nodeIndex].index;
+            if (maxMinLimits > 0 && value > maxMinLimits >> 24 && value < (maxMinLimits << 16) >> 24)
+            {
+                // See the comments on the struct above for more information about this format
+                return nodeIndex + ((maxMinLimits << 8) >> 24) + value - (maxMinLimits >> 24);
+            }
+
+            int leftBound = nodeIndex + 1, rightBound = nodeIndex + _parsingTrie[nodeIndex].valueOrNumChildren;
+            int midIndex = 0;
+            while (true)
+            {
+                if (leftBound > rightBound)  // if the search failed
+                {
+                    // this loop is necessary because binary search takes the floor
+                    // of the middle, which means it can give incorrect indices for insertion.
+                    // we should never iterate up more than two indices.
+                    while (midIndex < nodeIndex + _parsingTrie[nodeIndex].valueOrNumChildren
+                        && _parsingTrie[midIndex].valueOrNumChildren < value)
+                    {
+                        midIndex++;
+                    }
+                    return -midIndex;
+                }
+
+                midIndex = (leftBound + rightBound) / 2; // find the middle value
+
+                byte mValue = _parsingTrie[midIndex].valueOrNumChildren;
+
+                if (mValue < value)
+                    leftBound = midIndex + 1;
+                else if (mValue > value)
+                    rightBound = midIndex - 1;
+                else
+                    return midIndex;
+            }
         }
 
         // it might be worth compacting the data into a single byte array.
@@ -49,6 +124,7 @@ namespace System.Text
                 new byte[] { 43, 0, }, // plus sign 
                 new byte[] { 78, 0, 97, 0, 78, 0, }, // NaN
                 new byte[] { 69, 0, }, // E
+                new byte[] { 101, 0, }, // e
             };
 
             s_invariantUtf16 = new FormattingData(utf16digitsAndSymbols, Encoding.Utf16);
@@ -71,6 +147,7 @@ namespace System.Text
                 new byte[] { 43, }, // plus sign
                 new byte[] { 78, 97, 78, }, // NaN
                 new byte[] { 69, }, // E
+                new byte[] { 101, }, // e
             };
 
             s_invariantUtf8 = new FormattingData(utf8digitsAndSymbols, Encoding.Utf8);
@@ -91,6 +168,66 @@ namespace System.Text
             {
                 return s_invariantUtf8;
             }
+        }
+
+        /// <summary>
+        /// Parse the next byte in a byte array. Will return either a DigitOrSymbol value, an InvalidCharacter, or a Continue
+        /// </summary>
+        /// <param name="nextByte">The next byte to be parsed</param>
+        /// <param name="bytesParsed">The total number of bytes parsed (will be zero until a code unit is deciphered)</param>
+        /// <returns></returns>
+        public bool TryParseNextCodingUnit(ref byte[] buffer, ref int bufferIndex, out uint value)
+        {
+            int trieIndex = 0;
+            int codeUnitIndex = 0;
+
+            while (true)
+            {
+                if (_parsingTrie[trieIndex].valueOrNumChildren == 0)    // if numChildren == 0, we're on a leaf & we've found our value and completed the code unit
+                {
+                    if (VerifyCodeUnit(ref buffer, bufferIndex, codeUnitIndex, _parsingTrie[trieIndex].index))
+                    {
+                        bufferIndex += _digitsAndSymbols[_parsingTrie[trieIndex].index].Length - codeUnitIndex;
+                        value = (uint)_parsingTrie[trieIndex].index;  // return the parsed value
+                        return true;
+                    }
+                    else
+                    {
+                        value = 0;
+                        return false;
+                    }
+                }
+                else
+                {
+                    int search = BinarySearch(trieIndex, codeUnitIndex, buffer[bufferIndex]);    // we search the _parsingTrie for the nextByte
+
+                    if (search > 0)   // if we found a node
+                    {
+                        trieIndex = _parsingTrie[search].index;
+                        bufferIndex++;
+                        codeUnitIndex++;
+                    }
+                    else
+                    {
+                        value = 0;
+                        return false;
+                    }
+                }
+            }
+        }
+        private bool VerifyCodeUnit(ref byte[] buffer, int index, int codeUnitIndex, int digitOrSymbol)
+        {
+            int codeUnitLength = _digitsAndSymbols[digitOrSymbol].Length;
+            if (codeUnitIndex == codeUnitLength - 1)
+                return true;
+            
+            for (int i = 0; i < codeUnitLength - codeUnitIndex; i++)
+            {
+                if (buffer[i + index] != _digitsAndSymbols[digitOrSymbol][i + codeUnitIndex])
+                    return false;
+            }
+
+            return true;
         }
 
         public bool TryWriteDigit(ulong digit, Span<byte> buffer, out int bytesWritten)
--- a/src/System.Text.Primitives/System/Text/InvariantParser_signed.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_signed.cs
@@ -180,8 +180,89 @@ namespace System.Text
 				}
                 return true;
             }
+			else
+			{
+				int byteIndex = index;
+				int codeUnitsConsumed = 0;
+                while (byteIndex < utf8Text.Length)
+                {
+                    uint result;
+					int oldIndex = byteIndex;
+                    bool success = cultureAndEncodingInfo.TryParseNextCodingUnit(ref utf8Text, ref byteIndex, out result);
 
-            return false;
+                    if (!success || result > 9)
+                    {
+						if (bytesConsumed == 0 && result == (int)FormattingData.Symbol.MinusSign)
+						{
+							negative = true;
+							signed = true;
+                            bytesConsumed += byteIndex - oldIndex;
+							codeUnitsConsumed++;
+						}
+						else if (bytesConsumed == 0 && result == (int)FormattingData.Symbol.PlusSign)
+						{
+							negative = true;
+							signed = true;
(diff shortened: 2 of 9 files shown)


tests/System.Text.Primitives.Tests/ParserTests.tt
```diff
@@ -174,6 +199,68 @@ namespace System.Text.Primitives.Tests
...
+            Format.Parsed nf = new Format.Parsed('G');
+            bool result = InvariantParser.TryParse(UtfEncode(text, false), index, fd, nf, out parsedValue, out bytesConsumed);
+
+            Assert.Equal(expectSuccess, result);
+            Assert.Equal(expectedValue, parsedValue);
+            Assert.Equal(expectedBytesConsumed, bytesConsumed);
+        }
+
+		[Theory]
+        [InlineData("๑๑๑", true, 0, 111, 9)]
```
jonfortescue: In a future PR, these values will be generated most likely so that they can vary from type to type. Currently, they're all hardcoded just for the sake of being able to test something.

jonfortescue:
Just realized there is a significant bug, but it's the middle of the night so I won't fix it now. The parsers need to check to make sure the rest of the byte sequence is accurate once they identify a byte successfully

jonfortescue:
In the newest update:

* I have made FormattingData immutable once again, which means that the state of the state machine is held by the thread doing the parsing. This will make it so multithreaded parsing operations can be performed on the same FormattingData object.
* I added a `VerifyCodingUnit` which is designed to be run immediately after successfully parsing a coding unit to verify that the tree did not return an erroneous result. This fixes the bug I mentioned here (https://github.com/dotnet/corefxlab/pull/768#issuecomment-241358556)

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,309 @@ public struct FormattingData
...
         public enum Encoding
         {
             Utf16 = 0,
             Utf8 = 1,
         }
+        // used to determine if/how we should construct the parsing trie
+        // eventually, this could include all sorts of information, like encoding rules, etc.
+        public enum ParsingStates
+        {
+            NO_PARSING = 0, // don't construct it
```
KrzysztofCwalina: These names don't follow our casing conventions. You should use PascalCasing.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,309 @@ public struct FormattingData
...
+            PARSING = 1,    // construct it with upfront cost
+            LAZY_PARSING = 3 // construct it but allocate lazy. 3 is chosen so the bitwise check below work
```
KrzysztofCwalina: As we discussed, we should load the tree from a bin file into a flat array. At which point I don't think we will need to make it lazy.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,309 @@ public struct FormattingData
...
+        /// <param name="decimalSeparator">The bytes which map to the decimal separtor.</param>
+        /// <param name="groupSeparator">The bytes which map to the group separator</param>
+        /// <param name="infinity">The bytes which map to the representation of infinity</param>
+        /// <param name="minusSign">The bytes which map to the minus sign</param>
+        /// <param name="plusSign">The bytes which map to the plus sign</param>
+        /// <param name="nan">The bytes which map to the representation of NaN (not a number)</param>
+        /// <param name="exponent">The bytes which map to the exponent marker for floating-point types</param>
+        /// <param name="encoding">The specified encoding</param>
+        /// <param name="parsing"></param>
+        public FormattingData(byte[] digit0, byte[] digit1, byte[] digit2, byte[] digit3, byte[] digit4, byte[] digit5, byte[] digit6, byte[] digit7,
```
KrzysztofCwalina: Why are you adding this constructor? Take a look into System.Formatting.Globalization for samples (tests) showing how to construct non-invariant cultures.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,309 @@ public struct FormattingData
...
+            byte[] digit8, byte[] digit9, byte[] decimalSeparator, byte[] groupSeparator, byte[] infinity, byte[] minusSign, byte[] plusSign, byte[] nan,
+            byte[] exponent, byte[] exponentSecondary, Encoding encoding, byte parsing = (byte)ParsingStates.NO_PARSING) 
+            : this(new byte[][] { digit0, digit1, digit2, digit3, digit4, digit5, digit6, digit7, digit8, digit9,
+                decimalSeparator, groupSeparator, infinity, minusSign, plusSign, nan, exponent, exponentSecondary }, encoding, parsing)
+        {
+            // This constructor simple calls the base constructor
+        }
+        private class TrieNode
```
KrzysztofCwalina: As we discussed. Let's store the tree in an array of structs.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,309 @@ public struct FormattingData
...
+        {
+            public byte? ByteValue { get; set; } // nullable byte is used so that the null-value can be used           (0x02)
```
KrzysztofCwalina: Sorry, I don't understand this diagram. What's 0x03? Why is there a tuple in the right hand branch?If one of the byte sequences if simply 0x01 (as you say), why is the root of the tree 0x02?

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,309 @@ public struct FormattingData
...
+            // If the node is a leaf, we know we have reached the bottom of the trie.
+            if (node.IsLeaf)
+            {
+                node.IsLeaf = false; // this is now a branch, not a leaf
+                byte? oldByteValue = DisambiguateByteValue(oldIndex, level);
+                if (newByteValue == oldByteValue)   // this means there's another collision, so we have to go down one level more
+                {
+                    if (oldByteValue == null)   // if both values are null, we have identical code units, which is bad
+                    {
+                        throw new ArgumentException("Invalid mapping: two values map to the same code unit.");
```
KrzysztofCwalina: I don't think a developer callign this API will be able to figure out from the message what's going on. Could the message say something about two digits and/or symbols being encoded with the same sequences of bytes?

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,309 @@ public struct FormattingData
...
+            else
+                return _childTrieNodes[node.ChildIndices[m]].ByteValue;
+        }
+        // This binary search implementation returns a tuple:
+        // * a bool which represents whether the search completed successfully
+        // * an int representing either:
+        //      - the index of the item searched for
+        //      - the index of the location where the item should be placed to maintain a sorted list
+        // Because of the second function, a tuple is necessary since we cannot return -1 to represent a failed search 
+        private Tuple<bool, int> BinarySearch(TrieNode node, int level, byte? value)
```
KrzysztofCwalina: Tuple is a class. It causes an allocation. You should return a custom struct with two fields.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -93,6 +385,75 @@ public static FormattingData InvariantUtf8
             }
         }
 
+        /// <summary>
+        /// Parse the next byte in a byte array. Will return either a DigitOrSymbol value, an InvalidCharacter, or a Continue
```
KrzysztofCwalina: You should add an enum (backed by a byte) for this instead of returning a byte.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -49,6 +339,7 @@ static FormattingData()
                 new byte[] { 43, 0, }, // plus sign 
                 new byte[] { 78, 0, 97, 0, 78, 0, }, // NaN
                 new byte[] { 69, 0, }, // E
+                new byte[] { 101, 0, }, // e
```
KrzysztofCwalina: If you are adding a new symbol, you need to also change System.Text.Globalization, I think.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -93,6 +385,75 @@ public static FormattingData InvariantUtf8
             }
         }
 
+        /// <summary>
+        /// Parse the next byte in a byte array. Will return either a DigitOrSymbol value, an InvalidCharacter, or a Continue
+        /// </summary>
+        /// <param name="nextByte">The next byte to be parsed</param>
+        /// <param name="bytesParsed">The total number of bytes parsed (will be zero until a code unit is deciphered)</param>
+        /// <returns></returns>
+        public byte ParseNextByte(byte nextByte, out int bytesParsed, ref int nodeIndex, ref int level)
```
KrzysztofCwalina: Why not have an overload (or just one method) that takes a buffer, instead of having state machine APIs exposed publicly?

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -142,6 +503,9 @@ public enum Symbol : ushort
             PlusSign = 14,          
             NaN = 15,
             Exponent = 16,
+            ExponentSecondary = 17,
```
KrzysztofCwalina: Can you add a comment describing what this is? I think he other names come from the Unicode standard. I don't remember seeing "secondary exponent" there, so people might not be able to understand what this is.

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -142,6 +503,9 @@ public enum Symbol : ushort
...
+            Invalid = 18, // invalid character for parsing
```
KrzysztofCwalina: Can we avoid adding these two here? They are not symbols. It makes the public API confusing. Especially for formatting where symbol values are passed in (and passing in Invalid would be invalid).

src/System.Text.Primitives/System/Text/InvariantParser_signed.cs
```diff
@@ -180,8 +180,101 @@ public static partial class InvariantParser
 				}
                 return true;
             }
+			else
```
KrzysztofCwalina: Could you add perf tests comparing the two code paths (invariant and not).

KrzysztofCwalina:
>         public void ParseUtf8ByteArrayToDecimal(string text, bool expectSuccess, int index, decimal expectedValue, int expectedBytesConsumed)

Would it make sense to have the non-generated tests in a separate cs file instead of using the template?

---
Refers to: tests/System.Text.Primitives.Tests/ParserTests.tt:496 in d647eb0. [](commit_id = d647eb06cc254196982ff46936ee54dd68d7a212, deletion_comment = False)

KrzysztofCwalina:
:clock1:

src/System.Text.Primitives/System/Text/FormattingData.cs
jonfortescue: I was thinking that I'd need a separate symbol for when there was more than one way to represent a symbol ('e', 'E'), but now I just realized that was dumb and they should both point to 16. Changing this haha.

src/System.Text.Primitives/System/Text/InvariantParser_signed.cs
jonfortescue: yup yup

src/System.Text.Primitives/System/Text/FormattingData.cs
jonfortescue: Will do in next push.

src/System.Text.Primitives/System/Text/FormattingData.cs
jonfortescue: Yes. Will be there when we move this to the tool.

src/System.Text.Primitives/System/Text/FormattingData.cs
jonfortescue: oh, good catch. The diagram is wrong (I compacted the tree in the example).

The result is not a tuple, but an array sequence of bytes.

src/System.Text.Primitives/System/Text/FormattingData.cs
jonfortescue: Was doing it so I could hardcode for values inline. Removed, though.

src/System.Text.Primitives/System/Text/FormattingData.cs
jonfortescue: I think that we may end up needing such a parameter eventually for encodings with special rules. Not positive, but it's definitely something we might need to examine.

src/System.Text.Primitives/System/Text/FormattingData.cs
jonfortescue: removing anyway

jonfortescue:
Re: non-generated tests in a separate file -- that's actually a really good idea. I will probably implement that in the next PR.

src/System.Text.Primitives/System/Text/FormattingData.cs
KrzysztofCwalina: If we eventually needed, we will add it (or something like it).

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,210 @@ public struct FormattingData
...
             Utf16 = 0,
             Utf8 = 1,
         }
+        public struct TrieNode
+        {
+            public byte valueOrNumChildren;
+            public int index;
+        }
+
+        public FormattingData(byte[][] digitsAndSymbols, TrieNode[] parsingTrie, Encoding encoding)
```
KrzysztofCwalina: >TrieNode [](start = 57, length = 8)

Can we make it not public? the ctor and TrieNode type

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -12,19 +12,210 @@ public struct FormattingData
...
 
         public FormattingData(byte[][] digitsAndSymbols, Encoding encoding)
         {
             _digitsAndSymbols = digitsAndSymbols;
             _encoding = encoding;
+            _parsingTrie = null;
+        }
+
+        // Leaving this code here for now so we can move it to a tool to construct the parsing trie
+        //
```
KrzysztofCwalina: Can we remove the comment?

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -93,6 +286,59 @@ public static FormattingData InvariantUtf8
             }
         }
 
+        /// <summary>
+        /// Parse the next byte in a byte array. Will return either a DigitOrSymbol value, an InvalidCharacter, or a Continue
+        /// </summary>
+        /// <param name="nextByte">The next byte to be parsed</param>
+        /// <param name="bytesParsed">The total number of bytes parsed (will be zero until a code unit is deciphered)</param>
+        /// <returns></returns>
+        public uint ParseNextCodingUnit(ref byte[] buffer, ref int bufferIndex)
```
KrzysztofCwalina: >ParseNextCodingUnit [](start = 20, length = 19)

Please add a TODO to convert it to TryParse(byte[], int, out CodePoint, out int)

src/System.Text.Primitives/System/Text/FormattingData.cs
```diff
@@ -93,6 +286,59 @@ public static FormattingData InvariantUtf8
...
+                        trieIndex = _parsingTrie[search].index;
+                        bufferIndex++;
+                        codeUnitIndex++;
+                    }
+                    else
+                        return uint.MaxValue;
+                }
+            }
+        }
+        public bool VerifyCodeUnit(ref byte[] buffer, int index, int codeUnitIndex, int digitOrSymbol)
```
KrzysztofCwalina: >VerifyCodeUnit [](start = 20, length = 14)

make private

tests/System.Text.Primitives.Tests/ParserTests.cs
```diff
@@ -28,12 +28,92 @@ private byte[] UtfEncode(string s, bool utf16)
...
+                null, // so that it is != to uft8DigitsAndSymbols
+                new byte[] { 73, 110, 102, 105, 110, 105, 116, 121, },
+                new byte[] { 45, }, // minus sign
+                new byte[] { 43, }, // plus sign
+                new byte[] { 78, 97, 78, }, // NaN
+                new byte[] { 69, }, // E
+            };
+            var utf8ParsingTrie = new FormattingData.TrieNode[]
+            {
+                new FormattingData.TrieNode { valueOrNumChildren = 17, index = 0x3004390D },
```
KrzysztofCwalina: We need docs/comment describing the format.

jonfortescue: Done.

KrzysztofCwalina:
Good work. Thanks!

PR #772, opened 2016-08-25
Adding libuv.dll

KrzysztofCwalina:


davidfowl:
You can do this better. We have a libuv package now https://www.nuget.org/packages/Libuv/

PR #781, opened 2016-08-29
Adding support for parsing string and Utf8String

KrzysztofCwalina:
For now, just for boolean; I am trying to settle on the right overloads/signatures.
Similar work needs to be done for parsing other types (e.g. int)

### DIFF
--- a/src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
+++ b/src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
@@ -32,6 +32,79 @@ namespace System.Text
 
             return true;
         }
+
+        private static bool IsTrue(string text, int index)
+        {
+            if (text.Length - index < 4)
+                return false;
+
+            char firstChar = text[index];
+            if (firstChar != 't' && firstChar != 'T')
+                return false;
+
+            char secondChar = text[index + 1];
+            if (secondChar != 'r' && secondChar != 'R')
+                return false;
+
+            char thirdChar = text[index + 2];
+            if (thirdChar != 'u' && thirdChar != 'U')
+                return false;
+
+            char fourthChar = text[index + 3];
+            if (fourthChar != 'e' && fourthChar != 'E')
+                return false;
+
+            return true;
+        }
+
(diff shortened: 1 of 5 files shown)


KrzysztofCwalina:
@jonfortescue

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -83,6 +156,91 @@ private static bool IsFalse(byte[] utf8Text, int index)
...
+                return false;
+
+            Utf8CodeUnit fifthChar = text[4];
+            if (fifthChar != 'e' && fifthChar != 'E')
+                return false;
+
+            return true;
+        }
+
+        private static bool IsFalse(ReadOnlySpan<byte> utf8Bytes)
```
jonfortescue: If we're going to have a lot of overloads of these helper methods (w/ essentially identical code except for types), we might look into putting these in a T4 template so that the code is more maintainable.

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
```diff
@@ -112,21 +270,61 @@ private unsafe static bool IsFalse(byte* utf8Text, int index, int length)
...
+                    bytesConsumed = 5;
+                    value = false;
+                    return true;
+                }
+                else {
+                    return false;
+                }
+            }
+
+            return false;
```
jonfortescue: For now just invariant utf8? Culture-specific in a future PR?

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
KrzysztofCwalina: Once I setup perf runner, I will try to have just one overload that takes Span<byte>.

src/System.Text.Primitives/System/Text/InvariantParser_bool.cs
KrzysztofCwalina: That's how the byte[] overload works for now.

PR #785, opened 2016-08-31
Added non-allocating Base64 encoder/decoder

KrzysztofCwalina:
This is not the fastest implementation I can imagine, but at least it's non-allocating, i.e. operated passed in byte buffers.

### DIFF
--- a/src/System.Binary.Base64/project.json
+++ b/src/System.Binary.Base64/project.json
@@ -1,0 +1,34 @@
+{
+  "name": "System.Binary.Base64",  
+  "version": "0.1.0-*",
+  "description": "Non-allocating Base64 encoder and decoded",
+  "authors": [
+    "Microsoft Corporation"
+  ],
+  "copyright": "Microsoft Corporation, All rights reserved",
+  "packOptions": {
+    "tags": [
+      ".NET non-allocating Base64 encoder and decoder"
+    ],
+    "releaseNotes": "Pre-release package, for testing only",
+    "licenseUrl": "http://go.microsoft.com/fwlink/?LinkId=329770",
+    "iconUrl": "http://go.microsoft.com/fwlink/?LinkID=288859",
+    "projectUrl": "https://github.com/dotnet/corefxlab",
+    "requireLicenseAcceptance": true,
+  },
+  "buildOptions": {
+    "keyFile": "../../tools/Key.snk"
+  },
+  "dependencies": {
+    "System.Runtime": "4.0.0",
+    "System.Runtime.InteropServices": "4.0.0",
+    "System.Diagnostics.Debug": "4.0.0",
+    "System.Slices": { "target": "project" },
+    "System.Text.Utf8": { "target": "project" }
+  },
+  "frameworks": {
+    "netstandard1.0": {
+      "imports": [ "dotnet5.1" ]
+    }
+  }
+}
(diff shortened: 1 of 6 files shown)


KrzysztofCwalina:
as requested by @davidfowl

jonfortescue:
This LGTM

davidfowl:
System.Binary.Base64, a whole new package for this single class with 2 functions? NPM here we come 😄 .

src/System.Binary.Base64/project.json
```diff
@@ -0,0 +1,34 @@
...
+    "licenseUrl": "http://go.microsoft.com/fwlink/?LinkId=329770",
+    "iconUrl": "http://go.microsoft.com/fwlink/?LinkID=288859",
+    "projectUrl": "https://github.com/dotnet/corefxlab",
+    "requireLicenseAcceptance": true,
+  },
+  "buildOptions": {
+    "keyFile": "../../tools/Key.snk"
+  },
+  "dependencies": {
+    "System.Runtime": "4.0.0",
```
davidfowl: This should be 4.1.0

KrzysztofCwalina:
I might put it into System.Binary later. I just need to remove dependency on Utf8String

src/System.Binary.Base64/project.json
```diff
@@ -0,0 +1,34 @@
...
+    "System.Runtime.InteropServices": "4.0.0",
+    "System.Diagnostics.Debug": "4.0.0",
```
davidfowl: 4.0.11

src/System.Binary.Base64/project.json
```diff
@@ -0,0 +1,34 @@
...
+    "iconUrl": "http://go.microsoft.com/fwlink/?LinkID=288859",
+    "projectUrl": "https://github.com/dotnet/corefxlab",
+    "requireLicenseAcceptance": true,
+  },
+  "buildOptions": {
+    "keyFile": "../../tools/Key.snk"
+  },
+  "dependencies": {
+    "System.Runtime": "4.0.0",
+    "System.Runtime.InteropServices": "4.0.0",
```
davidfowl: 4.1.0

src/System.Binary.Base64/project.json
KrzysztofCwalina: Why?

davidfowl:
@KrzysztofCwalina WRT why the later versions of dependencies:
The latest package versions contain *older* DLL versions. Targeting netstandard 1.0 is enough to run *everywhere*. If you don't use the packages that have netstandard then you're forcing all consumers to use "imports" because older package versions don't have the netstandard moniker.

davidfowl:
Also do these get published on every commit or just once daily?

KrzysztofCwalina:
@davidfowl, I will merge and kick start package build.

KrzysztofCwalina:
I would love to target NS 1.0, but not without trimming. A few months ago somebody changed the dependencies to NS1.0 and it was a mess. Programs would not compile because there was something wrong with packages that the program was not even using (but are part of NS 1.0). And so I change it back to use individual contracts.

I don't want to target newer individual contracts, because I only use APIs available in S.R 4.0, and so why would I (if I don't use NS 1.0).

KrzysztofCwalina:
The package is now on https://dotnet.myget.org/feed/dotnet-corefxlab/package/nuget/System.Binary.Base64

PR #791, opened 2016-09-02
Base64.Encode performance improvements

KrzysztofCwalina:
... at least when destination is aligned.

### DIFF
--- a/src/System.Binary.Base64/System/Binary/Base64.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Text.Utf8;
+using System.Runtime.CompilerServices;
 
 namespace System.Binary
 {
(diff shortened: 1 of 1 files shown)


src/System.Binary.Base64/System/Binary/Base64.cs
```diff
@@ -43,6 +45,30 @@ public static void Encode(byte b0, byte b1, byte b2, out byte r0, out byte r1, o
             r3 = s_encodingMap[i3];
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static int Encode(ReadOnlySpan<byte> threeBytes)
+        {
+            byte b0 = threeBytes[0];
```
bartonjs: Debug.Assert(threeBytes.Length >= 3)?

src/System.Binary.Base64/System/Binary/Base64.cs
```diff
@@ -43,6 +45,30 @@ public static void Encode(byte b0, byte b1, byte b2, out byte r0, out byte r1, o
...
+            byte b1 = threeBytes[1];
+            byte b2 = threeBytes[2];
+
+            int i3 = b2 & 63;
```
bartonjs: Personally, I like bitmasks in hex

benaadams: `int i3 = b2 & 0b_0011_1111` :trollface:

src/System.Binary.Base64/System/Binary/Base64.cs
KrzysztofCwalina: Will do.

src/System.Binary.Base64/System/Binary/Base64.cs
KrzysztofCwalina: Will do. Yeah, the binary syntax would be great, but not yet available.

PR #608, opened 2016-02-09
Slices: Fix overflow in Cast<T,U>() which happens when  T is larger than U

omariom:
When T  <= U we don't pay the cost:
```C#
        [MethodImpl(MethodImplOptions.NoInlining)]
        public static void TestSpanCast(Span<byte> span)
        {
            var l = span.Cast<byte, long>();
        }
```
```ASM
; Assembly listing for method CoreClrTest.Program:TestSpanCast(struct)
; Lcl frame size = 56
       57                   push     rdi
       56                   push     rsi
       4883EC38             sub      rsp, 56
       488BF1               mov      rsi, rcx
       488D7C2420           lea      rdi, [rsp+20H]
       B906000000           mov      ecx, 6
       33C0                 xor      rax, rax
       F3AB                 rep stosd 
	   	   
       488BCE               mov      rcx, rsi
       4C8B01               mov      r8, gword ptr [rcx]
       4C8B4908             mov      r9, qword ptr [rcx+8]
       8B4110               mov      eax, dword ptr [rcx+16]
       99                   cdq      
       83E207               and      edx, 7
       03C2                 add      eax, edx
       C1F803               sar      eax, 3
       85C0                 test     eax, eax
       7506                 jne      SHORT G_M62539_IG03
       4533C0               xor      r8, r8
       4533C9               xor      r9, r9
G_M62539_IG03:
       488D542420           lea      rdx, bword ptr [rsp+20H]
       4C8902               mov      gword ptr [rdx], r8
       4C894A08             mov      qword ptr [rdx+8], r9
       894210               mov      dword ptr [rdx+16], eax
       4883C438             add      rsp, 56
       5E                   pop      rsi
       5F                   pop      rdi
       C3                   ret      

; Total bytes of code 77, prolog size 26 
```

Otherwise we check if it overflows:
```C#
        [MethodImpl(MethodImplOptions.NoInlining)]
        public static void TestSpanCast(Span<long> span)
        {
            var l = span.Cast<long, byte>();
        }
```
```ASM
; Assembly listing for method CoreClrTest.Program:TestSpanCast(struct)
; Lcl frame size = 56
       57                   push     rdi
       56                   push     rsi
       4883EC38             sub      rsp, 56
       488BF1               mov      rsi, rcx
       488D7C2420           lea      rdi, [rsp+20H]
       B906000000           mov      ecx, 6
       33C0                 xor      rax, rax
       F3AB                 rep stosd 

       488BCE               mov      rcx, rsi
[...]

PR #796, opened 2016-09-10
[Wip} Use Unsafe rather than re_write_il

benaadams:


src/System.Slices/System/PtrUtils.cs
```diff
@@ -2,16 +2,10 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
```
davidfowl: @jkotas Can you take a look here?

omariom:
@benaadams
Does it generate code of the same quality? Would be good to see perf  numbers as well.

omariom:
Usage of `AsPointer` requires pinning.

benaadams:
> Usage of AsPointer requires pinning.

Might need some new methods in Unsafe then; can't use fixed on T and GCHandle.Alloc seems a bit heavyweight

benaadams:
@omariom could call through to another function byref; though what happens currently when the asm is inlined?

jkotas:
> Usage of `AsPointer` requires pinning

Right. And pinning would have a perf hit. It can be done without a perf hit using C# 7 locals and returns. Similar to https://github.com/dotnet/coreclr/commit/0b9ba0413a005b68ad3a5aa1a1a26c0351961809.

jkotas:
> What happens currently when the asm is inlined?

The asm is operating with byrefs - it is what makes it GC safe.

benaadams:
Alas https://github.com/dotnet/corefx/pull/10540 with coreclr is currently out of my VS'15s league and VS 2015 finds them curious; so might have to pass the mantle on this one.

benaadams:
@jkotas do we need additional methods in `Unsafe`?

Currently have:
` static !!T& AsRef<T>(void* source)`

Additionally may need
```csharp
// "unknown" class, otherwise would need to chain 
// Unsafe.AsRef(Unsafe.As<T>(o))
static !!T& AsRef<T>(object source)` 
// heap structs
static !!T& AsRef<T>(T source)`
```

jkotas:
What would these methods do - what would their implementation in IL be?

src/System.Slices/System/SpanExtensions.cs
```diff
@@ -143,20 +143,16 @@ public static ReadOnlySpan<char> Slice(this string str, int start, int length)
...
             {
-                IntPtr count = PtrUtils.CountOfU<T, U>((uint)slice.Length);
-                unsafe
-                {
-                    // We can't compare IntPtrs, so have to resort to pointer comparison
-                    bool fits = (byte*)count <= (byte*)int.MaxValue;
-                    Contract.Requires(fits);
-                    countOfU = (int)count.ToPointer();
-                }
+                long count = ((uint)slice.Length * Unsafe.SizeOf<T>()) / Unsafe.SizeOf<U>();
```
omariom: `slice.Length * Unsafe.SizeOf<T>()` can [overflow](https://github.com/dotnet/corefxlab/pull/608)

benaadams:
> What would these methods do 

Return what they point at as a ref return; though I now think that can be just as easily done with a ref function.

src/System.Slices/System/PtrUtils.cs
```diff
@@ -43,114 +37,71 @@ sealed class PtrUtils
...
-            ldarg.2     // load the index
-            sizeof !!T  // load size of T
-            mul         // multiply the index and size of T
-            add         // add the result
-            ldobj !!T   // load a T value from the computed address
-            ret")]
-        public static T Get<T>(object obj, UIntPtr offset, UIntPtr index) { return default(T); }
+        public static unsafe T Get<T>(object obj, UIntPtr offset, UIntPtr index)
+        {
+            return Unsafe.Read<T>(*(byte**)Unsafe.AsPointer(ref obj) + offset.ToUInt64() + index.ToUInt32() * Unsafe.SizeOf<T>());
```
mattwarren: Maybe a dumb question, but as you're replacing some in-line IL with 3 method calls, do all the `Unsafe` methods get inlined as well? (I know they're [marked as assgressiveinlining](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/src/System.Runtime.CompilerServices.Unsafe.il), but have you verified they are?)

benaadams: The will do as the inliner thinks in IL. However these methods are invalid due to not being GC safe; need to use the ref return methods.

omariom:
@benaadams  I suspect that casting to array and fixing it will make getters and setter very slow.

benaadams:
Probably; its more a discovery on what `Unsafe` needs.

The `ref Add` methods don't quite work as its `ref + a bit` for the array offset

omariom:
@jkotas @benaadams 
Is it even possible to replace the  original trick Joe Dufy describes in his comment with `ref` operations?

jkotas:
The implementation of that trick is questionable. It is invalid IL according to the spec that is lucky to make it through the JIT - there is no implicit conversion from Object to byref.

If you run System.Slices on checked build of the JIT, you will see multiple asserts like `Assertion failed 'genActualType(op1->TypeGet()) == TYP_BYREF && genActualTypeIsIntOrI(op2->TypeGet())' in 'System.PtrUtils:ElemOffset(ref):int' (IL size 14)` because of that.

The implementation using S.R.CS.Unsafe and ref operations can look something like:

```
class RawData
{
    public byte Data;
}

...

public T this[int index]
{
    get
    {
       Contract.RequiresInRange(index, (uint)Length);
       if (Object != null)
           // This should use Unsafe.Add overload with IntPtr argument that does not exist yet. It was discussed in https://github.com/dotnet/corefx/issues/10451, 
           // but I did not add it because of it was not clear whether it is useful, what it should do exactly, and what would be right name for it.
           // return Unsafe.Add<T>(ref Unsafe.AddByteOffset<T>(ref Unsafe.As<RawData>(Object).Data, Offset), index);

           // In the meantime, the int overload will work fine, except for very large arrays...
           return Unsafe.Add<T>(ref Unsafe.As<T>(ref Unsafe.Add(ref Unsafe.As<RawData>(Object).Data, (int)Offset)), index);
       else
           return Unsafe.Add<T>(ref Unsafe.AsRef<T>((void *)Offset), index);
    }
}
```

It has extra check for `Object != null` that is hard to get rid of without the invalid IL.

benaadams:
`CopyBlock` also needs the `by ref` and `plus a bit` version.

jkotas:
> `CopyBlock` also needs the by ref and plus a bit version.

I think it is a good idea to add these overloads to S.R.CS.Unsafe in corefx now that we have a good use case for them.

benaadams:
@jkotas not sure of good api. Something like?
```csharp
static ref T Add<T>(ref object source, int byteInset, int elementOffset)
```
Are 3 types
* object native inset 0 bytes
* object `T[]` inset `arrayStart` bytes
* object `string` inset `stringStart` bytes

jkotas:
The API I had in mind was: `static ref T AddByteOffset<T>(ref T source, IntPtr byteOffset)`.

The problem with `static ref T Add<T>(ref object source, int byteInset, int elementOffset)` is that it is not possible to implement it using valid IL in performant way. S.R.CS.Unsafe is limited to what can be implemented in valid IL.

nietras:
> ```static ref T AddByteOffset<T>(ref T source, IntPtr byteOffset)```

:+1: for adding this to `Unsafe`, perhaps we can review this in a separate issue or pull request.

nietras:
@jkotas  wait... instead of `AddByteOffset` would it not be better to just add overloads to `Add` for `IntPtr` instead of a `AddByteOffset`, which is less useful? We did talk about `IntPtr` support but choose not to since use cases might be few. Adding `IntPtr` overload for both `Add` and `Subtract` seems more useful and within API design.

jkotas:
> `AddByteOffset` which is less useful?

It is more useful for this case. If we add the Add method that takes elementOffset as IntPtr, there will need to be ~3 calls to Unsafe methods instead of just one in this case.

I think it is reasonable to say: `int` is used for element offsets - indices are ints pretty much everywhere in .NET; `IntPtr` is used for byte offsets - used for the super low-level code.

nietras:
> more useful for this case. If we add the `Add` method that takes `elementOffset` as `IntPtr`, there will need to be ~3 calls to `Unsafe` methods instead of just one in this case.

Useful perhaps, but also less general in scope I would think, and there still needs to be 2 calls as far as I can tell, and the `As` calls are noops after jitting right? So is the convience that important in the general case?

Adding `Pinnable` as we have discussed would help too, though, so `RawData` is not needed.

benaadams:
Will need to capture the extra apis needed in a corefx Issue for Unsafe, then can close this?

KrzysztofCwalina:
@benaadams, what should we do with this one?

benaadams:
Need to capture the missing methods in corefx for Unsafe then close.

KrzysztofCwalina:
BTW, the Unsafe class is available only in the most recent NetStandard, correct? I wanted System.Slices to work on as many versions as possible.

jkotas:
> BTW, the Unsafe class is available only in the most recent NetStandard, correct?

Unsafe class is OOB, supported all the way from NetStandard 1.0.

davidfowl:
If we can get this working, the build time for this repository will be much better 😄

KrzysztofCwalina:
@jkotas, ah good to know.

benaadams:
Required functions are now in Unsafe thanks to @nietras 👍

nietras:
Thanks, can I ask what the plan is regarding this? Are you going to wait
until C# 7 is released?

KrzysztofCwalina:
How do I use Unsafe.AsPointer with the Span's fields in corfxlab? The fields are readonly and so I cannot pass them in by-ref to AsPointer.

jkotas:
Using `Unsafe.AsPointer(ref ...)` on Span fields does not sound right. What do you need to use the AsPointer for?

KrzysztofCwalina:
For example, I would like to remove [UnsafeUtilities.Get](https://github.com/dotnet/corefxlab/blob/master/src/System.Slices/System/UnsafeUtilities.cs#L111). Wat S.R..CS.Unsafe APIs should I use instead?

jkotas:
The current implementation for `UnsafeUtilities.Get` contains invalid IL. If you run it through the debug build of the JIT, you will hit asserts (see https://github.com/dotnet/corefxlab/pull/796#issuecomment-246667113 for details). It happens to generate a functional code today, but there are likely cases where the JIT will just blow up; or where the GC tracking information will be incorrect.

The Unsafe class does contain invalid IL, and so the approach needs adjustment. Here is an example how the Span indexer can be implemented using the recently added Unsafe methods:

```
class RawData
{
    internal T Data;
}

public T this[int index]
{
    get
    {
        Contract.RequiresInRange(index, (uint)Length);
        return Unsafe.Add(ref Unsafe.AddByteOffset(ref Unsafe.As<RawData>(Object).Data, Offset), index)
    }
...
```

- This implementation requires change in the meaning of the Offset field. It needs to be adjusted by the offset of RawData.Data field.
- This implementation is not compatible with runtimes that do not internally store array elements inline (e.g. it would not work with disassociative arrays). The existing implementation in corefxlab has the same problem.

nietras:
> needs to be adjusted by the offset of RawData.Data field.

@jkotas regarding this, do we need to add unsafe code for finding these offsets? E.g. something like:
```csharp
IntPtr ByteOffset<L, R>(ref L left, ref R right);
```
and will this also be useful instead of:
```csharp
[ILSub(@"
            ldarg.0
            ldc.i4 0
            ldelema !!T
            ldarg.0
            sub
            ret")]
        internal static int ElemOffset<T>(T[] arr)
 { throw new Exception("ILSub did not execute."); }
```

Or do we need to add something similar like this to `Unsafe`. I though this element offset might be found with:
```csharp
var a = new byte[10];
var offset = Unsafe.ByteOffset(ref a, ref a[0]);
```
but it may not be as performant due to bounds checking.

jkotas:
> Unsafe.ByteOffset(ref a, ref a[0]);

This would subtract address of `a` on the stack from address of first element of the array. Instead, it needs to be something like: `Unsafe.ByteOffset(ref a[0], ref Unsafe.As<RawData>(a).Data);`.

I agree that Unsafe.ByteOffset method should be useful (you can do the operation today, but it requires pinning). It may be a good idea to make both arguments to have same type - to make you write extra extra code to diff apples and oranges.

> due to bounds checking

I do not think we need to worry about bounds checking here. This is done in `Span<T>` constructor that has to do some bounds checking already. It may be possible to structure it such that the JIT will eliminate bounds check for this one as redundant.

nietras:
> Unsafe.ByteOffset(ref a[0], ref Unsafe.As<RawData>(a).Data);

Ha, so careful with the function signature, then mess up the example completely, yes that is correct.

> it requires pinning

Yes, this was what I thought would be good to avoid.

Ok, added a issue for this addition to `Unsafe` https://github.com/dotnet/corefx/issues/12593

PR #800, opened 2016-09-12
Added support for backpressure

KrzysztofCwalina:
David Fowler requested this so he can use IFromatter in his channels project.

For now, back pressure is synchronous. The caller (who gets StreamStatus.Backpressue) can decide to wait, retry, spin, or any other fallback it thinks is best in the situation.

As part of the change, I noticed that IFormatter basically a stream with FormattingData,
and so I am playing with the idea.

### DIFF
--- a/src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
+++ b/src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
@@ -8,14 +8,20 @@ namespace System.Text.Formatting
     // this interface would be implemented by types that want to support formatting, i.e. TextWriter/StringBuilder-like types.
     // the interface is used by an extension method in IFormatterExtensions.
     // One thing I am not sure here is if it's ok for these APIs to be synchronous, but I guess I will wait till I find a concrete issue with this.
-    public interface IFormatter
+    public interface IStream
     {
-        Span<byte> FreeBuffer { get; }
-        void CommitBytes(int bytes);
+        Span<byte> AvaliableBytes { get; }
+        void Advance(int byteCount);
 
-        /// <summary>desiredFreeBytesHint == -1 means "i don't care"</summary>
-        void ResizeBuffer(int desiredFreeBytesHint = -1);
+        /// <summary>Returns false if there is backpressure and the request can be fullfiled later. Throws if the request cannot be fullfiled.</summary>
+        bool TryEnsureAvaliable(int minimunByteCount = 1);
+    }
 
+    // this interface would be implemented by types that want to support formatting, i.e. TextWriter/StringBuilder-like types.
+    // the interface is used by an extension method in IFormatterExtensions.
+    // One thing I am not sure here is if it's ok for these APIs to be synchronous, but I guess I will wait till I find a concrete issue with this.
+    public interface IFormatter : IStream
+    {       
         FormattingData FormattingData { get; }
     }
 }
--- a/src/System.Text.Formatting/System/Text/Formatting/BufferFormatter.cs
+++ b/src/System.Text.Formatting/System/Text/Formatting/BufferFormatter.cs
@@ -1,4 +1,5 @@
-﻿using System.Buffers;
+﻿using System;
+using System.Buffers;
 using System.Text;
 
 namespace System.Text.Formatting
@@ -37,7 +38,7 @@ namespace System.Text.Formatting
             _count = 0;
         }
 
-        Span<byte> IFormatter.FreeBuffer
+        Span<byte> IStream.AvaliableBytes
         {
             get
             {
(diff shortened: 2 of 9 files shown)


KrzysztofCwalina:
@davidfowl

src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
```diff
@@ -8,14 +8,27 @@ namespace System.Text.Formatting
...
+    public interface IFormatter : IStream
+    {       
         FormattingData FormattingData { get; }
     }
+
+    public enum StreamStatus
+    {
+        OK,          // made the desired byte count avaliable
+        Eos,         // reached EOS and so the desired byte count will never be made avaliable
+        Backressure, // there is back pressure. the desired bytes are not avaliable, but can be in the future
```
svick: Typo: Backressure → Backpressure

src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
```diff
@@ -8,14 +8,27 @@ namespace System.Text.Formatting
...
+    // One thing I am not sure here is if it's ok for these APIs to be synchronous, but I guess I will wait till I find a concrete issue with this.
+    public interface IFormatter : IStream
+    {       
         FormattingData FormattingData { get; }
     }
+
+    public enum StreamStatus
+    {
+        OK,          // made the desired byte count avaliable
+        Eos,         // reached EOS and so the desired byte count will never be made avaliable
```
svick: Wouldn't spelling out the acronym here be better?

davidfowl:
It'll be good to get some examples that actually have backpressure. Maybe a fixed sized formatter that can only write n bytes at a time, after that it'll return BackPressure from EnsureAvailable. Then we can test out something that blocks, vs polls vs uses a task. I don't think you can make something efficient with the current model.

davidfowl:
@KrzysztofCwalina What about a callback based model? That itself is decoupled from any async model. You just need an optimistic way to check if there's back pressure, but you also need a way to get notified when the state changes.

src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
KrzysztofCwalina: Thanks. Will fix.

src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
KrzysztofCwalina: Sure. Thanks for the suggestion.

KrzysztofCwalina:
@davidfowl, I wrote such sample. But I don't think it is very useful for deciding whether it's the right design or not. 
I think we need to decide what we want the following method do:

```csharp
public static void Append<TFormatter>(
    this TFormatter formatter, 
    int value, 
    Format.Parsed format = default(Format.Parsed)
) where TFormatter : IFormatter
{
    int bytesWritten;
    while (!value.TryFormat(formatter.AvaliableBytes, format, formatter.FormattingData, out bytesWritten))  {
        formatter.EnsureAvaliable();
        bytesWritten = 0;
    }
    formatter.Advance(bytesWritten);
}
```
In addition, we need to decide what the callers of Append look like.

src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
```diff
@@ -8,14 +8,20 @@ namespace System.Text.Formatting
     // this interface would be implemented by types that want to support formatting, i.e. TextWriter/StringBuilder-like types.
     // the interface is used by an extension method in IFormatterExtensions.
     // One thing I am not sure here is if it's ok for these APIs to be synchronous, but I guess I will wait till I find a concrete issue with this.
-    public interface IFormatter
+    public interface IStream
     {
-        Span<byte> FreeBuffer { get; }
-        void CommitBytes(int bytes);
+        Span<byte> AvaliableBytes { get; }
```
davidfowl: typo should be `AvailableBytes`

davidfowl: What about the following:

```C#
public interface IStream
{
    Span<byte> AvailableBytes { get; }
    void Advance(int byteCount);
    bool TryEnsureAvailable(int minimunByteCount = 1);
    ValueTask FlushAsync();
}
```

benaadams: `Task FlushAsync()` unless its returning a value?

omariom:
no sure if it is relevant but the discussion is interesting

https://github.com/Reactive-Extensions/Rx.NET/issues/19

src/System.Text.Formatting/System/Text/Formatting/IFormatter.cs
KrzysztofCwalina: @davidfowl, I cannot answer your question before I know what we are planing to do in IFormatterExtensions.Append. It seems to me like the Append methods cannot be async, and so I am not sure FlushAsync makes sense on this interface. 

I prototypes some things yesterday, and currently I am trying to change the Append methods to TryAppend and then build spinning sync Appends and async AppendAsync on top of these TryAppends. I still have some problems with it, but it looks promising.

KrzysztofCwalina:
closing this one. We decided to handle backpressure outside of this interface.

PR #801, opened 2016-09-12
Obsoleted UnsafeBuffer and TryGetArray

KrzysztofCwalina:
Added TryGetArrayElseGetBuffer.

The existing APis need to be used together to make the interop safe.
The new API makes it more in the face.
This is how the API should be used:

```csharp
void Interop(Span<Byte> s){
    void* p;
    ArraySegment<byte> a;

    if(s.TryGetArrayOrGetPointer(out a, out p)){
        fixed(byte* pB = a.Array){
	    var pOffset = pB + a.Offset;
            DoInterop(pOffset. s.Length);
        }
    }
    else {
    	DoInterop((byte*)p, s.Length);
    }
}
```

This API might be removed altogether when we add proper pinning to spans.

### DIFF
--- a/src/System.Slices/System/ReadOnlySpan.cs
+++ b/src/System.Slices/System/ReadOnlySpan.cs
@@ -150,6 +150,29 @@ namespace System
             get { return Length == 0; }
         }
 
+        /// <summary>
+        /// Gets array if the slice is over an array, otherwise gets a pointer to memory.
+        /// </summary>
+        /// <returns>true if it's a span over an array; otherwise false (if over a pointer)</returns>
+        /// <remarks>This method can be used for interop, while we are waiting for proper pinning support. Make sure the array contents are read-only.</remarks>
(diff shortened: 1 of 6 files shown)


src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -150,6 +150,29 @@ public bool IsEmpty
...
 
+        /// <summary>
+        /// Gets array if the slice is over an array, otherwise gets a pointer to memory.
+        /// </summary>
+        /// <returns>true if it's a span over an array; otherwise false (if over a pointer)</returns>
+        /// <remarks>This method can be used for interop, while we are waiting for proper pinning support. Make sure the array contents are read-only.</remarks>
+        public unsafe bool TryGetArrayElseGetPointer(out ArraySegment<T> array, out void* pointer)
+        {
+            var a = Object as T[];
+            if (a == null)
```
omariom: what if `Object` is string?

PR #802, opened 2016-09-12
removed obsolete attributes

KrzysztofCwalina:
There are some scenarios that are not covered by the new APIs. 

For example, if the span wraps an array, but the calling code knows the array is already pinned and wants to get the pointer.


KrzysztofCwalina:
@davidfowl

davidfowl:
:shipit: !

Issue #687, opened 2016-06-06
Slices fails strong name validation on  Windows 7


omariom:
@livarcocc

It throws 
>FileLoadException: Could not load file or assembly 'System.Slices, Version=0.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' or one of its dependencies. Strong name validation failed. (Exception from HRESULT: 0x8013141A)

omariom:
Can it be caused by not having local admin rights?

benaadams:
And on Win10 
> Unhandled Exception: System.IO.FileLoadException: Could not load file or assembly 'System.Slices, Version=0.1.0.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' or one of its dependencies. Strong name validation failed. (Exception from HRESULT: 0x8013141A) ---> System.Security.SecurityException: Strong name validation failed. (Exception from HRESULT: 0x8013141A)

xenocons:
Hello,

This appears to be as a result of IL modifications after the dll is signed, it should just be a matter of resigning the binary after the IL modifications. I would love this fixed as it wreaks havoc with unattended deployments (azure)

Alternate workaround is to add a validation bypass via the registry for that pubkey token, or use sn.exe included with Windows SDK.

xenocons:
Just as a note, I stumbled upon this https://github.com/gluck/il-repack , which seems to offer a parameter to supply a keyfile to sign the resulting assembly, I am not sure if the current solution can do this, so it might be a nice move.

omariom:
> This appears to be as a result of IL modifications after the dll is signed

Yes, it started working after I commented out `keyFile` in projects.json

KrzysztofCwalina:
cc @davidfowl

KrzysztofCwalina:
I believe david fixed it.

PR #806, opened 2016-09-13
Resign System.Slices when using ilasm

davidfowl:
- Added keyfile to the ilasm step
- Fixed up the output paths of the xproj to match the latest tooling version. Also is more consistent with command line builds.

#687


KrzysztofCwalina:
Thanks!

PR #808, opened 2016-09-13
Fixes #807

### LINKED ISSUE #807: System.Buffers.Experimental: Cannot reuse a returned buffer
Returning a Span to the pool leads to an 0-length NativeBuffer (the native buffer is explicitly created that way in NativeBufferPool:Return) being returned to the bucket.   The next time this buffer is to be used (by another call to Rent on the same bucket), it will create a 0-length Span.  Which doesn't work.

Easy repro steps:
    let pool = System.Buffers.NativeBufferPool.Shared
    let u = pool.Rent 10 in pool.Return u
    let v = pool.Rent 10 in pool.Return v  // v.Length=0, and the .Return() call crashes

I think the bucket should just forcefully set the NativeBuffer's length before returning it.

KrzysztofCwalina:
I tried to fix just this bug, but discovered this component has some many other problems
that it probably needs to be rewritten from scratch.
I did such a rewrite, but had time to implement only a very limited support (see all thrown exceptions),
and quite naive synchronization logic.

I will filed an issue (#809) to properly implement this component.


KrzysztofCwalina:
cc @vlabrecque

PR #811, opened 2016-09-14
Expanding support for "multispans"

KrzysztofCwalina:
Added an abstraction that can support various implementations of containers of spans
Added a method that illustrates how to parse UInt32 out of a milti-span

The multispan abstraction (ISequence<T>) allows array based and linked implementations of sequences.
A linked-list like container, would implement the interface as follows:

        public T TryGetItem(ref Position position)
        {
            if (_head != null && position.IsValid && !position.IsEnd)
            {
                var node = (Node)position.ObjectPosition;
                if(node==null) { node = _head; }
                var result = node._item;
                if (node._next != null)
                {
                    position.ObjectPosition = node._next;
                }
                else
                {
                    position = Position.End;
                }
                return result;
            }

            position = Position.Invalid;
            return default(T);
        }

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/ISequence.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ISequence.cs
@@ -1,0 +1,66 @@
+﻿using System;
+using System.ComponentModel;
+using System.Runtime.CompilerServices;
+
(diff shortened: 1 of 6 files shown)


KrzysztofCwalina:
@davidfowl

PR #812, opened 2016-09-14
Added a type representing Span<byte> future

KrzysztofCwalina:
The type can be stored on the heap.
It's subject to tearing, and should not be used in multi-threaded context, or passed to user code.

The type is used as a buffer of an abstract buffer pool (either managed or native). 

The change also adds such abstract BufferPool type and makes both NativeBufferPool and 
ManagedBufferPool implement this abstraction.


KrzysztofCwalina:
cc @davidfowl

PR #817, opened 2016-09-15
explicit conversion to ArraySegment

benaadams:


benaadams:
Hmm.. the conversion for Span might not make sense as you are writing to something that doesn't "exist"

omariom:
We have implicit conversion operator for`ArraySegment` but explicit `ToArray` method.
I think both conversation must be aligned: either add `ToArraySegment` or implicit conversation to array.

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -143,6 +143,16 @@ internal ReadOnlySpan(object obj, UIntPtr offset, int length)
             return new ReadOnlySpan<T>(slice.Object, slice.Offset, slice.Length);
         }
 
+        /// <summary>
+        /// Gets ArraySegment containing the ReadOnlySpan's data.
+        /// </summary>
+        /// <remarks>This method can be used for interop, as array data is mutable, an array is allocated and copy of the data is returned in it.</remarks>
+        public static explicit operator ArraySegment<T>(ReadOnlySpan<T> span)
```
mgravell: For my money, I'd want it to be *shockingly obvious* that  I'm allocating here; by comparison to `CreateArray`, I wonder whether  `CreateArraySegment()` is a better thing here

KrzysztofCwalina: CreateArraySegment could also be an extension method, which I like, as I am not super thrilled by Span<T> now depending on ArraySegment<T>. I would prefer to keep Span<T> as simple as language primitives.

benaadams:
I don't think this is great anymore :)

PR #818, opened 2016-09-15
Added non-blocking formatters

KrzysztofCwalina:
These can be used to implement fully async formatting:

```csharp
public static async Task WriteAsync<TStream>(this TStream stream, T value) where TStream : IStream, IFormatter {
    while (true) {
        if (stream.TryAppend(value)) { // non-blocking formatter method
            await Task.CompletedTask;
            break;
        }
        await stream.FetchMoreBufferSpace(); // method on hypothetical IStream
    }
}
```


KrzysztofCwalina:
@davidfowl

PR #819, opened 2016-09-15
update package.ps1 to use $(Build.BuildNumber)

joshfree:
The Build number format is now specified in the Build Definition
and is a supplied parameter.

### DIFF
--- a/scripts/package.ps1
+++ b/scripts/package.ps1
@@ -1,6 +1,7 @@
 ﻿Param(
     [string]$Configuration="Debug",
-    [string]$ApiKey
+    [string]$ApiKey,
+    [string]$BuildVersion
 )
 
 $repoRoot = "$PSScriptRoot\.."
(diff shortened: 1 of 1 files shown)


davidfowl:
What is the build number @joshfree ?

scripts/package.ps1
```diff
@@ -21,8 +22,7 @@ Function Ensure-Nuget-Exists {
 Write-Host "** Building all NuGet packages. **"
 foreach ($file in [System.IO.Directory]::EnumerateFiles("$repoRoot\src", "project.json", "AllDirectories")) {
     Write-Host "Creating NuGet package for $file..."
-    $dateSuffix = [System.DateTime]::Now.ToString('eyyMMdd-1')
```
davidfowl: I think we need a fallback if `$BuildVersion = $null` so that we can build locally as well.

joshfree: thanks I made the field non-mandatory (kept the current default)

joshfree:
> What is the build number @joshfree ?

It'll be set in vsts to the current scheme, only the hardcoded -1 suffix will increment with each build (commit) that calendar day.

davidfowl:
> It'll be set in vsts to the current scheme, only the hardcoded -1 suffix will increment with each build (commit) that calendar day.

What scheme is that.

joshfree:
> What scheme is that.

https://dotnet.myget.org/feed/dotnet-corefxlab/package/nuget/System.Buffers/0.1.0-e160122-1

this branch currently uses this eYYMMdd-x pre-release suffix where x will now be $(rev:.r) instead of always being '1'.  It's goofy with the hardcoded e (ask @KrzysztofCwalina about the 'e' previously being a 'd' for 'day' 😄).  

Don't spend too much time on this scheme right now; this is just the first commit to turn on multiple builds per day.  I'll bump the package version numbers and drop the 'e' later.

Issue #827, opened 2016-09-20
Span<T> - find if one span contains another; perhaps IsSliceOf


mgravell:
A scenario I find myself needing is to see whether one span is a sub-span of another. This test is semantically valid for  both array and pointer spans, and is a related operation to `Slice`

Suggested API:

```
Span<T>
    bool IsSliceOf(Span<T> span);
    bool IsSliceOf(Span<T> span, out int offset);
```

where it is:
- `true` if both spans are array-based, for the exact same array, where the inner offet >= outer offset, and inner offset+count <= outer offset+count
- `true` if both spans are pointer-based, and inner ptr >= outer ptr and inner ptr+count <= outer ptr+count
- `false`  otherwise

or could be unified to:

```
    int GetSliceIndex(Span<T> span)
```

(returns -1 if not a slice; kinda ugly)

Would be happy to contribute towards the code, but wanted to discuss API first.

omariom:
What if the first Span is array based but the second one is pointer based created from the same **pinned** array?
Is it currently possible to  handle such case without pinning the array based Span?

mgravell:
that would be the `false` case, IMO; it could not have been created as a result of `Slice`

KrzysztofCwalina:
Can we get the same semantics for this span (three field) and the byref filed span in coreclr repo?

cc @jkotas

jkotas:
> What if the first Span is array based but the second one is pointer based created from the same pinned array?

`IsSliceOf` should detect the overlap in this case as well.

> Is it currently possible to handle such case without pinning the array based Span?

You either need to pin, or have more unsafe helpers that operate with byrefs.

> Can we get the same semantics for this span (three field) and the byref filed span in coreclr repo?

I think so.

mgravell:
@jkotas  it *feels* like it would be reasonable to *not* detect it, though; I guess what we're asking is:

> given `Span<T> foo` and `bar`, is there an `i` for which we could have called `foo = bar.Slice(i, {foo.Length});` (and optionally, tell me the `i`)

To my mind, *that* is what  `IsSlice` means. However, if `foo` and `bar` are of different fundamental types, the answer  to that question is simply: "no"

jkotas:
To me, Span should is a safe pointer to a block of memory. The internal representation of the block of memory (pinned vs. not-pinned) should not leak into behavior. If the internal representation leaks into behavior:
[...]

PR #829, opened 2016-09-20
Add Span<T>.IsSliceOf; addresses #827

mgravell:
Suggested code for #827

### DIFF
--- a/src/System.Slices/System/Span.cs
+++ b/src/System.Slices/System/Span.cs
@@ -400,6 +400,36 @@ namespace System
             return false;
         }
 
+        /// <summary>
+        /// Determines whether the current span is a slice of the supplied span
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool IsSliceOf(Span<T> parentSpan)
+        {
+            var elementSize = PtrUtils.SizeOf<T>();
+            return Object == parentSpan.Object // either both pointers (both null), or same array instance
+                && parentSpan.Offset.ToUInt64() <= Offset.ToUInt64() // parent must start earlier (or equal start)
+                && parentSpan.Offset.ToUInt64() + (ulong)(Length * elementSize) >= Offset.ToUInt64() + (ulong)(Length * elementSize); // parent must  end later (or equal end)
+        }
+
+        /// <summary>
+        /// Determines whether the current span is a slice of the supplied span
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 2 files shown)


mgravell:
Fixing problems; I screwed up the tests

src/System.Slices/System/Span.cs
```diff
@@ -400,6 +400,36 @@ public override bool Equals(object obj)
...
+        /// <summary>
+        /// Determines whether the current span is a slice of the supplied span
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool IsSliceOf(Span<T> parentSpan)
+        {
+            var elementSize = PtrUtils.SizeOf<T>();
+            return Object == parentSpan.Object // either both pointers (both null), or same array instance
+                && parentSpan.Offset.ToUInt64() <= Offset.ToUInt64() // parent must start earlier (or equal start)
+                && parentSpan.Offset.ToUInt64() + (ulong)(Length * elementSize) >= Offset.ToUInt64() + (ulong)(Length * elementSize); // parent must  end later (or equal end)
```
omariom: It will overflow in a legal case Length = 300000000 and  elementSize = 8

src/System.Slices/System/Span.cs
```diff
@@ -400,6 +400,36 @@ public override bool Equals(object obj)
...
 
+        /// <summary>
+        /// Determines whether the current span is a slice of the supplied span
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool IsSliceOf(Span<T> parentSpan)
+        {
+            var elementSize = PtrUtils.SizeOf<T>();
+            return Object == parentSpan.Object // either both pointers (both null), or same array instance
+                && parentSpan.Offset.ToUInt64() <= Offset.ToUInt64() // parent must start earlier (or equal start)
```
omariom: To make it a bit more efficient on x86 you can use `ToPointer()` instead of `ToInt64()`.

mgravell:
Huh, it won't let me re-open; dammit :/ Fixed the *obvious* bugs - looking at  the two points raised above

src/System.Slices/System/Span.cs
mgravell: fix pushed, thanks

src/System.Slices/System/Span.cs
mgravell: fix pushed, thanks; can only really apply to the left hand bound check, due to the math issue

mgravell:
Moving to https://github.com/dotnet/corefxlab/pull/830 because github refuses to let this one live

PR #834, opened 2016-09-21
Added Base64 encoding in place

KrzysztofCwalina:
cc: @mgravell, @davidfowl, @joshfree

### DIFF
--- a/src/System.Binary.Base64/System/Binary/Base64.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64.cs
@@ -107,6 +107,40 @@ namespace System.Binary
             return di; 
         }
 
+        /// <summary>
+        /// 
+        /// </summary>
+        /// <param name="buffer">Buffer containing source bytes and empty space for the encoded bytes</param>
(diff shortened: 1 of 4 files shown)


src/System.Binary.Base64/System/Binary/Base64.cs
```diff
@@ -107,6 +107,40 @@ public static int Encode(ReadOnlySpan<byte> source, Span<byte> destination)
...
+
+            // encode last pack to avoid conditional in the main loop
+            if (leftover != 0) {
+                var sourceSlice = buffer.Slice(sourceIndex, leftover);
+                var desitnationSlice = buffer.Slice(destinationIndex, 4);
+                destinationIndex -= 4;
+                Encode(sourceSlice, desitnationSlice);
+            }
+
+            for (int index = sourceIndex - 3; index>=0; index -= 3) {
```
KrzysztofCwalina: The loop could be optimized. There is no need to use the encode method that operates on Span<byte>

PR #841, opened 2016-09-23
Made a stub project to port code from ASP.NET

davidfowl:
/cc @KrzysztofCwalina @troydai


troydai:
Are we going to check in the stub first or just pile on this PR? And are we really want to just call it `UTF8`, not more generic?

KrzysztofCwalina:
@troydai, good questions. I hope that we can make it more generic or even unify with the corfx encodings, but for now I think it's ok to merge this so we can prototype quickly.

PR #842, opened 2016-09-25
remove all generic interfaces that put Span<T> in generic arg

### LINKED ISSUE #838: Remove IEnumerable<T> from Spans.
Since Span<T> will be stack-only, it cannot implement this interface.

adamsitnik:
@KrzysztofCwalina I have removed it also from Utf8String since it's based on Span as well

### DIFF
--- a/src/System.Slices/System/Contract.cs
+++ b/src/System.Slices/System/Contract.cs
@@ -83,6 +83,8 @@ namespace System
             }
         }
 
+        internal static InvalidOperationException InvalidOperationExceptionForBoxingSpans() => new InvalidOperationException("Spans must not be boxed");
+
         private static void ThrowArgumentException()
         {
             throw new ArgumentException();
--- a/src/System.Slices/System/ReadOnlySpan.cs
+++ b/src/System.Slices/System/ReadOnlySpan.cs
@@ -345,6 +345,8 @@ namespace System
             }
         }
 
+        public override bool Equals(object obj) { throw Contract.InvalidOperationExceptionForBoxingSpans(); }
+
         /// <summary>
         /// Checks to see if two spans point at the same memory.  Note that
         /// this does *not* check to see if the *contents* are equal.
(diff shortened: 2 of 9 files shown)


davidfowl:
We have to remove all interfaces that put `Span<T>` in the generic arg.

adamsitnik:
@davidfowl good point!

I have removed these interfactes + `override bool Equals(object obj)` implementation that was allowing to box spans.

src/System.Slices/System/ReadOnlySpan.cs
```diff
@@ -353,27 +349,14 @@ public override int GetHashCode()
...
-            return ReferenceEquals(other);
-        }
+        public bool Equals(ReadOnlySpan<T> other) => ReferenceEquals(other);
 
-        public bool Equals(Span<T> other)
-        {
-            return other.StructuralEquals(Object, Offset, Length);
-        }
-
-        public override bool Equals(object obj)
```
KrzysztofCwalina: Why is this removed?

benaadams: Should it just return false? Assume `ReadOnlySpan<T>` can't be boxed to object?

KrzysztofCwalina: or throw till we have boxing disabled.

adamsitnik: @KrzysztofCwalina I added throwing override for `bool Equals(object obj)`

src/System.Slices/System/Contract.cs
```diff
@@ -83,6 +83,8 @@ public static void RequiresInInclusiveRange(uint start, uint length, uint existi
             }
         }
 
+        internal static InvalidOperationException InvalidOperationExceptionForBoxingSpans() => new InvalidOperationException("Spans must not be boxed");
```
omariom: What are the benefits of having it cached?

omariom: Oh.. it is not a field.. Ignore :)

src/System.Slices/System/ReadOnlySpan.cs
KrzysztofCwalina: Great. thanks! ... and now the tests show all the bugs :-)

adamsitnik: @KrzysztofCwalina ups.. Sorry for that! Corrected!

PR #844, opened 2016-09-27
Removes Bytes and added Memory<T>

davidfowl:
- Removed UnsafePointer and TryGet* from Span<T> basically
disallowing unwrapping of the Span<T>.
- Memory<T> can be unwrapped using those methods
- Updated to pools to return Memory<byte> instead of Bytes
- Updated the samples to use Memory<T>
- Updated the unit test to unwrapped Span<T>
and made it use Memory<T> instead

Some of the call sites can use `Span<T>` when it supports pinning

/cc @KrzysztofCwalina


src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,132 @@
...
+            {
+                _memory = pointer;
+            }
+
+            _array = null;
+            _offset = 0;
+            _memoryLength = length;
+        }
+
+        public unsafe Memory(T[] array) : this(array, 0, array.Length)
```
ViIvanov: Is it OK that calling `new Memory<T>(default(T[]));` will cause a NullReferenceException because you have `array.Length`? May be, `array?.Length ?? 0` is better?

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,132 @@
...
+using System;
+using System.Runtime;
+
+namespace System.Buffers
+{
+    /// <summary>
+    /// Represents a Span<byte> factory</byte>
+    /// </summary>
+    /// <remarks>
+    /// This struct is not safe for multithreaded access, i.e. it's subject to struct tearing that can result in unsafe memory access.
```
jkotas: This should be changed to class...

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,132 @@
...
+            _array = null;
+            _offset = 0;
+            _memoryLength = length;
+        }
+
+        public unsafe Memory(T[] array) : this(array, 0, array.Length)
+        {
+        }
+
+        public unsafe Memory(T[] array, int offset, int length, bool isPinned = false)
```
KrzysztofCwalina: Why do we need isPinned? If the caller pinned the array, they should construct Memory with the ctor that takes a pointer.

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: Why? If it's immutable, we can keep it as a struct. I think in some of the early discussions/prototypes it was responsible for disposing and needed to be a class. Now, it's not disposable/mutable.

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,132 @@
...
+                    _memory = null;
+                }
+            }
+
+            _array = array;
+            _offset = offset;
+            _memoryLength = length;
+        }
+
+        public Span<T> Span
```
KrzysztofCwalina: I don't like that we have Slice(...) method and Span property (and some even differently named APIs in other palaces that do similar things). Should we standardize on having method Slice(), Slice(...), etc.?

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            return memory.Span;
+        }
+
+        public static implicit operator ReadOnlySpan<T>(Memory<T> memory)
+        {
+            return memory.Span;
+        }
+
+        public unsafe void* UnsafePointer
```
KrzysztofCwalina: We need to standardize these methods too. What about TryGetArray and TryGetPointer?

src/System.Slices/System/Buffers/Memory.cs
jkotas: This looks very much like the "Shared Data" type. I do not think you can make it safe if it is a struct.

KrzysztofCwalina: Does it need to be safe? We could make it an unsafe type that internals of the pipeline use very carefully. Then user code always gets spafe Spans.  
To me the reason for this type to be a struct is that the main usage will be to embed it in some "pipeline context" and it will just cost additional allocation without much improved safety.

jkotas: If it is a internal pipeline type, then it is fine. It should be called appropriately then, like `UnsafeMemory<T>`.

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,132 @@
...
+        {
+            get
+            {
+                if (_memory == null)
+                {
+                    throw new InvalidOperationException("The native pointer isn't available because the memory isn't pinned");
+                }
+
+                return (byte*)_memory + (UnsafeUtilities.SizeOf<T>() * _offset);
```
mgravell: int32 overflow?

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,132 @@
...
+        }
+
+        public int Length => _memoryLength;
+
+        public unsafe Memory<T> Slice(int offset, int length)
+        {
+            // TODO: Bounds check
+            if (_array == null)
+            {
+                return new Memory<T>((byte*)_memory + (UnsafeUtilities.SizeOf<T>() * offset), length);
```
mgravell: int32 overflow?

src/System.Slices/System/Buffers/Memory.cs
mgravell: @KrzysztofCwalina because there are scenarios where you want it to be available later as *either* an array and pointer; we've  seen this in Channels where some external APIs work with pointers, and some work with arrays. Neither by itself is *completely* useful. Now, you might say: "tough; you need to find another way to abstract  that", but...

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: @davidfowl, do you prefer Memory to be safe or cheap?

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: @mgravell, could you point me to code that depends on span being either pointer to array?

davidfowl: The default pool implementation in channels allocates a pre-pinned managed array. https://github.com/davidfowl/Channels/blob/master/src/Channels/MemoryPoolBlock.cs#L26

The same pool can be used for the managed socket API: https://github.com/davidfowl/Channels/blob/master/src/Channels.Networking.Sockets/SocketConnection.cs#L317

OR the libuv API without tweaks:
https://github.com/davidfowl/Channels/blob/master/src/Channels.Networking.Libuv/UvTcpConnection.cs#L199-L200

Now you could argue those 2 systems should just use different pools but if the user did all the hard work to pre-pinn the managed buffer passed in the both the unsafe pointer and the underlying array should be accessible.

src/System.Slices/System/Buffers/Memory.cs
davidfowl: Property looks better at the consumption places. What about `.Data`? Slice is an interesting method that we should also talk about

src/System.Slices/System/Buffers/Memory.cs
davidfowl: Yea, after looking at the API I think TryGetPointer makes more sense. Just means all the places need to throw their own exception 😄

src/System.Slices/System/Buffers/Memory.cs
davidfowl: Cheap because I was replacing code that was slicing up spans before so this ended up being the alternative. I need to allocate say a 4K chunk, have pieces of code write into it then get the underlying pointer (lets call that unwrapping) so I can pass it to a native API like libuv for outgoing writes. Maybe the pattern should be to expose a `Span<byte>` for reading an writing but still provide access to `Memory<byte>` just for unwrapping. 

I'd be OK making it a class if there was only one of them and they were pooled.

jkotas: > they were pooled for reuse

Just so I understand ... are you on plan to pool all per-request allocation? I think it is ok for prototyping to test the boundaries.

davidfowl: All buffers are pooled in this model. The buffer pool itself returns a Memory<T> which is pooled for reuse across requests.

KrzysztofCwalina: @davidfowl, "cheap" because if it's a struct you can store a hundred of them in a single object (array). If it's a class, you need 101 objects.

jkotas: Number of objects does not matter. The total size is what matters. There is still difference ... but the numbers are different; there are also other factors like cost of passing these around ... .

davidfowl: @jkotas fair but if `Memory<T>` and `Span<T>` map 1-1 right now so my naive POV is that it needs to be as cheap as a `Span<T>` or the usage pattern needs to be different. I'm having a hard time coming with an alternative plan way to avoid slicing the `Memory<T>` to show a sliding window of the full memory that is currently writable.

KrzysztofCwalina: Why don't we start with a unsafe struct and see how far we can go. If it does not work, we change it to a class.

jkotas: >  start with a unsafe struct 

I agree that it is a fine start for prototyping.

samples/LibuvWithNonAllocatingFormatters/Program.cs
```diff
@@ -89,9 +90,15 @@ static void RunLoop(bool log)
                         formatter.Format(" @ {0:O}", DateTime.UtcNow);
                     }
 
-                    var response = formatter.Buffer.Slice(0, formatter.CommitedByteCount); // formatter should have a property for written bytes
-
-                    connection.TryWrite(response);
+                    unsafe
+                    {
+                        fixed (byte* ptr = formatter.Buffer)
```
davidfowl: We could pass the pointer directly here but I just wanted to reduce the number of changes.

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -0,0 +1,147 @@
...
+                throw new ArgumentNullException(nameof(array));
+            }
+
+            unsafe
+            {
+                if (isPinned)
+                {
+                    // TODO: Use new Unsafe.AsPointer(ref array[0])
+                    // The caller pinned the array so we can safely get the pointer
+                    _memory = Unsafe.AsPointer(ref array[0]);
```
davidfowl: First use of unsafe here, though the comment is out of date.

davidfowl:
updated!

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: So it seems like it's a point in time issue, i.e. if the socket APIs accepter byte*, you would not need to get the array from the span/memory. You would just always use the pointer.

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: Do we even need the property? Can we start with the cast operator?

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -33,18 +34,24 @@ public unsafe Memory(void* pointer, int length)
...
         {
+            if (array == null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+
             unsafe
             {
                 if (isPinned)
                 {
```
ViIvanov: May be, it is better to use `unsafe` inside `if`/`else`? Like
```cs
if (isPinned)
{
    unsafe
    {
      // …
    }
}
else
{
    unsafe
    {
      // …
    }
}
```

src/System.Slices/System/Buffers/Memory.cs
davidfowl: We need this property, casting is ugly. I started without it and added it because of the uglyness. Extension methods are one of the big reasons you want this, also it's unclear when code starts to do this:

```C#
Span<byte> span = Memory;
```

All of the variable declarations go from `var` blah to `Span<T>` to force implicit conversions to work:

```C#
var span = Memory.Span;
```

davidfowl: See https://github.com/StackExchange/Channels.WebSockets/commit/f4598323b243129321ec94b834fdbca42be239f0

src/System.Slices/System/Buffers/Memory.cs
mgravell: See `UnsafeUtilities.GetElementAddress` in https://github.com/dotnet/corefxlab/pull/849

src/System.Slices/System/Buffers/Memory.cs
mgravell: See `UnsafeUtilities.GetElementAddress` in https://github.com/dotnet/corefxlab/pull/849

src/System.Slices/System/Buffers/Memory.cs
davidfowl: Agreed, but they don't today and this allows us to interop seamlessly without making `Span<T>` the messy type that lets you "peak behind the curtain" to see how it was constructed.

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: @davidfowl, it seems like a good suggestion

davidfowl: Yea, I did it in the latest changes.

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: Can we add a ctor that accepts both an array and a pointer, which would indicate that the array is pinned?

davidfowl: I used to have that at some point but this was cleaner. I can make that change though. /cc @mgravell

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: In this case, I still thing parameterless Slice() method is the best compromise. For example, we already use array.Slice() to create full span over an array.

src/System.Slices/System/Buffers/Memory.cs
benaadams: Taking a pointer forces the caller to also be unsafe; which is probably good

src/System.Slices/System/Buffers/Memory.cs
davidfowl: Slice with arguments returns another `Memory<T>` so there's inconsistency.

KrzysztofCwalina: Ah, that is indeed a problem.

KrzysztofCwalina: I am fine with Span property, until we come up with something better.

src/System.Slices/System/Buffers/Memory.cs
davidfowl: One more question, should memory point to the `array[0]`? Or `array[offset]`? Should the caller do that or should it be done by `Memory<T>`?

src/System.Slices/System/Buffers/Memory.cs
KrzysztofCwalina: SO, are you going to change it to TryGetPointer?

src/System.Slices/System/Buffers/Memory.cs
benaadams: if you don't do the ptr .ctor could do:

`_memory = Unsafe.AsPointer(ref array[offset]);`

src/System.Slices/System/Buffers/Memory.cs
davidfowl: Did already. The problem with these github reviews is that the comments don't update properly once the changes are made.

src/System.Slices/System/Buffers/Memory.cs
benaadams: Could do the current but with offset and use the pointer to confirm e.g.
```csharp
public unsafe Memory(T[] array, int offset, int length, void* pointer)
{
    if (array == null)
	{
	    throw new ArgumentNullException(nameof(array));
	}
	if (pointer != null)
	{
        _memory = Unsafe.AsPointer(ref array[offset]);
		
		if (_memory != pointer)
		{
		    throw new ArgumentException(nameof(pointer));
		}
	}
	// ....
}
```

davidfowl: Adding a bit more tests based on this

davidfowl:
Updated again

src/System.Slices/System/Buffers/Memory.cs
davidfowl: Upated

KrzysztofCwalina:
Looks good. Thanks to all!

PR #845, opened 2016-09-27
Refactored IFormatter based on discussions with the channels team

KrzysztofCwalina:
I split IFormatter into IOutput and ITextOutput.
IOutput represents a general output stream. ITextOutput represents an output stream of text.

I tried to completely get rid of ITextOutput, but then ran into a significant problem: the "formatters", e.g. StringFormatter, ResponseFormatter were not in control of the encoding. The code calling Append could write partial UTF8 and partial UTF16 to lets say response buffers, and the code managing the buffers would not be able to prevent it. 

Also, I tried to rename Enlarge to Ensure. The problem is that Ensure does not make much sense with the default parameter, yet the default parameter is very useful as the calling code often does not know how much space it needs (it just knows it needs more).

### DIFF
--- a/src/System.Slices/System/Buffers/IOutput.cs
+++ b/src/System.Slices/System/Buffers/IOutput.cs
@@ -1,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Buffers
+{
+    public interface IOutput
+    {
+        Span<byte> Buffer { get; }
+        void Advance(int bytes);
+
+        /// <summary>desiredBufferLength == 0 means "i don't care"</summary>
+        void Enlarge(int desiredBufferLength = 0);
+    }
+}
--- a/samples/LowAllocationWebServer/Framework/SharedData.cs
+++ b/samples/LowAllocationWebServer/Framework/SharedData.cs
@@ -191,7 +191,7 @@ namespace Microsoft.Net.Http
         public ResponseFormatter Headers { get { return new ResponseFormatter(_data, 2); } }
     }
 
-    public struct ResponseFormatter : IFormatter
+    public struct ResponseFormatter : ITextOutput
     {
         int _order;
         SharedData _data;
@@ -212,7 +212,7 @@ namespace Microsoft.Net.Http
             }
         }
 
-        public Span<byte> FreeBuffer
+        public Span<byte> Buffer
         {
             get
             {
@@ -221,14 +221,14 @@ namespace Microsoft.Net.Http
             }
         }
 
-        public void CommitBytes(int bytes)
+        public void Advance(int bytes)
         {
             _data.Commit(_order, bytes);
         }
 
-        public void ResizeBuffer(int desiredFreeBytesHint = -1)
+        public void Enlarge(int desiredBufferLength = 0)
         {
-            _data.Allocate(desiredFreeBytesHint == -1 ? 1024 : desiredFreeBytesHint, _order);
+            _data.Allocate(desiredBufferLength == 0 ? 1024 : desiredBufferLength, _order);
         }
     }
 }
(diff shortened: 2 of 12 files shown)


KrzysztofCwalina:
@davidfowl

src/System.Slices/System/Buffers/IOutput.cs
```diff
@@ -0,0 +1,14 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Buffers
+{
+    public interface IOutput
+    {
+        Span<byte> Buffer { get; }
```
davidfowl: `Span<T>` or `Memory<T>`?

src/System.Slices/System/Buffers/IOutput.cs
```diff
@@ -0,0 +1,14 @@
...
+        void Advance(int bytes);
+
+        /// <summary>desiredBufferLength == 0 means "i don't care"</summary>
+        void Enlarge(int desiredBufferLength = 0);
```
davidfowl: > Also, I tried to rename Enlarge to Ensure. The problem is that Ensure does not make much sense with the default parameter, yet the default parameter is very useful as the calling code often does not know how much space it needs (it just knows it needs more).


While I agree enlarge doesn't express "do nothing if I already have a big enough buffer".

src/System.Slices/System/Buffers/IOutput.cs
davidfowl: After thinking about it, this should be `Span<T>`

src/System.Slices/System/Buffers/IOutput.cs
KrzysztofCwalina: If you have enough, don't call Enlarge, i.e.
if(out.Buffer.Length < needed) out.Enlarge(needed);
out.Buffer.Set(data);

src/System.Slices/System/Buffers/IOutput.cs
KrzysztofCwalina: In general we should default to Span<T> (safer) and only use Memory<T> where Span<T> is not enough.

davidfowl:
Just so we're clear, this would be a valid pattern right?:

```C#
public static void Write(this IOutput output, Span<byte> source)
{
    if (output.Buffer.IsEmpty)
    {
        output.Enlarge();
    }

    // Fast path, try copying to the available memory directly
    if (source.TryCopyTo(output.Memory))
    {
        output.Advance(source.Length);
        return;
    }

    var remaining = source.Length;
    var offset = 0;

    while (remaining > 0)
    {
        var writable = Math.Min(remaining, output.Buffer.Length);

        output.Ensure(writable);

        if (writable == 0)
        {
            continue;
        }

        source.Slice(offset, writable).TryCopyTo(output.Buffer);

        remaining -= writable;
        offset += writable;

        output.Advance(writable);
    }
}
```

KrzysztofCwalina:
I am not sure what you specifically want the pattern to allow, but I think it will work with a small modification: Currently, the Resize's parameter is a hint. If the IOutput cannot resize to the desired size, it still tries to do the best it can to enlarge. And so your code would have to Resize, check the length of the new span, and then TryCopyTo as much as the new length is, not as much as it passed to Resize.

We can of course change the contract and demand that Resize's parameter is the minimum.

davidfowl:
> If the IOutput cannot resize to the desired size, it still tries to do the best it can to enlarge. 

It needs to fail if the desired size > what it can do. For e.g. in the default pool in channels, if you ask for more than 4K it'll explode. You can keep asking it to grow (between 0-4K, or by passing 0) and keep writing through. Which is what the above is doing.

KrzysztofCwalina:
OK. We can try this. I will merge and try to update the code to these semantics.

Issue #848, opened 2016-09-28
Alignment and 2 GiB overflow issues in Span<T>.TryCopyTo


mgravell:
Only affects the `IsPrimitiveValueType<T>` case;

- overflow for large `src.Length` and non-trivial `T`; might need to loop, since `cpblk` is limited to `uint` sizes (thus is not quite the same issue as #846)
- `CopyBlock` assumes that the pointers are aligned; otherwise, `cpblk` needs the `unaligned` prefix

(also impacts `ReadOnlySpan<T>`)

benaadams:
Raised issue for unaligned https://github.com/dotnet/corefx/issues/12243

PR #849, opened 2016-09-28
Fix multiple 2GiB overflow issues in Span<T>; fixes #846

### LINKED ISSUE #846: Extensive 2 GiB overflow problems in Span<T>
Basically, anywhere that uses `* UnsafeUtilities.SizeOf<T>()` is probably wrong - at least when it comes to huge spans and/or non-trivial `T`. Example:

    [Fact]
    public unsafe void SpanSplitShouldWorkForHugeSpans()
    {
        const long ThreeGiB = 3L * 1024L * 1024L * 1024L,
            TwoGiB = 2L * 1024L * 1024L * 1024L;
        var huge = Marshal.AllocHGlobal(new IntPtr(ThreeGiB));
        try
        {
            int totalItemCount = (int)(ThreeGiB / sizeof(Guid));
            int ovfLimit = (int)(TwoGiB / sizeof(Guid));
            int i = ovfLimit + 4;
            var span = new Span<Guid>((void*)huge, totalItemCount);
            var slice = span.Slice(i); // boom!
            slice[0] = Guid.NewGuid();
        }
        finally
        {
            Marshal.FreeHGlobal(huge);
        }
    }

I started looking at this after spotting an ovf in the `Memory<T>` code; came up with the following that fixes the *above* scenario:

        /// <summary>
        /// Computes the address of an element relative to a pointer -
        /// essentially returning simply: ((T*)root) + offset;
        /// 
        /// This cannot be expressed in C#, and the naive approach
        /// leads to int32 overflow problems; the "fix" to the naive
        /// approach (ulong) leads to suboptimal x86 performance.
        /// </summary>
        [ILSub(@"
            .maxstack 8
            ldarg.0 // load the base pointer
            ldarg.1 // load the offset
            conv.i  // convert the offset to a native integer
            sizeof !!T  // load size of T
            mul // multiply size by offset
            add // add product to base pointer
            ret // return")]
        internal static unsafe void* GetElementAddress<T>(void* root, int offset)
        {
            // slow version used for illustration only
            return ((byte*)root) + ((ulong)offset * (ulong)SizeOf<T>());
        }

(switching into `Slice`)

However, when I came to put a PR together I realized just how big the problem actually is. Thought I'd better flag it here rather than change a vast range of files.

Probably 20 areas impacted according to code search.

mgravell:
Note I've left  `TryCopyTo` as that needs some additional input rather than just fixing; raised as #848

### DIFF
--- a/src/System.Slices/System/MemoryUtils.cs
+++ b/src/System.Slices/System/MemoryUtils.cs
@@ -33,10 +33,10 @@ namespace System
                     byte* secondPointer = (byte*)UnsafeUtilities.ComputeAddress(second.Object, second.Offset).ToPointer();
 
                     int step = sizeof(void*) * 5;
-                    int totalBytesCount = first.Length * UnsafeUtilities.SizeOf<T>();
-                    byte* firstPointerLimit = firstPointer + (totalBytesCount - step);
+                    ulong totalBytesCount = (ulong)first.Length * (ulong)UnsafeUtilities.SizeOf<T>();
+                    byte* firstPointerLimit = firstPointer + (totalBytesCount - (ulong)step);
 
-                    if (totalBytesCount > step)
+                    if (totalBytesCount > (ulong)step)
                     {
                         while (firstPointer < firstPointerLimit)
                         {   // IMPORTANT: in order to get HUGE benefits of loop unrolling on x86 we use break instead of return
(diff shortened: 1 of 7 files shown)


KrzysztofCwalina:
looks good. thanks!

PR #851, opened 2016-09-28
Moved ISequence out of System.Buffers.Experimental

KrzysztofCwalina:
This is so we can implement APIs operating on sequences without a dependency on buffer pools.


davidfowl:
Are you going to add spanenumerator?

KrzysztofCwalina:
@davidfowl, yes that's the plan.

PR #852, opened 2016-09-29
Added SpanSequence

KrzysztofCwalina:
I am not sure if I am happy with the interface.
It is quite complex to implement. 
I had to change it from the TryGetNext design, as span enumerator cannot store Span<T> between calls to MoveNext and Current.

### DIFF
--- a/samples/LibuvWithNonAllocatingFormatters/Program.cs
+++ b/samples/LibuvWithNonAllocatingFormatters/Program.cs
@@ -90,15 +90,14 @@ namespace LibuvWithNonAllocatingFormatters
                         formatter.Format(" @ {0:O}", DateTime.UtcNow);
                     }
 
-                    unsafe
-                    {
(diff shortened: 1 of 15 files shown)


samples/LibuvWithNonAllocatingFormatters/Program.cs
```diff
@@ -90,15 +90,14 @@ static void RunLoop(bool log)
...
                     }
 
-                    unsafe
-                    {
-                        fixed (byte* ptr = formatter.Buffer)
-                        {
-                            // formatter should have a property for written bytes
-                            var response = new Memory<byte>(formatter.Buffer, 0, formatter.CommitedByteCount, pointer: ptr);
+                    var segment = formatter.Written;
+                    unsafe {
```
davidfowl: The formatting looks broken here 😉

PR #853, opened 2016-09-29
Fix Memory<T>.Empty

davidfowl:
- Turns out it didn't work after I added the null check
- Added tests

Found this when using it on Channels


KrzysztofCwalina:
looks good

PR #856, opened 2016-09-29
Simplified Multispan

KrzysztofCwalina:
Plus some small other cleanup


KrzysztofCwalina:
@davidfowl, BTW take a look at SequenceFormatter. It shows hot to format to multiple spans without leaving empty ends in some of the spans. There are probably some holes/bugs in the code, but we could experiment with this in the WritableBufferFormatter.

PR #861, opened 2016-10-03
Added LINQ-like operations for span sequences.

KrzysztofCwalina:


src/System.Buffers.Experimental/System/Buffers/SequenceExtensions.cs
```diff
@@ -1,16 +1,52 @@
...
+        public static void Do<T>(this ISpanSequence<T> buffer, ActionOfSpan<T> action)
+        {
+            Span<T> flat;
+            if (buffer.Count < 2) flat = buffer.Flatten();
+            if (buffer.TotalLength < 128) {
+                unsafe
+                {
+                    var stackArray = stackalloc byte[128];
+                    flat = new Span<T>(stackArray, 128);
+                }
```
svick: Shouldn't this also copy the data from `buffer` into `flat`, probably by using `TryCopyTo` like the other overload?

KrzysztofCwalina: Ah, good catch. Fixed in #862

Issue #860, opened 2016-10-03
Port ASP.NET Core's UTF8 url unescaper


troydai:
Port my UTF8 path unescape decoder from ASP.NET core. 

The project has been initialized in https://github.com/dotnet/corefxlab/commit/db532ee4431a6795c28b89b924998e3865516d8c

source: https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/UrlPathDecoder.cs

working branch: [troy-860](https://github.com/dotnet/corefxlab/tree/troy-860)

troydai:
@davidfowl please help to properly label this issue.

KrzysztofCwalina:
Hmmm, we should coordinate. I know @SedarG started to work on this. I thought you guys just checked in the skeleton of the APIs. 

If you still want to work on it, I will talk to @SedarG about developing the escape (encoder) counterparts.

troydai:
@davidfowl I'll assign this to you as I've left the team.

davidfowl:
This is done

PR #863, opened 2016-10-04
Add Url Unescape 

troydai:
Issue: #860

### DIFF
--- a/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
+++ b/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
@@ -1,19 +1,330 @@
-﻿using System;
-using System.Collections.Generic;
-using System.Threading.Tasks;
-
-namespace System.Text.Encodings.Web.Utf8
+﻿namespace System.Text.Encodings.Web.Utf8
 {
     public class UrlEncoder
     {
-        public static void UnescapeInPlace(Span<byte> data)
+        /// <summary>
+        /// Unescape a URL path
+        /// </summary>
+        /// <param name="source">The byte span represents a UTF8 encoding url path.</param>
+        /// <param name="destination">The byte span where unescaped url path is copied to.</param>
+        /// <returns>The length of the byte sequence of the unescaped url path.</returns>
+        public static int Decode(ReadOnlySpan<byte> source, Span<byte> destination)
         {
+            if (destination.Length < source.Length)
+            {
+                throw new ArgumentException(
+                    "Lenghth of the destination byte span is less then the source.",
+                    nameof(destination));
+            }
 
+            // This requires the destination span to be larger or equal to source span
+            source.CopyTo(destination);
+            return DecodeInPlace(destination);
         }
 
-        public static void Unescape(Span<byte> source, Span<byte> destination)
+        /// <summary>
+        /// Unescape a URL path in place.
+        /// </summary>
+        /// <param name="buffer">The byte span represents a UTF8 encoding url path.</param>
+        /// <returns>The number of the bytes representing the result.</returns>
+        /// <remarks>
+        /// The unescape is done in place, which means after decoding the result is the subset of 
+        /// the input span.
+        /// </remarks>
+        public static int DecodeInPlace(Span<byte> buffer)
         {
+            // the slot to read the input
+            var sourceIndex = 0;
 
+            // the slot to write the unescaped byte
+            var destinationIndex = 0;
+
+            while (true)
+            {
+                if (sourceIndex == buffer.Length)
+                {
+                    break;
+                }
+
+                if (buffer[sourceIndex] == '%')
+                {
+                    var decodeIndex = sourceIndex;
+
+                    // If decoding process succeeds, the writer iterator will be moved
+                    // to the next write-ready location. On the other hand if the scanned
+                    // percent-encodings cannot be interpreted as sequence of UTF-8 octets,
+                    // these bytes should be copied to output as is. 
+                    // The decodeReader iterator is always moved to the first byte not yet 
+                    // be scanned after the process. A failed decoding means the chars
+                    // between the reader and decodeReader can be copied to output untouched. 
+                    if (!DecodeCore(ref decodeIndex, ref destinationIndex, buffer))
+                    {
+                        Copy(sourceIndex, decodeIndex, ref destinationIndex, buffer);
+                    }
+
+                    sourceIndex = decodeIndex;
+                }
+                else
+                {
+                    buffer[destinationIndex++] = buffer[sourceIndex++];
+                }
+            }
+
+            return destinationIndex;
+        }
+
+        /// <summary>
+        /// Unescape the percent-encodings
+        /// </summary>
+        /// <param name="sourceIndex">The iterator point to the first % char</param>
+        /// <param name="destinationIndex">The place to write to</param>
+        /// <param name="end">The end of the buffer</param>
+        /// <param name="buffer">The byte array</param>
+        private static bool DecodeCore(ref int sourceIndex, ref int destinationIndex, Span<byte> buffer)
+        {
+            // preserves the original head. if the percent-encodings cannot be interpreted as sequence of UTF-8 octets,
+            // bytes from this till the last scanned one will be copied to the memory pointed by writer.
+            var byte1 = UnescapePercentEncoding(ref sourceIndex, buffer);
+            if (byte1 == -1)
+            {
+                return false;
+            }
+
+            if (byte1 == 0)
+            {
+                throw new InvalidOperationException("The path contains null characters.");
+            }
+
+            if (byte1 <= 0x7F)
+            {
+                // first byte < U+007f, it is a single byte ASCII
+                buffer[destinationIndex++] = (byte)byte1;
+                return true;
+            }
+
+            int byte2 = 0, byte3 = 0, byte4 = 0;
+
+            // anticipate more bytes
+            var currentDecodeBits = 0;
+            var byteCount = 1;
+            var expectValueMin = 0;
+            if ((byte1 & 0xE0) == 0xC0)
+            {
+                // 110x xxxx, expect one more byte
+                currentDecodeBits = byte1 & 0x1F;
+                byteCount = 2;
+                expectValueMin = 0x80;
+            }
+            else if ((byte1 & 0xF0) == 0xE0)
+            {
+                // 1110 xxxx, expect two more bytes
+                currentDecodeBits = byte1 & 0x0F;
+                byteCount = 3;
+                expectValueMin = 0x800;
+            }
+            else if ((byte1 & 0xF8) == 0xF0)
+            {
+                // 1111 0xxx, expect three more bytes
+                currentDecodeBits = byte1 & 0x07;
+                byteCount = 4;
+                expectValueMin = 0x10000;
+            }
+            else
+            {
+                // invalid first byte
+                return false;
+            }
+
+            var remainingBytes = byteCount - 1;
+            while (remainingBytes > 0)
+            {
+                // read following three chars
+                if (sourceIndex == buffer.Length)
+                {
+                    return false;
+                }
+
+                var nextSourceIndex = sourceIndex;
+                var nextByte = UnescapePercentEncoding(ref nextSourceIndex, buffer);
+                if (nextByte == -1)
+                {
+                    return false;
+                }
+
+                if ((nextByte & 0xC0) != 0x80)
+                {
+                    // the follow up byte is not in form of 10xx xxxx
+                    return false;
+                }
+
+                currentDecodeBits = (currentDecodeBits << 6) | (nextByte & 0x3F);
+                remainingBytes--;
+
+                if (remainingBytes == 1 && currentDecodeBits >= 0x360 && currentDecodeBits <= 0x37F)
+                {
+                    // this is going to end up in the range of 0xD800-0xDFFF UTF-16 surrogates that
+                    // are not allowed in UTF-8;
+                    return false;
+                }
+
+                if (remainingBytes == 2 && currentDecodeBits >= 0x110)
+                {
+                    // this is going to be out of the upper Unicode bound 0x10FFFF.
+                    return false;
+                }
+
+                sourceIndex = nextSourceIndex;
+                if (byteCount - remainingBytes == 2)
+                {
+                    byte2 = nextByte;
+                }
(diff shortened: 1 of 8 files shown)


src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+                    var decodeReader = reader;
+
+                    // If decoding process succeeds, the writer iterator will be moved
+                    // to the next write-ready location. On the other hand if the scanned
+                    // percent-encodings cannot be interpreted as sequence of UTF-8 octets,
+                    // these bytes should be copied to output as is. 
+                    // The decodeReader iterator is always moved to the first byte not yet 
+                    // be scanned after the process. A failed decoding means the chars
+                    // between the reader and decodeReader can be copied to output untouched. 
+                    if (!DecodeCore(ref decodeReader, ref writer, end, data))
```
KrzysztofCwalina: Why do you pass the "end" here? This information is already passed in data.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+        /// </summary>
+        /// <param name="reader">The iterator point to the first % char</param>
+        /// <param name="writer">The place to write to</param>
+        /// <param name="end">The end of the buffer</param>
+        /// <param name="buffer">The byte array</param>
+        private static bool DecodeCore(ref int reader, ref int writer, int end, Span<byte> buffer)
         {
+            // preserves the original head. if the percent-encodings cannot be interpreted as sequence of UTF-8 octets,
+            // bytes from this till the last scanned one will be copied to the memory pointed by writer.
+            var byte1 = UnescapePercentEncoding(ref reader, end, buffer);
```
KrzysztofCwalina: so this returns nullable byte. Do we have to do it? Can the calling code check the length of the buffer before calling this ? Would it be faster? I am asking because we need to be super careful with all these routines operating on single bytes. They can become quite expensive.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+            var remainingBytes = byteCount - 1;
+            while (remainingBytes > 0)
+            {
+                // read following three chars
+                if (reader == buffer.Length)
+                {
+                    return false;
+                }
+
+                var nextItr = reader;
```
KrzysztofCwalina: Could you name it without using the shortening?

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+            }
+
+            if (currentDecodeBits < expectValueMin)
+            {
+                // overlong encoding (e.g. using 2 bytes to encode something that only needed 1).
+                return false;
+            }
+
+            // all bytes are verified, write to the output
+            if (byteCount > 0)
```
KrzysztofCwalina: It might be faster to write an int into the span. Otherwise you might be doing lots of range checks.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+        /// will be removed beyond the location of % and -1 will be returned.
+        /// 
+        /// If the following two characters can't be successfully unescaped the 
+        /// <paramref name="scan"/> iterator will be move behind the % and -1 
+        /// will be returned.
+        /// </summary>
+        /// <param name="scan">The value to read</param>
+        /// <param name="end">The end of the buffer</param>
+        /// <param name="buffer">The byte array</param>
+        /// <returns>The unescaped byte if success. Otherwise return -1.</returns>
```
KrzysztofCwalina: What about null?

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+            {
+                return false;
+            }
+
+            if (byte1 == 0)
+            {
+                throw new InvalidOperationException("The path contains null characters.");
+            }
+
+            if (byte1 <= 0x7F)
```
KrzysztofCwalina: It seems like UnescapePercentEncoding can return -1. What happens then?

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+        /// <param name="data">The byte span represents a UTF8 encoding url path.</param>
+        /// <returns>
+        /// The length of the unescape sequence on the input byte span. The sequence begins
+        /// from the head of original byte span.
+        /// </returns>
+        /// <remarks>
+        /// The unescape is done in place, which means the result span is the subset of 
+        /// the input span. The beginnings of the spans are the same.
+        /// </remarks>
+        public static int UnescapeInPlace(Span<byte> data)
```
KrzysztofCwalina: I think we should call these Decode/Encode. It's consistent with other APIs we already shipped and with some terminology used even within this file.

tests/System.Text.Encodings.Web.Utf8.Tests/UnescapeUrlPathInPlaceTests.cs
```diff
@@ -0,0 +1,18 @@
...
+    public class UnescapeUrlPathInPlaceTests : UrlEncoderTests
+    {
+        protected override void TestCore(string raw, string expected)
+        {
+            var input = GetBytes(raw);
+
+            var len = UrlEncoder.UnescapeInPlace(input);
+            Assert.True(len <= input.Length);
+
+            var outputDecoded = Encoding.UTF8.GetString(input.Slice(0, len).ToArray());
```
KrzysztofCwalina: You should use "new Utf8String(input.Slice(0, len).ToString()"

tests/System.Text.Encodings.Web.Utf8.Tests/UnescapeUrlPathTests.cs
```diff
@@ -0,0 +1,34 @@
+﻿using Xunit;
+
+namespace System.Text.Encodings.Web.Utf8.Tests
+{
+    public class UnescapeUrlPathTests : UrlEncoderTests
+    {
```
KrzysztofCwalina: What's the reason for the base type?

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+        {
+            // the slot to read the input
+            var reader = 0;
```
troydai: sure.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
 {
     public class UrlEncoder
     {
-        public static void UnescapeInPlace(Span<byte> data)
+        /// <summary>
+        /// Unescape a URL path in place.
+        /// </summary>
+        /// <param name="data">The byte span represents a UTF8 encoding url path.</param>
+        /// <returns>
+        /// The length of the unescape sequence on the input byte span. The sequence begins
```
troydai: will update.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: will update.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -1,19 +1,336 @@
...
+            }
+            if (byteCount > 3)
+            {
+                buffer[writer++] = (byte)byte4;
+            }
+
+            return true;
+        }
+
+        private static void Copy(int begin, int end, ref int writer, Span<byte> buffer)
```
troydai: will do.

tests/System.Text.Encodings.Web.Utf8.Tests/UnescapeUrlPathTests.cs
troydai: two sets of tests. once for unescapse in place and the other for unescape and copy to a new byte span. the scenarios are the same but the verification are different. in one it will assert that it indeed an in-place operation, and for the other one it ensure the data is copied to another array.

uses a base class here allows two sets of test cases to reuse the same set of scenarios while offer different test logic.

tests/System.Text.Encodings.Web.Utf8.Tests/UnescapeUrlPathInPlaceTests.cs
troydai: will update.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: i'll revert the nullable and instead return -1 for false cases.

troydai: and the false case is not just about the length of the buffer but may also about the legitimacy of the byte representations follows the `%` therefore `UnescapsePercentEncoding` is responsible of determining if the next characters are acceptable.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: updated. there won't be `null` return any more.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: updated. `-1` case will be return early.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
KrzysztofCwalina: I would just name the first parameter "source" and the only parameter "buffer". This is to be consistent with Base64 encoder we already have.

tests/System.Text.Encodings.Web.Utf8.Tests/UnescapeUrlPathTests.cs
troydai: yup. will add.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: ok. looks good to me. will update.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: does `Span.CopyTo` work for the cases that the destination and source are overlapped?

davidfowl: The CoreCLR version does, but the corefxlab one doesn't https://github.com/dotnet/coreclr/blob/SpanOfT/src/mscorlib/src/System/Span.cs#L352.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: could you elaborate here?

KrzysztofCwalina: Feel free to postpone it (and add a TODO), but every time you write to the span, we do range checks today, and you are writing 4 times. Instead, you could create an int (with these 4 bytes) and use span.Write(int), which is in SpanExtensions to write one int.

troydai: I got it. The tricky part here it does not always write 4 bytes. It may write any thing between 1 to 4 bytes. To fully the logic will be a bit complex:
```
IF 1 byte:
    Write 1 byte
ELSE-IF 2 bytes:
    Compose 1 short
    Write 1 short
ELSE-IF 3 bytes:
    Compose 1 short
    Write 1 short
    Write 1 byte
ELSE-IF 4 bytes:
    Compose 1 int
    Write 1 int
```
The code will be a myth to the maintainer. And I'm not sure about the performance gain given the additional comparison and value manipulation.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
troydai: Will do.

troydai: I'll leave this one behind since the `Span` we uses here won't support overlap.

troydai: I can spend some time on expanding `corefxlab`'s `Span` and then get back to this.

tests/System.Text.Encodings.Web.Utf8.Tests/UnescapeUrlPathInPlaceTests.cs
troydai: Skip for now, the outcome of `Utf8String.ToString` is inconsistent with `Encoding.UTF8.GetString`. It may be an issue in `Utf8String`.

troydai: To be more specific, there are valid UTF8 coded byte span rejected by `Utf8String` under the reason "invalid characters".

For example: `%F0%A4%AD%A2` is decoded to `𤭢` in `Encoding.UTF8` but rejected in `Utf8String`.

troydai:
/to: @KrzysztofCwalina @davidfowl 
Addressed all the feedbacks except:
1. Span.CopyTo is not used because the current implementation cannot deal with overlap memory blocks.
2. Doesn't merge decoded bytes into larger primitive type before writing to avoid range check. I state my reason in the comments.
3. Doesn't use `Utf8String` in test. It rejects a few samples which are acceptable to `Encoding.UTF8`. It may be a bug to `Utf8String`.

KrzysztofCwalina:
@troydai, thanks. looks good for now. I will try to look into the Utf8String issue and CopyTo

KrzysztofCwalina:
@troydai, please resolve the merge conflict. Thanks.

troydai:
Rebase and squash the commits. Waiting for the CI validation.

PR #866, opened 2016-10-05
Enabled Json parser to parse spans

KrzysztofCwalina:
This is a large change. The parser was written to assume empty space at the end of the array containing the JSON payload. But in real scenarios, users often want to parse spans that don't contain any empty space. The change required many updates to allow storing parsed database in a separate buffer (either passed in, or allocated).


KrzysztofCwalina:
cc: @ahsonkhan

PR #870, opened 2016-10-06
Fixed #869

### LINKED ISSUE #869: Utf8String fails on
repro:

```c#            
var bytes = new byte[] { 0xF0, 0xA4, 0xAD, 0xA2 };
var utf8String = new Utf8String(bytes);
var str = "𤭢";
Assert.Equal(str, utf8String.ToString());
```

fails with "invalid character"

KrzysztofCwalina:
The existing code was checking if a code point is a surrogate characters, but it should be checking for whether the code point needs a surrogate characters.

Also the encoder was incorrectly computing the high surrogate value.


KrzysztofCwalina:
@krwq, @troydai

PR #875, opened 2016-10-07
Added ability to resize JsonParser's db

KrzysztofCwalina:
Also, made JsonObject and others disposable to return the scratch buffer to the pool


hammett:
@KrzysztofCwalina I thought Dispose should never ever throw exceptions (?)

KrzysztofCwalina:
Yeah, I would like to avoid it, but I am not sure what to do instead. Seems like not throwing here would be equivalent to swallowing an exception/error.

PR #881, opened 2016-10-13
Remove "torn" out of bounds risk

benaadams:
Order fields `pointer` => `length` => `offset` so any "torn reads" are within the data.

If they are set in a different order then a torn read could point somewhere random.

This assumes struct writes/copies are done in a forward order.

/cc @davidfowl @KrzysztofCwalina @jkotas


jkotas:
Both compilers and hardware can reorder the reads and writes. I do not think that you can guarantee anything by ordering the fields in certain way.

benaadams:
Hmm... true

PR #882, opened 2016-10-13
RFC: Memory struct inheritence

benaadams:
Memory->UnsafeMemory
Add Memory

The structs compose:
```csharp
struct Memory<T>
{
    byte[] Array;
    int Offset;
    int Length;
}

struct UnsafeMemory<T>
{
    void* Pointer;
    Memory<T> Memory;
}
```
In `Memory` the worst that can happen with a torn read is an Exception for a range check failure on the array and the array doesn't need extra protection as its GC handled.

In `UnsafeMemory` a torn read could lead to an out of boundary access and you may desire it to have an extra class reference for protection to detect disposal of the memory.

I think some of the uses of `UnsafeMemory` could be changed to `Memory`

/cc @jkotas @davidfowl @KrzysztofCwalina

### DIFF
--- a/samples/LibuvWithNonAllocatingFormatters/Program.cs
+++ b/samples/LibuvWithNonAllocatingFormatters/Program.cs
@@ -93,7 +93,7 @@ namespace LibuvWithNonAllocatingFormatters
                     var segment = formatter.Formatted;
                     unsafe {
                         fixed (byte* p = segment.Array) {
-                            var response = new Memory<byte>(segment.Array, segment.Offset, segment.Count, pointer: p);
+                            var response = new UnsafeMemory<byte>(segment.Array, segment.Offset, segment.Count, pointer: p);
                             connection.TryWrite(response);
                         }
                     }
--- a/samples/LowAllocationWebServer/Framework/SharedData.cs
+++ b/samples/LowAllocationWebServer/Framework/SharedData.cs
@@ -162,9 +162,9 @@ namespace Microsoft.Net.Http
                 Id = id;
             }
 
-            public Memory<byte> Commited
+            public UnsafeMemory<byte> Commited
             {
-                get { return new Memory<byte>(Array.Array, Array.Offset, Array.Count); }
+                get { return new UnsafeMemory<byte>(Array.Array, Array.Offset, Array.Count); }
             }
 
             public Span<byte> Free
--- a/samples/LowAllocationWebServer/Framework/TcpServer.cs
+++ b/samples/LowAllocationWebServer/Framework/TcpServer.cs
@@ -158,7 +158,7 @@ namespace Microsoft.Net.Sockets
             SocketImports.closesocket(_handle);
         }
 
-        public int Send(Memory<byte> buffer)
+        public int Send(UnsafeMemory<byte> buffer)
         {
             // This can work with Span<byte> because it's synchronous but we need pinning support
             unsafe {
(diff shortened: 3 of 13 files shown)


markrockcity:
System.Slices.Tests.SlicesTests.MemoryCtorSubArrayWithPinnedPtr [FAIL]
23:36:51       System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.
23:36:51       Stack Trace:
23:36:51         D:\j\workspace\Release_prtest58035862\tests\System.Slices.Tests\BasicUnitTests.cs(419,0): at System.Slices.Tests.SlicesTests.MemoryCtorSubArrayWithPinnedPtr()
23:36:51     System.Slices.Tests.SlicesTests.SliceArrayBackedMemory [FAIL]
23:36:51       System.ArgumentOutOfRangeException : Specified argument was out of the range of valid values.
23:36:51       Stack Trace:
23:36:51         D:\j\workspace\Release_prtest58035862\tests\System.Slices.Tests\BasicUnitTests.cs(440,0): at System.Slices.Tests.SlicesTests.SliceArrayBackedMemory()

KrzysztofCwalina:
Why do we need two types: safe and unsafe Memory?

benaadams:
The structs compose: 
```csharp
struct Memory<T>
{
    byte[] Array;
    int Offset;
    int Length;
}

struct UnsafeMemory<T>
{
    void* Pointer;
    Memory<T> Memory;
}
```
In `Memory` the worst that can happen with a torn read is an Exception for a range check failure on the array and the array doesn't need extra protection as its GC handled.

In `UnsafeMemory` a torn read could lead to an out of boundary access and you may desire it to have an extra class reference for protection to detect disposal of the memory.

davidfowl:
This basically defeats the purpose of having a single type to unify T[] and T*.

KrzysztofCwalina:
But it creates a two types with no polymorphic relationship between them. I already feel like we have too many memory-like types in the system (span, buffers, memory, etc.). 
Maybe Memory<T> can be removed and ArraySegment + extensions used?

Also, I would be interested in your thoughts on https://github.com/KrzysztofCwalina/corefxlab/blob/Memory/samples/Memory/Memory/Program.cs
I started to prototype it to solve some of the threading and lifetime issues, i.e. today, it seems like memory instances can be stored for later (on the heap) and mess up buffer pooling.

benaadams:
> This basically defeats the purpose of having a single type to unify T[] and T*.

> But it creates a two types with no polymorphic relationship between them. 

Internally Memory is convertible to UnsafeMemory and externally Span provides this unification as they both Slice to Span.

> Maybe Memory can be removed and ArraySegment + extensions used?

Alas ArraySegment gives you the full array access and it it up to the caller to index correctly; starting at offset and remaining within length. Memory provides the heap based slice which is just a window or segment.

KrzysztofCwalina:
@benaadams, do you still feel this PR has advantages over #902?

benaadams:
@KrzysztofCwalina currently I feel it might have a perf edge; but #902 has many more desirable properties.

KrzysztofCwalina:
Sounds good. Though I have to say I do worry about Memory<T> perf. I started to change JSON parser to use it and I am not sure I like it ;-). Maybe after we have true stack-only types, I convert the APIs back to use spans.

benaadams:
What concerns me currently is to get a Span is non-inlinable virtual dispatch; which would be bad for iterating in small span chunks. Also its quite large 8+4+4+8 = 24 bytes; slimming the id to an int would make it the same size as padding would push it back up. 

Thar's just eyeballing it though; will need to do some tests to see what it actually does. Its a much better starting point though.

PR #885, opened 2016-10-15
sp, and wording

benaadams:


docs/Span.md
```diff
@@ -245,7 +245,7 @@ We need to enable the existing array bounds check optimizations for Span – in
 Span\<T\> will support reinterpret cast conversions to Span\<byte\>. It will also support unsafe casts between arbitrary primitive types. The reason for this limitation is that some processors don’t support efficient unaligned memory access.
 
 ##Platform Support Plans
-We want to ship Span\<T\> as a nugget fat package available for .NET Standard 1.1 and above. Runtimes that support by-ref fields and returns will get the fast two filed span. Other runtimes will get the slower three-field span. 
+We want to ship Span\<T\> as a nuget heritage package available for .NET Standard 1.1 and above. Runtimes that support by-ref fields and returns will get the fast two filed span. Other runtimes will get the slower three-field span.
```
davidfowl: heritage package isn't a thing lol 😄

benaadams: For heritage packages for heritage runtime versions 😉

KrzysztofCwalina:
What are they really called?

davidfowl:
There's no name for this, @benaadams is trying to coin a term 😉

KrzysztofCwalina:
I would call them "adaptive packages"

docs/Span.md
KrzysztofCwalina: @weshaggard, @ericstj, what do we call these packages?

KrzysztofCwalina:
Closing for now, though it would be great to find an official name.
@terrajobst

docs/Span.md
terrajobst: We informally call them fat packages. Publicly, there is no name for that and I'm convinced we need one. The nature of packages allows having multiple assets per platform.

terrajobst:
See comment above

docs/Span.md
benaadams: Ah like a bundle that includes x86, x64 etc all in one; portable but with baggage

PR #901, opened 2016-10-18
Moved some open issues to language support section

KrzysztofCwalina:
@jaredpar, @MadsTorgersen


docs/Span.md
```diff
@@ -290,25 +290,53 @@ Since Span\<T\> will be a stack-only type, it’s not well suited as the general
...
+    ```
+
+3. Slicing syntax
+
+    C# compiler will add slicing syntax, and Memory\<T\>, Span\<T\>, and ReadOnlySpan\<T\> will support it. The details are TBD, but imagine that a[1..5] calls a.Slice(1, 5) or a.Slice(new Range(1, 5)). 
+    ```c#
+    Span<T> span = ...
+    var slice = span[1..5];
+    ```
+
```
cdrnet: I assume this will be compatible to the existing slicing support in F#?

PR #902, opened 2016-10-19
Adding a prototype of safe Memory<T>

KrzysztofCwalina:
Memory2\<T\> is a tearing-safe[er] version of Memory\<T\>. 

It supports revocation, i.e. the underlying buffer (MemoryManager\<T\>) can be disposed, which will "poison" all existing copies of Memory2\<T\>. This makes it buffer and object pooling friendly as it is much harder for buggy code to get away with use-after-free bugs (such bugs will almost always result in ObjectDisposedExceptions).   

Also, it supports "reserving" the buffer, i.e. communicating to MemoryManager\<T\> that Memory2\<T\> instance will need the buffer for longer than the lifetime of the current stack frame. This is useful for example when the buffer will be used for an asynchronous call (e.g. async interop call).

The tests illustrate the concepts in detail.

Note that the APIs are not yet hardened for all the races I would like to harden it for. For now, I want to mainly prototype the the basic semantics we want out of this type. But any feedback on races/threading issues is welcome too.

### DIFF
--- a/src/System.Slices/System/Buffers/Memory2.cs
+++ b/src/System.Slices/System/Buffers/Memory2.cs
@@ -1,0 +1,113 @@
+﻿using System.Buffers;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+namespace System
+{
+    public struct Memory2<T> 
+    {
+        MemoryManager<T> _manager;
+        long _id;
+        int _index;
+        int _length;
+
+        internal Memory2(MemoryManager<T> manager, long id)
+            : this(manager, id, 0, manager.GetSpan(id).Length)
+        { }
+
+        private Memory2(MemoryManager<T> manager, long id, int index, int length)
+        {
+            _manager = manager;
+            _id = id;
+            _index = index;
+            _length = length;
+        }
+
+        public static Memory2<T> Empty => EmptyManager.Shared.Memory;
+
+        public int Length => _length;
+
+        public bool IsEmpty => Length == 0;
+
+        public Memory2<T> Slice(int index)
+        {
+            return new Memory2<T>(_manager, _id, _index + index, _length - index);
+        }
+        public Memory2<T> Slice(int index, int length)
+        {
+            return new Memory2<T>(_manager, _id, _index + index, length);
+        }
+
+        public Span<T> Span => _manager.GetSpan(_id).Slice(_index, _length);
+
+        public DisposableReservation Reserve() => new DisposableReservation(_manager, _id);
+
+        public unsafe bool TryGetPointer(out void* pointer)
+        {
+            if (!_manager.TryGetPointer(_id, out pointer)) {
+                return false;
+            }
+            pointer = Add(pointer, _index);
+            return true;
+        }
+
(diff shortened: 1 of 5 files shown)


KrzysztofCwalina:
@vancem, @jkotas, @davidfowl

src/System.Slices/System/Buffers/Memory2.cs
```diff
@@ -0,0 +1,113 @@
...
+            return new Memory2<T>(_manager, _id, _index + index, _length - index);
+        }
+        public Memory2<T> Slice(int index, int length)
+        {
+            return new Memory2<T>(_manager, _id, _index + index, length);
+        }
+
+        public Span<T> Span => _manager.GetSpan(_id).Slice(_index, _length);
+
+        public DisposableReservation Reservation => new DisposableReservation(_manager, _id);
```
davidfowl: Can we make this a method instead? 

```C#
var result = await channel.ReadAsync();
var buffer  = result.Buffer;

var reservation = buffer.Reserve();

channel.Advance(buffer.End);
```

KrzysztofCwalina: Done.

src/System.Slices/System/Buffers/Memory2.cs
```diff
@@ -0,0 +1,113 @@
...
+            return new Memory2<T>(_manager, _id, _index + index, _length - index);
+        }
+        public Memory2<T> Slice(int index, int length)
+        {
+            return new Memory2<T>(_manager, _id, _index + index, length);
+        }
+
+        public Span<T> Span => _manager.GetSpan(_id).Slice(_index, _length);
+
+        public DisposableReservation Reserve() => new DisposableReservation(_manager, _id);
```
jkotas: This will need to be updated to match the language extensions once they are designed. I would expect that the Span property and Reserve method will become just one thing; and DisposableReservation won't exist at all.

src/System.Slices/System/Buffers/MemoryManagers.cs
```diff
@@ -0,0 +1,137 @@
...
+
+        protected override Span<T> GetSpanCore()
+        {
+            if (_array != null) {
+                return _array;
+            }
+            else {
+                unsafe
+                {
+                    return new Span<T>(_pointer, _array.Length);
```
omariom: accessing Length of null _array

src/System.Slices/System/Buffers/MemoryManager.cs
```diff
@@ -0,0 +1,81 @@
...
+            if (_id != id) throw new ObjectDisposedException(nameof(Memory2<T>));
+            Interlocked.Increment(ref _references);
+        }
+        public void ReleaseReference(long id)
+        {
+            if (_id != id) throw new ObjectDisposedException(nameof(Memory2<T>));
+            Interlocked.Decrement(ref _references);
+        }
+
+        public virtual void Initialize()
```
omariom: Why is it public and virtual?

Issue #907, opened 2016-10-20
BufferPool should be IBufferPool<T>


benaadams:
Its not particularly doing anything as an abstract class

If it was `IBufferPool<T>` then you could have a single buffer pool that handed out multiple types using the same memory.

e.g. 

```csharp
public class  NativeBufferPool 
    : IBufferPool<byte>, IBufferPool<sbyte>, 
      IBufferPool<char>, IBufferPool<short>, IBufferPool<ushort>
```

omariom:
Isn't it job of `Span` to handle different blittable types?

benaadams:
Its a more general type not sure why it needs the Span casts?

Else you'd be doing 
```csharp
var buffer = NativeBufferPool.Rent(length * sizeof(long));
var span = buffer.Span.Cast<byte, long>();
```

Additionally would mean `Memory<T>` is always `Memory<byte>` if you wanted to use any of the built in pooling and you'd always have to cast through span?

benaadams:
e.g. if you are working with `Memory<long>` and `Span<long>` why not use pools that give you them directly?

KrzysztofCwalina:
I worry about the following:
1. How will implementers of the interface decide which constructed types to implement?
2. Will implementing many of these interfaces significantly increase code size, especially for AOT runtimes.
3. It's hard to evolve interfaces (add members after they ship). Is this feature worth the risk that we won't be able to evolve the interface once we ship stable.

benaadams:
@KrzysztofCwalina 

1. Depends what they need; you've created a protected primitive pool that works over multiple memory types; its bigger than Span and Memory over bytes. For Managed it just needs to flow the T to the ArrayPool.

2. Not significantly, for multiple in the example above its just multiplying the length by sizeof in Rent; the actual Rent and Returns can call to a non-generic method. However if only `IBufferPool<byte>` is required that is all that needs to be implemented - its merely opening up the possibilities.

3. have a different hierarchy in https://github.com/dotnet/corefxlab/pull/909#issuecomment-255181177 are only 2 methods Rent and Return (return being an explicit implementation). 

Arguably adding new methods is a new interface and it can either be implemented via interface inheritance where they make sense or an entirely parallel interface where there are not cross cutting concerns (e.g. blocking and async in Streams are two parallel interfaces and auto implementing them on the base class has only been a source of pain other than for MemoryStream) or via extension methods where its a recomposition of the current methods?

davidfowl:
[...]

PR #908, opened 2016-10-20
IBufferPool

benaadams:
Resolves https://github.com/dotnet/corefxlab/issues/907


KrzysztofCwalina:
Please look at the comments from @KrzysztofCwalina and @davidfowl in the linked issue. We both have some concerns about the change. Do we feel like the concerns have been resolved?

shiftylogic:
The change itself looks good. I made the assumption that, given it was submitted as a PR and being code reviewed, that those design issues had been resolved (since they were from back in October).

I don't really see an issue with having something more generic like this (i.e. not fixed to byte). The interface is pretty simple and straightforward. However, we could switch back to a generic abstract class if the interface is a concern.

Also, given that Cast is not portable, this seems like a reasonable thing to do from my perspective. My mind could be changed though.

KrzysztofCwalina:
Yeah, no problem.
But I still think the feature does not pay for itself. I think single implementation of many typed pool is a rare scenario. But changing the abstraction to an interface has non-trivial costs:
a) interfaces are harder to evolve
b) interface dispatch is more expensive than virtual dispatch (in most cases)
c) it's unclear which constructed types should pools implement.

davidfowl:
I'm not a fan of making the pool generic. We already have ArrayPool for that. Also implementing more than one on the same type seems obscure...

benaadams:
I'd completely forgotten about this :)

> a) interfaces are harder to evolve

Inherit and implement the implement the inherited interface or add an extension method. 

What's harder to do is forcibly add methods to a type that doesn't expose them by adding them to the base class. 

With all the async-over-sync and sync-over-async problems by this automatic "upgrade" that the Stream, Database etc classes cause; I'm not really sure that's a bad thing... 

Though pipelines is [`IBufferPool`](https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines/IBufferPool.cs) rather than `BufferPool` anyway...

> b) interface dispatch is more expensive than virtual dispatch (in most cases)

Can use it to do generic dispatch rather than interface dispatch. Create a struct wrapper which also implements the interface over a class with non-virtual methods, and its faster and inlinable unlike virtual dispatch.

Though the proliferation of angle brackets with generics gets a bit sucky.

> c) it's unclear which constructed types should pools implement.

Not sure what you mean? `T : where T struct`?

> We already have ArrayPool for that. 

Unify the pools, have one interface/generic for all; looked at changing them all to generic struct allocators in https://github.com/dotnet/corefxlab/issues/924 so you could choose where your structs memory came form (native, managed, pooled etc)

Pipelines narrows the pool type to `byte`, which is fine and then also collapses the generics, so isn't an upstream problem. Just feel its a missed opportunity by having it as a api that is narrowed to `byte` when it could be much more.

Anyway, not hugely concerned 😄

benaadams:
> Also implementing more than one on the same type seems obscure...

Build it they will come, ImageSharp is looking to use ArrayPool for their `TColor` generic structs https://github.com/JimBobSquarePants/ImageSharp/pull/97#issuecomment-276658359

BufferPool isn't much different than ArrayPool; other than having other allocation methods (e.g. native) and counting methods out of the box. Just thinking of unifying the abstractions, rather than yet another block pool.

KrzysztofCwalina:
BTW, I think generic abstraction makes sense. As @benaadams said, people use arrays/buffers of different types. I just don't think one implementation needs to support many types.

PR #909, opened 2016-10-20
[No Merge] Owned memory perf

benaadams:
Still need to handle `Marshal.Free` in dispose

Bit ugly; but much less virtual dispatch, only in dispose for Empty; Dispose returns to the pool, can't get Memory or Span from the OwnedMemory when in disposed state

/cc @davidfowl @KrzysztofCwalina


benaadams:
```csharp
// Exposed for equality test on pool return
public interface IMemoryOwner
{
}

// Internal reference held in OwnedMemory, passed out as IMemoryOwner
public interface IMemoryDisposer<T> : IMemoryOwner
{
	void Return(OwnedMemory<T> buffer);
}

// Implemented by pool, return implemented explicit as `IMemoryDisposer.Return`
public interface IBufferPool<T> : IMemoryDisposer<T>, IDisposable
{
	OwnedMemory<T> Rent(int minimumBufferSize);
}
```

PR #912, opened 2016-10-21
Made Memory<T>.Span callpath non-virtual

KrzysztofCwalina:


KrzysztofCwalina:
@benaadams, I made access to span non-virtual. Some other methods are still virtual, but they won't be called as often and probably the virtual dispatch cost will be less significant as some of these methods do expensive work.

What do you think about this vs the PR you submitted? It seems like this one gives us the perf without losing the ability to create custom owned memory.

cc: @davidfowl

davidfowl:
Anyway we can measure before making this change? It messes with the abstraction. https://github.com/davidfowl/Channels/compare/davidfowl/react-to-memory-changes#diff-5771d95dc15ca396dd271baf4ea8f567R13. The `OwnedMemory<T>` implementation here is a backing array, offset and length (basically an array segment). I *could* pass null and make the pointer scoped but that feels too hacky. You could also make the Array argument an ArraySegment instead so that this scenario work just work.

KrzysztofCwalina:
@davidfowl, I added offset. I will measure perf tomorrow.

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -13,15 +14,46 @@ internal interface IKnown
...
 
-        static long _nextId = InitializedId + 1;
+        protected T[] _array;
+        protected IntPtr _pointer;
+        protected int _length;
+        protected int _offset;
 
-        const long InitializedId = long.MinValue;
-        
+        protected OwnedMemory(T[] array, int arrayOffset, IntPtr pointer, int length)
```
davidfowl: Just make it ArraySegment<byte> and IntPtr?

benaadams:
Added so tweaks to this in https://github.com/dotnet/corefxlab/pull/915

src/System.Slices/System/Buffers/OwnedMemory.cs
KrzysztofCwalina: But then when you just pass a pointer and length, you will have to create arraysegment. It's wasteful and not very intuitive.

KrzysztofCwalina:
Closing this one. I opened another PR with the same feature, but built on top of new Memory<T> APIs

PR #940, opened 2016-10-27
Made OwnerMemory field setters private

### LINKED ISSUE #937: OwnedMemory properties should be private set
Else you risk a derived class tearing the object

https://github.com/dotnet/corefxlab/blob/master/src/System.Slices/System/Buffers/OwnedMemory.cs#L13-L17

/cc @KrzysztofCwalina

KrzysztofCwalina:
Fixes #937 
@benaadams, thanks for filing the issue.


KrzysztofCwalina:
cc @davidfowl

PR #943, opened 2016-10-28
Moved NativeMemory and PinnedMemory to System.Buffers.Experimental

KrzysztofCwalina:
plus, myget or nuget don't seem to be handling https icons?

### DIFF
--- a/src/System.Slices/System/Buffers/OwnedArray.cs
+++ b/src/System.Slices/System/Buffers/OwnedArray.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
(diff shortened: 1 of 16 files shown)


src/System.Slices/System/Buffers/OwnedArray.cs
```diff
@@ -0,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Buffers
+{
+    public sealed class OwnedArray<T> : OwnedMemory<T>
+    {
```
davidfowl: Can you add an ArraySegment overload?

KrzysztofCwalina: done

PR #945, opened 2016-11-01
Start consolidation and rewriting of System.Text.PrimitiveParser

botaberg:
Introduce class PrimitiveParser2 with intent for it to replace PrimitiveParser. The class right now contains skeleton implementations of the overloads of TryParseBoolean and TryParseUInt32.

PrimitiveParser2.TryParseBoolean and PrimitiveParser2.TryParseUInt32 are the main overloads. The overloads in PrimitiveParser2.InvariantUtf8 and PrimitiveParser2.InvariantUtf16 are wrappers around the main overloads currently, but could provide performance improvements if separately implemented.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_UInt32.cs
+++ b/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_UInt32.cs
@@ -1,0 +1,58 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Text.Utf8;
+
+namespace System.Text
+{
+    public static partial class PrimitiveParser2
+    {
+        public static partial class InvariantUtf16
+        {
+            public unsafe static bool TryParseUInt32(char* text, int length, out uint value)
+            {
+                int consumed;
+                var span = new ReadOnlySpan<byte>(text, length);
+                return PrimitiveParser2.TryParseUInt32(span, out value, out consumed, EncodingData.InvariantUtf16);
+            }
+            public unsafe static bool TryParseUInt32(char* text, int length, out uint value, out int consumed)
+            {
+                var span = new ReadOnlySpan<byte>(text, length);
+                return PrimitiveParser2.TryParseUInt32(span, out value, out consumed, EncodingData.InvariantUtf16);
+            }
+            public static bool TryParseUInt32(ReadOnlySpan<byte> text, out uint value)
+            {
+                int consumed;
+                return PrimitiveParser2.TryParseUInt32(text, out value, out consumed, EncodingData.InvariantUtf16);
+            }
+            public static bool TryParseUInt32(ReadOnlySpan<byte> text, out uint value, out int consumed)
+            {
+                return PrimitiveParser2.TryParseUInt32(text, out value, out consumed, EncodingData.InvariantUtf16);
+            }
+            public static partial class Hex
+            {
+                public unsafe static bool TryParseUInt32(char* text, int length, out uint value)
+                {
+                    int consumed;
+                    var span = new ReadOnlySpan<byte>(text, length);
+                    return PrimitiveParser2.TryParseUInt32(span, out value, out consumed, EncodingData.InvariantUtf16, 'X');
+                }
+                public unsafe static bool TryParseUInt32(char* text, int length, out uint value, out int consumed)
+                {
+                    var span = new ReadOnlySpan<byte>(text, length);
+                    return PrimitiveParser2.TryParseUInt32(span, out value, out consumed, EncodingData.InvariantUtf16, 'X');
+                }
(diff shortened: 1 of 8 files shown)


botaberg:
@KrzysztofCwalina @terrajobst PTAL

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_UInt32.cs
```diff
@@ -0,0 +1,58 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Text.Utf8;
+
+namespace System.Text
+{
+    public static partial class PrimitiveParser2
```
KrzysztofCwalina: I know I sent you a sample with PrimitiveParser2, but it's because I forgot to remove the "2". I think we should just add these methods as overloads to the existing class.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_UInt32.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                var span = new ReadOnlySpan<byte>(text, length);
+                return PrimitiveParser2.TryParseUInt32(span, out value, out consumed, EncodingData.InvariantUtf16);
+            }
+            public static bool TryParseUInt32(ReadOnlySpan<byte> text, out uint value)
+            {
+                int consumed;
+                return PrimitiveParser2.TryParseUInt32(text, out value, out consumed, EncodingData.InvariantUtf16);
+            }
+            public static bool TryParseUInt32(ReadOnlySpan<byte> text, out uint value, out int consumed)
```
KrzysztofCwalina: One thing I was wondering is whether we should call this parameter consumed or consumedBytes/Chars. Otherwise it's not super clear if it counts chars or bytes in the UTF16 APIs.

KrzysztofCwalina:
This is a good first PR to get you familiar with CoreFxlab. 
In subsequent PRs, please always include tests as well.

Also, I assume that you will work on efficient implementations next. The implementation you have in this PR defeats the purpose of the perf optimized overlaods.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
```diff
@@ -0,0 +1,58 @@
...
+            public static bool TryParseUInt32(ReadOnlySpan<byte> text, out uint value)
+            {
+                int consumed;
+                return PrimitiveParser2.TryParseUInt32(text, out value, out consumed, EncodingData.InvariantUtf8);
+            }
+            public static bool TryParseUInt32(ReadOnlySpan<byte> text, out uint value, out int consumed)
+            {
+                return PrimitiveParser2.TryParseUInt32(text, out value, out consumed, EncodingData.InvariantUtf8);
+            }
+            public static partial class Hex
```
davidfowl: nit: add a space

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_UInt32.cs
davidfowl: I was wondering about that... 😄

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_UInt32.cs
KrzysztofCwalina: I think the method should consume bytes when parsing byte buffers and consume chars when parsing char buffers, regardless of the encoding.

botaberg: Does this mean there should be overloads that take ReadOnlySpan<char> as an argument? Or are there other ways to explicitly tell the parser that we are consuming a char buffer?

botaberg: I am not as familiar with the use cases of these parsing APIs, but it seems to me that users may want different Utf16 span overloads that take byte and char buffers. If we have to choose one policy, it could depend on which type of buffer is used more often in Utf16 scenarios.

KrzysztofCwalina: Let's go with ROS<char> for Utf16 overloads.

KrzysztofCwalina:
One more thing. After updating an existing PR, you need to squash the commits. In general, a PR should have just one commit (yours have 3)

jkotas:
> After updating an existing PR, you need to squash the commits.

The common practice in other repos is to squash on merge (pick squash and merge from the drop down on the Merge pull request button).

joshfree:
I signed off on the pre-merge signing tool; yes please squash on merge.

PR #947, opened 2016-11-02
Preparing for an experiment with OwnedMemory<T>

KrzysztofCwalina:
I would like to experiment with an OwnedMemory<T> abstraction. This is so we could implement it on T[] and String, and so allow creating Memory<T>/Span<T> out of these without an allocation (for the owned memory wrapper). 

In the past, when we changed OwnedMemory<T>.GetSpan to a virtual method, we saw a significant perf regression in some scenarios. This experiment is using an interface, which because of stub dispatch, might make the performance hit acceptable.

cc: @davidfowl, @jaredpar

### DIFF
--- a/src/System.Slices/System/Buffers/Memory.cs
+++ b/src/System.Slices/System/Buffers/Memory.cs
@@ -14,7 +14,7 @@ namespace System
         int _length;
 
         internal Memory(OwnedMemory<T> owner, long id)
-            : this(owner, id, 0, owner.GetSpan(id).Length)
+            : this(owner, id, 0, owner.GetSpanInternal(id).Length)
         { }
 
         private Memory(OwnedMemory<T> owner, long id, int index, int length)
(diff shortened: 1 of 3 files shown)


davidfowl:
Looks good.

benaadams:
`IKnown` is probably a bad name with this change; unless its a play on `IUnknown` which is also reference counted?

Would the add ref, remove ref calls be ignored on array and string when used directly? (and lifetime just GC controlled)

KrzysztofCwalina:
Yeah, I will rename IKnown if we make it public. This PR makes it public accidentally. I will make it back internal in a subsequent change.

Yeah, for array and string refcounting will be ignored. I am not sure if I like the implication of it for arrays. For strings, it's totally fine as they are immutable.

benaadams:
You can go via OwnedArray though for stronger ref counting? (I assume that interface will be kept)

KrzysztofCwalina:
I cannot use OwnedArray (nor OwnedMemory) for refcounting because it would force DisposableReservation to be a generic type.

benaadams:
Sorry I mean a user could choose to use OwnedArray instead for the stronger control

KrzysztofCwalina:
Ah, correct.

PR #948, opened 2016-11-02
Experimenting with string/array like type supporting Memory<T>

KrzysztofCwalina:
Experimenting with string/array like type (called Utf8String2) supporting Memory<T> without needing to inherit from or be wrapped in OwnedMemory<T>. We would use this to support Memory<T> over strings and T[].

Please don't merge this PR. The changes to Memory<T> have the potiential to regress perf in Channels. We need to run some perf tests first.

cc: @davidfowl,  @jaredpar, @benaadams, @terrajobst, @vancem, @jkotas, @joshfree

### DIFF
--- a/src/System.Slices/System/Buffers/OwnedMemory.cs
+++ b/src/System.Slices/System/Buffers/OwnedMemory.cs
@@ -111,21 +111,25 @@ namespace System.Buffers
         protected virtual void OnReferenceCountChanged(int newReferenceCount)
         { }
 
-        protected internal virtual DisposableReservation Reserve(ref ReadOnlyMemory<T> memory)
+        DisposableReservation IReadOnlyMemory<T>.Reserve(ref ReadOnlyMemory<T> memory)
+        {
+            return ReserveCore(ref memory);
+        }
+
+        protected virtual DisposableReservation ReserveCore(ref ReadOnlyMemory<T> memory)
         {
             return new DisposableReservation(this, Id);
         }
-
         #endregion
 
         #region Used by Memory<T>
-        void IKnown.AddReference(long id)
+        void IReferenceCounted.AddReference(long id)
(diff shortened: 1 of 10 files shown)


src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -177,26 +185,32 @@ Span<T> IMemory<T>.GetSpan(long id)
...
+    public interface IReadOnlyMemory<T> : IReferenceCounted
+    {
+        ReadOnlySpan<T> GetSpan(long id);
+
+        unsafe bool TryGetPointer(long id, out void* pointer);
 
-    internal interface IMemory<T> : IKnown
+        DisposableReservation Reserve(ref ReadOnlyMemory<T> memory);
+    }
+    public interface IMemory<T> : IReadOnlyMemory<T>
```
benaadams: yay :) Nice readonly inheritance 👍 List not being derived from ReadonlyList has always bugged me...

src/System.Text.Utf8/System/Text/Utf8/Utf8String2.cs
```diff
@@ -0,0 +1,144 @@
...
+        public ReadOnlyMemory<byte> Substring(int index, int length)
+        {
+            return Memory.Slice(index, length);
+        }
+        public ReadOnlyMemory<byte> Substring(int index)
+        {
+            return Memory.Slice(index);
+        }
+
+        Span<byte> Buffer => new Span<byte>(this, new UIntPtr(12), 20);
```
davidfowl: @jkotas was right, passing a ref would be so much better...

```C#
Span<byte> Buffer => Span<byte>.DangerousCreate(this, ref _b0, 20);
```

PR #955, opened 2016-11-03
Flesh out implementations of PrimitiveParser.TryParseBoolean overloads

botaberg:
Small changes to the implementation of the TryParseBoolean overloads:
- Moved parsing logic into more specific-case overloads
- More general case overloads call into the specific case overloads; before this PR the reverse was true
- Overloads that take byte* or char* as arguments are wrappers around the overloads that take ReadOnlySpan

cc: @KrzysztofCwalina

### DIFF
--- a/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
+++ b/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
@@ -6,29 +6,37 @@ using System.Text.Utf8;
 
 namespace System.Text
 {
-    public static partial class PrimitiveParser2
+    public static partial class PrimitiveParser
     {
         public static partial class InvariantUtf16
         {
             public unsafe static bool TryParseBoolean(char* text, int length, out bool value)
             {
-                int consumedBytes;
-                var span = new ReadOnlySpan<byte>(text, length);
-                return PrimitiveParser2.TryParseBoolean(span, out value, out consumedBytes, EncodingData.InvariantUtf16);
+                int consumed;
+                var span = new ReadOnlySpan<byte>(text, length * 2);
+                return PrimitiveParser.TryParseBoolean(span, out value, out consumed, EncodingData.InvariantUtf16);
             }
-            public unsafe static bool TryParseBoolean(char* text, int length, out bool value, out int consumedBytes)
+            public unsafe static bool TryParseBoolean(char* text, int length, out bool value, out int consumed)
             {
-                var span = new ReadOnlySpan<byte>(text, length);
-                return PrimitiveParser2.TryParseBoolean(span, out value, out consumedBytes, EncodingData.InvariantUtf16);
+                var span = new ReadOnlySpan<byte>(text, length * 2);
+                int bytesConsumed;
+                bool result = PrimitiveParser.TryParseBoolean(span, out value, out bytesConsumed, EncodingData.InvariantUtf16);
+                consumed = bytesConsumed / 2;
+                return result;
             }
-            public static bool TryParseBoolean(ReadOnlySpan<byte> text, out bool value)
+            public static bool TryParseBoolean(ReadOnlySpan<char> text, out bool value)
             {
-                int consumedBytes;
-                return PrimitiveParser2.TryParseBoolean(text, out value, out consumedBytes, EncodingData.InvariantUtf16);
+                int consumed;
+                var byteSpan = text.Cast<char, byte>();
+                return PrimitiveParser.TryParseBoolean(byteSpan, out value, out consumed, EncodingData.InvariantUtf16);
             }
-            public static bool TryParseBoolean(ReadOnlySpan<byte> text, out bool value, out int consumedBytes)
+            public static bool TryParseBoolean(ReadOnlySpan<char> text, out bool value, out int consumed)
             {
-                return PrimitiveParser2.TryParseBoolean(text, out value, out consumedBytes, EncodingData.InvariantUtf16);
+                var byteSpan = text.Cast<char, byte>();
+                int bytesConsumed;
+                bool result = PrimitiveParser.TryParseBoolean(byteSpan, out value, out bytesConsumed, EncodingData.InvariantUtf16);
+                consumed = bytesConsumed / 2;
+                return result;
             }
         }
     }
--- a/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
+++ b/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
@@ -1,0 +1,35 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Text.Utf8;
+
+namespace System.Text
+{
+    public static partial class PrimitiveParser2
+    {
+        public static partial class InvariantUtf16
+        {
+            public unsafe static bool TryParseBoolean(char* text, int length, out bool value)
+            {
+                int consumedBytes;
+                var span = new ReadOnlySpan<byte>(text, length);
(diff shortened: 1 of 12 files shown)


src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
```diff
@@ -12,31 +12,88 @@ public static partial class InvariantUtf16
...
-                return PrimitiveParser.TryParseBoolean(byteSpan, out value, out consumed, EncodingData.InvariantUtf16);
+                value = default(bool);
+                if (text.Length < 1)
+                {
+                    return false;
+                }
+
+                char firstCodeUnit = text[0];
+
+                if (firstCodeUnit == '1')
```
KrzysztofCwalina: Why do we have this? Is it because .NET Framework has this behavior? If yes, we should think whether we want to continue with this behavior. It just adds more branches and I am not sure it's useful.

src/System.Text.Primitives/System/Text/Parsing/PrimitiveParser2_bool.cs
```diff
@@ -10,6 +10,30 @@ public static partial class PrimitiveParser
     {
         #region helpers
 
+        private static bool IsTrue(ReadOnlySpan<byte> utf8Bytes)
+        {
+            if (utf8Bytes.Length < 4)
+                return false;
+
+            byte firstChar = utf8Bytes[0];
+            if (firstChar != 't' && firstChar != 'T')
```
KrzysztofCwalina: Is that the behavior we have in .NET? i.e. "TrUe" parses to true?

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
botaberg: I did some investigations and it seems the .NET framework method boolean.TryParse does not have this behavior. It will fail to parse the strings "0" or "1". This behavior should be removed. Thanks for bringing this up!

src/System.Text.Primitives/System/Text/Parsing/PrimitiveParser2_bool.cs
KrzysztofCwalina: OK, for now let's leave the implementation as-is. If we need more perf, we can add case-sensitive overload[s].

jkotas: You can do `if ((firstChar & ~0x20) == 'T')` to squeeze the extra branch out.

jkotas: May be a good idea to create small helper function for it that gets always inlined.

jkotas: Or to have no branches at all, you can do: return (span.ReadInt32LittleEndian() & ~0x20202020) != 'T' + 0x100 * 'R' + 0x10000 * 'U' + 0x1000000 * 'E';

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
```diff
@@ -12,31 +12,88 @@ public static partial class InvariantUtf16
         {
             public unsafe static bool TryParseBoolean(char* text, int length, out bool value)
             {
-                int consumed;
-                var span = new ReadOnlySpan<byte>(text, length * 2);
-                return PrimitiveParser.TryParseBoolean(span, out value, out consumed, EncodingData.InvariantUtf16);
+                ReadOnlySpan<char> span = new ReadOnlySpan<char>(text, length);
```
davidfowl: Doesn't this defeat he purpose of the overload? It'll be slower with the slow span I'd assume

src/System.Text.Primitives/System/Text/Parsing/PrimitiveParser2_bool.cs
davidfowl: We use this trick in kestrel for doing case insensitive comparisons on ascii text. We generate the constants at build time for header names.

botaberg:
The code is not as optimized as it could be. I can apply your recommendations regarding the TryParseBoolean overloads as they are easy to implement. However, before we put serious effort into optimizing this code, I think we need to finish creating all of the PrimitiveParser.TryParse* skeleton implementations (right now we only have TryParseBoolean and TryParseUInt32) and then prepare benchmarks for validation of our optimization work. This may require us to switch some of our higher-level parsing APIs such as HTML parsers over to using the newer PrimitiveParser APIs. 

Feel free to comment on this or chat with me if you think these priorities are incorrect. I had an insightful exchange offline with Krzysztof earlier which has shaped some of these priorities.

joshfree:
@botaberg what's the next step in cleaing up this PR to get it checked in?

https://ci.dot.net/job/dotnet_corefxlab/job/master/job/Release_prtest/376/console
09:21:03 === TEST EXECUTION SUMMARY ===
09:21:03    System.Text.Primitives.Tests  Total: 1309, Errors: 0, Failed: 6, Skipped: 1, Time: 1.493s

botaberg:
@joshfree I have one more revision coming soon - implementing the byte* and char* versions of TryParseBoolean and removing test cases that no longer apply.

botaberg:
At this point, if the CI checks pass, I would recommend we merge this PR. I have addressed some but not all of the feedback; notably the implementations still are not as optimized as they could be. There are more branches than necessary, and there is a cast from Span<byte> to Span<char>, which may be slow. I believe we should wait until we have benchmarks to hone in the performance of these APIs.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
```diff
@@ -12,13 +12,50 @@ public static partial class InvariantUtf16
         {
             public unsafe static bool TryParseBoolean(char* text, int length, out bool value)
             {
-                ReadOnlySpan<char> span = new ReadOnlySpan<char>(text, length);
-                return TryParseBoolean(span, out value);
+                value = default(bool);
+                if (length < 1)
```
KrzysztofCwalina: Would it be better to remove this check? It's not even accurate anymore, i.e. should be < 4. But then IsTrue does the same check again.

src/System.Text.Primitives/System/Text/Parsing/PrimitiveParser2_bool.cs
```diff
@@ -10,6 +10,54 @@ public static partial class PrimitiveParser
     {
         #region helpers
 
+        private unsafe static bool IsTrue(byte* utf8Bytes, int length)
+        {
+            if (length < 4)
+                return false;
+
+            byte firstChar = utf8Bytes[0];
+            if (firstChar != 't' && firstChar != 'T')
```
KrzysztofCwalina: Please file an issue to possibly implement a case-sensitive fast parse that simply reads and int and compares it to a constant (as per the discussion we had in the previous commit version.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_bool.cs
botaberg: Yes - I can do this before the PR is merged if desired.

src/System.Text.Primitives/System/Text/Parsing/PrimitiveParser2_bool.cs
botaberg: https://github.com/dotnet/corefxlab/issues/963

^ I have created an issue and assigned myself.

KrzysztofCwalina: thanks

joshfree:
thanks @botaberg

PR #964, opened 2016-11-08
Add more implementations of PrimitiveParser.TryParse* overloads

botaberg:
Add skeleton implementations for TryParseSByte, Int16, Int32, Int64, as well as their unsigned counterparts.

These methods currently utilize the implementations generated in PrimitiveParser_unsigned.tt, and thus they are not optimized. The value in this PR is in the API surface area - once the rest of the new PrimitiveParser surface area (TryParseDecimal, Float, etc) is implemented, we can use these APIs in meaningful benchmarks and hone their performance.


botaberg:
I can squash the extra commits...

botaberg:
@KrzysztofCwalina @joshfree @yizhang82

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_Int16.cs
```diff
@@ -0,0 +1,75 @@
...
+                return result;
+            }
+
+            public static partial class Hex
+            {
+                public unsafe static bool TryParseInt16(char* text, int length, out short value)
+                {
+                    int consumed;
+                    var span = new ReadOnlySpan<byte>(text, length * 2);
+                    return PrimitiveParser.TryParseInt16(span, out value, out consumed, EncodingData.InvariantUtf16, 'X');
```
KrzysztofCwalina: One thing I did not think through when I proposed the API is that there are two different hex formats: 'X' and 'x' (at least for formatting). How does the parser work? will it parse both 0xFF and 0xff? If not, what should we do about it?

botaberg: The parser can be made to parse 0xFF and 0xff (and perhaps 0xFf) as long as we think we can do this in a performant manner. I am unsure about the direction we should go if we determine that having "catch-all" Hex parsing APIs is not what we want.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_Int16.cs
```diff
@@ -0,0 +1,75 @@
...
+namespace System.Text
+{
+    public static partial class PrimitiveParser
+    {
+        public static partial class InvariantUtf16
+        {
+            public unsafe static bool TryParseInt16(char* text, int length, out short value)
+            {
+                int consumed;
+                var span = new ReadOnlySpan<byte>(text, length * 2);
```
yizhang82: We should prefer sizeof(char) instead of 2

KrzysztofCwalina: @botaberg, please change the PR to use sizeof(char). It's a great suggestion. Once you update the PR I will merge. We can tweak the hex parsing separately.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf16/InvariantUtf16_Int32.cs
```diff
@@ -13,12 +13,12 @@ public static partial class InvariantUtf16
...
+                var span = new ReadOnlySpan<byte>(text, length * sizeof(char));
                 return PrimitiveParser.TryParseInt32(span, out value, out consumed, EncodingData.InvariantUtf16);
             }
             public unsafe static bool TryParseInt32(char* text, int length, out int value, out int charactersConsumed)
             {
-                var span = new ReadOnlySpan<byte>(text, length * 2);
+                var span = new ReadOnlySpan<byte>(text, length * sizeof(char));
                 int bytesConsumed;
                 bool result = PrimitiveParser.TryParseInt32(span, out value, out bytesConsumed, EncodingData.InvariantUtf16);
                 charactersConsumed = bytesConsumed / 2;
```
KrzysztofCwalina: Same should be used in the "/ 2" lines.

PR #971, opened 2016-11-10
Add PrimitiveParser.TryParseDecimal overloads

botaberg:
Add PrimitiveParser.TryParseDecimal overloads and a few tests.


src/System.Text.Primitives/System/Text/Parsing/PrimitiveParser2_decimal.cs
```diff
@@ -0,0 +1,32 @@
...
+            }
+            else if (encoding.IsInvariantUtf16)
+            {
+                ReadOnlySpan<char> textChars = text.Cast<byte, char>();
+                int charactersConsumed;
+                bool result = InvariantUtf16.TryParseDecimal(textChars, out value, out charactersConsumed);
+                bytesConsumed = charactersConsumed * sizeof(char);
+                return result;
+            }
+
```
KrzysztofCwalina: Here, you need to call/execute the very general purpose parser for non-invariant cultures. Is that something you left for later? If yes, then please file an issue and put a comment in here with the issue number.

botaberg: I'll have to do this for all of the TryParse overloads outside of InvariantUtf8 and InvariantUtf16. Issue will be tracked soon.

PR #980, opened 2016-11-15
Introducing System.IO.Pipelines

davidfowl:
- Golang took the name channels so now we have pipelines

@KrzysztofCwalina


jchannon:
So what if Golang named it first, why can't 2 languages have the same named concept?

If a golang user comes to .net they will immediately know what channels is.

JesperTreetop:
@jchannon Yes, they will have an idea of what channels should be and be highly confused, because channels in Go are a completely different concept. What is now called Pipelines in .NET is an optimized specialization of streams. [In Go](https://tour.golang.org/concurrency/2), "Channels are a typed conduit through which you can send and receive values with the channel operator, `<-`."

stephentoub:
The project at https://github.com/dotnet/corefxlab/tree/master/src/System.Threading.Tasks.Channels maps much more closely in concept to Go's channel.

src/System.IO.Pipelines/DefaultReadableBufferExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+        /// <param name="stream">The target <see cref="Stream"/></param>
+        /// <returns></returns>
+        public static async Task CopyToAsync(this ReadableBuffer buffer, Stream stream)
+        {
+            foreach (var memory in buffer)
+            {
+                ArraySegment<byte> data;
+                if (memory.TryGetArray(out data))
+                {
+                    await stream.WriteAsync(data.Array, data.Offset, data.Count);
```
stephentoub: All of the awaits on tasks should use ConfigureAwait(false)

src/System.IO.Pipelines/DefaultReadableBufferExtensions.cs
```diff
@@ -0,0 +1,96 @@
...
+    {
+        /// <summary>
+        /// Copies a <see cref="ReadableBuffer"/> to a <see cref="Stream"/> asynchronously
+        /// </summary>
+        /// <param name="buffer">The <see cref="ReadableBuffer"/> to copy</param>
+        /// <param name="stream">The target <see cref="Stream"/></param>
+        /// <returns></returns>
+        public static async Task CopyToAsync(this ReadableBuffer buffer, Stream stream)
+        {
+            foreach (var memory in buffer)
```
stephentoub: Is it common/likely for buffer to have more than one "memory"?  If so, you could special case that and return the result of stream.WriteAsync directly, avoiding the extra async method frame.

src/System.IO.Pipelines/Gate.cs
```diff
@@ -0,0 +1,77 @@
...
+    internal class Gate : ICriticalNotifyCompletion
+    {
+        private static readonly Action _gateIsOpen = () => {};
+
+        private volatile Action _gateState;
+
+        /// <summary>
+        /// Returns a boolean indicating if the gate is "open"
+        /// </summary>
+        public bool IsCompleted => _gateState == _gateIsOpen;
```
stephentoub: >_gateState == _gateIsOpen [](start = 35, length = 25)

Unless it's misused, checking here and in other places in the file ```!= null``` should provide the same behavior as ```== _gateIsOpen```, and without needing to load the static field.

src/System.IO.Pipelines/Gate.cs
```diff
@@ -0,0 +1,77 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Resets the gate to continue blocking the waiter. This is called immediately after awaiting the signal.
+        /// </summary>
+        public void GetResult()
+        {
+            // Clear the active continuation to "reset" the state of this event
+            Interlocked.Exchange(ref _gateState, null);
```
stephentoub: >Interlocked.Exchange [](start = 12, length = 20)

Why an interlocked?

khalidabuhakmeh:
I like it!

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -0,0 +1,153 @@
...
+                input.Advance(sliced.End);
+
+                if (actual != 0)
+                {
+                    return new ValueTask<int>(actual);
+                }
+                else if (fin)
+                {
+                    return new ValueTask<int>(0);
+                }
```
stephentoub: Couldn't this just be:
```C#
if (action != 0 || fin)
{
    return new ValueTask<int>(actual);
}
```
?

src/System.IO.Pipelines/PipelineConnectionStream.cs
```diff
@@ -0,0 +1,199 @@
...
+                if (_cachedTask.Result != task.Result)
+                {
+                    // Needs .AsTask to match Stream's Async method return types
+                    _cachedTask = task.AsTask();
+                }
+            }
+            else
+            {
+                // Needs .AsTask to match Stream's Async method return types
+                _cachedTask = task.AsTask();
```
stephentoub: >_cachedTask = task.AsTask(); [](start = 16, length = 28)

I'm not sure this is a good thing to do.  First, the operation could fail, in which case you've cached a task that has an exception, and the next time you check _cachedTask.Result, it'll throw.  Second, you could end up in a situation where a call to ReadAsync blocks waiting on the _cachedTask.  I think you only want to cache it in the task.IsCompletedSuccessfully case.

src/System.IO.Pipelines/PipelineConnectionStream.cs
```diff
@@ -0,0 +1,199 @@
...
+
+        public override void Flush()
+        {
+            // No-op since writes are immediate.
+        }
+
+        public override Task FlushAsync(CancellationToken cancellationToken)
+        {
+            // No-op since writes are immediate.
+            return Task.FromResult(0);
```
stephentoub: >Task.FromResult(0) [](start = 19, length = 18)

Task.CompletedTask

src/System.IO.Pipelines/StreamExtensions.cs
```diff
@@ -0,0 +1,244 @@
...
+            {
+                if (task.IsFaulted)
+                {
+                    writer.CompleteReader(task.Exception);
+                }
+                else
+                {
+                    writer.CompleteReader();
+                }
+            });
```
stephentoub: Would be better as:
```C#
writer.CopyToAsync(stream).ContinueWith((t,s) =>
{
    var innerWriter = (PipelineReaderWriter)s;
    if (t.IsFaulted) innerWriter.CompleteReader(task.Exception.InnerException);
    else innerWriter.CompleteReader();
}, writer, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
```
That way, you avoid the closure/delegate allocation closing over state, the continuation is invoked as part of the antecedent task's completion (rather than queueing another work item, and the continuation will not be forced back to any custom TaskScheduler that was present.  Similar comments apply to other uses of ContinueWith in this PR.

src/System.IO.Pipelines.File/FileReader.cs
```diff
@@ -0,0 +1,194 @@
...
+        }
+
+        private class Box<T>
+        {
+            public Box(T value)
+            {
+                Value = value;
+            }
+            public T Value { get; set; }
+        }
```
stephentoub: System.Runtime.CompilerServices.StrongBox<T>

src/System.IO.Pipelines.File/FileReader.cs
```diff
@@ -0,0 +1,194 @@
...
+        {
+        }
+
+        public FileReader(PipelineReaderWriter input) : base(input)
+        {
+        }
+
+        // Win32 file impl
+        // TODO: Other platforms
+        public unsafe void OpenReadFile(string path)
```
stephentoub: Do you still need this now that FileStream has an optimized CopyToAsync implementation?  What further changes would be necessary there to obviate the need for this here?

src/System.IO.Pipelines.Networking.Sockets/Internal/Signal.cs
```diff
@@ -0,0 +1,85 @@
...
+{
+    /// <summary>
+    /// Very lightweight awaitable gate - intended for use in high-volume single-producer/single-consumer
+    /// scenario, in particular targeting the bridge between async IO operations
+    /// and the async method that is pumping the read/write queue. A key consideration is that
+    /// no objects (in particular Task/TaskCompletionSource) are allocated even in the await case. Instead,
+    /// a custom awaiter is provided. Works like a <see cref="ManualResetEvent "/> - the <see cref="Reset"/> method must
+    /// be called between operations.
+    /// </summary>
+    internal class Signal : INotifyCompletion
```
stephentoub: ICriticalNotifyCompletion

src/System.IO.Pipelines.Networking.Sockets/Internal/Signal.cs
```diff
@@ -0,0 +1,85 @@
...
+
+        private object SyncLock => this;
+
+        public Signal GetAwaiter() => this;
+
+        public void GetResult() { }
+
+        public void OnCompleted(Action continuation)
+        {
+            if (continuation != null)
```
stephentoub: > if (continuation != null) [](start = 11, length = 26)

This would only happen if it was misused... the compiler will never pass a null continuation.

src/System.IO.Pipelines.Networking.Sockets/Internal/Signal.cs
```diff
@@ -0,0 +1,85 @@
...
+            {
+                var oldValue = Interlocked.CompareExchange(ref _continuation, continuation, null);
+
+                if (ReferenceEquals(oldValue, _completedSentinel))
+                {
```
stephentoub: Same comments as on Gate about this.

samples/System.IO.Pipelines.Samples/HttpServer/HttpResponseStream.cs
```diff
@@ -0,0 +1,121 @@
...
+
+        public override void EndWrite(IAsyncResult asyncResult)
+        {
+            ((Task<object>)asyncResult).GetAwaiter().GetResult();
+        }
+
+        private Task WriteAsync(byte[] buffer, int offset, int count, CancellationToken cancellationToken, object state)
+        {
+            var tcs = new TaskCompletionSource<object>(state);
+            var task = WriteAsync(buffer, offset, count, cancellationToken);
```
stephentoub: >var task = WriteAsync(buffer, offset, count, cancellationToken); [](start = 12, length = 64)

In cases like this, where the async operation is likely to complete synchronously, you're better off adding a fast path, e.g.
```C#
Task task = WriteAsync(...);
if (task.Status == TaskStatus.RanToCompletion)
{
    return task;
}

var tcs = ...
...
```

adamralph:
Does it matter that golang might introduce something called pipelines later which is again completely different to this?

nzall:
@adamralph golang already has a concept called [pipelines](https://blog.golang.org/pipelines), which is a series of stages (goroutines running the same function) connected by channels. Since there is only a limited set of words in the English vocabulary that are suitable for software purposes, chances are that the good ones are already taken.

omariom:
@davidfowl 
What is the allocation profile of such approach compared to the traditional streams?

KrzysztofCwalina:
> using System.Buffers;

All the files should have the standard copyright/license notice

---
Refers to: src/System.IO.Pipelines/ArrayBufferPool.cs:1 in 56ee6e7. [](commit_id = 56ee6e7d1015c64d8b5cdc517da632391870ae56, deletion_comment = False)

src/System.IO.Pipelines/PipelineConnectionStream.cs
```diff
@@ -0,0 +1,199 @@
+﻿using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    public class PipelineConnectionStream : Stream
```
KrzysztofCwalina: >Stream [](start = 44, length = 6)

I think the adapters should be separated (namespacewise) from the pure-pipeline APIs

src/System.IO.Pipelines/PipelineReader.cs
```diff
@@ -0,0 +1,57 @@
...
+{
+    /// <summary>
+    /// Represents a pipeline from which data can be read.
+    /// </summary>
+    public abstract class PipelineReader : IPipelineReader
+    {
+        /// <summary>
+        /// The underlying <see cref="PipelineReaderWriter"/> the <see cref="PipelineReader"/> communicates over.
+        /// </summary>
+        protected readonly PipelineReaderWriter _input;
```
KrzysztofCwalina: Yeah, the fact that reader wraps reader/writers shows that we don't yet have a good name for the reader/writer.

src/System.IO.Pipelines/MemoryPool.cs
```diff
@@ -0,0 +1,243 @@
...
+
+        /// <summary>
+        /// Called to take a block from the pool.
+        /// </summary>
+        /// <returns>The block that is reserved for the called. It must be passed to Return when it is no longer being used.</returns>
+#if DEBUG
+        private MemoryPoolBlock Lease(
+            [CallerMemberName] string memberName = "",
+            [CallerFilePath] string sourceFilePath = "",
+            [CallerLineNumber] int sourceLineNumber = 0)
```
davidfowl: We'll need to figure out a way to get these diagnostics without being in debug mode. But these are fine for now

src/System.IO.Pipelines.Compression/Inflater.cs
```diff
@@ -0,0 +1,225 @@
...
+using System.IO;
+using System.IO.Compression;
+using System.Security;
+
+namespace System.IO.Pipelines.Compression
+{
+    /// <summary>
+    /// Provides a wrapper around the ZLib decompression API
+    /// </summary>
+    internal sealed class Inflater : IDisposable
```
KrzysztofCwalina: Yes, but we need to design clean low level APIs.

---
In reply to: [88018514](https://github.com/dotnet/corefxlab/pull/980#discussion_r88018514) [](ancestors = 88018514)

KrzysztofCwalina:
:clock1:

src/System.IO.Pipelines.File/FileReader.cs
davidfowl: This was a proof of concept done way before FileStream.CopyToAsync had those changes. Actually it would be better to use FileStream.CTA adapter directly now (at least on .NET Core) because you get the cross platform benefits. There are some questions as to whether we ship this at all but for now its just an experiment. Ideally we'd have a unix impl so we could see what that looks like.

src/System.IO.Pipelines.File/FileReader.cs
davidfowl: Nice

adamralph:
With reference to [@stephentoub's comment](https://github.com/dotnet/corefxlab/pull/980#issuecomment-260638418), is the implication that `System.IO.Pipelines` aligns more closely with golang pipelines ([referenced by @nzall](https://github.com/dotnet/corefxlab/pull/980#issuecomment-260660592))?

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
davidfowl: fin being true doesn't mean the buffer is empty. I'd rather keep the logic split for now and do further testing to make sure that length is 0 when fin is true (in this case).

stephentoub:
> With reference to @stephentoub's comment, is the implication that System.IO.Pipelines aligns more closely with golang pipelines (referenced by @nzall)?

A "pipeline" is a general pattern, a series of processing elements (often called stages) with data handed from one stage to the next.  A pipeline can be implemented in many languages/frameworks; in Go, it's typically implemented using a goroutine for each stage and using a channel as the mechanism for handed data between stages.  But the concept is in no way specific to Go.

src/System.IO.Pipelines/PipelineConnectionStream.cs
davidfowl: Suggestions?

src/System.IO.Pipelines/StreamExtensions.cs
davidfowl: Ah yea, I was being lazy here. Though passing of those arguments is one of the reasons I try to avoid continue with...

src/System.IO.Pipelines/PipelineConnectionStream.cs
KrzysztofCwalina: System.IO.Pipelines.Bridge? System.IO.Pipelines.Streams?

src/System.IO.Pipelines/Gate.cs
davidfowl: I'll file bugs for the Gate/Signal changes. I want to get this in asap since it's a giant "move the code over" PR. I'd prefer if there was a reusable TCS in the framework, these 2 types are basically that.

stephentoub: > I'd prefer if there was a reusable TCS in the framework, these 2 types are basically that.

I've considered that several times over the years.  I think it'd be reasonable to add something to TCS that would essentially allow it to dump its existing task and create a new one, but that's really not sufficient to replace these custom awaitables... for that, you'd not only need to make TCS reusable, but you'd need to allow a Task to go from completed to not completed, which could potentially break a whole lot of code if used incorrectly.  Such a task-resetting feature could only be used in a closed environment where the producer and consumer of the task coordinate, as otherwise the producer won't know when it's safe to reset (the same kind of issue that exists with pooled data structures in general).

davidfowl: Agreed, this is one of those systems. We just need primitives that give the power. Maybe a new namespace? `System.*.KnowWhatYouAreDoing`

adamralph:
@stephentoub thanks for the explanation. I'm trying to rationalise the rename from "channels" to "pipelines". I can see how the rename probably makes sense in this case, but basing the decision on another platform having "taken" the generic noun you intended to use feels a bit off somehow.

JesperTreetop:
@adamralph Look at the repository @stephentoub linked to - they're not changing the name of this thing "because we can't use names taken in other languages", but because they already are testing something which would be closer to channels and want to name that thing channels instead.

adamralph:
@JesperTreetop that sounds like a good reason. It's a very different reason to that given by the OP.

dazinator:
Couldn't resist commenting.. I have literally no idea what this is.. but I love it +1. Look forward to working it out through experimentation :-)

malachib:
I'm excited about this, it looks like a potentially fresher and better tool than Streams, especially for in-process / short-hop data movement

samples/System.IO.Pipelines.Samples/CompressionSample.cs
```diff
@@ -0,0 +1,41 @@
...
+                              .DeflateCompress(factory, CompressionLevel.Optimal)
+                              .DeflateDecompress(factory);
+
+                // Wrap the console in a pipeline writer
+                var output = factory.CreateWriter(Console.OpenStandardOutput());
+
+                // Copy from the file reader to the console writer
+                input.CopyToAsync(output).GetAwaiter().GetResult();
+
+                input.Complete();
```
dazinator: What happens if you forget to call Complete() ? Do you think there would be any benefit if `IPipelineReader` implemented `IDisposable`, and called Complete() in it's dispose?

src/System.IO.Pipelines/DefaultReadableBufferExtensions.cs
electricessence: Is this still true?

PR #987, opened 2016-11-16
CommandLine: Allow extra arguments

### LINKED ISSUE #825: Support unknown options
I would like to capture unknown options for a tool, that will be sent to another tool that takes arbitrary parameters used in a script. Is this possible somehow?
### Example

This is what invocation could look like:

```
> mytool.exe --verbosity diagnostic --unknown value1 value2 --unknownflag --hello world
```

Where `--unknown value1 value2 --unknownflag --hello world` are parameters sent to the script.

And this is what I would want to do when parsing:

``` csharp
var result = ArgumentSyntax.Parse(args, syntax =>
{
    syntax.AllowUnknownOptions = true;
    syntax.DefineOption("v|verbosity", Verbosity.Normal, VerbosityParser.Parse);
});

IDictionary<string, IList<string>> unknown = result.GetUnknownOptions();
```

I understand that something like this might be out of scope for what you're building here, and I'm not even sure if the API that I've proposed makes any sense outside of my needs, but I thought I would ask.

If this is something that is of interest, I would be more than happy to provide a PR.

bricelam:
Resolves #825

### DIFF
--- a/src/System.CommandLine/System/CommandLine/ArgumentSyntax.cs
+++ b/src/System.CommandLine/System/CommandLine/ArgumentSyntax.cs
@@ -26,6 +26,7 @@ namespace System.CommandLine
             HandleErrors = true;
             HandleHelp = true;
             HandleResponseFiles = true;
+            ErrorOnUnexpectedArguments = true;
         }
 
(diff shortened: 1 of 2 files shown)


KrzysztofCwalina:
looks good. thanks.

PR #1000, opened 2016-11-18
Changed DefaultWritableBufferExtensions.Write to take ReadOnlySpan<byte>

KrzysztofCwalina:
The Write method currently takes Span<byte> as a parameter.
It should take ReadOnlySpan<byte> as the method does not mutate the contents of the data it writes.
This will enable writing new Utf8String("Hello World!).Bytes, and no technology can succeed without supporting the Hello World. 

cc: @botaberg, @joshfree, @davidfowl, @AtsushiKan


davidfowl:
`new Utf8String("Hello World!).Bytes` is pretty ugly 😄

davidfowl:
Since `Span<T>` has an implicit conversion to `ROS<T>` it should be fine

KrzysztofCwalina:
> new Utf8String("Hello World!).Bytes` is pretty ugly 😄 

once we have UTF8 literals (and of course the representation for them) we will add an extension method to write it directlu. And so the sample will be: 
```c#
buffer.Write("Hello World"u8);
```

PR #1001, opened 2016-11-19
InvariantUtf8.TryParseUInt32 byte* overloads: Performance Improvements and Benchmarks

botaberg:
PrimitiveParser.InvariantUtf8.TryParseUInt32(byte* text, int length, out uint value)
PrimitiveParser.InvariantUtf8.TryParseUInt32(byte* text, int length, out uint value, out int BytesConsumed)

This change introduces more optimized implementations of the above parsing methods, as well as some xunit.performance benchmarks to compare these implementations to uint.TryParse and InternalParser.TryParse. These implementations are similar to the ones in InternalParser, which are already pretty fast. The main improvement is that they don't start checking for overflow until the tenth byte is parsed, since uint.MaxValue is ten digits long.

I have been running these benchmarks using xunit.performance, but visualization of results is still difficult. I have created a small summary of the test results here: [gist](https://gist.github.com/botaberg/c3a5d66e7afc8c72039cff26191cccd2)

We still have room to try different optimization strategies. I tried some creative "improvements" myself. Notable observations:
- Favoring ternary ?: moves over branching if-else statements doesn't seem to improve performance.
- Trying to manually unroll loops in source code doesn't seem to improve performance.
- Function calls, even with inlining, seem to worsen performance. This is especially true with out parameters; perhaps they are required to return on the stack?
- The overloads with the `out bytesConsumed` appear to be slower than those without. This is a compelling reason to keep the distinction between these overloads.
- The most successful performance improvements have been related to reducing the amount of instructions needed to parse each character.
- The InternalParser code is already a vast improvement over the baseline uint.TryParse. Good job Jon Fortescue! Perhaps there are other implementations to compare against.


botaberg:
@KrzysztofCwalina @joshfree 
@yizhang82 

More to come soon!

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
```diff
@@ -10,16 +11,128 @@ public static partial class PrimitiveParser
     {
         public static partial class InvariantUtf8
         {
+            static int UINT_OVERFLOW_LENGTH = 10;
+
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
KrzysztofCwalina: This method is too long to be inlined

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
```diff
@@ -10,16 +11,128 @@ public static partial class PrimitiveParser
...
+
+                if (length < UINT_OVERFLOW_LENGTH)
+                {
+                    // Length is less than OVERFLOW_LENGTH; overflow is not possible
+                    for (int index = 0; index < length; index++)
+                    {
+                        uint nextDigit = text[index] - 48u; // '0'
+                        if (nextDigit > 9)
+                        {
+                            return true;
```
KrzysztofCwalina: don't you need to check that at least one digit was successfully parsed?

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
```diff
@@ -10,16 +11,128 @@ public static partial class PrimitiveParser
...
+                        }
+                        value = value * 10 + nextDigit;
+                    }
+                }
+                
+                return true;
             }
+            
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
             public unsafe static bool TryParseUInt32(byte* text, int length, out uint value, out int bytesConsumed)
```
KrzysztofCwalina: Did you have a chance to measure if the additional parameter matters?

KrzysztofCwalina:
Looks very nice! Thanks for doing good work here.

PR #1002, opened 2016-11-19
Fixed link to span.md

svick:
Also changed absolute links to relative.

(The advantage of relative links is that they point to the version of the document for the current commit, even when that's not `master`.)


KrzysztofCwalina:
Thanks!

PR #1006, opened 2016-11-21
Expose LocalEndPoint property on SocketListener

stbraley:
When using a system assigned port for the SocketListener. The SocketListener class needs to expose the address of the socket so the caller can obtain the assigned port. I've added a property to expose the endpoint of the socket. If the socket is null, then the property will return a empty endpoint.

### DIFF
--- a/src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
+++ b/src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
@@ -86,6 +86,8 @@ namespace System.IO.Pipelines.Networking.Sockets
             }
         }
 
+        public EndPoint LocalEndPoint => _socket == null ? new IPEndPoint(IPAddress.None, 0) : _socket.LocalEndPoint;
+
         private Socket GetReusableSocket() => null; // TODO: socket pooling / re-use
 
         private void BeginAccept(SocketAsyncEventArgs args)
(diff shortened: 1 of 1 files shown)


davidfowl:
It makes sense to expose the entire socket.

src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
```diff
@@ -14,7 +14,7 @@ namespace System.IO.Pipelines.Networking.Sockets
     public class SocketListener : IDisposable
     {
         private readonly bool _ownsFactory;
-        private Socket _socket;
+        public Socket _socket;
```
davidfowl: Not like that, expose a readonly property...

davidfowl:
I read this incorrectly. I though you wanted to expose the client socket...

stbraley:
No, need to get the listening port number.

Modeling after the TcpListener in the core framework.

`var listener = new SocketListener();
 var endPoint = new IPEndPoint(IPAddress.Loopback, 0);
_socketListener.Start(_endPoint);

var port = ((IPEndPoint)listener.LocalEndpoint).Port;`

src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
```diff
@@ -86,6 +86,8 @@ public void Stop()
             }
         }
 
+        public Socket ListeningSocket { get { return _socket;} }
```
davidfowl: C#6 ify

```C#
public Socket ListeningSocket => _socket;
```

davidfowl:
LGTM

PR #1008, opened 2016-11-21
Implement PrimitiveParser.InvariantUtf8.TryParseUInt32 span and Hex overloads

botaberg:
Implement PrimitiveParser.InvariantUtf8.TryParseUInt32 span and Hex overloads, and add performance tests for these new implementations. Hex translation from ASCII/Utf8 byte to integer is done by indexing into a static readonly array; suggestions for improvement in this area or others are welcome.

Observations about performance so far:
- Hex parsing takes about the same time per character as parsing in base 10
- Parsing byte* seems to be slightly faster than ReadOnlySpan<byte>, but overall the two are very comparable

I don't have a gist with performance numbers this time around; my hope is that we can get some good analysis going soon.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
+++ b/src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
@@ -10,11 +10,10 @@ namespace System.Text
     {
         public static partial class InvariantUtf8
         {
-            #region static
-            const int s_uint32OverflowLength = 10;
-            const int s_uint32OverflowLengthHex = 8;
+            const int s_UInt32OverflowLength = 10;
+            const int s_UInt32OverflowLengthHex = 8;
 
-            static readonly byte[] s_hexLookup =
+            static readonly byte[] s_HexLookup =
             {
                 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,             // 15
                 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,             // 31
@@ -33,7 +32,6 @@ namespace System.Text
                 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF,             // 239
                 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF, 0xFF              // 255
             };
-            #endregion
             
             public unsafe static bool TryParseUInt32(byte* text, int length, out uint value)
             {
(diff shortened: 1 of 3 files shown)


src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
```diff
@@ -11,20 +11,44 @@ public static partial class PrimitiveParser
     {
         public static partial class InvariantUtf8
         {
-            static int UINT_OVERFLOW_LENGTH = 10;
+            #region static
+            private static readonly int UINT_OVERFLOW_LENGTH = 10;
```
jkotas: This should be const.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
```diff
@@ -151,35 +176,425 @@ public unsafe static bool TryParseUInt32(byte* text, int length, out uint value,
...
+                    }
+                    value = (uint)firstDigit;
+
+                    if (length <= UINT_OVERFLOW_LENGTH_HEX)
+                    {
+                        // Length is less than or equal to OVERFLOW_LENGTH; overflow is not possible
+                        for (int index = 1; index < length; index++)
+                        {
+                            byte nextByte = text[index];
+                            int nextDigit = hexLookup[nextByte];
```
jkotas: Note that the access to static variables come with static constructor checks - this line basically translates to:

```
if (has cctor run)
    Run static constructor
fetch hexLookup
```

It is often beneficial to cache the static variable in a local, to guarantee that the static cctor check is done just once.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
```diff
@@ -11,20 +11,44 @@ public static partial class PrimitiveParser
     {
         public static partial class InvariantUtf8
         {
-            static int UINT_OVERFLOW_LENGTH = 10;
+            #region static
+            private static readonly int UINT_OVERFLOW_LENGTH = 10;
+            private static readonly int UINT_OVERFLOW_LENGTH_HEX = 8;
+
+            static readonly int[] hexLookup =
```
jkotas: https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md say to prefix static fields with `s_`.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
jkotas: (and CamelCase)

KrzysztofCwalina:
> Parsing ReadOnlySpan seems to be slightly faster than byte*, but overall the two are very comparable

@botaberg, why would that be?

botaberg:
@KrzysztofCwalina After re-looking it seems the opposite is the case. Parsing byte* is faster than ReadOnlySpan<byte>. I wondered that myself after you brought it up. I will amend my PR notes.

KrzysztofCwalina:
@botaberg, what's the rough difference in perf?

botaberg:
@KrzysztofCwalina In my most recent tests, parsing a 10 digit byte* took about 0.75 times as long as ReadOnlySpan<byte>, and parsing a 1-digit byte* took about 0.9 times as long as ReadOnlySpan<byte>. However, the times for the variable length tests were nearly identical. Not sure what to make of this; the performance for each method also seems to change after rebuilds.

KrzysztofCwalina:
@botaberg, ok this is not too bad. Once we have the fast span, we probably will get even closer to parity

KrzysztofCwalina:
Looks good. Thanks! Nice PR.

src/System.Text.Primitives/System/Text/Parsing/InvariantUtf8/InvariantUtf8_UInt32.cs
KrzysztofCwalina: @botaberg, is that something you will still change? i.e. When you change this, and remove the "s_" from the constants, I think we can merge.

PR #1011, opened 2016-11-22
Record some more of the thinking for parsing

terrajobst:
@KrzysztofCwalina, thoughts?

### DIFF
--- a/docs/specs/parsing.md
+++ b/docs/specs/parsing.md
@@ -1,10 +1,15 @@
 # Parsing
 
+The purpose of this document is to put some stakes in the ground for a set of
+low-allocation parsing APIs.
+
+The current prototype is [here](https://github.com/dotnet/corefxlab/tree/master/src/System.Text.Primitives/System/Text/Parsing).
+
 ## Goals
 
 * Parses the following types with minimal allocations, ideally none:
-    - `bool`
-    - `byte`, `sbyte`
+    - `Boolean`
+    - `Byte`, `SByte`
     - `Int16`, `UInt16`
     - `Int32`, `UInt32`
     - `Int64`, `UInt64`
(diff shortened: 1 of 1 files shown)


docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
...
+                               out int bytesConsumed)
+```
+
+This allows the parsing to happen from a wide variety of data sources:
+
+* **Native memory**, by creating a span from the pointer + lengh.
+* **`byte[]`**, via an implicit conversion
+* **`Span<byte>`**, via an implicit conversion
+
+Depending on how we express encodings and cultures, it's likeky what we need to
```
jkotas: likeky -> likely

docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
...
+`ISequence<T>`, which is basically a more efficient `IEnumerable<T>`:
+
+```C#
+public static bool TryParseXyz<T>(this T memorySequence,
+                                  out Xyz value,
+                                  out int bytesConsumed)
+    where T : ISequence<ReadOnlyMemory<byte>>;
+```
+
+We'll need to decide if that's the right approach or not. Alternatively, we
```
jkotas: In particular, this API is likely to be too slow to be usable because of it will need to convert from `Memory<T>` to `Span<T>` every time is called.

I believe that the right approach is to have a stack-only type `struct Parser` that has the current Span being parsed cached in a field, and the `ISequence` to get more buffers to parse as another field.

docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
...
+
+* **Native memory**, by creating a span from the pointer + lengh.
+* **`byte[]`**, via an implicit conversion
+* **`Span<byte>`**, via an implicit conversion
+
+Depending on how we express encodings and cultures, it's likeky what we need to
+create additional arguments, defaulted to being UTF8 and invariant.
+
+The result of the method indicates whether parsing was successful. Since the
+entire point of these APIs is performance, we don't plan on adding APIs that
```
jkotas: How does one distinguish between the invalid input, and non-enough-characters in the input buffer?

docs/specs/parsing.md
jkotas: Also, the shape of the API leads to generic expansion bloat ... but that's a secondary problem.

docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
...
     - UTF16-BE
     - UTF16-LE
 * Parses from existing buffers without knowing up-front how many bytes the
   parsing needs to consume
 
-## Discussion
+## Focus on UTF8
+
+While we generally need to handle parsing for more than just UTF8 we belive it's
+beneficial to have an implemetnation that is and-optimized for UTF8 instead of
```
jkotas: implemetnation -> implementation 
and-optimized -> hand-optimized

docs/specs/parsing.md
KrzysztofCwalina: Additional, probably even bigger problem, is that such type needs to be disposable and disposed, otherwise the reservation will leak.

docs/specs/parsing.md
KrzysztofCwalina: The following code explains it the best: https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Formatting/System/Text/Parsing/SequenceParser.cs#L15

docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
     - `DateTime`, `DateTimeOffset`
     - `TimeSpan`
     - `Guid`
+    - `Uri`
 * Parses different cultures
     - e.g. English and German dates
```
KrzysztofCwalina: We should consider separating culture-dependent date-time parsing. I think it would be reasonable if it was handled by System.Text.Formatting library (which needs to be renamed but it is higher level text handling library using the primitives). The primitives library would handle UTC "R" format parsing only (which is needed for parsing networking protocols).

docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
...
-    - Needs to handle all cultures .NET can handle
+    - Needs to handle all cultures .NET Core can handle
+* Parse any outputs produced by [formatting](formatting.md)
+    - This ensures parsing & formatting behave symmetric
 * Parse from these encodings
-    - UTF8      <------------- THATS WHAT WE CARE ABOUT
+    - UTF8
     - ASCII
```
KrzysztofCwalina: We have to handle support ISO-8859-1. It's the encoding used by HTTP. It's a ASCII-based encoding and so we might as well handle ASCII too. The good thing is that I believe the current design can easily handle these, but we need to say that we care about more than just UTF8.

docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
...
     - UTF16-BE
     - UTF16-LE
 * Parses from existing buffers without knowing up-front how many bytes the
   parsing needs to consume
```
KrzysztofCwalina: This section should also talk about the goals for supporting all "formats", i.e. 'G', 'N', "x", "X", etc. and more broadly about compatibility with .NET, i.e. can this library parse everything that the current APIs can parse with the same results?

docs/specs/parsing.md
```diff
@@ -13,18 +18,75 @@
     - `DateTime`, `DateTimeOffset`
```
KrzysztofCwalina: BTW, as part of the larger cloud framework effort, I think we should consider adding UtcTime type. If we had that, we should consider removing these DateTime structures from the primitives (as far as parsing is concerned).

KrzysztofCwalina: Also, it might be worth stating that single, double, and decimal are lower priority.

KrzysztofCwalina: I would add BigInteger to the list. I think it might be required to be able to implement full JSON parsing support.

docs/specs/parsing.md
jkotas: We have discussed before having a language support for converting memory to span, something like:

```
fixed (Span<char> c = memory) // This handles reservation for Span - emits acquire/release reservation at the start/end of the scope
{
    ...
}
```

If this pattern is convention based and works on any types that implement methods with given names, it should just work for Parser as well. Something like:

```
fixed (Parser p = bufferSequence)
{
    ...
}
```

KrzysztofCwalina: There are many different parsers (JSON, XML, HTTP, etc). User code chooses which one they need to use. How owuld that work?

Also, it would be much preferable if parsers were created using ctors. It's more idiomatic to .net.

KrzysztofCwalina: Also, I feel like we are trying to solve a problem that is very rare (if not theoretical) and already exists in today's ASP.NET pipeline. The pipeline already uses pooled buffers and code can (and often pins) the buffers for interop (or other reasons) while a different thread can dispose the buffer (which makes it subject to use after free). We don't currently have check to prevent this.

In practice the problem boils down to whether a different thread disposes the owned memory between the call to _output.Buffer and the call to TryFormat.
```c#
        public void Append(int value)
        {
            int bytesWritten;
            while (!PrimitiveFormatter.TryFormat(value, _output.Buffer, out bytesWritten, default(TextFormat), _encoding))
            {
                _output.Enlarge();
            }
            _output.Advance(bytesWritten);
        }
```
Is solving this race condition worth the usability cost? The race condition occurs only if somebody writes very bad code (tries to use IOutput after a request (window) completed. And the usability hit is very severe (pinning you proposed above) or we require users to manually dispose, which I am pretty sure will result actual problems (leaks) in many apps.

jkotas: Yeah, there should be constructor or factory method call somewhere in that line - depends on what the convention expansion will look like exactly.

jkotas: > a problem that is very rare 

Rare problems still turn into a good exploits.

> Is solving this race condition worth the usability cost?

There is usability cost either way.

C# has very nice property today that you cannot introduce a use-after-free buffer overrun if you stay in safe C# and safe set of libraries (e.g. not calling anything under System.Runtime.InteropServices).

If we are changing this rule, we need to be explanation what is safe vs. dangerous that people need very careful about, and be prepared to build tooling for tracing down cases where somebody missed it. These needs to be done in more general context than just ASP.NET pipeline because of I expect that these APIs are going to be generally usable.

jkotas: The solution without language/runtime enforced safety seems to have a strong managed/unmanaged interop bias. So maybe a reasonable compromise is:

- Move Memory/OwnerMemory to System.Runtime.InteropServices namespace. This namespace is full of similar types that you need to be careful about using correctly.
- Make it non-generic, for byte buffers only. Generics do not make a lot of sense for maanaged/unmanaged interop.
- Consider reusing the existing `System.Runtime.InteropServices.SafeBuffer` for OwnedMemory because they are essentially the same types.

davidfowl: This is an interesting discussion. The point of this entire effort is to keep the performance we have today without many of the escape hatches (basically improving the primitives of the runtime and language to make writing idiomatic code without resorting to pointer manipulation). I'm not willing to make tradeoffs without data, so as we start porting kestrel over and changing to patterns we feel will have the best performance characteristics, we can measure and see where to draw the line. If we end up having to do all of the same things we did before in the kestrel codebase itself, then we've failed.

Time will tell 😄

joshfree:
@terrajobst can you also update the parsing.md to include the relative priorities we discussed over email:

Parsing
-	(pri 0) performance tests and improvement
-	(pri 0) support fast lower case and upper case hexadecimal parsing
-	(pri 1) remove old APIs
-	(pri 1) remove EncodingData(byte[][] symbols, TextEncoding encoding, Tuple<byte, int>[] parsingTrie)
-	(pri 1) add all primitive parser to PrimitiveParser (e.g. GUID, URI, date time, etc.)
-	(pri 1) add support for multi-span parsing
-	(pri 2) support ASCII
-	(pri 2) clean up code
-	(pri 2) full test coverage
-	(pri 2) support all formats
-	(pri 2) provide EncodingData.InvariantUtf16BE
-	(pri 2) write custom EncodingData generation tool

docs/specs/parsing.md
terrajobst: Added

terrajobst:
While we're still fine tuning the spec, let's merge what we have so far.

PR #1013, opened 2016-11-22
Move Span out of corefxlab.

AtsushiKan:


benaadams:
Woop! 👍

src/System.Slices/System/SpanExtensions.Legacy.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+namespace System
+{
+    public static partial class SpanExtensions
```
KrzysztofCwalina: I will remove these.

KrzysztofCwalina:
OK, there are some additional cleanup work items we need to do, but I will merge it for now and we can do the cleanup later. 
Thanks for doing this work!

PR #1016, opened 2016-11-23
Add harness and script to run xunit-performance tests

dpodder:


dpodder:
@brianrob @jorive @KrzysztofCwalina @botaberg PTAL

perf.cmd
```diff
@@ -0,0 +1,10 @@
+@echo off
```
KrzysztofCwalina: This file does not seem to be mentioned/described in the instructions. Is that intended? i.e. Should we use this file to run perf tests? 
Also, if yes, could we rename the file to something more descriprive, e.g. testperfromance.cmd

scripts/PerfHarness/PerfHarness.cs
```diff
@@ -0,0 +1,34 @@
...
+        // Assume test assemblies are colocated/restored next to the PerfHarness.
+        return Path.Combine(
+            Path.GetDirectoryName(Assembly.GetEntryAssembly().Location),
+            $"{testName}.dll"
+        );
+    }
+
+    private static string[] EnumeratePerfTests()
+    {
+        return new [] {
```
KrzysztofCwalina: I assume longer term you will just discover new test assemblies.

For now, could you add a note to the instructions saying that this array needs to be modified to add new tests to new assembly.

scripts/PerfHarness/PerfHarness.cs
```diff
@@ -0,0 +1,34 @@
...
+            "System.Binary.Base64.Tests",
```
KrzysztofCwalina: Should I assume longer term you will just discover tests automatically?

KrzysztofCwalina:
Thanks! Just one small comment.

scripts/PerfHarness/PerfHarness.cs
brianrob: The list here is actually the list of test assemblies and not tests.  Thus, we won't discover test assemblies, but once we know which are the test assemblies, we'll discover the tests within them (that works today).  Perhaps we should change this to "GetTestAssemblies()".

perf.cmd
brianrob: You can use this to run perf tests.  I suspect @dpodder was looking for ways to make things simple, but you should know that in the end, all you really need to do is dotnet run the PerfHarness exe, which will do all of the work, including pulling all of the dependencies required to actually run.  No more having to cherry-pick dependencies into a single directory to run.

brianrob:
LGTM as long as @KrzysztofCwalina's comments are addressed.

perf.cmd
dpodder: Right, the file is mostly here to serve as an example of just what you need to do: restore, then run, the PerfHarness project. I'm happy to rename the file, or outright remove it if it's more confusing than helpful.

scripts/PerfHarness/PerfHarness.cs
dpodder: Longer term this could certainly be updated to do some amount of self-discovery; I just kept that out for now to keep the focus on how the harness API is designed to work. Self-discovery itself can be implemented a number of ways, and that is discussion which I believe warrants its own separate thread. To make sure things are clear, I documented the requirements for adding new tests in the README file that I've also added in this PR.

KrzysztofCwalina:
Looks good. Thanks!

PR #1019, opened 2016-11-23
Added null checks around _method, _path and _httpVersion

niemyjski:


niemyjski:
Didn't see that PreservedBuffer was a struct.

PR #1021, opened 2016-11-23
Add support for cancelling the pending ReadAsync call

### LINKED ISSUE #1017: Consider adding CancellatonToken to ReadAsync or Cancel method to IPipelineReader
Right now, to force the await to continue, the caller needs to do a 0 byte write. This is a little bit hacky since that write needs to be synchronized with other writes into the same PipelineReaderWriter. The concern with a cancellation token passed into read is the allocation per read to register the callback that fires on cancellation. A cancel method would do the same thing but would avoid the per read allocation and be synchronized with the writer. The downside to that is that it's a different pattern that we currently use in .NET.

davidfowl:
- We need this to support things like graceful shutdown in kestrel
and to support yeilding the await in SignalR (for long polling)
- Unlike a 0 byte write, this is synchronized with the other state
in the PipelineReaderWriter and UnownedBufferReader so there are no
races/clashes
- Added IsCancelled to ReadResult so the caller can see what triggered
the await.

Fixes #1017

TODOs:
- Potentially look at the cancellation token approach as an extension method (less optimized by more idomatic)
- Maybe throw operation cancelled exception?
- Interpret 0 byte write as end of pipeline (more compatible with streams).

/cc @anurse @pakrym @halter73 @benaadams

### DIFF
--- a/src/System.IO.Pipelines/PipelineReaderWriter.cs
+++ b/src/System.IO.Pipelines/PipelineReaderWriter.cs
@@ -36,6 +36,7 @@ namespace System.IO.Pipelines
         private int _consumingState;
         private int _producingState;
         private object _sync = new object();
+        private bool _readingIsCancelled;
 
         // REVIEW: This object might be getting a little big :)
         private readonly TaskCompletionSource<object> _readingTcs = new TaskCompletionSource<object>();
@@ -366,19 +367,21 @@ namespace System.IO.Pipelines
             // Reading commit head shared with writer
             lock (_sync)
             {
-                if (!examined.IsDefault &&
+                if ((!examined.IsDefault &&
                     examined.Segment == _commitHead &&
                     examined.Index == _commitHeadIndex &&
-                    Reading.Status == TaskStatus.WaitingForActivation)
+                    Reading.Status == TaskStatus.WaitingForActivation) || _readingIsCancelled)
                 {
                     Interlocked.CompareExchange(
                         ref _awaitableState,
                         _awaitableIsNotCompleted,
                         _awaitableIsCompleted);
                 }
+
+                _readingIsCancelled = false;
             }
 
-            while (returnStart != returnEnd)
+            while (returnStart != null && returnStart != returnEnd)
             {
                 var returnSegment = returnStart;
                 returnStart = returnStart.Next;
@@ -465,6 +468,21 @@ namespace System.IO.Pipelines
         }
 
         /// <summary>
+        /// Cancel to currently pending call to <see cref="ReadAsync"/> without completing the <see cref="IPipelineReader"/>.
+        /// </summary>
+        public void CancelPendingRead()
+        {
+            // TODO: Can factor out this lock
+            lock (_sync)
+            {
+                // Mark reading is cancellable
+                _readingIsCancelled = true;
+
+                Complete();
+            }
+        }
+
+        /// <summary>
         /// Asynchronously reads a sequence of bytes from the current <see cref="IPipelineReader"/>.
         /// </summary>
         /// <returns>A <see cref="ReadableBufferAwaitable"/> representing the asynchronous read operation.</returns>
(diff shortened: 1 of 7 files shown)


src/System.IO.Pipelines/ReadResult.cs
```diff
@@ -3,16 +3,31 @@
...
 
+        /// <summary>
+        /// The <see cref="ReadableBuffer"/> that was read
+        /// </summary>
         public ReadableBuffer Buffer { get; }
 
+        /// <summary>
+        /// True if the currrent read was cancelled
+        /// </summary>
+        public bool IsCancelled { get; }
```
pakrym: We can store this as enum and use comparison to implement bool properties.

src/System.IO.Pipelines/UnownedBufferReader.cs
```diff
@@ -194,12 +195,13 @@ void IPipelineReader.Advance(ReadCursor consumed, ReadCursor examined)
             }
 
             // Again, we don't need an interlock here because Read and Write proceed serially.
-            if (!examined.IsDefault &&
+            if ((!examined.IsDefault &&
```
pakrym: This check looks to bulky. Split into local variables?

src/System.IO.Pipelines/PipelineReaderWriter.cs
```diff
@@ -465,6 +468,21 @@ private void SignalWriter(Exception exception)
         }
 
         /// <summary>
+        /// Cancel to currently pending call to <see cref="ReadAsync"/> without completing the <see cref="IPipelineReader"/>.
```
halter73: This also effectively cancels the next read if none is currently pending, no?

src/System.IO.Pipelines/ReadResult.cs
halter73: Make sure to use HasFlag 😉

src/System.IO.Pipelines/UnownedBufferReader.cs
halter73: There seems to be quite a bit of duplication between PipelineReaderWriter and UnownedBufferReader.

halter73: Never mind. Even this section is more different than I initially thought.

src/System.IO.Pipelines/PipelineReaderWriter.cs
```diff
@@ -366,19 +367,21 @@ public void AdvanceReader(ReadCursor consumed, ReadCursor examined)
             // Reading commit head shared with writer
             lock (_sync)
             {
-                if (!examined.IsDefault &&
+                if ((!examined.IsDefault &&
                     examined.Segment == _commitHead &&
                     examined.Index == _commitHeadIndex &&
-                    Reading.Status == TaskStatus.WaitingForActivation)
+                    Reading.Status == TaskStatus.WaitingForActivation) || _readingIsCancelled)
```
halter73: I don't think you should have merged this. I'm pretty sure this is the opposite of what you want. Here's what could happen:

1. Data arrives, ReadAsync() completes because of the new data. (i.e. _awaitableState is called invoking consumer code).
2. Before Advance() is called, CancelPendingRead() is called (probably on another thread, but could be the consuming thread).
3. Advance() is called, but it didn't examine up to the commit head.
4. The next call to await ReadAsync(), instead of completing immediately like one would expect, hangs until more data arrives (which may never happen). This is because `_readingIsCancelled` was true during the last call to Advance() causing the `_awaitableState` to be transitioned to `_awaitableIsNotCompleted`. Now IReadableBufferAwaiter.IsCompleted is false.

You should add a test for calling CancelPendingRead() between ReadAsync() and Advance().

davidfowl: Adding a test and fixing the bug. Nice catch

PR #1022, opened 2016-11-23
Reduce console output in PrimitiveParserPerfTests

botaberg:
As it stood, each perf test would spew out 1000 lines of fairly useless text to the console. This change removes this console output and replaces it with a single line expressing the name of the test as well as the string argument passed to it.


botaberg:
@KrzysztofCwalina PTAL

PR #1026, opened 2016-11-23
Added Memory.md section on safety

KrzysztofCwalina:
cc @davidfowl, @jkotas, @benaadams, @jaredpar, @joshfree, @botaberg, @vancem, @AtsushiKan

### DIFF
--- a/docs/specs/memory.md
+++ b/docs/specs/memory.md
@@ -231,6 +231,7 @@ The main negatives of this approach are additional complexity and larger size of
 The current prototype in the corfxlab repo implements this feature: https://github.com/dotnet/corefxlab/blob/master/src/System.Slices/System/Buffers/Memory.cs#L14
 
 ### IOwnedMemory\<T\>
+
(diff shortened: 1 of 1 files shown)


benaadams:
See how reference counting helps with outstanding Memory; but not sure how reference counting helps with outstanding Spans?

Span has no Dispose and it would need some kind of "out of scope" destructor addition to the runtime?

KrzysztofCwalina:
The point is that we can detect and fail with an error if OwnedMemory<T> is Disposed while ref count is > 0.

benaadams:
Ah; looked at the code that's pretty horrible 😁

Also isn't greatly safe
```csharp
var reservation = memory.AddReference();
var span = reservation.Span;
reservation.Dispose();
// trouble, no tooled compile warning about not disposing
var item = span[0];
```

benaadams:
Memory and its functions can be protected; but don't think there is much that can be done about Span itself

jkotas:
Right, there would need to be C# language support to make this look reasonable and to make the use after free patterns prohibited.

PR #1028, opened 2016-11-24
InvariantUtf8.TryParseUInt64

botaberg:
Add implementations for PrimitiveParser.InvariantUtf8.TryParseUInt64. Add performance tests for UInt64 as well. Refactor InvariantUtf8 class as well as the perf tests, consolidating some static members and helper methods in one file.


botaberg:
@KrzysztofCwalina PTAL

davidfowl:
This seems to have broken the following test:

```
     System.IO.Pipelines.Tests.ReadableBufferFacts.GetUInt64GivesExpectedValues [FAIL]
02:25:32       Assert.Equal() Failure
02:25:32       Expected: 10
02:25:32       Actual:   16
02:25:32       Stack Trace:
02:25:32         
```

botaberg:
@davidfowl Seems like this could be an issue with ReadableBufferExtensions.GetUInt64. This method uses the Hex version of TryParseUInt64. Is this intended? Before @KrzysztofCwalina 's port (https://github.com/dotnet/corefxlab/commit/24a135d395c2d376df0dc3a059023dcb3803a62f), it used the InternalParser's version and passed in TextFormat.HexUppercase, but the internal parser never had an implementation of hex parsing and ignored that parameter entirely.

PR #1030, opened 2016-11-24
Fix downgrades by updating to the latest corefx packages

### LINKED ISSUE #1018: Fix downgrade warnings in the build
We have 186 downgrade warnings since the move to System.Memory. This is because System.Memory is using the latest corefx dependencies but corefxlab is downgrading them to stable deps:

image

Can we make System.Memory use stable dependencies? Or should we make corefxlab use the latest dependencies as well?

/cc @weshaggard @joshfree

davidfowl:
- Turns out downgrades are bad and made it impossible to consume in the
current state outside of corefxlab.
- This PR also minimizes the individual references and uses the latest NetStandard.Library
package for core deps.

Fixes #1018

/cc @KrzysztofCwalina  (sorry bro 😄 )

### DIFF
--- a/src/System.Binary.Base64/project.json
+++ b/src/System.Binary.Base64/project.json
@@ -20,10 +20,7 @@
     "keyFile": "../../tools/Key.snk"
   },
   "dependencies": {
-    "System.Runtime": "4.1.0",
-    "System.Diagnostics.Debug": "4.0.11",
     "System.Slices": { "target": "project" },
-    "System.Memory": "4.4.0-beta-24721-02",
     "System.Text.Primitives": { "target": "project" }
   },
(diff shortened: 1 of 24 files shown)


KrzysztofCwalina:
Bye all the platfroms left behind :cry:

jkotas:
> Bye all the platforms left behind 

Why is this leaving platforms behind? 

nuget package version does not say anything about supported platforms, and I believe we are making sure that newer nuget package versions includes support for all platforms that the older version supported. (There are way too many version numbers involved and it is super hard to understand....)

davidfowl:
Jan is correct. Older platforms still work fine.

PR #1038, opened 2016-11-30
Port Seek methods from kestel

pakrym:
@davidfowl @halter73 @benaadams

### DIFF
--- a/src/System.IO.Pipelines/SeekExtensions.cs
+++ b/src/System.IO.Pipelines/SeekExtensions.cs
@@ -47,7 +47,9 @@ namespace System.IO.Pipelines
                     following = block.End - index;
                 }
                 ArraySegment<byte> array;
-                Debug.Assert(block.Memory.TryGetArray(out array));
+                var getArrayResult = block.Memory.TryGetArray(out array);
+                Debug.Assert(getArrayResult);
+
                 while (following > 0)
                 {
                     // Need unit tests to test Vector path
@@ -144,7 +146,9 @@ namespace System.IO.Pipelines
                     following = block.End - index;
                 }
                 ArraySegment<byte> array;
-                Debug.Assert(block.Memory.TryGetArray(out array));
+                var getArrayResult = block.Memory.TryGetArray(out array);
+                Debug.Assert(getArrayResult);
+
                 while (following > 0)
                 {
                     // Need unit tests to test Vector path
@@ -235,7 +239,8 @@ namespace System.IO.Pipelines
                     following = block.End - index;
                 }
                 ArraySegment<byte> array;
-                Debug.Assert(block.Memory.TryGetArray(out array));
+                var getArrayResult = block.Memory.TryGetArray(out array);
+                Debug.Assert(getArrayResult);
 
                 while (following > 0)
                 {
@@ -341,7 +346,8 @@ namespace System.IO.Pipelines
                     following = block.End - index;
                 }
                 ArraySegment<byte> array;
-                Debug.Assert(block.Memory.TryGetArray(out array));
+                var getArrayResult = block.Memory.TryGetArray(out array);
+                Debug.Assert(getArrayResult);
 
                 while (following > 0)
                 {
--- a/src/System.IO.Pipelines/SeekExtensions.cs
+++ b/src/System.IO.Pipelines/SeekExtensions.cs
@@ -1,0 +1,468 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    using Diagnostics;
+    using Numerics;
+    using Runtime.CompilerServices;
+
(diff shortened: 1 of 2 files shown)


KrzysztofCwalina:
Coincidentally, I was just on email thread with @jkotas about Span support for vectors. Here is code sample from Jan showing how to convert a span to a vector:
```c#
    public static partial class SpanExtensions
    {
        public static Vector<T> ReadVector<T>(this ReadOnlySpan<T> slice)
        {
            if (slice.Length < Vector<T>.Count)
            return Unsafe.As<T, Vector<T>>(ref slice.DangerousGetPinnableReference());
        }
    }
```
I think it would be great to write a reusable routine that finds a byte in Span<byte> and then use it from this code, or at least use the cast above instead of asserting that the span is an array.

pakrym:
@KrzysztofCwalina do you think we should also switch to iterating span directly instead of pinning and unsafe pointers?

benaadams:
Extension methods?  😢 

Need some by ref extensions or its going to be painful https://github.com/dotnet/roslyn/pull/15535

KrzysztofCwalina:
@pakrym, I am pretty sure it would regress some Kestrel code unacceptably, but I think we need to always try to use spans, if perf is unacceptable, use byte* as a workaround, and file an issue to fix Span<T>. When the issue is fixed, remove the workaround. 
Without this, it's a bit of chicken and egg problem. When I talk to the JIT team, they (rightfully so) ask for real code (i.e. not a microbenchmark) that shows bad perf, before they are willing to invest in expensive optimizations.

benaadams:
Extension methods?  😢 

ReadCursor is 16 bytes, probably some by ref extensions or its going to hit perf https://github.com/dotnet/roslyn/pull/15535 

Obviously the ReadableBuffer extensions will be worse as they are 48 bytes

/cc @jaredpar

Drawaes:
By ref extensions are critical when using structs to reduce allocations.
What is the plan for this, I mean VB has had it for how long?

pakrym:
@benaadams @Drawaes it's in PR (https://github.com/dotnet/roslyn/pull/15535). It takes some time to get language feature in.

KrzysztofCwalina:
@jaredpar, @MadsTorgersen, I know we talked about it, but I forgot: what's the pla for by ref this extension methods?

jaredpar:
@KrzysztofCwalina by ref extension methods are one of the features I have under the `Span<T>` umbrella.  I agree it's very necessary to have it included.

pakrym:
@KrzysztofCwalina `return Unsafe.As<T, Vector<T>>(ref slice.DangerousGetPinnableReference());` requires `ref returns` and C# which as far as I understand is not supported in CLI which we are using to build corefxlab.

jaredpar:
The full list of items I have under the `Span<T>` umbrella:

1. ref extension methods on value types
1. readonly ref return
1. readonly struct support
1. ref like type support 
1. The `Span<T>` / `Memory<T>` return safety issues

It's really hard to separate out a lot of these features.  For example 1-3 really need to be considered as a whole as they really impact the design of each other.  Likely we'll end up doing those as one whole feature.

benaadams:
> 1. ref extension methods on value types

Surely that's just VB.NET interop and parity, someone wrote a blog about it 😜 
https://blogs.msdn.microsoft.com/jaredpar/2008/12/11/extension-methods-and-byref-targets/

davidfowl:
@KrzysztofCwalina we need to have a design meeting about ReadableBuffer and ReadCursor.

KrzysztofCwalina:
@terrajobst, do we have time to discuss ReadableBuffer design at the next design meeting?

Drawaes:
I have some code examples of why the readable buffer API is a pain vs the writable buffer.  Picture this common scenario
https://github.com/Drawaes/corefxlab/blob/ManagedTls/src/System.IO.Pipelines.Networking.Tls/ManagedConnectionContext.cs#L142-L149

vs something like this for writing

https://github.com/Drawaes/corefxlab/blob/ManagedTls/src/System.IO.Pipelines.Networking.Tls/Managed/Handshake/HandshakeServerHelloWriter.cs#L15-L30

So as you can see the writable is a lot cleaner without having to slice all the time. But with the readable it involves a lot of slicing (and thus as @benaadams pointed out 18bytes + of copy).

_incase you are wondering that is a managed TLS_

jaredpar:
@benaadams 

> Surely that's just VB.NET interop and parity, someone wrote a blog about it 

I actually wrote that blog post as a fun poke at some of my coworkers.  At the time we were debating whether it was good practice or not to have extension methods which allowed null receivers.  I was in the camp of "no that's bad".  Hence I wrote that post as a joke about taking it to the next level. 

Little did I know that years later I'd be pushing for it as a feature :smile:

aL3891:
Any update on this? would love to see it merged :)

davidfowl:
Merging this so we can move forward with kestrel

PR #1045, opened 2016-12-06
Libuv Udp

grahamehorner:
A sample & proof of concept for Udp using Libuv

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -2,7 +2,7 @@
     [string]$Configuration="Debug",
     [string]$Restore="true",
     [string]$Channel="preview",
-    [string]$Version="1.0.0-preview2-003121"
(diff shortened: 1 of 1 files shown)


global.json
```diff
@@ -1,6 +1,6 @@
 {
   "projects": [ "src", "tests" ],
   "sdk": {
-    "version": "1.0.0-preview2-003121"
+    "version": "1.0.0-preview2-003131"
```
davidfowl: Don't change this unless you change the build script too.

grahamehorner:
Apologies & good spot thought I'd reversed this change

davidfowl:
@grahamehorner this is now super out of date. Do you mind taking another stab at it?

grahamehorner:
@davidfowl by happy to given the churn, any suggestion/guidance welcome ;)

PR #1047, opened 2016-12-07
Added ReferenceCounter

KrzysztofCwalina:
ReferenceCounter is a small datastructure to store an manipulate reference counts
It stores counts on per thread basis, which allows us to manipulate the counts without synchronization.
When the count needs to be read, it does need to be tallied up from all the threads.
We compute the tally without synchronizing with the thread-local modifications, and so
there is a possibility of a race here. But, the race is much less likely to occur than today
when Span\<T\> is "checked-out" from Memory\<T\> without incrementing the OwnedMemory\<T\> refcount,
and so it's an improvement. Also, we think we can eliminate the impact of the race at higher level,
i.e. in OwnedMemory\<T\> (at some additional memory barrier cost).

But the performance wins are very large when compared to contended Interlocked reference counting.
A quick test incrementing and decrementing the count 1M times on 10 threads shows the following results:

ReferenceCounter 206ms
Interlocked      13458ms

BTW, on 50 threads, the interlocked test takes minutes to complete. The ReferenceCounter one scales close to linearly.
 
CC: @joshfree, @jkotas, @vancem, @mjp41, @davidfowl, @jaredpar 

In a subsequent PR, I will port the current OwnedMemory\<T\> reference counting to use this counter. Possibly adding memory barriers between Dispose and AddReference.

### DIFF
--- a/src/System.Slices/System/Buffers/ReferenceCounter.cs
+++ b/src/System.Slices/System/Buffers/ReferenceCounter.cs
@@ -1,0 +1,213 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Runtime
+{
+    /// <summary>
+    /// Make sure the struct is not copied, i.e. pass it only by reference
+    /// </summary>
+    /// <remarks>
+    /// The  counter is not completly race-free. Reading GetGlobalCount and AddReference/Release are subject to a race.
+    /// </remarks>
+    public struct ReferenceCounter
+    {
+        // thread local counts that can be updated very efficiently
+        [ThreadStatic]
+        static ObjectTable t_threadLocalCounts;
+
+        // all thread local counts; these are tallied up when global count is comupted
+        static ObjectTable[] s_allTables = new ObjectTable[Environment.ProcessorCount];
+        static int s_allTablesCount;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void AddReference(object obj)
+        {
+            var localCounts = t_threadLocalCounts;
+            if (localCounts == null)
+            {
+                localCounts = AddThreadLocalTable();
+            }
+            localCounts.Increment(obj);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Release(object obj)
+        {
+            var localCounts = t_threadLocalCounts;
+            localCounts.Decrement(obj);
+        }
+
+        // TODO: can we detect if the object was only refcounted on one thread and its the current thread? If yes, we don't need to synchronize?
+        public uint GetGlobalCount(object obj)
+        {
+            var allTables = s_allTables;
+            lock (allTables)
+            {
+                uint globalCount = 0;
+                for (int index = 0; index < s_allTablesCount; index++)
+                {
+                    globalCount += allTables[index].GetCount(obj);
+                }
+                return globalCount;
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public uint GetThreadLocalCount(object obj)
+        {
+            var localCounts = t_threadLocalCounts;
+            if (localCounts == null) return 0;
+            return localCounts.GetCount(obj);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static ObjectTable AddThreadLocalTable()
+        {
+            Debug.Assert(t_threadLocalCounts == null);
+            var localCounts = new ObjectTable();
+            t_threadLocalCounts = localCounts;
+            lock (s_allTables)
+            {
+                    var allTables = s_allTables;
+                    if (s_allTablesCount >= allTables.Length)
+                    {
+                        var newAllTables = new ObjectTable[allTables.Length << 1];
+                        allTables.CopyTo(newAllTables, 0);
+                        s_allTables = newAllTables;
+                        allTables = newAllTables;
+                    }
+                    allTables[s_allTablesCount++] = localCounts;
+            }
+            return localCounts;
+        }
+    }
+
+    // This datastructure holds a collection of object to reference count mappings
+    sealed class ObjectTable
+    {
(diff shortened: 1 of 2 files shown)


src/System.Slices/System/Buffers/ReferenceCounter.cs
```diff
@@ -0,0 +1,197 @@
...
+            int index;
+            for (index = 0; index < _items.Length; index++)
+            {
+                // if found the object, just return the index 
+                if (ReferenceEquals(_items[index].Obj, obj))
+                {
+                    return index;
+                }
+
+                // the slot is free, rememeber it; we want to use the first free slot we found, if the object is not found later
```
joshfree: typo

tests/System.Slices.Tests/ReferenceCounterTests.cs
```diff
@@ -0,0 +1,57 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System.Runtime;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.Slices.Tests
+{
+    public class ReferenceCounterTests
```
davidfowl: Add a test that hits the resize code paths, both for ReferenceCounter and for ObjectTable. 

The tests for both pieces should also be separate.

src/System.Slices/System/Buffers/ReferenceCounter.cs
```diff
@@ -0,0 +1,203 @@
...
+    /// The  counter is not completly race-free. Reading GetGlobalCount and AddReference/Release are subject to a race.
+    /// </remarks>
+    public struct ReferenceCounter
+    {
+        // thread local counts that can be updated very efficiently
+        [ThreadStatic]
+        static ObjectTable t_threadLocalCounts;
+
+        // all thread local counts; these are tallied up when global count is comupted
+        static ObjectTable[] s_allTables = new ObjectTable[4];
```
davidfowl: Why 4 tables? Should it match number of Cores?

src/System.Slices/System/Buffers/ReferenceCounter.cs
```diff
@@ -0,0 +1,203 @@
...
+            var localCounts = t_threadLocalCounts;
+            if (localCounts == null) return 0;
+            return localCounts.GetCount(obj);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void AddNewObject(object obj)
+        {
+            var localCounts = t_threadLocalCounts;
+            lock (s_allTables)
```
davidfowl: Will this be a problem? Should this be lockfree? I guess it only happens the very first time you up the ref count on a thread.

src/System.Slices/System/Buffers/ReferenceCounter.cs
```diff
@@ -0,0 +1,203 @@
...
+            {
+                AddNewObject(obj);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Release(object obj)
+        {
+            var localCounts = t_threadLocalCounts;
+            localCounts.Decrement(obj);
```
davidfowl: Guard against null? Should never happen so maybe a debug assert?

tests/System.Slices.Tests/ReferenceCounterTests.cs
KrzysztofCwalina: will do.

src/System.Slices/System/Buffers/ReferenceCounter.cs
KrzysztofCwalina: good point

src/System.Slices/System/Buffers/ReferenceCounter.cs
KrzysztofCwalina: Yeah, I still don't have a well formed opinion on how much error checking we should be doing in our high performance libraries. The CLR already guards against null here, so what's the point of more guards? The only benefit I see is better exception message, but then I am not sure how much it's worth it. If I do add the check, then I have to play games with the inliner (as it does not like throw). Anyway, as I said I don't have an opinion that is firm here, so I can go either way.

src/System.Slices/System/Buffers/ReferenceCounter.cs
KrzysztofCwalina: It has to be thread safe. I might try to make it use lighter weight synchronization, but the code (the array based datastructure) is already super complex. Also, this does not even happen the first time the ref count is upped. It happens the first time an object is upped. If the object is returned to the pool and then reused (while its slot is not reused), the next time the object is reused an upped from 0 to 1, it will already be in the array (which BTW might cause leaks, and so we should discuss).

src/System.Slices/System/Buffers/ReferenceCounter.cs
```diff
@@ -0,0 +1,203 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public uint GetThreadLocalCount(object obj)
+        {
+            var localCounts = t_threadLocalCounts;
+            if (localCounts == null) return 0;
+            return localCounts.GetCount(obj);
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static void AddNewObject(object obj)
```
mjp41: What is the intent of this method? Why does it only do something if the thread static is null?

src/System.Slices/System/Buffers/ReferenceCounter.cs
```diff
@@ -0,0 +1,203 @@
...
+        {
+            var localCounts = t_threadLocalCounts;
+            lock (s_allTables)
+            {
+                if (localCounts == null)
```
mjp41: Should you perform this test outside the lock?  No need to grab it if you aren't doing anything.

src/System.Slices/System/Buffers/ReferenceCounter.cs
mjp41: Okay. I get this now.  It is not intended to be called from anywhere except AddReference. Perhaps, make it private?

tests/System.Slices.Tests/ReferenceCounterTests.cs
KrzysztofCwalina: done

PR #1049, opened 2016-12-08
Go back to stable versions

davidfowl:
- Now that System.Memory and Unsafe have been re-baselined
to stable dependencies, we can go back to using RTM bits for
corefx.
- Now we float for System.Memory and System.CompilerServices.Unsafe

/cc @KrzysztofCwalina 

I'm going go merge this when the CI passes. It'll unblock ASP.NET pulling new bits and will make @KrzysztofCwalina happy 😄


KrzysztofCwalina:
looks good.

PR #991, opened 2016-11-16
Tls Lib for Pipelines

Drawaes:
An attempt to make a low allocation, wicked fast TLS layer for Pipelines. Xplat, and server / multiple connection optimised...yeah

Issue #1060, opened 2016-12-12
Stitching multiple readable buffers together


Drawaes:
In TLS there is two layers of framing. The Record frame and the protocol frame. A record frame is wrapped inside the protocol frame. However the record frame can be much larger than the record frame. Therefore I need to be able to unwrap and decrypt the record fragments in place.

I then need to slice off the headers, and trailers to get the actual record data. At this point if I don't have the complete frame I will need to preserve and check multiple of these. It would be nice if at this point I didn't need to construct a collection or my own linked list but instead could append the start cursor of the second list to the end cursor of the first. 

I am not sure if this is possible today but as I believe they are linked lists of buffers internally this should be possible.

davidfowl:
Code samples would help me understand better here. Is this across multiple calls to ReadAsync? 
 
       ReadAsync()                ReadAsync()
[ header | payload ] -> [ header | payload ]

Is that what it looks like?  You want to string multiple payloads together into a single readable buffer?

Drawaes:
The other problem with the internal channel is that the process needs to be synchronised. Async won't work because the current payload may change the encryption state which will change the way the next [header] frame will be handled (encrypted or not, renegotiated keys etc)

One way to solve it but seems unclean is to hold a list of buffers that makes up the current payload. Obviously then I have to enumerate that and the readables which is a mess.

Drawaes:
@davidfowl does this kinda help (P.S. don't mention spacing, var it's mock code)
[Gist of Demo](https://gist.github.com/Drawaes/009d39f80ba6caf045b2b8372d4917fa)

Drawaes:
What would be a solution is if You could build an I SEQUENCE from a bunch of preserved readablebuffers. So in effect you would end up with a selectmany to use linq terminology

davidfowl:
> (P.S. don't mention spacing, var it's mock code)

My eyes

Drawaes:
I have got past this by making an internal channel for now. It's really a
sync operation so it's a bit over kill but it works

Drawaes:
[Managed TLS using an internal handshake channel](https://github.com/Drawaes/corefxlab/blob/ManagedTlsV3/src/System.IO.Pipelines.Networking.Tls.Managed/SecurePipeline.cs#L56-L64)

Shows how I have used an internal channel. This might actually be preferable if a new handshake happens during communication (renegotiation for certs etc).

Drawaes:
[...]

PR #1062, opened 2016-12-12
TLS Lib for Pipelines Take 2

Drawaes:
Reference #991

Now we know the issues that can be fixed with more design. The interop layer needs to be ported to the same as used in corefx. If I can get some help to copy that over into the library (what is there now) and get it compiling, I can then port over my pinvoke to that. The difficulty will be that I don't want to removed that custom bio, and avoid too much of it going to C++ if we possibly can. That part is key to the performance on OpenSsl


KrzysztofCwalina:
@davidfowl, @bartonjs

Drawaes:
Needs to use "corefx" interop wrappers

Drawaes:
Needs to handle multiple TLSPlainText blocks per record and TLSPlainText blocks fragmented over multiple records

Re : [rfc5246 section 6.2.1](https://tools.ietf.org/html/rfc5246#section-6.2.1)

Drawaes:
Potential solution is issue #1060

Drawaes:
I have a second set of demo code available. Currently it is windows only, I will finish up a couple of more cipher suite and extensions support. It uses Bcrypt/CNG/NCrypt but the rest of the protocol handling is managed

https://github.com/Drawaes/corefxlab/tree/ManagedTlsV2/src/System.IO.Pipelines.Networking.Tls.Managed

@blowdart I am not recommending you lot pic this up, but I am putting it out there as an alternative idea.

src/System.IO.Pipelines.Networking.Tls/CipherInfo.cs
```diff
@@ -0,0 +1,8 @@
+﻿namespace System.IO.Pipelines.Networking.Tls
+{
+    public struct CipherInfo
+    {
+        public string Name;
+        public int KeySizeInBits;
```
Drawaes: Yes that was a hang over from testing/checking negotiated ciphers during perf testing. I will remove it

src/System.IO.Pipelines.Networking.Tls/Internal/OpenSsl/CustomBio.cs
```diff
@@ -0,0 +1,256 @@
...
+        private static class WindowsLib
+        {
+            public const string CryptoDll = "libeay32.dll";
+            [DllImport(CryptoDll, CallingConvention = CallingConvention.Cdecl)]
+            public extern static void BIO_set_flags(ref bio_st bio, BioFlags flags);
+        }
+
+        private static class UnixLib
+        {
+            public const string CryptoDll = "libcrypto.so";
```
Drawaes: The plan here I believe is to "copy" (because for now its not in corefx but labs) all of the shim code across from the current crypto and reuse that.

src/System.IO.Pipelines.Networking.Tls/Internal/OpenSsl/CustomBio.cs
```diff
@@ -0,0 +1,256 @@
...
+            bio.num = -1;
+            return 1;
+        }
+
+        private static int WriteBio(ref bio_st bio, void* buff, int numberOfBytes)
+        {
+            var bufferHandle = GCHandle.FromIntPtr((IntPtr)bio.ptr);
+            var bufferPipeline = (IPipelineWriter)bufferHandle.Target;
+            int numberOfBytesRemaing = numberOfBytes;
+            var buffer = bufferPipeline.Alloc();
```
Drawaes: Agreed, as explained this was changed because the ASP.NET team wanted it, when it was in that domain. I am in line with var only where it is clear what the "var" will be

src/System.IO.Pipelines.Networking.Tls/Internal/OpenSsl/CustomBio.cs
```diff
@@ -0,0 +1,256 @@
...
+                int sizeToWrite = Math.Min(MaxBlockSize, numberOfBytesRemaing);
+                buffer.Ensure(sizeToWrite);
+                var writeSpan = new Span<byte>(buff, sizeToWrite);
+                writeSpan.CopyTo(buffer.Memory.Span);
+                buffer.Advance(sizeToWrite);
+                numberOfBytesRemaing -= sizeToWrite;
+            }
+            buffer.Commit();
+            bio.num_write = (uint)numberOfBytes;
+            BIO_set_flags(ref bio, BioFlags.BIO_FLAGS_WRITE | BioFlags.BIO_FLAGS_SHOULD_RETRY);
```
Drawaes: I need to recheck, but I stretching my memory back, there was an issue without it..... OpenSSL doc are.... not always right.

src/System.IO.Pipelines.Networking.Tls/Internal/OpenSsl/CustomBio.cs
```diff
@@ -0,0 +1,256 @@
...
+            return numberOfBytes;
+        }
+
+        public static int NumberOfWrittenBytes(InteropBio.BioHandle handle)
```
Drawaes: agreed.

src/System.IO.Pipelines.Networking.Tls/Internal/OpenSsl/CustomBio.cs
```diff
@@ -0,0 +1,256 @@
...
+            public int flags;                  /* extra storage */
+            public int retry_reason;
+            public int num;
+            public void* ptr;
+            public void* next_bio;// struct bio_st *next_bio;    /* used by filter BIOs */
+            public void* prev_bio;//struct bio_st *prev_bio;    /* used by filter BIOs */
+            public int references;
+            public ulong num_read;
+            public ulong num_write;
+        };
```
Drawaes: missing fields off the end? If so then it doesn't matter. If you want to abandon the custom bio that if fine, but then you are copying and the perf of the memory bio is terrible, you end up with a LOT of copying (if you copy x to the memory bio, then add bytes it makes a new array and copies the lot again).

src/System.IO.Pipelines.Networking.Tls/Internal/OpenSsl/Interop.cs
```diff
@@ -0,0 +1,451 @@
...
+            else
+            {
+                UnixLib.SSL_set_accept_state(ssl);
+            }
+        }
+        public static int SSL_write(IntPtr ssl, void* buf, int len) => IsWindows ? WindowsLib.SSL_write(ssl, buf, len) : IsOsx ? OsxLib.SSL_write(ssl, buf, len) : UnixLib.SSL_write(ssl, buf, len);
+        public static int SSL_read(IntPtr ssl, void* buf, int len) => IsWindows ? WindowsLib.SSL_read(ssl, buf, len) : IsOsx ? OsxLib.SSL_read(ssl, buf, len) : UnixLib.SSL_read(ssl, buf, len);
+        public static int SSL_do_handshake(IntPtr ssl) => IsWindows ? WindowsLib.SSL_do_handshake(ssl) : IsOsx ? OsxLib.SSL_do_handshake(ssl) : UnixLib.SSL_do_handshake(ssl);
+        public static SslErrorCodes SSL_get_error(IntPtr ssl, int errorIndetity) => IsWindows ? WindowsLib.SSL_get_error(ssl, errorIndetity) : IsOsx ? OsxLib.SSL_get_error(ssl, errorIndetity) : UnixLib.SSL_get_error(ssl, errorIndetity);
+        public static int SetKeys(IntPtr ctx, IntPtr certificates, IntPtr privateKey) => IsWindows ? WindowsLib.SetKeys(ctx, certificates, privateKey) : IsOsx ? OsxLib.SetKeys(ctx, certificates, privateKey) : UnixLib.SetKeys(ctx, certificates, privateKey);
```
Drawaes: because if there is no return value you can't use a single-line-double-ternary as far as I am aware, maybe you can

src/System.IO.Pipelines.Networking.Tls/OpenSslSecurityContext.cs
```diff
@@ -0,0 +1,141 @@
...
+            }
+        }
+
+        private unsafe Interop.AlpnStatus alpn_cb(IntPtr ssl, out byte* selProto, out byte selProtoLen, byte* inProtos, int inProtosLen, IntPtr arg)
+        {
+            byte* matchBuffer = (byte*)_alpnHandle.AddrOfPinnedObject();
+            for (int i = 0; i < inProtosLen;)
+            {
+                var len = inProtos[i];
+                var s = new Span<byte>(inProtos + i + 1, len);
```
Drawaes: size byte, could make it a const if needed or a comment

src/System.IO.Pipelines.Networking.Tls/Internal/OpenSsl/CustomBio.cs
Drawaes: As far as I am aware the flow always for these Bio objects is. You give it a template of the function pointers. Then you can only allocate via a Bio_New call then you can interact with that created object.

Drawaes:
TBH you are wasting your time going through all these little issues unless we can figure out the whole Shim/Interop SChannel etc concept. There is no point going into the finer detail. The OpenSsl shim won't work on windows which I get the MS corporate line on SChannel only. However it makes it difficult for me to write/dev/test, and a lot of work if I have to fire up linux containers and write code on them etc.

Until we figure out a solution or get some help from someone in that area, fixing each naming/style/etc comment seems pointless.

In the meantime, I am going to have a crack at a fully managed TLS with native crypto (OpenSsl/CNG). I have basic TLS 1.2 working, I have it using the Interop layer from CoreFX for CNG, once again I have it the OpenSSL no shim on windows issue so that is going direct for now.

One potential solution is if someone can show me how to compile the shim for openssl for windows that might be a solution for now (I in no way expect that this would be deployed but just used for dev).

bartonjs:
> One potential solution is if someone can show me how to compile the shim for openssl for windows 

That's never been tried.  We don't support the use of OpenSSL on Windows at all.  All parts of corefx using OpenSSL were tested on Linux.  (I would say "written", but I did the C# code writing on Windows and copied the binaries to Linux to test)

Drawaes:
>but I did the C# code writing on Windows and copied the binaries to Linux to test

That is basically what I want to do (I like my windows too much). But the flow for say running an XUnit test and step by step debugging, is it there or are you writing code blind and hoping for the best.

Not supporting OpenSsl on windows at all is a different issue. There are a number of really good reasons to do it, but I realize this is more a philosophical rather than a technical limitation of which @blowdart for instance has strong opinions on. But I really don't want to get into that here, for now I am happy to concede to SChannel on Windows/OpenSsl/SecureTransport, and to hide the windows/linux specific types inside and have a "provider" in front. The main reason for the OpenSsl Shim on windows is to easy development/testing of the core lib then shutting it off.

On the OpenSsl I might just open an issue against pipelines to discuss it.

bartonjs:
> But the flow for say running an XUnit test and step by step debugging, is it there or are you writing code blind and hoping for the best.

I'd say it's more "writing it like a service".  That is, add lots of logging.  Then, when done developing (unlike a service) rip out all the logging.

* College student: Taught printf()
* Intern: Interactive debuggers are great, you should really learn how to use them.
* New hire: NTSD is the best debugger ever.  Or, if you're a sissy, use WinDBG.
* A couple months later: That's great, but what you really need is Wireshark or Fiddler, see what the raw data was before some lower level library got in the way.
* Services developer: You can't ever attach a debugger to a running instance.  If you need to know something, log it.  If you think you might need to know something later, log it now.
* .NET Framework developer: You really should learn how to use SOS.
* .NET Core (Linux) developer: Use printf().

And, thus, my life has come full circle.


But, yeah, it's really like writing services.  You write a line of code, project evil spew from the Internet at it (in your mind), then write the next one.  And hope you have precisely one bug (about halfway through), because if it works on the first try it's suspicious.

Drawaes:
Yeah I hear that, try catch, log... nope not enough "I got to X" "I got to Y" logs are always painful. It's possible for me to do that, but then I would rather do something interesting with my nights. So here is the thing, by being able to dev on windows when I wrote this it was simple, then final testing on linux was a couple of tweaks and logging and it worked. That is why writing the openssl shim for windows for early dev work would make it all easier. Surely it's a recompile with binding to the windows version of OpenSsl?

KodrAus:
> the implementation should try to limit itself to SecureTransport's known limitations (like not supporting ALPN)

@bartonjs is a lowest common donominator approach the only option here? As an alternative could we provide intrinsics for whether or not certain features/extensions are supported and let callers enable them if they're available and needed?

bartonjs:
@KodrAus My comment about limiting to SecureTransport's limitations only applies to macOS.  The Windows and Linux versions are allowed to do whatever SChannel and OpenSSL can do.

Since we're trying to eliminate the OpenSSL dependency on macOS we should strive to not have a loss of functionality when this library switches from OpenSSL to SecureTransport.  (Being restrictive now and adding support for something later is way more palatable than restrictions creeping in later)

Drawaes:
Its out of date now, and unloved... putting it out of it's misery

PR #1064, opened 2016-12-13
Create pipelines-io.md

davidfowl:


docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
+## Efficient IO
+
+As part of the high performance work, we need new abstractions for doing efficient reading and writing to and from the network. The pattern we have settled on
+so far in pipelines is a model where buffers are pushed by a producer (`IPipelineWriter`) to a consumer (`IPipelineWriter`).
```
jkotas: Is IPipelineWriter really both the producer and the consumer?

davidfowl: oops

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+
+```C#
+private Task Produce(IPipelineWriter writer)
+{
+   while (true)
+   {
+       // Allocate a buffer from the writer
+       WritableBuffer buffer = writer.Alloc();
+       
+       var bytes = Encoding.UTF8.GetBytes(DateTime.Now.ToString());
```
KrzysztofCwalina: Should we be showing non-allocating APIs? i.e. use OutputFormatter<TOutput>

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+   public void Advance(int bytesWritten);
+   public void Commit();
+   public Task FlushAsync();
+}
+```
+
+There are extension methods like `Write(ReadOnlySpan<byte>)` which automate writing to the available memory and advancing the buffer.
+
+```C#
+public static void Write(this WritableBuffer buffer, ReadOnlySpan<byte> source)
```
KrzysztofCwalina: I think this method might become the pit of failure. People should be writing directly to WritableBuffer, not first to span and then copy to the buffer. I know there are cases where the data written is static and just needs to be copied, but: a) we should first show samples how to write directly to buffer, b) possibly rename this method to Copy, as it's meant only for copy scenarios.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+    {
+        ReadResult result = await reader.ReadAsync();
+        ReadableBuffer buffer = result.Buffer;
+        if (buffer.IsEmpty && result.IsCompleted)
+        {
+            break;
+        }
+
+        // Process the buffer
+        Consume(ref buffer);
```
KrzysztofCwalina: Why does it have to be passed by ref? Is it because the buffer is a large struct, or it needs to be mutated?

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+
+public struct PreservedBuffer : IDisposable
+{
+   public ReadableBuffer Buffer { get; }
+}
+```
+
+A `ReadCursor` represents a specific coordinate within a `BufferSegment` and a `ReadableBuffer` represents all of the data between 2 `ReadCursors`, Start and End. The `ReadableBuffer` is a point in time snapshot of the data the producer made available. 
+
+After processing the buffer, consumers call `IPipelineReader.Advance(consumed, observed)` indicate how much of the buffer was consumed and how much of the buffer was observed. The distinction is very important as consumed data can be returned to the system for re-use by another component. Marking observed data lets the consumer control over where it should be notified again (to an extent).
```
KrzysztofCwalina: Did you considered removing "observed" and assuming that everything in the buffer was observed? At least this being the default that could be explicitly changed, e.g. two overloads of Advance, one with used consumed.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+2. Multiple buffers. Today pipelines uses a linked list of buffers and provides a single abstraction over them, the `ReadableBuffer`. This multi buffer abstraction means all low level routines need to be implemented against a new multi buffer primitives or duplicated in pipelines.
+3. Structs or classes? Today `WritableBuffer` is tied to the underlying `IPipelineWriter` it was allocated from, this is to enable passing it to other methods without explicitly passing it by reference. This is very limiting as the type is now tied to the underlying writer. `ReadableBuffer` has the same problem. After peeking at bytes in the buffer, you have to make a new slice today because there's no good way to *move* the `ReadableBuffer`. This leads to lots of struct copies and `ReadableBuffer` is a big struct.
+4. Are parsers asynchronous or synchronous? In the model above, parsers are always stateful and synchronous. The code calling into the parser does the awaiting and calls the parser with more data when available. It means that in the future, we'd recommend people write synchronous stateful parsers (JSON.NET etc) and have asynchrony layered on top.
```
KrzysztofCwalina: Synchronous yes. But I am not sure about stateful. I implemented a prototype JSON parser where the caller needs to keep the state and re-pass it when it calls the parser again. If we get "no copy structs" then we can probably make the parsers stateful (as we will be able to make them structs and pass them around).

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+5. We need the primitives for reading and writing to be decoupled from pipelines. It should be possible to write to arbitrary scratch buffers of pooled/unpooled memory.
```
KrzysztofCwalina: Are they not decoupled? ISequence and IOutput are decoupled.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+6. Overlapping writes are a harder problem because writing isn't a single atomic call (Alloc(), Write(), Commit()). This isn't usually a *huge* problem because pipelines are single producer single consumer.
```
KrzysztofCwalina: You mean "pipes are single producer single consumer"? (not pipelines)

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
+## Efficient IO
+
+As part of the high performance work, we need new abstractions for doing efficient reading and writing to and from the network. The pattern we have settled on
```
jaredpar: Consider using I/O in general here, with networking as a specific example.  Pipelines should be applicable to all types of IO: file, IPC and network.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+so far in pipelines is a model where buffers are pushed by a producer (`IPipelineWriter`) to a consumer (`IPipelineReader`).
+
+```C#
+public interface IPipelineReader
+{
+    ReadableBufferAwaitable ReadAsync();
```
jaredpar: Do we need a cancellation token here?

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+    void CancelPendingRead();
+    void Complete(Exception exception = null);
+}
+```
+
+```C#
+public interface IPipelineWriter
+{
+    Task Writing { get; }
+    WritableBuffer Alloc(int minimumSize = 0);
```
jaredpar: Consider using the full word allocate here.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+    void Complete(Exception exception = null);
+}
+```
+
+The producer gets a handle on a `WritableBuffer` and the consumer gets a handle on a `ReadableBuffer` to do writes and reads respectively. 
+
+Memory in pipelines is managed as a linked list of `OwnedMemory<T>`:
+
+```C#
+public class BufferSegment
```
jaredpar: Consider using `sealed` here.  Given this is high performance code and JIT in some cases can take advantage of sealed it should be considered.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+    // The backing memory, usually leased from a pool
+    OwnedMemory<byte> Data { get; set; }
+    public int Start { get; set; }
+    public int End { get; set; }
+    
+    // The next segment
+    BufferSegment Next { get; set; }
+    
+    // The segment cannot be written to
+    public bool ReadOnly { get; set; }
```
jaredpar: Consider `IsReadOnly`.  This is the predominate pattern today in the BCL.  For example: `ICollection<T>.IsReadOnly`.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+    public int WritableBytes => Data.Length - End;
+}
+```
+
+A `BufferSegment` represents a link in a contiguous chain of buffers. New segments are added if the requested memory is less than the writable bytes in the current segment. What's interesting here is that there are 2 views to a `BufferSegment`, the ReadableBytes and the WritableBytes. The writing and reading primitives in pipelines build on this concept.
+
+### Writing
+
+```C#
+private Task Produce(IPipelineWriter writer)
```
jaredpar: Nit: `private async Task`

The body of the method uses `await` hence this needs to be marked as `async`

docs/specs/pipelines-io.md
davidfowl: Because it needs to be mutated. There's no way to move Start and End. It's just a pointer today.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+{
+    Task Writing { get; }
+    WritableBuffer Alloc(int minimumSize = 0);
+    void Complete(Exception exception = null);
+}
+```
+
+The producer gets a handle on a `WritableBuffer` and the consumer gets a handle on a `ReadableBuffer` to do writes and reads respectively. 
+
+Memory in pipelines is managed as a linked list of `OwnedMemory<T>`:
```
davidfowl: Will do

docs/specs/pipelines-io.md
davidfowl: That's fair.

docs/specs/pipelines-io.md
```diff
@@ -0,0 +1,196 @@
...
+### Threading
+
+When the producer calls `FlushAsync()` on the `WritableBuffer` that will call the consumer on the producer's thread.
+
+### Buffer ownership
+
+In the examples above, data is pushed from the producer to consumer. The consumer has ownership of the buffer between the call to `IPipelineReader.ReadAsync()` and `IPipelineReader.Advance()`. If the consumers needs ownership of the buffer outside of this scope, an explicit call to `ReadableBuffer.Preserve()` must be made before calling Advance. This will transfer ownership of the buffer to the consumer. `PreservedBuffers` must be disposed.
+
+## Problems
+1. When writing higher level protocol parsers, there is a need for efficient low level routines that can grab a byte at a time, and possible several bytes at a time. Further more, to build successful ecosystem of reusable parsing routines, we'd like these types of parsers to be built on the same primitives. Is `ReadableBuffer` that primitive or do we need something lower?
```
davidfowl: Because that's not a great API to use. IMO we need a type that looks more like `StreamReader` for `byte[]`.

docs/specs/pipelines-io.md
davidfowl: Yes pipes

docs/specs/pipelines-io.md
davidfowl: I specifically avoided that and added CancelPendingRead because of having to register the callback per `ReadAsync` call. See https://github.com/dotnet/corefxlab/issues/1017 for more details.

docs/specs/pipelines-io.md
davidfowl: I don't know why we still seal things. This is internal anyways. It's just to illustrate the architecture.

docs/specs/pipelines-io.md
davidfowl: We need examples of this. *Some* layer needs to be stateful.

docs/specs/pipelines-io.md
jaredpar: > I don't know why we still seal things. 

1. Performance: easier for code gen technologies (mostly JIT) to analyze sealed types. Result is it can emit better code at runtime.
2. Design: an unsealed type with no virtual methods doesn't really serve much of a purpose (other than to make JIT harder).

docs/specs/pipelines-io.md
KrzysztofCwalina: You need a language with no-copy struct support, or it's going to be a landmine field.

docs/specs/pipelines-io.md
davidfowl: It's not the exchange type we should be using for reading a bunch of buffers. It's like swapping `TextReader` for `IEnumerable<char[]>`.

KrzysztofCwalina: We should talk more, but I think exchange types, as their name implies, are just handles. They need to be good for passing around and representing general concepts. They don't need to be very usable. Stream is our current successful i/o exchange type and it's not very easy to use. We have TextReader building on top of stream to provide usability.
Having said that, I am very much open to improving ISequence. We should talk about what limitations we see today and what scenarios we would like it to support.

docs/specs/pipelines-io.md
KrzysztofCwalina: Would it be enough to poll? i.e. use IsCancelled?

docs/specs/pipelines-io.md
benaadams: Just checked this as always wondered... Sealed makes no difference; just annoys users, can't derive and add extra methods; performance is the same. Marking a function as virtual is where the cost is and only in calling it on the type that has it marked as virtual (base/parent type).

https://github.com/benaadams/PerfTesting/tree/master/src/DispatchTesting

docs/specs/pipelines-io.md
davidfowl: @KrzysztofCwalina poll? No, poll where? There's no point burning a thread. This pattern is more efficient. How much more depends on how many reads you do...

docs/specs/pipelines-io.md
benaadams: Sorry had tests wrong; sealed doesn't look like it does anything to perf. Same as non sealed derived class. And sealed non derived class is same as sealed non derived class. Updated numbers.

While interfaces are slower than abstract classes, if you can do an `as` cast the whole sequence is way faster than abstract classes; and `as` cast between classes makes no difference (likely because the overridden functions are same via base or derived; whether sealed or not)

benaadams: Bug for sealed not doing anything https://github.com/dotnet/coreclr/issues/1166

omariom: Currently jit doesn't take advantage of sealing (except when casting), but there jiras for that.

PR #1072, opened 2016-12-15
Improve race detection in OwnedMemory 

mjp41:
This design alters the synchronization to catch a race between
Dispose and AddReference.

By making Dispose of a OwnedMemory first invalidates the _id,
and then check for a zero reference count, and making
AddReference validate the id after increasing the count,
we guarantee that parallel calls to Dispose and Reserve at
least one will fail.

This change also detects some instances of double dispose.

### DIFF
--- a/src/System.Slices/System/Buffers/OwnedMemory.cs
+++ b/src/System.Slices/System/Buffers/OwnedMemory.cs
@@ -8,10 +8,13 @@ namespace System.Buffers
     {
         static long _nextId = InitializedId + 1;
         const long InitializedId = long.MinValue;
+        const long FreedId = long.MinValue + 1;
         int _referenceCount;
 
+        private long _id;
+
         public int Length { get; private set; }
-        protected long Id { get; private set; }
+        protected long Id { get { return _id; } }
         protected T[] Array { get; private set; }
         protected IntPtr Pointer { get; private set; }
         protected int Offset { get; private set; }
@@ -23,7 +26,7 @@ namespace System.Buffers
 
         protected OwnedMemory(T[] array, int arrayOffset, int length, IntPtr pointer = default(IntPtr))
         {
-            Id = InitializedId;
+            _id = InitializedId;
             Initialize(array, arrayOffset, length, pointer);
         }
 
@@ -69,11 +72,11 @@ namespace System.Buffers
         {
             Contract.Requires(array != null || pointer != null);
             Contract.Requires(array == null || arrayOffset + length <= array.Length);
-            if (!IsDisposed) {
+            if (!IsDisposed && Id!=InitializedId) {
                 throw new InvalidOperationException("this instance has to be disposed to initialize");
             }
 
-            Id = Interlocked.Increment(ref _nextId);
+            _id = Interlocked.Increment(ref _nextId);
             Array = array;
             Offset = arrayOffset;
             Length = length;
@@ -83,9 +86,9 @@ namespace System.Buffers
 
         public void Dispose()
         {
+            Interlocked.Exchange(ref _id,  FreedId);
             if (ReferenceCount != 0) throw new InvalidOperationException("outstanding references detected.");
             Dispose(true);
-            Id = InitializedId;
             Array = null;
             Pointer = IntPtr.Zero;
             Length = 0;
@@ -95,7 +98,7 @@ namespace System.Buffers
         protected virtual void Dispose(bool disposing)
         { }
 
-        public bool IsDisposed => Id == InitializedId;
+        public bool IsDisposed => Id == FreedId;
 
         public void AddReference()
         {
@@ -120,8 +123,13 @@ namespace System.Buffers
         #region Used by Memory<T>
         void IKnown.AddReference(long id)
         {
-            VerifyId(id);
             AddReference();
+            try {
+                VerifyId(id);
+            } catch (ObjectDisposedException e) {
+                Release();
+                throw e;
+            }
         }
 
         void IKnown.Release(long id)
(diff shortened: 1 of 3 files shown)


mjp41:
CC @KrzysztofCwalina

mjp41:
Sorry, I didn't realise this would affect the other tests.  I can fix most, but I am unsure what to do with `UnownedBuffer`.  It doesn't initialise the underlying `OwnedMemory`, as the `Id==InitializedId`, and `Initialize` is not called.

Should I add two special states to `OwnedMemory`, one for not intialized and one for freed.   

cc @davidfowl

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -120,8 +122,8 @@ protected internal virtual DisposableReservation Reserve(ref ReadOnlyMemory<T> m
         #region Used by Memory<T>
         void IKnown.AddReference(long id)
         {
-            VerifyId(id);
             AddReference();
+            VerifyId(id);
```
KrzysztofCwalina: Why is the order switched here?

tests/System.Slices.Tests/MemoryTests.cs
```diff
@@ -30,25 +31,72 @@ public void SimpleTestS()
...
-                using (memorySlice.Reserve()) { // increments the "outstanding span" refcount
-                    Assert.Throws<InvalidOperationException>(() => { // memory is reserved; cannot dispose
-                        owner.Dispose();
-                    });
-                    Span<byte> span = memorySlice.Span;
-                    span[0] = 255;
-                } // releases the refcount
-            }
+            Memory<byte> copyStoredForLater = default(Memory<byte>);
+            Assert.Throws<InvalidOperationException>(() => {
```
KrzysztofCwalina: The scope of this try-catch is quite large. This means the test will pass if the exception is thrown from let's say OwnedArray constructor. Is that the best we can do?

src/System.Slices/System/Buffers/OwnedMemory.cs
mjp41: So you have in Dispose you have 
   invalidate ID, followed by check reference count is zero
and here you have 
   ensure non-zero reference count, followed by validate ID. 

By performing the checks/validate after the action that will cause the other operation to fail you guarantee that at least one of the operations will fail if they race.

tests/System.Slices.Tests/MemoryTests.cs
mjp41: It is the implicit
```csharp
finally {
    owner.Dispose();
}
```
That is throwing the exception.  I could unpack the `using` statement?

If I disable the check for double calls to Dispose, then this would be reverted.  And other code does do double dispose on subclasses, which is why the tests are currently failing.

KrzysztofCwalina: Let's unpack the using block for now.

src/System.Slices/System/Buffers/OwnedMemory.cs
mjp41: We could catch the exception, Decrement the reference count, and rethrow the exception?  Or unpack VerifyId if you don't want the rethrow.

tests/System.Slices.Tests/MemoryTests.cs
KrzysztofCwalina: Ah, I just looked at the UnownedMemory failure. We need to decide if we want to allow double disposes on the abstraction.

src/System.Slices/System/Buffers/OwnedMemory.cs
KrzysztofCwalina: Let's rethrow, i.e. I think it's ok that we erase the callstack in this case as there is no user code on it ever.

tests/System.Slices.Tests/MemoryTests.cs
mjp41: We could make it so on the first call to Dispose, it calls Dispose(true), and after that Dispose(false) until it is reinitialized.  So it isn't an exception, but can be programmatically used.

KrzysztofCwalina: Dispose true/false already has meaning. I don't want to hijack it for something else. On the other hand, we have many types in the framework that allow multiple dispose calls. The subsequent ones are just nops. But in this case, maybe we want to detect them (without failing)

@davidfowl, @jkotas?

jkotas: The standard pattern is to allow multiple calls to Dispose. From Dispose documentation: `If an object's Dispose method is called more than once, the object must ignore all calls after the first one`

I think Dispose on OwnedMemory should follow it for consistency.

mjp41:
I have removed the detecting double Disposes.  It is unclear how far down the "abusing using" route we want to go.

I have made AddReference, release and rethrow if it fails to check the ID.

I have expanded the tests to more closely track where the exceptions occur.  More exceptions are now reported.  Previously, the API had the behavour that if it threw an exception, then it had not made any state changes, now that is not the case. @KrzysztofCwalina I could reestablish that behaviour for the sequential case, but it might have undesired consequences in a concurrent setting.

davidfowl:
@mjp41 I'll take a look and see why we're double disposing with unowned buffers (that's where the problem was right?)

mjp41:
@davidfowl, I think UnownedMemory had a pattern I hadn't expected as it doesn't call Initialize.  I added an additional state of FreedId, versus InitializeId.  To detect this kind of case.  I think most of the double disposes where from tests, where they did disposing at many points. 

You can just make the Dispose code track double deallocation.   Just raise an exception if the Exchange at the start of Dispose returns FreedId.  I can get back to you on where things are happening, but it will be this evening before I get a change to do this.

PR #1075, opened 2016-12-17
Modify PrimitiveParser template system; add templates TryParseBoolean

botaberg:
Re-implement PrimitiveParser.TryParseBoolean overloads using a template file
Move InternalParser TryParseUri implementation into PrimitiveParser, templated and with some modifications
Add performance tests for TryParseBoolean and TryParseUri (Utf8)
Add functional tests for TryParseUri (more to be done here)
Remove "unsafe" modifier from TryParseInteger implementations which use ReadOnlySpan


botaberg:
@KrzysztofCwalina @joshfree @yizhang82 Please take a look. If this is good, next steps will be filling out the "main" overloads - those outside of InvariantUtf8 and InvariantUtf16.

src/System.Text.Primitives/System/Text/Parsing/InvariantUri.tt
```diff
@@ -0,0 +1,113 @@
+﻿<#@ template hostspecific="false" language="C#" #>
```
KrzysztofCwalina: This needs some security focused review before we merge. Feel free to remove this one parser from this PR so we can merge sooner.

KrzysztofCwalina:
Great. Looks good. 
As I mentioned in the comment, I think parsing Uri should be a separate PR. It's more security sensitive and should be reviewed separately.

src/System.Text.Primitives/System/Text/Parsing/InvariantBool.cs
```diff
@@ -0,0 +1,267 @@
+﻿
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+// NOTE: This file is generated via a T4 template. Please do not edit this file directly. Any changes should be made
+// in InvariantUri.tt.
```
botaberg: Nice catch!

joshfree:
@shiftylogic could you take a quick look too?

botaberg:
I can pull out the Uri parsing into a separate PR. One thing to understand about this implementation of TryParseUri is that I tried to preserve the behavior of InternalParser.TryParseUri, as part of a broader effort to remove InternalParser. As a result, the behavior is slightly different from Uri.TryCreate and does not provide any performance improvement over Uri.TryCreate. In fact, Uri.TryCreate will necessarily be called if the Uri is valid.

An example of a significant difference: Uri.TryCreate will accept and parse '%' signs at appropriate locations within the Uri, whereas this TryParseUri implementation does not accept '%' as a valid character. To me, this means that the implementation is far complete. However, this is the behavior that was expected in InternalParser.TryParseUri, and has been preserved.

KrzysztofCwalina:
Yeah, now that I think about it, I think we need a totally different approach to parsing URI. In fact, I don't think we can use System.Uri as the representation of the result. System.Uri forces creating system.string, transcoding to UTF16, etc.

@davidfowl, what's going to be our non-allocating representation of an URI? i.e. do we have a non-allocating URI?

botaberg:
@KrzysztofCwalina I have removed the Uri parsing template and testing from this PR as per your request.

KrzysztofCwalina:
@dotnet-bot test this please

src/System.Text.Primitives/System/Text/Parsing/Internal/PrimitiveParser_bool.cs
```diff
@@ -81,6 +81,30 @@ private static bool IsTrue(Utf8String text)
             return true;
         }
 
+        private static bool IsTrue(ReadOnlySpan<byte> text)
+        {
+            if (text.Length < 4)
```
ahsonkhan: Wouldn't it be better for this to be (text.Length != 4) ?

src/System.Text.Primitives/System/Text/Parsing/Internal/PrimitiveParser_bool.cs
```diff
@@ -189,6 +213,34 @@ private static bool IsFalse(Utf8String text)
             return true;
         }
 
+        private static bool IsFalse(ReadOnlySpan<byte> text)
+        {
+            if (text.Length < 5)
```
ahsonkhan: Same here, text.Length != 5

PR #1077, opened 2016-12-21
Parsing spec

ahsonkhan:
Good to know for future: How small (whether in terms of impact or amount changed) does the change need to be for me to create a pull request? Should it be for any change, regardless?


jkotas:
> Should it be for any change, regardless?

Yes.

Also, it is a good practice to cleanup the git history when submitting the PR. Your change has 4 merge commits that would nice to rebase away.

PR #1079, opened 2016-12-21
Added Requirements section to encoding spec

KrzysztofCwalina:
cc: @ahsonkhan, @joshfree, @terrajobst, @davidfowl, @shiftylogic

### DIFF
--- a/docs/specs/encoding.md
+++ b/docs/specs/encoding.md
@@ -6,4 +6,13 @@ The existing .NET encoding APIs (System.Text.Encoding) don't work super well in 
 2. They don't work with Span\<T\>
 3. They have the overhead of virtual calls (which might be significan to transcoding small slices)
 
+The requirements for the new APIs are as follows:
+
+1. Transcode bytes contained in ReadOnlySpan<byte> into a passed in Span<byte>
+2. The API needs to handle running out of space in the output buffer and then continuing when an additional output buffer is passed in.
+3. The API needs to be fast (on par with native code implementations)
+4. Needs to do this with zero GC allocations
+5. Needs to be stateless (multithreaded)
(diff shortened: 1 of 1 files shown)


docs/specs/encoding.md
```diff
@@ -6,4 +6,13 @@ The existing .NET encoding APIs (System.Text.Encoding) don't work super well in
 2. They don't work with Span\<T\>
 3. They have the overhead of virtual calls (which might be significan to transcoding small slices)
 
+The requirements for the new APIs are as follows:
+
+1. Transcode bytes contained in ReadOnlySpan<byte> into a passed in Span<byte>
+2. The API needs to handle running out of space in the output buffer and then continuing when an additional output buffer is passed in.
+3. The API needs to be fast (on par with native code implementations)
+4. Needs to do this with zero GC allocations
+5. Needs to be stateless (multithreaded)
```
davidfowl: I'd love to understand how when we have requirement 2.

jkotas: It would tell you how much it consumed from the input buffer.

davidfowl: Which component is stateful? The one consuming this? What does that look like?

KrzysztofCwalina: The encoding APIs need to be stateless so we can call them on multiple threads without the need to allocate on every thread.

docs/specs/encoding.md
```diff
@@ -6,4 +6,13 @@ The existing .NET encoding APIs (System.Text.Encoding) don't work super well in
 2. They don't work with Span\<T\>
 3. They have the overhead of virtual calls (which might be significan to transcoding small slices)
```
joshfree: typo:  significan

KrzysztofCwalina: fixed

PR #1081, opened 2016-12-22
Changed EncodingData/TextEncoding architecture to support arbitrary encodings

KrzysztofCwalina:
Encoding of symbols in formatters and parsers has been handled by EncodingData. 
But formatting of text, e.g. StringFormatter.Append(string...), used hardcoded UTF8 or UTF16 encoding logic. This change allows text formatting to be controlled by TextEncoder abstraction, and will allows us to support additional encodings (e.g. ISO8859, ASCII, etc.)

TextEncoder needs specific APIs that can handle encoding from and into Spans. If we ever add such APIs to System.Text.Encoding (and meet some other requirements), we will be able to remove TextEncoder and use System.Text.Encoding instead. 

Also, the details of APIs in this PR are not finalized. I just wanted the basic architecture to be reasonably on track. For example, it now seems like EncodingData should be renamed. But let's talk about these details at the API review.


KrzysztofCwalina:
cc: @joshfree, @ahsonkhan, @shiftylogic, @terrajobst, @botaberg

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter_time.cs
```diff
@@ -182,19 +182,19 @@ static bool TryFormatDateTimeFormatO(DateTimeOffset value, bool isDateTime, Span
         {
```
ahsonkhan: We renamed the parameter formattingData to encodingData in the above TryFormatDateTimeFormat method, why not here?

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -2,13 +2,293 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.ComponentModel;
+using System.Text.Utf8;
 
 namespace System.Text {
 
-    public enum TextEncoding : byte
+    public abstract class TextEncoder
     {
```
ahsonkhan: Should the file name be changed to TextEncoder instead of TextEncoding?

ahsonkhan:
Otherwise, LGTM.

src/System.Text.Primitives/System/Text/TextEncoding.cs
KrzysztofCwalina: Yeah, I will rename in a subsequent PR. I diod not want to rename it in this one as there were so many changes to the file that git would most probably not detect it as a rename but rather delete + new file.

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter_time.cs
KrzysztofCwalina: The parameter names are very inconsistent. We need to settle on the name for EncodingData and then rename all these parameters. I just renamed a couple opportunistically.

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter_time.cs
```diff
@@ -129,23 +129,23 @@ public static bool TryFormat(this DateTime value, Span<byte> buffer, TextFormat
             }      
         }
 
-        static bool TryFormatDateTimeFormagG(DateTime value, Span<byte> buffer, EncodingData formattingData, out int bytesWritten)
+        static bool TryFormatDateTimeFormagG(DateTime value, Span<byte> buffer, EncodingData encodingData, out int bytesWritten)
```
KrzysztofCwalina: Good catch. I will fix.

PR #1082, opened 2016-12-22
Make Buffer/Cursor ToString output control characters

pakrym:
@davidfowl

### DIFF
--- a/src/System.IO.Pipelines/ReadableBuffer.cs
+++ b/src/System.IO.Pipelines/ReadableBuffer.cs
@@ -346,67 +346,9 @@ namespace System.IO.Pipelines
             var sb = new StringBuilder();
             foreach (var memory in this)
             {
-                int length = memory.Span.Length;
-                for (int i = 0; i < length; i++)
-                {
-                    char c = (char)memory.Span[i];
-                    AppendCharLiteral(c, sb);
-                }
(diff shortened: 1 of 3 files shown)


src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -349,13 +349,66 @@ public override string ToString()
...
-                    byte b = memory.Span[i];
-                    sb.Append((char)b);
+                    char c = (char)memory.Span[i];
+                    AppendCharLiteral(c, sb);
                 }
             }
             return sb.ToString();
         }
 
+        internal static void AppendCharLiteral(char c, StringBuilder sb)
```
davidfowl: Move this static elsewhere

src/System.IO.Pipelines/SpanExtensions.cs
```diff
@@ -0,0 +1,71 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Text;
+
+namespace System.IO.Pipelines
+{
+    public class SpanExtensions
```
davidfowl: Make this all internal.

PR #1084, opened 2016-12-22
Fix race condition in ResizingThreadTableWorks.

mjp41:
I believe this fixes an intermittent fault on ResizingThreadTableWorks.
The race could be observed by inserting a Thread.Sleep(10) as the first
command of the thread body, then the test fails consistently.

The WaitHandles are now made in sequential code, so there is no race on
adding to the threads list, with other additions, or the wait at the
end of the test.

@KrzysztofCwalina @joshfree does this fix the issue you were observing?


KrzysztofCwalina:
Thanks!

PR #1092, opened 2017-01-03
Add Cursor to ReadableBufferReader and dispatch from Pipe

pakrym:
@davidfowl


davidfowl:
@pakrym why is the CI failing?

pakrym:
@davidfowl because dispatched continuation was not awaited, removed dispatch from this PR.

src/System.IO.Pipelines/WritableBuffer.cs
```diff
@@ -109,5 +109,14 @@ public Task FlushAsync()
         {
             return _pipe.FlushAsync();
         }
+
+        /// <summary>
+        /// Signals the <see cref="IPipelineReader"/> data is available.
+        /// </summary>
+        /// <returns>A task that completes when the data is fully flushed.</returns>
+        public Task SignalAsync()
```
davidfowl: I'm not a fan of this. What does it look like to use it?

src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -223,7 +227,12 @@ public bool TrySliceTo(byte b1, out ReadableBuffer slice, out ReadCursor cursor)
         public ReadableBuffer Slice(int start, int length)
         {
             var begin = _start.Seek(start);
-            return Slice(begin, begin.Seek(length));
+            var end = begin.Seek(length);
```
davidfowl: Why not bounds check as part of seek? Why do it after seeking?

aL3891:
`public ReadCursor Move(ReadCursor cursor, int count)`, yes please :)

src/System.IO.Pipelines/WritableBuffer.cs
pakrym: https://github.com/aspnet/KestrelHttpServer/commit/b1342aa4a5ab1975590a7179ac4d96ce22af6ac3#diff-4837ca204b80ef23b080588614cea7f8L302

src/System.IO.Pipelines/ReadableBuffer.cs
pakrym: There is a check in seek to make sure we are not Seeking past the end of segments, but we need also to check that we are not past the `End`. I was thinking about passing End to Seek and handling everything at one place.

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -49,7 +49,7 @@ public static ValueTask<int> ReadAsync(this IPipelineReader input, Span<byte> de
                 var inputBuffer = result.Buffer;
 
                 var fin = result.IsCompleted;
-                var sliced = inputBuffer.Slice(0, destination.Length);
+                var sliced = inputBuffer.Slice(0, Math.Min(inputBuffer.Length, destination.Length));
```
pakrym: Yes, `Slice` lost it's "Slice at most" behavior and became "Slice exactly"

src/System.IO.Pipelines/ReadableBuffer.cs
davidfowl: Yes, pass end to Seek

src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -551,5 +577,26 @@ bool ISequence<ReadOnlyMemory<byte>>.TryGet(ref Position position, out ReadOnlyM
             }
             return true;
         }
+
+        public ReadCursor Move(ReadCursor cursor, int count)
+        {
+            var newCursor = cursor.Seek(count);
```
davidfowl: Also rename to Advance ala @aL3891 's comment

pakrym: Advance is to similar to AdvanceReader

PR #1093, opened 2017-01-03
Support integer parsing in non-invariant or custom cultures

botaberg:
Implement the main overloads of the integer parsing routines - those that take EncodingData and TextFormat as arguments - and remove references to the InternalParser implementation.

Parsing in Hexadecimal will only work if the format is InvariantUtf8, and it will support the InvariantUtf16 format with more work into the parsing trees in EncodingData. My understanding is that supporting hexadecimal parsing outside invariant cultures is lower priority.

Leading sign symbols (+, -, etc.) are supported when parsing signed integers, but unsupported when parsing unsigned integers. Thousands separators, decimal points, and other symbols are unsupported by these implementations. Support for these characters may be straightforward to implement, as soon as our policy decision are laid out.

These implementations look at TextFormat to determine if the parsing should be in hexadecimal, but otherwise ignore the information in TextFormat.

I added tests for these parsing methods, but Utf16 testing is disabled because I could not find a parsing tree for Utf16. Once we are able to generate parsing trees, that part of the testing can be enabled. Included in the test suite are Thai-encoding tests for signed integer parsing.


botaberg:
@KrzysztofCwalina @joshfree @yizhang82

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
```diff
@@ -0,0 +1,301 @@
...
+                    bytesConsumed = charsConsumed * sizeof(char);
+                    return result;
+                }
+                else
+                {
+                    // TODO: EncodingData.InvariantUtf8 and EncodingData.InvariantUtf16 don't work for parsing right now
+                    // because their parsing trees are undefined. A custom EncodingData with a manually-created parsing
+                    // tree must be used instead. Default to InvariantUtf8 parsing for now, but after autogeneration of
+                    // parsing trees is implemented, this should return false.
+                    return InvariantUtf8.Hex.TryParseByte(text, out value, out bytesConsumed);
```
KrzysztofCwalina: Is there a reason we would not simply throw NotImplementedException for now?

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
```diff
@@ -0,0 +1,301 @@
...
+
+// NOTE: This file is generated via a T4 template. Please do not edit this file directly. Any changes should be made
+// in Unsigned.tt.
+
+namespace System.Text
+{
+    public static partial class PrimitiveParser
+    {
+        public static bool TryParseByte(ReadOnlySpan<byte> text, out byte value, out int bytesConsumed, EncodingData encoding = default(EncodingData), TextFormat format = default(TextFormat))
+        {
```
KrzysztofCwalina: I think this should start with 

if(encoding.IsInvariantUtf8) { // call fast paths }
else {
   // the code that you have below, i.e. way below where you actually use the parse tire.
}

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
```diff
@@ -0,0 +1,301 @@
...
+                if (!success || nextSymbol > 9)
+                {
+                    bytesConsumed = index;
+                    value = (byte) parsedValue;
+                    return true;
+                }
+
+                // If parsedValue > (byte.MaxValue / 10), any more appended digits will cause overflow.
+                // if parsedValue == (byte.MaxValue / 10), any nextDigit greater than 5 implies overflow.
+                if (parsedValue > byte.MaxValue / 10 || (parsedValue == byte.MaxValue / 10 && nextSymbol > 5))
```
KrzysztofCwalina: Why don't you implement the same check as in the invariant parser? i.e. parse without overflow checks up to some value and then do checks.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
```diff
@@ -0,0 +1,301 @@
...
+                {
+                    bytesConsumed = 0;
```
KrzysztofCwalina: Design questions: should we return the current index when parsing fails? This would allow programs (e.g. editor) to show to the user the exact location where parsers failed.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
KrzysztofCwalina: Also, since other formats besides the default are not supported, you should throw NotSupportedException for all formats that are not yet supported.

botaberg: Got it

botaberg: Should we throw, or just return false?

Also, are we ever planning on supporting formats where the number of digits is specified? It makes sense to support these from a formatting perspective, but perhaps not from a parsing perspective since it seems like the only effect would be to invalidate some strings that don't match the format.

botaberg: One thing about this is that these code paths will go untested until we update the tests to use EncodingData.InvariantUtf8/16.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
botaberg: I think we would want to return false here, instead of defaulting to InvariantUft8 parsing. However, I wanted there to be some way for us to do hex parsing. Hex parsing is currently impossible without making an assumption about the encoding.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
botaberg: I thought this might be less useful outside an Invariant setting, since symbols can be of arbitrary length. For instance, each Utf16 digit is 2 bytes long and each Thai digit is 3 bytes long.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
KrzysztofCwalina: Can you update the tests to call these overloads with both the built-in invariant encoding and with custom invariant encoding to exercise all code paths? It should be super simple.

botaberg: Built-in invariant encoding does not work yet. No parsing trie.

        public EncodingData(byte[][] symbols, TextEncoder encoder)
        {
            _symbols = symbols;
            _encoder = encoder;
            _parsingTrie = null;
            _parsingTrie = CreateParsingTire(_symbols);
        }

        private ParsingTrieNode[] CreateParsingTire(byte[][] _digitsAndSymbols)
        {
            // TODO: this needs to be implemented;
            return null;
        }

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
KrzysztofCwalina: I am not sure I understand. If the encoding is invariant, your code should either work or throw NYI exceptions. If the encoding is not invariant, for now we should probably throw NotSupportedException, and then think if we want to support it later and how.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
KrzysztofCwalina: Ah makes sense.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
KrzysztofCwalina: I would throw. We return false if the data passed in is not parable. The format is just not supported (i.e. it will never succeed; no reason to call eh API with unsupported format).

KrzysztofCwalina: Can you just add the tire to the built-in encodings? Use the one from the tests.

botaberg: Ah, thanks.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
botaberg: I could implement something based on the number of digits parsed though. I'm working on your other requests and will push the changes soon - then perhaps I can work on this as well.

KrzysztofCwalina:
Looks good. Thanks!

PR #1095, opened 2017-01-04
Add buildtools to CoreFXLab.

crummel:
- Build stuff is mostly copied from CoreFX.
- We retain the ability to build the old way for now.
- .xprojs are replaced by .csprojs.
- Product changes were bruteforced to make it build, they definitely require review.


crummel:
I'm going to clean up the .csprojs a bit, they were copied from CoreFX as well and I don't think everything in them applies to CoreFXLab.
Source changes and other concerns:
- Adding netcoreapp1.0 as a target framework was needed to get several of the projects to compile.  I don't know why this is.  Is it a problem?
- System.IO.Pipelines.File gave me "System.Threading.Overlapped 4.3.0 is not compatible with net451" so I upgraded it to net46, which cascaded back through the System.IO.Pipelines projects.  Is this a real problem or is something wrong with how I'm building it?
- Buildtools includes analyzers that catch some DLL imports that aren't cross-platform.  I've disabled these in those source files, but we could also turn the analyzers off globally.
- Some projects needed more dependencies declared.  I'm not sure why this is.
- Tests are not yet converted.
- Would someone mind trying the debugging scenario that we were concerned this would break?
@joshfree @KrzysztofCwalina @shiftylogic @ahsonkhan

davidfowl:
@crummel are these 2017 enabled? I'm not a fan of using completely custom everything.

crummel:
@davidfowl I haven't tried with 2017, I'll spin up a machine and see if it works.  Is your concern just using the projects in VS?  That should work, although we weren't sure if we needed to extra work to support debugging.

davidfowl:
> @davidfowl I haven't tried with 2017, I'll spin up a machine and see if it works. Is your concern just using the projects in VS? That should work, although we weren't sure if we needed to extra work to support debugging.

More concerned that we're not dogfooding what people are using in real life. CoreFx and CoreCLR builds are very special compared to corefxlab. These are regular class libraries that should be using the regular .NET Core tooling that we're shipping with VS2017.

joshfree:
Thanks for bringing the latest build tools to corefxlab @crummel 

@KrzysztofCwalina any concerns before this gets merged in?

crummel:
@joshfree @KrzysztofCwalina I do get a build error in 2017, I'm looking into it.

dependencies.props
```diff
@@ -0,0 +1,138 @@
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
davidfowl: What is this?

src/System.Binary.Base64/System.Binary.Base64.csproj
```diff
@@ -0,0 +1,65 @@
+<?xml version="1.0" encoding="utf-8"?>
+<Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
davidfowl: These files should be using the new SDK syntax.

src/System.Binary.Base64/System.Binary.Base64.csproj
```diff
@@ -0,0 +1,65 @@
...
+  <Import Project="$([MSBuild]::GetDirectoryNameOfFileAbove($(MSBuildThisFileDirectory), dir.props))\dir.props" />
+  <PropertyGroup>
+    <ProjectGuid>b9ecbaf7-524d-4134-ae37-eff7802b221f</ProjectGuid>
```
davidfowl: All of this needs to be deleted. These projects should look like this https://github.com/aspnet/Security/blob/dev/src/Microsoft.AspNetCore.Authentication.Cookies/Microsoft.AspNetCore.Authentication.Cookies.csproj

init-tools.cmd
```diff
@@ -0,0 +1,83 @@
+@if not defined _echo @echo off
```
weshaggard: Why aren't you using the bootstraping scripts?

src/System.Binary.Base64/System.Binary.Base64.csproj
weshaggard: You will need to pull in a newer CLI in order to use the new SDK syntax. I think CoreFxLab would be a great candidate for dogfooding that build portion of the CLI, that is a little trickier to to in CoreFx right now.

davidfowl: DOG FOOOD!

joshfree:
> I do get a build error in 2017, I'm looking into it.

@crummel how is this going?

crummel:
@joshfree I can push out a new version that's just the SDK 15.0 changes, but I'm still getting errors with adding buildtools to that.  If we just want the project changes and come back later to actually add buildtools to the projects I can get an updated PR out today.

joshfree:
If you want to break up the work into multiple PRs that's fine - so long as VS use isn't regressed in the repo.

crummel:
@davidfowl @joshfree @weshaggard I've updated to use the latest project syntax and confirmed debugging in VS2017 works.  I'll do the bootstrap change separately.

joshfree:
> I've updated to use the latest project syntax and confirmed debugging in VS2017 works.  I'll do the bootstrap change separately.

OK sounds good.  Thanks @crummel

KrzysztofCwalina:
Sounds good. Let's merge the change once if passes the tests.

KrzysztofCwalina:
@dotnet-bot test this please

PR #1097, opened 2017-01-05
First round of updates to Channels library

stephentoub:
I'm starting to revisit this library to improve the APIs, performance, etc.  This is a first round of changes.

Primary changes:
- Added GetAwaiter to the IReadableChannel interface.  This allows channels to provide their own awaitable implementation if they can do so in an optimized way, e.g. the unbounded SPSC channel can do so allocation-free, reusing the awaiter implementation rather than creating new tasks.  Channels that don't want to do anything special can achieve the same behavior as exists today by implementing GetAwaiter like: ```public ValueAwaiter<T> GetAwaiter() => new ValueAwaiter<T>(ReadAsync());```
- Added support for the bounded channel to control what happens when a write occurs on a full channel; current options include waiting (the same as the existing behavior), or dropping an existing element from the channel, either oldest or newest.
- Removed some of the channels I'd initially added as an experiment but that aren't likely to be shipped, e.g. a channel that wraps a task.
- ClosedChannelException was made public, so that it can be caught and so that custom channels can throw it.
- Split the Channels class into three classes.  Channels now contains the methods meant to be called statically (e.g. the creation methods), and ChannelsExtensions holds the extension methods for manipulating channel instances; the cleaner separation makes for a better IntelliSense experience and is helpful in code maintenance.  The internal/private helpers that were on Channels have been put onto an internal ChannelsUtilities class.
- Other refactorings, such as moving all of the channels types from being nested in Channel to being top-level.  I'm considering exposing these as public (there are pros and cons to that), and this is a step in that direction if we choose to do so (in which case it's possible Channel and its Create methods could be removed).

cc: @anurse, @KrzysztofCwalina


src/System.Threading.Tasks.Channels/System/Runtime/CompilerServices/AutoResetAwaiter.cs
```diff
@@ -0,0 +1,155 @@
...
+using System.Diagnostics;
+using System.Runtime.ExceptionServices;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>Provides a reusable object that can be awaited by a consumer and manually completed by a producer.</summary>
+    /// <typeparam name="TResult">The type of data being passed from producer to consumer.</typeparam>
+    public sealed class AutoResetAwaiter<TResult> : IAwaiter<TResult>
```
anurse: Public! I love it :). We've been looking for things like this. /cc @davidfowl 

Is there a ManualResetAwaiter? (He says, before finishing reviewing the code ;P)

src/System.Threading.Tasks.Channels/System/Runtime/CompilerServices/ValueAwaiter.cs
```diff
@@ -0,0 +1,142 @@
...
+namespace System.Runtime.CompilerServices
+{
+    /// <summary>
+    /// Provides an awaiter that can be created around an existing value or around
+    /// another async operation, represented by either a <see cref="Task{TResult}"/>
+    /// or an <see cref="IAwaiter{T}"/>. 
+    /// </summary>
+    /// <typeparam name="TResult">Specifies the type of the result.</typeparam>
+    [StructLayout(LayoutKind.Auto)]
+    public struct ValueAwaiter<TResult> : IAwaiter<TResult>
```
anurse: More goodness! Is this going to live in Channels or is this just a place to stash it right now? Seems more generally useful.

src/System.Threading.Tasks.Channels/System/Runtime/CompilerServices/AutoResetAwaiter.cs
stephentoub: > Is there a ManualResetAwaiter?

Not yet.  I've only added things as I've needed them.  It's quite possible I'll need one of these, too, in which case I'll add it.

src/System.Threading.Tasks.Channels/System/Runtime/CompilerServices/ValueAwaiter.cs
stephentoub: Right now it's a place to stash it.  It could for example potentially move to System.Threading.Tasks.Extensions.dll (where ValueTask is).

PR #1098, opened 2017-01-05
Adding utf8 encoding performance comparison test

ahsonkhan:
Comparing the implementation in CoreCLR and the naive implementation in Corefxlab.

image

This gives us an idea of where we could be if we added APIs to Encoding and where we should at least be with corefxlab APIs.


jkotas:
> This gives us an idea of where we could be if we added APIs to Encoding

I do not understand what you meant by this.

tests/System.Text.Primitives.Encoding.Performance.Tests/EncodingPerfComparisonTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            }
+        }
+
+        public string GenerateString(int charLength)
+        {
+            Random rand = new Random();
+            var plainText = new StringBuilder();
+            for (int j = 0; j < charLength; j++)
+            {
+                plainText.Append((char)rand.Next(char.MinValue, char.MaxValue));
```
jkotas: String composed of completely random Unicode characters like this do not exist in practice. The strings you generate this way will tend to also be invalid because of they will contain invalid surrogate combinations - I assume that's why you had to do the roundtrip using Encoding.Unicode to get rid of the invalid surrogates.

Instead, you should test:

- ASCII strings (ie 1 byte UTF8 encoding)
- Cyrilic strings (ie 2 byte UTF8 encoding) 
- Chinese string (ie 3 byte UTF8 encoding)
- Combinations of the above

ahsonkhan:
>  This gives us an idea of where we could be if we added APIs to Encoding

> I do not understand what you meant by this.

The way the tests are written, they try to encode system.string from a ReadOnlySpan and output Span<byte> encoded in utf8. They show perf if we were to add APIs to System.Text.Encodings that behave this way.

tests/System.Text.Primitives.Encoding.Performance.Tests/EncodingPerfComparisonTests.cs
jkotas: I think I did not get the culture names for the different sizes right ... but you got the idea.

ahsonkhan: Do you mean something similar to the following?
https://github.com/dotnet/corefxlab/blob/master/tests/System.Text.Primitives.Tests/Encoding/PerformanceTests.cs#L74

I realized the issue with unpaired surrogates. I wanted to have a string with variable byte characters within it rather than separate strings of fixed byte length characters.

jkotas:
> They show perf if we were to add APIs to System.Text.Encodings that behave this way.

I agree that it would be nice to add APIs that take Span/ReadOnlySpan to Encoding eventually; but it has nothing to do with the perf characteristics. The APIs that take Span/ReadOnlySpan with these performance characteristics can live anywhere.

tests/System.Text.Primitives.Encoding.Performance.Tests/EncodingPerfComparisonTests.cs
jkotas: Yes, something like this.

scripts/PerfHarness/PerfHarness.cs
```diff
@@ -30,6 +30,7 @@ private static string[] GetTestAssemblies()
             "Benchmarks",
             "System.Binary.Base64.Tests",
             "System.Text.Primitives.Tests",
+            "System.Text.Primitives.Encoding.Performance.Tests"
```
KrzysztofCwalina: Why a new project? I think you should just add it to System.Text.Primitives.Tests

tests/System.Text.Primitives.Encoding.Performance.Tests/EncodingPerfComparisonTests.cs
```diff
@@ -0,0 +1,68 @@
...
+            string unicodeString = GenerateString(1000);
+            ReadOnlySpan<char> characters = unicodeString.Slice();
+
+            Text.Encoding utf8 = Text.Encoding.UTF8;
+
+            int utf8Length = utf8.GetByteCount(unicodeString);
+            byte[] utf8Buffer = new byte[utf8Length];
+            Span<byte> span;
+
+            char[] charArray = characters.ToArray();
```
jkotas: > Could you also add a test where you convert span to array inside the test loop?

If you were to use the current encoding APIs, it should use the unsafe overloads that take pointers to avoid the unnecessary copy and array allocation.

scripts/PerfHarness/PerfHarness.cs
ahsonkhan: Sure.

tests/System.Text.Primitives.Encoding.Performance.Tests/EncodingPerfComparisonTests.cs
KrzysztofCwalina: ... assuming we don't want to use unsafe code overloads.

ahsonkhan: > Could you also add a test where you convert span to array inside the test loop? I would simulate what we would get if we used current encoding APIs for buffer formatting.

I do not understand. Can you please clarify?

tests/System.Text.Primitives.Encoding.Performance.Tests/EncodingPerfComparisonTests.cs
joshfree: http://stackoverflow.com/questions/6136800/unicode-test-strings-for-unit-tests

tests/System.Text.Primitives.Encoding.Performance.Tests/EncodingPerfComparisonTests.cs
jkotas: > assuming we don't want to use unsafe code overloads.

I do not think such number is interesting.

tests/System.Text.Primitives.Tests/EncodingPerfComparisonTests.cs
```diff
@@ -0,0 +1,85 @@
...
+            byte[] utf8Buffer = new byte[utf8Length];
+            Span<byte> span;
+
+            char[] charArray = characters.ToArray();
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                    if (convertToArrayInLoop)
+                        utf8.GetBytes(characters.ToArray(), 0, characters.Length, utf8Buffer, 0);
```
jkotas: As I have said, I do not think that it is interesting to measure the performance if you convert to array in every loop iteration. It is not how production quality code would have been ever written.

ahsonkhan: > the performance if you convert to array in every loop iteration.

It seems, there is contention here due to difference of opinion. I am fine either way (keeping it or removing it) when an agreement is reached.

One thing I would like clarification, you stated that repeated conversion within a loop wouldn't happen. However, isn't this equivalent to a single conversion in user code? The loop is there to take multiple samples, no?

ahsonkhan:
Can we merge this? Why is the pre-merge signing tool still pending for 24 hours?

Issue #1091, opened 2017-01-03
Productize Encoding APIs


KrzysztofCwalina:
Pri 0
- [x] Make encoding between UTF8 and UTF16 (both ways) fast

Pri 1
- [x] Hide System.Text.Utf8 and System.Text.Utf16 namespaces, except for Utf8String
- [x] Add TextEncoder overloads hinted in comments in source.
- [x] Move UnicodeCodePoint to some subnamespace, or hide for now, if not needed by bullet above
- [ ] Support other encodings, e.g. 8859-*
- [x] We need ability to continue (as opposed to restart) encoding after output buffer space has been exhausted

Pri 2
- [x] Rename EncodingData to something better
- [x] Rename TextEncoder.Id type, and Scheme property

KrzysztofCwalina:
@joshfree, @terrajobst, @shiftylogic, @ahsonkhan, @botaberg, I added this issue to track work related to formatting APIs. Feel free to add/tweak workitems.

joshfree:
>> System.Text.Utf8 and System.Text.Utf16 namespaces need to be designed properly

Is this just a catch all item or can you quickly enumerate some additional issues here?

KrzysztofCwalina:
Let's just change the workitem to "hide the APIs except for Utf8String". I will make the update.

KrzysztofCwalina:
@joshfree, @ahsonkhan, I updated the workitem list. I tried to make it clearer, added priorities, and sorted in the order I think the work should be done.

KrzysztofCwalina:
All of the items except for 8859-* support are done. I created #1727 to track the remaining item and so closing this issue.

PR #1103, opened 2017-01-10
Add TextEncoder overloads hinted in comments in source.

ahsonkhan:
As part of Issue #1091 - Productize Encoding APIs


src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8EncodedCodePoint.cs
```diff
@@ -33,6 +33,27 @@ public Utf8EncodedCodePoint(char character) : this()
...
+            {
+                throw new ArgumentOutOfRangeException("character", "Surrogate characters are not allowed");
+            }
+
+            unsafe
+            {
+                fixed (byte* encodedData = &_byte0)
+                {
+                    var buffer = new Span<byte>(encodedData, 4);
+                    if (!Utf8Encoder.TryEncodeCodePoint(codePoint, buffer, out _length))
```
KrzysztofCwalina: How fast do we need this ctor to be? If we need it to be fast, I think you should implement custom routine to convert UCP to each of the fields. I worry that creating span and then calling the TryEncodeCodePoint method will be too slow. But maybe you should measure.

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -121,6 +123,156 @@ public override bool TryEncodeFromUtf16(ReadOnlySpan<char> utf16, Span<byte> buf
...
+                {
+                    bytesWritten = 0;
+                    return false;
+                }
+                buffer[bytesWritten++] = (byte)codepoint;
+            }
+            else
+            {
+                Utf8EncodedCodePoint encoded;
+                encoded = new Utf8EncodedCodePoint(codePoint);
```
KrzysztofCwalina: Yeah, this seems wasteful. You encode the code point to byte fileds inside Utf8EncodedCodePoint (via Span), then you take the values out of the fileds one-by-one and write them to the span. You should rewrite the routine to directly write to span

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -25,11 +25,13 @@ public enum Id  : byte
...
-        // TryEncodeFromUnicode(UnicodeCodePoint codePoint, Span<byte> buffer, out int bytesWritten);
-        // TryDecodeToUnicode(Span<byte> encoded, ReadOnlySpan<UnicodeCodePoint> decoded, out int bytesWritten);
-        // TryDecodeToUnicode(Span<byte> encoded, UnicodeCodePoint decoded, out int bytesWritten);
+        public abstract bool TryEncodeFromUnicode(ReadOnlySpan<UnicodeCodePoint> codePoints, Span<byte> buffer, out int bytesWritten);
+
+        public abstract bool TryEncodeFromUnicode(UnicodeCodePoint codePoint, Span<byte> buffer, out int bytesWritten);
+
+        public abstract bool TryDecodeToUnicode(Span<byte> encoded, Span<UnicodeCodePoint> decoded, out int bytesWritten);
+
+        public abstract bool TryDecodeToUnicode(Span<byte> encoded, out UnicodeCodePoint decoded, out int bytesWritten);
```
KrzysztofCwalina: I know you did not do it in the first place, but could you please split the abstract encoder from the concrete ones (source file wise)?

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -121,6 +123,156 @@ public override bool TryEncodeFromUtf16(ReadOnlySpan<char> utf16, Span<byte> buf
...
 
+        public override bool TryEncodeFromUnicode(ReadOnlySpan<UnicodeCodePoint> codePoints, Span<byte> buffer, out int bytesWritten)
+        {
+            var avaliableBytes = buffer.Length;
+            int bytesWrittenForCodePoint = 0;
+            bytesWritten = 0;
+
+            for (int i = 0; i < codePoints.Length; i++)
+            {
+                if (!TryEncodeFromUnicode(codePoints[i], buffer.Slice(bytesWritten), out bytesWrittenForCodePoint))
```
KrzysztofCwalina: I don't think we can slice for every single code point. It's too expensive. I am speaking about the routines that we need to get very fast (e.g. encoding between UTF8 and UTF16 and vice versa). Some other routines (for simplicity sake) can of course slice. I am not sure if this one is one of these or not.

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -121,6 +123,156 @@ public override bool TryEncodeFromUtf16(ReadOnlySpan<char> utf16, Span<byte> buf
...
+                        }
+                    }
+                }
+
+                bytesWritten += encoded.Length;
+            }
+            return true;
+        }
+
+        public override bool TryDecodeToUnicode(Span<byte> encoded, Span<UnicodeCodePoint> decoded, out int bytesWritten)
```
KrzysztofCwalina: This is a design question, maybe we can discuss it at some point: I think for long buffer transformation routines, we might want a different pattern than TryXxx. I think it might be worth exploring something that would transform up to some point and then returned how many units of work it completed till it ran out of buffer space. The caller would recall the routine with a new buffer.

@shiftylogic, @terrajobst

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -121,6 +123,156 @@ public override bool TryEncodeFromUtf16(ReadOnlySpan<char> utf16, Span<byte> buf
...
+            {
+                input = (byte)(input << 1);
+                count++;
+            }
+            return count;
+        }
+
         public override bool TryEncodeString(string value, Span<byte> buffer, out int bytesWritten)
         {
             var avaliableBytes = buffer.Length;
```
KrzysztofCwalina: This routine should now just get a span out of the string and call the TryEncode(Span, Span) method.

After you do this and improve perf of the main TryEncode routine, it would be good to include perf results in the PR description.

src/System.Text.Primitives/System/Text/TextEncoding.cs
jkotas: The existing Encoder/Decoder.Convert is pretty close to what you need (without the `completed` out argument because of your encoders are state free):

https://github.com/dotnet/corefx/blob/master/src/System.Runtime/ref/System.Runtime.cs#L7046

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8EncodedCodePoint.cs
jkotas: Is this class going away as the TODO at the top says? It does not really matter what it does if we are going to delete it anyway...

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -25,11 +25,13 @@ public enum Id  : byte
...
         public abstract bool TryEncodeFromUtf16(ReadOnlySpan<char> utf16, Span<byte> buffer, out int bytesWritten);
 
-        // TODO: I think we also need to add the following to support transcoding
-        // TryEncodeFromUnicode(ReadOnlySpan<UnicodeCodePoint> codePoints, Span<byte> buffer, out int bytesWritten);
-        // TryEncodeFromUnicode(UnicodeCodePoint codePoint, Span<byte> buffer, out int bytesWritten);
-        // TryDecodeToUnicode(Span<byte> encoded, ReadOnlySpan<UnicodeCodePoint> decoded, out int bytesWritten);
-        // TryDecodeToUnicode(Span<byte> encoded, UnicodeCodePoint decoded, out int bytesWritten);
+        public abstract bool TryEncodeFromUnicode(ReadOnlySpan<UnicodeCodePoint> codePoints, Span<byte> buffer, out int bytesWritten);
+
+        public abstract bool TryEncodeFromUnicode(UnicodeCodePoint codePoint, Span<byte> buffer, out int bytesWritten);
```
jkotas: Design question: Is this abstract method worth it? It super rare that you need to encode one character; and if you hit the rare case that needs it, it can be done as non-virtual wrapper over the method that takes spans.

src/System.Text.Primitives/System/Text/TextEncoding.cs
jkotas: These encoding/decoding routines should have same or better performance than the (unsafe) encoding/decoding routines built into framework. You may want to start by copying that implementation over and adjusting it to match - what you got now is way behind the framework implementation performance-wise.

jkotas: > This routine should now just get a span out of the string and call the TryEncode(Span, Span) method

Also, this method should be implemented in the base class - it should not be virtual.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8EncodedCodePoint.cs
ahsonkhan: I did not write this code. It is also, already in master. I am not sure why it showed up in the diff for this PR.

https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8EncodedCodePoint.cs

src/System.Text.Primitives/System/Text/TextEncoding.cs
ahsonkhan: Which namespace and directory should I put the UT16EncodingLE and Utf8TextEncoding implementation classes in?

src/System.Text.Primitives/System/Text/TextEncoding.cs
ahsonkhan: If I do not slice, I have to essentially copy the
```C#
TryEncodeFromUnicode(UnicodeCodePoint codePoint, Span<byte> buffer, out int bytesWritten) 
```
implementation into the for loop for 

```C#
TryEncodeFromUnicode(ReadOnlySpan<UnicodeCodePoint> codePoints,...
```

If not, how else do we avoid slicing without code duplication?

src/System.Text.Primitives/System/Text/TextEncoding.cs
ahsonkhan: Should I use the implementation found here directly:
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L205

src/System.Text.Primitives/System/Text/TextEncoding.cs
KrzysztofCwalina: I would use System.Text.Utf16 and System.Text.Utf8 respectively.

src/System.Text.Primitives/System/Text/TextEncoding.cs
KrzysztofCwalina: Yeah.

src/System.Text.Primitives/System/Text/TextEncoding.cs
ahsonkhan: Don't we already have that here:
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/TextEncoding.cs#L45

So, should we just remove the override methods?

src/System.Text.Primitives/System/Text/TextEncoding.cs
shiftylogic: You can also pass a target index for writing to the output buffer. It means a bit of extra math, but avoids the Slice operation.

src/System.Text.Primitives/System/Text/TextEncoding.cs
ahsonkhan: I will address this closer to the end of the PR discussions.

src/System.Text.Primitives/System/Text/TextEncoding.cs
ahsonkhan: This issue has been addressed by removing the individual code point encoder/decoders.

src/System.Text.Primitives/System/Text/TextEncoding.cs
KrzysztofCwalina: For now, I would just remove the override. Only add it back if you think we need and can implement better perf.

KrzysztofCwalina:
Please rebase/squash the commits

PR #1105, opened 2017-01-11
Adding TextEncoder overloads and splitting the classes into separate …

ahsonkhan:
…source files.

 - Cleaning up and addressing PR comments from the closed PR #1103 

As part of Issue #1091 - Productize Encoding APIs


ahsonkhan:
I closed the previous PR because I pulled in commits from master, accidentally.

The changes here:
- split the Utf8 and Utf16 encoding classes into separate source files in the System.Text.Utf8 and System.Text.Utf16 namespaces
- removed the override methods for encoding string for both Utf8 and Utf16.
```C#
public override bool TryEncodeString(string value, Span<byte> buffer, out int bytesWritten)
```

ahsonkhan:
cc @KrzysztofCwalina can this be merged?

KrzysztofCwalina:
I am not sure which of these changes are new code and which are just the split. I would like to know that so I can review new code.

ahsonkhan:
> I am not sure which of these changes are new code and which are just the split. I would like to know that so I can review new code.

The only difference is the removal of the TryEncodeString override methods, and the namespace change:

```C#
using System.Text.Utf8;
namespace System.Text.Utf16
{
...
namespace System.Text.Utf8 
{
```

The rest has remained unchanged from the previous PR.

KrzysztofCwalina:
OK. In this case looks good.

PR #1108, opened 2017-01-11
More changes to channels

stephentoub:
- Main change: replace interface with abstract base classes.  This enables us to provide default implementations for methods (e.g. GetAwaiter) that can then be overridden as needed for either behavior or performance by derived classes.  It also allows us to add additional such virtuals in the future as we find need.  And to do all this while having the Channel.Create methods return the base type that provides all of the functionality and without requiring casting or type checks.
- Boosts code coverage to ~99% line coverage and ~98% branch coverage.  Remaining uncovered blocks and branches are all subject to race conditions as to whether they get hit or not.
- Removed an exception getting thrown and caught in AsyncEnumerator, resulting in spurious first-chance exceptions in the debugger.
- Added a GetAwaiter to writable channels, behaving the same as WaitToWriteAsync.
- Fix several functional bugs, e.g. in handling of observables

cc: @KrzysztofCwalina, @anurse


src/System.Threading.Tasks.Channels/System/Threading/Tasks/Channels/Channel_2.cs
```diff
@@ -0,0 +1,120 @@
...
+        //   type could override them all to avoid the extra virtual hop, but then that's a bunch of extra
+        //   code to override all of these members multiple times.
+        // - Either every additional member on one of the halfs needs to be duplicated here, or we have a
+        //   potential inconsistency with some operations showing up on the halfs but not on the whole. 
+
+        ///// <summary>
+        ///// Gets a <see cref="Task"/> that completes when no more data will ever
+        ///// be available to be read from this channel.
+        ///// </summary>
+        //public Task Completion => In.Completion;
```
KrzysztofCwalina: Why are all these commented out?

KrzysztofCwalina:
looks good!

src/System.Threading.Tasks.Channels/System/Threading/Tasks/Channels/Channel_2.cs
stephentoub: See the comment right above. I don't know if we want them or not.

davidfowl:
Can you also update the readme with the new usage patterns?

stephentoub:
> Can you also update the readme with the new usage patterns?

Thanks for the reminder. Done.

Issue #1106, opened 2017-01-11
Parsing routines fail when called with the default format


KrzysztofCwalina:
Repro:

        ulong value;
        int consumed;
        if(PrimitiveParser.TryParseUInt64(new byte[] { (byte)'1' }, out value, out consumed)
        {
            Console.WriteLine(value);
        }

Result:
System.NotImplementedException was unhandled
  HResult=-2147467263
  Message=Format with precision not supported.
  Source=System.Text.Primitives
 
Expected:
prints '1'

botaberg:
Taking a look now.

botaberg:
This is due to the observation that default(TextFormat) has precision 0 rather than NoPrecision:

if (format.HasPrecision)
{
       throw new NotImplementedException("Format with precision not supported.");
}

Should the default be precision 0 rather than NoPrecision? What is the difference between precision==0 and precision==NoPrecision?

KrzysztofCwalina:
NoPrecision means that precision was not specified. Precision==0 means it was specified as 0. The default being 0 will screw up some formatting routines. I think there are two solutions:
1. When entering methods, convert the default format to NoPrecision.
2. Treat 0 value of the filed, as "no precision specified" and when the precision is explicitly set to zero, set the filed to some value that means 'zero", i.e. use one bit (high bit) in the precision filed to mean "the value was set"

I think the first option is better. The second option means that all code that wants to red the precision would have to clear the bit.

KrzysztofCwalina:
@botaberg fixed it.

PR #1109, opened 2017-01-12
In parsing methods, check for default(EncodingData) and default(TextFormat)

botaberg:
PrimitiveParser's main overloads - those which take EncodingData and TextFormat objects as parameters - are currently failing when given the default values for those objects. This change introduces better behavior. If default(EncodingData) is passed, the parser will default to InvariantUtf8 parsing. If default(TextFormat) is passed, the parser will not interpret the format as having precision, as was the case previously.

This is a fix for Issue #1106.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Parsing/Signed.cs
+++ b/src/System.Text.Primitives/System/Text/Parsing/Signed.cs
@@ -11,7 +11,12 @@ namespace System.Text
     {
         public static bool TryParseSByte(ReadOnlySpan<byte> text, out sbyte value, out int bytesConsumed, EncodingData encoding = default(EncodingData), TextFormat format = default(TextFormat))
         {
-            if (format.HasPrecision)
+            if (encoding == default(EncodingData))
+            {
+                encoding = EncodingData.InvariantUtf8;
+            }
+
+            if (!format.IsDefault && format.HasPrecision)
             {
                 throw new NotImplementedException("Format with precision not supported.");
             }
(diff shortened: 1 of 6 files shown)


tests/System.Text.Primitives.Tests/PrimitiveParserIntegerTests.cs
```diff
@@ -161,6 +161,11 @@ public unsafe void ParseByteDec(string text, bool expectSuccess, byte expectedVa
             Assert.Equal(expectedValue, parsedValue);
             Assert.Equal(expectedConsumed, consumed);
 
+			result = PrimitiveParser.TryParseByte(utf8Span, out parsedValue, out consumed);
```
KrzysztofCwalina: Why the indent here is one less than above?

botaberg: Darn! It's because I haven't changed my VS settings to turn spaces into tabs when editing .tt files.

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
```diff
@@ -11,7 +11,12 @@ public static partial class PrimitiveParser
     {
         public static bool TryParseSByte(ReadOnlySpan<byte> text, out sbyte value, out int bytesConsumed, EncodingData encoding = default(EncodingData), TextFormat format = default(TextFormat))
         {
-            if (format.HasPrecision)
+            if (encoding == default(EncodingData))
+            {
+                encoding = EncodingData.InvariantUtf8;
+            }
+
+            if (!format.IsDefault && format.HasPrecision)
```
KrzysztofCwalina: For the bug fix, I think it's ok, but I think it would be good to open an issue and design the precondition contract for these routines and implement them at the top of the routine. Otherwise: a) it's not clear what combination of arguments is valid, b) we are checking some things (e.g. IsDefault) more than one time.

KrzysztofCwalina:
Looks good.

Issue #1070, opened 2016-12-15
Productize Parsing APIs


KrzysztofCwalina:
Parsing APIs design document is at https://github.com/dotnet/corefxlab/blob/master/docs/specs/parsing.md

We identified the following work items required to productize the APIs:

Priority 0
- [x] remove old APIs (.Internal sub-namespace)
- [x] remove `EncodingData(byte[][] symbols, TextEncoding encoding, Tuple<byte, int>[] parsingTrie)`
- [ ] remove all custom parsers from System.IO.Pipelines (and other corefxlab components)
- [x] make sure APIs can support non-invariant cultures (by implementing and testing prototype)

Priority 1
- [x] implement parsing APIs for integer types
- [x] support UTF8 and UTF16 invariant
- [x] performance tests and improvement for integer parsing
- [x] support fast lower case and upper case hexadecimal parsing
- [x] generate parsers using T4 templates
- [x] support Boolean
- [x] full test coverage (including .NET compatibility tests)
- [ ] ASCII and ISO-8859-1 (required by networking protocols)

Priority 2
- [ ] full support for non-invariant cultures
- [ ] support BigInteger parsing
- [ ] support URI parsing
- [x] support DateTime, DateTimeOffset 
- [x] support GUID
- [x] support 'G', 'D', 'R', 'X' formats
- [x] clean up code, including performance tests (we have too many of them)
- [x] add GC allocation tests (the low level APIs should not allocate)
- [-] add performance tests comparing to C++ performance
- [ ] add support for parsing discontinuous buffers (higher level APIs)
- [x] do formal API review

Priority 3
- [x] support Single, Double
- [x] support Decimal
- [x] support TimeSpan
- [x] support all numeric and date time format strings.
- [ ] support UTF16BE

KrzysztofCwalina:
I moved the workitems from the spec to this issue. It's easier to update.

@joshfree, @botaberg, @terrajobst, @davidfowl

botaberg:
Thanks, @KrzysztofCwalina !

KrzysztofCwalina:
@botaberg, I updated the list. Mainly rearranged priorities

GSPP:
Are `byte*` overloads even necessary? Can they not be fully replaced with the span version?

ahsonkhan:
> Are `byte*` overloads even necessary? Can they not be fully replaced with the span version?

You are right, they are not necessary if we replace them with APIs that take span.

[...]

PR #1111, opened 2017-01-12
Remove implementation of and references to InternalParser

botaberg:
System.Text.Internal.InternalParser contains many prototype implementations of parsing methods. Most of these implementations have been replaced by the more standardized APIs in System.Text.PrimitiveParser, although many concepts from these prototypes have been brought over. This change removes the implementation and all references to InternalParser, as part of the effort to productize the parsing APIs ( #1070).


botaberg:
@KrzysztofCwalina @yizhang82 Please take a look.

KrzysztofCwalina:
thanks!

PR #1116, opened 2017-01-12
OwnedMemory tidy and performance testing

mjp41:
Main changes
* Made Reference Count value not directly accessible, only whether it is zero or not.  
* Added a setting for whether reservations should use:
   * Interlocked Reference Counting (slow),
   * Reference Counter (almost safe), or
   * Nothing (unsafe).
* Added a simple performance test for reservations that does parameter sweep. 

Minor tidying to 
* Remove unused Reserve method.
* Make DisposableReservations be able to return the Span, as in the design doc.

### DIFF
--- a/src/System.IO.Pipelines/BufferSegment.cs
+++ b/src/System.IO.Pipelines/BufferSegment.cs
@@ -85,11 +85,11 @@ namespace System.IO.Pipelines
 
         public void Dispose()
         {
-            Debug.Assert(_buffer.ReferenceCount >= 1);
+            Debug.Assert(_buffer.HasOutstandingReferences);
 
             _buffer.Release();
 
-            if (_buffer.ReferenceCount == 0)
+            if (!_buffer.HasOutstandingReferences)
             {
                 _buffer.Dispose();
             }
--- a/src/System.Slices/System/Buffers/Memory.cs
+++ b/src/System.Slices/System/Buffers/Memory.cs
@@ -55,7 +55,7 @@ namespace System
 
         public Span<T> Span => _owner.GetSpanInternal(_id).Slice(_index, _length);
 
-        public DisposableReservation Reserve() => new DisposableReservation(_owner, _id);
+        public DisposableReservation<T> Reserve() => new DisposableReservation<T>(_owner, _id);
 
         public unsafe bool TryGetPointer(out void* pointer)
         {
--- a/src/System.Slices/System/Buffers/OwnedMemory.cs
+++ b/src/System.Slices/System/Buffers/OwnedMemory.cs
@@ -114,10 +114,6 @@ namespace System.Buffers
         protected virtual void OnZeroReferences()
         { }
 
-        protected internal virtual DisposableReservation Reserve(ref ReadOnlyMemory<T> memory)
-        {
-            return new DisposableReservation(this, Id);
-        }
 
         #endregion
 
--- a/src/System.Slices/System/Buffers/OwnedMemory.cs
+++ b/src/System.Slices/System/Buffers/OwnedMemory.cs
@@ -195,21 +195,23 @@ namespace System.Buffers
         unsafe bool TryGetPointer(long id, out void* pointer);
     }
 
-    public struct DisposableReservation : IDisposable
+    public struct DisposableReservation<T> : IDisposable
     {
-        IKnown _owner;
+        OwnedMemory<T> _owner;
         long _id;
 
-        internal DisposableReservation(IKnown owner, long id)
+        internal DisposableReservation(OwnedMemory<T> owner, long id)
         {
             _id = id;
             _owner = owner;
-            _owner.AddReference(_id);
+            ((IKnown)_owner).AddReference(_id);
         }
+
+        public Span<T> Span => _owner.Span;
 
         public void Dispose()
         {
-            _owner.Release(_id);
+            ((IKnown)_owner).Release(_id);
             _owner = null;
         }
     }
--- a/src/System.Slices/System/Buffers/OwnedMemory.cs
+++ b/src/System.Slices/System/Buffers/OwnedMemory.cs
@@ -18,7 +18,7 @@ namespace System.Buffers
         protected T[] Array { get; private set; }
         protected IntPtr Pointer { get; private set; }
         protected int Offset { get; private set; }
-        public int ReferenceCount { get { return _referenceCount; } }
+        public bool HasOutstandingReferences { get { return _referenceCount != 0; } }
 
         private OwnedMemory() { }
 
@@ -87,7 +87,7 @@ namespace System.Buffers
         public void Dispose()
         {
             Interlocked.Exchange(ref _id,  FreedId);
-            if (ReferenceCount != 0) throw new InvalidOperationException("outstanding references detected.");
+            if (HasOutstandingReferences) throw new InvalidOperationException("outstanding references detected.");
             Dispose(true);
             Array = null;
             Pointer = IntPtr.Zero;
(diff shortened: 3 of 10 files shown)


mjp41:
@KrzysztofCwalina I made the System.Slices.Tests part of the performance harness.  Should I create a separate project for System.Slices performance?

@davidfowl I removed direct access to an OwnedMemory's reference count, and instead provide a method for checking if it has no references. This seemed to fit with your use case in the Pipelines.

src/System.IO.Pipelines/BufferSegment.cs
```diff
@@ -85,11 +85,11 @@ public BufferSegment(OwnedMemory<byte> buffer, int start, int end)
 
         public void Dispose()
         {
-            Debug.Assert(_buffer.ReferenceCount >= 1);
+            Debug.Assert(!_buffer.HasZeroReferences);
```
KrzysztofCwalina: Could we make it to be HasReferences? HasZeroReferences when prefixed with '!' is hard to parse (double negation).

src/System.Slices/System/Buffers/Memory.cs
```diff
@@ -55,7 +55,7 @@ public Memory<T> Slice(int index, int length)
 
         public Span<T> Span => _owner.GetSpanInternal(_id).Slice(_index, _length);
 
-        public DisposableReservation Reserve() => new DisposableReservation(_owner, _id);
+        public DisposableReservation<T> Reserve() => new DisposableReservation<T>(_owner, _id);
```
KrzysztofCwalina: I used to have this type as a generic type and it was a problem. Somethimes you want to store many reservations on a collection, and generic reservations mean you need many different types of collections.

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -4,6 +4,12 @@
 
 namespace System.Buffers
 {
+    public enum Mode {RefCount, LocalRefCount, None};
+
+    public class OwnedMemorySettings {
+        public static Mode s_mode = Mode.RefCount;
```
KrzysztofCwalina: If you make DisposableReservation non-generic, you could just add this to DisposableReservation, so there is no need for a new type. I know this type is probably just temporary (for experiments), but we try to minimize number of types in our APIs.

Also, s_mode does not follow our naming guidelines. Name it 'Mode'

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -4,6 +4,12 @@
 
 namespace System.Buffers
 {
+    public enum Mode {RefCount, LocalRefCount, None};
```
KrzysztofCwalina: I would call these

public enum ReferenceCountingMethod {
    Interlocked,
    ReferenceCounter,
    None
}

tests/System.Slices.Tests/MemoryTests.cs
```diff
@@ -108,41 +110,111 @@ public void RacyAccess()
...
+
+        public static IEnumerable<object[]> ReservationPerformanceData =>
+            MakePermutations(MakeArray(1,4,10),
+                             MakeArray(100,1000),
+                             MakeArray(1,2,4,8,12),
+                             MakeArray(Mode.None, Mode.RefCount, Mode.LocalRefCount));
+
+
+        [MemberData(nameof(ReservationPerformanceData))]
+        [Benchmark]
```
KrzysztofCwalina: Re your question about the project for the tests:

I think it's ok for these tests to be here or in the Benchmarks project. I would not create a separate test project for these.

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -198,21 +207,41 @@ internal interface IMemory<T> : IKnown
...
 
-        internal DisposableReservation(IKnown owner, long id)
+        internal DisposableReservation(OwnedMemory<T> owner, long id)
         {
             _id = id;
             _owner = owner;
-            _owner.AddReference(_id);
+            switch(OwnedMemorySettings.s_mode) {
+                case Mode.RefCount:
+                    ((IKnown)_owner).AddReference(_id);
```
KrzysztofCwalina: If you keep the reservation generic, you don't need the cast here.

src/System.Slices/System/Buffers/Memory.cs
mjp41: So if DisposableReservation is non-generic then it can't give you access to the Span directly.  Could you elaborate a little on the use case, as storing reservations in a collection to me sounds dangerous, particularly with the ReferenceCounter.

KrzysztofCwalina: The idea was that the pipeline would store tokens for all buffer segments,a nd then dispose them all when it's done with the buffer. But I just realized in such scenario, all tokens will be Disposable<byte>.

Let's just make it generic then and see how it feels to use it.

src/System.IO.Pipelines/BufferSegment.cs
benaadams: May be confusion with `ContainsReferences` and ilk? Perhaps `HasOutstandingReferences`?

benaadams: Or better keep it in sync with usage `HasReservations` or change `DisposableReservation`  to `DisposableReference`

mjp41: I think `DisposableReference` and  ` HasOutstandingReferences` makes the most sense.  As `AddReference` exists on the API surface as well. 

Would `AutoReference` be better than `DisposableReference`?

KrzysztofCwalina: I named it DisposableReference to highlight to users that they need to dispose it.

benaadams: `AutoReference` sounds like it deals with disposal for you; would need something like "Destructible Types" https://github.com/dotnet/roslyn/issues/161 for an auto

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -198,21 +207,41 @@ internal interface IMemory<T> : IKnown
...
+                    break;
+                case Mode.LocalRefCount:
+                    ReferenceCounter.AddReference(_owner);
+                    break;
+                case Mode.None:
+                    break;
+            }
         }
 
+        public Span<T> Span => _owner.Span;
```
KrzysztofCwalina: Sorry, I just noticed this: Could you at least for now not remove this but instead rename it to DangerousGetSpan().
Even if we settle on refcounting option being the default, we want to have the ability to get span without refcounting in cases where the caller knows that it is safe, e.g. the caller owns the memory lifetime or it is GCed memory.

mjp41: I can do this, but for clarity which paths are considered Dangerous?  For instance, I think Memory<T>.Span is going to be Dangerous, should that also be renamed?
To me only 

```
using(var res = Memory<T>.Reserve()) {
     var s = res.Span;
     ...
}
```

Is the only way to get a Span from a Memory that doesn't have `Dangerous` in the external API?

KrzysztofCwalina: Yes, we wanted to do the same rename on Memory

KrzysztofCwalina:
@dotnet-bot test this please

KrzysztofCwalina:
BTW, let's merge it and then do the renames later.

mjp41:
Sounds good. Thanks for the comments @krzysztofCwalina and @benaadams.

PR #1125, opened 2017-01-14
Add parsing trie generation algorithm

botaberg:
This is a work item in #1070: Productize Parsing APIs.

The EncodingData type allows our parsing APIs to parse arbitrarily-encoded text. The user supplies specifies the encoding's symbols in the form of a list of byte sequences (byte[][]). Before this change, the user was also forced to supply a parsing trie (prefix tree) for the parsing algorithms to search through. The trie is structured as an array and is optimized for search speed, so it is difficult to generate manually. This change introduces an algorithm that generates a parsing trie structure based on the list of byte sequences given in the constructor.


botaberg:
@KrzysztofCwalina @yizhang82 PTAL

src/System.Text.Primitives/System/Text/EncodingData.cs
```diff
@@ -365,8 +346,196 @@ private bool VerifySuffix(ReadOnlySpan<byte> buffer, int codeUnitIndex, Symbol s
...
         struct ParsingTrieNode
         {
-            public byte valueOrNumChildren;
+            public byte ValueOrNumChildren;
             public int IndexOrSymbol;
         }
+
+        private class AbstractParsingTrie
+        {
+            private interface Node
```
jkotas: It would be nice to add a comment that this interface is added just for code clarity, but otherwise useless.

src/System.Text.Primitives/System/Text/EncodingData.cs
```diff
@@ -365,8 +346,196 @@ private bool VerifySuffix(ReadOnlySpan<byte> buffer, int codeUnitIndex, Symbol s
...
+            {
+            }
+
+            private struct InternalNode : Node
```
jkotas: Since you are boxing this struct anyway to cast it to the interface, it would be better to either make it a class that inherits from List<Node>; or to have just one struct type with SymbolByteOrIndex field.

KrzysztofCwalina:
Hmm, this is very expensive (allocation-wise) way to create the tire. I know most code will do it once, but it's overly allocating for now good reason. Can you not avoid boxing the nodes? i.e. create List<Node>, use the list to store new nodes, but use list indexes to "link" them.

botaberg:
I was thinking of rewriting some of this to use fewer resources, and it seems based on the feedback that it would be a good idea. I'll send an update soon with the progress.

src/System.Text.Primitives/System/Text/EncodingData.cs
```diff
@@ -365,8 +345,153 @@ private bool VerifySuffix(ReadOnlySpan<byte> buffer, int codeUnitIndex, Symbol s
         // CC = the max value, and DD = the max value's index in the same coord-system as BB.
         struct ParsingTrieNode
         {
-            public byte valueOrNumChildren;
+            public byte ValueOrNumChildren;
             public int IndexOrSymbol;
         }
+        
+        private static int CompareSuffix(Suffix x, Suffix y)
```
KrzysztofCwalina: Should this be CompareTo method in Suffix struct?

src/System.Text.Primitives/System/Text/EncodingData.cs
```diff
@@ -365,8 +345,153 @@ private bool VerifySuffix(ReadOnlySpan<byte> buffer, int codeUnitIndex, Symbol s
...
+
+        private struct SuffixClump
+        {
+            public byte BeginningByte;
+            public List<Suffix> Suffixes;
+
+            public SuffixClump(byte beginningByte)
+            {
+                BeginningByte = beginningByte;
+                Suffixes = new List<Suffix>();
```
KrzysztofCwalina: Would it be possible to estimate the capacity we need and pass it to the ctor?

src/System.Text.Primitives/System/Text/EncodingData.cs
```diff
@@ -365,8 +345,153 @@ private bool VerifySuffix(ReadOnlySpan<byte> buffer, int codeUnitIndex, Symbol s
         // CC = the max value, and DD = the max value's index in the same coord-system as BB.
         struct ParsingTrieNode
         {
-            public byte valueOrNumChildren;
+            public byte ValueOrNumChildren;
             public int IndexOrSymbol;
         }
+
```
KrzysztofCwalina: I would prefer to move the tire generation algorithm to a separate file. It's quite involved.

KrzysztofCwalina: Also, could you describe the algorithm and the role of each of the types in the comments.

src/System.Text.Primitives/System/Text/EncodingData.cs
botaberg: The capacity required would be between 1 and the number of symbols, which is currently 17. A capacity of 20 or so would guarantee that no resizing is needed. I can make the ctor set this, like: Suffixes = new List<Suffix>(20);

src/System.Text.Primitives/System/Text/EncodingData.cs
botaberg: For the separate file, would you also want a separate type? Or is it fine to keep this in EncodingData?

src/System.Text.Primitives/System/Text/EncodingData.cs
KrzysztofCwalina: Sounds good.

src/System.Text.Primitives/System/Text/EncodingData.cs
KrzysztofCwalina: Your call if you'd like a separate type or not.

KrzysztofCwalina:
Looks good. Thanks!

PR #1131, opened 2017-01-18
Various fixes in pipelines

pakrym:
1. Introduce separate conditional variables to enable\disable stack trace tracking
2. Fix `Cursor` of `ReadableBufferReader`
3. Add diagnostics that prevents `Read` on channel that had `CompleteReading` called on it
3. Add diagnostics that prevents `Alloc` on channel that had `CompleteWriting` called on it
4. Add diagnostics that prevents `CompleteReading` on channel that has reading operation in progress (is in between `ReadAsync` and `AdvanceReader`)
5. Add diagnostics that prevents `CompleteWriting` on channel that has writing operation in progress (is in between `Alloc` and `Commit`)
6. Reorder `OwnerEmptyMemory` static initializes so it doesn't get null array.
7. Fix `OwnedMemory.Initialize` to correctly check for pointer being null


src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -33,7 +31,7 @@ public struct ReadableBuffer : ISequence<ReadOnlyMemory<byte>>
         /// <summary>
         /// Determines if the <see cref="ReadableBuffer"/> is empty.
         /// </summary>
-        public bool IsEmpty => Length == 0;
+        public bool IsEmpty => _first.IsEmpty && Length == 0;
```
halter73: 👍

halter73: I'm wondering if it's better to optimize even in the case when _first is empty. I'm a little worried that there might be a pathological (but still reasonably plausible) scenario where the consumer will read till the end of a segment of Memory and then check IsEmpty repeatedly on new Readable buffers. This could be particularly bad if the consumer falls far enough behind the producer.

MemoryPoolIterator.IsEnd short-circuits aggressively.

pakrym: I think we just need to prevent this scenario from happening or verify that it does not happen (you should never be left with an empty block at the beginning of a pipe)

Drawaes: I have seen it in stress testing my TLS library. Not sure if it was a bug that has gone but I tried to encrypt zero bytes during a foreach over a readable at the start but there was data after that. I had to put an if in that was ugly. Not sure if it is still the case.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -561,10 +616,16 @@ private void Dispose()
             Debug.Assert(Writing.IsCompleted, "Not completed writing");
             Debug.Assert(Reading.IsCompleted, "Not completed reading");
 
-            // TODO: Review throw if not completed?
+            GC.SuppressFinalize(this);
```
davidfowl: ?

src/System.IO.Pipelines/project.json
```diff
@@ -14,7 +14,13 @@
   },
   "buildOptions": {
     "allowUnsafe": true,
-    "keyFile": "../../tools/Key.snk"
+    "keyFile": "../../tools/Key.snk",
+    "define": [
```
davidfowl: :\

src/System.IO.Pipelines/Pipe.cs
pakrym: Accidental will fix.

src/System.IO.Pipelines/project.json
pakrym: Very useful to be able to disable  them, makes tests run 1000x faster. Maybe we need to move these settings to settings class but this suffices for now.

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -201,6 +197,11 @@ private void OnRead(UvStreamHandle handle, int status)
                     task.ContinueWith((t, state) => ((UvTcpConnection)state).StartReading(), this);
                 }
             }
+
+            if (normalDone || _input.Writing.IsCompleted)
```
davidfowl: Any comment to go here?

src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -86,20 +86,15 @@ internal void Initialize()
...
             return builder.ToString();
         }
 
         protected override void Dispose(bool disposing)
         {
-            Pool.Return(this);
-
             base.Dispose(disposing);
+
+            Pool.Return(this);
```
davidfowl: ?

pakrym: We need to dispose first then return because we can get into situation where we dispose newly leased block.

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
pakrym: We didn't complete on `normalDone` before.

src/System.IO.Pipelines/MemoryPoolBlock.cs
davidfowl: Add a comment

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
davidfowl: ok

davidfowl:
Ci failing in debug.

pakrym:
@davidfowl not anymore

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
```diff
@@ -171,13 +175,16 @@ protected virtual void Dispose(bool disposing)
         {
             if (disposing)
             {
-                _output.CompleteWriter();
-                _output.CompleteReader();
+                _cancelationTokenSource.Cancel();
+                _output.CancelPendingRead();
 
-                _input.CompleteWriter();
+                Task.WhenAll(_sendTask, _receiveTask);
```
davidfowl: WaitAll?

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
```diff
@@ -39,6 +40,9 @@ public class SocketConnection : IPipelineConnection
         private PipelineFactory _factory;
         private Pipe _input, _output;
         private Socket _socket;
+        private Task _receiveTask;
+        private Task _sendTask;
+        private CancellationTokenSource _cancelationTokenSource = new CancellationTokenSource();
```
davidfowl: why not just a volatile bool?

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
benaadams: `var yolo = Task.WhenAll(_sendTask, _receiveTask).ContinueWith((t) => { /*... Everything else... */})`?

pakrym:
@davidfowl used `bool` and `WaitAll`

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
```diff
@@ -455,6 +461,7 @@ private void RecycleSmallBuffer(ref ArraySegment<byte> buffer)
         private async Task<ArraySegment<byte>> ReceiveInitialDataUnknownStrategyAsync(SocketAsyncEventArgs args)
         {
 
+            Console.WriteLine("ReceiveInitialDataUnknownStrategyAsync");
```
davidfowl: Remove

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
```diff
@@ -39,6 +40,9 @@ public class SocketConnection : IPipelineConnection
         private PipelineFactory _factory;
         private Pipe _input, _output;
         private Socket _socket;
+        private Task _receiveTask;
+        private Task _sendTask;
+        private bool _stopping;
```
davidfowl: volatile?

benaadams:
>  used `bool` and `WaitAll`

Is there definitely going to be no delay on the awaits you are waiting for? If you shutdown 200 connections simultaneously there is no way the `WaitAll` can starve the threadpool?

pakrym:
@benaadams there should not be, at this point it just waits for methods to break loops and return. Maybe it's not the best pattern going forward, but at least it avoids race we had here before.

benaadams:
@pakrym I'd have gone with something like
```csharp
Task.WhenAll(_sendTask, _receiveTask)
    .ContinueWith((task, state) => 
        { 
            // if (task.IsFaulted) log exeception... ?
            var connection = (SocketConnection)state;
            
            connection._output.CompleteWriter();
            connection._input.CompleteReader();

            GC.SuppressFinalize(connection)
            connection._socket?.Dispose();
            connection._socket = null;
            if (connection._ownsFactory) { connection._factory?.Dispose(); }
            connection._factory = null;
        }, 
        this);
```
So its an non-blocking callback rather than a synchronous thread block; but I haven't traced it through, so might not matter...

src/System.IO.Pipelines/ReadCursor.cs
```diff
@@ -103,59 +101,60 @@ internal int GetLength(ReadCursor end)
             }
         }
 
-        internal ReadCursor Seek(int bytes, ReadCursor end)
+        internal ReadCursor Seek(int bytes, ReadCursor end, bool trustingEnd = false)
```
davidfowl: rename this arg to something more descriptive

PR #1132, opened 2017-01-18
Redesign of text encoding public surface area

shiftylogic:
* Text encoding public API surface proposal.
* Some optimizations to encoding (still a lot of work to do here).

@KrzysztofCwalina @ahsonkhan


KrzysztofCwalina:
@shiftylogic, could you copy and paste resulting API shape from the API review tool? Thanks.

shiftylogic:
This is the current public surface for encoding.

```c#
public abstract class TextEncoder {
        public static readonly TextEncoder Utf16;
        public static readonly TextEncoder Utf8;
        protected TextEncoder();
        public TextEncoder.Id Scheme { get; }
        public abstract bool TryDecode(ReadOnlySpan<byte> data, Span<byte> utf8, out int bytesConsumed, out int bytesWritten);
        public abstract bool TryDecode(ReadOnlySpan<byte> data, Span<char> utf16, out int bytesConsumed, out int charactersWritten);
        public abstract bool TryDecode(ReadOnlySpan<byte> data, Span<uint> utf32, out int bytesConsumed, out int codePointsWritten);
        public abstract bool TryDecode(ReadOnlySpan<byte> data, out string text, out int bytesConsumed);
        public bool TryEncode(char value, Span<byte> data, out int bytesWritten);
        public abstract bool TryEncode(ReadOnlySpan<byte> utf8, Span<byte> data, out int bytesConsumed, out int bytesWritten);
        public abstract bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> data, out int charactersConsumed, out int bytesWritten);
        public abstract bool TryEncode(ReadOnlySpan<uint> utf32, Span<byte> data, out int codePointsConsumed, out int bytesWritten);
        public abstract bool TryEncode(string text, Span<byte> data, out int bytesWritten);
        public enum Id : byte {
            Ascii = (byte)2,
            Utf16 = (byte)0,
            Utf8 = (byte)1,
        }
    }
```

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
```diff
@@ -2,102 +2,204 @@
...
         {
-            if (buffer.Length < 2)
-            {
-                codePoint = default(UnicodeCodePoint);
-                encodedBytes = default(int);
-                // buffer too small
+            charactersConsumed = 0;
+            charactersWritten = 0;
+
+            // TODO: Is this really what we want?
```
KrzysztofCwalina: I think we should return true and zeros for consumed and written.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
```diff
@@ -2,102 +2,204 @@
...
             }
+
             return true;
         }
 
+        #endregion Encoding implementation
+
+        #region Surrogate helpers
+
+        private static bool IsSurrogate(uint codePoint)
```
KrzysztofCwalina: This is not blocking this PR, but should these helpers be public somewhere?

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
 {
     class Utf16TextEncodingLE : TextEncoder
     {
-        public override bool TryEncodeFromUtf8(ReadOnlySpan<byte> utf8, Span<byte> buffer, out int bytesWritten)
+        #region Encoding Constants
+
+        // TODO: Some of these members are needed only in Utf16LittleEndianEncoder.
+        //       Should we add the usage of them to UnicodeCodePoint class and merge this class with it?
+        private const uint Utf16HighSurrogateFirstCodePoint = 0xD800;
```
KrzysztofCwalina: Are we repeating these? We have the same constants above. BTW, maybe such constants should also be public somewhere.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
...
+        private const uint Utf16HighSurrogateLastCodePoint = 0xDFFF;
+        private const uint Utf16LowSurrogateFirstCodePoint = 0xDC00;
+        private const uint Utf16LowSurrogateLastCodePoint = 0xDFFF;
+
+        private const uint Utf16SurrogateRangeStart = Utf16HighSurrogateFirstCodePoint;
+        private const uint Utf16SurrogateRangeEnd = Utf16LowSurrogateLastCodePoint;
+
+        // To get this to compile with dotnet cli, we need to temporarily un-binary the magic values
+        private const byte b0000_0111U = 7;
```
KrzysztofCwalina: I think C#7 supports binary literals you could use after we merge the tools: https://github.com/dotnet/roslyn/issues/215

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
...
-            foreach (var cp in new Utf8String(utf8).CodePoints)
-            {
-                if (!Utf16LittleEndianEncoder.TryEncodeCodePoint(cp, buffer.Slice(bytesWritten), out justWritten))
-                {
-                    bytesWritten = 0;
-                    return false;
-                }
-                bytesWritten += justWritten;
-            }
+            var utf16 = data.Cast<byte, char>();
```
KrzysztofCwalina: This cast is not portable. In fact this method will be removed. You need to use AsBytes when casting to Span<byte> and NonPortableCast when casting to anything else.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
...
 
-        public override bool TryEncodeFromUnicode(ReadOnlySpan<UnicodeCodePoint> codePoints, Span<byte> buffer, out int bytesWritten)
+        public override bool TryDecode(ReadOnlySpan<byte> data, Span<char> utf16, out int bytesConsumed, out int charactersWritten)
         {
-            var availableBytes = buffer.Length;
-            var inputLength = codePoints.Length;
-            int bytesWrittenForCodePoint = 0;
-            bytesWritten = 0;
+            // TODO: Other methods validate that the input stream contains valid sequences as they are consumed.
+            //       Currently, this is a copy operation with no validation. What is the right thing here?
```
KrzysztofCwalina: I think no validation is fine for now. If we discover that we need validation, we could add an option to do validation.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
...
+        public override bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> data, out int charactersConsumed, out int bytesWritten)
+        {
+            // TODO: Other methods validate that the input stream contains valid sequences as they are consumed.
+            //       Currently, this is a copy operation with no validation. What is the right thing here?
 
-                decoded[i] = decodedCodePoint;
-                bytesWritten += bytesWrittenForCodePoint;
+            var charBuffer = data.Cast<byte, char>();
+
+            if (utf16.Length > charBuffer.Length)
```
KrzysztofCwalina: I thought we would encode as much as fits in the output.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
...
+            charactersConsumed = utf16.Length;
             return true;
         }
 
-        public override bool TryEncodeChar(char value, Span<byte> buffer, out int bytesWritten)
+        public override bool TryEncode(ReadOnlySpan<uint> utf32, Span<byte> data, out int charactersConsumed, out int bytesWritten)
+        {
+            int written;
+            var utf16 = data.Cast<byte, char>();
+            var result = Utf16LittleEndianEncoder.TryEncode(utf32, utf16, out charactersConsumed, out written);
```
KrzysztofCwalina: BTW, I think we should consider creating PrimitiveEncoder class, similar to PrimitiveParser and Formatter, and add nested Utf8 and Utf16 classes with these static methods. This comment is not blocking this PR.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
...
+            if (bytesWritten > data.Length)
             {
                 bytesWritten = 0;
                 return false;
             }
-            buffer[0] = (byte)value;
-            buffer[1] = (byte)(value >> 8);
-            bytesWritten = 2;
+
+            unsafe
```
KrzysztofCwalina: I think we have a Slice extension method for string that would let you do this without unsafe code.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
shiftylogic: I originally did that, but it broke some test cases. I made the assumption that it was desired.

I agree with you. I'll change it to succeed when input is empty and fix the tests.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
shiftylogic: Do we see people needing this? It was something specific to UTF-16 encoding.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
shiftylogic: There was a comment left in UnicodeConstants that it might be useful to merge with the UTF-16 encoder.

I thought about this for a bit and came to the conclusion that only the UTF-16 encoder needs these values. The other copy currently exists in an old / existing class that will eventually be removed (and is currently hidden from the surface).

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8TextEncoding.cs
```diff
@@ -1,308 +1,96 @@
...
-                    continue;
-                }
-                else if (char.IsSurrogate(codePoint))
-                {
-                    if (tempCounter + 4 > availableBytes)
-                    {
-                        bytesWritten = 0;
-                        return false;
-                    }
+            var buffer = new char[len];
```
KrzysztofCwalina: Since you know the string lengh, I would do the following:
```c#
var str = new string(' ', len);
fixed(char* p = string){
    var charSpan = new Span<char>(p, len);
}
```

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
shiftylogic: FYI, a goal was to get this into the code base so we could continue iterating on performance without me having to continually merge crazy changes back into mine.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
shiftylogic: It is so much easier to work with appropriate sized spans. Do you have a recommendation on what we should do here? For internal implementation, I would really like to cast up to the appropriate size. Also, I suspect optimizations will require the ability to cast as well for reading larger units.

Thoughts?

tests/System.Text.Primitives.Tests/Encoding/Utf8EncoderTests.cs
```diff
@@ -4,79 +4,6 @@ namespace System.Text.Utf8.Tests
 {
     public class Utf8EncoderTests
```
KrzysztofCwalina: Could you please add some tests testing the scenario of partial encoding/decoding, i.e. give the method too small output buffer, and when it return false, slice the input buffer, give it a new output buffer, and then call the method again to keep encoding/decoding.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
jkotas: I expect that you will end up rewriting these in unsafe code with raw pointers to get the performance on par with what's in corelib today...

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
shiftylogic: That would mean that we can't optimize here with a simple buffer copy. If we are doing UTF-16 to UTF-16 (i.e. nothing really), then assuming that the input is correct means a buffer copy. Otherwise, we have to do a full decode / encode to validate the data before doing the copy.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
KrzysztofCwalina: We should think about it. We don't want to add APIs that will not be used, but on the other hand there is risk of people reimplementing these all the time.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
shiftylogic: @jkotas I agree with you. Most of the implementation in this change is not far from what was already there. The goal was to just change the surface. We will be iterating on performance over the next bit.

KrzysztofCwalina: For now I would just call NonProtableCast.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8TextEncoding.cs
shiftylogic: Thanks

tests/System.Text.Primitives.Tests/Encoding/Utf8EncoderTests.cs
shiftylogic: Will do.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
shiftylogic: Thanks.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
```diff
@@ -2,102 +2,204 @@
...
+            // TODO: Is this really what we want?
+            if (utf16.Length == 0)
                 return false;
-            }
 
-            uint codePointValue = buffer.Read<ushort>();
-            encodedBytes = 2;
-            if (UnicodeCodePoint.IsSurrogate((UnicodeCodePoint)codePointValue))
+            for (int i = 0; i < utf16.Length; i++)
             {
```
ahsonkhan: I think we should store utf16.Length (and utf32.Length) in a local variable.

shiftylogic: Shouldn't this be a standard compiler optimization? Old JavaScript interpreters used to be dumb and not do this, but most of them do now. I'm be shocked if the C# compiler didn't already do this.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
```diff
@@ -2,102 +2,204 @@
...
+        /// </summary>
+        /// <param name="utf32">A span containing a sequence of UTF-32 characters.</param>
+        /// <param name="utf16">A span to write the UTF-16 encoded data into.</param>
+        /// <param name="charactersConsumed">On exit, contains the number of code points that were consumed from the UTF-32 character span.</param>
+        /// <param name="charactersWritten">An output parameter to store the number of bytes written to <paramref name="utf8"/></param>
+        /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
+        public static bool TryEncode(ReadOnlySpan<uint> utf32, Span<char> utf16, out int charactersConsumed, out int charactersWritten)
         {
-            if (!UnicodeCodePoint.IsSupportedCodePoint(codePoint))
+            charactersConsumed = 0;
```
ahsonkhan: I noticed a performance increase if you write to a local variable and copy over to the out parameter right before returning rather than constantly writing to the out parameter directly.

shiftylogic: That's interesting. Do you mind if we hold that change for the next optimization pass? Much of this code is likely to be churned as we try to make it faster.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
```diff
@@ -2,102 +2,204 @@
...
+        /// <summary>
+        /// Encodes a single UTF-32 character into UTF-16.
+        /// </summary>
+        /// <param name="codePoint">A UTF-32 character to encode.</param>
+        /// <param name="utf16">A span to write the UTF-16 encoded data into.</param>
+        /// <param name="index">An index into the <paramref name="utf16"/> span to to the write operation.</param>
+        /// <param name="charactersWritten">An output parameter to store the number of bytes written to <paramref name="utf16"/></param>
+        /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
+        public static bool TryEncode(uint codePoint, Span<char> utf16, int index, out int charactersWritten)
+        {
```
ahsonkhan: Should this be public? I thought we didn't need single unit encodings (char, uint, etc).

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
```diff
@@ -2,102 +2,204 @@
...
                 }
             }
+
             return true;
         }
 
+        #endregion Encoding implementation
+
+        #region Surrogate helpers
+
```
ahsonkhan: These helpers and constants are used for encoding from utf 16 to utf8. So, utf8 needs them too.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
ahsonkhan: These helpers and constants are used for encoding from utf 16 to utf8. So, utf8 needs them too.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
shiftylogic: It is on an internal class, so not really public. It is also being used outside the class by another internal class. I suspect it might go away with optimizations, but we'll have to see.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
```diff
@@ -7,113 +7,145 @@ namespace System.Text.Utf16
 {
     class Utf16TextEncodingLE : TextEncoder
     {
-        public override bool TryEncodeFromUtf8(ReadOnlySpan<byte> utf8, Span<byte> buffer, out int bytesWritten)
+        #region Encoding Constants
+
+        // TODO: Some of these members are needed only in Utf16LittleEndianEncoder.
+        //       Should we add the usage of them to UnicodeCodePoint class and merge this class with it?
+        private const uint Utf16HighSurrogateFirstCodePoint = 0xD800;
+        private const uint Utf16HighSurrogateLastCodePoint = 0xDFFF;
```
ahsonkhan: This should be 0xDBFF, if we are repeating them.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
shiftylogic: I believe that is no longer true in this code. Anything UTF-16 related in routed through UTF-16 encoder and vice versa for UTF-8 code. That way the appropriate logic stays contextually together.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncodingLE.cs
shiftylogic: Those constants are now removed. I meant to have them on just the static class, but forgot to remove them when I split the classes.

src/System.Text.Primitives/System/Text/UnicodeHelpers.cs
```diff
@@ -0,0 +1,42 @@
...
+
+        public static bool IsSupportedCodePoint(uint codePoint)
+        {
+            return codePoint < FirstNotSupportedCodePoint;
+        }
+
+        public static bool IsBmp(uint codePoint)
+        {
+            return codePoint < 0x10000;
+        }
```
ahsonkhan: Can we have 0x10000 as a constant?

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
ahsonkhan: Of course. That should be fine.

src/System.Text.Primitives/System/Text/TextEncoding.cs
```diff
@@ -21,28 +21,172 @@ public enum Id  : byte
         public readonly static TextEncoder Utf8 = new Utf8TextEncoding() { Scheme = Id.Utf8 };
         public readonly static TextEncoder Utf16 = new Utf16TextEncodingLE() { Scheme = Id.Utf16 };
```
ahsonkhan: Do we still need Id and Scheme?

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
```diff
@@ -2,102 +2,204 @@
...
+            return codePoint >= Utf16HighSurrogateFirstCodePoint && codePoint <= Utf16HighSurrogateLastCodePoint;
+        }
+
+        #endregion Surrogate helpers
+
+        #region Old stuff
+
         // TODO: Should we rewrite this to not use char.ConvertToUtf32 or is it fast enough?
         public static bool TryDecodeCodePointFromString(string s, int index, out UnicodeCodePoint codePoint, out int encodedChars)
         {
```
ahsonkhan: Are we getting rid of UnicodeCodePoint completely? If so, should this be out uint?

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
jkotas: > think we should store utf16.Length (and utf32.Length) in a local variable

For arrays, RyuJIT generates better code when the Length property is used directly. Check https://github.com/dotnet/corefx/pull/15192 for example.

Once the RyuJIT has the bounds check elimination for Span, I expect that it is going to be the case for Span as well. So storing it a local variable will actually turn into de-optimization... .

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
shiftylogic: I'd like to. It just isn't fully done yet.

Note that the function you mentioned is existing code and inside an internal class. It is currently used within Utf8String which also needs to be looked at. Once Utf8String has been designed / worked on, this function should be stripped.

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
ahsonkhan: That is good to know. Do you when we can expect to see this improvement?

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
ahsonkhan: Perfect :)

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
jkotas: Before `Span<T>` ships. Not sure about the exact timeline.

shiftylogic:
@KrzysztofCwalina @ahsonkhan 
Requested changes should be done.

ahsonkhan:
LGTM

KrzysztofCwalina:
LGTM. Thanks!

ahsonkhan:
As part of Issue #1091 - Rename TextEncoder.Id type, and Scheme property

PR #1133, opened 2017-01-18
Added Multi-Segment HTTP Parser

KrzysztofCwalina:
The existing HTTP parser (HttpRequest.Parse) cannot parse payloads arriving in multiple buffers.
This change adds the ability to parse requests arriving in multiple-buffers. It uses BytesReader added yesterday to parse the HTTP payload.

The old single segment parser was moved to System.Text.Http.SingleSegment. I will remove it completely later.


KrzysztofCwalina:
@shiftylogic, @botaberg, @ahsonkhan

PR #1134, opened 2017-01-19
EphemeralBuffers for Windows and Unix, where the memory won't be page…

Drawaes:
Why ? 

Well storing perm secrets should be done on things like TPM's, however for items like session keys it is normally okay to store them unencrypted in memory. However perfect forward secrecy for instance can be badly compromised if the memory is paged out to disk, as it can hang around for a long time and is easier to attack than crossing into a processes memory space.

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/EphemeralBufferWindows.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/EphemeralBufferWindows.cs
@@ -13,7 +13,7 @@ namespace System.Buffers.Experimental.System.Buffers
         private IntPtr _memory;
         private int _bufferCount;
         private int _bufferSize;
-        private ConcurrentQueue<SecureMemory> _buffers = new ConcurrentQueue<SecureMemory>();
+        private ConcurrentQueue<EphemeralMemory> _buffers = new ConcurrentQueue<EphemeralMemory>();
         private UIntPtr _totalAllocated;
         private byte[] _emptyData;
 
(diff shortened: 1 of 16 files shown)


Drawaes:
The buffers are zero'ed on return, and on dispose of the entire pool. VirtualLock on windows is used and MLock on unix. For the Interop I used (although needed to add to ) files from the PAL layers in Corefx, so it should be  compliant generally there.

Drawaes:
@davidfowl these are what we talked about for in memory secrets

KrzysztofCwalina:
System.Buffers.Experimental is a pure portable library. I would not want interop code in it. I will let @davidfowl review the rest of the PR, but if he does want to add it to corfxlab, it should be in a different package, not in System.Buffers.Experimental.

src/System.Buffers.Experimental/System/Buffers/EphemeralBufferWindows.cs
```diff
@@ -0,0 +1,92 @@
...
+            if (!buffer2.Rented)
+            {
+                return;
+            }
+            buffer2.Rented = false;
+            _emptyData.CopyTo(buffer.Span);
+            _buffers.Enqueue(buffer2);
+        }
+
+        sealed class SecureMemory : OwnedMemory<byte>
```
benaadams: `EphemeralMemory` 😛

Drawaes: Whoops, missed that in the rename will change.

Drawaes:
Sure it can be moved, just seemed like it fit will all the other buffers, where would it go... ?

KrzysztofCwalina:
I would not mind a separate package/dll, e.g. System.Buffers.Ephemeral. Having said that, don't do too much work till @davidfowl confirms that he wants this in corfxlab (unless you are sure).

Drawaes:
I am not sure :) I am using it in my library currently and will continue to do so until/if it becomes part of the lab.

davidfowl:
Seems pretty specialized to go into System.Buffers... It should stay in  the SSL lib for now.

Drawaes:
Fair enough

PR #1139, opened 2017-01-20
Unsafe Utf8Encoder.TryEncode from Utf16 - performance increase

ahsonkhan:
As part of Issue #1091 - Make encoding between UTF8 and UTF16 (both ways) fast

Using DangerousGetPinnableReference() to get a pointer to the input and output spans to see if we can 
get improvements in performance.

**==============Here are the latest results:==============**
image

In all cases, there is a 2-4x performance gain over using Span IndexOf.
Comparing to System.Text.Encoding, the ASCII range encoding is still much slower, but all other ranges are on-par.

**Note:** I haven't run the corefxlab tests in VS 2017 yet, but I confirmed the correctness of the encoding in a stand-alone console app. I couldn't open the solution and wasn't sure how to run them using CLI.

***==============Previous entries (no longer relevant, please ignore)==============***
~~In all cases, there is a ~~2-4x~~ 15% performance gain over using Span IndexOf.~~

~~Comparing to System.Text.Encoding, the ASCII range encoding is still slower, but all other ranges are on-par.~~
~~Comparing to System.Text.Encoding, encoding is still 2-4x slower.~~

~~![image]~~IGNORE(https://cloud.githubusercontent.com/assets/6527137/22141068/fb1b08f2-dea4-11e6-8851-61fb88f3709f.png)

~~**Edit:** There is still some ways to go given the new results once I pinned the pointers.~~
~~![image]~~IGNORE(https://cloud.githubusercontent.com/assets/6527137/22168923/2bf033d2-df25-11e6-9868-5b77860e4d15.png)

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
@@ -21,16 +21,6 @@ namespace System.Text.Utf8
         private const byte b1110_0000U = 0xE0; //224
         private const byte b1111_0000U = 0xF0; //240
         private const byte b1111_1000U = 0xF8; //248
-        
-        private const ushort Utf16HighSurrogateFirstCodePoint = 0xD800;
-        private const ushort Utf16HighSurrogateLastCodePoint = 0xDBFF;
-        private const ushort Utf16LowSurrogateFirstCodePoint = 0xDC00;
-        private const ushort Utf16LowSurrogateLastCodePoint = 0xDFFF;
-
-        private const byte Utf8OneByteLastCodePoint = 0x7F;
-        private const ushort Utf8TwoBytesLastCodePoint = 0x7FF;
-        private const ushort Utf8ThreeBytesLastCodePoint = 0xFFFF;
-        private const uint Utf8FourBytesLastCodePoint = 0x10FFFF;
 
         #endregion Constants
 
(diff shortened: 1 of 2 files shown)


ahsonkhan:

ahsonkhan:
error CS7085: By-reference return type 'ref T' is not supported.

When can we start writing code using ref in corefxlab?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -290,53 +300,119 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
-            bytesWritten = 0;
+            int bytesWrittenTemp = 0;
+            int i = 0;
+            char codePoint = '\0';
 
-            for (int i = 0; i < availableCharacters; i++)
+            unsafe
             {
-                var ch = utf16[i];
+                var utf16Ref = Unsafe.AsPointer(ref utf16.DangerousGetPinnableReference());
```
jkotas: The use of `Unsafe.AsPointer` is not ok - you have to use `fixed` block to get unsafe pointer from Span.

KrzysztofCwalina:
The numbers look good! We might want to tweak the ASCII range code in subsequent PRs.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -290,53 +300,119 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                var utf8Ref = Unsafe.AsPointer(ref utf8.DangerousGetPinnableReference());
 
-                if ((ushort)ch <= 0x7f) // this if block just optimizes for ascii
+                var utf16CharRef = (char*)utf16Ref;
+                var utf8ByteRef = (byte*)utf8Ref;
+
+                i = 0;
```
KrzysztofCwalina: Why do you set it to zero again?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -290,53 +300,119 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+
+                ascii:
+                for (; i < inputCharacters; i++)
```
KrzysztofCwalina: I wonder if the different in ASCII performance is not because of loop unrolling (done in CLR code). As @jkotas mentioned before, have you looked into the existing CLR implementation to see how this is done there?

KrzysztofCwalina:
Also, please don't delete the Span based implementation. Please move it to a benchmark test. We want to be running this benchmark periodically to drive span performance down to be in the ballpark of the pointer one.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: Yes.

Also, keep in mind that a lot of real world workloads are in the form `<a lot of ascii> <small non-ascii string> <a lot of ascii> <small non-ascii string> <a lot of ascii> ...`. You may want to add a specific tests for that. The CoreCLR implementation does not take the slow path when it hits the first non-ascii strings, It recovers to use the unrolled implementation.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -290,53 +300,119 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                        *(utf8ByteRef + bytesWrittenTemp) = (byte)(((surrogateCalculation >> 6) & b0011_1111U) | b1000_0000U);
+                        bytesWrittenTemp++;
+                        *(utf8ByteRef + bytesWrittenTemp) = (byte)(((surrogateCalculation) & b0011_1111U) | b1000_0000U);
+                        bytesWrittenTemp++;
+                    }
+                    else if (codePoint <= Utf8ThreeBytesLastCodePoint)
                     {
-                        codePoint = (uint)ch;
-                        consumed = 1;
+                        if (bytesWrittenTemp + 3 > outputBytesAvailable)
```
jkotas: These checks as written have security hole - buffer overrun triggered by integer overflow

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -290,53 +300,119 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                    {
+                        if (bytesWrittenTemp + 4 > outputBytesAvailable)
+                        {
+                            goto need_more;
+                        }
+
+                        if (++i >= inputCharacters)
+                            throw new ArgumentException("Invalid surrogate pair.", nameof(utf16));
+
+                        var lowSurrogate = *(utf16CharRef + i);
```
jkotas: Nit: Coding guidelines allow use of `var` only if the type shows up the right hand side. It is not the case here.

jkotas:
What is the exact benchmark that you are using for these measurements? I am able to reproduce your results (and also - they are surprisingly good given the implementation). I see: 3.4x slower for ASCII range, 1.7x slower for 0x7FF range, ... .

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: That is not necessary. I will remove it. It's an artifact of moving code around.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: @jkotas, Hmm, yes I did think about that ("bursty" ascii) which is why if you notice, there are two for loops, and once I see an ascii character, I assume the upcoming characters are also ascii unless observed otherwise.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: Yes, will change.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: Good ... just preserve these characteristics as you add the unrolling.

ahsonkhan:
@jkotas,
I use the method directly (copying over the implementation) in a separate .NET Core console application. The character string is randomly generated with characters within each range.

So for ASCII, midCodePoint = 0, maxCodePoint = 0x7F.

```C#
    public static long EncodingPerformanceTestUsingCorefxlabPointer(int Iterations, int minCodePoint, int maxCodePoint, bool useSpecialString = false)
    {
        string unicodeString = useSpecialString ? GetSpecialString() : GenerateString(CharLength, minCodePoint, maxCodePoint);
        ReadOnlySpan<char> characters = unicodeString.Slice();
        int utf8Length = CharLength * 4;
        byte[] utf8Buffer = new byte[utf8Length];
        Span<byte> span = new Span<byte>(utf8Buffer);

        Stopwatch timer = new Stopwatch();
        long duration = 0;

        if (!TryEncodeFromUtf16(characters, span, out int consumed, out int encodedBytes))
        {
            throw new Exception(); // this should not happen
        }

        timer.Restart();
        for (int i = 0; i < Iterations; i++)
        {
            if (!TryEncodeFromUtf16(characters, span, out consumed, out encodedBytes))
            {
                throw new Exception(); // this should not happen
            }
        }
        duration += timer.ElapsedMilliseconds;
#if OUTPUT
        Console.Write("EncodePointer (");
        Console.Write("{0:X2}-", minCodePoint);
        Console.Write("{0:X2}", maxCodePoint);
        Console.WriteLine("): " + duration + "ms");
#endif
        return duration;
    }
```

ahsonkhan:
> Also, please don't delete the Span based implementation. Please move it to a benchmark test. We want to be running this benchmark periodically to drive span performance down to be in the ballpark of the pointer one.

I wanted to keep it initially to see code diff. I will move it for performance before the PR is ready to merge.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: Putting it in a fixed block gives the following compiler error:

```C#
fixed (void* utf16Ref = Unsafe.AsPointer(ref utf16.DangerousGetPinnableReference()))
{
```

> You cannot use the fixed statement to take the address of an already fixed expression

jkotas: You should not be using `Unsafe.AsPointer` at all here (or pretty much never in any code in this project). It should be: 

```
            fixed (char * pUtf16 = &utf16.DangerousGetPinnableReference())
            fixed (byte * pUtf8 = &utf8.DangerousGetPinnableReference())
```

jkotas:
> I just call the following method 10x.

And what does the .NET Core counterpart that you are comparing this with looks like?

ahsonkhan:
> And what does the existing .NET Core counterpart that you are comparing this with looks like?

It is similar, calling the Encoding.UTF8.GetBytes() method.

image


```C#
    public static long EncodingPerformanceTestUsingCoreCLR(int Iterations, int minCodePoint, int maxCodePoint, bool useSpecialString = false)
    {
        string unicodeString = useSpecialString ? GetSpecialString() : GenerateString(CharLength, minCodePoint, maxCodePoint);
        //if (useSpecialString) Console.WriteLine(unicodeString);
        ReadOnlySpan<char> characters = unicodeString.Slice();

        Encoding utf8 = Encoding.UTF8;

        int utf8Length = utf8.GetByteCount(unicodeString);
        byte[] utf8Buffer = new byte[utf8Length];
        Span<byte> span;

        char[] charArray = characters.ToArray();
        Stopwatch timer = new Stopwatch();
        long duration = 0;

        utf8.GetBytes(charArray, 0, characters.Length, utf8Buffer, 0);
        span = new Span<byte>(utf8Buffer);

        timer.Restart();
        for (int i = 0; i < Iterations; i++)
        {
            utf8.GetBytes(charArray, 0, characters.Length, utf8Buffer, 0);
        }
        duration += timer.ElapsedMilliseconds;
#if OUTPUT
        Console.Write("CoreCLR (");
        Console.Write("{0:X2}-", minCodePoint);
        Console.Write("{0:X2}", maxCodePoint);
        Console.WriteLine("): " + duration + "ms");
#endif
        return duration;
}

```

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: @jkotas, the initial results of testing what you suggested are similar to the ASCII range (looking at around 100+ ascii chars with <10 non-ascii chars around them).
If the ratio of non-ascii characters increases, the perf gets closer to existing encoder.

ahsonkhan:
Question @KrzysztofCwalina, can we make the entire method unsafe, that is, change the signature to:
```C#
public static unsafe bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, out int charactersConsumed, out int bytesWritten)
```

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -21,6 +21,16 @@ internal static class Utf8Encoder
         private const byte b1110_0000U = 0xE0; //224
         private const byte b1111_0000U = 0xF0; //240
         private const byte b1111_1000U = 0xF8; //248
+        
+        private const ushort Utf16HighSurrogateFirstCodePoint = 0xD800;
+        private const ushort Utf16HighSurrogateLastCodePoint = 0xDBFF;
```
shiftylogic: If these are the same constants that exist in the UTF16 encoder, moving them to a common place makes sense. I only put them in the UTF16 encoder because I was working at getting rid of the UnicodeCodePoint.

I suggest to put them as internal constants on UnicodeHelpers.cs and rename that class to EncodingHelpers or something like that. Then we can put all the constants used across classes there.

ahsonkhan: That is what I was thinking. It makes sense :)
Shall I move the helper functions too (like IsSurrogate/etc)?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
shiftylogic: Yes, but for now, make everything internal. We talked about making them public, but it isn't really necessary at the moment and may not make sense.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: Making this change significantly drops the performance (almost back to Span Index).

image

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
shiftylogic: Yes, but for now, make everything internal. We talked about making them public, but it isn't really necessary at the moment and may not make sense.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: Can you please elaborate further on how this happens and help me find a way to resolve it?

jkotas: You need to write the code such that it handles the boundary conditions well. In this case, consider what will happen for: 

```
outputBytesAvailable = 0x7FFFFFFE;
bytesWrittenTemp = 0x7FFFFFFD;

if (bytesWrittenTemp + 3 > outputBytesAvailable) // bytesWrittenTemp + 3 will overflow will give you -MaxInt, and this condition will be incorrectly false
```

In this case, you can fix it by rewriting the condition as:
```
                  if (bytesWrittenTemp > outputBytesAvailable - 3)
```

Also, you may want to take some of the security trainings about security bugs caused by integer overflows… .

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: I likely means that your buggy implementation using AsPointer was not operating correctly - it might have been operating on corrupted data, hard to tell. You can make things run really-really fast when they do not have to work...

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: My security professor will be disappointed in me. I should have caught this. I suppose I was focusing on perf predominantly.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: I am validating correctness (at least for valid characters) and comparing the results of the encoding to the current implementation. The logic is also similar to code that passed the correctness tests in corefxlab.

jkotas: You are not doing this validation while running the perf test in the code fragment below. The bug with AsPointer is a stress bug. Everything will work fine when you run it just once; it will only break once you run longer and GC happens to move things around.

ahsonkhan: How can I expose this bug? I increased the character length to 100K characters and validating a thousand times (instead of once), the runtime is over 5 minutes.

jkotas: It is not very interesting to try to figure out why the buggy code is fast. 

Instead, you need to try to figure out why the correct code is slow. E.g. look at the disassembly and see where the cycles are getting wasted.

ahsonkhan: Looking at the IL, I realized why there was such a big performance degradation when I pinned the pointers and have resolved it. The results look more promising now.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -298,120 +298,110 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
-                            {
-                                if (bytesWrittenTemp > outputBytesAvailable - 2)
-                                {
-                                    goto need_more;
-                                }
-                                *(pUtf8 + bytesWrittenTemp) = (byte)(((codePoint >> 6) & b0001_1111U) | b1100_0000U);
-                                bytesWrittenTemp++;
-                                *(pUtf8 + bytesWrittenTemp) = (byte)(((codePoint) & b0011_1111U) | b1000_0000U);
-                                bytesWrittenTemp++;
+                                goto need_more;
```
jkotas: I do not think that the charactersConsumed will be always right if you take this path - even if `++i >= inputCharacters` is false, `i` is still going to be incremented. Make sure to add the test for this... .

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -298,120 +298,110 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                            bytesWrittenTemp++;
+                            *(bytesWrittenTemp + pUtf8) = (byte)(((surrogateCalculation) & b0011_1111U) | b1000_0000U);
+                            bytesWrittenTemp++;
+                        }
+                        else
+                        {
+                            if (bytesWrittenTemp > outputBytesAvailable - 3)
                             {
-                                throw new ArgumentException("Invalid surrogate pair.", nameof(utf16));
+                                goto need_more;
```
jkotas: Where did the handling of invalid surrogate pairs go? You will just keep asking for more now if there is malformed input. Make sure to add a test for this... .

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: Thanks for catching that. I will add the test cases for these corner cases soon along with full branch coverage.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: For this specific case, I realized we will never reach this else branch (hence I removed it). We already have a multiplexer for all char values between 0-FFFF in other if/else branches. 

Generally, talking with Robert/Krzysztof, we just return false and how much was consumed rather than throwing, when we get invalid surrogate pairs. The fallback/recovery would be built on top.

jkotas: You current implementation will encode invalid input that contains just low surrogates as bogus output. There is no way to build fallback/recovery behavior on top of it...

Here is the check that deals with this case in the CoreCLR implementation: https://github.com/dotnet/coreclr/blob/b638af3a4dd52fa7b1ea1958164136c72096c25c/src/mscorlib/src/System/Text/UTF8Encoding.cs#L713

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -288,55 +298,121 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                        codePoint = *(pUtf16 + i);
+
+                        if (codePoint > Utf8OneByteLastCodePoint)
+                        {
+                            goto non_ascii;
+                        }
+                        if (bytesWrittenTemp > outputBytesAvailable - 1)
+                        {
+                            goto need_more;
+                        }
```
KrzysztofCwalina: I think you could squeeze more perf by having a simpler loop over minimum of input characters and avaliable space (the minumum computed before entering the loop and recomputed whenever you detect a non ascii char)

jkotas: Yep, that's what the CoreCLR impl does. 

Once you start doing some of these optimizations, you will likely hit problems with register pressure: things will become slower because they won't fit into registers anymore. The fix for that is to use `pointer` that fits into one register instead of `base+index` that takes two registers.

ahsonkhan: I agree and I was playing around in this area to get some of the insights from the CoreCLR implementation but I am having trouble working around the output bytes available check. For me to accurately consume characters and set bytes written, I need to have this if (bytesWrittenTemp > outputBytesAvailabe - X) check everywhere because I can't be sure if the character is 1 byte or up to 4.

For example, in this loop unrolling in CoreCLR, how exactly do we check and return when the output buffer is full?
```C#
                    while (pSrc < pStop) {
                        ch = *(int*)pSrc;
                        int chc = *(int*)(pSrc+2);
                        if (((ch | chc) & unchecked((int)0xFF80FF80)) != 0) {
                            goto LongCodeWithMask;
                        }

                        // Unfortunately, this is endianess sensitive
#if BIGENDIAN
                        *pTarget = (byte)(ch>>16);
                        *(pTarget+1) = (byte)ch;
                        pSrc += 4;
                        *(pTarget+2) = (byte)(chc>>16);
                        *(pTarget+3) = (byte)chc;
                        pTarget += 4;
#else // BIGENDIAN
                        *pTarget = (byte)ch;
                        *(pTarget+1) = (byte)(ch>>16);
                        pSrc += 4;
                        *(pTarget+2) = (byte)chc;
                        *(pTarget+3) = (byte)(chc>>16);
                        pTarget += 4;
#endif // BIGENDIAN
                    }
```

tests/System.Text.Primitives.Tests/Encoding/Utf8EncoderTests.cs
```diff
@@ -4,6 +4,522 @@ namespace System.Text.Utf8.Tests
...
+            Assert.Equal(0, charactersConsumed);
+            Assert.Equal(0, bytesWritten);
+        }
+
+        [Fact]
+        public void TestEncodingInputBufferMaxSize()
+        {
+            int maxCharArraySize = int.MaxValue / 4; // TODO: want this to be int.MaxValue (need gcAllowVeryLargeObjects = true)
+            char[] charArray = new char[maxCharArraySize];
+
```
ahsonkhan: How can I get a int.MaxValue sized array so I can make sure there aren't Integer OverFlow issues resulting in incorrect encoding?
Right now CLR throws an OutofMemoryException for anything over 2GB.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -433,16 +509,16 @@ internal static bool TryEncodeCodePoint(uint codePoint, Span<byte> utf8, int ind
...
+            if (codePoint <= Utf8TwoBytesLastCodePoint)
                 return 2;
 
-            if (codePoint <= 0xFFFF)
+            if (codePoint <= Utf8ThreeBytesLastCodePoint)
                 return 3;
 
-            if (codePoint <= 0x1FFFFF)
+            if (codePoint <= Utf8FourBytesLastCodePoint)
                 return 4;
```
ahsonkhan: FYI: Utf8FourBytesLastCodePoint = 0x10FFFF, the previous GetNumberOfEncodedBytes() method was incorrect as it was using the constant value of 0x1FFFFF.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: As per the suggestion, I added a test case where the input buffer only contains low surrogates to ensure correctness in this scenario. There is also a test that contains a high surrogate which isn't paired with a low surrogate.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: I added a test case where the last character in the buffer is a high surrogate to cover this edge case.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: I will address this issue in a separate PR. Created an issue to track this: #1147

ahsonkhan:
> > Also, please don't delete the Span based implementation. Please move it to a benchmark test. We want to be running this benchmark periodically to drive span performance down to be in the ballpark of the pointer one.

> I wanted to keep it initially to see code diff. I will move it for performance testing before the PR is ready to merge.

I have created an issue to track this: #1148

tests/System.Text.Primitives.Tests/Encoding/Utf8EncoderTests.cs
jkotas: You either need more recent CoreCLR with for https://github.com/dotnet/coreclr/issues/8847 that enabled gcAllowVeryLargeObjects by default; or you can use unmanaged buffers like what Span tests in corefx do https://github.com/dotnet/corefx/blob/master/src/System.Memory/tests/Span/Clear.cs#L254

ahsonkhan:
@dotnet-bot Test Windows Debug please

ahsonkhan:
@dotnet-bot Test Windows Release please

ahsonkhan:
With the latest changes (loop unrolling and reducing output bounds check) we are now within 15% of the CoreCLR impl (in the ASCII range) and essentially on par in the rest (+/- 10%).

If the output buffer is more than 3x larger than the input buffer, the performance becomes on par, even in the ASCII range.

However, this still does not contain byte alignment for example:
https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/Text/UTF8Encoding.cs#L870

ahsonkhan:
cc @KrzysztofCwalina, @jkotas, @shiftylogic.

jkotas:
> The code is starting to look ugly with some duplication.

Yes, it does not look very promising. It is about as long as CoreCLR implementation, and it does not do everything yet. And it contains more bad patterns like goto labels compared to the CoreCLR implementation...

I have submitted a PR with the implementation based on the CoreCLR decoder with what I think should be used as the baseline optimized implementation.

ahsonkhan:
We should still consider adding the tests from this PR.

jkotas:
> We should still consider adding the tests from this PR.

Agree. Could you please redo the PR to contain just the tests?

ahsonkhan:
@dotnet-bot test this please.

PR #1140, opened 2017-01-20
Remove the single char variant of TryEncode from TextEncoder

shiftylogic:
Removing this variant force people to think about whether they really want to encode a single character. We could consider an extension method outside System.Text.Primitives core code base. For now, most of the users were in prototypes, tests, or somewhere that wrapping the "unsafe" portion was easy.

### DIFF
--- a/src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
+++ b/src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
@@ -335,7 +335,7 @@ namespace System.Text.Json
     static class Utf8StringExtensions
     {
         private static readonly char[] EncodingChars = new char[] { '"' };
-        private static readonly ReadOnlySpan<char> QuoteSpan = new ReadOnlySpan<char>(EncodingChars, 4, 1);
+        private static readonly ReadOnlySpan<char> QuoteSpan = new ReadOnlySpan<char>(EncodingChars, 0, 1);
 
         // TODO: this should be properly implemented 
         // currently it handles formatting to UTF8 only.
--- a/src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
+++ b/src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
@@ -10,7 +10,17 @@ namespace System.Text.Json
 {
     public class JsonDynamicObject : DynamicObject, IBufferFormattable
     {
-        Dictionary<JsonProperty, JsonValue> _properties;
+        #region Constants
+
+        private static readonly char[] EncodingChars = new char[] { '{', '}', ',', ':', '"' };
+        private static readonly ReadOnlySpan<char> OpenCurlySpan = new ReadOnlySpan<char>(EncodingChars, 0, 1);
+        private static readonly ReadOnlySpan<char> CloseCurlySpan = new ReadOnlySpan<char>(EncodingChars, 1, 1);
+        private static readonly ReadOnlySpan<char> CommaSpan = new ReadOnlySpan<char>(EncodingChars, 2, 1);
+        private static readonly ReadOnlySpan<char> ColonSpan = new ReadOnlySpan<char>(EncodingChars, 3, 1);
+
+        #endregion Constants
+
+        private readonly Dictionary<JsonProperty, JsonValue> _properties;
 
         public JsonDynamicObject() : this(new Dictionary<JsonProperty, JsonValue>()) { }
 
@@ -167,8 +177,9 @@ namespace System.Text.Json
         public bool TryFormat(Span<byte> buffer, out int written, TextFormat format, EncodingData formattingData)
         {
             written = 0;
+            int consumed;
             int justWritten;
-            if(!formattingData.TextEncoder.TryEncode('{', buffer, out justWritten)) {
+            if(!formattingData.TextEncoder.TryEncode(OpenCurlySpan, buffer, out consumed, out justWritten)) {
                 return false;
             }
             written += justWritten;
@@ -181,7 +192,7 @@ namespace System.Text.Json
                 if(firstProperty) { firstProperty = false; }
                 else
                 {
-                    if (!formattingData.TextEncoder.TryEncode(',', buffer.Slice(written), out justWritten))
+                    if (!formattingData.TextEncoder.TryEncode(CommaSpan, buffer.Slice(written), out consumed, out justWritten))
                     {
                         return false;
                     }
@@ -193,7 +204,7 @@ namespace System.Text.Json
                     written = 0; return false;
                 }
                 written += justWritten;
-                if (!formattingData.TextEncoder.TryEncode(':', buffer.Slice(written), out justWritten))
+                if (!formattingData.TextEncoder.TryEncode(ColonSpan, buffer.Slice(written), out consumed, out justWritten))
                 {
                     return false;
                 }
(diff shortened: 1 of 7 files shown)


src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -10,7 +10,17 @@ namespace System.Text.Json
 {
     public class JsonDynamicObject : DynamicObject, IBufferFormattable
     {
-        Dictionary<JsonProperty, JsonValue> _properties;
+        #region Constants
```
shiftylogic: FYI, this project does not have unsafe enabled (yet?) and I didn't really think it was necessary right now. If you don't like this pattern, I can change it.

shiftylogic:
@KrzysztofCwalina @ahsonkhan

shiftylogic:
@KrzysztofCwalina @ahsonkhan

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -10,7 +10,17 @@ namespace System.Text.Json
...
+
+        private static readonly char[] EncodingChars = new char[] { '{', '}', ',', ':', '"' };
+        private static readonly ReadOnlySpan<char> OpenCurlySpan = new ReadOnlySpan<char>(EncodingChars, 0, 1);
+        private static readonly ReadOnlySpan<char> CloseCurlySpan = new ReadOnlySpan<char>(EncodingChars, 1, 1);
+        private static readonly ReadOnlySpan<char> CommaSpan = new ReadOnlySpan<char>(EncodingChars, 2, 1);
+        private static readonly ReadOnlySpan<char> ColonSpan = new ReadOnlySpan<char>(EncodingChars, 3, 1);
```
KrzysztofCwalina: I think this is fine for now, but once the compiler checks are implemented, you won't be able to have span fields in classes. At that point, this will become expensive. 

I think it might be worth adding single character to to code-point encode methods to the encoders. What we don't want is single character to a buffer.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -167,8 +177,9 @@ public override bool TrySetMember(SetMemberBinder binder, object value)
         public bool TryFormat(Span<byte> buffer, out int written, TextFormat format, EncodingData formattingData)
         {
             written = 0;
+            int consumed;
             int justWritten;
-            if(!formattingData.TextEncoder.TryEncode('{', buffer, out justWritten)) {
+            if(!formattingData.TextEncoder.TryEncode(OpenCurlySpan, buffer, out consumed, out justWritten)) {
```
KrzysztofCwalina: Yeah, it would be good to measure the perf impact of this.

shiftylogic: From what live code to this change, it is equal or faster with change. The reason is that TryEncode with a single char just wrapped the char in a span and called the span version anyways.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
shiftylogic: This would be part of the existing JSON work item. I suspect that there is probably work that needs to be done it we productive this JSON serializer.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: Is there a work item to make a fast single char encoder rather than having the implementation that you removed?

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -323,6 +334,9 @@ public bool TryFormat(Span<byte> buffer, out int written, TextFormat format, Enc
 
     static class Utf8StringExtensions
     {
+        private static readonly char[] EncodingChars = new char[] { '"' };
+        private static readonly ReadOnlySpan<char> QuoteSpan = new ReadOnlySpan<char>(EncodingChars, 0, 1);
+
```
ahsonkhan: Why not create and use the QuoteSpan from the EncodingChars on line 15?

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -10,7 +10,17 @@ namespace System.Text.Json
 {
     public class JsonDynamicObject : DynamicObject, IBufferFormattable
     {
-        Dictionary<JsonProperty, JsonValue> _properties;
+        #region Constants
+
+        private static readonly char[] EncodingChars = new char[] { '{', '}', ',', ':', '"' };
```
ahsonkhan: Do we ever use the '"' char here?

tests/System.Text.Formatting.Tests/CustomTypeFormatting.cs
```diff
@@ -25,8 +25,14 @@ public bool TryFormat(Span<byte> buffer, out int bytesWritten, TextFormat format
...
 
             char symbol = _inMonths ? 'm' : 'y';
+            int consumed;
             int symbolBytes;
-            if (!encoding.TextEncoder.TryEncode(symbol, buffer.Slice(bytesWritten), out symbolBytes)) return false;
+
+            unsafe
+            {
+                ReadOnlySpan<char> symbolSpan = new ReadOnlySpan<char>(&symbol, 1);
+                if (!encoding.TextEncoder.TryEncode(symbolSpan, buffer.Slice(bytesWritten), out consumed, out symbolBytes)) return false;
```
ahsonkhan: Wrap return false in curly brackets? Looking at other areas in this PR, for consistency.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
shiftylogic: There is no work item for this at the moment. However, I think it is best we see some real scenarios first and determine if it is necessary. Is it possible to implement a high performance JSON processor without a single-character method?

We can always add APIs, but it is much harder to remove them.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
shiftylogic: It was in a different class. None of this code will stay long-term. This change is a move towards getting rid of an undesirable method.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
shiftylogic: I added that because of the other usage initially then realized I was in 2 separate classes and forgot to remove it. I'll strip it out.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
jkotas: > but once the compiler checks are implemented

Fast span in the runtime will fail on this as well.

shiftylogic: Hmm... okay. I'll do the less desirable way then to avoid problems.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: Just going off from Krzysztof's comment:
> but there are real cases where we need to encode single chars, and we should think how to do it efficiently without creating a performance trap 

But I agree that we can add it later when required.

ahsonkhan:
LGTM

PR #1142, opened 2017-01-21
Replace custom parsing routine in UrlEncoder with PrimitiveParser routine

botaberg:
In order to unescape percent encodings, UrlEncoder had a custom implementation of a Utf8 hex parser. This proposed change replaces this custom parser with PrimitiveParser.InvariantUtf8.Hex.TryParseByte, which can accomplish the same goal.

This is part of a broader effort to remove all custom parsers from corefxlab components: see #1070.

### DIFF
--- a/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
+++ b/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
@@ -251,80 +251,29 @@
             {
                 return -1;
             }
+            
+            if (scan + 2 > buffer.Length)
+            {
+                return -1;
+            }
+            
+            byte value;
+            int bytesConsumed;
+            bool result = PrimitiveParser.InvariantUtf8.Hex.TryParseByte(buffer.Slice(scan, 2), out value, out bytesConsumed);
 
-            var probe = scan;
-
-            var value1 = ReadHex(ref probe, buffer);
-            if (value1 == -1)
+            if (result == false || bytesConsumed != 2)
             {
                 return -1;
             }
 
-            var value2 = ReadHex(ref probe, buffer);
-            if (value2 == -1)
-            {
-                return -1;
-            }
-
-            if (SkipUnescape(value1, value2))
-            {
-                return -1;
-            }
-
-            scan = probe;
-            return (value1 << 4) + value2;
-        }
-
-
-        /// <summary>
-        /// Read the next char and convert it into hexadecimal value.
-        /// 
-        /// The <paramref name="scan"/> index will be moved to the next
-        /// byte no matter no matter whether the operation successes.
-        /// </summary>
-        /// <param name="scan">The index of the byte in the buffer to read</param>
-        /// <param name="buffer">The byte span from which the hex to be read</param>
-        /// <returns>The hexadecimal value if successes, otherwise -1.</returns>
-        private static int ReadHex(ref int scan, Span<byte> buffer)
-        {
-            if (scan == buffer.Length)
-            {
-                return -1;
-            }
-
-            var value = buffer[scan++];
-            var isHead = ((value >= '0') && (value <= '9')) ||
-                         ((value >= 'A') && (value <= 'F')) ||
-                         ((value >= 'a') && (value <= 'f'));
-
-            if (!isHead)
-            {
-                return -1;
-            }
-
-            if (value <= '9')
-            {
-                return value - '0';
-            }
-            else if (value <= 'F')
-            {
-                return (value - 'A') + 10;
-            }
-            else // a - f
-            {
-                return (value - 'a') + 10;
-            }
-        }
-
-        private static bool SkipUnescape(int value1, int value2)
-        {
             // skip %2F - '/'
-            if (value1 == 2 && value2 == 15)
+            if (value == 0x2F)
             {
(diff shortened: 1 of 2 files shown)


botaberg:
@KrzysztofCwalina PTAL

KrzysztofCwalina:
Have you measured the performance impact of this change?

KrzysztofCwalina:
@botaberg, can you answer the question?

botaberg:
@KrzysztofCwalina I haven't measured the performance impact of the change, or looked through the performance test suite to see if there are existing tests for this scenario.

botaberg:
@KrzysztofCwalina It looks like there are no performance tests for the UrlEncoder. We may be able to make a simple performance test suite by following the patterns in System.Text.Encoding.Web.Utf8.Tests.

KrzysztofCwalina:
Yeah, that would be great.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -251,80 +251,29 @@ private static int UnescapePercentEncoding(ref int scan, Span<byte> buffer)
             {
                 return -1;
             }
-
-            var probe = scan;
-
-            var value1 = ReadHex(ref probe, buffer);
-            if (value1 == -1)
+            
+            if (scan + 2 > buffer.Length)
```
shiftylogic: Possible integer overflow here. You should change this to be a subtraction on Length instead, or you end up having to do extra checks for overflow.

Also, the increment above has same issue.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -251,80 +251,29 @@ private static int UnescapePercentEncoding(ref int scan, Span<byte> buffer)
...
-        /// The <paramref name="scan"/> index will be moved to the next
-        /// byte no matter no matter whether the operation successes.
-        /// </summary>
-        /// <param name="scan">The index of the byte in the buffer to read</param>
-        /// <param name="buffer">The byte span from which the hex to be read</param>
-        /// <returns>The hexadecimal value if successes, otherwise -1.</returns>
-        private static int ReadHex(ref int scan, Span<byte> buffer)
-        {
-            if (scan == buffer.Length)
+            if (result == false || bytesConsumed != 2)
```
ahsonkhan: Why not just have !result || bytesConsumed != 2 rather than doing the comparison to false?

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
botaberg: Thanks! Will follow up with an update.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
botaberg: 'result == false' and '!result' produce the same IL:
```
  IL_003d:  brfalse.s  IL_0043
  IL_003f:  ldloc.1
  IL_0040:  ldc.i4.2
  IL_0041:  beq.s      IL_0045
  IL_0043:  ldc.i4.m1
  IL_0044:  ret
```
I personally prefer 'result == false', but will defer to any style guidelines in place.

ahsonkhan: OK. I just haven't seen this style used anywhere and I would imagine it is more prone to errors like result = false instead of result == false. I am not sure if there are specific guidelines around this. I would be curious if one exists.

haha: "...but this question will likely solicit debate, arguments, polling, or extended discussion."
http://stackoverflow.com/questions/7535818/using-if-bool-vs-if-bool-false-in-c-sharp

botaberg: Good point! I searched corefxlab and found many instances where 'if (!booleanVar)' is used and only one instance (besides mine) where 'if (booleanVar == false)' is used. I'll change this PR to use the other pattern.

botaberg:
I have done a performance analysis on this change and reported on it in this gist:

https://gist.github.com/botaberg/e5b909260719942714b94bde32bde588

The 'inner loop' of PrimitiveParser is sound, but it is bundled up in such a way that it is hard to take advantage of it in this parsing scenario. Please read the description of the investigation - scroll down to the bottom for my comment.

shiftylogic:
I had another chance to look at this code. A problem I noticed is that it is effectively doing a double copy. The first thing it does in Decode is copy the entire source into the destination. Then it does in-place decoding. As soon as it hits the first percent encoding, the algorithm is now copying data within the destination buffer for the entire rest of the decoding process.

You should attempt to implement the decode method to operate on the spans without the first copy. That would also make it easy to have an internal implementation that works on raw pointers for the underlying span data, which would likely speed this up even more.

ahsonkhan:
@botaberg, any updates to this PR?

ahsonkhan:
@botaberg, do we want to merge this PR?

ahsonkhan:
@botaberg, should we close this PR?

ahsonkhan:
@botaberg, @shiftylogic, please re-open if needed. Closing.

PR #1143, opened 2017-01-21
Fix != operators in Memory<T>

someone-with-default-username:


KrzysztofCwalina:
Wow, how did that survive for so long. Thanks for catching it and submitting a fix. Could you just please add a test to make sure does not ever regress? Thanks.

someone-with-default-username:
I have not really checked out repo. Just was reading code in web browser :)

KrzysztofCwalina:
ok, I will add the tests.

PR #1150, opened 2017-01-27
[WIP] Add backpressure to pipelines

pakrym:
Still not sure about Flush returning bool

@davidfowl @halter73

### DIFF
--- a/src/System.IO.Pipelines/FlushAsyncAwaitable.cs
+++ b/src/System.IO.Pipelines/FlushAsyncAwaitable.cs
@@ -1,12 +1,11 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Runtime.CompilerServices;
 
 namespace System.IO.Pipelines
 {
-    public struct FlushAsyncAwaitable : ICriticalNotifyCompletion
+    internal struct FlushAsyncAwaitable : ICriticalNotifyCompletion
     {
         private readonly IFlushAwaiter _awaiter;
 
--- a/src/System.IO.Pipelines/FlushAsyncAwaitable.cs
+++ b/src/System.IO.Pipelines/FlushAsyncAwaitable.cs
@@ -1,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace System.IO.Pipelines
+{
+    public struct FlushAsyncAwaitable : ICriticalNotifyCompletion
+    {
+        private readonly IFlushAwaiter _awaiter;
+
+        public FlushAsyncAwaitable(IFlushAwaiter awaiter)
+        {
+            _awaiter = awaiter;
+        }
+
+        public bool IsCompleted => _awaiter.IsCompleted;
+
+        public bool GetResult() => _awaiter.GetResult();
+
+        public FlushAsyncAwaitable GetAwaiter() => this;
+
+        public void UnsafeOnCompleted(Action continuation) => _awaiter.OnCompleted(continuation);
+
+        public void OnCompleted(Action continuation) => _awaiter.OnCompleted(continuation);
+    }
+}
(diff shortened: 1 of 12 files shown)


src/System.IO.Pipelines/PipelineFactory.cs
```diff
@@ -17,6 +17,7 @@ namespace System.IO.Pipelines
     public class PipelineFactory : IDisposable
     {
         private readonly IBufferPool _pool;
+        public int MaximumPipeSize { get; set; }
```
davidfowl: Should be an argument to create

pakrym:
@davidfowl added min/max, internal

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -52,15 +53,23 @@ public class Pipe : IPipelineReader, IPipelineWriter, IReadableBufferAwaiter
...
         /// <summary>
         /// Initializes the <see cref="Pipe"/> with the specifed <see cref="IBufferPool"/>.
         /// </summary>
         /// <param name="pool"></param>
-        public Pipe(IBufferPool pool)
+        /// <param name="maximumSize"></param>
+        public Pipe(IBufferPool pool, long maximumSizeLow = 0, long maximumSizeHigh = 0)
         {
             _pool = pool;
+            _maximumSizeHight = maximumSizeHigh;
```
davidfowl: _maximumSizeHight -> _maximumSizeHigh

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -393,6 +432,9 @@ public void AdvanceReader(ReadCursor consumed, ReadCursor examined)
 
             if (!consumed.IsDefault)
             {
+                var consumedBytes = ReadCursor.GetLength(_readHead, _readHead.Start, consumed.Segment, consumed.Index);
+                Interlocked.Add(ref _length, -consumedBytes);
```
davidfowl: The locking around length is inconsistent. In some cases you update it inside of _sync and in this case you're using an interlocked. Move this code lower into the lock.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -436,55 +475,32 @@ public void AdvanceReader(ReadCursor consumed, ReadCursor examined)
...
-        void IPipelineWriter.Complete(Exception exception) => CompleteWriter(exception);
 
-        /// <summary>
-        /// Marks the pipeline as being complete, meaning no more items will be written to it.
-        /// </summary>
-        /// <param name="exception">Optional Exception indicating a failure that's causing the pipeline to complete.</param>
-        public void CompleteWriter(Exception exception = null)
-        {
-            if (_producingState != State.NotActive)
+            if (_length < _maximumSizeLow)
```
davidfowl: Here as well.

src/System.IO.Pipelines/FlushAsyncAwaitable.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+using System.Runtime.CompilerServices;
+
+namespace System.IO.Pipelines
+{
+    public struct FlushAsyncAwaitable : ICriticalNotifyCompletion
```
KrzysztofCwalina: Some of these types are what I would call "plumbing". We should consider moving them to some obscure sub namespace. But let's chat about it when we do an API review.

pakrym: I'll make it internal for now.

tests/System.IO.Pipelines.Tests/BackpressureTests.cs
```diff
@@ -0,0 +1,141 @@
...
+            var result = _pipe.ReadAsync().GetAwaiter().GetResult();
+            var consumed = result.Buffer.Move(result.Buffer.Start, 33);
+            _pipe.AdvanceReader(consumed, consumed);
+
+            Assert.True(flushAsync.IsCompleted);
+            Assert.True(flushAsync.Result);
+        }
+
+        [Fact]
+        public void FlushAsyncAwaitableDoesNotCompletesWhenReaderAdvancesUnderHight()
```
davidfowl: High*

PR #1151, opened 2017-01-27
Update corefxlab to VS2017.

crummel:
- Upgrade xproj/project.json to csproj.
- Upgrade CLI version to one that supports dotnet msbuild.
- Make minor changes to build scripts to support dotnet msbuild.
- Upgrade dotnet-install to latest version.
- This builds and debugs in VS2017.
- This builds from the command line with the upgraded CLI.
- Tests work and pass, with the exception of:
  - Tests that reference xunit.performance.core can not be discovered and will not be run in either VS or the command line.
  - Tests that use resx files fail with a MissingManifestResourceException in VS and will not be discovered on the command line.
- This does not include buildtools changes.


davidfowl:
These project files need some love. Did you very the automatic migration produces identical assets as the project.json.

crummel:
@davidfowl I spot-checked the properties and those all seemed fine.  For functional stuff, the project.json build appears to have included transitive references implicitly, which was not the case after conversion, so those all had to be added, as well as some framework references which I would have thought would also be needed in the project.jsons.  Upgrading to NETCore.Library 1.1.0 also caused a lot of version conflicts on existing references, so I updated those (mostly to 4.3.0 or latest across the board, except for System.Reflection.Metadata.Cil, which depends on an old version of System.Reflection.Metadata).  The auto-migration also had some issues with the projects that don't include all the .cs files in their directory, so I manually fixed those.  So I don't think they're identical, but given that we're upgrading I think I got as close as possible.  I logged https://github.com/dotnet/corefxlab/issues/1155 to fix the remaining test issues.

Issue #1155, opened 2017-01-27
Tests need to be fixed for VS2017 upgrade


crummel:
I disabled some tests that were failing after the upgrade.
For individual tests, they are tagged with "VS2017".
I also took projects that used xunit.performance.core out of the build, I believe these were failing because that package needs to be upgraded to the latest SDK as well:
- Benchmarks
- System.Binary.Base64.Tests
- System.Slices.Tests
- System.Text.Json.Tests

Projects that used RESX files also failed:
- System.CommandLine.Tests
- System.Collections.Generic.MultiValueDictionary.Tests
- System.IO.Pipelines.Tests
- System.Text.Formatting.Globalization.Tests
- System.Time.Tests

I'm not familiar enough with resource stuff to know why those are an issue.

ahsonkhan:
System.Text.Primitives.Tests can also not be discovered since it references xunit.performance.core

ahsonkhan:
System.Text.Json.Tests do not reference xunit.performance.core. They shouldn't be skipped.

davidfowl:
@ahsonkhan is this done?

ahsonkhan:
I need to remove performance tests (and dependency to xunit.performance.core) from the following projects and move them to Benchmarks:
System.Binary.Base64.Tests
System.Slices.Tests

These tests (along with the functional unit tests within these projects) don't show up in VS test explorer as long as they have benchmark tests in them. I am trying to resolve the issue with the xunit.perf team.

This should be complete:

> Projects that used RESX files also failed:
System.CommandLine.Tests
System.Collections.Generic.MultiValueDictionary.Tests
System.IO.Pipelines.Tests
System.Text.Formatting.Globalization.Tests
System.Time.Tests`

ahsonkhan:
This issue has been fixed by updating the package reference to stable versions of xunit and xunit.runner.visualstudio, and by updating to a new xunit.performance.core package (along with PRs referenced in this issue).

PR #1156, opened 2017-01-29
UTF8 encoder based on the CoreCLR implementation

jkotas:


jkotas:
cc @KrzysztofCwalina, @jkotas, @shiftylogic

I took the CoreCLR implementation of the UTF8 encoder and fixed it up to fit the Span-based encoder signature with minimal refactoring. It is ~2x faster for the ASCII range than current https://github.com/dotnet/corefxlab/pull/1139.

@ahsonkhan Could you please run your perf tests on it?

jkotas:
Also, see my comments in https://github.com/dotnet/coreclr/pull/9187 . I think we should work towards having same workhorse method for here as well as in CoreCLR. We will be able to share any improvements made here with CoreCLR that way.

KrzysztofCwalina:
Looks good. 

Just one question: do you think it would be possible to share the source with the corclr implementation? Even if it just meant to be able to easily copy and paste a whole file between these two repos.

jkotas:
> do you think it would be possible to share the source with the corclr implementation? 

Yes. It is what I am basically asking for in https://github.com/dotnet/coreclr/pull/9187. (The side-effect is that it will make the CoreCLR implementation faster.)

ahsonkhan:
> @ahsonkhan Could you please run your perf tests on it?

So, I ran my perf test and this implementation is indeed faster (faster than the current CoreCLR impl as well) :)

image

Given that the implementation is faster and similar in complexity/code length as the state I ended up reaching, it makes sense to use this over mine.

However, looking at the different encoder/decoders in System.Text.Primitives, how do we plan to share the implementations or keep them in sync? And if we aren't doing that, how do we go about maintaining them separately? I could imagine us having implementations that could diverge from CoreCLR (different signatures, other encodings).

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -288,55 +310,274 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
-            for (int i = 0; i < availableCharacters; i++)
-            {
-                var ch = utf16[i];
+                // assume that JIT will enregister pSrc, pTarget and ch
 
-                if ((ushort)ch <= 0x7f) // this if block just optimizes for ascii
+                // don't fall into the fast decoding loop if we don't have enough characters
+                // Note that if we don't have enough bytes, pStop will prevent us from entering the fast loop.
+                while (pSrc < pEnd - 13)
                 {
```
ahsonkhan: Can you explain, maybe in comments, why 13 is special in this case?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -288,55 +310,274 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                            pTarget += 4;
+#endif // BIGENDIAN
+                        }
+                        continue;
+
+                        LongCodeWithMask:
+#if BIGENDIAN
+                        // be careful about the sign extension
+                        ch = (int)(((uint)ch) >> 16);
+#else // BIGENDIAN
```
ahsonkhan: What is sign extension?

jkotas:
> how do we plan to share the implementations or keep them in sync? 

Yes, we should try to use same the core encoding/decoding methods between the different places and avoid diverging them. Check the discussion in 
https://github.com/dotnet/coreclr/pull/9187#discussion_r98521142

> And if we aren't doing that, how do we go about maintaining them separately?

The low-tech solution to do that is to add a comment that if you change one copy, you should update other copy to match. I do not expect that the encoders/decoders will change that frequently, so it should work fine to do it that way.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -269,6 +270,27 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+        // byte* flavor just for parity
+        unsafe private static int PtrDiff(byte* a, byte* b)
+        {
+            return (int)(a - b);
+        }
+
+        private static bool InRange(int ch, int start, int end)
+        {
+            return (uint)(ch - start) <= (uint)(end - start);
+        }
```
ahsonkhan: Should these methods go into the Encoder Helper class?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -288,55 +310,274 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                                chd = unchecked((sbyte)0xE0) | (ch >> 12);
+                            }
+                            else
+                            {
+                                // 4 byte encoding - high surrogate + low surrogate
+                                // if (!IsHighSurrogate(ch))
+                                if (ch > HIGH_SURROGATE_END)
+                                {
+                                    // low without high -> bad, try again in slow loop
+                                    goto NeedMore;
```
ahsonkhan: Comment needs updating.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: https://en.wikipedia.org/wiki/Sign_extension

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -288,55 +310,274 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                            if (pTarget >= pAllocatedBufferEnd - 3)
+                                goto NeedMore;
+
+                            // 4 byte encoding - high surrogate + low surrogate
+                            // if (!IsHighSurrogate(ch))
+                            if (ch > HIGH_SURROGATE_END)
+                            {
+                                // low without high -> bad, try again in slow loop
+                                goto NeedMore;
+                            }
```
ahsonkhan: Comment needs updating.

jkotas: Fixed.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: Fixed.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: Maybe - once there is a good reason for it.

jkotas: It will make it harder to share the same implementation between different places though... .

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: Done.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -288,55 +310,278 @@ internal static bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int by
...
+                            ch = *(int*)pSrc;
+                            int chc = *(int*)(pSrc + 2);
+                            if (((ch | chc) & unchecked((int)0xFF80FF80)) != 0)
+                            {
+                                goto LongCodeWithMask;
+                            }
+
+                            // Unfortunately, this is endianess sensitive
+#if BIGENDIAN
+                            *pTarget = (byte)(ch >> 16);
```
ahsonkhan: Do we need to worry about endianness if we only support encoding/decoding utf16 LE?

ahsonkhan:
LGTM.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: .NET Core runs on little endian machines only today, and the .NET Core framework makes assumptions about it - either silently, or by throwing when running on big endian machine. So we do not worry about it yet, but we may need to start worrying about it eventually.

benaadams: Also dropping it would be a change from current coreclr if the method is shared with coreclr.

jkotas:
@dotnet-bot test this please

ahsonkhan:
> The low-tech solution to do that is to add a comment that if you change one copy, you should update other copy to match. I do not expect that the encoders/decoders will change that frequently, so it should work fine to do it that way.

Can we add such a comment here so it is explicit? Also in CoreCLR?

ahsonkhan:
@dotnet-bot test this please

jkotas:
@dotnet-bot test Windows Release  please

jkotas:
@dotnet-bot test this please

ahsonkhan:
@dotnet-bot test this please

ahsonkhan:
I am not sure why the CI keeps failing.

jkotas:
@ahsonkhan I do not see the tests that you have added running. When I run `dotnet test` in under tests\System.Text.Primitives.Tests, it just prints `Starting test execution, please wait...` without any signs of the tests running. How are you running these tests?

ahsonkhan:
I believe that the latest VS 2017 build changes (PR #1151) have disabled System.Text.Primitives.Tests (along with some others). I think it is because System.Text.Primitives.Tests references xunit.performance.core. 
> Tests that reference xunit.performance.core can not be discovered and will not be run in either VS or the command line.

I created a separate local project and added an xunit project where I was writing/running these tests (along with the ref return method to code the TryEncode method).

So, currently, I don't think the tests are running. However, I did copy over your method implementation into my standalone app and ran my new tests against it.

There is an issue open for this, but I am not sure how to make progress on it yet (Issue #1155).

jkotas:
>  did copy over your method implementation into my standalone app and ran my new tests against it

Thanks!

PR #1158, opened 2017-01-30
API Cleanup

shiftylogic:
* Normalizing argument order between Parsing / Formatting.
* Slight tweaks to formatting public surface.


shiftylogic:
@ahsonkhan @KrzysztofCwalina @botaberg

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.cs
```diff
@@ -15,39 +15,39 @@ namespace System.Text
     public static partial class PrimitiveFormatter
     {
         #region Integers
-        public static bool TryFormat(this byte value, Span<byte> buffer, out int bytesWritten, TextFormat format, EncodingData encoding)
+        public static bool TryFormat(this byte value, Span<byte> buffer, out int bytesWritten, TextFormat format = default(TextFormat), EncodingData encoding = default(EncodingData))
```
KrzysztofCwalina: How did you chose between the two alternative orderings we could use? i.e. why do you think format followed by encoding is better than the other way around?

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
```diff
@@ -1,4 +1,10 @@
 ﻿
+
+
+
+
+
+
```
KrzysztofCwalina: Why do we have these empty lines?

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
```diff
@@ -1,4 +1,10 @@
 ﻿
+
+
+
+
+
+
```
KrzysztofCwalina: same

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.cs
shiftylogic: I suspect the case of passing an EncodingData is going to be much less frequent than formatting information. So, I'd argue to put the more commonly used parameter first.

I could be swayed with a better argument.

KrzysztofCwalina: Sounds good. Thanks.

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
shiftylogic: I suspect that is an artifact of the T4 templates. It is one of my dislikes of T4 templates. The template engine does a very poor job of handling spacing / newlines / etc.

src/System.Text.Primitives/System/Text/Parsing/Unsigned.cs
shiftylogic: I suspect that is an artifact of the T4 templates. It is one of my dislikes of T4 templates. The template engine does a very poor job of handling spacing / newlines / etc.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -273,9 +273,9 @@ public bool TryFormat(Span<byte> buffer, out int written, TextFormat format, Enc
                 switch (_type)
                 {
                     case JsonReader.JsonValueType.String:
-                        return _value.TryFormatQuotedString(buffer, format, formattingData, out written);
+                        return _value.TryFormatQuotedString(buffer, out written, format, formattingData: formattingData);
                     case JsonReader.JsonValueType.Number:
-                        return _value.TryFormat(buffer, format, formattingData, out written);
+                        return _value.TryFormat(buffer, out written, format, formattingData: formattingData);
                     case JsonReader.JsonValueType.Object:
```
ahsonkhan: I thought the guideline was to have out parameters be at the end. Is that not the case?

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.cs
```diff
@@ -15,39 +15,39 @@ namespace System.Text
...
         {
             return IntegerFormatter.TryFormatUInt64(value, 1, buffer, out bytesWritten, format, encoding);
         }
 
-        public static bool TryFormat(this sbyte value, Span<byte> buffer, out int bytesWritten, TextFormat format, EncodingData encoding)
+        public static bool TryFormat(this sbyte value, Span<byte> buffer, out int bytesWritten, TextFormat format = default(TextFormat), EncodingData encoding = default(EncodingData))
         {
```
ahsonkhan: I thought the guideline was to have out parameters be at the end. Is that not the case? On the other hand, I also thought that default parameters go in the end. Is this the recommended approach when you have both?

KrzysztofCwalina: This is the official .NET guideline, but it means the parameters could not be defaulted. The guidelines were developed before we allowed default parameters in public APIs, and now these two guidelines (ok to use default and out parameters being last) seem not compatible with each other.

@terrajobst, we should talk about the guidelines at some point and see if we need to tweak them. But for now, I think this change is ok, and will give us more experience when we talk about the guidelines and how to adjust them.

PR #1161, opened 2017-01-30
Fix packaging to look for csprojs instead of project.jsons.

crummel:


scripts/package.ps1
```diff
@@ -20,7 +20,7 @@ Function Ensure-Nuget-Exists {
 }
 
 Write-Host "** Building all NuGet packages. **"
-foreach ($file in [System.IO.Directory]::EnumerateFiles("$repoRoot\src", "project.json", "AllDirectories")) {
+foreach ($file in [System.IO.Directory]::EnumerateFiles("$repoRoot\src", "System*.csproj", "AllDirectories")) {
```
davidfowl: Why not *.csproj?

crummel: Microsoft.CodeAnalysis.StackOnlyTypes is the exception.  It was always an old-style project, so the old project.json build and package scripts didn't do anything with it.  It's also not part of the corefxlab solution so it doesn't build in VS.

KrzysztofCwalina:
thanks!

PR #1166, opened 2017-02-01
Added the ability to schedule callbacks on an IScheduler

davidfowl:
- To avoid forcing the consuming code to dispatch onto the right
thread, we add the ability for the pipe to run the callback on an
IScheduler via IScheduler.Schedule.
- It's more efficient to do it in the pipe implementation itself
rather than on the consuming side (things don't need to be scheduled
if data is already available).
- The default scheduler is the inline scheduler (the consumer runs on the
producer's thread)
- Also added a TaskRunScheduler which does Task.Run.
- Changed the libuv networking implementation to schedule outbound callbacks on
libuv thread. Made UvThread an IScheduler.
- Added PipeOptions to remove the number of overloads on PipelineFactory.Create


src/System.IO.Pipelines/PipeOptions.cs
```diff
@@ -0,0 +1,17 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    public class PipeOptions
```
KrzysztofCwalina: Why is it a class? Could it be an immutable struct?

davidfowl: The only problem with making it a struct is that I'd have to make all of the properties set during construction. I don't really want to require that sine this was about making these parameters optional.

KrzysztofCwalina: Would it be ok if these were settable after the pipe is constructed?

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -62,24 +64,33 @@ public class Pipe : IPipelineReader, IPipelineWriter, IReadableBufferAwaiter, IF
         /// Initializes the <see cref="Pipe"/> with the specifed <see cref="IBufferPool"/>.
         /// </summary>
         /// <param name="pool"></param>
+        /// <param name="scheduler"></param>
         /// <param name="maximumSizeLow"></param>
         /// <param name="maximumSizeHigh"></param>
-        public Pipe(IBufferPool pool, long maximumSizeLow = 0, long maximumSizeHigh = 0)
+        public Pipe(IBufferPool pool, IScheduler scheduler = null, long maximumSizeLow = 0, long maximumSizeHigh = 0)
```
pakrym: Reference options object?

tests/System.IO.Pipelines.Tests/SchedulerFacts.cs
```diff
@@ -0,0 +1,85 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.IO.Pipelines.Tests
+{
+    public class SchedulerFacts
+    {
+        [Fact]
+        public async Task ReadAsyncCallbackRunsOnSchedulerThread()
```
pakrym: Check FlushAsync too.

src/System.IO.Pipelines/PipeOptions.cs
davidfowl: Nope.

tests/System.IO.Pipelines.Tests/SchedulerFacts.cs
davidfowl: What about FlushAsync?

pakrym: That it's callback is scheduled too.

davidfowl: Oh that is broken then. I need this code to only affect the consumer not the producer. The producer is using Task so you'll get TPL behavior.

src/System.IO.Pipelines/PipeOptions.cs
benaadams: readonly nullables?

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -64,36 +65,38 @@ public class Pipe : IPipelineReader, IPipelineWriter, IReadableBufferAwaiter, IF
...
+                throw new ArgumentException(nameof(options.MaximumSizeHigh) + " should be greater or equal to " + nameof(options.MaximumSizeLow), nameof(options.MaximumSizeHigh));
             }
+
             _pool = pool;
-            _scheduler = scheduler ?? InlineScheduler.Default;
-            _maximumSizeHigh = maximumSizeHigh;
-            _maximumSizeLow = maximumSizeLow;
-            _awaitableState = _awaitableIsNotCompleted;
-            _flushAwaitableState = _awaitableIsCompleted;
+            _readerScheduler = options.ReaderScheduler ?? InlineScheduler.Default;
```
pakrym: Make options immutable and keep the reference, Pipe is giant as is.

davidfowl: See the discussion above. Making it immutable means I need to have overloads for all combinations of fields.

pakrym: Or use immutable builder like `.WithBlala(12)`

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -343,15 +357,21 @@ public void AdvanceWriter(int bytesWritten)
...
         {
             if (_producingState == State.Active)
             {
                 // Commit the data as not already committed
                 Commit();
             }
 
-            return await SignalWriterAsync();
+            // TODO: Can factor out this lock
+            lock (_sync)
```
pakrym: Just remove this lock, I don't think there is anything in Resume to protect and running callback via inline scheduler would cause unnecessary overlocking.

davidfowl:
@dotnet-bot test this please

PR #1167, opened 2017-02-01
Sample: changed response writer to not use multi-segment buffers

KrzysztofCwalina:
Instead the formatter/writer sends the data once it runs out of buffer space. 
I can do it now as I switched to chunked transfer.


KrzysztofCwalina:
@dotnet-bot test this please

PR #1169, opened 2017-02-02
Re-enabling System.Text.Primitives functional tests by...

ahsonkhan:
... separating out perf tests into a separate project.


ahsonkhan:
@dotnet-bot Test Windows Release please

src/System.Text.Primitives/Properties/AssemblyInfo.cs
```diff
@@ -10,6 +10,13 @@
 [assembly: AssemblyProduct("System.Text.Primitives")]
 [assembly: AssemblyTrademark("")]
 
+[assembly: InternalsVisibleTo("System.Text.Primitives.Performance.Tests,PublicKey="+
+"0024000004800000940000000602000000240000525341310004000001000100039ac461fa5c82"+
+"c7dd2557400c4fd4e9dcdf7ac47e3d572548c04cd4673e004916610f4ea5cbf86f2b1ca1cb824f"+
+"2a7b3976afecfcf4eb72d9a899aa6786effa10c30399e6580ed848231fec48374e41b3acf88119"+
+"31343fc2f73acf72dae745adbcb7063cc4b50550618383202875223fc75401351cd89c44bf9b50"+
+"e7fa3796")]
+
```
KrzysztofCwalina: Why is this required?

tests/System.Text.Primitives.Tests/Encoding/RandomTests.cs
```diff
@@ -147,8 +147,8 @@ public void ToString(string expected, Utf8String s)
         [Theory]
         [MemberData("StringEqualsTestCases_EmptyStrings")]
         [MemberData("StringEqualsTestCases_SimpleStrings")]
-        public unsafe void StringEquals(bool expected, Utf8String s1, Utf8String s2)
-        {
+        public unsafe void StringEqualsWithThreeArgs(bool expected, Utf8String s1, Utf8String s2)
+        {   // TODO: investigate why the tests fail if we have two methods with the same name StringEquals
```
KrzysztofCwalina: What do you mean by that? Two methods with the same name on different types? Are the types tests?

shiftylogic: @KrzysztofCwalina Something changed in either XUnit or the compiler / tool-chain such that two tests with the exact same name but different arguments cause an issue.

In this case, there are 2 tests names 'StringEquals'. The bottom one (logical code order) takes 1 argument and has 2 inline test cases. For some reason, those test cases attempt to run using the 3 argument variant of the function and thus throw an exception at invocation time.

Renaming the function solves the problem, though, it might be worth filing a bug in any case. This seems like an odd behavior considering it worked before we upgraded to VS2017.

src/System.Text.Primitives/Properties/AssemblyInfo.cs
shiftylogic: Quick scan, it doesn't look like any of the performance tests you moved are calling internal code. If that is true, you can remove this.

ahsonkhan: EncodingPerfComparisonTests actually uses Utf8Encoder.TryEncode which is inaccessible due to its protection level. The UTF8Encoder class is internal.

scripts/PerfHarness/PerfHarness.cs
```diff
@@ -27,10 +27,10 @@ private static string GetTestAssembly(string testName)
     private static string[] GetTestAssemblies()
     {
         return new [] {
-            "Benchmarks",
-            "System.Binary.Base64.Tests",
-            "System.Text.Primitives.Tests",
-            "System.Slices.Tests"
+            //"Benchmarks",
```
shiftylogic: Did you intend to check in this file?

ahsonkhan: Yes, I forgot to uncomment the assemblies. Will fix.

ahsonkhan:
@dotnet-bot test please

crummel:
@ahsonkhan LGTM.

src/System.Text.Primitives/Properties/AssemblyInfo.cs
jkotas: This sounds broken to me. I think it is important to expose the highest performance static workhorse methods as public so that folks who want to just encode piece of UTF8 without the overhead of virtual calls, casts and other crap can do so.

shiftylogic: @jkotas I agree on this and I think the static implementation classes will become public at some point.

PR #1172, opened 2017-02-03
Optimized Memory<T> Access

KrzysztofCwalina:
At the standup today, we talked about how accessing Span from Memory from Memory from OwnedMemory shows in profiles. I spent some thine looking at what we do in these code paths, and there are some obvious inefficiencies.  

a) Getting Span from Memory first creates a full Span for the OwnedMemory and then slices the Span to get Span corresponding to Memory's subrange. This change eliminates this slicing.
b) Getting Memory from OwnedMemory is done by passing OwnedMemory to Memory ctor. The ctor then gets Span out of OwnedMemory to determine the length of itself. This change eliminates the need to get the Span by calling a different Memory ctor that explicitly takes length parameter.

Unfortunately I cannot compare perf wins as out perf benchmarks for pipelines (Benchmarks.dll) are currently broken. I will try to fix the benchmarks and amend this PR.

cc: @davidfowl, @pakrym, @shiftylogic, @ahsonkhan

### DIFF
--- a/src/System.Slices/System/Buffers/OwnedMemory.cs
+++ b/src/System.Slices/System/Buffers/OwnedMemory.cs
@@ -50,7 +50,9 @@ namespace System.Buffers
             Initialize(array, arrayOffset, length, pointer);
         }
 
-        public Memory<T> Memory => new Memory<T>(this, Id);
+        public Memory<T> Memory => new Memory<T>(this, Id, 0, Length);
+        public ReadOnlyMemory<T> ReadOnlyMemory => new ReadOnlyMemory<T>(this, Id, 0, Length);
+
         public Span<T> Span
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -133,8 +135,6 @@ namespace System.Buffers
 
         protected virtual void OnZeroReferences()
         { }
-
-
         #endregion
 
         #region Used by Memory<T>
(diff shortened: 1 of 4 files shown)


src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -178,16 +178,25 @@ bool IMemory<T>.TryGetArray(long id, out ArraySegment<T> buffer)
...
         {
             VerifyId(id);
-            return Span;
+            var array = Array;
+            if (array != null)
+                return new Span<T>(array, Offset + index, length);
+            else unsafe
+                {
+                    if (length > Length - index) throw new ArgumentOutOfRangeException();
+                    var offsetBytes = Unsafe.SizeOf<T>() * index;
```
jkotas: Integer overflow...

KrzysztofCwalina: Are you sure? The check above ensures that either index is smaller than Length or length is negative. if length is negative, the next line (ctor) will fail. If index is smaller than Length, then the multiplication cannot overflow or we will have much bigger problems.  

Having said that, I just stole the implementation from System.Memory.Span<T>.Slice. Though I am not sure all the checks are necessary (as mentioned above).

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -178,16 +178,47 @@ bool IMemory<T>.TryGetArray(long id, out ArraySegment<T> buffer)
...
+                    if ((uint)index > (uint)Length || (uint)length > (uint)(Length - index))
+                        ThrowArgHelper();
+                    IntPtr newPtr = Add<T>(Pointer, index);
+                    return new Span<T>(newPtr.ToPointer(), length);
+                }
+        }
+
+        // TODO: this is taken from SpanHelpers. If we move this type to System.Memory, we should remove this helper
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static IntPtr Add<TItem>(IntPtr start, int index)
```
jkotas: No need to be generic over `Item` here - the method is in generic type already.

jkotas: Also, did you mean for this method to be public?

src/System.Slices/System/Buffers/OwnedMemory.cs
jkotas: I think there wes problem with multiplication - what happens if the result is more than Int32.MaxInt. Writing correct code that operates with native ints is real pain... .

tests/System.Buffers.Experimental.Tests/BytesReader.cs
```diff
@@ -10,7 +10,7 @@ namespace System.Slices.Tests
 {
     public partial class ReadOnlyBytesTests
     {
-        [Fact(Skip = "[VS2017] Error: The active Test Run was aborted.")]
+        [Fact]
```
davidfowl: I kept seeing this, is it working now?

KrzysztofCwalina: I removed the Skip and it worked.

ahsonkhan:
> Unfortunately I cannot compare perf wins as out perf benchmarks for pipelines (Benchmarks.dll) are currently broken. I will try to fix the benchmarks and amend this PR.

Were you able to get the benchmark tests running?

PR #1179, opened 2017-02-05
Avoid unnecessary branches when fixing non-null/empty arrays

justinvp:
When we know an array is not null or empty at the point where `fixed` is used, we can reduce the number of branches (and IL size of the method body) by using the address of the first element.

Based on the discussion here: https://github.com/dotnet/corefx/pull/15432#discussion_r97684916

Related:

 - https://github.com/dotnet/coreclr/pull/9115
 - https://github.com/dotnet/coreclr/pull/9188
 - https://github.com/dotnet/corefx/pull/15597
 - https://github.com/dotnet/corert/pull/2668

### DIFF
--- a/src/System.IO.Pipelines.Networking.Libuv/Interop/SockAddr.cs
+++ b/src/System.IO.Pipelines.Networking.Libuv/Interop/SockAddr.cs
@@ -85,7 +85,7 @@ namespace System.IO.Pipelines.Networking.Libuv.Interop
             {
                 // otherwise IPv6
                 var bytes = new byte[16];
-                fixed (byte* b = bytes)
+                fixed (byte* b = &bytes[0])
                 {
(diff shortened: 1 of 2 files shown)


src/System.IO.Pipelines.Networking.Libuv/Interop/SockAddr.cs
```diff
@@ -85,7 +85,7 @@ public unsafe IPEndPoint GetIPEndPoint()
             {
                 // otherwise IPv6
                 var bytes = new byte[16];
```
benaadams: Shame the `IPAddress(byte* address` overloads are internal, else could `stackalloc` this

davidfowl:
@dotnet-bot test Windows Release please

davidfowl:
👍

PR #1182, opened 2017-02-06
Add IPipe interface, Rename IPipeline* to IPipe*

pakrym:
@davidfowl


src/System.IO.Pipelines/Pipe.cs
```diff
@@ -763,4 +749,10 @@ private static class CancelledState
             public static int CancellationObserved = 2;
         }
     }
+
+    public interface IPipe
```
davidfowl: Move this to another file

davidfowl:
@dotnet-bot test this please

PR #1194, opened 2017-02-08
Davidfowl/fix pipeline tests

davidfowl:


src/System.IO.Pipelines.Networking.Libuv/UvTcpClient.cs
```diff
@@ -53,7 +50,7 @@ private static void OnConnection(UvConnectRequest req, int status, Exception exc
 
             var connection = new UvTcpConnection(client._thread, client._connectSocket);
 
-            client._connectTcs.TrySetResult(connection);
+            Task.Run(() => client._connectTcs.TrySetResult(connection));
```
benaadams: Could go heavy on the ifdefs and for more recent frameworks do it in ctor:
```csharp
_connectTcs = new TaskCompletionSource<UvTcpConnection>(TaskCreationOptions.RunContinuationsAsynchronously);
```

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
```diff
@@ -167,30 +179,40 @@ internal static void RecycleSocketAsyncEventArgs(SocketAsyncEventArgs args)
...
         {
-            if (disposing)
+            if (!_disposed)
             {
+                _disposed = true;
                 _stopping = true;
                 _output.Reader.CancelPendingRead();
 
-                Task.WaitAll(_sendTask, _receiveTask);
+                await Task.WhenAll(_sendTask, _receiveTask);
```
benaadams: As mentioned in https://github.com/dotnet/corefxlab/pull/1131#issuecomment-275172475 you could turned it into a `WhenAll` continuation? Not sure you really care about when it finishes with all the `async void` about..?

src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
```diff
@@ -137,22 +137,19 @@ private void OnAccept(SocketAsyncEventArgs e)
             // will cause a stack-dive in the sync (backlog) case
         }
 
-        private async void ExecuteConnection(SocketConnection conn)
+        private async void ExecuteConnection(SocketConnection connection)
```
benaadams: `async void` 😱

src/System.IO.Pipelines.Networking.Libuv/UvTcpClient.cs
halter73: Any particular reason this is preferable?

davidfowl: I'll do it when we stop targeting net45.

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
halter73: I prefer await over ContinueWith. With the latter, you're much more likely to run into unobserved exceptions.

src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
davidfowl: Yah, nobody is tracking these tasks.

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
benaadams: But then you await them in an `async void`; surely an unobserved exception is preferable to a straight out app crash?

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
```diff
@@ -167,30 +179,40 @@ internal static void RecycleSocketAsyncEventArgs(SocketAsyncEventArgs args)
...
         }
 
+        /// <summary>
+        /// Releases all resources owned by the connection
+        /// </summary>
+        protected virtual void Dispose(bool disposing)
+        {
+            if (disposing)
+            {
+                GC.SuppressFinalize(this);
```
halter73: Why even implement the disposing patter if this is the only virtual method in the class? The pattern also recommends putting the call to SuppressFinalize in the non-virtual Dispose() method, presumably so it can't be overridden.

src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
halter73: Is this considered the same, better or worse than `async Task` with the caller doing `var ignore =`?

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
halter73: I'd prefer the crash. Then we'd know about it.

src/System.IO.Pipelines.Networking.Sockets/SocketListener.cs
benaadams: You can catch/report/log on `var ignore =` exceptions with the `TaskScheduler.UnobservedTaskException` handler, and in 4.5+ they don't cause an app crash. `async void` exceptions cut through the app like butter and bring it all down.

benaadams: Is why the logging threadpool wrapper in Kestrel; throwing an exception out of a QUWI does the same thing.

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
benaadams: Doesn't look like `_sendTask` or `_receiveTask` will throw as they catch almost everything, and have an empty catch for the rest; still on't tell you about it though.

`AsyncVoidMethodBuilder` captures the current `SynchronizationContext` and posts stuff back to that, starting run, completion, exception etc; whereas `AsyncTaskMethodBuilder` stuffs in the Task.

Feels like code smell though; but I not know enough about `async void` other than a having a general aversion from use; but it does interact with the sync context in a way `async Task` doesn't seem to which means it may behave differently depending on runtime environment?

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
davidfowl: Maybe I'll remove Dispose from `IPipelineConnection`, it really serves no purpose at this point.

src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs
benaadams: Alternatively, what's the point of `_sendTask` and `_receiveTask` they don't really do anything? Change the ctor to
```
_socketTask = Task.WhenAll(ReceiveFromSocketAndPushToWriterAsync(), ReadFromReaderAndWriteToSocketAsync());
```
Just store the socket task? Then that await becomes
```
await _socketTask;
```
You could even expose that task and find if the socket had errored etc

PR #1195, opened 2017-02-08
Await writes so natural back pressure is applied

davidfowl:
Not 100% sure about the perf throughput of this change yet, various a bit on my machine. It also reduces writing in parallel but that can be brought back by not awaiting the task. We can also have a maximum concurrency level if we want to restrict the amount of outstanding writes (like Kestrel does).

/cc @halter73 @pakrym @benaadams


src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -121,47 +117,25 @@ protected virtual void Dispose(bool disposing)
...
-        }
-
-        private void EndWrite(UvWriteReq writeReq)
-        {
-            _pendingWrites--;
-
-            _thread.WriteReqPool.Return(writeReq);
+            try
+            {
+                await writeReq.WriteAsync(_handle, buffer);
```
halter73: Nice. Sooo much cleaner. Do you have any benchmark results before and after?

benaadams: Agreed is much nicer

PR #1197, opened 2017-02-08
Finalizer issue with OwnedNativeMemory

mjp41:
The finalizer could fail if the object had got into a bad state due
to Reservation having been leaked. This occured in the test due
to trying to dispose an object while having reservations. This is
a possible fix, but it may be better to decide on how finalization
and Dispose should interact.

There are a few questions about the right error model:
* Should we raise an error if a Release is on an object that has been 
  attempted to be freed? Or should that succeed.  Currently, if fails, 
  but I think this might be a bad idea. 
* Should finalization of an object potentially raise an exception if it 
  has effectively been leaked?

@KrzysztofCwalina @joshfree @davidfowl

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/OwnedMemory.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/OwnedMemory.cs
@@ -23,7 +23,7 @@ namespace System.Buffers
 
         ~OwnedNativeMemory()
         {
-            Dispose();
+            Dispose(false);
         }
 
         protected override void Dispose(bool disposing)
--- a/src/System.Slices/System/Buffers/OwnedMemory.cs
+++ b/src/System.Slices/System/Buffers/OwnedMemory.cs
@@ -107,14 +107,16 @@ namespace System.Buffers
             Interlocked.Exchange(ref _id,  FreedId);
             if (HasOutstandingReferences) throw new InvalidOperationException("outstanding references detected.");
             Dispose(true);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        { 
+            _id = FreedId;
             _array = null;
             _pointer = IntPtr.Zero;
             _length = 0;
             _arrayIndex = 0;
         }
-
-        protected virtual void Dispose(bool disposing)
-        { }
 
         public bool IsDisposed => Id == FreedId;
 
--- a/tests/System.Buffers.Experimental.Tests/MemoryTests.cs
+++ b/tests/System.Buffers.Experimental.Tests/MemoryTests.cs
@@ -7,7 +7,7 @@ namespace System.Slices.Tests
 {
     public class MemoryTests
     {
-        [Fact(Skip = "flaky")]
+        [Fact]
         public void SimpleTestS()
         {            
             using(var owned = new OwnedNativeMemory(1024)) {
@@ -43,7 +43,7 @@ namespace System.Slices.Tests
             }
         }
 
-        [Fact(Skip = "flaky")]
+        [Fact]
         public void NativeMemoryLifetime()
         {
             Memory<byte> copyStoredForLater;
@@ -81,7 +81,7 @@ namespace System.Slices.Tests
             });
         }
 
-        [Fact(Skip = "flaky")]
+        [Fact]
         public unsafe void PinnedArrayMemoryLifetime()
         {
             Memory<byte> copyStoredForLater;


KrzysztofCwalina:
adding @jkotas

davidfowl:
> Should we raise an error if a Release is on an object that has been
attempted to be freed? Or should that succeed. Currently, if fails,
but I think this might be a bad idea.

What do you mean by "has been attempted to be freed". Do you mean a call to `Dispose()` then `Release()`?

> Should finalization of an object potentially raise an exception if it
has effectively been leaked?

I don't think finalization is supposed to throw. It's a last ditch effort to clean up things that may not have been cleaned up with proper disposal.

mjp41:
>>Should we raise an error if a Release is on an object that has been
>>attempted to be freed? Or should that succeed. Currently, if fails,
>>but I think this might be a bad idea.

>What do you mean by "has been attempted to be freed". Do you mean a call to  Dispose()  then  Release() ?

Yeah, so 
```csharp
try {
using(var r = o.DisposalableReservation())
   o.Dispose();   // This currently throws
} // The implicit o.Release(); also throws.
```
If the `Release` succeeded, then the finalizer would succeed given the previous code. I think Release should just succeed, and not check it is on a valid owner? That is 
```csharp
        void IKnown.Release(long id)
        {
            VerifyId(id);
            Release();
        }
```
should become
```csharp
        void IKnown.Release(long id)
        {
           // VerifyId(id);
            Release();
        }
```

mjp41:
We have potential security holes if an OwnerNative becomes unreachable, and is then resurrected.

KrzysztofCwalina:
Wouldn't it be bad if anybody could release OwnedMemory? A copy of reservation token could release somebody else's OwnedMemory (that was returned and re-leased from the pool).

mjp41:
If reservation tokens can be copied, then we are unsafe, with or without that check. This is part of the design that worries me.

KrzysztofCwalina:
We don't have a mechanism to prevent the tokens from being copied. Unless we make the token to be a class.

But, I think it's less bad for a single logical opetation to screw up tokens than for two independent logical operations being able to release their memory.

Drawaes:
Zero the memory on dispose/release... then doesn't really matter if someone bring back memory there won't be anything of use in there.

KrzysztofCwalina:
oops, accidentally closed.

KrzysztofCwalina:
The memory is pooled. We don't want "unauthorized" threads to bring back somebody's memory. Clearing won't help as the memory might be re-rented from a pool and filled with data again.

mjp41:
If the OwnedMemory object is not reused, then it I think it would be safe to do zeroing in the right places, but if we pool OwnedMemory instances as the design doc suggests. Then that would not be sufficient.

Drawaes:
I agree, however from a security standpoint, I would say without a wipe there is a chance another thread picks up some "old" presumed gone data and then say stuffs it onto a socket.

davidfowl:
Whatever we do the finalizer can't throw.

KrzysztofCwalina:
I agree. We should not throw from finalizers. We might want to add trace events instead.

src/System.Buffers.Experimental/System/Buffers/OwnedMemory.cs
```diff
@@ -23,7 +23,7 @@ public OwnedNativeMemory(int length) : this(length, Marshal.AllocHGlobal(length)
 
         ~OwnedNativeMemory()
         {
-            Dispose();
+            Dispose(false);
```
KrzysztofCwalina: This means we would free the memory only in Dispose, but not in Finalize?

src/System.Slices/System/Buffers/OwnedMemory.cs
```diff
@@ -107,15 +107,16 @@ public void Dispose()
             Interlocked.Exchange(ref _id,  FreedId);
             if (HasOutstandingReferences) throw new InvalidOperationException("outstanding references detected.");
             Dispose(true);
+        }
+
+        protected virtual void Dispose(bool disposing)
+        { 
             _array = null;
```
KrzysztofCwalina: Should the _id be "freed" too?

mjp41:
I have removed the change to Release.  This is just the change to stop the finalizer throwing for now.

KrzysztofCwalina:
@vancem, would you recommend tracing from the finalizer to communicate errors in lifetime management of OwnedMemory?

KrzysztofCwalina:
Should the memory tests be enabled back?

mjp41:
@KrzysztofCwalina it looks like the tests were disable because I had put some performance tests in there?

```ps
        $file -match "System.Slices.Tests") {
        Write-Warning "Skipping tests in $file. These tests use the xunit.performance package which needs to be updated for dotnet SDK rc4."
        continue;
```

Should I factor the tests differently?

KrzysztofCwalina:
Sorry for not being clearer. I was talking about the tests we had to disabled because the filnalizer issue were failing the build: https://github.com/dotnet/corefxlab/blob/master/tests/System.Buffers.Experimental.Tests/MemoryTests.cs

mjp41:
Added back. Was the test always failing in CI?

KrzysztofCwalina:
@dotnet-build-bot test this please

KrzysztofCwalina:
I will run the tests one more time before merging. 
Otherwise, looks good.

PR #1201, opened 2017-02-08
Update common.props

davidfowl:
Actually strong name the assemblies


KrzysztofCwalina:
looks good

halter73:
I did a search for "SignAssembly" in the corefx repo and noticed there were no matching results in the code. I found https://github.com/dotnet/corefx/pull/2194 which seems to imply true is the default in that repo. Anyone know why? Is there an equivalent to the KoreBuild repo that sets this to true?

PR #1202, opened 2017-02-09
Improvements to ReadOnlyBytes

KrzysztofCwalina:
cc: @shiftylogic, @ahsonkhan, @davidfowl 

Added more through tests for slicing
Found a bug and fixed it
Optimized slicing to empty ROBs

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -107,10 +107,16 @@ namespace System.Buffers
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public ReadOnlyBytes Slice(int index, int length)
         {
+            if(length==0) return ReadOnlyBytes.Empty;
+
             var first = First;
             if (first.Length >= length + index)
             {
-                return new ReadOnlyBytes(first.Slice(index, length));
+                var slice = first.Slice(index, length);
+                if(slice.Length > 0) {
+                    return new ReadOnlyBytes(first.Slice(index, length));
+                }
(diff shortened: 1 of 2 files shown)


src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -129,9 +135,20 @@ public ReadOnlyBytes Slice(int index)
             else {
                 var first = First;
                 if (first.Length > index) {
-                    return new ReadOnlyBytes(first.Slice(index), _rest);
+                    var slice = first.Slice(index);
+                    if(slice.Length > 0 || _rest!=null) {
+                        return new ReadOnlyBytes(slice, _rest);
```
ahsonkhan: _rest can be null?

KrzysztofCwalina: Yes. If the multi-segment structure contains just one segment.

ahsonkhan: I see. To confirm, this should be an 'or' (as in we could end up with ReadOnlyBytes (0, _rest) or ReadOnlyBytes(slice, null)? Both are valid?
```C#
if(slice.Length > 0 || _rest!=null) { 
```

KrzysztofCwalina: Yes, it should be an OR. If slice.length is zero but _rest is not null, we don't want to return an empty ROB. Same if slice is longer than zero.

davidfowl:
The formatting is all messed up 😄

PR #1204, opened 2017-02-09
Don't use tasks to track completion of the pipe

davidfowl:
/cc @pakrym @halter73 @benaadams

### DIFF
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -706,7 +706,7 @@ namespace System.IO.Pipelines
 
             public void ThrowIfFailed()
             {
-                if (_exception != _completedNoException)
+                if (_exception != null && _exception != _completedNoException)
                 {
                     throw _exception;
                 }
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -692,22 +692,21 @@ namespace System.IO.Pipelines
 
         private struct Completion
         {
-            private int _completed;
+            private static readonly Exception _completedNoException = new Exception();
+
             private Exception _exception;
 
-            public bool IsCompleted => _completed == 1;
+            public bool IsCompleted => _exception != null;
 
             public void TryComplete(Exception exception = null)
             {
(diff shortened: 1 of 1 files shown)


benaadams:
Made sense before; but this makes even more sense. 

Ignoring the switching of reader and writer... 😉

pakrym:
I like it. But we can decrease size of pipe object by storing only exception and having marker exception for completed state.

benaadams:
`NoException` 😉

davidfowl:
I considered that but I didn't want to go that route yet because the size of pipe isn't really a problem as we aren't making one per request.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -645,23 +625,8 @@ private static void Reset(ref Action awaitableState)
...
-            else
-            {
-                taskCompletionSource.TrySetResult(null);
-            }
-        }
-
         private void Dispose()
         {
-            Debug.Assert(Writing.IsCompleted, "Not completed writing");
-            Debug.Assert(Reading.IsCompleted, "Not completed reading");
```
halter73: Why did you remove these asserts?

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -713,18 +678,41 @@ bool IWritableBufferAwaiter.GetResult()
...
+
+            public void TryComplete(Exception exception = null)
+            {
+                // Set the exception object to the exception passed in or a sentinel value
+                Interlocked.CompareExchange(ref _exception, exception ?? _completedNoException, null);
+            }
+
+            public void ThrowIfFailed()
+            {
+                if (_exception != null && _exception != _completedNoException)
```
halter73: Why not just initialize _exception to be _completedNoException?

halter73: Nvm. CompareExchange duh.

src/System.IO.Pipelines/Pipe.cs
davidfowl: laziness

halter73: Were tests failing?

PR #1207, opened 2017-02-09
Extract awaitable support into a struct

pakrym:
@davidfowl


halter73:
:shipit:

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -89,12 +84,11 @@ public Pipe(IBufferPool pool, PipeOptions options = null)
...
 
             _pool = pool;
-            _readerScheduler = options.ReaderScheduler ?? InlineScheduler.Default;
-            _writerScheduler = options.WriterScheduler ?? InlineScheduler.Default;
             _maximumSizeHigh = options.MaximumSizeHigh;
             _maximumSizeLow = options.MaximumSizeLow;
-            _readerCallback = _awaitableIsNotCompleted;
-            _writerCallback = _awaitableIsCompleted;
+
+            _readerAwaitable = new Awaitable(false, options.ReaderScheduler ?? InlineScheduler.Default);
```
davidfowl: Use optional parameter syntax and put the bool last. Makes it more obvious what it is.

src/System.IO.Pipelines/Pipe.Awaitable.cs
```diff
@@ -0,0 +1,108 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    internal partial class Pipe
+    {
+        private struct Awaitable
```
halter73: Nit: Make PipeAwaitable a non-nested internal struct instead of using partial classes.

src/System.IO.Pipelines/Pipe.OperationState.cs
```diff
@@ -0,0 +1,64 @@
...
+#endif
+                }
+            }
+
+            private static class State
+            {
+                public static int NotActive = 0;
+                public static int Active = 1;
+            }
+
```
halter73: nit

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -448,8 +395,6 @@ void IPipeReader.Advance(ReadCursor consumed, ReadCursor examined)
                 _length -= consumedBytes;
                 resumeWriter = _length < _maximumSizeLow;
 
-                // Change the state from observed -> not cancelled. We only want to reset the cancelled state if it was observed
-                Interlocked.CompareExchange(ref _cancelledState, CancelledState.NotCancelled, CancelledState.CancellationObserved);
```
davidfowl: Where does this happen?

src/System.IO.Pipelines/System.IO.Pipelines.csproj
```diff
@@ -5,6 +5,7 @@
     <TargetFrameworks>net451;netstandard1.3</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageIconUrl>http://go.microsoft.com/fwlink/?linkid=833199</PackageIconUrl>
+    <DefineConstants Condition="'$(Configuration)' == 'Debug'">OPERATION_LOCATION_TRACKING;COMPLETION_LOCATION_TRACKING</DefineConstants>
```
davidfowl: You need to keep the default `<DefineConstants>`.

```C#
+    <DefineConstants Condition="'$(Configuration)' == 'Debug'">$(DefineConstants);OPERATION_LOCATION_TRACKING;COMPLETION_LOCATION_TRACKING</DefineConstants>
```

src/System.IO.Pipelines/Pipe.cs
pakrym: PipeAwaitable.Reset

src/System.IO.Pipelines/PipeAwaitable.cs
```diff
@@ -0,0 +1,105 @@
...
+        }
+
+        public void Reset()
+        {
+            // Change the state from observed -> not cancelled. We only want to reset the cancelled state if it was observed
+            Interlocked.CompareExchange(ref _cancelledState,
+                CancelledState.NotCancelled,
+                CancelledState.CancellationObserved);
+
+            if (_cancelledState != CancelledState.CancellationRequested)
```
benaadams: Take return value from the `CompareExchange` and use that instead?

benaadams: Take return value from the `CompareExchange` and use that instead? Rather than re-reading the value you've just cache invalidated

pakrym: It's done in other PR: https://github.com/dotnet/corefxlab/pull/1209

benaadams: Concurrent PRs on concurrency, nice touch 😉

halter73: I don't think this PR is meant to affect any of the logic. It's about code cleanliness.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -10,20 +10,20 @@ namespace System.IO.Pipelines
     /// <summary>
     /// Default <see cref="IPipeWriter"/> and <see cref="IPipeReader"/> implementation.
     /// </summary>
-    internal class Pipe : IPipe, IPipeReader, IPipeWriter, IReadableBufferAwaiter, IWritableBufferAwaiter
+    internal partial class Pipe : IPipe, IPipeReader, IPipeWriter, IReadableBufferAwaiter, IWritableBufferAwaiter
```
davidfowl: Remove the partial?

PR #1209, opened 2017-02-10
Unlock some parts of pipes

pakrym:
Locking only around commit head and index.

@davidfowl @halter73 thoughts?

### DIFF
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -258,16 +258,17 @@ namespace System.IO.Pipelines
                 // Always move the commit head to the write head
                 _commitHead = _writingHead;
                 _commitHeadIndex = _writingHead.End;
+            }
 
-                _length += _currentWriteLength;
+            var currentLenght = Interlocked.Add(ref _length, _currentWriteLength);
+
                 // Do not reset if reader is complete
                 if (_maximumSizeHigh > 0 &&
-                    _length >= _maximumSizeHigh &&
+                currentLenght >= _maximumSizeHigh &&
                     !_readerCompletion.IsCompleted)
                 {
                     _writerAwaitable.Reset();
                 }
-            }
 
             // Clear the writing state
             _writingHead = null;
@@ -354,9 +355,6 @@ namespace System.IO.Pipelines
                 ThrowHelper.ThrowInvalidOperationException(ExceptionResource.CompleteWriterActiveProducer, _producingState.Location);
             }
 
-            // TODO: Review this lock?
-            lock (_sync)
-            {
                 _writerCompletion.TryComplete(exception);
 
                 _readerAwaitable.Resume();
@@ -366,7 +364,6 @@ namespace System.IO.Pipelines
                     Dispose();
                 }
             }
-        }
 
         // Reading
 
@@ -386,13 +383,14 @@ namespace System.IO.Pipelines
                 _readHead.Start = consumed.Index;
             }
 
-            bool resumeWriter;
+            var currentLength = Interlocked.Add(ref _length, -consumedBytes);
+            // Change the state from observed -> not cancelled. We only want to reset the cancelled state if it was observed
+            Interlocked.CompareExchange(ref _cancelledState, CancelledState.NotCancelled, CancelledState.CancellationObserved);
 
             // Reading commit head shared with writer
+            bool consumedEverything;
             lock (_sync)
             {
-                _length -= consumedBytes;
-                resumeWriter = _length < _maximumSizeLow;
 
 
                 var consumedEverything = examined.Segment == _commitHead &&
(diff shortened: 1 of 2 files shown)


src/System.IO.Pipelines/Pipe.cs
```diff
@@ -293,15 +293,16 @@ internal void Commit()
...
 
-                _length += _currentWriteLength;
-                // Do not reset if reader is complete
-                if (_maximumSizeHigh > 0 &&
-                    _length >= _maximumSizeHigh &&
-                    !_readerCompletion.IsCompleted)
-                {
-                    Reset(ref _writerCallback);
-                }
+            var currentLenght = Interlocked.Add(ref _length, _currentWriteLength);
```
halter73: Considering that _length is always updated immediately after or immediately prior to acquiring the _sync lock, is introducing a new memory barrier worth it? Is this measurably better than what we had before?

halter73: Also, channeling my inner @cesarbs, nit: currentLength.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -440,28 +437,26 @@ void IPipeReader.Advance(ReadCursor consumed, ReadCursor examined)
...
-                if (consumedEverything && _cancelledState != CancelledState.CancellationRequested)
-                {
-                    Reset(ref _readerCallback);
-                }
+            // We reset the awaitable to not completed if
+            // 1. We've consumed everything the producer produced so far
+            // 2. Cancellation wasn't requested
+            if (consumedEverything && cancelledState != CancelledState.CancellationRequested)
+            {
+                Reset(ref _readerCallback);
```
halter73: There's still a chance that CancelPendingRead() was called between the CompareExchange() and the call to Reset. This would cause the subsequent call to ReadAsync() to stall even though in the Pipe is in the CancellationObserved state. I think it could be changed to the following:

```c#
// We reset the awaitable to not completed if we've consumed everything the producer produced so far
if (consumedEverything)
{
    Reset(ref _readerCallback);
}

// Change the state from observed -> not cancelled. We only want to reset the cancelled state if it was observed
var cancelledState = Interlocked.CompareExchange(ref _cancelledState, CancelledState.NotCancelled, CancelledState.CancellationObserved);

if (consumedEverything && cancelledState == CancelledState.CancellationRequested)
{
    // Rollback the awaitable reset if a cancellation was requested
    Resume(_readerScheduler, ref _readerCallback);
}
```

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -258,16 +258,17 @@ internal void Commit()
                 // Always move the commit head to the write head
                 _commitHead = _writingHead;
                 _commitHeadIndex = _writingHead.End;
+            }
+
+            var currentLenght = Interlocked.Add(ref _length, _currentWriteLength);
```
davidfowl: spelling and formatting.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -258,15 +259,16 @@ internal void Commit()
...
 
-                _length += _currentWriteLength;
-                // Do not reset if reader is complete
-                if (_maximumSizeHigh > 0 &&
-                    _length >= _maximumSizeHigh &&
-                    !_readerCompletion.IsCompleted)
-                {
-                    _writerAwaitable.Reset();
-                }
+            var currentLenght = Interlocked.Add(ref _length, _currentWriteLength);
```
davidfowl: spelling - `currentLength`

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -386,26 +384,23 @@ void IPipeReader.Advance(ReadCursor consumed, ReadCursor examined)
...
-                    _readerAwaitable.Reset();
-                }
+                currentLength = (_length -= consumedBytes);
+                consumedEverything = examined.Segment == _commitHead &&
+                                     examined.Index == _commitHeadIndex &&
+                                     !_writerCompletion.IsCompleted;
+            }
+            // We reset the awaitable to not completed if
+            // 1. We've consumed everything the producer produced so far
+            // 2. Cancellation wasn't requested
```
davidfowl: Move this comment inside of `PipeAwaitable`

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -493,29 +478,24 @@ ReadableBufferAwaitable IPipeReader.ReadAsync()
 
         private void Dispose()
         {
-            // TODO: Review throw if not completed?
```
davidfowl: Since we're removing the lock here should we throw on other operations if disposed is true? That'll at least make sure that nobody is misusing the thing.

halter73: For "Dispose()" to be called both the reader and writer would need to be completed. Both reading and writing operations already throw if called after their respective sides complete.

davidfowl: kewl

davidfowl:
We need some micro benchmarks for pipe so we can see if this makes things better or worse.

PR #1210, opened 2017-02-10
Updating the Microsoft.NET.Test.Sdk to a newer version

ahsonkhan:
- This resolves the VS issue and now all tests run in VS.

You should now be able to click Run All Tests in the Test Explorer and they will run to completion rather than aborting mid way (especially for System.Buffers.Experimental.Tests).

### DIFF
--- a/tests/Benchmarks/Benchmarks.csproj
+++ b/tests/Benchmarks/Benchmarks.csproj
@@ -19,7 +19,7 @@
     <GenerateAssemblyProductAttribute>false</GenerateAssemblyProductAttribute>
   </PropertyGroup>
   <ItemGroup>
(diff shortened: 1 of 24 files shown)


KrzysztofCwalina:
Thanks. Looks good.

ahsonkhan:
Please let me know if you get "Test aborted" errors after this change. I got the "fix" but the root cause was unknown so if it occurs again, I will chase the VS Unit Test team again (after the VS 2017 release).

ahsonkhan:
> Please let me know if you get "Test aborted" errors after this change. I got the "fix" but the root cause was unknown so if it occurs again, I will chase the VS Unit Test team again (after the VS 2017 release).

cc @shiftylogic

ahsonkhan:
Partially fixes #1155

PR #1213, opened 2017-02-11
Porting performance tests to xunit perf

ahsonkhan:
Moving Utf8StringPerformanceTests out of System.Text.Primitives.Tests and into System.Text.Primitives.Performance.Tests since these are performance tests.

This should reduce our CI (and build/test) time by a few minutes, but cause running performance tests to be even longer. Hopefully, we can get the xunit.performance update where we can run specific performance tests.

**Note:** I made changes in PerformanceTests.cs so that the diff is easier to see. This file will be deleted in the next commit and we will only have the Utf8StringPerformanceTests.cs file (these have duplicate copies of the code).


ahsonkhan:
Diff visible here:
https://github.com/dotnet/corefxlab/pull/1213/commits/73b1f43e231ba4e0d99458c4684990f31382cc60#diff-c9b483d639c86eac8468c47d46797180

ahsonkhan:
cc @KrzysztofCwalina

ahsonkhan:
@dotnet-bot test this please.

PR #1217, opened 2017-02-14
Fix ExecuteCopyToAsync to not return void

### LINKED ISSUE #1216: System.IO.Pipelines Unhandled exception on threadpool
When writing an adapter for Kestrel (not the pipeline version of kestrel) it seems that when calling

ReadAsync() on a connection that is then closed with no more data I get an exception that I cannot catch that brings down the entire Kestrel process

kestrel crash

The code is at

[Sample Server](https://github.com/Drawaes/Leto/tree/master/samples/SampleASPNetServer)
[Code that blows](https://github.com/Drawaes/Leto/blob/master/src/Leto.Tls13/SecurePipelineConnection.cs#L155)

This pipeline is actually the stream coming from Kestrel that should be encrypted and then output

Drawaes:
should fix
#1216

### DIFF
--- a/src/System.IO.Pipelines/PipeFactory.cs
+++ b/src/System.IO.Pipelines/PipeFactory.cs
@@ -49,9 +49,9 @@ namespace System.IO.Pipelines
             return pipe;
         }
 
-        private async void ExecuteCopyToAsync(Pipe pipe, Stream stream)
+        private Task ExecuteCopyToAsync(Pipe pipe, Stream stream)
         {
-            await stream.CopyToAsync(pipe);
+            return stream.CopyToAsync(pipe);
         }
 
         public IPipeConnection CreateConnection(NetworkStream stream)
--- a/src/System.IO.Pipelines/PipeFactory.cs
+++ b/src/System.IO.Pipelines/PipeFactory.cs
@@ -67,24 +67,6 @@ namespace System.IO.Pipelines
             return pipe;
         }
 
-        public IPipeWriter CreateWriter(IPipeWriter writer, Func<IPipeReader, IPipeWriter, Task> consume)
-        {
-            var pipe = new Pipe(_pool);
-
-            consume(pipe, writer);
-
-            return pipe;
-        }
-
-        public IPipeReader CreateReader(IPipeReader reader, Func<IPipeReader, IPipeWriter, Task> produce)
-        {
-            var pipe = new Pipe(_pool);
-
-            produce(reader, pipe);
-
-            return pipe;
-        }
-
         public void Dispose() => _pool.Dispose();
     }
 }
--- a/src/System.IO.Pipelines/PipeFactory.cs
+++ b/src/System.IO.Pipelines/PipeFactory.cs
@@ -62,20 +62,7 @@ namespace System.IO.Pipelines
             }
 
             var pipe = new Pipe(_pool);
-
-            pipe.CopyToAsync(stream).ContinueWith((task, state) =>
-            {
-                var innerPipe = (Pipe)state;
-                if (task.IsFaulted)
-                {
-                    innerPipe.Reader.Complete(task.Exception.InnerException);
-                }
-                else
-                {
-                    innerPipe.Reader.Complete();
-                }
-            },
-            pipe, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
+            var ignore = pipe.CopyToEndAsync(stream);
 
             return pipe;
         }
--- a/src/System.IO.Pipelines/PipeFactory.cs
+++ b/src/System.IO.Pipelines/PipeFactory.cs
@@ -55,7 +55,7 @@ namespace System.IO.Pipelines
             {
                 await stream.CopyToAsync(pipe);
             }
-            catch (Exception ex)
+            catch(Exception ex)
             {
                 pipe.Writer.Complete(ex);
                 return;
@@ -98,9 +98,7 @@ namespace System.IO.Pipelines
         {
             var pipe = new Pipe(_pool);
 
-            consume(pipe, writer).ContinueWith(t =>
-            {
-            });
+            consume(pipe, writer);
 
             return pipe;
         }
@@ -109,9 +107,7 @@ namespace System.IO.Pipelines
         {
             var pipe = new Pipe(_pool);
 
-            produce(reader, pipe).ContinueWith(t =>
-            {
-            });
+            produce(reader, pipe);
 
             return pipe;
         }
--- a/src/System.IO.Pipelines/PipeFactory.cs
+++ b/src/System.IO.Pipelines/PipeFactory.cs
@@ -45,13 +45,22 @@ namespace System.IO.Pipelines
             }
 
             var pipe = new Pipe(_pool);
-            ExecuteCopyToAsync(pipe, stream);
+            var ignore = ExecuteCopyToAsync(pipe, stream);
             return pipe;
         }
 
-        private Task ExecuteCopyToAsync(Pipe pipe, Stream stream)
+        private async Task ExecuteCopyToAsync(Pipe pipe, Stream stream)
         {
-            return stream.CopyToAsync(pipe);
+            try
+            {
+                await stream.CopyToAsync(pipe);
+            }
+            catch (Exception ex)
+            {
+                pipe.Writer.Complete(ex);
+                return;
+            }
+            pipe.Writer.Complete();
         }
 
         public IPipeConnection CreateConnection(NetworkStream stream)
@@ -99,7 +108,11 @@ namespace System.IO.Pipelines
         public IPipeReader CreateReader(IPipeReader reader, Func<IPipeReader, IPipeWriter, Task> produce)
         {
             var pipe = new Pipe(_pool);
-            produce(reader, pipe);
+
+            produce(reader, pipe).ContinueWith(t =>
+            {
+            });
+
             return pipe;
         }
 
--- a/src/System.IO.Pipelines/PipeFactory.cs
+++ b/src/System.IO.Pipelines/PipeFactory.cs
@@ -45,22 +45,8 @@ namespace System.IO.Pipelines
             }
 
             var pipe = new Pipe(_pool);
-            var ignore = ExecuteCopyToAsync(pipe, stream);
+            var ignore = stream.CopyToEndAsync(pipe);
             return pipe;
-        }
-
-        private async Task ExecuteCopyToAsync(Pipe pipe, Stream stream)
-        {
-            try
-            {
-                await stream.CopyToAsync(pipe);
-            }
-            catch (Exception ex)
-            {
-                pipe.Writer.Complete(ex);
-                return;
-            }
-            pipe.Writer.Complete();
         }
 
         public IPipeConnection CreateConnection(NetworkStream stream)
--- a/src/System.IO.Pipelines/PipeFactory.cs
+++ b/src/System.IO.Pipelines/PipeFactory.cs
@@ -55,7 +55,7 @@ namespace System.IO.Pipelines
             {
                 await stream.CopyToAsync(pipe);
             }
-            catch(Exception ex)
+            catch (Exception ex)
             {
                 pipe.Writer.Complete(ex);
                 return;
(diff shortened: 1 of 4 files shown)


src/System.IO.Pipelines/PipeFactory.cs
```diff
@@ -49,9 +49,9 @@ public IPipeReader CreateReader(Stream stream)
             return pipe;
         }
 
-        private async void ExecuteCopyToAsync(Pipe pipe, Stream stream)
+        private Task ExecuteCopyToAsync(Pipe pipe, Stream stream)
         {
-            await stream.CopyToAsync(pipe);
```
halter73: This now results in a CS4014 in CreateReader. This could cause a problem once we reenable warnings as errors which I'm sure will happen at some point. You can silence this with a pragma or `var ignore = ...`.

Aditionally, shouldn't there be a try/catch that sets an error on the pipe using IPipeWriter.Complete(ex) if CopyToAsync(pipe) throws? This would also prevent unobserved task execptions. Arguably, you wouldn't need to change this method to be Task returning if you did this, but I don't mind keeping it as async Task just in case Complete(ex) throws or something which it shouldn't

benaadams: > Aditionally, shouldn't there be a try/catch that sets an error on the pipe using IPipeWriter.Complete(ex) if CopyToAsync(pipe) throws? 

Should follow the `CreateWriter` pattern?

benaadams: Else the reader will get stuck forevers

halter73: Exactly. I don't love ContinueWith though since it doesn't handle exceptions the stream might throw synchronously.

Drawaes:
Updated to the same pattern as the write.

src/System.IO.Pipelines/PipeFactory.cs
```diff
@@ -45,13 +45,27 @@ public IPipeReader CreateReader(Stream stream)
             }
 
             var pipe = new Pipe(_pool);
-            ExecuteCopyToAsync(pipe, stream);
+
+            ExecuteCopyToAsync(pipe, stream).ContinueWith((task, state) =>
+            {
+                var innerpipe = (Pipe)state;
+                if(task.IsFaulted)
```
davidfowl: nit: fix formatting (space between if and `(`)

src/System.IO.Pipelines/PipeFactory.cs
davidfowl: It can't throw synchronously here though so not an issue. It's inside of an async method.

halter73: @davidfowl That's just not true. [The Stream.CopyToAsync(IPipeWriter) extension method being called](https://github.com/dotnet/corefxlab/blob/565a10a6c5be31528ce94b2c85417a0c2d74cf02/src/System.IO.Pipelines/StreamExtensions.cs#L83) can absolutely throw synchronously since it's simply calling through to virtual Stream.CopyToAsync() method with a stream wrapper for the pipeline.

The Stream.CopyToAsync(IPipeWriter) extension method is not an "async" method, and neither is ExecuteCopyToAsync() any longer. There's no guarantee that the Stream.CopyToAsync() method isn't overridden to throw synchronously in some scenarios.

Try/catch not only reads easier, it also better avoids confusion because it's easy to incorrectly assume that a method cannot throw synchronously when it really can.

src/System.IO.Pipelines/PipeFactory.cs
```diff
@@ -99,7 +113,11 @@ public IPipeWriter CreateWriter(IPipeWriter writer, Func<IPipeReader, IPipeWrite
         public IPipeReader CreateReader(IPipeReader reader, Func<IPipeReader, IPipeWriter, Task> produce)
         {
             var pipe = new Pipe(_pool);
-            produce(reader, pipe);
+
+            produce(reader, pipe).ContinueWith(t =>
+            {
+            });
```
halter73: This isn't observing anything. Not only in the sense it is ignored, but also because the Task.Exception property is never accessed. This will still raise an UnobservedTaskException just like before.

Once again, I recommend calling this in an async method and using try/catch. Use `var ignore =` or a pragma to disable warnings if need be. Even async void is defensible if completely wrapped by a try/catch.

src/System.IO.Pipelines/PipeFactory.cs
davidfowl: This method should just be converted back into an async method (I didn't realize we changed that).

halter73:
Nit: removing 	6027527 from this PR would make it a little easier to review.

src/System.IO.Pipelines/PipeFactory.cs
davidfowl: I also agree that moving it into an async method with try catch is easier to read and we should make that change here. Move it all into the async method and do the same for reads and writes.

halter73: ContinueWith() can also be confusing because that in and of itself doesn't observe the task. The Task.Exception property needs to be accessed. Look below for an example of this confusion.

ContinueWith() can die in a 🔥 try/catch is clearly superior. We have multiple examples in one PR already of ContinueWith() being done wrong.

Drawaes:
Fixed the random commit that snuck in that should have been a rebase... Changed to async, added try catch. Changed the write side a bit as well. See what you think

src/System.IO.Pipelines/PipeFactory.cs
benaadams: Or follow the pattern in `AsPipelineWriter` that's been removed where `task.IsFaulted` is checked; but try/catch is as per the other `CreateReader` probably easier to reason about

Drawaes:
Probably can remove the empty continuation now because the try catch will get any exceptions?

src/System.IO.Pipelines/PipeFactory.cs
```diff
@@ -45,13 +45,22 @@ public IPipeReader CreateReader(Stream stream)
...
+            try
+            {
+                await stream.CopyToAsync(pipe);
+            }
+            catch (Exception ex)
+            {
+                pipe.Writer.Complete(ex);
+                return;
+            }
+            pipe.Writer.Complete();
```
halter73: After further thought, maybe the pipe.Writer.Complete(ex/null) calls should be moved to StreamExtensions.CopyToAsync() itself.

@davidfowl Do you agree? If so, I don't think StreamExtensions.CopyToAsync() should ever throw since you should observe any exceptions by reading from the pipe.

src/System.IO.Pipelines/PipeFactory.cs
```diff
@@ -99,7 +108,11 @@ public IPipeWriter CreateWriter(IPipeWriter writer, Func<IPipeReader, IPipeWrite
         public IPipeReader CreateReader(IPipeReader reader, Func<IPipeReader, IPipeWriter, Task> produce)
```
halter73: Sorry, because this isn't really related to the PR at all, but @davidfowl why would you ever need this API in a world where you already have PipeFactory.Create()?

src/System.IO.Pipelines/PipeFactory.cs
halter73: @Drawaes. I just saw that you did change StreamExtensions.CopyToAsync() to call Complete() and swallow any exceptions. In that case both the try/catch and the calls to Complete() in this method seem redundant.

halter73:
@Drawaes The empty continuations were never doing anything to begin with. The CLR still treats the exceptions as unobserved if you don't access the Task.Exception property.

src/System.IO.Pipelines/StreamExtensions.cs
```diff
@@ -78,9 +64,18 @@ public static IPipeReader AsPipelineReader(this Stream stream, CancellationToken
...
+            try
+            {
+                await stream.CopyToAsync(new PipelineWriterStream(writer));
+            }
+            catch (Exception ex)
+            {
+                writer.Complete(ex);
+                return;
+            }
+            writer.Complete();
```
Drawaes: I am not sure about that @halter73 in fact maybe it should be removed. If there is no exception is the expected behavior that the pipe is completed. I think that would be more likely to be named CopyToEndAsync or similar.

halter73: Interesting point. I think we should definitely complete the writer in the case of an error, but I can see the argument for leaving the pipe open after a successful copy by default. What I don't like is calling Complete() both here and ExecuteCopyToAsync(). Pick one.

Drawaes: My problem then is the Execute is more you want it to finish the pipe because you are actively wrapping the stream but for the copy to you don't want to. In both cases you want to complete the pipeline if there is an exception. I think the best idea is to just return the task from the copytoasync and catch and close in the execute.

Drawaes:
So my thought here is that if you are calling CopyToAsync you will be handling errors/completion yourself you don't really expect the pipe to be taken control of. Say if you open a filestream for whatever reason and you want to copy the file to the stream, and it fails because of the stream you probably don't want the stream to kill the pipe. If on the otherhand you are wrapping the pipe into a stream then it should take control of it.

If you prefer it the other way I am happy to flip it... your choice

halter73:
@Drawaes I think I agree with you now, but it would be nice to have some built-in method that converts a stream to a pipe while handling completion and errors. I think that's something a lot of people will want to do without having to deal with all the try/catch/complete ceremony themselves. It probably shouldn't be called CopyToAsync though, since the stream API sets up bad expectations in that regard.

I'm not sure what we should call this new method though. Maybe the name should include "Convert".

Drawaes:
@halter73 I have changed it to have a CopyToEndAsync() which is kind of a pair to the ReadToEndAsync on the pipe. It has the complete(ex/null) behavior and therefore we don't need the private ExecuteCopy anymore. See what you think

src/System.IO.Pipelines/StreamExtensions.cs
```diff
@@ -29,21 +29,7 @@ public static IPipeWriter AsPipelineWriter(this Stream stream)
...
-                    innerPipe.Reader.Complete(task.Exception.InnerException);
-                }
-                else
-                {
-                    innerPipe.Reader.Complete();
-                }
-            }, 
-            pipe, CancellationToken.None, TaskContinuationOptions.ExecuteSynchronously, TaskScheduler.Default);
-
+            pipe.CopyToAsync(stream);
```
halter73: This CopyToAsync() could throw, right? For example if Stream.WriteAsync() throws, I think we should observe it. I also think it makes sense to still complete the reader since the intention was to read the entire pipe anyway.

src/System.IO.Pipelines/PipeFactory.cs
davidfowl: Yes. The idea is to not expose the pipe but now that we have `IPipe` it's conceivable that most of these methods just go away.

src/System.IO.Pipelines/StreamExtensions.cs
Drawaes: Hahaha, yeah sorry I wasn't really thinking about the Pipe -> Stream as it wasn't my original problem but you are right the whole thing needs to be consistent and those other API's could be removed. I was trying for a surgical strike rather than open heart surgery... give me a few minutes to fix it up

Drawaes:
Okay, both sides are the same now. I have done the create reader/writer delete as another commit to remove them and updated the compression library to change to the create. If you don't want it in here that is fine I can remove the commit, but I figured it's in the same area and a quick change.

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -67,6 +67,25 @@ public static ValueTask<int> ReadAsync(this IPipeReader input, Span<byte> destin
...
+            try
+            {
+                await input.CopyToAsync(stream, bufferSize, cancellationToken);
+            }
+            catch (Exception ex)
+            {
+                input.Complete(ex);
+                return;
+            }
+            return;
```
halter73: Shouldn't there also be a call to `input.Complete()` here?

Drawaes: Yes, I have fixed it... I will chuck a couple of tests in tonight and submit a new PR.

halter73: Awesome. I was also wondering if AsPipelineWriter should be calling this method instead of CopyToAsync() so exceptions are handled.

PR #1222, opened 2017-02-16
Added SIMD-based IndexOf

KrzysztofCwalina:
IndexOf is very commonly used in parsers. This PR is the first stab at implementing vectorized IndexOf operating over Span<T>. 

I added a simple benchmark, and it shows that vectorized IndexOf is ~3x faster than the sequential IndexOf for searching items far into the buffer. It is significantly slower if the searched for item is at index 0. The break even point is around index 30.

Test Name            /                            Average	[time]
IndexOfBench.VectorizedIndexOf	  0.091765931	
IndexOfBench.SpanIndexOf	          0.29441888	

Note, that I tried to implement a hybrid algorithm, i.e. sequential for indexes 0-32 and then vectorized. Unfortunately the overhead of the branch, slicing, etc. was so high that it almost nullified the gains for indexes 0-30 and then made perf slightly worse for other indexes. I might play with this idea a bit more later.

One more thing, there is a small different between Span and ReadOnlySpan IndexOfVectorized implementations. Span has the GetItem method that returns a by ref (i.e. no copy). ReadOnlySpan does not and so I have to use the indexer (which copies a large vector). @jaredpar, are we getting readonly by refs soon? It would be a great match for ReadOnlySpan.

cc: @ahsonkhan, @shiftylogic, @vancem, @jkotas, @joshfree, @benaadams, @davidfowl, @sivarv

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
@@ -3,6 +3,7 @@
 
 using System.Collections.Sequences;
 using System.Diagnostics;
+using System.Numerics;
 using System.Runtime.CompilerServices;
 
 namespace System.Buffers
@@ -117,5 +118,107 @@ namespace System.Buffers
 
             return -1;
         }
+
+        static readonly int s_longSize = Vector<ulong>.Count;
+        static readonly int s_byteSize = Vector<byte>.Count;
+
+        public static int IndexOfVectorized(this Span<byte> buffer, byte value)
+        {
+            Debug.Assert(s_longSize == 4 || s_longSize == 2);
+
+            var byteSize = s_byteSize;
+
+            if (buffer.Length < byteSize * 2 || !Vector.IsHardwareAccelerated) return buffer.IndexOf(value);
+
+            Vector<byte> match = new Vector<byte>(value);
+            var vectors = buffer.NonPortableCast<byte, Vector<byte>>();
+            var zero = Vector<byte>.Zero;
+
+            for (int vectorIndex = 0; vectorIndex < vectors.Length; vectorIndex++)
+            {
+                var vector = vectors.GetItem(vectorIndex);
+                var result = Vector.Equals(vector, match);
+                if (result != zero)
+                {
+                    var longer = Vector.AsVectorUInt64(result);
+                    Debug.Assert(s_longSize == 4 || s_longSize == 2);
+
+                    var candidate = longer[0];
+                    if (candidate != 0) return vectorIndex * byteSize + IndexOf(candidate);
+                    candidate = longer[1];
+                    if (candidate != 0) return 8 + vectorIndex * byteSize + IndexOf(candidate);
+                    if (s_longSize == 4)
+                    {
+                        candidate = longer[2];
+                        if (candidate != 0) return 16 + vectorIndex * byteSize + IndexOf(candidate);
+                        candidate = longer[3];
+                        if (candidate != 0) return 24 + vectorIndex * byteSize + IndexOf(candidate);
+                    }
+                }
+            }
+
+            var processed = vectors.Length * byteSize;
+            var index = buffer.Slice(processed).IndexOf(value);
+            if (index == -1) return -1;
+            return index + processed;
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public static int IndexOfVectorized(this ReadOnlySpan<byte> buffer, byte value)
+        {
+            Debug.Assert(s_longSize == 4 || s_longSize == 2);
+
+            var byteSize = s_byteSize;
+
+            if (buffer.Length < byteSize * 2 || !Vector.IsHardwareAccelerated) return buffer.IndexOf(value);
+
+            Vector<byte> match = new Vector<byte>(value);
+            var vectors = buffer.NonPortableCast<byte, Vector<byte>>();
+            var zero = Vector<byte>.Zero;
+
+            for (int vectorIndex = 0; vectorIndex < vectors.Length; vectorIndex++)
+            {
+                var vector = vectors[vectorIndex];
+                var result = Vector.Equals(vector, match);
+                if (result != zero)
+                {
+                    var longer = Vector.AsVectorUInt64(result);
+                    var candidate = longer[0];
+                    if (candidate != 0) return vectorIndex * byteSize + IndexOf(candidate);
+                    candidate = longer[1];
+                    if (candidate != 0) return 8 + vectorIndex * byteSize + IndexOf(candidate);
+                    if (s_longSize == 4)
+                    {
+                        candidate = longer[2];
+                        if (candidate != 0) return 16 + vectorIndex * byteSize + IndexOf(candidate);
+                        candidate = longer[3];
+                        if (candidate != 0) return 24 + vectorIndex * byteSize + IndexOf(candidate);
+                    }
+                }
+            }
+
+            var processed = vectors.Length * byteSize;
+            var index = buffer.Slice(processed).IndexOf(value);
+            if (index == -1) return -1;
+            return index + processed;
+        }
+
+        // used by IndexOfVectorized
+        static int IndexOf(ulong next)
+        {
+            // Flag least significant power of two bit
+            var powerOfTwoFlag = (next ^ (next - 1));
+            // Shift all powers of two into the high byte and extract
+            var foundByteIndex = (int)((powerOfTwoFlag * _xorPowerOfTwoToHighByte) >> 57);
+            return foundByteIndex;
+        }
+
+        const ulong _xorPowerOfTwoToHighByte = (0x07ul |
+                                                0x06ul << 8 |
+                                                0x05ul << 16 |
+                                                0x04ul << 24 |
+                                                0x03ul << 32 |
+                                                0x02ul << 40 |
+                                                0x01ul << 48) + 1;
     }
 }
--- a/scripts/PerfHarness/PerfHarness.csproj
+++ b/scripts/PerfHarness/PerfHarness.csproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
     <DebugType>portable</DebugType>
     <AssemblyName>PerfHarness</AssemblyName>
     <OutputType>Exe</OutputType>
--- a/src/System.Buffers.Experimental/System.Buffers.Experimental.csproj
+++ b/src/System.Buffers.Experimental/System.Buffers.Experimental.csproj
@@ -13,5 +13,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="System.Buffers" Version="4.3.0" />
+    <PackageReference Include="System.Numerics.Vectors" Version="4.3.0" />
   </ItemGroup>
 </Project>
--- a/tests/Benchmarks/Benchmarks.csproj
+++ b/tests/Benchmarks/Benchmarks.csproj
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <TargetFramework>netcoreapp1.0</TargetFramework>
+    <TargetFramework>netcoreapp1.1</TargetFramework>
     <AllowUnsafeBlocks>False</AllowUnsafeBlocks>
     <AssemblyOriginatorKeyFile>../../tools/test_key.snk</AssemblyOriginatorKeyFile>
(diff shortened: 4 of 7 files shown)


joshfree:
@mellinoe could you also take a quick look?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public static int IndexOfVectorized(this ReadOnlySpan<byte> buffer, byte value)
+        {
+            Debug.Assert(s_longSize == 4 || s_longSize == 2);
+
+            var byteSize = s_byteSize;
+
+            if (buffer.Length < byteSize * 2 || !Vector.IsHardwareAccelerated) return buffer.IndexOf(value);
+
+            Vector<byte> match = new Vector<byte>(value);
```
benaadams: As above

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+
+        public static int IndexOfVectorized(this Span<byte> buffer, byte value)
+        {
+            Debug.Assert(s_longSize == 4 || s_longSize == 2);
+
+            var byteSize = s_byteSize;
+
+            if (buffer.Length < byteSize * 2 || !Vector.IsHardwareAccelerated) return buffer.IndexOf(value);
+
+            Vector<byte> match = new Vector<byte>(value);
```
benaadams: Needs jit fix https://github.com/dotnet/coreclr/pull/7683 to be performant; else use 
```csharp
Vector<byte> match = Vector.AsVectorByte(new Vector<uint>(value * 0x01010101u));
```

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+        static readonly int s_longSize = Vector<ulong>.Count;
+        static readonly int s_byteSize = Vector<byte>.Count;
+
+        public static int IndexOfVectorized(this Span<byte> buffer, byte value)
+        {
+            Debug.Assert(s_longSize == 4 || s_longSize == 2);
+
+            var byteSize = s_byteSize;
+
+            if (buffer.Length < byteSize * 2 || !Vector.IsHardwareAccelerated) return buffer.IndexOf(value);
```
benaadams: Test `!Vector.IsHardwareAccelerated` first so the Jit eliminates everything in the function. Might be worth adding a definitely inlined indirection shim?

```csharp
public static int IndexOfVectorized(this Span<byte> buffer, byte value)
{
    if (Vector.IsHardwareAccelerated && buffer.Length >= Vector<byte>.Count)
    {
        return buffer.IndexOfVectorizedImpl(value);
    }

    return buffer.IndexOf(value);
}
```

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        public static int IndexOfVectorized(this ReadOnlySpan<byte> buffer, byte value)
+        {
+            Debug.Assert(s_longSize == 4 || s_longSize == 2);
+
+            var byteSize = s_byteSize;
+
+            if (buffer.Length < byteSize * 2 || !Vector.IsHardwareAccelerated) return buffer.IndexOf(value);
```
benaadams: As above

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
 
             return -1;
         }
+
+        static readonly int s_longSize = Vector<ulong>.Count;
+        static readonly int s_byteSize = Vector<byte>.Count;
+
+        public static int IndexOfVectorized(this Span<byte> buffer, byte value)
+        {
+            Debug.Assert(s_longSize == 4 || s_longSize == 2);
```
benaadams: With `AVX-512` this could go to `s_longSize == 8`; going above 64 bytes is probably unlikely in near term as cache line is 64 bytes which changing would probably break lots of assumptions in software

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+
+            var byteSize = s_byteSize;
```
benaadams: Jit team told me off for not using `Vector<byte>.Count` directly; as they can't infer intrincs like auto loop unrolling https://github.com/dotnet/coreclr/pull/8001

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+            var zero = Vector<byte>.Zero;
+
+            for (int vectorIndex = 0; vectorIndex < vectors.Length; vectorIndex++)
+            {
+                var vector = vectors.GetItem(vectorIndex);
+                var result = Vector.Equals(vector, match);
+                if (result != zero)
+                {
+                    var longer = Vector.AsVectorUInt64(result);
+                    Debug.Assert(s_longSize == 4 || s_longSize == 2);
```
benaadams: Might not be true on AVX -512

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+
+                    var candidate = longer[0];
```
benaadams: Change to for loop using `Vector<ulong>.Count` as limit https://github.com/dotnet/coreclr/pull/7912

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+                    if (candidate != 0) return vectorIndex * byteSize + IndexOf(candidate);
```
benaadams: break and continue as inline returns make codegen nasty; and loop for unrolling as above dotnet/coreclr#7912
```csharp
ulong candidate = 0;
int longIndex = 0;
for (; longIndex < Vector<ulong>.Count; longIndex++)
{
    var candidate = longer[longIndex];
    if (candidate == 0) continue;
    break;
}
return 8 * longIndex + vectorIndex * Vector<byte>.Count + IndexOf(candidate);
```

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+            }
+
+            var processed = vectors.Length * byteSize;
+            var index = buffer.Slice(processed).IndexOf(value);
+            if (index == -1) return -1;
+            return index + processed;
+        }
+
+        // used by IndexOfVectorized
+        static int IndexOf(ulong next)
```
benaadams: Force inline?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -117,5 +118,107 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyMe
...
+
+            Vector<byte> match = new Vector<byte>(value);
+            var vectors = buffer.NonPortableCast<byte, Vector<byte>>();
+            var zero = Vector<byte>.Zero;
+
+            for (int vectorIndex = 0; vectorIndex < vectors.Length; vectorIndex++)
+            {
+                var vector = vectors.GetItem(vectorIndex);
+                var result = Vector.Equals(vector, match);
+                if (result != zero)
```
benaadams: Zero test directly rather than via variable; easier for Jit to pick up https://github.com/dotnet/coreclr/pull/7367
```csharp
!result.Equals(Vector<byte>.Zero)
```

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: I had it as such loop. Was 10% slower.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: Ah yeah. I think @sivarv told me about it.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
benaadams: /cc @JosephTremoulet

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: Good point. Will do

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: Yeah, I know about this issue. The question is: is it worth doing the workaround above? It's going to be slower when the fix is in.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: I thought I tested it and local was faster, but you are right that it should be slower. I will retest and possibly change.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: Thus the assert :-)

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: How did the attribute disappear from the PR? Seriously :-)

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
benaadams: Could `#ifdef` for current and above? (> Desktop 4.6.3? + Coreclr 1.2?) Don't really know the version numbers 😄

KrzysztofCwalina:
Hmmm, I made most of changes and the code got slower. Let me merge as is and I will try your suggestions one by one after I am back from vacations (next Tuesday).

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
halter73: @benaadams Will the JIT emit AVX-512 instructions on processors that support it today?

@KrzysztofCwalina Does the corefx(lab) CI run tests in Debug mode? If the JIT changes to support AVX-512, do any of the CI servers run a Xeon Phi processor or whatever it would take for this Debug.Assert to fail?

If we were to ship a System.Buffers package that didn't support a `Vector<ulong>.Count` of 8 (or  greater), could IndexOfVectorized simply skip over matching bytes and continue the for loop? If that were the case, Kestrel couldn't use it. That would be a security issue as that could cause Kestrel to read requests differently than proxies in front of it. Hopefully the server would just fall over instead.

Drawaes: Xeon Phi won't help you, it's a co-processor card you have to specifically target (usually with an intel c++ complier). I think some of the Sandy Bridge EP Xeon's have it, but I suspect it will be not a straight exposure of the registers because the AVX512 spec is a bit all over the shop.

benaadams:
I think its easier just to fix it rather than doing a `NonPortableCast`? https://github.com/dotnet/corefx/pull/16222/commits/186d3e6311885257625ac05689d90374072d754e

benaadams:
Added PR with changes in referenced corefx PR https://github.com/dotnet/corefxlab/pull/1231 - not sure how the xunit benchmark works; its seems to generate very very large csv files without much context, that than need to be interpreted?

PR #1223, opened 2017-02-16
Experiment to enable delayed disposal

mjp41:
Accessing the `ReferenceCounter` for every single `Dispose` could be expensive.

We make `ReferenceCounter` able to return a `BloomFilter` that over-approximates the elements that are referenced.  This means that a single scan of all the thread local state can enable many items to be disposed.

The `DelayedDisposer` class then wraps up this approach by batching up multiple `Dispose` calls, and checking the `BloomFilter` when it has sufficient items to continue.

Currently, there are two weaknesses
*  `BloomFilter` is very naive. It just takes bytes from the hashcode to calculate the different hashes. It does not adapt to the number of elements it is storing. 
* `DelayedDisposer` does not shrink back its window of elements if it is no longer necessary to be so large.

cc @KrzysztofCwalina @jkotas @davidfowl

### DIFF
--- a/src/System.Slices/System/Buffers/BloomFilter.cs
+++ b/src/System.Slices/System/Buffers/BloomFilter.cs
@@ -1,0 +1,46 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using System;
+
+
+namespace System.Runtime {
+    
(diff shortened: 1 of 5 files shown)


ahsonkhan:
@mjp41, is this PR still active? If so, please resolve the conflicts.

mjp41:
I don't think this is in the plan any more, so closing.

PR #1226, opened 2017-02-17
Stream To Pipe Unit tests and Bugfix ref #1217

Drawaes:
Doing these tests it's highlighted a mismatch in the behavior. The CopyToAsync completes the pipe however this is not the same behavior as the stream side. I think we might what to change this behavior thoughts?

### DIFF
--- a/src/System.IO.Pipelines/PipelineConnectionExtensions.cs
+++ b/src/System.IO.Pipelines/PipelineConnectionExtensions.cs
@@ -81,41 +81,54 @@ namespace System.IO.Pipelines
         {
             try
             {
-                await input.CopyToAsync(stream, bufferSize, cancellationToken);
+                // TODO: Use bufferSize argument
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    var result = await input.ReadAsync();
+                    var inputBuffer = result.Buffer;
+                    try
+                    {
+                        if (inputBuffer.IsEmpty && result.IsCompleted)
+                        {
+                            input.Complete();
+                            return;
+                        }
+
+                        await inputBuffer.CopyToAsync(stream);
+                    }
+                    finally
+                    {
+                        input.Advance(inputBuffer.End);
+                    }
+                }
             }
             catch (Exception ex)
             {
                 input.Complete(ex);
-                return;
             }
-            return;
         }
 
         public static Task CopyToAsync(this IPipeReader input, Stream stream)
         {
-            return input.CopyToAsync(stream, 4096, CancellationToken.None);
+            return input.CopyToAsync(stream, 4096);
         }
 
-        public static async Task CopyToAsync(this IPipeReader input, Stream stream, int bufferSize, CancellationToken cancellationToken)
+        public static async Task CopyToAsync(this IPipeReader input, Stream stream, int bufferSize)
         {
-            // TODO: Use bufferSize argument
-            while (!cancellationToken.IsCancellationRequested)
+            var result = await input.ReadAsync();
+            var inputBuffer = result.Buffer;
+            try
             {
-                var result = await input.ReadAsync();
-                var inputBuffer = result.Buffer;
-                try
+                if (inputBuffer.IsEmpty && result.IsCompleted)
                 {
-                    if (inputBuffer.IsEmpty && result.IsCompleted)
-                    {
-                        return;
-                    }
+                    return;
+                }
 
-                    await inputBuffer.CopyToAsync(stream);
-                }
-                finally
-                {
-                    input.Advance(inputBuffer.End);
-                }
+                await inputBuffer.CopyToAsync(stream);
+            }
+            finally
+            {
+                input.Advance(inputBuffer.End);
             }
         }
 
--- a/tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
+++ b/tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
@@ -26,12 +26,12 @@ namespace System.IO.Pipelines.Tests
                 {
                     var reader = pipe.Reader.ReadAsync();
                 });
-                Assert.Equal<byte>(buffer, stream.ToArray());
+                Assert.Equal(buffer, stream.ToArray());
             }
         }
 
         [Fact]
-        public async Task CopyFromPipeToEndThrow()
+        public async Task CopyFromPipeToEndObserveStreamThrow()
         {
             var buffer = new byte[1000];
             (new Random()).NextBytes(buffer);
@@ -50,7 +50,7 @@ namespace System.IO.Pipelines.Tests
         }
 
         [Fact]
-        public async Task CopyFromPipeNoEndThrow()
+        public async Task CopyFromPipeObserveStreamThrow()
         {
             var buffer = new byte[1000];
             (new Random()).NextBytes(buffer);
@@ -99,13 +99,13 @@ namespace System.IO.Pipelines.Tests
                 var reader = await pipe.Reader.ReadAsync();
 
                 Assert.True(reader.IsCompleted);
-                Assert.Equal<byte>(buffer, reader.Buffer.ToArray());
+                Assert.Equal(buffer, reader.Buffer.ToArray());
                 pipe.Reader.Advance(reader.Buffer.End);
             }
         }
 
         [Fact]
-        public async Task CopyFromStreamToEndThrow()
+        public async Task CopyFromStreamToEndObserveStreamThrow()
         {
             using (var pipeFactory = new PipeFactory())
             using (var stream = new FakeExceptionStream())
@@ -121,7 +121,7 @@ namespace System.IO.Pipelines.Tests
         }
 
         [Fact]
-        public async Task CopyFromStreamNoEndThrow()
+        public async Task CopyFromStreamNoEndObserveStreamThrow()
         {
             using (var pipeFactory = new PipeFactory())
             using (var stream = new FakeExceptionStream())
@@ -147,7 +147,7 @@ namespace System.IO.Pipelines.Tests
                 var reader = await pipe.Reader.ReadAsync();
 
                 Assert.False(reader.IsCompleted);
-                Assert.Equal<byte>(buffer, reader.Buffer.ToArray());
+                Assert.Equal(buffer, reader.Buffer.ToArray());
                 pipe.Reader.Advance(reader.Buffer.End);
             }
         }
--- a/tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
+++ b/tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
@@ -1,0 +1,168 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.IO.Pipelines.Tests
+{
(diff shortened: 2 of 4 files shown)


Drawaes:
cc @halter73 @davidfowl

halter73:
I would change CopyToAsync to not complete the pipe since we now have CopyToEndAsync.

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -88,7 +87,7 @@ public static Task CopyToEndAsync(this IPipeReader input, Stream stream)
                 input.Complete(ex);
```
halter73: Shouldn't AsPipelineWriter() call this method instead of CopyToAsync() so exceptions are handled?

davidfowl:
@Drawaes tests are failing and there are 9 commits. Can you fix the PR to:
- Be on top of master
- Fix whatever is failing

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+            (new Random()).NextBytes(buffer);
+            using (var pipeFactory = new PipeFactory())
+            using (var stream = new MemoryStream(buffer))
+            {
+                var pipe = pipeFactory.Create();
+                await stream.CopyToAsync(pipe.Writer);
+                var reader = await pipe.Reader.ReadAsync();
+
+                Assert.False(reader.IsCompleted);
+                Assert.Equal<byte>(buffer, reader.Buffer.ToArray());
```
davidfowl: nit: get rid of the generic arg.

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.IO.Pipelines.Tests
+{
+    public class PipeToStreamFacts
+    {
+        [Fact]
+        public async Task CopyFromPipeToEnd()
```
davidfowl: The test name should have the expected behavior. CopyFromPipeToEndThrowsWhen???

ahsonkhan:
> Yolo test rerun
FYI: You can re-run the tests by commenting: "@dotnet-bot test this" or specify which tests to rerun: "@dotnet-bot test Windows Debug" You shouldn't need to re-commit unnecessarily.

Drawaes:
Good to know

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -81,41 +81,54 @@ public static Task CopyToEndAsync(this IPipeReader input, Stream stream)
         {
             try
             {
-                await input.CopyToAsync(stream, bufferSize, cancellationToken);
+                // TODO: Use bufferSize argument
+                while (!cancellationToken.IsCancellationRequested)
```
halter73: If this method is going to take a CancellationToken, it makes sense to register a cancellation callback that calls CancelPendingRead() in case the IPipeReader is idle.

davidfowl: Lets do that in another PR.

halter73: Also, should input.Complete(TCE) be called if a cancellation is requested during the copy?

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -81,41 +81,54 @@ public static Task CopyToEndAsync(this IPipeReader input, Stream stream)
...
+
+                        await inputBuffer.CopyToAsync(stream);
+                    }
+                    finally
+                    {
+                        input.Advance(inputBuffer.End);
+                    }
+                }
             }
             catch (Exception ex)
```
halter73: I'm not sure it makes as much sense to swallow exceptions when copying to a Stream since the implementer of the stream can't observe the exception when IPipeReader.ReadAsync() throws.

There's also a decent chance the producer won't observe exceptions from Stream.WriteAsync() either if it throws after all the produced data has already been "flushed".

What do you think @davidfowl?

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
halter73: Agreed.

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+            var buffer = new byte[1000];
+            (new Random()).NextBytes(buffer);
+            using (var pipeFactory = new PipeFactory())
+            using (var stream = new FakeExceptionStream())
+            {
+                var pipe = pipeFactory.Create();
+                await pipe.Writer.WriteAsync(buffer);
+                pipe.Writer.Complete();
+                await pipe.Reader.CopyToEndAsync(stream);
+                await Assert.ThrowsAsync<InvalidOperationException>(async () =>
```
halter73: CopyFromPipeToEndObserveStreamThrow doesn't seem like the appropriate name for what this is testing.

If this was really observing the stream throwing, wouldn't this be an IOException? It seems more like CopyToEndAsync called `IPipeReader.Complete(IOException)`, the the producer was no longer around to observe it. This is a good example of the issue I described in my comment in PipelineConnectionExtensions.cs.

What's really happening here is that `IPipeReader.ReadAsync()` can't be called after  `IPipeReader.Complete()` is with or without an exception.

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+        {
+            var buffer = new byte[1000];
+            (new Random()).NextBytes(buffer);
+            using (var pipeFactory = new PipeFactory())
+            using (var stream = new MemoryStream())
+            {
+                var pipe = pipeFactory.Create();
+                await pipe.Writer.WriteAsync(buffer);
+                await pipe.Reader.CopyToAsync(stream);
+                var reader = pipe.Reader.ReadAsync();
```
halter73: I can see the argument for not calling `IPipeWriter.Complete()` when a stream is fully copied to a pipe so even more data can be copied. I don't see why you would ever not call `IPipeReader.Complete()` when the data from the pipe has been fully consumed.

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+
+                Assert.False(reader.IsCompleted);
+                Assert.Equal<byte>(buffer, stream.ToArray());
+            }
+        }
+
+        [Fact]
+        public async Task CopyFromStreamToEnd()
```
halter73: Like the first test, this could also use a better name. Maybe CopyFromStreamToEndCompletesPipeWriter.

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+                var reader = await pipe.Reader.ReadAsync();
+
+                Assert.True(reader.IsCompleted);
+                Assert.Equal(buffer, reader.Buffer.ToArray());
+                pipe.Reader.Advance(reader.Buffer.End);
+            }
+        }
+
+        [Fact]
+        public async Task CopyFromStreamToEndObserveStreamThrow()
```
halter73: Maybe CopyFromStreamToEndObserveStreamThrowsFromPipeReadAsync.

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+
+                await Assert.ThrowsAsync<IOException>(async () =>
+                {
+                    var reader = await pipe.Reader.ReadAsync();
+                });
+            }
+        }
+
+        [Fact]
+        public async Task CopyFromStreamNoEndObserveStreamThrow()
```
halter73: Maybe CopyFromStreamNoEndObserveStreamThrowsFromCopyToAsync.

tests/System.IO.Pipelines.Tests/PipeToStreamFacts.cs
```diff
@@ -0,0 +1,168 @@
...
+                var pipe = pipeFactory.Create();
+                await Assert.ThrowsAsync<IOException>(async () =>
+                {
+                    await stream.CopyToAsync(pipe.Writer);
+                });
+            }
+        }
+
+        [Fact]
+        public async Task CopyFromStreamNoEnd()
```
halter73: Maybe CopyFromStreamNoEndDoesNotCompletePipeWriter

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -81,41 +81,54 @@ public static Task CopyToEndAsync(this IPipeReader input, Stream stream)
...
+                // TODO: Use bufferSize argument
+                while (!cancellationToken.IsCancellationRequested)
+                {
+                    var result = await input.ReadAsync();
+                    var inputBuffer = result.Buffer;
+                    try
+                    {
+                        if (inputBuffer.IsEmpty && result.IsCompleted)
+                        {
+                            input.Complete();
```
davidfowl: I don't like this control flow. Do it at the end of the method in a finally or something.

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
halter73: Registering to the token can be done in another PR, but it still shouldn't be possible to exit this method without Completing the IPipeReader. The easiest solution for now is to remove the CancellationToken overload until it's properly supported.

As I mentioned in my comment in the "CopyFromPipeNoEnd()" test, it doesn't make sense to have separate CopyToAsync and CopyToEndAsync methods in this direction since you always finish reading the IPipeReader in either method (unless Stream.WriteAsync() throws I guess). Might as well call IPipeReader .Complete().

Even when Stream.WriteAsync() throws, a lot of the time the producer will no longer be producing data making it unable to observe the exception as is the case in the "CopyFromPipeToEndObserveStreamThrow" test.

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -81,41 +81,54 @@ public static Task CopyToEndAsync(this IPipeReader input, Stream stream)
...
         public static Task CopyToAsync(this IPipeReader input, Stream stream)
         {
-            return input.CopyToAsync(stream, 4096, CancellationToken.None);
+            return input.CopyToAsync(stream, 4096);
         }
 
-        public static async Task CopyToAsync(this IPipeReader input, Stream stream, int bufferSize, CancellationToken cancellationToken)
+        public static async Task CopyToAsync(this IPipeReader input, Stream stream, int bufferSize)
         {
-            // TODO: Use bufferSize argument
```
halter73: Why remove this comment? bufferSize is still unused in this method right?  Can we just remove the bufferSize argument?

Drawaes:
These methods are moving so let's close this change out

PR #1228, opened 2017-02-18
Reduce implied param passing

benaadams:
Change causes `OwnedMemory``1.get_Memory` to evaporate from the stack traces

4th param for instance call is passed via stack rather than reg; so cut memory op by looking it up later, rather than precall; and spilling a param to stack; `0` is also implied by this .`ctor` so doesn't need to be passed.

(this `RCX`, owner `RDX`, id `R8`, index `R9`, length `stack`) 
=> (this `RCX`, owner `RDX`, length `R8`) 

It is an extremely hot function

### DIFF
--- a/src/System.Slices/System/Buffers/Memory.cs
+++ b/src/System.Slices/System/Buffers/Memory.cs
@@ -19,7 +19,15 @@ namespace System
         readonly int _index;
         readonly int _length;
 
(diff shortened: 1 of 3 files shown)


benaadams:
Change causes `OwnedMemory``1.get_Memory` to evaporate from the stack traces

PR #1237, opened 2017-02-22
Enumerator for matches (wip)

benaadams:


benaadams:
Added some tests; `MatchesEnumerator` is much larger than I was aiming for 😢

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -253,6 +253,114 @@ public unsafe static int IndexOfVectorized(this ReadOnlySpan<byte> buffer, byte
...
+        }
+
+        public static string ToAsciiString(this ReadOnlySpan<byte> buffer)
+        {
+            if (buffer.Length == 0)
+            {
+                return string.Empty;
+            }
+
+            return string.Empty;
```
ahsonkhan: Should it be returning string.Empty?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -311,14 +419,291 @@ private static Vector<byte> GetVector(byte vectorByte)
...
+
+            public MatchesEnumerator GetEnumerator()
+            {
+                return new MatchesEnumerator(_buffer, _value, _start, _upperBound);
+            }
+        }
+
+        public struct MatchesEnumerator
+        {
+            private ReadOnlySpan<byte> _buffer; // don't make readonly, methods called on it
```
ahsonkhan: Can you explain this comment? _buffer is a ReadOnlySpan.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
benaadams: No; they are just filler; need to move that and sample function somewhere else. Was mostly for testing the api.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
benaadams: Meant not to qualify it as a readonly field e.g.

private **readonly** ReadOnlySpan<byte> _buffer;

As the function `.DangerousGetPinnableReference()` is called on it and readonly structs get copied when you call methods on them; so as not to potentially mutate the field struct [which results in a perf hit](https://codeblog.jonskeet.uk/2014/07/16/micro-optimization-the-surprising-inefficiency-of-readonly-fields/)

benaadams:
Is there a way of limiting what the PerHarness runs; rather than running everything?

benaadams:
Is there a way of limiting what the PerHarness runs; rather than running everything? It seems to be working so would like to test vs indexOf & vector IndexOf

ahsonkhan:
> Is there a way of limiting what the PerHarness runs; rather than running everything? It seems to be working so would like to test vs indexOf & vector IndexOf

The temporary workaround for now:
You could comment out the other assemblies here and only have your assembly (which contains only your perf tests) to run:
https://github.com/dotnet/corefxlab/blob/master/scripts/PerfHarness/PerfHarness.cs#L34

We already have an ask for the xunit.performance team to enable us to run perf tests by class type/name.

KrzysztofCwalina:
cc @shiftylogic 

Would it be better to have a more general purpose API? e.g.
 
// returns true if delimiter (stopAt) not hit
bool IndexesOf(ReadOnlySpan<byte> buffer, ReadOnlySpan<byte> value, Span<int> indexes, out int foundCount, ReadOnlySpan<byte> stopAt)?

Headers could be parsed as follows:
```c#
var indexes = stackalloc Span<byte>(20);
while(buffer.IndexesOf("\r\n"u8, indexes, out var found, "\r\n\r\n"u8)){
    for(int i=0; i<foundCount; i++){
        yield return buffer.Slice(indexes[i], indexes[i+1] - indexes[i]);
    }
}
```

benaadams:
`yield` would box though `IEnumerable`? so couldn't use spans?

KrzysztofCwalina:
Sorry. I did not mean that the calling code would literally yield and/or use spans. i.e. I don't known what the calling code would want to do, and it's the whole point of the more general API. If the code wanted to implement (on top of the more general API) an enumerable, it could. If it wanted to materialize strings and add them to a dictionary, it could. If it wanted to just record the indexes/ranges for later, it could. etc.

benaadams:
Param `Span indexes` would be `ref` and `found` not needed? So:
```csharp
bool IndexesOf(ReadOnlySpan buffer, ReadOnlySpan value, ref Span foundIndexes, ReadOnlySpan stopAt)?
```
Max to find specified by ingoing length and number found specified by out coming length? Or that confusing?

ahsonkhan:
> Max to find specified by ingoing length and number found specified by out coming length? Or that confusing?

I am confused.

KrzysztofCwalina:
buffer.Length before the call would indicate maximum number of indexes to find. buffer.Length after the call would indicate the number of indexes found.

I do think such API is a bit more difficult to understand what's going on, but if the perf benefit is there, we should consider.

benaadams:
Was me that was confused; forgot you could update a Span's data without passing it by `ref`; have been mixing too many concepts in too many PRs 😆

davidfowl:
@benaadams close?

PR #1239, opened 2017-02-23
Lock around awaitable state

pakrym:
@halter73 @davidfowl

### DIFF
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -331,7 +331,13 @@ namespace System.IO.Pipelines
                 return new ReadableBuffer(); // Nothing written return empty
             }
 
-            return new ReadableBuffer(new ReadCursor(_commitHead, _commitHeadIndex), new ReadCursor(_writingHead, _writingHead.End));
+            ReadCursor readStart;
+            lock (_sync)
+            {
+                readStart = new ReadCursor(_commitHead, _commitHeadIndex);
+            }
+
+            return new ReadableBuffer(readStart, new ReadCursor(_writingHead, _writingHead.End));
         }
 
         private ReadableBuffer Read()
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -44,6 +44,8 @@ namespace System.IO.Pipelines
         private PipeOperationState _producingState;
 
         private bool _disposed;
+        private IScheduler _readerScheduler;
+        private IScheduler _writerScheduler;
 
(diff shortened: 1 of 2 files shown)


pakrym:
Leaving `.Resume` in `PipeAwaitable` is questionable.

halter73:
Agreed about `.Resume`. Just call the scheduler directly.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -386,14 +402,20 @@ void IPipeReader.Advance(ReadCursor consumed, ReadCursor examined)
...
                                      !_writerCompletion.IsCompleted;
+
+                if (oldLength >= _maximumSizeLow &&
+                    _length < _maximumSizeLow)
+                {
+                    continuation = _writerAwaitable.Complete();
+                }
             }
             // We reset the awaitable to not completed if we've consumed everything the producer produced so far
             if (consumedEverything)
```
halter73: Reset() needs to be called inside the lock.

src/System.IO.Pipelines/PipeAwaitable.cs
```diff
@@ -85,32 +83,36 @@ public void OnCompleted(Action continuation, ref PipeCompletion completion)
                 Task.Run(continuation);
```
halter73: nit: remove the Interlocked.Exchange above.

davidfowl:
I really want to add tests for this somehow...

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -11,6 +11,10 @@ namespace System.IO.Pipelines
     /// </summary>
     internal class Pipe : IPipe, IPipeReader, IPipeWriter, IReadableBufferAwaiter, IWritableBufferAwaiter
     {
+        // This sync objects protects the following state:
+        // 1. _commitHead & _commitHeadIndex
```
halter73: You need to expand the lock in AllocateWriteHead too, to cover all access to _commitHead. While it might not be technically necessary yet, it will be if/when we allow the consuming thread to return the commit head once it's caught up in order to reduce the memory utilization if idle connections. This is what socket input does. See https://github.com/aspnet/KestrelHttpServer/pull/1120

davidfowl: I think mimicking socket input at this point is strictly necessary. So let's overlock unnecessarily but parity would be good.

halter73: Same goes for Append, AsReadableBuffer

halter73:
:shipit: pending final comments.

PR #1240, opened 2017-02-23
Various changes related to Kestrel updates

davidfowl:
- Added GetAsciiString extension method on `Span<byte>`
- Store the `Memory<byte>` on BufferSegment
- Removed unused read cursor out param on TryGetBuffer
- Remove bounds check from internal readable buffer ctor
- Fixed Cursor property on ReadableBufferReader
- Added ReadableBufferReader.Skip

/cc @benaadams @halter73 @pakrym

### DIFF
--- a/src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
+++ b/src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
@@ -86,7 +86,7 @@ namespace System.IO.Pipelines.Text.Primitives
         {
             uint value;
             int consumed;
-            if(!buffer.TryParseUInt32(out value, out consumed))
+            if (!buffer.TryParseUInt32(out value, out consumed))
             {
                 throw new InvalidOperationException("could not parse uint");
             }
@@ -109,14 +109,14 @@ namespace System.IO.Pipelines.Text.Primitives
                 ArraySegment<byte> data;
                 if (buffer.First.TryGetPointer(out pointer))
                 {
-                    if (!PrimitiveParser.InvariantUtf8.TryParseUInt64((byte*)pointer, len, out value)) 
+                    if (!PrimitiveParser.InvariantUtf8.TryParseUInt64((byte*)pointer, len, out value))
                     {
                         throw new InvalidOperationException();
                     }
                 }
                 else if (buffer.First.TryGetArray(out data))
                 {
-                    if (!PrimitiveParser.InvariantUtf8.TryParseUInt64(data.Array, out value)) 
+                    if (!PrimitiveParser.InvariantUtf8.TryParseUInt64(data.Array, out value))
                     {
                         throw new InvalidOperationException();
                     }
@@ -131,7 +131,8 @@ namespace System.IO.Pipelines.Text.Primitives
                 var data = stackalloc byte[len];
                 buffer.CopyTo(new Span<byte>(data, len));
                 addr = data;
-                if (!PrimitiveParser.InvariantUtf8.TryParseUInt64(addr, len, out value)) {
+                if (!PrimitiveParser.InvariantUtf8.TryParseUInt64(addr, len, out value))
+                {
                     throw new InvalidOperationException();
                 }
             }
@@ -139,7 +140,8 @@ namespace System.IO.Pipelines.Text.Primitives
             {
                 // Heap allocated copy to parse into array (should be rare)
                 var arr = buffer.ToArray();
-                if (!PrimitiveParser.InvariantUtf8.TryParseUInt64(arr, out value)) {
+                if (!PrimitiveParser.InvariantUtf8.TryParseUInt64(arr, out value))
+                {
                     throw new InvalidOperationException();
                 }
 
@@ -147,6 +149,31 @@ namespace System.IO.Pipelines.Text.Primitives
             }
 
             return value;
+        }
+
+        public unsafe static string GetAsciiString(this Span<byte> span)
+        {
+            if (span.IsEmpty)
+            {
+                return null;
+            }
+
+            var asciiString = new string('\0', span.Length);
+
+            fixed (char* outputStart = asciiString)
+            {
+                var output = outputStart;
+
+                fixed (byte* buffer = &span.DangerousGetPinnableReference())
+                {
+                    if (!AsciiUtilities.TryGetAsciiString(buffer, output, span.Length))
+                    {
+                        throw new InvalidOperationException();
+                    }
+                }
+            }
+
+            return asciiString;
         }
 
         /// <summary>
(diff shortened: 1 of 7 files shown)


src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -80,7 +86,7 @@ public int Take()
 
         private void MoveNext()
         {
-            if (_enumerator.MoveNext())
+            if (_enumerator.MoveNext() && !_enumerator.Current.IsEmpty)
```
pakrym: How is this correct? Having empty segment in the middle of the pipe should work fine.

benaadams:
Its as now. But I think the ascii throw should be a seek option rather than string creation? (i.e. so it throws before even getting to string creation) Though not for this PR

src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -41,11 +41,17 @@ public ReadableBufferReader(ReadCursor start, ReadCursor end) : this()
 
         public int Index => _overallIndex;
 
+        public Span<byte> Span => _currentMemory;
```
pakrym: Why are you exposing this?

davidfowl: Doesn't need to be exposed right now. I was playing with different patterns  earlier. I can remove it for now.

src/System.IO.Pipelines/ReadableBufferReader.cs
davidfowl: Should probably be a while like inside of the ctor. If you have an empty segment and then you peek. It will explode.

src/System.IO.Pipelines/ReadableBufferReader.cs
pakrym: Remove it please I prefer keeping RBR forward only reader API for now.

src/System.IO.Pipelines/ReadableBufferReader.cs
pakrym: Yeah, we need to gracefully handle this case.

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
```diff
@@ -109,14 +109,14 @@ public unsafe static ulong GetUInt64(this ReadableBuffer buffer)
                 ArraySegment<byte> data;
                 if (buffer.First.TryGetPointer(out pointer))
```
KrzysztofCwalina: @shiftylogic, you might want to take a look at this. It's unfortunate that this code has to fish out the ArraySegment and/or pointer out of the buffer instead of simply getting Span and parsing it. Maybe the code could simply use span and still get equal or better performance.

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
```diff
@@ -149,6 +151,31 @@ public unsafe static ulong GetUInt64(this ReadableBuffer buffer)
             return value;
         }
 
+        public unsafe static string GetAsciiString(this Span<byte> span)
```
KrzysztofCwalina: @shiftylogic, TextEncoder should be able to do this equally fast.

src/System.IO.Pipelines/ReadableBufferReader.cs
pakrym: I think you need to merge this but file and issue and assign it to me to fix it and add tests, I'd like to keep turnaround shorter.

davidfowl: Skipping empty spans on MoveNext is fine. We should actually be calling `MoveNext()` from the ctor

src/System.IO.Pipelines/ReadableBufferReader.cs
davidfowl: Will do

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
benaadams: Does it have ascii only? (i.e. not utf8)

tests/System.IO.Pipelines.Tests/ReadableBufferReaderFacts.cs
```diff
@@ -39,71 +49,117 @@ public void PeekReturnsMinuOneByteInTheEnd()
...
+            var buffer = BufferUtilities.CreateBuffer(new[] { new byte[] { 1 }, new byte[] { 2 }, new byte[] { 3 } });
+            var reader = new ReadableBufferReader(buffer);
+
+            Assert.Throws<ArgumentOutOfRangeException>(() => reader.Skip(4));
+        }
+
+        [Fact]
+        public void TakeTraversesSegments()
+        {
+            var buffer = BufferUtilities.CreateBuffer(new[] { new byte[] { 1 }, new byte[] { 2 }, new byte[] { 3 } });
```
pakrym: BufferUtilities.CreateBuffer has params. Remove outer array creation

tests/System.IO.Pipelines.Tests/ReadableBufferReaderFacts.cs
```diff
@@ -39,71 +49,117 @@ public void PeekReturnsMinuOneByteInTheEnd()
...
-                Assert.Equal(-1, reader.Take());
-            }
+            var reader = new ReadableBufferReader(ReadableBuffer.Create(new byte[] { 1, 2, 3, 4, 5 }, 0, 5));
+            Assert.Throws<ArgumentOutOfRangeException>(() => reader.Skip(6));
+        }
+
+        [Fact]
+        public void CtorFindsFirstNonEmptySegment()
+        {
+            var buffer = BufferUtilities.CreateBuffer(new[] { new byte[] { }, new byte[] { 1 } });
```
pakrym: If you don't care about values use `int[]` overload BufferUtilities.CreateBuffer(0, 1)

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
shiftylogic: Has it been tried to call buffer.First.Span and the primitive parsing routines that take a Span? I'd be curious to see how much difference there was in performance.

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
shiftylogic: @KrzysztofCwalina Are you referring to the span => Ascii string conversion? We'd need to add an Ascii encoder, but it should be pretty fast / easy.

davidfowl: Yes!

benaadams: Note byte->ascii string can get very fancy after https://github.com/dotnet/corefx/pull/16276

PR #1241, opened 2017-02-23
Add performance tests to Pipelines, optimize enumerator performance a bit

pakrym:
```
Before:

            Method |      Mean |    StdErr |    StdDev |    Median |        RPS | Allocated |
------------------ |---------- |---------- |---------- |---------- |----------- |---------- |
  MemoryEnumerator | 2.7179 us | 0.0419 us | 0.2296 us | 2.6041 us | 367,925.16 |       0 B |
 SegmentEnumerator | 1.6205 us | 0.0207 us | 0.1135 us | 1.6058 us | 617,082.52 |       0 B |

After:
            Method |      Mean |    StdErr |    StdDev |        RPS | Allocated |
------------------ |---------- |---------- |---------- |----------- |---------- |
  MemoryEnumerator | 2.4062 us | 0.0234 us | 0.1279 us | 415,591.36 |       0 B |
 SegmentEnumerator | 1.4740 us | 0.0178 us | 0.0974 us | 678,441.85 |       0 B |
```

@davidfowl @benaadams @KrzysztofCwalina


src/System.IO.Pipelines/SegmentEnumerator.cs
```diff
@@ -33,50 +35,55 @@ public SegmentEnumerator(ReadCursor start, ReadCursor end)
...
+                    _startIndex = _segment.Start;
                 }
             }
 
+            _current = new SegmentPart()
+            {
+                Segment = segment,
+                Start = start,
+                End = end,
+            };
```
benaadams: just update `_current` rather than calling `.ctor` + updates `{ ... }`?
```csharp
_current.Segment = segment;
_current.Start = start;
_current.End = end;
```
Might be good, might not...

benaadams:
`SegmentEnumerator` doesn't implement `IEnumerable` so ditch `Reset`?

src/System.IO.Pipelines/MemoryEnumerator.cs
```diff
@@ -17,7 +17,7 @@ public struct MemoryEnumerator
         public MemoryEnumerator(ReadCursor start, ReadCursor end)
         {
             _segmentEnumerator = new SegmentEnumerator(start, end);
-            _current = Memory<byte>.Empty;
+            _current = default(Memory<byte>);
```
davidfowl: Why this change?

pakrym: `Memory<byte>.Empty` is slow

davidfowl:
The performance project is failing.

src/System.IO.Pipelines/MemoryEnumerator.cs
davidfowl: What happens when you access `default(Memory<byte>)`? I think it explodes.

pakrym: It's `default(Memory<byte>)` before you call `MoveNext` and after it returns false, and for enumerator it's undefined behavior.

KrzysztofCwalina: I will make Empty faster. The current implementation is pretty naive. It calls OwnedMemory.Memory every time get_Empty is called.

PR #1244, opened 2017-02-23
Basic implementation of vectorized TryIndicesOf

shiftylogic:
Based on some simple micro-benchmarks, using TryIndicesOf vs IndexOf+Slice in a loop is anywhere from 3x faster (for finding elements 8 bytes apart) to 8x faster (for elements ~120 bytes apart).

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
@@ -246,6 +246,7 @@ namespace System.Buffers
 
         private unsafe static bool TryIndicesOf(ref byte searchSpace, byte value, int length, Span<int> indices, out int numberOfIndices)
         {
+            var result = false;
             numberOfIndices = 0;
 
             fixed (byte* pSearchSpace = &searchSpace)
@@ -319,6 +320,7 @@ namespace System.Buffers
                     }
 
                     // No Matches
+                    result = true;
                     break;
 
             exitFixed:;
@@ -328,7 +330,7 @@ namespace System.Buffers
                 }
             }
 
-            return numberOfIndices > 0;
+            return result && numberOfIndices < indices.Length;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
@@ -218,7 +218,118 @@ namespace System.Buffers
         exitFixed:;
                 return offset;
             }
+        }
 
+        public static bool TryIndicesOf(this Span<byte> buffer, byte value, Span<int> indices, out int numberOfIndices)
+        {
+            var length = buffer.Length;
+            if (length == 0 || indices.Length == 0)
+            {
+                numberOfIndices = 0;
+                return false;
+            }
+
+            return TryIndicesOf(ref buffer.DangerousGetPinnableReference(), value, length, indices, out numberOfIndices);
+        }
+
+        public static bool IndicesOf(this ReadOnlySpan<byte> buffer, byte value, Span<int> indices, out int numberOfIndices)
+        {
+            var length = buffer.Length;
+            if (length == 0 || indices.Length == 0)
+            {
+                numberOfIndices = 0;
+                return false;
+            }
+
+            return TryIndicesOf(ref buffer.DangerousGetPinnableReference(), value, length, indices, out numberOfIndices);
+        }
+
+
+        private unsafe static bool TryIndicesOf(ref byte searchSpace, byte value, int length, Span<int> indices, out int numberOfIndices)
+        {
+            numberOfIndices = 0;
+
+            fixed (byte* pSearchSpace = &searchSpace)
+            {
+                var searchStart = pSearchSpace;
+                var offset = 0;
+
+                while (true)
+                {
+                    if (Vector.IsHardwareAccelerated)
+                    {
+                        // Check Vector lengths
+                        if (length - Vector<byte>.Count >= offset)
+                        {
+                            Vector<byte> values = GetVector(value);
+                            do
+                            {
+                                var vFlaggedMatches = Vector.Equals(Unsafe.Read<Vector<byte>>(searchStart + offset), values);
+                                if (!vFlaggedMatches.Equals(Vector<byte>.Zero))
+                                {
+                                    // Found match, reuse Vector values to keep register pressure low
+                                    values = vFlaggedMatches;
+                                    break;
+                                }
+
+                                offset += Vector<byte>.Count;
+                            } while (length - Vector<byte>.Count >= offset);
+
+                            // Found match? Perform secondary search outside out of loop, so above loop body is small
+                            if (length - Vector<byte>.Count >= offset)
+                            {
+                                // Find offset of first match
+                                offset += LocateFirstFoundByte(values);
+                                // goto rather than inline return to keep function smaller
+                                goto exitFixed;
+                            }
+                        }
+                    }
+
+                    ulong flaggedMatches = 0;
+                    // Check ulong length
+                    while (length - sizeof(ulong) >= offset)
+                    {
+                        flaggedMatches = SetLowBitsForByteMatch(*(ulong*)(searchStart + offset), value);
+                        if (flaggedMatches != 0)
+                        {
+                            // Found match
+                            break;
+                        }
+
+                        offset += sizeof(ulong);
+                    }
+
+                    // Found match? Perform secondary search outside out of loop, so above loop body is small
+                    if (length - sizeof(ulong) >= offset)
+                    {
+                        // Find offset of first match
+                        offset += LocateFirstFoundByte(flaggedMatches);
+                        // goto rather than inline return to keep function smaller
+                        goto exitFixed;
+                    }
+
+                    // Haven't found match, scan through remaining
+                    for (; offset < length; offset++)
+                    {
+                        if (*(searchStart + offset) == value)
+                        {
+                            // goto rather than inline return to keep loop body small
+                            goto exitFixed;
+                        }
+                    }
+
+                    // No Matches
+                    break;
+
+            exitFixed:;
+                    indices[numberOfIndices++] = offset++;
+                    if (numberOfIndices >= indices.Length)
+                        break;
+                }
+            }
+
+            return numberOfIndices > 0;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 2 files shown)


shiftylogic:
@KrzysztofCwalina @davidfowl @ahsonkhan

shiftylogic:
FYI, I have some tests in a private project. I'll move them over to our test project.

ahsonkhan:
Can we have a test/have a graph which shows the perf difference for finding elements n bytes apart? Would be interesting to visualize how 'n' affects perf for 0 < n < 100.

ahsonkhan:
> TryIndicesOf 

Nit: Let's all use either IndicesOf or IndexesOf but not both ;)
So, we are using IndicesOf then?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -218,7 +218,118 @@ private static unsafe int IndexOfVectorized(ref byte searchSpace, byte value, in
...
+            if (length == 0 || indices.Length == 0)
+            {
+                numberOfIndices = 0;
+                return false;
+            }
+
+            return TryIndicesOf(ref buffer.DangerousGetPinnableReference(), value, length, indices, out numberOfIndices);
+        }
+
+        public static bool IndicesOf(this ReadOnlySpan<byte> buffer, byte value, Span<int> indices, out int numberOfIndices)
```
ahsonkhan: Should both TryIndicesOf and IndicesOf be public?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -218,7 +218,118 @@ private static unsafe int IndexOfVectorized(ref byte searchSpace, byte value, in
         exitFixed:;
                 return offset;
             }
+        }
+
+        public static bool TryIndicesOf(this Span<byte> buffer, byte value, Span<int> indices, out int numberOfIndices)
```
ahsonkhan: ReadOnlySpan instead of Span?

shiftylogic:
The pattern we have so far is that if you are writing into a buffer, if should be a "TryXxxxx" with a bool return. We have to convey failure which includes no indices found.

This is clearly open for debate. I'd like to hear some other opinions.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
shiftylogic: This was a mistype.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
shiftylogic: You need both Span and ReadOnlySpan. I have both. I just mistyped the name of one of the methods.

benaadams:
Just saw this on the twitter streams: https://twitter.com/jaredpar/status/834884511930331136

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -218,7 +218,117 @@ private static unsafe int IndexOfVectorized(ref byte searchSpace, byte value, in
...
+                                offset += Vector<byte>.Count;
+                            } while (length - Vector<byte>.Count >= offset);
+
+                            // Found match? Perform secondary search outside out of loop, so above loop body is small
+                            if (length - Vector<byte>.Count >= offset)
+                            {
+                                // Find offset of first match
+                                offset += LocateFirstFoundByte(values);
+                                // goto rather than inline return to keep function smaller
+                                goto exitFixed;
```
KrzysztofCwalina: What if this vector contains more than one match?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -218,7 +218,117 @@ private static unsafe int IndexOfVectorized(ref byte searchSpace, byte value, in
...
+            if (length == 0 || indices.Length == 0)
+            {
+                numberOfIndices = 0;
+                return false;
+            }
+
+            return TryIndicesOf(ref buffer.DangerousGetPinnableReference(), value, length, indices, out numberOfIndices);
+        }
+
+        public static bool TryIndicesOf(this ReadOnlySpan<byte> buffer, byte value, Span<int> indices, out int numberOfIndices)
```
KrzysztofCwalina: It seems that to be able to parse HTTP, we will need:
1. The ability to match Span<byte> values
2. The ability to terminate the search when some terminator found. 

Is that an API that we can build on top of this one?

KrzysztofCwalina:
@ahsonkhan, I would also be interested in measurements for elements less than 120 bytes apart. I would say that in a typical http request, header newlines are ~15-30 bytes apart

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
davidfowl: Can you actually grab my kestrel branch and use this API? It would be great to show a real code sample

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
pakrym: How do you know if it's returned because all values were found vs there is no enough space in  `indices` span?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: Yeah, it would be great if you could validate/test this new API in the context of the following routine: https://github.com/aspnet/KestrelHttpServer/blob/davidfowl/changes/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/Frame.cs#L1492

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
shiftylogic: The code advances the offset up to the first byte then starts all over at the new offset (ignoring everything before it). I've tested this with bytes ranges that are both small (way less than a single vector) and large (several vectors). It is always correct and faster than indexof in a loop.

There is no doubt that this could be optimized more. But this alone is a perf win.

shiftylogic:
@KrzysztofCwalina Below is a table from my testing. Times are 1,000,000 iterations searching for a single byte with the specified gap between search hits and a span size of 4 KB.

| Bytes Apart | IndexOf + Slice | IndicesOf | Gain |
|---|---|---|---|
| 8 | 10865 ms | 3235 ms | 3.3x |
| 16 | 7307 ms | 1816 ms | 4x |
| 32 | 4333 ms | 1105 ms | 4x |
| 60 | 2935 ms | 627 ms | 4.6x |
| 80 | 2738 ms | 453 ms | 6x |
| 120 | 2730 ms | 339 ms | 8x |

benaadams:
Now we are cooking 👍

Also amortizes the cost of `fixed`

ahsonkhan:
@shiftylogic That is awesome. Is the gains linear proportional to "Bytes Apart"? What is the curve shape?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: I think it's ok for the initial merge, but I think it would be worth playing with an implementation that does not restart the search after finding a match. And that is for two reasons:
1. Restarting work costs. I think this is partially the reason for bytes apart having such large impact on the perf.
2. Today, it does not pay off to try to 32-byte align the vectorized data. I think the reason is that we laways generate unaligned load instructions for SIMD registers. I have heard this will change at some point, and then we will have to change the implementation to 32-byte align and then restarting (which messes up alignment) would not be an option.

KrzysztofCwalina:
cc: @sivarv, @vancem, @mellinoe, @CarolEidt

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
shiftylogic: Agreed. I attempted it briefly yesterday but it actually killed the performance (the way I was doing it). This is definitely something that needs to be done. It will just take a bit more work to get right.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
shiftylogic: I'll give this a try this morning and let you know how it goes.

KrzysztofCwalina:
looks good.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
benaadams: Could do in follow up PR. Something like adding another loop around the vector/ulong 
```csharp
do 
{
    var index = LocateFirstFoundByte(values);
    values[index] = 0;
    indices[numberOfIndices++] = offset + index;
    if (numberOfIndices >= indices.Length)
    {
        // exit
    }
} while (!values.Equals(Vector<byte>.Zero))
offset += Vector<byte>.Count;
```
Might require some work to get the most effective loop meshing.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
shiftylogic: Yeah, I meant to return false when search terminated due to running out of buffer space for the indices.

That would make the logic as follows (which is same as we do for encoders):
- If result == true, search completed entire buffer and indices contains all hits.
- If result == false and numberOfIndices == indices.Length, we found some hits and returned them to you, but ran out of room in the destination buffer. There may or may not be more hits further on in your data.
- If result == false and numberOfIndices returns 0, there were no hits (or you passed us an empty destination buffer so that's on you).

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -319,6 +320,7 @@ private unsafe static bool TryIndicesOf(ref byte searchSpace, byte value, int le
                     }
 
                     // No Matches
+                    result = true;
```
ahsonkhan: No Matches => result = true?

shiftylogic: Correct. See discussion above with the set of exit states.

pakrym: Yes, result does not mean that anything is found it means that the end was reached

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
benaadams: Actually might be able to separate `LocateFirstFoundByte` into two and ` & ~index` on the long from its iterations; might be messy though.

shiftylogic: Thanks @benaadams. I'll try to re-visit this next week after I attempt to use this new method in some real scenario.

Issue #1252, opened 2017-02-25
Add System.IO.Pipelines.Performance.Tests to the solution and update build script


ahsonkhan:
This was disabled in PR #1246.

Please add it to the corefxlab solution and add it back in the build script.

https://github.com/dotnet/corefxlab/blob/master/scripts/build.ps1#L50

ahsonkhan:
@pakrym

PR #1255, opened 2017-02-27
Fix performance project

pakrym:
Fixes: https://github.com/dotnet/corefxlab/issues/1252

@ahsonkhan

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -44,12 +44,6 @@ foreach ($file in [System.IO.Directory]::EnumerateFiles("$PSScriptRoot\..\src", 
 }
 
 foreach ($file in [System.IO.Directory]::EnumerateFiles("$PSScriptRoot\..\tests", "*.csproj", "AllDirectories")) {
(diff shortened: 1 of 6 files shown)


scripts/build.ps1
```diff
@@ -44,12 +44,6 @@ foreach ($file in [System.IO.Directory]::EnumerateFiles("$PSScriptRoot\..\src",
 }
 
 foreach ($file in [System.IO.Directory]::EnumerateFiles("$PSScriptRoot\..\tests", "*.csproj", "AllDirectories")) {
-    if ($file -match "System.IO.Pipelines.Performance.Tests")
```
ahsonkhan: I thought you didn't want these tests to run as part of the regular build? Why remove this?

joshfree:
LGTM

scripts/build.ps1
pakrym: Test runner does not fail if there is no tests, it would just build and don't running anything. So there is no real need to special case `System.IO.Pipelines.Performance.Tests`.

Issue #1232, opened 2017-02-21
Add Span.IndexOfAny(span)


benaadams:
search 3 with `new Span((void*)&int, 0, 3)`

/cc @davidfowl

ahsonkhan:
This is complete, correct? As of: https://github.com/dotnet/corefx/pull/17357

shiftylogic:
I believe that is correct. We have the necessary variants of IndexOfAny and they are vectorized.

PR #1257, opened 2017-03-01
Faster seek implementation using span.IndexOfVectorized

davidfowl:
There's likely a gain and a loss here since we're using the `MemoryEnumerator` to look at the data. Will measure the differences.

### DIFF
--- a/src/System.IO.Pipelines/MemoryEnumerator.cs
+++ b/src/System.IO.Pipelines/MemoryEnumerator.cs
@@ -27,6 +27,8 @@ namespace System.IO.Pipelines
 
         internal BufferSegment CurrentSegment => _segmentEnumerator.Current.Segment;
 
+        internal int CurrentSegmentStartIndex => _segmentEnumerator.Current.Start;
+
         /// <summary>
         /// Moves to the next <see cref="Memory{Byte}"/> in the <see cref="ReadableBuffer"/>
         /// </summary>
--- a/src/System.IO.Pipelines/MemoryEnumerator.cs
+++ b/src/System.IO.Pipelines/MemoryEnumerator.cs
@@ -27,8 +27,6 @@ namespace System.IO.Pipelines
 
         internal BufferSegment CurrentSegment => _segmentEnumerator.Current.Segment;
 
-        internal int CurrentSegmentStartIndex => _segmentEnumerator.Current.Start;
-
         /// <summary>
         /// Moves to the next <see cref="Memory{Byte}"/> in the <see cref="ReadableBuffer"/>
         /// </summary>
--- a/src/System.IO.Pipelines/ReadCursorOperations.cs
+++ b/src/System.IO.Pipelines/ReadCursorOperations.cs
@@ -10,15 +10,17 @@ namespace System.IO.Pipelines
     {
         public static int Seek(ReadCursor begin, ReadCursor end, out ReadCursor result, byte byte0)
         {
-            var enumerator = new MemoryEnumerator(begin, end);
+            var enumerator = new SegmentEnumerator(begin, end);
             while (enumerator.MoveNext())
             {
-                var span = enumerator.Current.Span;
-                var segment = enumerator.CurrentSegment;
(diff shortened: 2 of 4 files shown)


src/System.IO.Pipelines/MemoryEnumerator.cs
```diff
@@ -27,6 +27,8 @@ public MemoryEnumerator(ReadCursor start, ReadCursor end)
 
         internal BufferSegment CurrentSegment => _segmentEnumerator.Current.Segment;
 
+        internal int CurrentSegmentStartIndex => _segmentEnumerator.Current.Start;
```
pakrym: Abstraction leaks.

davidfowl: I'm going to use the Segment enumerator directly.

pakrym:
LGTM after using `SegmentEnumerator`

src/System.IO.Pipelines/ReadCursorOperations.cs
```diff
@@ -2,371 +2,94 @@
...
+                var span = enumerator.Current.Span;
+                var segment = enumerator.CurrentSegment;
 
-            var index = begin.Index;
-            var wasLastBlock = block.Next == null;
-            var following = block.End - index;
-            int byteIndex = int.MaxValue;
+                int index1 = span.IndexOfVectorized(byte0);
+                int index2 = span.IndexOfVectorized(byte1);
+                int index3 = span.IndexOfVectorized(byte2);
```
KrzysztofCwalina: This is quite inefficient. We setup the buffer vector 3 times. Please file an issue and assign it to @shiftylogic to provide an API that amortizes the buffer vector creation across multiple searches, e.g. IndexesOf(buffer, values, indexes)

src/System.IO.Pipelines/ReadCursorOperations.cs
```diff
@@ -2,371 +2,94 @@
...
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static Vector<byte> GetVector(byte vectorByte)
+        private static int MinIndex(int v1, int v2)
         {
-            // Vector<byte> .ctor doesn't become an intrinsic due to detection issue
-            // However this does cause it to become an intrinsic (with additional multiply and reg->reg copy)
-            // https://github.com/dotnet/coreclr/issues/7459#issuecomment-253965670
-            return Vector.AsVectorByte(new Vector<uint>(vectorByte * 0x01010101u));
+            v1 = v1 == -1 ? int.MaxValue : v1;
```
KrzysztofCwalina: So our searches return -1 for not found and then this routine is required to find the minimun index. If we had a IndexesOf method it would return the minimum index found in the first slot of the results span.

cc: @shiftylogic

src/System.IO.Pipelines/ReadCursorOperations.cs
davidfowl: Yes we know it's inefficient, the bug is here https://github.com/dotnet/corefxlab/issues/1232. Luckily, we don't call this anywhere in kestrel right now.

src/System.IO.Pipelines/ReadCursorOperations.cs
davidfowl: Agreed, see the above comment.

src/System.IO.Pipelines/System.IO.Pipelines.csproj
```diff
@@ -11,6 +11,7 @@
     <ProjectReference Include="..\System.Slices\System.Slices.csproj" />
     <ProjectReference Include="..\System.Binary\System.Binary.csproj" />
     <ProjectReference Include="..\System.Collections.Sequences\System.Collections.Sequences.csproj" />
+    <ProjectReference Include="..\System.Buffers.Experimental\System.Buffers.Experimental.csproj" />
```
davidfowl: /cc @KrzysztofCwalina we should we move these methods to corefx?

PR #1263, opened 2017-03-02
Only await non-complete writes

benaadams:
Is busy on the async

Valid change?

/cc @pakrym @davidfowl @halter73

### DIFF
--- a/src/System.IO.Pipelines/PipelineConnectionExtensions.cs
+++ b/src/System.IO.Pipelines/PipelineConnectionExtensions.cs
@@ -31,7 +31,13 @@ namespace System.IO.Pipelines
         private static Task FlushAsync(WritableBuffer writeBuffer)
         {
             var awaitable = writeBuffer.FlushAsync();
-            return awaitable.IsCompleted ? _completedTask : FlushAsyncAwaited(awaitable);
+            if (awaitable.IsCompleted)
+            {
+                awaitable.GetResult();
+                return _completedTask;
+            }
+
+            return FlushAsyncAwaited(awaitable);
         }
 
         private static async Task FlushAsyncAwaited(WritableBufferAwaitable awaitable)
--- a/src/System.IO.Pipelines/PipelineConnectionExtensions.cs
+++ b/src/System.IO.Pipelines/PipelineConnectionExtensions.cs
@@ -19,16 +19,24 @@ namespace System.IO.Pipelines
 
     public static class PipelineWriterExtensions
     {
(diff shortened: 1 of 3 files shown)


src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -19,16 +19,30 @@ public static Stream GetStream(this IPipeConnection connection)
...
+
         public static Task WriteAsync(this IPipeWriter output, Span<byte> source)
         {
             var writeBuffer = output.Alloc();
             writeBuffer.Write(source);
-            return AwaitFlushAsync(writeBuffer);
+            return FlushAsync(writeBuffer);
+        }
+
+        private static Task FlushAsync(WritableBuffer writeBuffer)
```
KrzysztofCwalina: How often pipleline async call complete synchronously? i.e. should we be using ValueTask?

davidfowl: There's no non generic ValueTask AFAIK. Caching the completed task here is fine.

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -19,16 +19,30 @@ public static Stream GetStream(this IPipeConnection connection)
...
+        {
+            var awaitable = writeBuffer.FlushAsync();
+            if (awaitable.IsCompleted)
+            {
+                awaitable.GetResult();
```
JamesNK: Out of interest: what is the purpose of the `GetResult()` here?

davidfowl: Throw an exception (if there was any)

benaadams: Is also `WritableBufferAwaitable` is custom awaiter on top of a [`IWritableBufferAwaiter` interface](https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines/IReadableBufferAwaiter.cs#L17-L24)  rather than a Task so can make less assumptions on how it behaves.

So for example it could be a reusable auto-reseting type; that would become incomplete once it completed/returned on an await; so calling `GetResult()` maintains the normal flow whatever that is.

benaadams: > Throw an exception (if there was any)

Gah wonder if this is related https://github.com/aspnet/KestrelHttpServer/issues/1435; might need to return a faulted Task... :-/

PR #1265, opened 2017-03-02
Small improvements to GetAsciiString

KrzysztofCwalina:
AsciiToString is showing as 3% of time in basic Kestrel benchmarks.

This adds perf tests for AsciiToString (leaving baseline in).
Also, the change contains small optimizations to the routine:

1. The routine used to be called on pointers, and then wrapped to work with spans. But the wrapper cannot be inlined. This change makes the wrapper inlinable. 
2. The validity checks were simplified.
3. Removed a couple of calls to Span.Length.

The performance improvements are small, but quite consistent.

Once we get SIMD widening APIs wired to JIT, I will try to rewrite the routine using SIMD.

```
      Test Name                                                                                                 | Metric                | Iterations |    AVERAGE |    STDEV.S |        MIN |        MAX
    :--------------------------------------------------------------------------------------------------------- |:--------------------- |:----------:| ----------:| ----------:| ----------:| ----------:
     GetAsciiStringBench.GetAsciiString(text: "/plaintext")                                                    | Duration              |    339     |     29.615 |      0.641 |     28.702 |     34.433
     GetAsciiStringBench.GetAsciiString(text: "/plaintext")                                                    | Instructions Retired  |    339     | 3.154E+008 | 4.911E+005 | 3.148E+008 | 3.195E+008
     GetAsciiStringBench.GetAsciiString(text: "/plaintext")                                                    | Branch Mispredictions |    339     |  67771.280 |   4035.207 |  61440.000 |  86016.000
     GetAsciiStringBench.GetAsciiString(text: "/plaintext")                                                    | Cache Misses          |    339     |  84529.841 |  46846.107 |  49152.000 | 5.530E+005
     GetAsciiStringBench.GetAsciiString(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...)         | Duration              |    156     |     64.886 |      2.809 |     62.976 |     82.883
     GetAsciiStringBench.GetAsciiString(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...)         | Instructions Retired  |    156     | 7.028E+008 | 8.133E+005 | 7.019E+008 | 7.094E+008
     GetAsciiStringBench.GetAsciiString(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...)         | Branch Mispredictions |    156     | 1.888E+005 |  11117.607 | 1.761E+005 | 2.908E+005
     GetAsciiStringBench.GetAsciiString(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...)         | Cache Misses          |    156     | 3.041E+005 | 3.025E+005 | 1.802E+005 | 2.220E+006
     GetAsciiStringBench.GetAsciiStringBaseline(text: "/plaintext")                                            | Duration              |    329     |     30.479 |      0.864 |     29.239 |     38.258
     GetAsciiStringBench.GetAsciiStringBaseline(text: "/plaintext")                                            | Instructions Retired  |    329     | 3.335E+008 | 5.508E+005 | 3.329E+008 | 3.383E+008
     GetAsciiStringBench.GetAsciiStringBaseline(text: "/plaintext")                                            | Branch Mispredictions |    329     |  70017.945 |   4776.711 |  61440.000 | 1.229E+005
     GetAsciiStringBench.GetAsciiStringBaseline(text: "/plaintext")                                            | Cache Misses          |    329     |  90286.298 |  85000.095 |  53248.000 | 8.724E+005
     GetAsciiStringBench.GetAsciiStringBaseline(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...) | Duration              |    140     |     72.405 |      0.774 |     71.343 |     77.400
     GetAsciiStringBench.GetAsciiStringBaseline(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...) | Instructions Retired  |    140     | 8.326E+008 | 4.052E+005 | 8.318E+008 | 8.334E+008
     GetAsciiStringBench.GetAsciiStringBaseline(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...) | Branch Mispredictions |    140     | 2.095E+005 |   4719.556 | 1.966E+005 | 2.212E+005
     GetAsciiStringBench.GetAsciiStringBaseline(text: "text/plain,text/html;q=0.9,application/xhtml+xml;q"...) | Cache Misses          |    140     | 2.411E+005 |  70829.991 | 1.925E+005 | 8.356E+005

```


KrzysztofCwalina:
cc: @ahsonkhan, @shiftylogic, @joshfree, @vancem

KrzysztofCwalina:
I just updated the test to also collect instructions retired. For the longer header the PR gets them down from 8.326E+008 to 7.028E+008

src/System.IO.Pipelines.Text.Primitives/AsciiUtilities.cs
```diff
@@ -13,6 +13,76 @@ internal class AsciiUtilities
...
+                i++;
+                *output = (char)*input;
+                output++;
+                input++;
+            }
+
+            return isValid <= 127;
+        }
+
+        public static unsafe bool TryGetAsciiStringBaseline(byte* input, char* output, int count)
```
jkotas: What's the point of keeping this around?

KrzysztofCwalina: I kept it for now to do baseline benchmarking. I will remove before merge.

KrzysztofCwalina: Actually, let me remove now. I will just keep the comparison table.

tests/Benchmarks/GetAsciiStringBench.cs
```diff
@@ -0,0 +1,34 @@
...
+        foreach (var iteration in Benchmark.Iterations) {
+            using (iteration.StartMeasurement()) {
+                for (int i = 0; i < s_loops; i++) {
+                    str = bytes.GetAsciiString();
+                    len += str.Length;
+                }
+            }
+        }
+        return len;
+    }
```
jorive: You could change the inner loop to look more like this:
```
[Benchmark(InnerIterationCount=1000000)]
void TestMethod()
{
      for(int i=0; i<Benchmark.InnerIterationCount; i++)
        // test code here
}
```

KrzysztofCwalina: will do

corefxlab.sln
```diff
@@ -131,7 +130,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Text.Json.Tests.Dyna
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Text.Primitives.Performance.Tests", "tests\System.Text.Primitives.Performance.Tests\System.Text.Primitives.Performance.Tests.csproj", "{04970622-6C7A-4FB4-814A-9BFE051A2E68}"
 EndProject
-Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "System.IO.Pipelines.Performance.Tests", "tests\System.IO.Pipelines.Performance.Tests\System.IO.Pipelines.Performance.Tests.csproj", "{E68CF803-CC9F-458A-B0C7-1C2E0A0A3E4A}"
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.IO.Pipelines.Performance.Tests", "tests\System.IO.Pipelines.Performance.Tests\System.IO.Pipelines.Performance.Tests.csproj", "{E68CF803-CC9F-458A-B0C7-1C2E0A0A3E4A}"
```
ahsonkhan: Are these changes to the sln file intended?

KrzysztofCwalina: I think they are caused by new version of VS. I have already submitted a couple of PR after updating to new VS and every time I had to roll back the sln changes. I think it's time to update the sln fine, unless you guys see a reason not to?

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
```diff
@@ -151,31 +151,31 @@ public unsafe static ulong GetUInt64(this ReadableBuffer buffer)
             return value;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public unsafe static string GetAsciiString(this Span<byte> span)
         {
-            if (span.IsEmpty)
-            {
+            var len = span.Length;
+            if (len == 0) {
```
ahsonkhan: This is faster than Span.IsEmpty?

```C#
public bool IsEmpty => _length == 0;
```
That is surprising. Why is that?

tests/Benchmarks/Properties/AssemblyInfo.cs
```diff
@@ -19,3 +20,5 @@
 // COM, set the ComVisible attribute to true on that type.
 [assembly: ComVisible(false)]
 
+[assembly: MeasureInstructionsRetired]
+
```
ahsonkhan: @jorive, I thought these results are captured by default now and hence we don't need to add this explicitly.

corefxlab.sln
ahsonkhan: On every commit? Each time we update VS (somewhat daily), the solution file will get updated when we open it in VS. Also, why is the GUID for the System.IO.Pipelines.Performance.Tests.csproj different in this case?

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
KrzysztofCwalina: It's not faster, but note that later in this routine, Span.Length is accessed several times, and so the new code accesses it once and stores the value in a local. But now I have cached Length and so I may as well use it to check if the span is empty.

tests/Benchmarks/Properties/AssemblyInfo.cs
KrzysztofCwalina: Some counters are in the default set. This one is not.

tests/Benchmarks/GetAsciiStringBench.cs
jorive: Yes, you are right. Currently the API does not divide by the number of inner iterations, and it is something we will take a look at, but it will break current behavior (we might add it under a switch). In addition, we could add options to override this at runtime.

PR #1266, opened 2017-03-03
IWritableBufferAwaiter fast path & Tasks cleanup

benaadams:
/cc @davidfowl

### DIFF
--- a/src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
+++ b/src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
@@ -1,10 +1,8 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Buffers;
 using System.Diagnostics;
-using System.Linq;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -18,8 +16,6 @@ namespace System.IO.Pipelines
     {
         private static readonly Action _awaitableIsCompleted = () => { };
         private static readonly Action _awaitableIsNotCompleted = () => { };
-
-        private static Task _completedTask = Task.FromResult(0);
 
         private Action _awaitableState;
 
--- a/src/System.IO.Pipelines/PipelineWriterExtensions.cs
+++ b/src/System.IO.Pipelines/PipelineWriterExtensions.cs
@@ -7,8 +7,6 @@ namespace System.IO.Pipelines
 {
     public static class PipelineWriterExtensions
     {
-        private readonly static Task _completedTask = Task.FromResult(0);
-
         public static Task WriteAsync(this IPipeWriter output, Span<byte> source)
         {
             var writeBuffer = output.Alloc();
@@ -19,12 +17,13 @@ namespace System.IO.Pipelines
         private static Task FlushAsync(WritableBuffer writeBuffer)
         {
             var awaitable = writeBuffer.FlushAsync();
-            if (awaitable.IsCompleted)
+            if (awaitable.IsCompletedSuccessfully)
             {
-                awaitable.GetResult();
-                return _completedTask;
+                return TaskUtilities.CompletedTask;
             }
 
+            // Failed results we pass to the awaited form so it can return
+            // a faulted task rather than throwing directly
             return FlushAsyncAwaited(awaitable);
         }
 
--- a/src/System.IO.Pipelines/System.IO.Pipelines.csproj
+++ b/src/System.IO.Pipelines/System.IO.Pipelines.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>An abstraction for doing efficient asynchronous IO</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFrameworks>netstandard1.1;netstandard1.3</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageIconUrl>http://go.microsoft.com/fwlink/?linkid=833199</PackageIconUrl>
     <!--<DefineConstants Condition="'$(Configuration)' == 'Debug'">$(DefineConstants);OPERATION_LOCATION_TRACKING;COMPLETION_LOCATION_TRACKING</DefineConstants>-->
--- a/src/System.IO.Pipelines/TaskUtilities.cs
+++ b/src/System.IO.Pipelines/TaskUtilities.cs
@@ -1,0 +1,25 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    internal class TaskUtilities
+    {
+        public static readonly Task<int> ZeroTask = Task.FromResult(0);
+#if NETSTANDARD1_3
+        public static readonly Task CompletedTask = Task.CompletedTask;
+        public static Task FromException(Exception exception)
+                        => Task.FromException(exception);
+#else
+        public static readonly Task CompletedTask = ZeroTask;
+        public static Task FromException(Exception exception)
+        {
+            var tcs = new TaskCompletionSource<int>();
+            tcs.SetException(exception);
+            return tcs.Task;
+        }
+#endif
+    }
+}
--- a/src/System.IO.Pipelines.Extensions/PipelineConnectionStream.cs
+++ b/src/System.IO.Pipelines.Extensions/PipelineConnectionStream.cs
@@ -1,8 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
-using System.IO;
 using System.Threading;
 using System.Threading.Tasks;
 
@@ -10,8 +8,7 @@ namespace System.IO.Pipelines
 {
     public class PipelineConnectionStream : Stream
     {
-        private readonly static Task<int> _initialCachedTask = Task.FromResult(0);
-        private Task<int> _cachedTask = _initialCachedTask;
+        private Task<int> _cachedTask = TaskUtilities.ZeroTask;
 
         private readonly IPipeConnection _connection;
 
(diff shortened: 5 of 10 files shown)


src/System.IO.Pipelines/PipelineConnectionExtensions.cs
```diff
@@ -33,8 +31,16 @@ private static Task FlushAsync(WritableBuffer writeBuffer)
             var awaitable = writeBuffer.FlushAsync();
             if (awaitable.IsCompleted)
             {
-                awaitable.GetResult();
-                return _completedTask;
+                try
+                {
+                    awaitable.GetResult();
+                }
+                catch(Exception e)
```
davidfowl: nit: space

benaadams:
Needs tests 😢

davidfowl:
Only for WriteAsync with an exception.

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
JamesNK: Why call `awaitable.GetResult()` to rethrow and then catch and wrap?

Why not test if `awaitable.IsFaulted` and return the exception on the task?

benaadams: `awaitable` is a a `WritableBufferAwaitable` and doesn't have an `IsFauted` property

JamesNK: I see! Carry on then.

JamesNK: p.s. if there was a fully declared type instead of `var` then I wouldn't have this confusion. Just saying... 😝

davidfowl: var all the things

Drawaes:
Unless you are talking to the corefx team instead of the asp.net team. Then it's "don't var all the things"

src/System.IO.Pipelines/PipelineConnectionExtensions.cs
MaximRouiller: From your own docs:
https://msdn.microsoft.com/en-us/library/ff926074.aspx#Implicitly-Typed-Local-Variables

davidfowl: Microsoft is a big company. We have our own style guide

Drawaes:
Hahaha... u are in corefxlab... just saying

davidfowl:
Needs a rebase

benaadams:
Rebased, moved more common

src/System.IO.Pipelines/TaskUtilities.cs
```diff
@@ -0,0 +1,28 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    internal class TaskUtilities
+    {
+        public static readonly Action AwaitableIsCompleted = () => { };
```
davidfowl: This doesn't belong here... I would put these back

src/System.IO.Pipelines/PipelineWriterExtensions.cs
```diff
@@ -19,12 +17,14 @@ public static Task WriteAsync(this IPipeWriter output, Span<byte> source)
         private static Task FlushAsync(WritableBuffer writeBuffer)
         {
             var awaitable = writeBuffer.FlushAsync();
-            if (awaitable.IsCompleted)
+            if (awaitable.IsCompletedSuccessfully)
             {
-                awaitable.GetResult();
-                return _completedTask;
+                awaitable.GetResult(); // Is this needed?
```
benaadams: Is this needed

JamesNK: If `GetResult()` is throwing an error then you will need to catch it and return a faulted task, no?

benaadams: Introduced IsCompletedSucessfully property so the exceptional path will go to the await path avoiding the try catch in the non exception route

JamesNK: Sounds like GetResult isn't needed if it will never throw. What other purpose does it have?

benaadams: Missed the cancelled case; fixed so isn't needed anymore. Cancelled also gets pushed out to the await

src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
```diff
@@ -66,6 +62,11 @@ public UnownedBufferReader()
 
         bool IReadableBufferAwaiter.IsCompleted => IsCompleted;
 
+        bool IReadableBufferAwaiter.IsCompletedSuccessfully => 
+            IsCompleted && 
+            _cancelledState == CancelledState.NotCancelled && 
+            !Reading.IsCanceled && !Reading.IsFaulted;
```
benaadams: This is a bit horrible

benaadams:
Should be some upstream changes like `result.IsCompleted` should change to `result.IsCompletedSuccessfully` in [FrameOfT](https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/FrameOfT.cs#L69) else it might loose cancellation or exception?

src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
davidfowl: Yea, I don't even like this change anymore...

benaadams: Trouble is this will throw rather than returning a faulted task https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines.Extensions/PipelineReaderExtensions.cs#L11-L22

davidfowl: Maybe it's time to use the async support for ValueTask then.

benaadams: Less allocaty, but still the slow path, I'll back out the changes for `IReadableBufferAwaiter` and raise an issue instead 😄

benaadams:
@davidfowl better?

src/System.IO.Pipelines/System.IO.Pipelines.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>An abstraction for doing efficient asynchronous IO</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFrameworks>netstandard1.1;netstandard1.3</TargetFrameworks>
```
davidfowl: Nooooooo

benaadams: `#if NETSTANDARD1_3` wouldn't work without it?

jkotas: Why not bump the one and only target to netstandard1.3 ?

benaadams:
dotnet:master
```
 Method |                       Type |      Mean |    StdDev |      Op/s | Allocated |
------- |--------------------------- |---------- |---------- |---------- |---------- |
 Output |                 LiveAspNet | 3.4662 us | 0.0456 us | 288499.85 |      69 B |
 Output |           PlaintextChunked | 1.8278 us | 0.0228 us | 547092.70 |      66 B |
 Output | PlaintextChunkedWithCookie | 2.6012 us | 0.0251 us | 384433.01 |      71 B |
 Output |        PlaintextWithCookie | 2.8224 us | 0.0755 us | 354307.26 |      72 B |
 Output |       TechEmpowerPlaintext | 1.9834 us | 0.0268 us | 504191.24 |      67 B |
```
PR
```
 Method |                       Type |      Mean |    StdDev |      Op/s | Allocated |
------- |--------------------------- |---------- |---------- |---------- |---------- |
 Output |                 LiveAspNet | 3.3890 us | 0.0356 us | 295072.68 |      69 B |
 Output |           PlaintextChunked | 1.7965 us | 0.0159 us | 556634.56 |      66 B |
 Output | PlaintextChunkedWithCookie | 2.5036 us | 0.0164 us | 399422.37 |      71 B |
 Output |        PlaintextWithCookie | 2.6238 us | 0.0316 us | 381131.95 |      72 B |
 Output |       TechEmpowerPlaintext | 1.8985 us | 0.0148 us | 526728.58 |      67 B |
```

pakrym:
@benaadams how is `IsCompletedSuccessfully` helping?

src/System.IO.Pipelines/TaskUtilities.cs
```diff
@@ -0,0 +1,25 @@
...
+        public static Task FromException(Exception exception)
+                        => Task.FromException(exception);
+#else
+        public static readonly Task CompletedTask = ZeroTask;
+        public static Task FromException(Exception exception)
+        {
+            var tcs = new TaskCompletionSource<int>();
+            tcs.SetException(exception);
+            return tcs.Task;
+        }
```
halter73: Why not just use this for netstandard1.3 instead of compiling two assemblies?

halter73:
@pakrym I would assume that it's the elision of the call to `IWritableBufferAwaiter.GetResult()`.

src/System.IO.Pipelines/TaskUtilities.cs
benaadams: Bump the framework version for Pipelines? Sounds way above my decision making ability 😄

benaadams:
> It's asymmetric between reader and writer

Added it reader also; but took it out of this PR; to look at separately as its a bit messy for UnownedBuffer

> how is IsCompletedSuccessfully helping?

Can't throw the exception from `GetResult` directly as it expects a faulted task; so would need to wrap it in a try/catch and then create a faulted task. Try/catch will doom any hope of inlining in common path.

> Can't we rearrange GetResult to get pretty much the same asm after inlining?

No, its doing a lot more, and it can't inline as its a call through an interface stub (that returns a struct which is then discarded)

benaadams:
Will have tests in morning

src/System.IO.Pipelines/TaskUtilities.cs
halter73: I probably wasn't clear. I was suggesting going back to just targeting netstandard1.1 and using just this code. It only seems like it might be slightly slower for the "exceptional" case.

ahsonkhan:
> Will have tests in morning

@benaadams, any updates to the PR?

ahsonkhan:
@davidfowl, @pakrym, what are the plans for this PR?

PR #1268, opened 2017-03-03
Moved optimized ASCII encoder to System.Text.Primitives

KrzysztofCwalina:
I tried to make it into a TryDecode method, but it was 5% slower for some strings.
Maybe we can try one more time once we have Kestrel benchmarks at goal and stable.

cc: @shiftylogic, @davidfowl, @ahsonkhan 

Also, i tried to make it into a nested PrimitiveEncoder.Ascii class, but we don't support extension methods on nested types.


ahsonkhan:
LGTM

PR #1273, opened 2017-03-04
Make MemoryPool ConcurrentQueue invariant

benaadams:
Makes the array in the ConcurrentQueue invariant


davidfowl:
Educate me here, what does this do?

PR #1277, opened 2017-03-06
Slim ReadableBufferReader

benaadams:
Now: 8 + 4 + 4 + 4 + 4 = 24 bytes + `sizeof(Span<byte>)`
Previously: 4 + 4 + 4 + (8 + 8 + 4 + 4 + (8 + 4 + 4)) = 52 bytes + `sizeof(Span<byte>)`
1 object ref rather than 3

Smaller `Skip` - faster in same span
Smaller `get Cursor`
Added zero cost `RemainingBytes` property

/cc @davidfowl @pakrym

### DIFF
--- a/src/System.IO.Pipelines/ReadableBufferReader.cs
+++ b/src/System.IO.Pipelines/ReadableBufferReader.cs
@@ -82,6 +82,8 @@ namespace System.IO.Pipelines
 
         public int ConsumedBytes => _consumedBytes;
 
+        public int RemainingBytes => _remainingBytes;
+
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public int Peek()
         {
--- a/src/System.IO.Pipelines/ReadableBufferReader.cs
+++ b/src/System.IO.Pipelines/ReadableBufferReader.cs
@@ -16,45 +16,30 @@ namespace System.IO.Pipelines
         private int _remainingBytes;
         private int _consumedBytes;
 
-        public ReadableBufferReader(ReadableBuffer buffer) : this()
+        public ReadableBufferReader(ReadableBuffer buffer)
         {
             var start = buffer.Start;
             var length = buffer.Length;
             var segment = start.Segment;
             var startIndex = start.Index;
 
-            Initalize(segment, startIndex, length);
-        }
+            while (segment != null && segment.End == startIndex)
+            {
+                segment = segment.Next;
+                if (segment == null)
+                {
+                    break;
+                }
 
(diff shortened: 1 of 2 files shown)


src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -1,132 +1,178 @@
...
-            _end = false;
-            _index = 0;
-            _consumedBytes = 0;
-            _enumerator = new SegmentEnumerator(start, end);
-            _currentSpan = default(Span<byte>);
-            MoveNext();
-        }
-
-        public bool End => _end;
+            var length = start.GetLength(end);
```
pakrym: This can be expensive.

src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -1,132 +1,178 @@
...
             return value;
         }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
-        private void MoveNext()
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Skip(int length)
         {
-            while (_enumerator.MoveNext())
+            if ((uint)_remainingBytes < (uint)length)
```
pakrym: Why casts?

src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -1,132 +1,178 @@
...
-            while (!_end && length > 0)
+            segment = segment.Next;
+            var remaining = skip;
+            var segmentLength = segment.ReadableBytes;
+            while (segmentLength <= remaining || segmentLength == 0)
             {
-                if ((_index + length) < _currentSpan.Length)
+                remaining -= segmentLength;
+
+                if (segment.Next == null && _remainingBytes == 0)
```
pakrym: I'm not a fan of having segment traverse logic in all the places

pakrym:
Benchmarks?

src/System.IO.Pipelines/ReadableBufferReader.cs
benaadams: tests both length being greater than remaining and length being negative (as then as `uint` it would be greater than `int.MaxValue`) in single test

benaadams:
Bugged

benaadams:
Make `First` calculated; calculate `Length` upfront, drop unused `.ctor` 

Still bugged (end cursor)

ahsonkhan:
@benaadams, what is the state of this PR?

davidfowl:
Going to close this for now since it's not critical

PR #1279, opened 2017-03-07
Add IndexOfVectorized that takes an index and count

benaadams:
Parallel of  #1278 

/cc @ahsonkhan


davidfowl:
No tests?

KrzysztofCwalina:
same question as in #1278: what's the break even point? 

The reason I am asking is that it would be great if we could simply tell people you can slice and do regular IndexOf in majority (or all) situations. We need to know when such guidance is not true.

davidfowl:
Seems like we might not need this

PR #1280, opened 2017-03-07
Remove 16 bytes from BufferSegment

benaadams:
Not sure caching memory here does much? Always calling it once and stashing, vs calling very light .ctor on use and inlining it?

/cc @davidfowl

### DIFF
--- a/src/System.IO.Pipelines/BufferSegment.cs
+++ b/src/System.IO.Pipelines/BufferSegment.cs
@@ -1,7 +1,6 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System;
 using System.Buffers;
 using System.Diagnostics;
 using System.Text;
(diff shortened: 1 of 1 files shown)


davidfowl:
Need to see how this affects Kestrel. Originally this showed up on the profile but we have since then gone through lots of iterations...

benaadams:
Used to have the check on owner so was much more expensive

KrzysztofCwalina:
Given BufferSegment is a class, do you think these 16 bytes matter?

benaadams:
Is allocated for every request; so increases the likelyhood the GC will become interested

benaadams:
Is allocated for every request (read); so increases the likelyhood the GC will become interested; if its not adding benefit is there a need for it?

KrzysztofCwalina:
If they are not adding benefits, we should of course not have them (the 16 bytes).

davidfowl:
I'd like to re-measure with this change.

davidfowl:
I think this would make the perf of writing worse.

PR #1281, opened 2017-03-07
Move less used methods and experiments out of main Pipelines project and general cleanup

pakrym:
@davidfowl


src/System.IO.Pipelines.Extensions/PipelineReaderExtensions.cs
```diff
@@ -0,0 +1,40 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.IO.Pipelines
+{
+    public static class PipelineReaderExtensions
+    {
+
```
davidfowl: nit:space

davidfowl:
/cc @anurse @pranavkm 

We need to mirror this new package once we merge this.

davidfowl:
See if you still need these references:

https://github.com/dotnet/corefxlab/blob/7f8d486eecf0d45393c294cb578b38bbc954507d/src/System.IO.Pipelines/System.IO.Pipelines.csproj#L11-L14

Also you should be able to move:
- ArrayBufferPool.cs
- CommonVectors.cs

davidfowl:
Does this method even work:

https://github.com/dotnet/corefxlab/blob/4e45ac3d9d2ffc41fe4c0d494931b3474e5a6fff/src/System.IO.Pipelines/PipelineReaderExtensions.cs#L15-L46

Span in async methods are illegal no?

pakrym:
@davidfowl yeah, it won't work, we need to start running tests on 2.0

PR #1283, opened 2017-03-08
ReadableBuffer cleanup and changes

davidfowl:
- Make first computed to save some room
- Move Equals and StartsWith to extensions

/cc @benaadams I took the computed first from https://github.com/dotnet/corefxlab/pull/1277/files

### DIFF
--- a/samples/System.IO.Pipelines.Samples/HttpClient/LibuvHttpClientHandler.cs
+++ b/samples/System.IO.Pipelines.Samples/HttpClient/LibuvHttpClientHandler.cs
@@ -161,15 +161,12 @@ namespace System.IO.Pipelines.Samples
                             responseBuffer = responseBuffer.Slice(1);
                             if (responseBuffer.Length == 0)
                             {
-                                ch = responseBuffer.First.Span[0];
-                            }
-                            responseBuffer = responseBuffer.Slice(1);
-
-                            if (ch == -1)
-                            {
                                 break;
                             }
-                            else if (ch == '\n')
+                            ch = responseBuffer.First.Span[0];
+                            responseBuffer = responseBuffer.Slice(1);
+
+                            if (ch == '\n')
                             {
                                 consumed = responseBuffer.Start;
                                 needMoreData = false;
--- a/samples/System.IO.Pipelines.Samples/HttpClient/LibuvHttpClientHandler.cs
+++ b/samples/System.IO.Pipelines.Samples/HttpClient/LibuvHttpClientHandler.cs
@@ -147,18 +147,22 @@ namespace System.IO.Pipelines.Samples
 
                     while (!responseBuffer.IsEmpty)
                     {
-                        var ch = responseBuffer.Peek();
-
-                        if (ch == -1)
+                        if (responseBuffer.Length == 0)
                         {
                             break;
                         }
+
+                        var first = responseBuffer.First.Span;
+                        int ch = first[0];
 
                         if (ch == '\r')
                         {
                             // Check for final CRLF.
                             responseBuffer = responseBuffer.Slice(1);
-                            ch = responseBuffer.Peek();
+                            if (responseBuffer.Length == 0)
+                            {
+                                ch = responseBuffer.First.Span[0];
+                            }
                             responseBuffer = responseBuffer.Slice(1);
 
                             if (ch == -1)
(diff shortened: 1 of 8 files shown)


src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -56,8 +59,8 @@ internal ReadableBuffer(ReadCursor start, ReadCursor end)
         {
             _start = start;
             _end = end;
-            start.TryGetBuffer(end, out _first);
-            _length = -1;
+            _length = start.GetLength(_end);
```
benaadams: `_length = start.GetLength(end);`? Skip the RAW hazard? Probably doesn't matter but [hazards](https://en.wikipedia.org/wiki/Hazard_(computer_architecture)#Read_after_write_.28RAW.29) keep me up at night

benaadams:
Would be major rework; but once length is calculated you could throw away the cursors and trim to 16 bytes
```csharp
public struct ReadableBuffer 
{
    private BufferSegment _segment;
    private int _index;
    private int _length;
}
``` 
Then traverse the segments with length rather than checking whether `== ._end`

benaadams:
Though it depends how much `End` cursor is used (and whether it can be replaced with length); as reconstructing it would be more expensive.

src/System.IO.Pipelines.Extensions/ReadableBufferExtensionsExtensions.cs
```diff
@@ -3,7 +3,7 @@
 
 namespace System.IO.Pipelines
 {
-    public static class TrySliceToExtensions
+    public static class ReadableBufferExtensionsExtensions
```
pakrym: `ExtensionsExtensions`

src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -40,7 +36,14 @@ public struct ReadableBuffer : ISequence<ReadOnlyMemory<byte>>
         /// <summary>
         /// The first <see cref="Memory{Byte}"/> in the <see cref="ReadableBuffer"/>.
         /// </summary>
-        public Memory<byte> First => _first;
+        public Memory<byte> First
```
pakrym: Make this a method so it would be more obvious that this is not cached?

davidfowl: Don't want I can add a method and deprecate this so it doesn't break immediately.

samples/System.IO.Pipelines.Samples/HttpClient/LibuvHttpClientHandler.cs
```diff
@@ -147,18 +147,22 @@ public LibuvHttpClientHandler()
...
                             // Check for final CRLF.
                             responseBuffer = responseBuffer.Slice(1);
-                            ch = responseBuffer.Peek();
+                            if (responseBuffer.Length == 0)
+                            {
+                                ch = responseBuffer.First.Span[0];
+                            }
                             responseBuffer = responseBuffer.Slice(1);
 
                             if (ch == -1)
```
benaadams: Will be == '\r' if `length > 0` and something else `if length == 0`? But never -1?

davidfowl: Crap, I messed this up. I need to rewrite all of this to use the `ReadableBufferReader`

src/System.IO.Pipelines/ReadableBuffer.cs
pakrym: Should've done this. :)

davidfowl: This made all of the code ugly. Filed this https://github.com/dotnet/corefxlab/issues/1285

davidfowl: I'll do it though. I want to rewrite a bunch of the samples.

mjp41:
Does it make sense to expose a FirstSpan? A lot of uses do First.Span...?

davidfowl:
> Does it make sense to expose a FirstSpan? A lot of uses do First.Span...?

That's not bad. The idea here is to deter people from using this type for looking at the data. Just get the span and pass it all the way down the stack. https://github.com/aspnet/KestrelHttpServer/blob/13519b6079043f6b800b70011175912abed57e19/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs#L34-L67

Get the span once then party on it and pass it down as long as possible. 

For more complex parsing, you make a `ReadableBufferReader` and enumerate spans on that directly:

https://github.com/aspnet/KestrelHttpServer/blob/13519b6079043f6b800b70011175912abed57e19/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs#L260-L378

PR #1289, opened 2017-03-08
Improve pipe throughput

pakrym:
1. Expand locks instead of interlocked
2. Avoid double locking
3. Inline `Read` so we don't copy big struct around

### DIFF
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -115,7 +115,7 @@ namespace System.IO.Pipelines
                 {
                     try
                     {
-                        AllocateWriteHead(minimumSize);
+                        AllocateWriteHeadUnsynchronized(minimumSize);
                     }
                     catch (Exception)
                     {
@@ -139,7 +139,7 @@ namespace System.IO.Pipelines
                 // Changing commit head shared with Reader
                 lock (_sync)
                 {
-                    segment = AllocateWriteHead(count);
+                    segment = AllocateWriteHeadUnsynchronized(count);
                 }
             }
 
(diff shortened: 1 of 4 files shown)


src/System.IO.Pipelines/Pipe.cs
```diff
@@ -566,7 +547,31 @@ ReadResult IReadableBufferAwaiter.GetResult()
...
+                ReadCursor readEnd;
+                // No need to read end if there is no head
+                var head = _readHead;
+                if (head == null)
+                {
+                    readEnd = new ReadCursor(null);
+                }
+                else
+                {
+                    // Reading commit head shared with writer
```
davidfowl: nit: delete extra whitespace space

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -566,7 +547,31 @@ ReadResult IReadableBufferAwaiter.GetResult()
...
+
+                    readEnd = new ReadCursor(_commitHead, _commitHeadIndex);
+
+                }
+
+                _readingState.Begin(ExceptionResource.AlreadyReading);
```
davidfowl: Why not do it at the top? I guess it doesn't matter?

pakrym: Doesn't matter

davidfowl:
## baselien

|                  Method |        Mean |    StdErr |     StdDev |          RPS | Allocated |
|------------------------ |------------ |---------- |----------- |------------- |---------- |
| ParseLiveAspNetTwoTasks | 500.8234 ns | 8.4644 ns | 46.3615 ns | 1,996,711.75 |       1 B |
|   ParseLiveAspNetInline | 369.7139 ns | 3.8723 ns | 21.2096 ns | 2,704,793.85 |       1 B |

## PR

|                  Method |        Mean |     StdDev |          RPS |
|------------------------ |------------ |----------- |------------- |
| ParseLiveAspNetTwoTasks | 288.8195 ns | 14.6526 ns | 3,462,370.46 |
|   ParseLiveAspNetInline | 257.2562 ns |  7.5016 ns | 3,887,175.88 |

KrzysztofCwalina:
nice!

Issue #1287, opened 2017-03-08
GCHandle like API for pinning span/memory


davidfowl:
Today we have TryGetPointer but it's not safe to use unless you're using native memory or you pinned the underlying buffer. A safer API for pinning and unpinned outside of lexical scope and across async calls is needed. Maybe this should only be on memory and not span (though there might be cases for the latter).

/cc @jkotas @KrzysztofCwalina @mjp41

jkotas:
I think it should be something like:

```
struct PinnedMemoryHandle
{
     void Alloc<T>(Memory<T> memory);
     void Free();
     void * PinnedPointer { get; }
}
```

The usage pattern is same as GCHandle.

mjp41:
I like an API like @jkotas is proposing.  I think it makes sense, and helps me understand the correct usage of TryGetPointer.  Does this mean that TryGetArray would go away, and we would do the relevant GCHandle for pinning in Memory?

mjp41:
Would it be nicer for `Memory` to return one of these structs?

```csharp
PinnedMemoryHandle Memory<T>.GetPinnedMemoryHandle()

struct PinnedMemoryHandle {
    void Free();
    IntPtr GetPinnedPtr()
}
```

Possibly, using IDisposable?

Drawaes:
I think it should support IDisposable like

https://github.com/Drawaes/Leto/blob/master/src/Leto.Tls13/Internal/PointerFromMemory.cs

But obviously natively

jkotas:
These are "no copy" structs holding unmanaged resource. You have to be careful about not creating multiple copies,  disposing one copy, and still accessing other copy. It is different from the usual IDispose contract where calling dispose multiple times is fine, and it does not matter on which copy.

The shape of GCHandle APIs encourages the right behavior of never creating copies, and makes it clear to people that it is not an ordinary IDisposable type. 

So this is an API design question whether we like GCHandle prior art for this, or whether we would prefer something else.

jkotas:
> https://github.com/Drawaes/Leto/blob/master/src/Leto.Tls13/Internal/PointerFromMemory.cs

This is class. I agree that if this was a class, the IDisposable would be fine for it. However, my proposal is a struct to avoid heap allocations.

Drawaes:
[...]

PR #1290, opened 2017-03-09
Added PinnedMemoryHandle<T> to replace TryGetPointer

pdeligia:
Proposal to address issue #1287 

- Added API (similar to `GCHandle`) for scoped access to internal pointers of `Memory<T>` and `ReadOnlyMemory<T>`.
- Partially ported to the new `GetPinnedMemoryHandle` API.

CC @KrzysztofCwalina @davidfowl @jkotas @mjp41

### DIFF
--- a/src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
+++ b/src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
@@ -125,7 +125,7 @@ namespace System.IO.Pipelines.Compression
 
                     unsafe
                     {
-                        var handle = memory.GetPinnedMemoryHandle();
+                        var handle = memory.Pin();
                         handles.Add(handle);
                         _deflater.SetInput((IntPtr)handle.PinnedPointer, memory.Length);
                     }
@@ -135,7 +135,7 @@ namespace System.IO.Pipelines.Compression
                         unsafe
                         {
                             writerBuffer.Ensure();
-                            var handle = writerBuffer.Memory.GetPinnedMemoryHandle();
+                            var handle = writerBuffer.Memory.Pin();
                             handles.Add(handle);
                             int written = _deflater.ReadDeflateOutput((IntPtr)handle.PinnedPointer, writerBuffer.Memory.Length);
                             writerBuffer.Advance(written);
@@ -161,7 +161,7 @@ namespace System.IO.Pipelines.Compression
                     {
                         writerBuffer.Ensure();
                         var memory = writerBuffer.Memory;
-                        var handle = memory.GetPinnedMemoryHandle();
+                        var handle = memory.Pin();
                         handles.Add(handle);
                         int compressedBytes;
                         flushed = _deflater.Flush((IntPtr)handle.PinnedPointer, memory.Length, out compressedBytes);
@@ -182,7 +182,7 @@ namespace System.IO.Pipelines.Compression
                     {
                         writerBuffer.Ensure();
                         var memory = writerBuffer.Memory;
-                        var handle = memory.GetPinnedMemoryHandle();
+                        var handle = memory.Pin();
                         handles.Add(handle);
                         int compressedBytes;
                         finished = _deflater.Finish((IntPtr)handle.PinnedPointer, memory.Length, out compressedBytes);
@@ -241,12 +241,12 @@ namespace System.IO.Pipelines.Compression
                     {
                         unsafe
                         {
-                            var handle = memory.GetPinnedMemoryHandle();
+                            var handle = memory.Pin();
                             handles.Add(handle);
                             _inflater.SetInput((IntPtr)handle.PinnedPointer, memory.Length);
 
                             writerBuffer.Ensure();
-                            handle = writerBuffer.Memory.GetPinnedMemoryHandle();
+                            handle = writerBuffer.Memory.Pin();
                             handles.Add(handle);
                             int written = _inflater.Inflate((IntPtr)handle.PinnedPointer, writerBuffer.Memory.Length);
                             writerBuffer.Advance(written);
--- a/src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
+++ b/src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
@@ -2,6 +2,8 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System;
+using System.Buffers;
+using System.Collections.Generic;
 using System.IO.Compression;
 using System.Threading.Tasks;
 
@@ -100,6 +102,8 @@ namespace System.IO.Pipelines.Compression
 
             public async Task Execute(IPipeReader reader, IPipeWriter writer)
             {
+                List<MemoryHandle> handles = new List<MemoryHandle>();
+
                 while (true)
                 {
                     var result = await reader.ReadAsync();
@@ -122,8 +126,8 @@ namespace System.IO.Pipelines.Compression
                     unsafe
                     {
                         var handle = memory.GetPinnedMemoryHandle();
+                        handles.Add(handle);
                         _deflater.SetInput((IntPtr)handle.PinnedPointer, memory.Length);
-                        handle.Free();
                     }
 
                     while (!_deflater.NeedsInput())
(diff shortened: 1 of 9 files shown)


src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
```diff
@@ -197,27 +189,17 @@ public unsafe static string GetAsciiString(this ReadableBuffer buffer)
...
                             throw new InvalidOperationException();
                         }
                     }
-                    else
+                    finally
                     {
-                        ArraySegment<byte> data;
-                        if (memory.TryGetArray(out data))
-                        {
-                            fixed (byte* ptr = &data.Array[0])
```
mjp41: Using the new API, fixed has changed into `GCHandle` pinning. Is this acceptable?

src/System.Slices/System/Buffers/PinnedMemoryHandle.cs
```diff
@@ -0,0 +1,59 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace System
```
KrzysztofCwalina: I think it would be better to move this type to System.Buffers namespace. In fact, I think everything, we will move everything related to Memory<T> to this sub namespace. Possibly the only types that will stay in System are Span and ReadOnlySpan.

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
jkotas: This should get the Span out of Memory and fixed on call DangerousGetPinnableReference() instead. Much cheaper than allocating GCHandle.

src/System.Slices/System/Buffers/PinnedMemoryHandle.cs
```diff
@@ -0,0 +1,59 @@
...
+{
+    public unsafe struct PinnedMemoryHandle<T>
```
KrzysztofCwalina: I would just call it MemoryHandle, pinned is implied, just like in GCHandle.

src/System.Slices/System/Buffers/PinnedMemoryHandle.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                ArraySegment<T> buffer;
+                if (_owner.TryGetArrayInternal(out buffer))
+                {
+                    _handle = GCHandle.Alloc(buffer.Array, GCHandleType.Pinned);
+                    _pointer = Memory<T>.Add((void*)_handle.AddrOfPinnedObject(), buffer.Offset + index);
+                }
+                else
+                {
+                    throw new InvalidOperationException("Memory cannot be pinned");
```
KrzysztofCwalina: I am not sure this is what we want. It sees to be breaking the abstraction. I think for pointer based memory, we should just assume its pinned. 

@jkotas? @davidfowl?

src/System.Slices/System/Buffers/ReadOnlyMemory.cs
```diff
@@ -58,6 +58,8 @@ public DisposableReservation<T> Reserve()
         {
             return _owner.Memory.Reserve();
         }
+
+        public unsafe PinnedMemoryHandle<T> GetPinnedMemoryHandle() => new PinnedMemoryHandle<T>(_owner, _index);
```
KrzysztofCwalina: Should this be called simply Pin()?

KrzysztofCwalina:
Thanks! This is a nice and needed change.

src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
```diff
@@ -121,33 +121,20 @@ public WritableDeflateTransform(CompressionLevel compressionLevel, int bits)
...
-                        {
-                            _deflater.SetInput((IntPtr)inPointer, memory.Length);
-                        }
-                        else
-                        {
-                            throw new InvalidOperationException("Pointer needs to be pinned");
-                        }
+                        var handle = memory.GetPinnedMemoryHandle();
+                        _deflater.SetInput((IntPtr)handle.PinnedPointer, memory.Length);
+                        handle.Free();
```
jkotas: Are you sure that the deflater does not need this pinned longterm?

If it does not need to be pinned longterm, this should again use fixed + DangerousGetPinnableReference

src/System.Slices/System/Buffers/PinnedMemoryHandle.cs
jkotas: > pinned is implied, just like in GCHandle.

Pinned is not implied in GCHandle...

jkotas: Why does it need to be generic?

KrzysztofCwalina: How so? I meant GCHandle is not called "PinningGCHandle" yet everybody uses it for pinning.

KrzysztofCwalina: Has a field OwnedMemory<T>

jkotas: Why does it need a OwnedMemory field? It just unnecessary overhead to let you go back from this one back to the OwnedMemory.

KrzysztofCwalina: To call Release

jkotas: Only if you create it with GCHandleType.Pinned. Other GCHandleTypes has nothing to do with pinning.

I am fine with this being called MemoryHandle ... I was just pointing out that the pinning is not implied in GCHandle.

KrzysztofCwalina: OK, fair.

mjp41: This should perhaps use an IKnown, then we don't have to keep the <T>.

src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
jkotas: Ok. I am surprised to see that we are trying to make interop safe - that is lost cause because of it is inherently unsafe. But we gave up on safety for regular code where it would actually matter...

src/System.Slices/System/Buffers/PinnedMemoryHandle.cs
jkotas: Ok. I am surprised to see that we are trying to make interop safe - that is lost cause because of it is inherently unsafe. But we gave up on safety for regular code (for now) where it would actually matter...

src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
mjp41: I think you are right.  We probably need to keep all the MemoryHandles, until we get to the _deflater.Dispose(), then free them all.

src/System.Slices/System/Buffers/PinnedMemoryHandle.cs
mjp41: The GCHandle has to be released, so from a programmer perspective it is no more work.  We can drop it.

src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
davidfowl: Yes it needs to be pinned long term.

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
```diff
@@ -105,25 +105,17 @@ public unsafe static ulong GetUInt64(this ReadableBuffer buffer)
             if (buffer.IsSingleSpan)
             {
                 // It fits!
-                void* pointer;
-                ArraySegment<byte> data;
-                if (buffer.First.TryGetPointer(out pointer))
+                var handle = buffer.First.GetPinnedMemoryHandle();
```
davidfowl: This should just pin the span

src/System.Net.Libuv/System/Net/Libuv/Stream.cs
```diff
@@ -86,24 +86,26 @@ public unsafe void TryWrite(Memory<byte> data)
...
             EnsureNotDisposed();
 
-            void* pointer;
-            if (!data.TryGetPointer(out pointer))
-            {
-                throw new InvalidOperationException("Pointer not available");
-            }
-
-            IntPtr ptrData = (IntPtr)pointer;
+            var handle = data.GetPinnedMemoryHandle();
```
davidfowl: Should pin the span here

KrzysztofCwalina:
Looks good to me.

PR #1309, opened 2017-03-12
Attempt to use the buffer pool from System.Buffers.Experimental

davidfowl:
- Remove Return from the buffer pool. It means that disposing owned memory
 also returns it to the pool which is more user friendly IMO. One realization with this design is the fact that the `BufferPool` is just a way to get a pooled `OwnedMemory<byte>`. To all consumers of `OwnedMemory<byte>`, their usage doesn't change (that's a nice side effect IMO).
- All of the checks that were in Return that determined if the `OwnedMemory<byte>` originated from the pool can be removed since the `OwnedMemory<byte>` is self describing.
- There's no auto release on ref count = 0, the user still has the
dispose the buffer manually.


KrzysztofCwalina:
I worry a bit that the abstraction is still not general purpose enough (e.g. cannot be implemented by existing pools, seems to hardcode the shared pool, etc.), but I think this PR is an improvement over what we currently have, so let's merge it.

davidfowl:
> I worry a bit that the abstraction is still not general purpose enough (e.g. cannot be implemented by existing pools, seems to hardcode the shared pool, etc.), but I think this PR is an improvement over what we currently have, so let's merge it.

Why do you think that?

KrzysztofCwalina:
1. It cannot be implemented by ArrayPool<T> because ArrayPool<T> does not rent ownedmemory
2. Shared pool is hardcoded, or OwnedMmeory would have to carry a reference to the pool it came from and if we change OwnedMemory to an interface and implement it on T[], we won't be able to add a reference to the pool to T[].

davidfowl:
> It cannot be implemented by ArrayPool<T> because ArrayPool<T> does not rent ownedmemory

Sure it can but you just need to allocate the `OwnedMemory<T>` wrapper. That can be pooled as well. This is what pipelines does for it's default pool.

> Shared pool is hardcoded

That's not a design issue, just an implementation issue.

> or OwnedMmeory would have to carry a reference to the pool it came from and if we change OwnedMemory to an interface and implement it on T[], we won't be able to add a reference to the pool to T[].

Is that still on the table? To implement new interfaces on Array?

KrzysztofCwalina:
> Is that still on the table? To implement new interfaces on Array?

I would like to experiment with it.

benaadams:
> OwnedMmeory would have to carry a reference to the pool it came from and if we change OwnedMemory to an interface and implement it on T[], we won't be able to add a reference to the pool to T[].

Unless you are using a multi pool which is taking mixed mode `OwnedMemory` then its fairly straight forward?

```csharp
public interface OwnedMemory<T> :  IDisposable
{
    // ...
}

public Array<T> : OwnedMemory<T>
{
    public void OwnedMemory<T>.Dispose()
    {
        // Does nothing, just gets GC'd
    }
}
```

Drawaes:
I am not sure how I implement the dispose now, I guess I just carry a pointer back to the pool? 
[Ephemeral Pool](https://github.com/Drawaes/Leto/blob/v2/src/Leto/Internal/EphemeralBufferPoolUnix.cs)

benaadams:
Question is can you return to any pool or only pool you were rented from? If only pool you were rented from then it shouldn't be manually returned?

Drawaes:
Well you should probably limit it to the pool you came from... but actually it would work to any pool should stop that though because the pools will have different block sizes. Not sure you would want to return a windows buffer to the Unix pool... might work with the new linux subsystem but might cause a nuclear meltdown.

Drawaes:
So when you say it shouldn't be returned... you are saying a reference to the parent pool and use dispose or ref counting to return?

Issue #1298, opened 2017-03-10
Implement Memory<T> final design


ahsonkhan:
This probably needs to be broken down into additional issues/specific work items.

KrzysztofCwalina:
[ ] - Measure performance impact of adding poison pill ID back
[ ] - Measure performance impact of making abstraction to be filed-less.
[ ] - Move Memory and friends to System.Buffers namespace
[ ] - Move Memory and friends to System.Buffers.Primitives dll and package
[ ] - Change CopyTo to copy as much as possible and return int representing number of items copied
[ ] - Do we need both Pin() and Reserve() methods?
[ ] - Do we need TryGetArray and TryGetPointer?

benaadams:
> Do we need TryGetArray and TryGetPointer?

I think? the pinnable ref on span removes the need for these?

KrzysztofCwalina:
@benaadams, I think it covers TryGetPointer. TryGetArray might be needed for interop with existing APIs that operate on arrays only. But then, I am not sure we want to ship such leaky abstraction.

jkotas:
> TryGetArray might be needed for interop with existing APIs

It is necessary for *efficient* interop. 

It is pretty common for *efficient* interop to be leaky. And if you do not provide the official way to do it, people have to resort to hacks to do it.

The purist view won for `ImmutableArray<T>` and you see hacks like this to grow everywhere https://github.com/dotnet/roslyn/blob/614299ff83da9959fa07131c6d0ffbc58873b6ae/src/Compilers/Core/Portable/InternalUtilities/ImmutableArrayInterop.cs#L30 . I think these hacks are much worse than having official leaky API for it.

davidfowl:
> Measure performance impact of adding poison pill ID back

Make sure that includes measuring in Kestrel as well.

> Do we need TryGetArray and TryGetPointer?

Yes, lets keep both.

> Do we need both Pin() and Reserve() methods?

I'm not sure we ever needed Reserve really...

KrzysztofCwalina:
@davidfowl, I thought I added Reserve because you needed it.

davidfowl:
@KrzysztofCwalina yea, but it's not required anymore. Pipelines manually ref counts by directly using OwnedMemory<byte>`

mjp41:
> > Do we need TryGetArray and TryGetPointer?

> Yes, lets keep both.

[...]

PR #1312, opened 2017-03-14
Remove TryGetPointer call from UvTcpConnection

KodrAus:
Following the discussion in #1298

This is a really simple code change to remove the call to `TryGetPointer` in `UvTcpConnection`. It looks like the semantics are _similar_, but I don't think I'm fully following where that memory comes from and when it gets cleaned up, so want to make sure I'm not missing something important.

It probably also has a slight performance penalty.

I don't expect this is correct yet, just a starting point. Once this one's sorted I'll look at the RIO impl.

### DIFF
--- a/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
+++ b/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
@@ -25,6 +25,7 @@ namespace System.IO.Pipelines.Networking.Libuv
 
         private Task _sendingTask;
         private WritableBuffer? _inputBuffer;
+        private MemoryHandle? _inputBufferHandle;
 
         public UvTcpConnection(UvThread thread, UvTcpHandle handle)
         {
@@ -159,6 +160,10 @@ namespace System.IO.Pipelines.Networking.Libuv
                 // See the note at http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb.
                 _inputBuffer?.Commit();
                 _inputBuffer = null;
+
+                _inputBufferHandle?.Free();
+                _inputBufferHandle = null;
+
                 return;
             }
 
@@ -188,7 +193,10 @@ namespace System.IO.Pipelines.Networking.Libuv
             else
             {
                 var inputBuffer = _inputBuffer.Value;
+                var inputBufferHandle = _inputBufferHandle.Value;
+
                 _inputBuffer = null;
+                _inputBufferHandle = null;
 
                 inputBuffer.Advance(readCount);
                 inputBuffer.Commit();
@@ -215,6 +223,8 @@ namespace System.IO.Pipelines.Networking.Libuv
                         }
                     }
                 }
+
+                inputBufferHandle.Free();
             }
 
             if (normalDone)
@@ -231,13 +241,12 @@ namespace System.IO.Pipelines.Networking.Libuv
         private unsafe Uv.uv_buf_t OnAlloc(UvStreamHandle handle, int status)
         {
             var inputBuffer = _input.Writer.Alloc(2048);
+            var pinnedHandle = inputBuffer.Memory.Pin();
 
             _inputBuffer = inputBuffer;
+            _inputBufferHandle = pinnedHandle;
             
-            // REVIEW: Does this leak?
-            var pinned = inputBuffer.Memory.Pin();
-
-            return handle.Libuv.buf_init((IntPtr)pinned.PinnedPointer, inputBuffer.Memory.Length);
+            return handle.Libuv.buf_init((IntPtr)pinnedHandle.PinnedPointer, inputBuffer.Memory.Length);
         }
     }
 }
--- a/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
+++ b/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
@@ -191,7 +191,6 @@ namespace System.IO.Pipelines.Networking.Libuv
                 var inputBuffer = _inputBuffer.Value;
 
                 _inputBuffer = null;
-                _inputBufferPin = null;
 
                 inputBuffer.Advance(readCount);
                 inputBuffer.Commit();
--- a/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
+++ b/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
@@ -6,6 +6,7 @@ using System.Collections.Generic;
 using System.IO;
 using System.Threading.Tasks;
 using System.IO.Pipelines.Networking.Libuv.Interop;
+using System.Buffers;
 
 namespace System.IO.Pipelines.Networking.Libuv
 {
@@ -232,15 +233,11 @@ namespace System.IO.Pipelines.Networking.Libuv
             var inputBuffer = _input.Writer.Alloc(2048);
 
             _inputBuffer = inputBuffer;
+            
+            // REVIEW: Does this leak?
+            var pinned = inputBuffer.Memory.Pin();
 
-            void* pointer;
-            if (!inputBuffer.Memory.TryGetPointer(out pointer))
-            {
-                throw new InvalidOperationException("Pointer must be pinned");
-            }
-
-
-            return handle.Libuv.buf_init((IntPtr)pointer, inputBuffer.Memory.Length);
+            return handle.Libuv.buf_init((IntPtr)pinned.PinnedPointer, inputBuffer.Memory.Length);
         }
     }
 }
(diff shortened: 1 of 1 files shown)


src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -232,15 +233,11 @@ private unsafe Uv.uv_buf_t OnAlloc(UvStreamHandle handle, int status)
             var inputBuffer = _input.Writer.Alloc(2048);
 
             _inputBuffer = inputBuffer;
+            
+            // REVIEW: Does this leak?
+            var pinned = inputBuffer.Memory.Pin();
```
davidfowl: You should call Free in the OnRead callback

KodrAus:
@davidfowl Like this? It feels kind of weird having to call `Free` _and_ null the buffer variable, but maybe it's not actually a big deal.

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -6,6 +6,7 @@
 using System.IO;
 using System.Threading.Tasks;
 using System.IO.Pipelines.Networking.Libuv.Interop;
+using System.Buffers;
```
davidfowl: Sort usings

davidfowl:
> @davidfowl Like this? It feels kind of weird having to call Free and null the buffer variable, but maybe it's not actually a big deal.

Yea, the problem is that OnRead can get called without OnAlloc so you have to know if you did indeed call .Pin so we don't try to double free (decrement the ref count)

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -214,6 +223,8 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
                         }
                     }
                 }
+
+                inputBufferHandle.Free();
```
davidfowl: Why is this only freeing in the else case?

KodrAus: I've only calling `Free` in the blocks where `_inputBuffer` is null'd.

davidfowl: I think you have to call it everywhere there's a commit

KodrAus: Hmm sorry, do you mind explaining why that's the case? If we commit but keep the buffer then wouldn't it be a problem to drop the handle if it's still going to be used?

davidfowl: > Hmm sorry, do you mind explaining why that's the case? 

Why would that be a problem? The pin only needs to last as long as the native call that copies data into it. From then the GC is free to move it around. Once the OnRead callback is called it can be freed, otherwise, you may as well keep the `TryGetPointer` calls.

> If we commit but keep the buffer then wouldn't it be a problem to drop the handle if it's still going to be used?

Why? `OnAlloc` would be called again and we would pin the buffer again. The cases where `OnRead` gets called before `OnAlloc` are pretty rare. The usual pattern is `OnAlloc`, `OnRead`

KodrAus: Ok I think I get it now, I was conflating the lifetime of the handle with the lifetime of the buffer, rather than when it needs to be pinned.

KrzysztofCwalina:
cc: @mjp41

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -24,6 +22,7 @@ public class UvTcpConnection : IPipeConnection
 
         private Task _sendingTask;
         private WritableBuffer? _inputBuffer;
+        private MemoryHandle? _inputBufferPin;
```
mjp41: That you have had to implement this with a nullable type, makes we wonder if we should make Free callable several times, and only the first does something.

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -179,6 +180,7 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
                 error = new IOException(uvError.Message, uvError);
 
                 _inputBuffer?.Commit();
```
mjp41: Should _inputBuffer be set to `null` at this point?

davidfowl: It's possible OnAlloc didn't get called.

mjp41: So if OnAlloc hasn't been called, won't it already be `null`?

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
KodrAus: I was going to get around to asking if we should make `Free` idempotent. I've explicitly always set it to `null` in the unpin method because it's not safe to call multiple times. So that rules out the possibility of that happening

Drawaes:
If your are aiming for parity with gchandle the there should be an "IsAllocated" property
Then you can do Is allocated okay then free

Drawaes:
This allows a user who knows for sure it will be allocated to not have to pay the price of a check. But a user who is unsure can implement the check

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
davidfowl: That's why we use ?

mjp41:
LGTM, shall we raise an issue to make `Free` idempotent, or provide a IsAllocated? (@davidfowl, @KrzysztofCwalina)

KrzysztofCwalina:
Why would we want to allow multiple calls to free? Isn't it a bug to call it multiple times? If multiple threads call it, it seems like it's left to change that one of the threads won't need the handle after another thread frees it.

benaadams:
> Why would we want to allow multiple calls to free? Isn't it a bug to call it multiple times?

Both `MemoryHandle` and its internal `GCHandle` are pass by copy structs so you'd either need to deal with that if you want to catch it.

I assume GC Handle's `extren InternalFree` is idempotent; so is only the multi-call to `_owner.Release();` that needs caution?

jkotas:
> Why would we want to allow multiple calls to free? Isn't it a bug to call it multiple times?

It is convenient to allow calling Free if the handle was not allocated at all. Considers something like this:

```
class MyAsyncReader
{
    MemoryHandle<byte> _handle;

    ...

    ~MyReader()
    {
        // This can be null if an exception happened before the _handle was allocated so you need to check for IsAllocated
        _handle.Free();
    }
}
```

Similar if you do cleanup using try/finally blocks. You either need to structure your try/finally blocks just right, or check for IsAllocated before calling Free.

It matches what C/C++ does. `operator delete` or `free` can be passed null and do nothing in that case.

`GCHandle.Free` is over engineered in a bad way in this regard: It throws for null handle and thus requires you to check for IsAllocated; but has Interlocked.CompareExchange to protect against people calling it on multiple threads.

I do not have opinion on this yet. Just listing some reasons why it may be a good idea to allow it…

KrzysztofCwalina:
If you don't have the lifetime logic totally under control of a single thread, calling IsAllocated does not help either. It has a race and it would just make finding bugs harder.

Drawaes:
You might have the life time sorted .. it is a struct right? So consider a gchandle which was the Api guideline  rather than having to have nullable you can test in a destructor if it is already freed. If so don't call free.

KodrAus:
> Both MemoryHandle and its internal GCHandle are pass by copy structs so you'd need to deal with that if you want to catch it.

@benaadams Hmm, that's interesting, so you can't tie the `Free` method to a particular instance of the handle, since there may be more handles floating around that correspond to the one ref count.

My feeling is that because this is a public API we should make an effort to avoid footguns like multiple calls to `Free`. Since it's pass-by-copy you can't really store that as state on the handle itself, right?

benaadams:
>  Since it's pass-by-copy you can't really store that as state on the handle itself, right?

You can solve it for a class member or use within a single function stack space/async; its whether to go further and try to solve it for multiple copies?

Drawaes:
Ahh the footgun known as GCHandle... I am missing some toes due to that very thing
[Readonly Structs Copy When Calling a Method](https://github.com/dotnet/roslyn/issues/17310#issuecomment-282569477) 
At least it will be consistent with the current GCHandle 😈

KodrAus:
Is this discussion about `Free` likely to block this PR as it stands? Or can it be moved to a new issue and worked out independently?

KrzysztofCwalina:
I think we should merge with a change to allow multiple calls to Free, or merge as-is and then change to allow multiple calls.
@davidfowl?

mjp41:
If you merge as is. I can fix free today.

davidfowl:
Sure this is fine. I still don't want to remove TryGetPointer as yet.

PR #1315, opened 2017-03-14
Remove Span APIs that are not implemented.

ahsonkhan:


davidfowl:
I personally think we should go further and remove all of the APIs on `Span<char>` as well. They are sooo underused. Probably have tons of bugs.

KrzysztofCwalina:
Looks good. And I, regrettably, agree with @davidfowl. I would love having these string manipulation APIs, but we have ran out of time, it seems.

ahsonkhan:
We can always add APIs back when future scenarios require them.

PR #1317, opened 2017-03-14
Moving IndexOfVectorized (and tests) to SpanExtensions as part of cleanup

ahsonkhan:
Along with TryIndicesOf Span Extension methods.


ahsonkhan:
Don't merge yet, please.

ahsonkhan:
This can be merged now.

davidfowl:
Go ahead 😄

KrzysztofCwalina:
@benaadams submitted a PR to make IndexOf vectorized. If these two PR don't have too many conflicts, it's good to go.

benaadams:
They shouldn't have any conflicts; though `IndexOfVectorized` will become superfluous once its worked its way though the build.

@KrzysztofCwalina Is also the "early exit" variant that probably wants adding here for when you have a long input and expect to find the result in the first several bytes?

KrzysztofCwalina:
> @KrzysztofCwalina Is also the "early exit" variant that probably wants adding here for when you have a long input and expect to find the result in the first several bytes?

Yes. Last time I measured, if the buffer is long and the searched for item is in slot 0, the vectorized implementation was significantly slower (as in 10x slower).

Issue #1314, opened 2017-03-14
[Pre-release] Productize System.Span and ReadOnlySpan


ahsonkhan:
**Pri 0** - _Must complete for release_
- [x] Finalize API shape of SpanExtensionsLab and merge with SpanExtensions - #1292
- [x] Add tests to increase code coverage - https://github.com/dotnet/corefx/issues/17078
- [ ] Add stress tests - #1313
- [x] Remove System.Slices - #1295
- [x] Consider changing CopyTo methods to copy as much as possible and return int - https://github.com/dotnet/corefx/issues/16888
- [x] Should we rename Span<T> to Memory<T> and Memory<T> to something else - https://github.com/dotnet/corefx/issues/16997
- [x] Improve Span.StartsWith(ReadOnlySpan<T>|ReadOnlySpan<byte>) performance + investigate API that takes index - #1294
- [x] Span/ReadonlySpan TryCopy/CopyTo are very slow (slow span) - https://github.com/dotnet/corefx/issues/16840
- [x] System.Memory.Tests crashing on Windows outer loop runs - https://github.com/dotnet/corefx/issues/16492
- [x] Fix array variance checks in slow Span - https://github.com/dotnet/corefx/issues/14533
- [x] Remove System.Buffers.Experimental dependency from Pipelines.

**Pri 1** - _Nice to have_
- [ ] Debugger proxies don't work on Spans - https://github.com/dotnet/corefx/issues/14376
- [x] Test: System.SpanTests.ReadOnlySpanTests/IndexOverflow failed with "System.OutOfMemoryException" - https://github.com/dotnet/corefx/issues/16593
- [x] Span: DangerousCreate should avoid argument validation - https://github.com/dotnet/corefx/issues/14486

**Pri 2** - _Nice to have_

- [x] Fix Spanbench benchmark tests for fast span to be more reliable - https://github.com/dotnet/coreclr/issues/10099
- [ ] Add slow span benchmarks to benchview - https://github.com/dotnet/coreclr/issues/10098
- [x] Add IndexOfAny overload - #1293

---
**Future** - _Not part of this release_
~- [ ] Add a way to efficiently compare memory - https://github.com/dotnet/corefx/issues/16878~
~- [ ] Span<T> == null throws ArgumentNullException - https://github.com/dotnet/corefx/issues/16730~
[...]

PR #1319, opened 2017-03-15
Preparing for productization

shiftylogic:
- new trimmed solution file.
- Removed System.Slices.
- Created a new System.Buffers.Primitives assembly.
- Moved memory related structs (Memory<T>, OwnedMemory, etc) into System.Buffers.Primitives.
- Moved span extensions that are being used to System.Buffers.Primitives.
- Moved anything else in System.Slices into System.Buffers.Experimental for now.

@KrzysztofCwalina @ahsonkhan


ahsonkhan:
> new trimmed solution file.

It is hard to discern the diff between this new sln file and the old one. Can you rename it in a separate commit or some other workaround so we can see the diff first?

shiftylogic:
This is a "new" file. The old one is still there. This just makes it so you don't have to load and build everything whenever you want to work on the production stuff.

scripts/build.ps1
```diff
@@ -22,7 +22,7 @@ $dotnetExePath="$PSScriptRoot\..\dotnet\dotnet.exe"
 
 if ($Restore -eq "true") {
     Write-Host "Restoring all packages"
-    Invoke-Expression "$dotnetExePath restore"
+    Invoke-Expression "$dotnetExePath restore corefxlab.sln"
```
ahsonkhan: Is this necessary now because we have two sln files?

shiftylogic: yes

ahsonkhan:
You are missing a change in the PerfHarness.cs
https://github.com/dotnet/corefxlab/blob/c1a0338149df653b50948a7ef0c1ff32f150e829/scripts/PerfHarness/PerfHarness.cs#L52

Please run the benchmarks as well as part of the testing for this change.

ahsonkhan:
LGTM

KrzysztofCwalina:
Assuming there are no changes to the large green files that seem to be just moved, it looks good.

ahsonkhan:
Part of https://github.com/dotnet/corefxlab/issues/1314

PR #1320, opened 2017-03-16
Renaming IndexOfVectorized to IndexOf

ahsonkhan:
- Bringing original IndexOf from corefx and naming it ~IndexOfNon_Vectorized~ SequentialIndexOf (for comparison).
- The default implementation of IndexOf is now vectorized (https://github.com/dotnet/corefx/pull/17143)

@KrzysztofCwalina, is the method name fine?
cc @davidfowl


src/System.Buffers.Primitives/System/SpanExtensions.cs
```diff
@@ -38,124 +38,122 @@ public static bool StartsWith<T>(this ReadOnlySpan<T> items, ReadOnlySpan<T> sli
             return items.Slice(0, slice.Length).SequenceEqual(slice);
         }
 
-        public unsafe static int IndexOfVectorized(this Span<byte> buffer, byte value)
+        public static int IndexOfNon_Vectorized(this Span<byte> span, byte value)
```
jkotas: It should be without underscore, I think

benaadams:
Is conveying implementation detail? Maybe
`IndexOfFastStart`

KrzysztofCwalina:
What about calling it IndexOfSequential? or SequentialIndexOf.

src/System.Buffers.Primitives/System/SpanExtensions.cs
```diff
@@ -38,124 +38,122 @@ public static bool StartsWith<T>(this ReadOnlySpan<T> items, ReadOnlySpan<T> sli
...
-                    }
-                }
-                // Not found
-                offset = -1;
-                shortExit:
-                return offset;
+                length -= 8;
+
+                if (value == Unsafe.Add(ref searchSpace, index))
+                    goto Found;
```
KrzysztofCwalina: Should this be called Found0 for consistency?

src/System.Buffers.Primitives/System/SpanExtensions.cs
```diff
@@ -38,124 +38,122 @@ public static bool StartsWith<T>(this ReadOnlySpan<T> items, ReadOnlySpan<T> sli
...
-        private static unsafe int IndexOfVectorized(byte* searchSpace, int length, byte value)
+        private static unsafe int SequentialIndexOf(ref byte searchSpace, byte value, int length)
         {
-            var offset = 0;
-            // If length < vector length the jump over Vector dominates the search; as the Vector section is quite chunky
-            // So do an early search and exit
-            if (length < Vector<byte>.Count)
+            Debug.Assert(length >= 0);
+
+            IntPtr index = (IntPtr)0; // Use IntPtr for arithmetic to avoid unnecessary 64->32->64 truncations
```
KrzysztofCwalina: I would use IntPtr.Zero

KrzysztofCwalina:
looks good.

src/System.Buffers.Primitives/System/SpanExtensions.cs
ahsonkhan: Sure, but I grabbed the code from corefx and it has goto Found in many places:

https://github.com/dotnet/corefx/blob/482335d774791c7a43a157300242f73442ffcad3/src/System.Memory/src/System/SpanHelpers.T.cs#L59

https://github.com/dotnet/corefx/blob/482335d774791c7a43a157300242f73442ffcad3/src/System.Memory/src/System/SpanHelpers.byte.cs#L61

Should I fix this in both repos?

src/System.Buffers.Primitives/System/SpanExtensions.cs
ahsonkhan: Similar to above, should this be fixed everywhere?

https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/SpanHelpers.T.cs#L53
https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/SpanHelpers.byte.cs#L55

src/System.Buffers.Primitives/System/SpanExtensions.cs
jkotas: I have intentionally named it this way: there is `index` and no `index + 0` here either; and Found0 looked odd in the last loop that goes by single element.

src/System.Buffers.Primitives/System/SpanExtensions.cs
jkotas: There are many hundreds instances of both forms in corefx. (My rule of thumb - if there are two equivalent forms, use the one that has fewer characters.)

benaadams: In https://github.com/dotnet/coreclr/pull/9473 I found `IntPtr.Zero` to generally be worse as it does a memory load as a `readonly static` struct rather than a conversion that is then elided. 

Is noted as the reason for the [existence of `internal IsNull()` ](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/src/System/IntPtr.cs#L32-L37)

jkotas: `IntPtr.Zero` is treated as intrinsic by the JIT and should be JITed as zero constant. Look for `CORINFO_FIELD_INTRINSIC_ZERO` with optimizations on. If you do not see it happening, it is a bug that should be fixed.

The comment on IsNull predates the `CORINFO_FIELD_INTRINSIC_ZERO` optimization.

src/System.Buffers.Primitives/System/SpanExtensions.cs
KrzysztofCwalina: ok. sounds good

src/System.Buffers.Primitives/System/SpanExtensions.cs
KrzysztofCwalina: Seems like we should do the cleanup separately (if we even want).

benaadams: Will have a look at it again and raise issue if I come across it again (I think it was less happy to inline using `IntPtr.Zero`)

PR #1321, opened 2017-03-16
Replace TryGetPointer with MemoryHandle in tests

pdeligia:
This change removes `TryGetPointer` from the System.IO.Pipelines.Tests, and replaces it with `MemoryHandle` and `Span.DangerousGetPinnableReference`.

There is now only a single place left that is using `TryGetPointer` (in RioTcpConnection).

### DIFF
--- a/tests/System.Buffers.Primitives.Tests/MemoryTests.cs
+++ b/tests/System.Buffers.Primitives.Tests/MemoryTests.cs
@@ -235,6 +235,72 @@ namespace System.Buffers.Tests
             reservation.Dispose();
             Assert.Equal(true, owned.IsDisposed);
         }
+
+        [Fact]
+        public void PinAddReferenceReleaseTest()
+        {
+            var array = new byte[1024];
+            OwnedArray<byte> owned = array;
+            var memory = owned.Memory;
+            Assert.False(owned.HasOutstandingReferences);
+            var h = memory.Pin();
+            Assert.True(owned.HasOutstandingReferences);
+            h.Free();
+            Assert.False(owned.HasOutstandingReferences);
+        }
+
+        [Fact]
+        public void MemoryHandleFreeUninitialized()
+        {
+            var h = default(MemoryHandle);
+            h.Free();
+        }
+
+        [Fact]
+        public void MemoryHandleDoubleFree() 
+        {
+            var array = new byte[1024];
+            OwnedArray<byte> owned = array;
+            var memory = owned.Memory;
+            var h = memory.Pin();
+            Assert.True(owned.HasOutstandingReferences);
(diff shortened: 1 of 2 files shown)


tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
```diff
@@ -212,6 +212,8 @@ private void EqualsDetectsDeltaForAllLocations(ReadableBuffer slice, byte[] expe
             }
         }
 
+        private static List<MemoryHandle> _handles = new List<MemoryHandle>();
```
davidfowl: Locals for tests please

mjp41:
I have expanded this to add tests for the idempotent free and other aspects of Pin, that were not tested previously.

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
```diff
@@ -235,6 +235,94 @@ public void AutoDispose()
...
+            h.Free();
+            Assert.True(owned.HasOutstandingReferences);
+            h.Free();
+            Assert.True(owned.HasOutstandingReferences);
+            owned.Release();
+            Assert.False(owned.HasOutstandingReferences);
+        }
+
+
+        WeakReference LeakHandle()
```
davidfowl: @jkotas is this sort of code reliable?

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
```diff
@@ -235,6 +235,94 @@ public void AutoDispose()
...
+
+        void DoGC()
+        {
+            GC.Collect(2);
+            GC.WaitForPendingFinalizers();
+            GC.Collect(2);
+        }
+
+        [Fact]
+        void PinGCArrayTest1()
```
davidfowl: @jkotas is this sort of code reliable?

jkotas: Checking that something was not collected (PinGCArrayTest1) is reliable.

Checking that something got collected (PinGCArrayTest2) tends to have reliability issues. The object lifetime can be extended for variety of reasons: the codegen can decide to inline LeakHandle and extend the GC reporting lifetime for the locals; or conservative GC reporting can extend lifetime of arbitrary objects (conservative reporting its not used in CoreCLR today but it may change ... it is used in CoreRT in certain situations).

davidfowl: Yea, I think I'll get rid of that test.

/cc @mjp41

mjp41:
Dropped the test that depended on GC.Collect collecting.

PR #1323, opened 2017-03-16
Removing experimental dependency from pipelines

shiftylogic:
- Moved BufferPool to Buffer primitives.
- Removed experimental assembly dependency from pipelines.


src/System.IO.Pipelines/System.IO.Pipelines.csproj
```diff
@@ -9,7 +9,7 @@
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\System.Buffers.Primitives\System.Buffers.Primitives.csproj" />
-    <ProjectReference Include="..\System.Buffers.Experimental\System.Buffers.Experimental.csproj" />
+    <ProjectReference Include="..\System.Collections.Sequences\System.Collections.Sequences.csproj" />
```
ahsonkhan: What is the purpose of this change?

shiftylogic: Buffers.Experimental indirectly pulled in the Collections.Sequences project. Pipelines depended on both. We are breaking the dependency on Experimental, so I had to add an explicit dependency on Collections.Sequences.

shiftylogic:
@davidfowl @KrzysztofCwalina

src/System.Buffers.Primitives/System/Buffers/OwnedPinnedArray.cs
```diff
@@ -6,37 +6,6 @@
 
 namespace System.Buffers
 {
-    public class OwnedNativeMemory : OwnedMemory<byte>
```
davidfowl: I actually think these implementations could stay in Experimental. /cc @KrzysztofCwalina ?

shiftylogic: I misunderstood a comment you made when we were chatting. Thought you needed those types. I'll move them back.

KrzysztofCwalina:
looks good.

PR #1332, opened 2017-03-17
Add cancellation token support to FlushAsync and ReadAsync

pakrym:
@halter73 @davidfowl

### DIFF
--- a/src/System.IO.Pipelines/PipeAwaitable.cs
+++ b/src/System.IO.Pipelines/PipeAwaitable.cs
@@ -14,11 +14,30 @@ namespace System.IO.Pipelines
 
         private CancelledState _cancelledState;
         private Action _state;
+        private CancellationToken _cancellationToken;
+        private CancellationTokenRegistration _cancellationTokenRegistration;
 
         public PipeAwaitable(bool completed)
         {
             _cancelledState = CancelledState.NotCancelled;
             _state = completed ? _awaitableIsCompleted : _awaitableIsNotCompleted;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public CancellationTokenRegistration AttachToken(CancellationToken cancellationToken, Action<object> callback, object state)
+        {
+            CancellationTokenRegistration oldRegistration;
+            if (!cancellationToken.Equals(_cancellationToken))
+            {
+                oldRegistration = _cancellationTokenRegistration;
+                _cancellationToken = cancellationToken;
+                if (_cancellationToken.CanBeCanceled)
+                {
+                    _cancellationToken.ThrowIfCancellationRequested();
+                    _cancellationTokenRegistration = _cancellationToken.Register(callback, state);
+                }
+            }
+            return oldRegistration;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
@@ -109,6 +128,8 @@ namespace System.IO.Pipelines
                 {
                     Reset();
                 }
+
+                _cancellationToken.ThrowIfCancellationRequested();
 
                 return true;
             }
--- a/tests/System.IO.Pipelines.Tests/FlushAsyncCancellationTests.cs
+++ b/tests/System.IO.Pipelines.Tests/FlushAsyncCancellationTests.cs
@@ -2,12 +2,113 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Threading;
+using System.Threading.Tasks;
 using Xunit;
 
 namespace System.IO.Pipelines.Tests
 {
     public class FlushAsyncCancellationTests : PipeTest
     {
+        [Fact]
+        public void GetResultThrowsIfFlushAsyncCancelledAfterOnCompleted()
+        {
+            var onCompletedCalled = false;
+            var cancellationTokenSource = new CancellationTokenSource();
+            var buffer = Pipe.Writer.Alloc(PipeTest.MaximumSizeHigh);
+            buffer.Advance(PipeTest.MaximumSizeHigh);
+
+            var awaiter = buffer.FlushAsync(cancellationTokenSource.Token);
+
+            awaiter.OnCompleted(() =>
+            {
+                onCompletedCalled = true;
+                Assert.Throws<OperationCanceledException>(() => awaiter.GetResult());
+            });
+
+            var awaiterIsCompleted = awaiter.IsCompleted;
+
+            cancellationTokenSource.Cancel();
+
+            Assert.False(awaiterIsCompleted);
+            Assert.True(onCompletedCalled);
+        }
+
+        [Fact]
+        public void GetResultThrowsIfFlushAsyncCancelledBeforeOnCompleted()
+        {
+            var onCompletedCalled = false;
+            var cancellationTokenSource = new CancellationTokenSource();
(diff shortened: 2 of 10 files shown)


src/System.IO.Pipelines/PipeAwaitable.cs
```diff
@@ -10,16 +10,34 @@ internal struct PipeAwaitable
...
         }
 
+
+        public void AttachToken(CancellationToken cancellationToken, Action<object> callback, object state)
+        {
+            if (cancellationToken != _cancellationToken)
+            {
+                _cancellationTokenRegistration.Dispose();
+                _cancellationToken = cancellationToken;
+                if (_cancellationToken != CancellationToken.None)
```
davidfowl: _cancellationToken.CanBeCancelled

tests/System.IO.Pipelines.Tests/PipeCancellationTokenTests.cs
```diff
@@ -0,0 +1,162 @@
...
+                MaximumSizeHigh = 65,
+                MaximumSizeLow = 6
+            });
+        }
+
+        public void Dispose()
+        {
+            _pipe.Writer.Complete();
+            _pipe.Reader.Complete();
+            _pipeFactory?.Dispose();
```
davidfowl: Why ?.Dispose

src/System.IO.Pipelines/PipeAwaitable.cs
```diff
@@ -10,16 +10,34 @@ internal struct PipeAwaitable
...
             _cancelledState = CancelledState.NotCancelled;
             _state = completed ? _awaitableIsCompleted : _awaitableIsNotCompleted;
         }
 
+
+        public void AttachToken(CancellationToken cancellationToken, Action<object> callback, object state)
+        {
+            if (cancellationToken != _cancellationToken)
+            {
+                _cancellationTokenRegistration.Dispose();
```
davidfowl: Any chance on deadlocks here? i.e. Waiting for completion callbacks to run.

src/System.IO.Pipelines/PipeAwaitable.cs
```diff
@@ -52,12 +70,14 @@ public void Reset()
 
         public Action OnCompleted(Action continuation, ref PipeCompletion completion)
         {
+            _cancellationToken.ThrowIfCancellationRequested();
```
davidfowl: We shouldn't throw here, just in ObserveCancellation

src/System.IO.Pipelines/PipeAwaitable.cs
```diff
@@ -83,6 +103,8 @@ public Action Cancel()
 
         public bool ObserveCancelation()
         {
+            _cancellationToken.ThrowIfCancellationRequested();
```
davidfowl: Should this throw after changing the `_cancelledState`?

tests/System.IO.Pipelines.Tests/PipeCancellationTokenTests.cs
```diff
@@ -0,0 +1,162 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.IO.Pipelines.Tests
+{
+    public class PipeCancellationTokenTests
```
davidfowl: You need tests where you interleave cancellation tokens and `CancelPendingRead` calls.

src/System.IO.Pipelines/PipeAwaitable.cs
pakrym: We need to throw here so `GetResult` throws

src/System.IO.Pipelines/PipeAwaitable.cs
pakrym: We need it here so `GetResult` throws if token got cancelled after `OnComplete`

tests/System.IO.Pipelines.Tests/FlushAsyncCancellationTests.cs
```diff
@@ -0,0 +1,140 @@
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace System.IO.Pipelines.Tests
+{
+    public class FlushAsyncCancellationTests: PipeTest
+    {
+
```
davidfowl: nit: space

davidfowl:
@pakrym I don't see a test that catches the cancellation token on await

pakrym:
@davidfowl to test what?

davidfowl:
@pakrym actual usage. You're always manually calling OnCompleted.

pakrym:
I will add one

src/System.IO.Pipelines/PipeAwaitable.cs
```diff
@@ -10,16 +10,34 @@ internal struct PipeAwaitable
...
+
+        public void AttachToken(CancellationToken cancellationToken, Action<object> callback, object state)
+        {
+            if (cancellationToken != _cancellationToken)
+            {
+                _cancellationTokenRegistration.Dispose();
+                _cancellationToken = cancellationToken;
+                if (_cancellationToken.CanBeCanceled)
+                {
+                    _cancellationToken.ThrowIfCancellationRequested();
```
davidfowl: Do we need this? or this? The calling code can do this right?

pakrym: This one is so Flush/ReadAsync would throw immediately if passed cancelled token

src/System.IO.Pipelines/PipeAwaitable.cs
halter73: This will deadlock if:

- Dispose is called while the token is being canceled
- AND the registered callback calls a Pipe method the attempts to acquire the _sync lock.

This seems like a realistic enough scenario to warrant using [SafeRegister](https://github.com/aspnet/KestrelHttpServer/blob/rel/1.1.1/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Infrastructure/CancellationTokenExtensions.cs#L11).

pakrym:
Perf dropped
```
                  Method |        Mean |    StdErr |     StdDev |          RPS |
------------------------ |------------ |---------- |----------- |------------- |
 ParseLiveAspNetTwoTasks | 276.2471 ns | 2.7874 ns | 15.2672 ns | 3,619,947.20 |
   ParseLiveAspNetInline | 199.4010 ns | 4.9525 ns | 27.1262 ns | 5,015,020.70 |

                  Method |        Mean |    StdErr |     StdDev |          RPS |
------------------------ |------------ |---------- |----------- |------------- |
 ParseLiveAspNetTwoTasks | 390.1697 ns | 4.3209 ns | 23.6664 ns | 2,562,987.42 |
   ParseLiveAspNetInline | 276.2824 ns | 4.7347 ns | 25.9332 ns | 3,619,484.60 |
```

pakrym:
After inlining AttachToken:

```
                  Method |        Mean |    StdErr |     StdDev |          RPS |
------------------------ |------------ |---------- |----------- |------------- |
 ParseLiveAspNetTwoTasks | 357.9763 ns | 5.7727 ns | 31.6186 ns | 2,793,480.68 |
   ParseLiveAspNetInline | 231.4328 ns | 2.4783 ns | 13.5743 ns | 4,320,908.14 |
```

pakrym:
@davidfowl merging this.

PR #1333, opened 2017-03-18
Cleaning up tests

ahsonkhan:
- Removing System.Text.Primitives.Performance.Tests.csproj
- Removing S.T.P.Performance.Tests from InternalsInvisibleTo in AssemblyInfo.

Essentially reverted this change (https://github.com/dotnet/corefxlab/pull/1169) since this issue (https://github.com/dotnet/corefxlab/issues/1155) has been fixed.

- During the test runs in VS, I was getting the warning below so I reduced the names of different strings in the test to fit under 449:
`Contents of string 'System.Text.Utf8.Tests.Utf8EncoderTests.EncodeAllUnicodeCodePoints(encoder: Utf16TextEncoderLE { Encoding = Utf16, IsInvariantUtf16 = True, IsInvariantUtf8 = False }, systemTextEncoder: UnicodeEncoding { BodyName = "utf-16", CodePage = 1200, DecoderFallback = DecoderReplacementFallback { DefaultString = "�", MaxCharCount = 1 }, EncoderFallback = EncoderReplacementFallback { DefaultString = "�", MaxCharCount = 1 }, EncodingName = "Unicode", ... })' exceeds max of '449', string has been truncated.`

- Some of the tests were duplicates and so I removed them (i.e. removed duplicate InlineData sets) to avoid the following warnings in the test discovery:
`A test with the same name 'System.Text.Utf8.Tests.Utf8StringTests.IndexOfTests (528eaefb89640cbfcc9dcbed4204971fd8a25bf7)' already exists. This test is not added to the test window.`
11 such duplicate tests were removed so now:
`========== Discover test finished: 439 found (0:00:04.1816092) ==========` matches what we see in the test window (which was 428 and not 439).

- Added try/catch block around `Marshal.AllocHGlobal(new IntPtr(ThreeGiB))` to catch OutOfMemoryExceptions and not have tests fail locally when allocation fails.


KrzysztofCwalina:
looks good

PR #1337, opened 2017-03-19
Move classes out of the same file

davidfowl:
- Added OwnedNativeMemory and OwnedPinnedArray class files


src/System.Buffers.Experimental/System/Buffers/OwnedNativeMemory.cs
```diff
@@ -0,0 +1,36 @@
+﻿using System.Runtime.InteropServices;
```
JunTaoLuo: nit:copyright

PR #1346, opened 2017-03-20
Cleanup dependencies and file names

### LINKED ISSUE #1345: Remove dependency from System.Text.Primitives to System.Buffers.Primitives
System.Text.Primitives should depend on System.Memory only.

Today, there are two bad dependencies that need to be removed:
1. Extension method that copies bytes from array to Span\<T\>
2. Extension Slice(byte[] ...)

The extensions should be moved to System.Memory or the code should be changed.

KrzysztofCwalina:
The filenames are renamed because of a PR that change the typenames.

This fixes #1345


shiftylogic:
👍

PR #1347, opened 2017-03-20
Throw if advancing past the Memory size

pakrym:
@davidfowl

### DIFF
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -298,7 +298,7 @@ namespace System.IO.Pipelines
 
                 if (_writingHead == null)
                 {
-                    throw new InvalidOperationException("Cannot advance without buffer allocated");
+                    ThrowHelper.ThrowInvalidOperationException(ExceptionResource.AdvancingWithNoBuffer);
                 }
 
                 Debug.Assert(_writingHead != null);
@@ -310,7 +310,7 @@ namespace System.IO.Pipelines
 
                 if (bufferIndex > buffer.Length)
                 {
-                    throw new InvalidOperationException("Cannot advance past memory size");
+                    ThrowHelper.ThrowInvalidOperationException(ExceptionResource.AdvancingPastBufferSize);
                 }
 
                 _writingHead.End = bufferIndex;
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -293,9 +293,14 @@ namespace System.IO.Pipelines
         internal void AdvanceWriter(int bytesWritten)
         {
             EnsureAlloc();
-
             if (bytesWritten > 0)
             {
+
+                if (_writingHead == null)
+                {
+                    throw new InvalidOperationException("Cannot advance without buffer allocated");
+                }
+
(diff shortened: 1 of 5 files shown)


src/System.IO.Pipelines/Pipe.cs
```diff
@@ -293,17 +293,25 @@ internal void CommitUnsynchronized()
...
                 Debug.Assert(!_writingHead.ReadOnly);
                 Debug.Assert(_writingHead.Next == null);
 
                 var buffer = _writingHead.Buffer;
                 var bufferIndex = _writingHead.End + bytesWritten;
 
-                Debug.Assert(bufferIndex <= buffer.Length);
+                if (bufferIndex > buffer.Length)
+                {
+                    throw new InvalidOperationException("Cannot advance past memory size");
```
benaadams: Very hot function `ThrowHelper.Throw...`?

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -293,17 +293,25 @@ internal void CommitUnsynchronized()
         internal void AdvanceWriter(int bytesWritten)
         {
             EnsureAlloc();
-
             if (bytesWritten > 0)
             {
+
+                if (_writingHead == null)
+                {
+                    throw new InvalidOperationException("Cannot advance without buffer allocated");
```
benaadams: Very hot function `ThrowHelper.Throw...`?

src/System.IO.Pipelines/Pipe.cs
pakrym: Oh, what was I thinking :(

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -293,17 +293,25 @@ internal void CommitUnsynchronized()
         internal void AdvanceWriter(int bytesWritten)
         {
             EnsureAlloc();
-
             if (bytesWritten > 0)
             {
+
```
davidfowl: nit: space

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -293,17 +293,25 @@ internal void CommitUnsynchronized()
...
+                if (_writingHead == null)
+                {
+                    ThrowHelper.ThrowInvalidOperationException(ExceptionResource.AdvancingWithNoBuffer);
+                }
+
                 Debug.Assert(_writingHead != null);
```
davidfowl: Remove the assert

tests/System.IO.Pipelines.Tests/BackpressureTests.cs
```diff
@@ -170,5 +170,16 @@ public void FlushAsyncAwaitableResetsOnCommit()
 
             Assert.False(flushAsync.IsCompleted);
         }
+
+        [Fact]
+        public void ThrowsOnAllocDuringFlush()
```
davidfowl: I just realized this will make the stream code impossible to write... https://github.com/aspnet/KestrelHttpServer/blob/751a0e2e7e92fc2adc23092aeec5117a2afe5a1b/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/SocketOutput.cs#L83. We can't do this one.

Issue #1292, opened 2017-03-10
Finalize API shape of SpanExtensionsLab and merge with SpanExtensions


ahsonkhan:
Merge the Span Extensions from corefxlab into corefx:
corefx: https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/SpanExtensions.cs
corefxlab: https://github.com/dotnet/corefxlab/blob/master/src/System.Slices/System/SpanExtensions.cs

davidfowl:
Also https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs

ahsonkhan:
@davidfowl, can you open a separate issue for that? Thanks :)

davidfowl:
What would the issue be? Figure out where IndexOfVectorized should live?

ahsonkhan:
I am not sure. I am not familiar with System.Buffers.Experimental or which of the methods implemented there are used/useful. If IndexOfVectorized is the only one relevant to you, that could be the issue.

jkotas:
We should have just one IndexOf that is vectorized and that lives in System.Memory in corefx. The cut-off between non-vectorized and vectorized IndexOf performance is pretty low (IIRC, <10 bytes), and so having just one vectorized IndexOf is the right default choice. If somebody needs IndexOf hyper-optimized for just a few bytes, they can build their local specialized version of it.

What needs to happen for it is to get System.Memory building in corefx with reference to System.Vectors. @benaadams submitted PR for it a while ago, but got lost in the intricacies of the corefx build system. Let me know if you need help moving it forward.

davidfowl:
That would be ideal.

KrzysztofCwalina:
I like the default to be vectorized, but I would not remove the optimized one and ask people to implement it on their own. Why don't we rename IndexOfVectorized to IndexOf and then move the the non-vectorized IndexOf back to corfxlab buffers.experimental and call it something liek EarlyIndexOf

jkotas:
Sounds good to me. 

I do not plan to work on it, but it should be easy to do because of I have added the build plumbing for using SIMD in System.Memory in the corefx makefiles.

benaadams:
Will add it; had problem before because of the SIMD reference

benaadams:
Added PR for Vector path for IndexOf to corefx https://github.com/dotnet/corefx/pull/17143

ahsonkhan:
[...]

PR #1348, opened 2017-03-21
Partial cleanup of SpanExtensionsLab

ahsonkhan:
Working towards - https://github.com/dotnet/corefxlab/issues/1314 / https://github.com/dotnet/corefxlab/issues/1292

- Move to System.Binary
    - public static T Read<T>(this ReadOnlySpan<byte> slice) where T : struct;
    - public static T Read<T>(this Span<byte> slice) where T : struct;
    - public static bool TryRead<T>(this ReadOnlySpan<byte> slice, out T value) where T : struct;
    - public static bool TryRead<T>(this Span<byte> slice, out T value) where T : struct;
    - public static bool TryWrite<T>(this Span<byte> slice, T value) where T : struct;
    - public static void Write<T>(this Span<byte> slice, T value) where T : struct;

- Moved to SpanExtensions: (PR https://github.com/dotnet/corefx/pull/17322)
    - public static bool StartsWith(this ReadOnlySpan<byte> bytes, ReadOnlySpan<byte> slice);
    - public static bool StartsWith<T>(this ReadOnlySpan<T> items, ReadOnlySpan<T> slice) where T : struct, IEquatable<T>;

- Move to BuffersExtension in experimental
    - ~public static int IndexOf(this ReadOnlySpan<byte> buffer, byte value0, byte value1);~
    - ~public static int IndexOf(this ReadOnlySpan<byte> buffer, byte value0, byte value1, byte value2);~
    - ~public static int IndexOf(this Span<byte> buffer, byte value0, byte value1);~
    - ~public static int IndexOf(this Span<byte> buffer, byte value0, byte value1, byte value2);~
    - public static int SequentialIndexOf(this ReadOnlySpan<byte> span, byte value);
    - public static int SequentialIndexOf(this Span<byte> span, byte value);
    - public static bool TryIndicesOf(this ReadOnlySpan<byte> buffer, byte value, Span<int> indices, out int numberOfIndices);
    - public static bool TryIndicesOf(this Span<byte> buffer, byte value, Span<int> indices, out int numberOfIndices);

- Remove (this should be on ArraySegment)
    - public static ArraySegment<T> Slice<T>(this ArraySegment<T> source, int count);

- Delete:
    - public static bool StartsWith(this ReadOnlySpan<char> str, ReadOnlySpan<char> value);

**NOTE:**
Moving the following to `System.Buffers.Experimental\System\Buffers\BufferExtensions.cs` causes `System.IO.Pipelines` to take a dependency on `System.Buffers.Experimental`.
```C#
    - public static int IndexOf(this ReadOnlySpan<byte> buffer, byte value0, byte value1);
    - public static int IndexOf(this ReadOnlySpan<byte> buffer, byte value0, byte value1, byte value2);
    - public static int IndexOf(this Span<byte> buffer, byte value0, byte value1);
    - public static int IndexOf(this Span<byte> buffer, byte value0, byte value1, byte value2);
```
The IndexOf with multiple byte values are used in `ReadCursorOperations.Seek`, here:
https://github.com/aspnet/KestrelHttpServer/blob/f6e5e74d9551df57effda19c856352608374cee7/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/KestrelHttpParser.cs#L273

This function is implemented here:
https://github.com/dotnet/corefxlab/blob/3bd0ffdef64b4510b408200cf4281d24b5742abd/src/System.IO.Pipelines/ReadCursorOperations.cs#L32

What should we do about this dependency? This is the only that System.IO.Pipelines depends on Buffers.Experimental.
If we remove those Seek overloads (and the one reference to it in kestrel), we could remove this dependency.

**Edit:**
Adding back IndexOf overloads to SpanExtensionsLab (in Buffes.Primitives). Will remove them once they are ported over to corefx (in System.Memory.dll).


davidfowl:
> Moving the following to System.Buffers.Experimental\System\Buffers\BufferExtensions.cs causes System.IO.Pipelines to take a dependency on System.Buffers.Experimental.

I'm not a fan of this. Move those methods to corefx proper before breaking this package.

KrzysztofCwalina:
Yeah, if the APIs are used by pipelines, and we know they have benefits, we should just have them in SpanExtensions. BTW, I think it would be ok to move the APIs to SpanExtensions without proving they have benefits (perf) and measure later.

KrzysztofCwalina:
Looks good. Thanks!

KrzysztofCwalina:
Ah, I think I merged Ben's rename and now there is a conflict. Could you please resolve and then just merge? Thanks.

ahsonkhan:
@KrzysztofCwalina, resolved the conflicts.

PR #1358, opened 2017-03-22
Adding extension method Slice on string which returns ReadOnlySpan<char>

ahsonkhan:
Moving over from System.Memory.dll as part of #1314 / #1292.

Removed here: https://github.com/dotnet/corefx/pull/17407

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
@@ -23,6 +23,84 @@ namespace System.Buffers
             return array.Items.Slice(0, array.Count);
         }
 
+        /// <summary>
+        /// Creates a new readonly span over the portion of the target string.
+        /// </summary>
+        /// <param name="text">The target string.</param>
+        /// <exception cref="System.ArgumentNullException">Thrown when <paramref name="text"/> is null.</exception>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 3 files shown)


src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -23,6 +23,84 @@ public static class BufferExtensions
...
+        {
+            if (text == null)
+                throw new ArgumentNullException(nameof(text));
+
+            int textLength = text.Length;
+
+            if ((uint) start > (uint) textLength)
+                throw new ArgumentOutOfRangeException(nameof(start));
+
+            if (textLength == 0) return ReadOnlySpan<char>.Empty;
```
KrzysztofCwalina: Should this check be done after subtracting start?

ahsonkhan: Yes. I would imagine this would be for performance reasons and not for correctness, right?
I ask this because if (textLength - start) == 0, we just pass in 0 length to the DangerousCreate method.

ahsonkhan:
Are we waiting for Slice methods to be removed from System.Memory.dll (https://github.com/dotnet/corefx/pull/17407) before merging this?

ahsonkhan:
cc @KrzysztofCwalina, can this be merged?

ahsonkhan:
cc @KrzysztofCwalina, should this be merged?

Do we still want slice on string or is the usage where we use AsSpan() fine?

Issue #1355, opened 2017-03-22
Complete writer based on examined instead of consumed?


tmds:
I have created a Pipe setting MaximumSize{Low,Heigh} to 1 to minimize the buffering.

When I read the data I want to do something after I have sufficient bytes:
```
while (true)
{
    var input = await reader.ReadAsync();
    // check completed, etc.
    if (input.Length < MinExpected)
    {
        reader.Advance(consumed: input.Start, examined: input.End);
        continue;
    }
    // use data
    reader.Advance(consumed: input.End, examined: input.End);
}
```
When the read returns less than MinExpected, the code deadlocks. Currently the completion of the writer is [based on the consumed data](https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines/Pipe.cs#L402-L409). This would work if it was based on the examined data.

davidfowl:
> When the read returns less than MinExpected, the code deadlocks. Currently the completion of the writer is based on the consumed data. 

What do you mean the code "dead locks" you just mean the awaitable isn't complete right?

tmds:
> What do you mean the code "dead locks" you just mean the awaitable isn't complete right?

The reader is waiting on ReadAsync.
The writer is waiting on FlushAsync.

tmds:
Repro:
```
using System.IO.Pipelines;
using System.Threading.Tasks;

namespace ConsoleApplication
{
    public class Program
    {
        public static void Main(string[] args)
        {
            var pipeFactory = new PipeFactory();
            var pipe = pipeFactory.Create(new PipeOptions() { MaximumSizeHigh = 1, MaximumSizeLow = 1 });
            Task.WaitAll(new[] { WriteToPipe(pipe.Writer), ReadFromPipe(pipe.Reader) });
        }

        public static async Task WriteToPipe(IPipeWriter writer)
        {
            while (true)
            {
                var buffer = writer.Alloc(1);
                buffer.Ensure(1);
                buffer.Advance(1);
                System.Console.WriteLine("Awaiting Flush");
                await buffer.FlushAsync();
                System.Console.WriteLine("-> Flush returns");
            }
        }

        public static async Task ReadFromPipe(IPipeReader reader)
        {
            while (true)
            {
                System.Console.WriteLine("Awaiting Read");
                var rb = await reader.ReadAsync();
                var buffer = rb.Buffer;
                System.Console.WriteLine("-> Read returns");
                if (buffer.Length < 5)
                {
[...]

PR #1359, opened 2017-03-23
Pipe: complete writerAwaitable based on examined instead of consumed bytes

tmds:
Implements https://github.com/dotnet/corefxlab/issues/1355

### DIFF
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -390,10 +390,20 @@ namespace System.IO.Pipelines
             BufferSegment returnStart = null;
             BufferSegment returnEnd = null;
 
+            int examinedBytes = 0;
+            if (!examined.IsDefault)
+            {
+                examinedBytes = new ReadCursor(_readHead).GetLength(examined);
+            }
+
             int consumedBytes = 0;
             if (!consumed.IsDefault)
             {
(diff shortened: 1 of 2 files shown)


davidfowl:
Few things:
- I'm not sure we want this.
- This is an extremely hot path and this it looks like we're computing the length for both consumed and examined every time. Did you run the micro benchmarks to see how this affects throughput?

tmds:
> I'm not sure we want this.
> This is an extremely hot path and this it looks like we're computing the length for both consumed and examined every time. Did you run the micro benchmarks to see how this affects throughput?
> You don't have any tests.

Let's figure out first if it is needed.

tmds:
> This is an extremely hot path and this it looks like we're computing the length for both consumed and examined every time.

I've optimized it a bit already, taking into account that examined >= consumed. Also consumed and examined will be in the same segment most of the time so `consumed.GetLength(examined)` will be in the inlined case.

davidfowl:
> I've optimized it a bit already, taking into account that examined >= consumed. Also consumed and examined will be in the same segment most of the time so consumed.GetLength(examined) will be in the inlined case.

Where are the before and after benchmarks? And the unit tests.

ahsonkhan:
> Where are the before and after benchmarks? And the unit tests.

@tmds, please update the PR.

davidfowl:
We decided not to take this

PR #1364, opened 2017-03-24
Add WritableBufferWriter for fast writes

pakrym:
Comparison with writefast that kestrel uses:

```
                                          Method |        Mean |    StdErr |     StdDev |          RPS |
------------------------------------------------ |------------ |---------- |----------- |------------- |
                      WriteFastPlaintextResponse | 519.8520 ns | 6.5837 ns | 36.0603 ns | 1,923,624.52 |
 WriteableBufferWriterWriteFastPlaintextResponse | 374.1911 ns | 6.0379 ns | 33.0709 ns | 2,672,431.50 |
```
/cc @benaadams

### DIFF
--- a/src/System.IO.Pipelines/WritableBuffer.cs
+++ b/src/System.IO.Pipelines/WritableBuffer.cs
@@ -10,17 +10,17 @@ namespace System.IO.Pipelines
     /// </summary>
     public struct WritableBuffer : IOutput
     {
-        private Pipe _pipe;
+        internal Pipe Pipe;
 
         internal WritableBuffer(Pipe pipe)
         {
-            _pipe = pipe;
+            Pipe = pipe;
         }
 
         /// <summary>
         /// Available memory.
         /// </summary>
-        public Buffer<byte> Buffer => _pipe.Buffer;
+        public Buffer<byte> Buffer => Pipe.Buffer;
 
         /// <summary>
         /// Returns the number of bytes currently written and uncommitted.
@@ -46,7 +46,7 @@ namespace System.IO.Pipelines
         /// </summary>
         public ReadableBuffer AsReadableBuffer()
         {
-            return _pipe.AsReadableBuffer();
+            return Pipe.AsReadableBuffer();
         }
 
         /// <summary>
@@ -55,14 +55,14 @@ namespace System.IO.Pipelines
         /// </summary>
         /// <param name="count">number of bytes</param>
         /// <remarks>
-        /// Used when writing to <see cref="Buffer"/> directly. 
+        /// Used when writing to <see cref="Buffer"/> directly.
         /// </remarks>
         /// <exception cref="ArgumentOutOfRangeException">
         /// More requested than underlying <see cref="IBufferPool"/> can allocate in a contiguous block.
         /// </exception>
         public void Ensure(int count = 1)
         {
-            _pipe.Ensure(count);
+            Pipe.Ensure(count);
         }
 
         /// <summary>
@@ -71,7 +71,7 @@ namespace System.IO.Pipelines
         /// <param name="buffer">The <see cref="ReadableBuffer"/> to append</param>
         public void Append(ReadableBuffer buffer)
         {
-            _pipe.Append(buffer);
+            Pipe.Append(buffer);
         }
 
         /// <summary>
@@ -83,7 +83,7 @@ namespace System.IO.Pipelines
         /// <exception cref="ArgumentOutOfRangeException"><paramref name="bytesWritten"/> is negative.</exception>
         public void Advance(int bytesWritten)
         {
-            _pipe.AdvanceWriter(bytesWritten);
+            Pipe.AdvanceWriter(bytesWritten);
         }
 
         /// <summary>
@@ -95,7 +95,7 @@ namespace System.IO.Pipelines
         /// </remarks>
         public void Commit()
         {
-            _pipe.Commit();
+            Pipe.Commit();
         }
 
         /// <summary>
@@ -105,7 +105,7 @@ namespace System.IO.Pipelines
         /// <returns>A task that completes when the data is fully flushed.</returns>
         public WritableBufferAwaitable FlushAsync()
         {
-            return _pipe.FlushAsync();
+            return Pipe.FlushAsync();
         }
     }
 }
--- a/src/System.IO.Pipelines/WritableBuffer.cs
+++ b/src/System.IO.Pipelines/WritableBuffer.cs
@@ -10,17 +10,17 @@ namespace System.IO.Pipelines
     /// </summary>
     public struct WritableBuffer : IOutput
     {
-        internal Pipe Pipe;
+        private Pipe _pipe;
 
         internal WritableBuffer(Pipe pipe)
         {
-            Pipe = pipe;
+            _pipe = pipe;
         }
 
         /// <summary>
         /// Available memory.
         /// </summary>
-        public Buffer<byte> Buffer => Pipe.Buffer;
+        public Buffer<byte> Buffer => _pipe.Buffer;
 
         /// <summary>
         /// Returns the number of bytes currently written and uncommitted.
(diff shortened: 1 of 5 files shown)


tests/System.IO.Pipelines.Tests/PipeLengthTests.cs
```diff
@@ -106,7 +102,7 @@ public void LengthNotChangeAfterReadAdvanceExamine()
         [Fact]
         public void ByteByByteTest()
         {
-            WritableBuffer writableBuffer;
+            WritableBuffer writableBuffer = default(WritableBuffer);
```
pakrym: Compiler just randomly started to error because of uninitialized variable on this line.

src/System.IO.Pipelines/WritableBufferWriter.cs
```diff
@@ -0,0 +1,82 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.IO.Pipelines
+{
+    public struct WritableBufferWriter
```
benaadams: `BufferWriter` 

`Writable` bit is a little superfluous as its already in the name

tests/System.IO.Pipelines.Tests/WritableBufferWriterFacts.cs
```diff
@@ -0,0 +1,109 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.Linq;
+using Xunit;
+
+namespace System.IO.Pipelines.Tests
+{
+    public class WritableBufferWriterFacts: IDisposable
```
benaadams: With earlier rename `BufferWriterFacts` etc

src/System.IO.Pipelines/WritableBufferWriter.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        public void Write(byte[] source)
+        {
+            Write(source, 0, source.Length);
+        }
+
+        public void Write(byte[] source, int offset, int length)
+        {
+            if (length <= _span.Length && length > 0)
```
benaadams: Single test it
```
(uint)length <= (uint)_span.Length
```
If length is less than zero then the `uint` version of it will be greater than `(uint)int.Max`

src/System.IO.Pipelines/WritableBufferWriter.cs
```diff
@@ -0,0 +1,82 @@
...
+                Advance(length);
+                return;
+            }
+
+            WriteMultiBuffer(source, offset, length);
+        }
+
+        private void WriteMultiBuffer(byte[] source, int offset, int length)
+        {
+            var remaining = length;
```
benaadams: Might not make much difference (not common path) and would need to look as asm; but wonder if arranging it like
```csharp
private void WriteMultiBuffer(byte[] source, int offset, int length)
{
    var finalOffset = offset + length;

    while (offset < finalOffset)
    {
        if (_span.Length == 0)
        {
            Ensure();
        }

        var writable = Math.Min(finalOffset - offset, _span.Length);

        Unsafe.CopyBlockUnaligned(ref _span.DangerousGetPinnableReference(), ref source[offset], (uint)writable);

        offset += writable;
        
        Advance(writable);
    }
}
```

Would do anything

benaadams: One less read + write per loop (`remaining`)
Update `offset` before calling `Advance` so its result is ready earlier for next loop check (i.e. its not used by anything later in loop)

benaadams:
Nice 👍

src/System.IO.Pipelines/WritableBufferWriter.cs
davidfowl: we shoul keep this name for now and rename all of the things together

src/System.IO.Pipelines/WritableBuffer.cs
```diff
@@ -10,17 +10,17 @@ namespace System.IO.Pipelines
     /// </summary>
     public struct WritableBuffer : IOutput
     {
-        private Pipe _pipe;
+        internal Pipe Pipe;
```
davidfowl: Revert this

src/System.IO.Pipelines/WritableBufferWriter.cs
```diff
@@ -0,0 +1,82 @@
...
+            _writableBuffer = writableBuffer;
+            _span = writableBuffer.Buffer.Span;
+        }
+
+        public Span<byte> Span => _span;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Advance(int count)
+        {
+            _span = _span.Slice(count);
```
davidfowl: Ugh I guess this will be fine on 2.0

tests/System.IO.Pipelines.Tests/WritableBufferWriterFacts.cs
davidfowl: nit space

src/System.IO.Pipelines/WritableBufferWriter.cs
```diff
@@ -0,0 +1,82 @@
...
+        public WritableBufferWriter(WritableBuffer writableBuffer)
+        {
+            _writableBuffer = writableBuffer;
+            _span = writableBuffer.Buffer.Span;
+        }
+
+        public Span<byte> Span => _span;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Advance(int count)
```
davidfowl: One major difference between this and the `ReadableBufferReader` is that advice doesn't support walking multiple blocks.

src/System.IO.Pipelines/WritableBufferWriter.cs
pakrym: Ugh, ok.

pakrym: Is double cast faster then 2 comparisons?

src/System.IO.Pipelines/WritableBufferWriter.cs
pakrym: Because it's semantics is different. It's not skip.

davidfowl: I plan to rename Skip to Advance for symmetry 😄

davidfowl:
I think you should copy some tests from here:

https://github.com/dotnet/corefxlab/blob/master/tests/System.IO.Pipelines.Tests/WritableBufferFacts.cs

src/System.IO.Pipelines/WritableBufferWriter.cs
pakrym: I would skip this for now because it's not on the hot path.

src/System.IO.Pipelines/WritableBufferWriter.cs
benaadams: Yes as it doesn't actually cast for ints etc; the asm just uses different instructions.

benaadams: Yes as it doesn't actually cast for ints etc; the asm just uses different instructions. 

So its one comparison with no conversion vs comparison + branch + comparison. Also with arrays the unsigned test will eliminate a second range check (if you were doing it against offset)

benaadams: Actually the test is completely wrong; as it only works for offset 0 😢 Should be

`if ((uint)offset > (uint)_span.Length || (uint)length > (uint)(_span.Length - offset))`

src/System.IO.Pipelines/WritableBufferWriter.cs
benaadams: Is like a Pen-Pineapple-Apple-Pen https://www.youtube.com/watch?v=kN5cqf63kb8

src/System.IO.Pipelines/WritableBufferWriter.cs
benaadams: Raised issue https://github.com/dotnet/corefxlab/issues/1366

PR #1368, opened 2017-03-24
Extracted HTTP Parser from Kestrel

KrzysztofCwalina:
I talked with @davidfowl about extracting the parser at some point, and the point has come.
@geoffkizer, once it merges, you can use it for your experiments. Also, feel free to submit patches.

The parser if taken as-is from Kestrel, with the only significant change being the removal of ASP.NET ILogger support. The logger is used in one place only (when bad request exception is thrown). I will try to clean the code in a subsequent PRs. 

@davidfowl, @geoffkizer, @ahsonkhan, @shiftylogic


src/System.Text.Http.Parser/BadHttpRequestException.cs
```diff
@@ -0,0 +1,193 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+using System.IO;
+
+namespace System.Text.Http.Parser
+{
+    internal static class StatusCodes
+    {
```
stephentoub: We seem to be missing some of these in https://github.com/dotnet/corefx/blob/master/src/System.Net.Primitives/src/System/Net/HttpStatusCode.cs#L8
Ideally we'd just add the missing values (https://github.com/dotnet/corefx/issues/4382) and then use that.

src/System.Text.Http.Parser/AsciiUtilities.cs
```diff
@@ -0,0 +1,79 @@
+﻿// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
```
stephentoub: Is this the right license to use?

src/System.Text.Http.Parser/HttpMethod.cs
```diff
@@ -0,0 +1,20 @@
+// Copyright (c) .NET Foundation. All rights reserved.
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace System.Text.Http.Parser
+{
+    public enum HttpMethod: byte
```
stephentoub: We should rationalize this with https://github.com/dotnet/corefx/blob/master/src/System.Net.Http/src/System/Net/Http/HttpMethod.cs#L9.

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -0,0 +1,524 @@
...
+                c == '\'' ||
+                c == '*' ||
+                c == '+' ||
+                c == '-' ||
+                c == '.' ||
+                c == '^' ||
+                c == '_' ||
+                c == '`' ||
+                c == '|' ||
+                c == '~';
```
stephentoub: I'm sure it's been looked at, but is this faster than a lookup table?  Maybe it is for the common-case and thus doing it this way is a win?

src/System.Text.Http.Parser/HttpUtilities.cs
```diff
@@ -0,0 +1,330 @@
...
+            methodNames[(byte)HttpMethod.Connect] = "CONNECT";
+            methodNames[(byte)HttpMethod.Options] = "OPTIONS";
+            return methodNames;
+        }
+
+        private unsafe static ulong GetAsciiStringAsLong(string str)
+        {
+            Debug.Assert(str.Length == 8, "String must be exactly 8 (ASCII) characters long.");
+
+            Span<byte> buffer = new byte[8];
```
stephentoub: Any reason not to use stackalloc here and below?

src/System.Text.Http.Parser/HttpVersion.cs
```diff
@@ -0,0 +1,12 @@
...
+// Licensed under the Apache License, Version 2.0. See License.txt in the project root for license information.
+
+namespace System.Text.Http.Parser
+{
+    public enum HttpVersion
+    {
+        Unknown = -1,
+        Http10 = 0,
+        Http11 = 1
+    }
```
stephentoub: Similarly good at some point to rationalize this with https://github.com/dotnet/corefx/blob/master/src/System.Net.Primitives/src/System/Net/HttpVersion.cs

src/System.Text.Http.Parser/PipelineExtensions.cs
```diff
@@ -0,0 +1,24 @@
...
+
+using System;
+using System.Buffers;
+using System.IO.Pipelines;
+using System.Runtime.CompilerServices;
+using System.Threading.Tasks;
+
+namespace System.Text.Http.Parser.Internal
+{
+    internal static class PipelineExtensions
```
stephentoub: Is this the right name?

benaadams:
Which is the primary source? (i.e. have a PR I need to rebase into the Parser should to go to corefxlab or Kestrel)?

src/System.Text.Http.Parser/BadHttpRequestException.cs
KrzysztofCwalina: @davidfowl, do you know why you are not using the System.Net enum? I assume because it did not exist when you guys wrote the code?

@stephentoub, are you going to add these values?

davidfowl:
We don't be using this parser in kestrel in the short term. We'll use it to experiment with other buffer primitives for parsing.

You should also port the perf tests.

src/System.Text.Http.Parser/AsciiUtilities.cs
KrzysztofCwalina: Good catch. I will update.

src/System.Text.Http.Parser/HttpUtilities.cs
davidfowl: It's only called once from a static field. I don't think it matters.

src/System.Text.Http.Parser/BadHttpRequestException.cs
stephentoub: > @stephentoub, are you going to add these values?

We should.  Obviously I'm not going to before this PR gets merged, but we should run it though the API process and get them added.

davidfowl: No it's internal. I dunno if we even have a System.Net dependency when we wrote it. It's just s bunch of ints so it doesn't matter anyways.

src/System.Text.Http.Parser/HttpParser.cs
davidfowl: Didn't measure a lookup table.

src/System.Text.Http.Parser/HttpUtilities.cs
KrzysztofCwalina: This method is called once to initialize statics, but I will add a TODO to optimize it.

src/System.Text.Http.Parser/HttpMethod.cs
KrzysztofCwalina: It shipped in Kestreal 1.0. https://github.com/aspnet/KestrelHttpServer/blob/d3694f085a2df0c51286fdc4369bca8012d308fc/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpMethod.cs

@davidfowl, what's your compat bar for APIs in internal namespaces?

davidfowl: 0 not that it matters as we don't expose these to user code. We call it pubternal.

src/System.Text.Http.Parser/HttpParser.cs
KrzysztofCwalina: This method is called for uncommon verbs (e.g. not for HTTP GET), so it does not matter much. But yes, the fastest way would probably be to create an array mapping from byte to bool

src/System.Text.Http.Parser/PipelineExtensions.cs
KrzysztofCwalina: For now it it. The parser today depends on ReadableBuffer, which is part of pipelines. I will try to move the parser to ReadOnlyBytes in a subsequent PRs, but I need perf tests first.

benaadams:
`HttpMethod` and `HttpVersion` are problematic to change to the corefx variants since they are full blown objects rather than enums so require both `ReferenceEquals` rather than simple equality and are ptrs so they can't be compressed together to reduce param passing

e.g. am looking at it for a [`HttpRequestLineParseInfo`](https://github.com/benaadams/KestrelHttpServer/blob/9a5ce325cc791b17b6418b590724d907a41d6647/src/Microsoft.AspNetCore.Server.Kestrel/Internal/Http/HttpRequestLineParseInfo.cs) struct

benaadams:
e.g. ReferenceEquals is (on pre-jitted readonly static; so no base .cctor init call)
```
00007FFF08A266FA  mov         rcx,1AEA0625858h  
00007FFF08A26704  mov         rcx,qword ptr [rcx]  
00007FFF08A26707  cmp         rcx,qword ptr [rsi+0A8h]  
00007FFF08A2670E  je          00007FFF08A267C6 
```
Enum compare is
```
00007FFF08A56923  cmp         dword ptr [rsi+1D4h],65h  
00007FFF08A5692A  je          00007FFF08A5697E  
```

So has an 2 additional loads (and pointer chase)

KrzysztofCwalina:
@benaadams, I would not use the System.Net classes. But we should rename these types to at least not conflict with naming.

benaadams:
>we should rename these types to at least not conflict with naming.

ParsedXxx? Don't think they are user exposed

KrzysztofCwalina:
Once this parser is public, the types (HttpVersion and HttpMethod) are exposed. They are parameters to IHttpRequestLineHandler

PR #1375, opened 2017-03-25
Remove TryGetPointer from RioTcpConnection

KodrAus:
Part of #1298

Moves the `GetSegmentFromMemory` method into the `RioThread` and uses `Pin` instead of `TryGetPointer`. I think it's safe to not keep the `MemoryHandle` around because we verify the memory is from a `Slab` in the memory pool using its address, which we know is pinned.

So now it's `RioThread`s responsibility to make sure its buffers are pinned, which is better because it's the one creating the `MemoryPool`.

I'm not sure about error handling, if it turns out the memory isn't from a known slab then the RIO buffer id is `IntPtr.Zero`, but we're not explicitly dealing with this.

### DIFF
--- a/src/System.IO.Pipelines.Networking.Windows.RIO/Internal/RioThread.cs
+++ b/src/System.IO.Pipelines.Networking.Windows.RIO/Internal/RioThread.cs
@@ -7,6 +7,7 @@ using System.Diagnostics;
 using System.Runtime.InteropServices;
 using System.Threading;
 using System.IO.Pipelines.Networking.Windows.RIO.Internal.Winsock;
+using System.Buffers;
 
 namespace System.IO.Pipelines.Networking.Windows.RIO.Internal
 {
@@ -92,7 +93,7 @@ namespace System.IO.Pipelines.Networking.Windows.RIO.Internal
             }
         }
 
-        public IntPtr GetBufferId(IntPtr address, out long startAddress)
+        private IntPtr GetBufferId(IntPtr address, out long startAddress)
         {
             var id = IntPtr.Zero;
             startAddress = 0;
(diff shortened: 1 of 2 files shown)


benaadams:
I think that looks good

/cc @davidfowl

davidfowl:
@KodrAus You need to call Free so the ref count doesn't just go up forever.

KodrAus:
@davidfowl Derp. Fixed. That's a nice segway into there are no tests for any of this code. Looks like the `libuv` TCP server at least has a stress test. Should RIO have the same when tested on a Windows box?

KodrAus:
@dnfclas Show me the CLA

src/System.IO.Pipelines.Networking.Windows.RIO/Internal/RioThread.cs
```diff
@@ -116,6 +116,23 @@ public IntPtr GetBufferId(IntPtr address, out long startAddress)
             return id;
         }
 
+        public unsafe RioBufferSegment GetSegmentFromMemory(Buffer<byte> memory)
+        {
+            var pin = memory.Pin();
```
davidfowl: Add a comment about why it's ok to unpin here because the memory needs to be pinned already? How do we enforce that though? There's no way to know if the backing memory is already pinned...

davidfowl: *But* since we create the pool, it's ok. Just add the comment and I'll merge it.

KodrAus: How's that?

PR #1377, opened 2017-03-25
Debug output relevant data for MemoryPoolBlock

benaadams:
Been down the wrong rabbit hole for a while due to what this was telling me...

/cc @davidfowl

### DIFF
--- a/src/System.IO.Pipelines/MemoryPoolBlock.cs
+++ b/src/System.IO.Pipelines/MemoryPoolBlock.cs
@@ -84,7 +84,7 @@ namespace System.IO.Pipelines
         public override string ToString()
         {
             var builder = new StringBuilder();
-            SpanExtensions.AppendAsLiteral(Buffer.Span, builder);
+            SpanExtensions.AppendAsLiteral(Buffer.Span.Slice(_offset, _length), builder);
             return builder.ToString();
         }
 


src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -84,7 +84,7 @@ internal void Initialize()
         public override string ToString()
         {
             var builder = new StringBuilder();
-            SpanExtensions.AppendAsLiteral(Buffer.Span, builder);
+            SpanExtensions.AppendAsLiteral(Buffer.Span.Slice(_offset, _length), builder);
```
davidfowl: Haha 😄

benaadams:
Maybe this is right already but [`OwnedBuffer`](https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs#L52-L53) is wrong

```csharp
        public Buffer<T> Buffer => new Buffer<T>(this, 0, Length);
        public ReadOnlyBuffer<T> ReadOnlyBuffer => new ReadOnlyBuffer<T>(this, 0, Length);
```

Should it be passing in Offset/_arrayIndex rather than 0?

PR #1379, opened 2017-03-25
ReadableBufferReader Skip split

benaadams:
Inlined same span for Skip
Non-inlined for skip span

Inlined enumerator next (`AdvanceSegment`) for .ctor and start of SkipMulti as always called
Non-inlined for Take and SkipMulti loop

SkipMulti to do while

Reversed if tests (cold branch prediction doesn't predict jump forward; loops jump backwards and Exceptions throws are moved to end of method to make them a jump forward)

### DIFF
--- a/src/System.IO.Pipelines/ReadableBufferReader.cs
+++ b/src/System.IO.Pipelines/ReadableBufferReader.cs
@@ -11,7 +11,7 @@ namespace System.IO.Pipelines
         private int _index;
         private SegmentEnumerator _enumerator;
         private int _consumedBytes;
-        private bool _end;
+        private bool _hasData;
 
         public ReadableBufferReader(ReadableBuffer buffer) : this(buffer.Start, buffer.End)
         {
@@ -19,7 +19,7 @@ namespace System.IO.Pipelines
 
         public ReadableBufferReader(ReadCursor start, ReadCursor end)
         {
-            _end = false;
+            _hasData = true;
             _index = 0;
             _consumedBytes = 0;
             _currentSpan = default(Span<byte>);
@@ -28,7 +28,7 @@ namespace System.IO.Pipelines
             AdvanceSegmentInlined();
         }
 
-        public bool End => _end;
+        public bool End => !_hasData;
 
         public int Index => _index;
 
@@ -37,7 +37,7 @@ namespace System.IO.Pipelines
             get
             {
                 var part = _enumerator.Current;
-                var index = !_end ? _index : _currentSpan.Length;
+                var index = _hasData ? _index : _currentSpan.Length;
 
                 return new ReadCursor()
                 {
@@ -54,7 +54,7 @@ namespace System.IO.Pipelines
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public int Peek()
         {
-            if (!_end)
+            if (_hasData)
             {
                 return _currentSpan[_index];
             }
@@ -65,7 +65,7 @@ namespace System.IO.Pipelines
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public int Take()
         {
-            if (!_end)
+            if (_hasData)
             {
                 var index = _index;
                 var value = _currentSpan[index];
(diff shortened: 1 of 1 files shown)


src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -101,23 +101,40 @@ private void MoveNext()
...
+            var remaining = _currentSpan.Length - _index;
+            if (length < remaining)
+            {
+                _index += length;
+            }
+            else
+            {
+                SkipMultiSegment(length - remaining);
+            }
+
```
davidfowl: nit: Remove space

benaadams: fixed

benaadams:
Rebased; tweaked inlines, updated summary

src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -37,12 +38,17 @@ public ReadCursor Cursor
             {
                 var part = _enumerator.Current;
 
-                if (_end)
+                ReadCursor cursor;
```
davidfowl: You can improve this by newing up a read cursor and setting the members

benaadams: Changed

pakrym:
Benchmark?

src/System.IO.Pipelines/ReadableBufferReader.cs
```diff
@@ -53,78 +54,107 @@ public ReadCursor Cursor
...
-            return _currentSpan[_index];
+
+            return -1;
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public int Take()
         {
-            if (_end)
+            if (!_end)
```
pakrym: Why can't JIT do this? Do we need to file a bug?

benaadams: Why would the JIT do it, how would it know to do it? You'd need to do PGO runs.

It does for exceptions (and correctly constructed throw helpers) as they should be uncommon paths; i.e. it moves the actual exception `throw ...` or call to `void Throw() -> throw` to end of method so if the "is invalid" test passes it is conditional jump forward (non predicted) to call the exception code.

It does for while and for loops, where possible; turning them into do while loops, so the test is at the end and its a conditional jump backwards (predicted) from the test. Backwards jump are predicted as they are loops and loops normally loop.

Ternary operators you want to put the more likely item first e.g. `var value = condition ? likelyResult : unlikelyResult` as it gets turned into:

```csharp
if (condition)
{
    value = likelyResult;
}
else
{ 
    value = unlikelyResult;
}
```
which becomes
```asm
test condition
jmpNotTrue second ; jmp forward not predicted
    value = likelyResult 
    jmp end
second:
    value = unlikelyResult
end:
```

...however the C# compiler on the other hand will reverse this code and change it to 
```csharp
if (_end)
{
   return -1;
}
...
```
```
.method public hidebysig instance int32 
    Take() cil managed 
  {
    .maxstack 4

    // [66 13 - 66 22]
    IL_0000: ldarg.0      // this
    IL_0001: ldfld        bool System.IO.Pipelines.ReadableBufferReader::_end
    IL_0006: brfalse.s    IL_000a

    // [68 17 - 68 27]
    IL_0008: ldc.i4.m1    
    IL_0009: ret          
```
defeating the change and deoptimizing the code; so I'd consider that a bug in csharp compiler

benaadams: Have to reverse the truth/false to get it to create right codegen; adding commit

pakrym: Does this give a measurable improvement?

benaadams:
Something made it happy; something made is unhappy; so will try to make the PR smaller and back to its original intent

benaadams:
Hold up... Can actually measure branch mispredicts now 😄  http://adamsitnik.com/Hardware-Counters-Diagnoser/

ahsonkhan:
> Benchmark?

@benaadams, any perf numbers?

PR #1380, opened 2017-03-25
(BufferEnumerator) Don't use Empty for discardable result 

benaadams:


pakrym:
Good change. How much did it take in non backtrace view?

benaadams:
@pakrym Its not a common path; but it is significant on that path

PR #1382, opened 2017-03-27
OwnedBuffer should avoid accidentally initializing into sentinel values

mgravell:
After many allocations, `OwnedBuffer`'s `_id` will wrap around; we should ensure that it doesn't accidentally select the sentinel values

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
@@ -97,8 +97,12 @@ namespace System.Buffers
             if (!IsDisposed && Id!=InitializedId) {
                 throw new InvalidOperationException("this instance has to be disposed to initialize");
             }
-
-            _id = (int)Interlocked.Increment(ref _nextId);
+            
+            do
+            {
+                _id = (int)Interlocked.Increment(ref _nextId);
+            } while(_id == InitializedId || _id == FreedId); // if we wrap around and collide with sentinels: pick again
+            
             _array = array;
             _arrayIndex = arrayOffset;
             _length = length;


davidfowl:
We'll likely move OwnedBuffer to be simpler (see https://github.com/dotnet/corefxlab/compare/master...KrzysztofCwalina:ExtensibleBuffers). Hopefully removing ids from the core.

mgravell:
nice; since you're here, @davidfowl, can I ask about the current timescales on the buffer stuff? I'm in active discussion on a similar topic in protobuf-net where someone is asking me about protobuf-net supporting pool allocators ([if you're curious](https://github.com/mgravell/protobuf-net/blob/696e56792f1e78f14b4da53b80a542866b41db3e/src/protobuf-net.Test/CustomSerializers.cs#L60-L131). At the moment my examples/tests are based on basic `ArraySegment<byte>` etc, but I'm trying to ensure that the API works equally with the upcoming stuff

davidfowl:
@mgravell Hmm, likely near to the end of year. There's lots of work happening here in the runtime, libraries and compiler, we'll be iterating on it aggressively though so it would be great if you could use it in a branch and give some real feedback.

mgravell:
@davidfowl ok; what I'll do, then, is start a fresh branch of protobuf-net looking at pipelines and related memory structures. A little part of me has been despearately hoping the "generators" stuff would materialize in some semi-usable state (since code-gen is going to be key)  - but since that seems stalled I'll probably rock with [CodeGeneration.Roslyn](https://github.com/AArnott/CodeGeneration.Roslyn) and pray that there's something reusable at the end of it :)

davidfowl:
> A little part of me has been despearately hoping the "generators" stuff would materialize in some semi-usable state (since code-gen is going to be key) 

I feel your pain 😭

ahsonkhan:
@mgravell, what should we do with the PR?

mgravell:
@ahsonkhan if the code still uses these identifiers, then there is technically a bug (although it will take a few billion allocations to see it), so I'd say: merge it, and **if and when** they go away, then fine: no longer a problem. But it isn't really up to me :)

mgravell:
I'll check when I'm at a PC to see the current state - I should have done that before commenting, but I suck at humaning.

mgravell:
@ahsonkhan problem no longer exists; closing

PR #1384, opened 2017-03-28
Added ability to parse headers represented  by ReadOnlyBytes

KrzysztofCwalina:
Note that the parser does not yet handle headers straddling spans. It will throw NotImplementedException in such case. I will add this capability later.

The parser is slightly faster than the ReadableBuffer parser, but I need to implement straddling headers and possibly removal of white space trailing header values, so I will refrain form publishing benchmarks for now.

The ReadableBuffer does more than one pass over bytes in some cases. This parser is single pass. But the ReadableBuffer used vectorized operations for searching some delimiters; this parser is sequential only.

Also, I improved performance tests to make results more consistent, but increasing inner iterations and unrolling the inner loop.

Lastly, I changed the handler callbacks to take ReadOnlySpan<byte> instead of taking Span<byte>. This is because ReadOnlyBytes does not give access to r/w data. ReadableBuffer (contrary to its name) givves access to r/w data.

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/BytesReader.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BytesReader.cs
@@ -50,6 +50,21 @@ namespace System.Buffers
 
         public byte Peek() => _currentSegment.Span[_currentSegmentIndex];
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public byte Read()
+        {
+            var result = _currentSegment.Span[_currentSegmentIndex];
+            _index++;
+            var unreadLength = _currentSegment.Length - _currentSegmentIndex;
+            if (1 < unreadLength) {
+                _currentSegmentIndex++;
+            }
+            else {
+                AdvanceNextSegment(1, advancedInCurrent: unreadLength);
+            }
+            return result;
+        }
+
         public ReadOnlySpan<byte> Unread => _currentSegment.Span.Slice(_currentSegmentIndex);
 
         public TextEncoder Encoder => _encoder;
--- a/src/System.Buffers.Experimental/System/Buffers/BytesReader.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BytesReader.cs
@@ -50,21 +50,6 @@ namespace System.Buffers
 
         public byte Peek() => _currentSegment.Span[_currentSegmentIndex];
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public byte Read()
-        {
-            var result = _currentSegment.Span[_currentSegmentIndex];
-            _index++;
-            var unreadLength = _currentSegment.Length - _currentSegmentIndex;
-            if (1 < unreadLength) {
-                _currentSegmentIndex++;
-            }
-            else {
-                AdvanceNextSegment(1, advancedInCurrent: unreadLength);
-            }
-            return result;
-        }
-
         public ReadOnlySpan<byte> Unread => _currentSegment.Span.Slice(_currentSegmentIndex);
 
         public TextEncoder Encoder => _encoder;
--- a/src/System.Text.Http.Parser/HttpParser.cs
+++ b/src/System.Text.Http.Parser/HttpParser.cs
@@ -354,13 +354,14 @@ namespace System.Text.Http.Parser
                 while (true) {
                     index++;
                     // get next byte 
-                    while (index == length) {
+                    while (length <= index) {
                         if (rest == null) {
                             RejectRequest(RequestRejectionReason.InvalidRequestHeadersNoCRLF);
                         }
                         span = rest.First.Span;
+                        length = span.Length;
                         rest = rest.Rest;
-                        index = -1;
+                        index = 0;
                         straddling = true;
                         throw new NotImplementedException("call to OnHeader below needs to account for straddling spans");
                     }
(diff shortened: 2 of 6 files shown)


KrzysztofCwalina:
cc: @geoffkizer, @davidfowl, @ahsonkhan, @shiftylogic

src/System.Buffers.Experimental/System/Buffers/BytesReader.cs
```diff
@@ -50,6 +50,21 @@ public BytesReader(IReadOnlyBufferList<byte> bytes, TextEncoder encoder) : this(
 
         public byte Peek() => _currentSegment.Span[_currentSegmentIndex];
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public byte Read()
+        {
+            var result = _currentSegment.Span[_currentSegmentIndex];
+            _index++;
```
ahsonkhan: Why is this being incremented here? It is not used in this method.

src/System.Buffers.Experimental/System/Buffers/BytesReader.cs
```diff
@@ -50,6 +50,21 @@ public BytesReader(IReadOnlyBufferList<byte> bytes, TextEncoder encoder) : this(
...
+            var unreadLength = _currentSegment.Length - _currentSegmentIndex;
+            if (1 < unreadLength) {
```
ahsonkhan: nit: reverse order.

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -340,6 +340,94 @@ public class HttpParser : IHttpParser
...
+            int index = -1;
+            State state = State.ParsingName;
+            int nameStart = 0, nameEnd = 0, valueStart = 0, valueEnd = 0;
+            bool sawCRLF = false;
+            bool straddling = false;
+            fixed (byte* pBuffer = &span.DangerousGetPinnableReference()) {
+                while (true) {
+                    index++;
+                    // get next byte 
+                    while (index == length) {
```
ahsonkhan: Why is this a while loop and not an if?

KrzysztofCwalina: because ReadOnlyBytes can contain empty segments

KrzysztofCwalina: But now, I noticed there is a bug here. I initally had index starting at 0 then I changed it to -1, and now the check does not work. I will fix.

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -340,6 +340,94 @@ public class HttpParser : IHttpParser
...
+                            valueStart = index;
+                        }
+                        else if (next == ByteCR) RejectRequest(RequestRejectionReason.InvalidCharactersInHeaderName);
+                        // else skip whitespace
+                    }
+                    else if (state == State.ArferCR) {
+                        if (next == ByteLF) {
+                            sawCRLF = true;
+                            if (nameEnd != 0) {
+                                if (!straddling) {
```
ahsonkhan: When will straddling be true?

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -340,6 +340,94 @@ public class HttpParser : IHttpParser
...
+                                    ReadOnlySpan<byte> nameBuffer = new ReadOnlySpan<byte>(pBuffer + nameStart, nameEnd - nameStart);
+                                    ReadOnlySpan<byte> valueBuffer = new ReadOnlySpan<byte>(pBuffer + valueStart, valueEnd - valueStart);
+                                    handler.OnHeader(nameBuffer, valueBuffer);
+                                    nameEnd = 0;
```
ahsonkhan: You reset nameEnd to 0. Why? Do you need to reset valueEnd as well?

src/System.Text.Http.Parser/HttpParser.cs
KrzysztofCwalina: When the while loop moves to the next segment after ParsingName started and before header is processed.

src/System.Text.Http.Parser/HttpParser.cs
KrzysztofCwalina: I use it in state == State.ArferCR to discover empty line, in which case nameEnd will never be set.

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -340,6 +340,95 @@ public class HttpParser : IHttpParser
...
+                            state = State.ArferCR;
+                        }
+                        // TODO: should value characters be validated?
+                        // TODO: should whitespace after all values (i.e. right before CRLF) be removed here?
+                        // else just keep advancing chracters in name
+                    }
+                    else if (state == State.ParsingName) {
+                        if (next == ByteColon) {
+                            state = State.BeforValue;
+                            nameEnd = index;
```
ahsonkhan: Should this be index - 1? Otherwise, aren't you including the ByteColon as part of the name?

davidfowl:
I'd suggest not making anymore changes until you've at least ported the unit tests.

src/System.Text.Http.Parser/HttpParser.cs
KrzysztofCwalina: the colon is not part of the name

KrzysztofCwalina:
Yeah, I would love to port the uni tests, but the tests use pipelines, so it's not super straightforward.

davidfowl:
Making changes without the tests make this a prototype at best. I'm not sure how you make these changes with any confidence. Do you know how many times we broke the tests making optimizations. IMO the benchmarks aren't valid with out tests

PR #1385, opened 2017-03-28
Adding new target framework (netcoreapp2.0)

ahsonkhan:
Also upgrading the build/install script to grab the latest dotnet cli tool.

System.Reflection.Metadata no longer has the interface `ISignatureTypeProvider<TType>` and now contains `ISignatureTypeProvider<TType, TGenericContext>`
https://github.com/dotnet/corefx/blob/master/src/System.Reflection.Metadata/src/System/Reflection/Metadata/Decoding/ISignatureTypeProvider.cs#L7
Therefore, I have disabled the build and test of the project System.Reflection.Metadata.Cil (and removed it from the solution) for now. This needs to be investigated further and fixed (I will file an issue).

The following test projects do not target netcoreapp2.0:
- Benchmarks
- System.Drawing.Graphics.Tests
- System.IO.Pipelines.Performance.Tests
- System.IO.Pipelines.Tests
- System.Text.Http.Tests
- System.Text.Primitives.Tests

This is because when I update the csproj to target netcoreapp2.0, I get the following error message:
`Starting test execution, please wait...
No test is available in D:\GitHub\Fork\corefxlab\tests\Benchmarks\bin\Debug\netcoreapp2.0\Benchmarks.dll. Make sure that installed test discoverers & executors, platform & framework version settings are appropriate and try again.`

The following three test projects already have the issue stated above, prior to any changes:
- Benchmarks
- System.Drawing.Graphics.Tests
- System.IO.Pipelines.Performance.Tests

This needs to be investigated further and fixed (I will file an issue).

`Span<T>` cannot be used within lambda expressions and hence I have changed tests accordingly, for example:
**Before:**
```C#
Assert.Throws<ArgumentNullException>(() => { Span<byte> span = new Span<byte>(null, 0, 0); });
```
**After:**
```C#
try
{
    Span<byte> span = new Span<byte>(null, 0, 0);
    Assert.True(false);
}
catch (Exception ex)
{
    Assert.True(ex is ArgumentNullException);
}
```

I have disabled certain tests that I wasn't sure how to fix (I will file an issue).
**Here are some example tests that were disabled:**
- [Fact(Skip = "System.TypeLoadException : A value type containing a by-ref instance field, such as Span<T>, cannot be used as the type for a static field.")]
JsonObjectTests.ParseNestedJson
- [Fact(Skip = "System.InvalidProgramException : Common Language Runtime detected an invalid program.")]
ReadOnlyBytesTests.SingleSegmentBytesReader
- [Fact(Skip = "System.TypeLoadException : The generic type 'System.Nullable`1' was used with an invalid instantiation in assembly 'System.Private.CoreLib")]
CompositeFormattingTests.CompositeFormattingBasics
- [Fact(Skip = "System.BadImageFormatException : An attempt was made to load a program with an incorrect format.")]
CompositeFormattingTests.CompositeFormattingUtf8String


ahsonkhan:
cc @shiftylogic, @KrzysztofCwalina, @davidfowl 
Please take a look.

davidfowl:
I stared in this in a branch. I think you should use it as inspiration to avoid some of the duplication here. See davidfowl/netcoreapp20

ahsonkhan:
> I stared in this in a branch. I think you should use it as inspiration to avoid some of the duplication here. See davidfowl/netcoreapp20

OK. Are you able to build the solution in VS after targeting net core 2.0? I cannot build in VS anymore and can only build the solution via command line.

ahsonkhan:
@davidfowl, why do you have these?
`<Import Project="..\..\tools\dependencies.props" />`

What is inside dependencies.props?

davidfowl:
> OK. Are you able to build the solution in VS after targeting net core 2.0? I cannot build in VS anymore and can only build the solution via command line.

Everything works in VS with .NET Core 2.0. You only have to update the tests and samples. The libraries are irrelevant.

> What is inside dependencies.props? I couldn't find this file in your branch.

Ah! The gitignore is missing props files. 

> It seems like you would want your changes on top of the changes I have made here.

No, I'd want my changes instead of what you have 😄 . The code changes are fine, the project file changes aren't right.

ahsonkhan:
@davidfowl, shall I close this PR then and possibly open a new one with just the test code changes?

davidfowl:
Make a PR with the code changes only. Close this one out

ahsonkhan:
Closing this one (I opened a new PR #1401).

Issue #1233, opened 2017-02-21
WritableBuffer: add support for non-contiguous Write


tmds:
The WSARecvFrom/recvmsg calls support receiving to a non-contiguous memory region.

I don't think the WritableBuffer supports allocating for such operations.

I.e. doing something like:
```C#
int allocated = 0;
while (allocated < needed)
{
   allocated += writableBuffer.Length;
   buffers.Add(writableBuffer.Memory);
   if (allocated < needed)
   {
       // adds a new segment available for writing at end of write head list
       // doesn't update the writeHead
       // Writable.Memory points to new segment
       writableBuffer.AllocSegment();
   }
}

int received = recvmsg(buffers);

// crosses segments
// updates writeHead
// writeHead.Next = null, Allocated segments after writeHead are Disposed
// Writable.Memory points to writeHead
writableBuffer.AdvanceWriter(received);
```

Ensure and Append should throw InvalidOperationException when writeHead.Next != null.

tmds:
@davidfowl makes sense?

davidfowl:
Hmm, no this doesn't work today. I need to think about it a bit.

tmds:
The ReaderBuffer non-contiguous read:
```C#
foreach (var memory in readableBuffer)
{
    buffers.Add(memory);
}
int sent = sendmsg(buffers);
reader.Advance(...);
```

@davidfowl does this look okay? What would be a good way to calculate the argument for the Advance call?

tmds:
@davidfowl any thoughts so far?

davidfowl:
Sorry dude, I'm heads down on getting back the kestrel perf right now. Haven't had enough free time recently to think about this.

tmds:
@davidfowl can I send a PR for this?

This is what I have in mind:
Add a `BufferSegment _readTail` to `Pipe`.
The current behavior would be `_readTail=_readHead`.
`AllocSegment` would allocate at the `_readTail` and return the `Memory`.
There are some debug asserts for `_readHead.Next==null`, these would become: `if (_readHead!=_readTail) throw InvalidOperation(...)`.
`Advance` would cross `BufferSegments`, and remaining `BufferSegments` would be `Disposed`.

davidfowl:
[...]

PR #1386, opened 2017-03-28
Add ReadableBuffer.Create(OwnedBuffer<byte>, offset, length)

tmds:
This API is useful for receiving data from the network to be added to a Pipe. Network APIs accept buffers and fill those buffers to some extent. This API allows the user to allocate a number of buffers, pass them to a receive API and then Append the filled ranges to a Pipe via WritableBuffer.Append(ReadableBuffer).

https://github.com/dotnet/corefxlab/issues/1233

@davidfowl @benaadams

### DIFF
--- a/src/System.IO.Pipelines/ReadableBuffer.cs
+++ b/src/System.IO.Pipelines/ReadableBuffer.cs
@@ -255,6 +255,32 @@ namespace System.IO.Pipelines
             return new ReadableBuffer(new ReadCursor(segment, offset), new ReadCursor(segment, offset + length));
         }
 
+        /// <summary>
+        /// Create a <see cref="ReadableBuffer"/> over an OwnedBuffer.
+        /// </summary>
+        public static ReadableBuffer Create(OwnedBuffer<byte> data, int offset, int length)
+        {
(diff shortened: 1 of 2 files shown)


src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -255,6 +255,32 @@ public static ReadableBuffer Create(byte[] data, int offset, int length)
             return new ReadableBuffer(new ReadCursor(segment, offset), new ReadCursor(segment, offset + length));
         }
 
+        /// <summary>
+        /// Create a <see cref="ReadableBuffer"/> over an OwnedBuffer.
+        /// </summary>
+        public static ReadableBuffer Create(OwnedBuffer<byte> data, int offset, int length)
```
davidfowl: There's a bunch of duplication here. Why not call this from the Array version?

benaadams:
Orthogonal to this; it does bug me that `ReadableBuffer Create(byte[] data)` goes via an overload that then does bounds checking on length and offset; which are fine because they are just the normal array bounds that have just been read... So its checking against itself

PR #1389, opened 2017-03-28
Fix Write(source, offset, length) fast path check

pakrym:
/cc @davidfowl  @benaadams


benaadams:
Will close https://github.com/dotnet/corefxlab/pull/1390 😄

PR #1394, opened 2017-03-29
Added TryRead to IPipeReader

davidfowl:
- To help support returning ValueTask and scenarios where
consumers can avoid creating the async state machine for as long as possible.
- The main difference between calling ReadAsync and TryRead is the fact
that if it returns false, you don't need to call Advance(though right now,
it throws).
- Makes it possible to implement synchronous API a bit more efficiently if the data is ready or the flush won't cause back pressure

PS: I'll look at implementing `TryWrite` but it might be awkward since writes are already synchronous and you still need to `FlushAsync` for the reader to see it. TryAlloc might be more appropriate but I haven't thought through it enough.

**EDIT: Will update with TryFlush** 

TryFlush feels pretty meh, but symmetry 😄 . It might have to go because calling `FlushAsync` after `TryFlush` attempts to signal to reader again which feels pretty wrong.

### DIFF
--- a/src/System.IO.Pipelines/Pipe.cs
+++ b/src/System.IO.Pipelines/Pipe.cs
@@ -322,7 +322,33 @@ namespace System.IO.Pipelines
             } // and if zero, just do nothing; don't need to validate tail etc
         }
 
+        internal bool TryFlush(out FlushResult result)
+        {
+            // Always signal the reader
+            SignalReader();
+
+            result = new FlushResult();
+            lock (_sync)
+            {
+                if (!_writerAwaitable.IsCompleted)
+                {
+                    return false;
+                }
+
+                GetResult(ref result);
+                return true;
+            }
+        }
+
         internal WritableBufferAwaitable FlushAsync()
+        {
+            // Signal the reader
+            SignalReader();
+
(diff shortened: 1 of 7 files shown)


Drawaes:
TryFlush would be useful as well ;) 
Use case multi framed and layered protocols that need innerpipes to handle fragmentation but are sync in nature.
[Example code](https://github.com/Drawaes/Leto/blob/v2/src/Leto/RecordLayer/RecordHandler.cs#L55)

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -322,8 +322,34 @@ internal void AdvanceWriter(int bytesWritten)
             } // and if zero, just do nothing; don't need to validate tail etc
         }
 
+        internal bool TryFlush(out FlushResult result)
+        {
+            // Always signal the reader
+            SignalReader();
```
pakrym: Remove lock from signal reader and move it under this methods lock

pakrym: Also call it SignalReaderUnsyncronized do nobody uses it without lock by accident.

PR #1401, opened 2017-03-30
Adding new target framework (netcoreapp2.0) for tests and samples

ahsonkhan:
- Upgraded the build script to grab the newer version of the dotnet cli tool
- Created a common dependencies.props so only one place needs to be updated to pick up newer package versions
- Fixed a lot of the tests that were failing since they were boxing `Span<T>`

`Span<T>` cannot be used within lambda expressions and hence I have changed tests accordingly, for example:
**Before:**
```C#
Assert.Throws<ArgumentNullException>(() => { Span<byte> span = new Span<byte>(null, 0, 0); });
```
**After:**
```C#
try
{
    Span<byte> span = new Span<byte>(null, 0, 0);
    Assert.True(false);
}
catch (Exception ex)
{
    Assert.True(ex is ArgumentNullException);
}
```
- Removed Span and Utf8String being passed in as MemberData in unit tests
- Disabled certain tests that were failing which I wasn't sure how to fix (I will file an issue for specific failure types)

**Here are some example test failure messages that resulted in those tests being disabled:**
> [Benchmark(Skip = "The generic type 'System.Collections.Generic.KeyValuePair`2' was used with an invalid instantiation in assembly 'System.Private.CoreLib")]

> [Fact(Skip = "System.InvalidProgramException : Common Language Runtime detected an invalid program.")]

> [Fact(Skip = "System.TypeLoadException : A value type containing a by-ref instance field, such as Span, cannot be used as the type for a static field.")]

> [Fact(Skip = "System.TypeLoadException : The generic type 'System.Nullable`1' was used with an invalid instantiation in assembly 'System.Private.CoreLib")]

> [Fact(Skip = "System.BadImageFormatException : An attempt was made to load a program with an incorrect format.")]

> [Fact(Skip = "System.TypeLoadException : The generic type 'System.Action`4' was used with an invalid instantiation in assembly 'System.Private.CoreLib")]

**List of issues opened to address disabled tests:**
- Issue #1403 
- Issue #1404 
- Issue #1405

### DIFF
--- a/tests/System.Buffers.Primitives.Tests/BasicUnitTests.cs
+++ b/tests/System.Buffers.Primitives.Tests/BasicUnitTests.cs
@@ -121,13 +121,29 @@ namespace System.Buffers.Tests
         [Fact]
         public void ByteSpanCtorWithRangeThrowsArgumentExceptionOnNull()
         {
-            Assert.Throws<ArgumentNullException>(() => { Span<byte> span = new Span<byte>(null, 0, 0); });
+            try
+            {
+                Span<byte> span = new Span<byte>(null, 0, 0);
+                Assert.True(false);
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex is ArgumentNullException);
+            }
         }
 
         [Fact]
         public void ByteReadOnlySpanCtorWithRangeThrowsArgumentExceptionOnNull()
         {
-            Assert.Throws<ArgumentNullException>(() => { ReadOnlySpan<byte> span = new ReadOnlySpan<byte>(null, 0, 0); });
+            try
+            {
+                ReadOnlySpan<byte> span = new ReadOnlySpan<byte>(null, 0, 0);
+                Assert.True(false);
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex is ArgumentNullException);
+            }
         }
 
         [Theory]
@@ -153,7 +169,15 @@ namespace System.Buffers.Tests
         [InlineData(new byte[3] { 0, 0, 0 }, 15, 0)]
         public void ByteSpanCtorWithRangeThrowsArgumentOutOfRangeException(byte[] bytes, int start, int length)
         {
-            Assert.Throws<ArgumentOutOfRangeException>(() => { Span<byte> span = new Span<byte>(bytes, start, length); });
+            try
+            {
+                Span<byte> span = new Span<byte>(bytes, start, length);
+                Assert.True(false);
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex is ArgumentOutOfRangeException);
+            }
         }
 
         [Theory]
@@ -197,10 +221,16 @@ namespace System.Buffers.Tests
         public void ByteSpanSliceWithRangeThrowsArgumentOutOfRangeException1(byte[] bytes, int start, int length)
         {
             var span = new Span<byte>(bytes);
-            Assert.Throws<ArgumentOutOfRangeException>(() =>
+
+            try
             {
                 Span<byte> slice = span.Slice(start, length);
-            });
+                Assert.True(false);
+            }
+            catch (Exception ex)
+            {
+                Assert.True(ex is ArgumentOutOfRangeException);
+            }
         }
 
         [Theory]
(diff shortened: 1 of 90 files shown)


ahsonkhan:
cc @shiftylogic, @davidfowl, @KrzysztofCwalina

tests/System.Buffers.Primitives.Tests/EqualityTests.cs
```diff
@@ -17,8 +17,7 @@ public void SpansMustNotBeBoxed()
 
             try
             {
-                span.Equals((object)span);
```
shiftylogic: Why do we even have these span tests in here? These same tests (or at least most of them) exist in corefx as well.

tests/System.Buffers.Primitives.Tests/BasicUnitTests.cs
```diff
@@ -121,13 +121,29 @@ public void ByteReadOnlySpanCtorWithRangeValidCases(byte[] bytes, int start, int
         [Fact]
         public void ByteSpanCtorWithRangeThrowsArgumentExceptionOnNull()
         {
-            Assert.Throws<ArgumentNullException>(() => { Span<byte> span = new Span<byte>(null, 0, 0); });
+            try
```
shiftylogic: First, I think we should remove the tests that are duplicates of ones in CoreFx. Also, I'd move any non-duplicate tests that are specifically testing span to the CoreFx tests.

Then, I'd also look at how CoreFx uses the "DontBox" concept in the tests to effectively do exactly what this code is doing. It ends up being one line and cleaner looking with the help of a small static helper.

tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
```diff
@@ -354,7 +354,7 @@ private unsafe void TestValue(ref ReadableBuffer readBuffer, ulong value)
             }
         }
 
-        [Fact]
+        [Fact(Skip = "System.TypeLoadException : A value type containing a by-ref instance field, such as Span<T>, cannot be used as the type for a class instance field.")]
```
shiftylogic: I think disabling these tests is the wrong approach. In this case, the issue is caused by the fact that the async function is creating a closure around the local span to capture them for re-entering the function when an await returns. This only happens if the span is used on both sides of an await, either directly or indirectly.

We need some documented patterns on things that work and don't work. However, you should be able to re-write these tests in a way that will allow them to work correctly without the error.

tests/System.IO.Pipelines.Tests/SocketsFacts.cs
```diff
@@ -176,6 +175,9 @@ public void CanCreateWorkingEchoServer_PipelineSocketServer_NonPipelineClient()
 
         static async Task<Tuple<int, int, int>> PingClient(IPipeConnection connection, int messagesToSend)
         {
+            Span<byte> _ping = new Span<byte>(Encoding.ASCII.GetBytes("PING"));
```
shiftylogic: This is almost certainly being caused by the WriteAsync extension method on IPipeWriter (WriteAsync). Async methods cannot take Span<T> as an argument.

tests/System.Text.Http.Tests/GivenAnHttpHeaders.cs
```diff
@@ -25,46 +25,32 @@ public class GivenAnHttpHeaders
...
-                {
-                    headers = new Span<byte>(buffer, bytes.Length);
-                }
-            }
-
-            _httpHeaders = new HttpHeadersSingleSegment(headers);
-        }
-
-        [Fact]
+        [Fact(Skip = "System.TypeLoadException : The generic type 'System.Collections.Generic.KeyValuePair`2' was used with an invalid instantiation in assembly 'System.Private.CoreLib")]
```
shiftylogic: HttpHeadersSingleSegment stores a Utf8String as a member variable but also implements IEnumerable. I'm pretty sure that will result in a Span on the heap.

tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
ahsonkhan: I agree with you. However, I do not have the capacity to work on these right now. This is why I have filed issues to track this work for later or so others can take a shot at it if they can.

tests/System.Buffers.Primitives.Tests/EqualityTests.cs
ahsonkhan: Agreed.

tests/System.Buffers.Primitives.Tests/BasicUnitTests.cs
ahsonkhan: I agree. In some scenarios, however, you cannot use DontBox and have to do this try catch (when you are not creating a new span). This is not one of those scenarios.

tests/System.Text.Http.Tests/GivenAnHttpHeaders.cs
ahsonkhan: Utf8String implements IEnumerable as well. I did not see issues there.

https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8String.CodePointEnumerable.cs#L11

tests/System.IO.Pipelines.Tests/SocketsFacts.cs
ahsonkhan: Filed issue #1405

tests/System.Buffers.Primitives.Tests/BasicUnitTests.cs
ahsonkhan: Filed issue #1403

tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
ahsonkhan: Filed issue #1404

ahsonkhan:
@shiftylogic, I think we should merge this PR and work on the issues raised on top.
We should leverage the compiler support to flag/prevent future breaking test.

@davidfowl, good to go?

PR #1402, opened 2017-03-30
Expose OwnedBuffer.TryGetPointer

tmds:
There is no way to get the pointer of an OwnedBuffer which is needed to support https://github.com/dotnet/corefxlab/issues/1233.

@davidfowl @benaadams

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/Buffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Buffer.cs
@@ -65,11 +65,7 @@ namespace System.Buffers
 
         public unsafe bool TryGetPointer(out void* pointer)
         {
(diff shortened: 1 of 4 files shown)


tmds:
Not sure what you mean. Is there a public API that allows me to get the pointer? I'm using an OwnedBuffer from the MemoryPool so it's pinned already.

Drawaes:
Var handle = OwnedBuffer.Buffer.Pin()
If it's pinned already it should be a noop
Then handle.pinnedpointer

tmds:
Hadn't seen that. Nice, Thanks!

Drawaes:
No problem..  :)

KodrAus:
Note you need to keep the resulting `MemoryHandle` live and call `Free` when you don't need the pointer anymore or you'll leak memory.

PR #1406, opened 2017-03-30
Build all projects listed in the solution file, including samples

ahsonkhan:
cc @KrzysztofCwalina, @shiftylogic, @crummel 

This is in sync with the restore step which restores only the projects that are in the corefxlab.sln.

It also syncs the behavior when building from Visual Studio and from the command line.


davidfowl:
You can just build the solution file.

ahsonkhan:
> You can just build the solution file.

Yes, but then you do not get the error reporting specific to failing projects and number of projects failing. If that is not crucial, then building the solution is a simple solution.

davidfowl:
> Yes, but then you do not get the error reporting specific to failing projects and number of projects failing. If that is not crucial, then building the solution is a simple fix.

You should never need to manually parse the solution file. Just build it.

ahsonkhan:
Passing the solution file to dotnet.exe build and dotnet.exe test reduced the end to end build/test time to half. (from around 8 minutes to around 4)!

ahsonkhan:
@dotnet-bot test

ahsonkhan:
@dotnet-bot test this

ahsonkhan:
@dotnet-bot test this please

ahsonkhan:
@dotnet-bot test this please

ahsonkhan:
Do we want this change? Thought it reduces the build/test time, it makes really hard to debug what failed if a failure does occur (you don't see which test failed, or which project). This is a side effect of building the solution file instead of individual csproj files.

@shiftylogic, @KrzysztofCwalina, @davidfowl

crummel:
@ahsonkhan You might want to try the testing single-threaded while leaving the build multi-threaded.  I know we do this in some projects and that seems like a lot of why it's harder to go through the output and find failures.

ahsonkhan:
@shiftylogic, good to go?

shiftylogic:
Does building the individual csproj files still decrease build times over building the whole solution?

shiftylogic:
Forget what I said. All you are adding is better failure detection.

You are good to go.

ahsonkhan:
> Does building the individual csproj files still decrease build times over building the whole solution?

I still build the whole solution (not individual csproj). The tests are being run individually (and they are slower) using dotnet test *.csproj.

PR #1421, opened 2017-04-04
Remove WriteAsync overloads that take span

pakrym:
@davidfowl

### DIFF
--- a/src/System.IO.Pipelines/PipelineWriterExtensions.cs
+++ b/src/System.IO.Pipelines/PipelineWriterExtensions.cs
@@ -9,7 +9,12 @@ namespace System.IO.Pipelines
     {
         private readonly static Task _completedTask = Task.FromResult(0);
 
-        public static Task WriteAsync(this IPipeWriter output, Span<byte> source)
+        public static Task WriteAsync(this IPipeWriter output, byte[] source)
+        {
+            return WriteAsync(output, new ArraySegment<byte>(source));
+        }
+
+        public static Task WriteAsync(this IPipeWriter output, ArraySegment<byte> source)
         {
             var writeBuffer = output.Alloc();
             writeBuffer.Write(source);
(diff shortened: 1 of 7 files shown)


ahsonkhan:
cc @shiftylogic.

@pakrym, do we want to remove the use of Span or can we re-write the tests to not use async?

src/System.IO.Pipelines/PipelineWriterExtensions.cs
```diff
@@ -9,7 +9,12 @@ public static class PipelineWriterExtensions
     {
         private readonly static Task _completedTask = Task.FromResult(0);
 
-        public static Task WriteAsync(this IPipeWriter output, Span<byte> source)
+        public static Task WriteAsync(this IPipeWriter output, byte[] source)
+        {
+            return WriteAsync(output, new ArraySegment<byte>(source));
+        }
+
+        public static Task WriteAsync(this IPipeWriter output, ArraySegment<byte> source)
```
ahsonkhan: Is this change unnecessary?

pakrym: Why is it unnecessary?

pakrym:
@ahsonkhan these tests do not test span so I'm fine with keeping them like this.

src/System.IO.Pipelines/PipelineWriterExtensions.cs
ahsonkhan: There isn't anything inherently wrong with having a WriteAsync overloads that take span. If you actually don't want this overload, then feel free to remove. The compiler won't complain if you had this method take a span since it isn't async. The tests were async.

pakrym: But it's no very useful because most clients are `async` methods and having this overload causes them to blow up.

PR #1426, opened 2017-04-05
Cannot use Span as a generic type

ahsonkhan:
Fixing some tests that were previously skipped because Span was used as a generic type within List<T>.

cc @shiftylogic, @botaberg, @davidfowl

### DIFF
--- a/tests/System.Text.Primitives.Tests/PrimitiveParserUInt32PerfTests.cs
+++ b/tests/System.Text.Primitives.Tests/PrimitiveParserUInt32PerfTests.cs
@@ -279,23 +279,28 @@ namespace System.Text.Primitives.Tests
             }
         }
 
-        [Benchmark(Skip = "The generic type 'System.Collections.Generic.List`1' was used with an invalid instantiation in assembly 'System.Private.CoreLib")]
-        private unsafe static void PrimitiveParserByteSpanToUInt32_VariableLength()
+        [Benchmark]
+        [InlineData(0)]
+        [InlineData(1)]
+        [InlineData(2)]
+        [InlineData(3)]
+        [InlineData(4)]
+        [InlineData(5)]
+        [InlineData(6)]
+        [InlineData(7)]
+        [InlineData(8)]
+        [InlineData(9)]
+        private unsafe static void PrimitiveParserByteSpanToUInt32_VariableLength(int index)
         {
-            List<ReadOnlySpan<byte>> byteSpanList = new List<ReadOnlySpan<byte>>();
-            foreach (string text in s_UInt32TextArray)
-            {
-                byte[] utf8ByteArray = Encoding.UTF8.GetBytes(text);
-                ReadOnlySpan<byte> utf8ByteSpan = new ReadOnlySpan<byte>(utf8ByteArray);
-                byteSpanList.Add(utf8ByteSpan);
-            }
+            string text = s_UInt32TextArray[index];
+            byte[] utf8ByteArray = Encoding.UTF8.GetBytes(text);
+            ReadOnlySpan<byte> utf8ByteSpan = new ReadOnlySpan<byte>(utf8ByteArray);
             foreach (var iteration in Benchmark.Iterations)
             {
                 using (iteration.StartMeasurement())
                 {
                     for (int i = 0; i < LoadIterations; i++)
                     {
-                        ReadOnlySpan<byte> utf8ByteSpan = byteSpanList[i % 10];
                         uint value;
                         PrimitiveParser.InvariantUtf8.TryParseUInt32(utf8ByteSpan, out value);
                         DoNotIgnore(value, 0);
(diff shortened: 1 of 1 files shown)


tests/System.Text.Primitives.Tests/PrimitiveParserUInt32PerfTests.cs
```diff
@@ -327,23 +332,28 @@ private unsafe static void PrimitiveParserByteSpanToUInt32_BytesConsumed(string
...
-            List<ReadOnlySpan<byte>> byteSpanList = new List<ReadOnlySpan<byte>>();
-            foreach (string text in s_UInt32TextArray)
-            {
-                byte[] utf8ByteArray = Encoding.UTF8.GetBytes(text);
-                ReadOnlySpan<byte> utf8ByteSpan = new ReadOnlySpan<byte>(utf8ByteArray);
-                byteSpanList.Add(utf8ByteSpan);
-            }
+            string text = s_UInt32TextArray[index];
+            byte[] utf8ByteArray = Encoding.UTF8.GetBytes(text);
+            ReadOnlySpan<byte> utf8ByteSpan = new ReadOnlySpan<byte>(utf8ByteArray);
```
davidfowl: var

tests/System.Text.Primitives.Tests/PrimitiveParserUInt32PerfTests.cs
```diff
@@ -481,23 +491,26 @@ private unsafe static void PrimitiveParserByteSpanToUInt32Hex(string text)
...
-            List<ReadOnlySpan<byte>> byteSpanList = new List<ReadOnlySpan<byte>>();
-            foreach (string text in s_UInt32TextArrayHex)
-            {
-                byte[] utf8ByteArray = Encoding.UTF8.GetBytes(text);
-                ReadOnlySpan<byte> utf8ByteSpan = new ReadOnlySpan<byte>(utf8ByteArray);
-                byteSpanList.Add(utf8ByteSpan);
-            }
+            string text = s_UInt32TextArrayHex[index];
+            byte[] utf8ByteArray = Encoding.UTF8.GetBytes(text);
+            ReadOnlySpan<byte> utf8ByteSpan = new ReadOnlySpan<byte>(utf8ByteArray);
```
davidfowl: var

tests/System.Text.Primitives.Tests/PrimitiveParserUInt32PerfTests.cs
```diff
@@ -529,23 +542,26 @@ private unsafe static void PrimitiveParserByteSpanToUInt32Hex_BytesConsumed(stri
...
```
davidfowl: var

tests/System.Text.Primitives.Tests/PrimitiveParserUInt32PerfTests.cs
```diff
@@ -279,23 +279,28 @@ private unsafe static void PrimitiveParserByteSpanToUInt32(string text)
             }
         }
 
-        [Benchmark(Skip = "The generic type 'System.Collections.Generic.List`1' was used with an invalid instantiation in assembly 'System.Private.CoreLib")]
-        private unsafe static void PrimitiveParserByteSpanToUInt32_VariableLength()
+        [Benchmark]
+        [InlineData(0)]
```
shiftylogic: The size of the list that was being created is fixed. Can't you just use an array sized at run-time and do effectively the same test without having to run this 10 different times?

ahsonkhan: I tried that. You cannot create an array of Span.
ReadOnlySpan<byte>[] byteSpanList = new ReadOnlySpan<byte>[10];

..\..\dotnet\dotnet.exe run -c Release

Unhandled Exception: System.TypeLoadException: Could not create array type 'System.ReadOnlySpan`1[System.Byte][]' from assembly 'System.Private.CoreLib, Version=4.0.0.0, Culture=neutral, PublicKeyToken=7cec85d7bea7798e'.
   at ConsoleApp.Program.Main(String[] args)

shiftylogic: But you can create an array of byte arrays, then implicit cast to span kicks in during the perf loop.

Does somethign like this not work:
```csharp
        [Benchmark]
        private unsafe static void PrimitiveParserByteSpanToUInt32_VariableLength()
        {
            byte[][] byteSpanList = (byte[][])Array.CreateInstance(typeof(byte[]), s_UInt32TextArray.Length);
            for (var i = 0; i < s_UInt32TextArray.Length; i++)
            {
                byteSpanList[i] = Encoding.UTF8.GetBytes(s_UInt32TextArray[i]);
            }

            {
                {
                    for (int i = 0; i < LoadIterations; i++)
                    {
                        ReadOnlySpan<byte> utf8ByteSpan = byteSpanList[i % 10];
                        uint value;
                        PrimitiveParser.InvariantUtf8.TryParseUInt32(utf8ByteSpan, out value);
                        DoNotIgnore(value, 0);
                    }
                }
            }
        }
```

ahsonkhan: That would work, but you are then adding the cost of constructing a span in your benchmark.

shiftylogic: The previous test was pulling values from a list. This should be faster than it was, and we only have to look at a single benchmark entry instead of 10.

These are all about relativity anyways. They only make sense when compared to prior runs or implementation differences.

ahsonkhan: If you look at some of the other tests, they essentially do the same thing:
https://github.com/dotnet/corefxlab/blob/master/tests/System.Text.Primitives.Tests/PrimitiveParserUInt32PerfTests.cs#L466

Instead of passing in index to a string[], they just pass in the strings themselves. I could do that, if you prefer.

shiftylogic: With this PR or your suggested changes, how are PrimitiveParserByteSpanToUInt32_VariableLength and PrimitiveParserByteSpanToUInt32 different other than different input strings? Isn't the first supposed to be doing.

ahsonkhan: I suppose. I don't really understand why we have two. This isn't testing for correctness so I don't see the significance of modifying the length of the string being parsed during each inner loop.

shiftylogic: If the test is not relevant, add a few more cases to PrimitiveParserByteSpanToUInt32 and remove the other, or make the other something more meaningful.

Again, there are benchmarks so are more about relativity of performance run-to-run than they are about correctness.

ahsonkhan: I will create an array of byte arrays, then implicit cast to span as you suggested to keep the semantics of the original tests. We can revisit later if necessary.

Issue #1419, opened 2017-04-04
Investigate why ReadOnlyBytes tests fail with System.InvalidProgramException


ahsonkhan:
Example: https://github.com/dotnet/corefxlab/blob/master/tests/System.Buffers.Experimental.Tests/BytesReader.cs#L14
```C#
[Fact]
public void SingleSegmentBytesReader()
{
    ReadOnlyBytes bytes = Create("AB CD#EF&&");
    var reader = new BytesReader(bytes);

    var ab = reader.ReadBytesUntil((byte)' ');
    Assert.Equal("AB", ab.ToString(TextEncoder.Utf8));

    reader.Advance(1);
    var cd = reader.ReadBytesUntil((byte)'#');
    Assert.Equal("CD", cd.ToString(TextEncoder.Utf8));

    reader.Advance(1);
    var ef = reader.ReadBytesUntil(new byte[] { (byte)'&', (byte)'&' });
    Assert.Equal("EF", ef.ToString(TextEncoder.Utf8));

    reader.Advance(2);

    //Assert.True(reader.IsEmpty);
}
```

The sample test causes the error:
`System.InvalidProgramException : Common Language Runtime detected an invalid program.`

Why?

cc @KrzysztofCwalina

shiftylogic:
The ToString extension method is passing a stack only structure to a StringBuilder function. Instead of converting the Utf8String to a string, it is boxing it by calling the "object" variant of Append. I changed the method to explicitly call ToString on the Utf8String instance and re-enabled the tests.

#1428

PR #1428, opened 2017-04-05
Re-enabling and fixing the BytesReader tests

shiftylogic:
From #1419


shiftylogic:
@ahsonkhan

tests/System.Buffers.Experimental.Tests/BytesReader.cs
```diff
@@ -10,13 +10,14 @@ namespace System.Slices.Tests
...
     {
-        [Fact(Skip = "System.InvalidProgramException : Common Language Runtime detected an invalid program.")]
+        [Fact]
         public void SingleSegmentBytesReader()
         {
             ReadOnlyBytes bytes = Create("AB CD#EF&&");
             var reader = new BytesReader(bytes);
 
             var ab = reader.ReadBytesUntil((byte)' ');
+            Assert.True(ab.HasValue);
```
ahsonkhan: Why is this necessary?

shiftylogic: It isn't really necessary. I added it during my investigation. However, it is a correct test but mostly benign.

Issue #1235, opened 2017-02-22
Span taking Enumerators/Enumerable can't implement IEnumerable(or)


benaadams:
As they could be boxed via interface and take Span off the stack e.g. 
```csharp
        public struct CodePointEnumerable : IEnumerable<uint>, IEnumerable
        {
            private ReadOnlySpan<byte> _buffer;
```

ahsonkhan:
@shiftylogic, would this be part of the UTF8String re-work?
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8String.CodePointEnumerable.cs#L11

ahsonkhan:
These types related to Utf8String need to be re-designed:
- CodePointReverseEnumerator
- CodePointEnumerator
- CodePointEnumerable

cc @KrzysztofCwalina

PR #1433, opened 2017-04-06
Fixing use of Span/Utf8String in tests

ahsonkhan:
Re-enabling tests that were disabled during migration to net core 2.0 here: https://github.com/dotnet/corefxlab/pull/1401

- Utf8String cannot be used as a generic in `Nullable<T>`.
- Cannot use Utf8String as a static.
- Cannot box Utf8String by calling GetType on it.
-  Cannot use Span or Utf8String as a static + removing a duplicate test.
-  Cannot cast Utf8String as an object.
- Removing tests (commenting out) around JsonDynamicObject and HttpHeadersSingleSegment.

Opened issues to address change requests to these types since they need to be redesigned.
- JsonDynamicObject / JsonLazyDynamicObject - #1434
- HttpHeadersSingleSegment - #1435
- Utf8String CodePointReverseEnumerator / CodePointEnumerator / CodePointEnumerable - #1235

cc @shiftylogic, @KrzysztofCwalina

### DIFF
--- a/src/System.Text.Http/System/Text/Http/IFormatterHttpExtensions.cs
+++ b/src/System.Text.Http/System/Text/Http/IFormatterHttpExtensions.cs
@@ -9,18 +9,18 @@ namespace System.Text.Http
     public static class IFormatterHttpExtensions
     {
         //TODO: Issue #387: In the Http extensions of IFormatter, we need to ensure that all the characters follow the basic rules of rfc2616
-        private static readonly Utf8String Http10 = new Utf8String("HTTP/1.0");
-        private static readonly Utf8String Http11 = new Utf8String("HTTP/1.1");
-        private static readonly Utf8String Http20 = new Utf8String("HTTP/2.0");
+        private static readonly string Http10 = "HTTP/1.0";
+        private static readonly string Http11 = "HTTP/1.1";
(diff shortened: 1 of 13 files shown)


ahsonkhan:
@dotnet-bot test this please

src/System.Text.Http/System/Text/Http/IFormatterHttpExtensions.cs
```diff
@@ -9,18 +9,18 @@ namespace System.Text.Http
     public static class IFormatterHttpExtensions
     {
         //TODO: Issue #387: In the Http extensions of IFormatter, we need to ensure that all the characters follow the basic rules of rfc2616
-        private static readonly Utf8String Http10 = new Utf8String("HTTP/1.0");
-        private static readonly Utf8String Http11 = new Utf8String("HTTP/1.1");
-        private static readonly Utf8String Http20 = new Utf8String("HTTP/2.0");
+        private static readonly string Http10 = "HTTP/1.0";
```
davidfowl: Maybe these should be `byte[]`

ahsonkhan: What is the rationale behind that?

Issue #1436, opened 2017-04-06
ISequence<ReadOnlyBuffer<byte>>.ToSpan extension causing hang


shiftylogic:
When used with ReadableBuffer, in the case of a single span or multi-span buffer, ToSpan hangs indefinitely. I suspect the issue is somewhere in ReadableBuffer.TryGet implementation.

shiftylogic:
#1438

PR #1438, opened 2017-04-06
Fix for ReadableBuffer.ToSpan hang

shiftylogic:
Fix for #1436


shiftylogic:
@ahsonkhan @davidfowl @pakrym

tests/System.Buffers.Experimental.Tests/BufferExtensionTests.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System.IO.Pipelines;
+using System.IO.Pipelines.Testing;
+using Xunit;
+
+namespace System.Buffers.Experimental.Tests
+{
+    public class BufferExtensionTests
```
ahsonkhan: Why are these tests in Buffers.Experimental?

ahsonkhan:
@dotnet-bot test this please

tests/System.Buffers.Experimental.Tests/BufferExtensionTests.cs
```diff
@@ -0,0 +1,35 @@
...
+    {
+        [Fact]
+        public void ToSpanOnReadableBufferSingleSpan()
+        {
+            byte[] data = { 0, 1, 2, 3, 4, 5, 6, 7 };
+
+            var readable = BufferUtilities.CreateBuffer(data);
+            var span = readable.ToSpan();
```
ahsonkhan: Isn't this test testing `ToSpan(this ReadableBuffer buffer)`
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Http.Parser/PipelineExtensions.cs#L15

tests/System.Buffers.Experimental.Tests/BufferExtensionTests.cs
shiftylogic: They are testing an extension method being added into that assembly, so that makes sense to me.

tests/System.Buffers.Experimental.Tests/BufferExtensionTests.cs
shiftylogic: It does not run that version of ToSpan. It runs the one in buffer experimental. That ToSpan is internal and private to pipelines.

PR #1442, opened 2017-04-06
Add reset to pipes

pakrym:


src/System.IO.Pipelines/IPipe.cs
```diff
@@ -8,5 +8,6 @@ public interface IPipe
     {
         IPipeReader Reader { get; }
         IPipeWriter Writer { get; }
+        void Reset();
```
davidfowl: Now this is hella hacky 😄 . Reset on the interface!

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -85,8 +85,15 @@ public Pipe(BufferPool pool, PipeOptions options = null)
...
             _writerScheduler = options.WriterScheduler ?? InlineScheduler.Default;
+            ResetState();
+        }
+
+        private void ResetState()
+        {
             _readerAwaitable = new PipeAwaitable(completed: false);
             _writerAwaitable = new PipeAwaitable(completed: true);
+            _readerCompletion = default(PipeCompletion);
+            _writerCompletion = default(PipeCompletion);
```
davidfowl: What about all of the other fields? Who resets length?

src/System.IO.Pipelines/IPipe.cs
pakrym: Nah, it's nice, owner of entire pipe should reset.

src/System.IO.Pipelines/Pipe.cs
pakrym: Good catch, would add a test.

src/System.IO.Pipelines/IPipe.cs
pakrym: We can do TryReset() or throw

PR #1443, opened 2017-04-07
Referencing and using custom C# compiler from roslyn myget feed

ahsonkhan:
- ~**Using msbuild to build the solution (instead of dotnet cli) <= do we want to do this?**~
- Adding a new myget feed for roslyn
- Including a package reference to Microsoft.Net.Compilers in all the projects
- Adding RoslynVersion property

From @VSadov: 
> As a first step we are using 2.3.0-rdonly-ref-61530-07 here.
NOTE: This compiler version does not implement any checks for Span<T>. That will come in the next change.

cc @KrzysztofCwalina, @shiftylogic, @VSadov. @joshfree

PR #1444, opened 2017-04-07
referencing C# compiler from a myget feed

VSadov:
As a first step we are using 2.3.0-rdonly-ref-61530-07 here.

NOTE: This compiler version does not implement any checks for Span<T>. That will come in the next change.


ahsonkhan:
Duplicate of #1443 

Also, it turns out, you do not need to run dotnet msbuild.

PR #1454, opened 2017-04-07
Switching to the new toolset compiler.

VSadov:
The new version is
https://dotnet.myget.org/feed/roslyn/package/nuget/Microsoft.Net.Compilers/2.3.0-rdonly-ref-61607-04

The new version enforces the following rules:
- `Span<T>` and `ReadonlySpan<T>` are stack-only types. (no boxing, invalid use in async, etc...)
- `Span<T>` and `ReadonlySpan<T>` cannot be passed via a writeable parameter.

There were several violations that I fixed.

NOTE: we are not yet validating the usage of span-containing structs. There could be more indrect violations of stack safety through that.

### DIFF
--- a/src/System.Text.Http.Parser/HttpParser.cs
+++ b/src/System.Text.Http.Parser/HttpParser.cs
@@ -36,10 +36,18 @@ namespace System.Text.Http.Parser
                 consumed = buffer.Move(consumed, lineIndex + 1);
                 span = span.Slice(0, lineIndex + 1);
             }
-            else if (buffer.IsSingleSpan || !TryGetNewLineSpan(ref buffer, ref span, out consumed))
+            else if (buffer.IsSingleSpan)
             {
-                // No request line end
                 return false;
+            }
+            else
+            {
+                span = TryGetNewLineSpan(ref buffer, out consumed);
+                if (span.Length == 0)
+                {
+                    // No request line end
+                    return false;
+                }
             }
 
             // Fix and parse the span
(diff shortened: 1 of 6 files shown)


VSadov:
CC: @jaredpar @OmarTawfik

tests/Benchmarks/Helpers/Server.cs
```diff
@@ -35,14 +35,14 @@ public static void RunSingleSegmentParser(int numberOfRequests, int concurrentCo
                             break;
                         }
 
-                        var requestBytes = input.First.Span;
+                        var requestBuffer = input.First;
 
-                        if (requestBytes.Length != 492)
+                        if (requestBuffer.Length != 492)
```
ahsonkhan: Is requestBuffer.Length == requestBuffer.Span.Length?

ahsonkhan:
Otherwise, LGTM.

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -581,17 +589,17 @@ private static unsafe bool Contains(byte* searchSpace, int length, byte value)
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private static bool TryGetNewLineSpan(ref ReadableBuffer buffer, ref Span<byte> span, out ReadCursor end)
+        private static Span<byte> TryGetNewLineSpan(ref ReadableBuffer buffer, out ReadCursor end)
```
shiftylogic: Is this more efficient, or just not allowed anymore?

tests/Benchmarks/Helpers/Server.cs
VSadov: Yes. Both have the same `Buffer<T>_length` value.

---
In reply to: [110496917](https://github.com/dotnet/corefxlab/pull/1454#discussion_r110496917) [](ancestors = 110496917)

src/System.Text.Http.Parser/HttpParser.cs
VSadov: ref/out spans are not allowed. 
Not sure if more efficient, I'd guess it is the same.

---
In reply to: [110497909](https://github.com/dotnet/corefxlab/pull/1454#discussion_r110497909) [](ancestors = 110497909)

PR #1360, opened 2017-03-23
Improved Pooling

benaadams:
Pool `BufferSegment`s
Add `ThreadStatic` single cache for `MemoryPool`
Test for negative `MemoryPool` block size request
Don't test `block.Slab != null` as its always true
Tidy up `MemoryPool`  `#if`

/cc @pakrym @davidfowl

PR #1459, opened 2017-04-09
Remove dangling ref from UnownedBufferReader

benaadams:
Discovered when pooling BufferSegments in https://github.com/dotnet/corefxlab/pull/1360

Isn't a problem currently as they just get GC'd eventually; but is a problem if they get pooled as something else updates the buffer and then its no longer empty.

/cc @davidfowl


ahsonkhan:
@davidfowl, are you signing off on this PR?

benaadams:
Is a bug; its holding onto the segment and Disposing it, while relying on `_head.ReadableBytes == 0`; when it should just drop the segment rather than holding onto it past disposal; and using the null check on `_head` to indicate nothing more to read.

Issue #57, opened 2015-03-27
Add support for non-invariant cultures


KrzysztofCwalina:
There needs to be a way to get FormattingData instances for other cultures.
There are two ways to do it: either get the data from .NET CultureInfo or write a parser for the standard Unicode database.

PR #1462, opened 2017-04-11
Primitive formatting enhancements for Time and Uuid

shiftylogic:
* Optimized for Invariant UTF-8 and UTF-16. 
* Formats are matched with existing ToString(InvariantCulture).
* Test coverage.
* Performance improvements (2-7x depending on scenario).


src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
```diff
@@ -0,0 +1,144 @@
...
+        {
+            Unsafe.Add(ref buffer, index) = (byte)HexTable[value >> 4];
+            Unsafe.Add(ref buffer, index + 1) = (byte)HexTable[value & 0xF];
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int WriteFractionDigits(long value, int digitCount, ref byte buffer, int index)
+        {
+            for (var i = FractionDigits; i > digitCount; i--)
+                value = DivMod(value, 10, out long m);
```
ahsonkhan: We don't use `out long m` anywhere, should we have an overload that doesn't return an out parameter?

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
```diff
@@ -0,0 +1,144 @@
...
+        #region Math Helper methods
+
+        /// <summary>
+        /// We don't have access to Math.DivRem, so this is a copy of the implementation.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static long DivMod(long numerator, long denominator, out long modulo)
+        {
+            long div = numerator / denominator;
+            modulo = numerator - (div * denominator);
```
ahsonkhan: Is there a perf difference between this and just using the % operator?

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
```diff
@@ -0,0 +1,144 @@
...
+            return div;
+        }
+
+        #endregion Math Helper methods
+
+        #region Character counting helper methods
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int CountDigits(long n)
```
ahsonkhan: I tried this implementation and it is faster:
```C#
n = Math.Abs(n);
int length = 1;
while ((n /= 10) >= 1)
{
    length++;
}
return length;
```

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
```diff
@@ -0,0 +1,144 @@
...
+using System.Runtime.CompilerServices;
+
+namespace System.Text
+{
+    // All the helper methods in this class assume that the by-ref is valid and that there is
+    // enough space to fit the items that will be written into the underlying memory. The calling
+    // code must have already done all the necessary validation.
+    internal static class FormattingHelpers
+    {
+        private const int FractionDigits = 7;
```
ahsonkhan: Why is this 7? Can you explain what FractionDigits represents?

src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf16.cs
```diff
@@ -0,0 +1,322 @@
...
+            Unsafe.Add(ref utf16Bytes, 5) = Slash;
+
+            FormattingHelpers.WriteDigits(value.Year, 4, ref utf16Bytes, 6);
+            Unsafe.Add(ref utf16Bytes, 10) = Space;
+
+            FormattingHelpers.WriteDigits(value.Hour, 2, ref utf16Bytes, 11);
+            Unsafe.Add(ref utf16Bytes, 13) = Colon;
+
+            FormattingHelpers.WriteDigits(value.Minute, 2, ref utf16Bytes, 14);
+            Unsafe.Add(ref utf16Bytes, 16) = Colon;
```
ahsonkhan: Do we get 16 from adding 2 and 14 from the call to WriteDigits? If so, should this be a calculation stored in a variable?
```C#
digitCount = 2;
index = 14;
Unsafe.Add(ref utf16Bytes, digitCount + index) = Colon;
```

src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf8.cs
```diff
@@ -0,0 +1,315 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Text
+{
+    internal static class InvariantUtf8TimeFormatter
```
ahsonkhan: Does it make sense to extract out the duplicate code between InvariantUtf16TimeFormatter and InvariantUtf8TimeFormatter into helpers?

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter_time.cs
```diff
@@ -1,69 +1,39 @@
...
                 case 'R':
                     return TryFormatDateTimeRfc1123(value.UtcDateTime, buffer, out bytesWritten, encoder);
+
                 case 'O':
-                    return TryFormatDateTimeFormatO(value.UtcDateTime, false, buffer, out bytesWritten, encoder);
+                    return TryFormatDateTimeFormatO(value.DateTime, value.Offset, buffer, out bytesWritten, encoder);
+
                 case 'G':
-                    return TryFormatDateTimeFormatG(value.DateTime, buffer, out bytesWritten, encoder);
+                    return TryFormatDateTimeFormatG(value.DateTime, offset, buffer, out bytesWritten, encoder);
```
ahsonkhan: If format.Symbol == 'G' and format.IsDefault == false, offset = NullOffset. Do we want that?

tests/System.Text.Primitives.Tests/Formatting/TimeTests.cs
```diff
@@ -0,0 +1,184 @@
...
+            var expected = ts.ToString(format.ToString(), CultureInfo.InvariantCulture);
+
+            var span = new Span<byte>(new byte[256]);
+            Assert.True(PrimitiveFormatter.TryFormat(ts, span, out int written, format, encoder));
+
+            var actual = TestHelper.SpanToString(span.Slice(0, written), encoder);
+            Assert.Equal(expected, actual);
+        }
+
+        static readonly Random Rnd = new Random();
```
ahsonkhan: Use a seed value for consistency between runs.

src/System.Text.Primitives/System/Text/Formatting/Internal/UuidFormatter.Utf16.cs
```diff
@@ -0,0 +1,92 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Text
+{
+    internal static class InvariantUtf16UuidFormatter
```
ahsonkhan: Rename as GUID instead of UUID?

ahsonkhan:
Otherwise, LGTM.

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
shiftylogic: In most cases, that out param is used since that is the purpose of this function. However, this particular usage is actually leftover from when I was using a custom implementation of DivMod10 that was faster but obscure and only worked on positive integers up to 1 billion or so. I ended up reverting to the Math.DivRem implementation for easier maintenance.

I will switch this to a straight division since my optimized DivMod10 is gone now.

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
shiftylogic: The % operator is *really* expensive. The compiler should do this for us (like the C++ compiler does), but it doesn't. This trick is actually coded into the Math library as a DivRem method, but I don't have access to it here. My testing shows this to be significantly faster.

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
shiftylogic: I saw different results on my dev machine. I might re-visit this. Though, if I change it, I will likely just unroll the loop completely.

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
shiftylogic: The spec for fractions on time formats is fixed at exactly 7 characters. I have no idea why this is, but it is part of most time specs. I'll add a comment to this.

src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf16.cs
shiftylogic: Extra variables means more IL instructions. These are constants and therefore can be optimized.

src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf8.cs
shiftylogic: They are fundamentally different enough that you can't merge them without using generics or some level of indirection since they operate using different sized characters and types. The point of separating them was to avoid indirection and abstraction.

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter_time.cs
shiftylogic: That is exactly what I expect.

shiftylogic: FYI, the way this is done mirrors pretty closely how DateTime.ToString works.

shiftylogic: I have tests for the "default" case and this is passing based on what DateTimeOffset.ToString returns. For some reason, the system DateTime.ToString "default" (though it is documented as 'G') actually does something slightly different than explicitly using 'G'. I don't claim to understand why, but these APIs now follow exactly what the system does in all the cases we support.

tests/System.Text.Primitives.Tests/Formatting/TimeTests.cs
shiftylogic: I thought about that, but I don't know if I like it. This will have better coverage over time, and in failure cases, the logs will have enough information to replicate the failure if we end up hitting a bad case.

I could be swayed, if this is something we MUST do.

src/System.Text.Primitives/System/Text/Formatting/Internal/FormattingHelpers.cs
ahsonkhan: Sounds good.

src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf16.cs
ahsonkhan: Alright. Maybe add a comment to make the relationship explicit then and explain what the constants are? I am just thinking to make it easier in terms of maintainability or code changes/refactoring.

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter_time.cs
ahsonkhan: That seems odd, but I suppose we should match the DateTime.ToString behaviour.
I wonder if the source can provide some insight (I haven't looked deeply into this, but for reference):
http://referencesource.microsoft.com/#mscorlib/system/globalization/datetimeformat.cs,908

tests/System.Text.Primitives.Tests/Formatting/TimeTests.cs
ahsonkhan: I think it makes sense to keep test runs consistent. If the test does catch a bug, we want it to be deterministic and always fail rather than seeing intermittent failures which could cause confusion (scenario: seeing a red in CI, re-test assuming some CI infra issue, and it goes green). If we want higher coverage, should we be explicit in our test cases and brute force a large input set?

> better coverage over time

I suppose that would be OK but imo being explicit in the edge cases/etc would be better.

src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf8.cs
ahsonkhan: Aren't these parts of the methods exactly the same (aside from the difference that one takes a ref char while the other takes ref byte)?

https://github.com/shiftylogic/corefxlab/blob/9b46b6344f2feae9e4ffdff1fc074ec42203dad3/src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf8.cs#L76-L113

https://github.com/shiftylogic/corefxlab/blob/9b46b6344f2feae9e4ffdff1fc074ec42203dad3/src/System.Text.Primitives/System/Text/Formatting/Internal/TimeFormatter.Utf16.cs#L78-L115

shiftylogic: One writes char, the other byte. Without generics, how would you make this common? If they were each operating on the same sized span and writing the same data, it would be trivial.

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter_time.cs
shiftylogic: My implementations are derived from that code. That is how I found the difference. For some reason, they treat the default (i.e. no format specifier) as a modified 'G'.

tests/System.Text.Primitives.Tests/Formatting/UuidTests.cs
```diff
@@ -0,0 +1,68 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Globalization;
+using Xunit;
+
+namespace System.Text.Primitives.Tests
+{
+    public class UuidTests
```
ahsonkhan: Rename as guid? Along with filename.

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.Guid.cs
```diff
@@ -0,0 +1,20 @@
...
+        public static bool TryFormat(this Guid value, Span<byte> buffer, out int bytesWritten, TextFormat format = default(TextFormat), TextEncoder encoder = null)
+        {
+            encoder = encoder == null ? TextEncoder.Utf8 : encoder;
+
+            if (encoder.IsInvariantUtf8)
+                return InvariantUtf8GuidFormatter.TryFormat(value, buffer, out bytesWritten, format);
+            else if (encoder.IsInvariantUtf16)
+                return InvariantUtf16GuidFormatter.TryFormat(value, buffer, out bytesWritten, format);
+            else
+                throw new NotImplementedException();
```
KrzysztofCwalina: It would be good to have issues filed (or some other tracking mechanism) for all the NotImplementedExceptions.

shiftylogic: This should already be covered by issue #57. There are variants of that issue for parsing and encoding as well.

Issue #1466, opened 2017-04-14
Expose TextEncoder.TryComputeEncodedBytes on the public API


mgravell:
It is often the case that you need to know the length *before* you start writing, for network protocol purposes; the new `TextEncoder` API does not expose an API for this that is consistent with the inputs allowed of `TryEncode`.

Suggest: add appropriate `abstract bool TryComputeEncodedBytes(..., out int bytesNeeded)` methods that proxy through to the pre-existing `internal` implementations where they already exist, allowing this to be surfaced on the public API.

Example PR #1467 inbound.

PR #1467, opened 2017-04-14
TextEncoder: expose TryComputeEncodedBytes on the public API

mgravell:
related to #1466

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
@@ -152,9 +152,23 @@ namespace System.Text.Utf16
             return true;
         }
 
+        public override bool TryComputeEncodedBytes(ReadOnlySpan<byte> utf8, out int bytesNeeded)
+        {
+            int characters;
+            var result = Utf8Encoder.TryComputeStringLength(utf8, out characters);
+            bytesNeeded = characters * 2;
+            return result;
+        }
+
         public override bool TryComputeEncodedBytes(ReadOnlySpan<char> utf16, out int bytesNeeded)
         {
             bytesNeeded = utf16.Length * sizeof(char);
+            return true;
+        }
+
+        public override bool TryComputeEncodedBytes(string text, out int bytesNeeded)
+        {
+            bytesNeeded = text.Length * sizeof(char);
             return true;
         }
 
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
@@ -152,6 +152,15 @@ namespace System.Text.Utf16
             return true;
         }
 
+        public override bool TryComputeEncodedBytes(ReadOnlySpan<char> utf16, out int bytesNeeded)
+        {
+            bytesNeeded = utf16.Length * sizeof(char);
+            return true;
+        }
+
+        public override bool TryComputeEncodedBytes(ReadOnlySpan<uint> utf32, out int bytesNeeded)
+            => Utf16LittleEndianEncoder.TryComputeEncodedBytes(utf32, out bytesNeeded);
+
         #endregion Encoding implementation
     }
 }
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
@@ -154,8 +154,7 @@ namespace System.Text.Utf16
 
         public override bool TryComputeEncodedBytes(ReadOnlySpan<byte> utf8, out int bytesNeeded)
         {
-            int characters;
-            var result = Utf8Encoder.TryComputeStringLength(utf8, out characters);
+            var result = Utf8Encoder.TryComputeStringLength(utf8, out int characters);
             bytesNeeded = characters * sizeof(char);
             return result;
         }
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
@@ -156,7 +156,7 @@ namespace System.Text.Utf16
         {
             int characters;
             var result = Utf8Encoder.TryComputeStringLength(utf8, out characters);
-            bytesNeeded = characters * 2;
+            bytesNeeded = characters * sizeof(char);
             return result;
         }
 
--- a/tests/System.Text.Primitives.Tests/Encoding/RandomTests.cs
+++ b/tests/System.Text.Primitives.Tests/Encoding/RandomTests.cs
@@ -789,5 +789,14 @@ namespace System.Text.Primitives.Tests
                 Assert.Equal(expectedBytes, actual);
             }
         }
+
+        [Fact]
+        public void TryComputeEncodedBytesIllegal_Utf8()
+        {
+            string text = Utf8.Tests.Utf8EncoderTests.GenerateOnlyInvalidString(20);
+
+            int bytes;
+            Assert.False(TextEncoder.Utf8.TryComputeEncodedBytes(text, out bytes));
+        }
     }
 }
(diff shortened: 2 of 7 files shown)


src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
```diff
@@ -152,6 +152,15 @@ public override bool TryEncode(string text, Span<byte> encodedBytes, out int byt
             return true;
         }
 
+        public override bool TryComputeEncodedBytes(ReadOnlySpan<char> utf16, out int bytesNeeded)
```
shiftylogic: There is a missing scenario for calculating the encoded bytes from a UTF-8 buffer to whatever encoding the encoder is targeting. The pattern for TryEncode/TryDecode was overloads for UTF-8, UTF-16, and UTF-32.

mgravell:
good spot! I can fix that in the PR if the concept is OK.

shiftylogic:
I'm cool with exposing these functions.

@KrzysztofCwalina ?

mgravell:
@shiftylogic methods taking utf8 added in https://github.com/dotnet/corefxlab/pull/1467/commits/6f934e909b06023e7379d5ca2ee5602c24750e30 - all the logic already existed, was just plumbing

mgravell:
by the way, I'm totally open to naming discussions - I have zero investment in the name, other than it was consistent with the internal implementation

tests/System.Text.Primitives.Tests/Encoding/RandomTests.cs
```diff
@@ -725,5 +725,69 @@ public void ConstructingFromUtf16StringWithEmptyStringDoesNotAllocate()
...
+        public void TryComputeEncodedBytesShouldMatchEncoding_Utf16(string value)
+            => TryTryComputeEncodedBytesShouldMatchEncoding(value, TextEncoder.Utf16, Encoding.Unicode);
+
+        static unsafe void TryTryComputeEncodedBytesShouldMatchEncoding(string value, TextEncoder encoder, Encoding encoding)
+        {
+            int expectedBytes = encoding.GetByteCount(value);
+
+            // test via string input
+            int actual;
+            Assert.True(encoder.TryComputeEncodedBytes(value, out actual));
```
ahsonkhan: Can you add a test case where this will return false as well?

mgravell: @ahsonkhan added - note that I also had to fix the pre-existing internal implementation, as it could throw rather than return false sometimes

src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16TextEncoderLE.cs
```diff
@@ -152,6 +152,29 @@ public override bool TryEncode(string text, Span<byte> encodedBytes, out int byt
             return true;
         }
 
+        public override bool TryComputeEncodedBytes(ReadOnlySpan<byte> utf8, out int bytesNeeded)
+        {
+            int characters;
+            var result = Utf8Encoder.TryComputeStringLength(utf8, out characters);
```
davidfowl: needs moar c# 7 out int1

mgravell: If we're going c# 7, personally I'd love me some tuples on these Try methods, rather than "returns bool, one or more outs"...

davidfowl: Thats public API 😄

mgravell: Yes. It would still work, though, and is arguably more friendly than the out approach. Is the house policy "tuples on internal API only"? If it is: fine

mgravell: addressed in https://github.com/dotnet/corefxlab/pull/1467/commits/a9a906bfedb0b626ae4f355966326aa634b96a6e

KrzysztofCwalina:
why do we need these methods? So far the philosophy has been to use TryEncode methods and then either slice, or recall the method to endoce more, and then either stitch or copy. 
I worry that these methods will encourage people to do encoding twice (for all practical purposes)

ahsonkhan:
> why do we need these methods?

> From @mgravell, issue #1466:
It is often the case that you need to know the length *before* you start writing, for network protocol purposes; the new `TextEncoder` API does not expose an API for this that is consistent with the inputs allowed of `TryEncode`.

> I worry that these methods will encourage people to do encoding twice (for all practical purposes)

@mgravell, @shiftylogic: thoughts?

mgravell:
@KrzysztofCwalina  as I noted in #1466 - sometimes you really need it. To use a specific example, I'm currently "kicking the tires" of the span/pipelines/etc code by porting protobuf-net to it. As it happens, the protobuf format uses byte-length-prefixed string encoding. Now, you *could* say "reserve the space, encode into the +4 area, then backfill the length prefix when you know it" - but here's the kicker protobuf uses variable length "varint" encoding for the length - which could be between 1 and 5 bytes long (unless I do something really nasty but technically legal in the spec, but: eugh).

Obviously one can try to optimize for the short case (length prefixes of <= 127 can be written in a single byte) - and you can see what I'm doing [https://github.com/mgravell/protobuf-net/blob/async/src/AsyncSpike/SimpleUsage.cs](here), in the `PipeWriter.WriteStringWithLengthPrefix` method - if the string is < 128 characters, I *hope* that it will encode in < 128 bytes, and give it a whirl without calculating the length first - and write the single byte in afterwards. However, for longer strings I would *either* need to know the **exact** length ahead of time, or I'd need to try encoding it multiple times speculatively. The other problem is that if I have to reserve and backfill the length prefix, then I **can't commit or flush that data** - I'd have to buffer everything in memory.

To make it even more fun: when encoding sub-objects, the *object itself* is byte-length-prefixed; I need to know the exact length of an *entire object* before I start writing it. If that object contains strings, or contains sub-objects that contain strings, then I need to know *their* encoded length *without* actually needing to encode anything - just to write the object prefix. You can see how this works in `NullWriter.WriteStringWithLengthPrefix`.

So: does that explain my motivation from a real world scenario?

Obviously in both the cases that I've outlined, I'm *currently* using `string` and `Encoding.GetByteCount` - but that kinda isn't the point; I'm exploring *what I have* - everything I've said remains true of many general purpose network protocols and serialization formations. Sure, some use sentinel suffixes (newline, nil, etc) - but not all.

mgravell:
@KrzysztofCwalina as a slightly off-topic aside; the *other* way for me to address this problem *in the specific case of pipelines* is to re-implement https://github.com/davidfowl/Channels/pull/86/files which includes a mechanism to take a `Pipe` reservation (for the maximum size required), encode past the reservation, then backfill the reservation and truncate the reservation if it turned out to be oversized. This is something I'll need to ask @davidfowl  about - is that API still ok in concept? The fun thing about that API/PR is that it was *also* one of the motivations behind `IsSliceOf` (https://github.com/dotnet/corefxlab/pull/830) - see, all this still is holistically inter-connected :) So: **in the specific case of pipelines**, and **if** we get `IsSliceOf` merged, and **if** I re-implement the pipelines reserve/truncate API, **then** I can do everything without needing to calculate the length first.

However, that **only** addresses pipelines - it doesn't help at all for the more general case of "writing stuff to a buffer" - and very deliberately one of the other targets in protobuf-net's `AsyncSpike` branch is: `BufferWriter`; so I'd *still* need it for that :)

mgravell:
@KrzysztofCwalina  FWIW - I *am* already optimizing for the "the worst case and best case encodings would both have the same size prefix", but: that just isn't enough in all cases, because a lot of times they best and worst case encodings will be different prefix sizes.

In other news: how do you feel about something that exposes the per-encoding min/max encoded length for a given number of characters? very much like `Encoding.GetMaxByteCount` (except ... with sensible return values, because *damn* that API is weird).

At the moment I just use hard knowledge of the specific encoding (and a quick code-check of `GetNumberOfEncodedBytes`) and some `int maxPayloadBytes = value.Length << 2;` fun - that doesn't allow for abstraction over a `TextEncoder`, though.

Emphasis: I don't actively need this (my hack is fine) - just asking whether it is another related missing part of this puzzle.

KrzysztofCwalina:
Have you tried to encode to a temporary buffer and then copy into the pipe?

KrzysztofCwalina:
> I'm cool with exposing these functions.

Sounds good to me. Please feel free to merge.

Drawaes:
I have found memories of reservations in channels. At the moment I take a buffer at a point and go back and write the header which is mucky. Length prefixed protocols are very common and the content can be large enough that you don't want to write it and then copy as it defeats the zero copy ethos.

PR #1473, opened 2017-04-18
Fix memory pool broken by virtualiztion change

pakrym:


src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -84,6 +84,12 @@ public override string ToString()
             return builder.ToString();
         }
 
+        internal void Initialize()
+        {
+            // This is VERY bad, but is required while there is no re-initialization support
+            base.Dispose(false);
```
ahsonkhan: Can you help me understand why this is bad?

pakrym: https://msdn.microsoft.com/en-us/library/b1yfkh5e(v=vs.110).aspx 

> The Boolean parameter disposing indicates whether the method was invoked from the IDisposable.Dispose implementation or from the finalizer. The Dispose(bool) implementation should check the parameter before accessing other reference objects (e.g., the resource field in the preceding sample). Such objects should only be accessed when the method is called from the IDisposable.Dispose implementation (when the disposing parameter is equal to true). If the method is invoked from the finalizer (disposing is false), other objects should not be accessed. The reason is that objects are finalized in an unpredictable order and so they, or any of their dependencies, might already have been finalized.

PR #1465, opened 2017-04-13
Making OwnedBuffer virtual and adding a virtual Pin method

ahsonkhan:
cc @KrzysztofCwalina, @shiftylogic, @davidfowl

PR #1475, opened 2017-04-18
Adding IsDisposed checks to Span getter and GetArray/Pointer Internal

ahsonkhan:
cc @KrzysztofCwalina, @pakrym, @davidfowl

Addressing comments:
https://github.com/dotnet/corefxlab/pull/1465#discussion_r112050277
https://github.com/dotnet/corefxlab/pull/1465#discussion_r112050352

### DIFF
--- a/samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
+++ b/samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Buffers;
 using System.Collections.Sequences;
+using System.Runtime;
 
 namespace Microsoft.Net.Http
 {
@@ -41,7 +42,7 @@ namespace Microsoft.Net.Http
         {
             get
             {
-                if (IsDisposed) ThrowObjectDisposed();
+                if (IsDisposed) ThrowHelper.ThrowObjectDisposedException(nameof(OwnedBuffer));
                 return _array;
             }
         }
@@ -125,14 +126,9 @@ namespace Microsoft.Net.Http
             _next = null;
         }
 
-        public override Span<byte> GetSpan(int index, int length)
-        {
-            return Span.Slice(index, length);
-        }
-
         protected override bool TryGetArrayInternal(out ArraySegment<byte> buffer)
         {
-            if (IsDisposed) ThrowObjectDisposed();
+            if (IsDisposed) ThrowHelper.ThrowObjectDisposedException(nameof(OwnedBuffer));
             buffer = new ArraySegment<byte>(_array);
             return true;
         }
--- a/samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
+++ b/samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
@@ -37,7 +37,14 @@ namespace Microsoft.Net.Http
 
         public override int Length => _array.Length;
 
-        public override Span<byte> Span => _array;
+        public override Span<byte> Span
+        {
+            get
+            {
+                if (IsDisposed) ThrowObjectDisposed();
+                return _array;
+            }
+        }
 
         public int CopyTo(Span<byte> buffer)
         {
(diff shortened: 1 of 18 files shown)


samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
```diff
@@ -120,12 +127,12 @@ protected override void Dispose(bool disposing)
 
         public override Span<byte> GetSpan(int index, int length)
```
pakrym: Why do we need this method? What is the case where it's implementation is different from slicing `.Span`?

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
```diff
@@ -120,12 +127,12 @@ protected override void Dispose(bool disposing)
...
         {
-            if (IsDisposed) ThrowObjectDisposed();
             return Span.Slice(index, length);
         }
 
         protected override bool TryGetArrayInternal(out ArraySegment<byte> buffer)
         {
+            if (IsDisposed) ThrowObjectDisposed();
```
pakrym: Having `ThrowObjectDisposed` and `ThrowIndexOutOfRange` on a public abstract class is bad, use `ThrowHelper`

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
```diff
@@ -71,6 +71,24 @@ static void TestLifetime(OwnedBuffer<byte> owned)
         }
 
 
+        [Fact]
+        public void TestThrowOnAccessAfterDipose()
+        {
+            var array = new byte[1024];
+            OwnedBuffer<byte> owned = array;
```
pakrym: What about all other `Owned*` implemenations?

ahsonkhan: Many are in samples or experimental. I will add tests for other implementations in System.Buffer.Primitives.

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
ahsonkhan: Yes, we can remove the method.

src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -39,7 +39,14 @@ protected MemoryPoolBlock(MemoryPool pool, MemoryPoolSlab slab, int offset, int
 
         public override int Length => _length;
 
-        public override Span<byte> Span => new Span<byte>(Slab.Array, _offset, _length);
+        public override Span<byte> Span
+        {
+            get
+            {
+                if (IsDisposed) ThrowHelper.ThrowObjectDisposedException(nameof(MemoryPoolBlock));
+                return new Span<byte>(Slab.Array, _offset, _length);
```
KrzysztofCwalina: Is there a perfromance difference between the Span ctor that takes array vs the one that takes a pointer? Might be worth filing and issue to find it out, i.e. don't block this PR on that investigation.

src/System.IO.Pipelines/ThrowHelper.cs
```diff
@@ -52,6 +52,11 @@ public static void ThrowArgumentOutOfRangeException_BufferRequestTooLarge(int ma
             throw GetArgumentOutOfRangeException_BufferRequestTooLarge(maxSize);
         }
 
+        public static void ThrowObjectDisposedException(string objectName)
+        {
+            throw GetObjectDisposedException(objectName);
+        }
+
```
KrzysztofCwalina: BTW, the reason I had these methods on OwnedBuffer (as opposed on throw helper) is that the ThrowHelper is no public, and so each assembly that implements OwnedBuffer needs its own ThrowHelper. @pakrym.

KrzysztofCwalina:
in general looks good.

src/System.IO.Pipelines/ThrowHelper.cs
benaadams: Trouble is OwnedBuffer is generic `<T>` which means all T's get there own ThrowHelper as well

pakrym: And having throw helpers on public API is no good

pakrym:
@KrzysztofCwalina  We need to figure our OwnedBuffer recycling and fix dispose implementation therer.

src/System.IO.Pipelines/ThrowHelper.cs
KrzysztofCwalina: Yeah, good point about the cost of generic instantiations

PR #1480, opened 2017-04-19
Optimize decode

mgravell:
Adds a baseline **only** to illustrate decode performance


mgravell:
dammit, should have started from master; FML

PR #1485, opened 2017-04-20
Rename ThrowHelper classes so they don't clash when merged into on assembly

pakrym:


ahsonkhan:
> Looks awful but get it in so we can get unbroken.

We can revert this change after 2.0.

PR #1489, opened 2017-04-20
Add more cancellation tests

pakrym:


tests/System.IO.Pipelines.Tests/ReadAsyncCancellationTests.cs
```diff
@@ -338,6 +338,22 @@ public void ReadAsyncCompletedAfterPreCancellation()
...
+        public void ReadAsyncReturnsIsCancelOnCancelPendingReadBeforeGetResult()
+        {
+            var awaitable = Pipe.Reader.ReadAsync();
+            Assert.False(awaitable.IsCompleted);
+            awaitable.OnCompleted(() => {});
+
+            Pipe.Writer.WriteAsync(new byte[] {});
+            Pipe.Reader.CancelPendingRead();
+
+            Assert.True(awaitable.IsCompleted);
```
halter73: Why isn't this assert in the other test?

pakrym: No reason

tests/System.IO.Pipelines.Tests/FlushAsyncCancellationTests.cs
```diff
@@ -215,6 +215,25 @@ public void FlushAsyncCompletedAfterPreCancellation()
...
+        {
+            var writableBuffer = Pipe.Writer.Alloc(MaximumSizeHigh);
+            writableBuffer.Advance(MaximumSizeHigh);
+            var awaitable = writableBuffer.FlushAsync();
+
+            Assert.False(awaitable.IsCompleted);
+            awaitable.OnCompleted(() => {});
+
+            Pipe.Reader.Advance(Pipe.Reader.ReadAsync().GetResult().Buffer.End);
+            Pipe.Writer.CancelPendingFlush();
```
halter73: I think it makes sense to test flipping the order of the above two lines to verify Advance won't clobber the canceled state.

pakrym: Adding

tests/System.IO.Pipelines.Tests/FlushAsyncCancellationTests.cs
```diff
@@ -215,6 +215,44 @@ public void FlushAsyncCompletedAfterPreCancellation()
         }
 
         [Fact]
+        public void FlushAsyncReturnsIsCancelOnCancelPendingFlushBeforeGetResult()
```
cesarbs: nit: `ReturnsIsCancelled`

PR #1491, opened 2017-04-21
Add On[Reader|Writer]Completed callbacks to pipe

pakrym:


pakrym:
Do you want two methods on `IScheduller`? With and without state?

davidfowl:
Just one. With state.

pakrym:
What do we do with awaitable completions, they are `Action`'s? How to turn them into `Action<object>`?

davidfowl:
> What do we do with awaitable completions, they are Action's? How to turn them into Action<object>?

We aren't changing awaitable completions. But the conversion is simple. Just pass the callback as the state object. Or am I misunderstanding.

pakrym:
Adding another level of indirection to the hot path. I'll measure it.

halter73:
Can we just disallow more than one of each completed callback in order to easily add a state parameter? It's a little gross, but if we we're going to allocate anyway we could use a Task.

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,30 +1,75 @@
...
+                _exception = exception ?? _completedNoException;
+            }
+            return _callbackCount > 0;
+        }
+
+        public void AttachCallback(Action<Exception, object> callback, object state)
+        {
+            if (IsCompleted)
+            {
+                PipelinesThrowHelper.ThrowInvalidOperationException(ExceptionResource.AttachingToCompletedPipe);
```
davidfowl: This should not throw. It should just fire immediately.

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,30 +1,75 @@
...
-            Interlocked.CompareExchange(ref _exception, exception ?? _completedNoException, null);
+            if (_exception == null)
+            {
+                // Set the exception object to the exception passed in or a sentinel value
+                _exception = exception ?? _completedNoException;
+            }
+            return _callbackCount > 0;
+        }
+
+        public void AttachCallback(Action<Exception, object> callback, object state)
```
davidfowl: This needs to be thread safe.

davidfowl: I see it happens in the lock,  good!

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -38,21 +83,49 @@ public bool IsCompletedOrThrow()
             return false;
         }
 
-        private void ThrowFailed()
+        public void InvokeCallbacks()
         {
-            throw _exception;
+            Debug.Assert(IsCompleted);
+
+            if (_callbacks == null)
```
davidfowl: These are fire once callbacks right? Why not clear out the list here after invoking?

```
var callbacks = _callbacks;
_callbacks = null;
```

davidfowl: Ah, I see we reuse the array.

src/System.IO.Pipelines/PipelinesThrowHelper.cs
```diff
@@ -153,6 +153,9 @@ private static string GetResourceString(ExceptionResource argument, string locat
                 case ExceptionResource.BackpressureDeadlock:
                     resourceString = "Advancing examined to the end would cause pipe to deadlock because FlushAsync is waiting";
                     break;
+                case ExceptionResource.AttachingToCompletedPipe:
```
davidfowl: Remove

src/System.IO.Pipelines/TaskRunScheduler.cs
```diff
@@ -10,9 +10,9 @@ public class TaskRunScheduler : IScheduler
     {
         public static TaskRunScheduler Default = new TaskRunScheduler();
 
-        public void Schedule(Action action)
+        public void Schedule(Action<object> action, object state)
         {
-            Task.Run(action);
+            Task.Factory.StartNew(action, state);
```
davidfowl: 👍

tests/System.IO.Pipelines.Tests/UnownedBufferReaderFacts.cs
```diff
@@ -619,6 +619,11 @@ public void Complete(Exception exception = null)
                 _pipe.Reader.Complete(exception);
             }
 
+            public void OnWriterCompleted(Action<Exception, object> callback, object state)
+            {
+                throw new NotImplementedException();
```
davidfowl: Fix

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
```diff
@@ -0,0 +1,333 @@
...
+            {
+                continuationRan = true;
+            });
+            pipe.Reader.Complete();
+
+            Assert.True(callbackRan);
+        }
+
+
+        private class CallCountScheduler:IScheduler
```
davidfowl: nit formatting

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
```diff
@@ -0,0 +1,333 @@
...
+            Assert.False(awaiter.IsCompleted);
+            awaiter.OnCompleted(() =>
+            {
+                continuationRan = true;
+            });
+            pipe.Reader.Complete();
+
+            Assert.True(callbackRan);
+        }
+
```
davidfowl: nit:empty space

src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
```diff
@@ -243,6 +243,11 @@ void IPipeReader.Complete(Exception exception)
             }
         }
 
+        public void OnWriterCompleted(Action<Exception, object> callback, object state)
+        {
+            throw new NotImplementedException();
```
davidfowl: Hmmmm, we should fix this. Do a simple implementation based on task. Store a tcs internally and make that work.

src/System.IO.Pipelines/PipeCompletion.cs
davidfowl: Rename to AddCallback

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -38,21 +83,49 @@ public bool IsCompletedOrThrow()
...
+                    // we do not allow registering null callbacks
+                    // safe to assume we reached end of callback list
+                    break;
+                }
+                completionCallback.Callback.Invoke(_exception, completionCallback.State);
+            }
         }
 
-        public string Location
+        public void Reset()
```
davidfowl: Reset needs to happen inside of a lock.

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
```diff
@@ -0,0 +1,333 @@
...
+            _pipeFactory = new PipeFactory();
+        }
+
+        public void Dispose()
+        {
+            _pipeFactory.Dispose();
+        }
+
+        [Fact]
+        public void OnReaderCompletedThrowsAfterComplete()
```
davidfowl: Shouldn't throw.

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
```diff
@@ -0,0 +1,333 @@
...
+        {
+            var pipe = _pipeFactory.Create();
+            pipe.Reader.Complete();
+
+            Assert.Throws<InvalidOperationException>(() => pipe.Writer.OnReaderCompleted((exception, state) => { }, null));
+        }
+
+        [Fact]
+        public void OnWriterCompletedThrowsAfterComplete()
```
davidfowl: Shouldn't throw.

davidfowl:
We call TryComplete in this one exceptional case and it won't fire the callback https://github.com/dotnet/corefxlab/blob/179c97da84b55e7d095927cb3c065eab1e336c74/src/System.IO.Pipelines/PipeAwaitable.cs#L92.

src/System.IO.Pipelines/PipeCompletion.cs
benaadams: Shame can't add `Debug.Assert(Monitor.IsEntered(_sync))`

src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
pakrym: This PR is big enough, and I don't want to write tests for `UnownedBufferReader`. Lets decide if we want to keep it at all.

pakrym:
@davidfowl I will look at that case, that code is old and strange

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,30 +1,75 @@
...
 
-using System.Threading;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
 namespace System.IO.Pipelines
 {
     internal struct PipeCompletion
     {
+        private const int InitialCallbacksSize = 4;
```
halter73: Could we just make this 1?

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,30 +1,75 @@
...
 #endif
+        private Exception _exception;
+        private PipeCompletionCallback[] _callbacks;
+        private int _callbackCount;
+
+        public string Location
+        {
+            get
+            {
+#if COMPLETION_LOCATION_TRACKING
```
halter73: Having just used this, I think we should have a single define for all "LOCATION_TRACKING".

src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
davidfowl: Just add the feature instead of throwing.

src/System.IO.Pipelines/PipeCompletion.cs
pakrym: I'll implement it as a different PR.

ahsonkhan:
@dotnet-bot test this please

pakrym:
:up::date:

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,58 +1,137 @@
...
+            foreach (var completionCallback in _callbacks)
+            {
+                if (completionCallback.Callback == null)
+                {
+                    // we do not allow registering null callbacks
+                    // safe to assume we reached end of callback list
+                    break;
+                }
+
+                completionCallback.Callback.Invoke(Exception, completionCallback.State);
```
halter73: Do we guard against callbacks throwing, to ensure all callbacks get invoked?

src/System.IO.Pipelines.Extensions/UnownedBufferReader.cs
```diff
@@ -243,6 +243,11 @@ void IPipeReader.Complete(Exception exception)
             }
         }
 
+        public void OnWriterCompleted(Action<Exception, object> callback, object state)
+        {
+            _readingTcs.Task.ContinueWith(task => callback(task.Exception, state));
```
halter73: Nit: Use the overload of ContinueWith that takes state.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -472,13 +481,19 @@ void IPipeReader.Complete(Exception exception)
...
             lock (_sync)
             {
+                runCompletionCallbacks = _readerCompletion.TryComplete(exception);
                 awaitable = _writerAwaitable.Complete();
             }
+
+            if (runCompletionCallbacks)
+            {
+                TrySchedule(_writerScheduler, _callReaderCompletedCallbacks, this);
+            }
```
halter73: Nit: newline.

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,58 +1,137 @@
...
+
+            if (_callbacks == null)
+            {
+                _callbacks = new PipeCompletionCallback[InitialCallbacksSize];
+            }
+
+            var newIndex = _callbackCount;
+            _callbackCount++;
+
+            if (_callbackCount == _callbacks.Length)
```
halter73: Isn't this doubling the size of the array one callback earlier than necessary. Is this just so you leave a sentinel (default) value at the end of the array for InvokeCallbacks()? It makes more sense to me to have InvokeCallbacks() use _callbackCount instead.

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,58 +1,137 @@
...
+            return true;
         }
 
-        public string Location
+        public void InvokeCallbacks()
         {
-            get
+            Debug.Assert(IsCompleted);
+
+            if (_callbacks == null)
```
halter73: It doesn't look like this method is being called with a lock which seems very bad.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -472,13 +481,19 @@ void IPipeReader.Complete(Exception exception)
...
             TrySchedule(_writerScheduler, awaitable);
 
             if (_writerCompletion.IsCompleted)
```
halter73: How is it safe to have this and `_readerCompletion.IsCompleted` outside of the lock?

src/System.IO.Pipelines/PipeCompletion.cs
halter73: This is particularly funny now that InitialCallbacksSize is 1. The first callback to be added causes a 1 element array to be allocated only to be immediately discarded to be replaced with a 2 element array. :rofl:

src/System.IO.Pipelines/Pipe.cs
davidfowl: I think these do need to be inside of the lock. What happens if `_writerCompletion.IsCompleted` becomes true after this check? Won't it just leak memory?

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -487,6 +502,25 @@ void IPipeReader.Complete(Exception exception)
...
+
+            bool completeInline;
+            lock (_sync)
+            {
+                completeInline = _writerCompletion.AddCallback(callback, state);
+            }
+
+            if (completeInline)
+            {
+                callback(_writerCompletion.Exception, state);
```
davidfowl: Maybe schedule these?

pakrym: @halter73 has different opinion.

pakrym:
:up::date:

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -556,8 +616,18 @@ private static void TrySchedule(IScheduler scheduler, Action action)
...
             if (action != null)
             {
-                scheduler.Schedule(action);
+                scheduler.Schedule(_scheduleContinuation, action);
+            }
+        }
+
+        private static void TrySchedule(IScheduler scheduler, Action<object> action, object state)
+        {
+            if (action == null)
```
halter73: Nit: why have this if inverted but not the one in the original TrySchedule?

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,58 +1,121 @@
...
             _completionLocation = Environment.StackTrace;
 #endif
-            // Set the exception object to the exception passed in or a sentinel value
-            Interlocked.CompareExchange(ref _exception, exception ?? _completedNoException, null);
+            if (Exception == null)
+            {
+                // Set the exception object to the exception passed in or a sentinel value
+                Exception = exception ?? _completedNoException;
+            }
+            return _callbackCount > 0;
```
halter73: What's the benefit of checking the _callbackCount inside the lock? It seems to me this could be checked outside the lock if its only purpose is to avoid unnecessarily scheduling Invoke[Reader|Writer]CompletionCallbacks.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -709,18 +789,41 @@ private void WriterCancellationRequested()
             TrySchedule(_writerScheduler, action);
         }
 
+        private void InvokeReaderCompletionCallbacks()
+        {
+            PipeCompletionCallbacks action;
+            lock (_sync)
+            {
+                action = _readerCompletion.GetCallbacks();
```
halter73: GetCallbacks() should be called inline and only Execute() should be scheduled. Otherwise, the pipe could be reset before InvokeReaderCompletionCallbacks() gets run by the scheduler. If doing this requires allocating or running the callbacks inline, maybe we should just use a TCS.

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -461,6 +470,11 @@ void IPipeReader.Advance(ReadCursor consumed, ReadCursor examined)
             TrySchedule(_writerScheduler, continuation);
         }
 
+        private bool IsEverythingExamined(ReadCursor examined)
+        {
+            return examined.Segment == _commitHead && examined.Index == _commitHeadIndex;
```
pakrym: Yes

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,58 +1,126 @@
...
-using System.Threading;
+using System.Buffers;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
 
 namespace System.IO.Pipelines
 {
     internal struct PipeCompletion
     {
+        private static readonly ArrayPool<PipeCompletionCallback> CompletionCallbackPool = ArrayPool<PipeCompletionCallback>.Shared;
```
davidfowl: This is going to be super overkill for most cases 😄

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
```diff
@@ -0,0 +1,453 @@
...
+                callbackRan = true;
+            }, null);
+            pipe.Writer.Complete();
+
+            Assert.True(callbackRan);
+            Assert.Equal(1, scheduler.CallCount);
+        }
+
+        [Fact]
+        public void OnReaderCompletedExceptionSurfacesToWriterScheduler()
```
davidfowl: Who observes these exceptions?

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
```diff
@@ -0,0 +1,453 @@
...
+            pipe.Writer.Complete();
+            pipe.Reset();
+
+            Assert.True(callbackRan);
+            callbackRan = false;
+
+            pipe.Reader.Complete();
+            pipe.Writer.Complete();
+
+            Assert.Equal(1, scheduler.CallCount);
```
davidfowl: `Assert.False(callbackran)`

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -592,9 +657,14 @@ private void Dispose()
...
             Action awaitable;
+            bool doubleCompletion;
             lock (_sync)
             {
-                awaitable = _readerAwaitable.OnCompleted(continuation, ref _writerCompletion);
+                awaitable = _readerAwaitable.OnCompleted(continuation, out doubleCompletion);
+            }
+            if (doubleCompletion)
+            {
+                Writer.Complete(PipelinesThrowHelper.GetInvalidOperationException(ExceptionResource.NoConcurrentOperation));
```
davidfowl: Isn't this broken? Won't it *not* yield both continuations?

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
```diff
@@ -0,0 +1,453 @@
...
+        public PipeCompletionCallbacksTests()
+        {
+            _pipeFactory = new PipeFactory();
+        }
+
+        public void Dispose()
+        {
+            _pipeFactory.Dispose();
+        }
+
```
davidfowl: Add a test that completes the reader in the writer callback and vice versa.

src/System.IO.Pipelines/PipeCompletion.cs
```diff
@@ -1,58 +1,126 @@
...
+            if (_callbackCount == 0)
             {
-#if COMPLETION_LOCATION_TRACKING
-                return _completionLocation;
-#else
                 return null;
-#endif
             }
+
+            var callbacks = new PipeCompletionCallbacks(CompletionCallbackPool, _callbackCount, Exception, _callbacks);
```
davidfowl: You need to pass null for the exception if the exception is `_completedNoException`, right now this shows up as always having an exception.

src/System.IO.Pipelines/Pipe.cs
pakrym: It would yiled one that passed in because it got returned from OnComplete and the one registered when Complete runs

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
pakrym: No, it throws on scheduler (Exception Surfaces To Writer Scheduler)

tests/System.IO.Pipelines.Tests/PipeCompletionCallbacksTests.cs
pakrym: This is not recommended use case because it can hit "Cannot complete during writing" but I will add a test

src/System.IO.Pipelines/Pipe.cs
pakrym: AwaitingFlushAsyncAwaitableTwiceCompletesReaderWithException checks both tasks for `            Assert.Equal(true, task2.IsCompleted);
`

PR #1493, opened 2017-04-22
Attempt at new UTF-8 Decoder implementation

ahsonkhan:
Trying a new UTF-8 decoder implementation that @davidfowl pointed us to: http://bjoern.hoehrmann.de/utf-8/decoder/dfa/

**Performance:**
Here are some preliminary performance results.
charLength = 1000
Results for strings with characters from specific codepoint ranges.
image

**Questions:**
- Can this implementation be made even faster?
**Note:** I tried to keep other optimizations to a minimum to see if this approach is faster than the previous one (as is). So, I am looking for optimizations related to this change (**not** something like use [pointers](https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L329) via DangerousGetPinnableReference()).

- Keep in mind, this implementation does end up causing an allocation of a byte array of length 400 (utf8d). Is that acceptable?

**To Do:**
If this is a good idea, we should remove the old [TryDecodeCodePoint](https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L125) implementation and update all [usage/references](https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L765).

**Benchmark Method:**
```C#
public static void After(int charLength, int minCodePoint, int maxCodePoint, bool shouldOutput = true)
{
    string unicodeString = GenerateString(charLength, minCodePoint, maxCodePoint);
    ReadOnlySpan<char> characters = unicodeString.AsSpan();

    Encoding utf8 = Encoding.UTF8;

    int utf8Length = utf8.GetByteCount(unicodeString);
    byte[] utf8Buffer = new byte[utf8Length];

    char[] charArray = characters.ToArray();
    utf8.GetBytes(charArray, 0, characters.Length, utf8Buffer, 0);
    Span<byte> span = new Span<byte>(utf8Buffer);

    Encoding unicode = Encoding.Unicode;
    int utf16Length = unicode.GetByteCount(unicodeString);
    char[] utf16Buffer = new char[utf16Length];
    Span<char> output = new Span<char>(utf16Buffer);

    Stopwatch sw = new Stopwatch();
    sw.Restart();
    TryDecode(span, output, out int x, out int y);
    sw.Stop();
    if (shouldOutput) Console.WriteLine(sw.ElapsedTicks);
}
```


src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -27,6 +27,24 @@ internal static class Utf8Encoder
         private const byte b1110_0000U = 0xE0; //224
         private const byte b1111_0000U = 0xF0; //240
         private const byte b1111_1000U = 0xF8; //248
+        private const byte b1111_1111U = 0xFF; //255
+    
+        public static byte[] utf8d = {
```
benaadams: readonly

benaadams:
> Keep in mind, this implementation does end up causing an allocation of a byte array of length 400 (utf8d). Is that acceptable?

Once per application run?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -56,22 +74,29 @@ public static bool TryDecode(ReadOnlySpan<byte> utf8, Span<char> utf16, out int
...
+                        return false;
 
-                charactersWritten += written;
-                bytesConsumed += consumed;
+                    bytesConsumed = consumed;
+                    charactersWritten += written;
+                }
             }
-
+            if (state != 0) return false;
```
mgravell: nit suggestion: `return state == 0;` - won't impact perf, though

mgravell:
I like it; looks like a no-brainer win to me. Would be good to leave a url reference to `http://bjoern.hoehrmann.de/utf-8/decoder/dfa/` just above the lookup table, IMO - so people reviewing the code don't have to ask "wat?", but: very nice.

This also retains full compatibility with the "all ASCII special-case via SIMD" PR, if we want (I'm not presuming that the SIMD PR will or won't be merged at some point - just remarking that it can still work side-by-side).

The 400 bytes is perfectly reasonable and justified IMO.

jkotas:
What is the performance relative to the UTF8 decoder in CoreCLR? It is what you need to match and beat. I am pretty sure that we will end up doing same thing for it as what we have done for UTF8 encoder to match it - copy the implementation from CoreCLR (https://github.com/dotnet/corefxlab/pull/1156).

jkotas:
> Would be good to leave a url reference to http://bjoern.hoehrmann.de/utf-8/decoder/dfa/ just above the lookup table

More than that. You need to preserve the copyrights and add reference to THIRD-PARTY-NOTICES

shiftylogic:
Caveat: We haven't done any optimization work on the encoders yet.

Having said that, prior to this change, we were ~30x slower than CoreCLR for ASCII range, ~7x slower for 0x80 - 0x7FF, and ~5x slower for the rest of the code point range.

With this change, we get to ~17x, ~5x, ~4x respectively.

@jkotas is correct that this decoding algorithm is likely not going to beat the one in CoreCLR.

jkotas:
> With this change, we get to ~17x, ~5x, ~4x respectively.

So why spend time on this, and not just copy the CoreCLR implementation instead?

ahsonkhan:
@shiftylogic, I am not able to get results like yours.

image

ahsonkhan:
> I am not able to get results like yours.
Now I can see the delta (utf8 to utf16 is where there is a big difference):

image

ahsonkhan:
I wonder why UTF8 to UTF32 is faster compared to CoreCLR with this change (when UTF8 to UTF16 is so much slower). @jkotas, any ideas? Is [this](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/Text/UTF32Encoding.cs#L521) sub-optimal or is there a mistake in the testing?

jkotas:
> Is this sub-optimal or is there a mistake in the testing?

I think the DecodingUtf32PerformanceTestUsingCoreCLR is broken: it takes an span UTF8 bytes, does not portable cast on them to turn them into bogus UTF16 characters, and then measures how fast it takes to convert these bogus UTF16 characters to UTF32.

jkotas:
There is no reasonable equivalent of the TryDecode method that goes from UTF8 to UTF32 in CoreCLR because of it never showed up so far as something that we really need and that needs to be fast.

There is `Encoding.Convert`, but it is convenience method with very poor performance: it allocates internal buffer, goes through the array multiple times, ... .

ahsonkhan:
> turn them into bogus UTF16 characters, and then measures how fast it takes to convert these bogus UTF16 characters to UTF32.

Ah yes. That test is measuring UTF16 to UTF32 and not what I want, which is UTF8 to UTF32. But why is it bogus UTF16 characters?

> There is no reasonable equivalent of the TryDecode method that goes from UTF8 to UTF32 in CoreCLR 

Should we then not have it either? I.e. remove:
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/TextEncoder.cs#L152
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/TextEncoder.cs#L207

jkotas:
> But why is it "bogus" UTF16 characters?

The test takes UTF8 bytes and casts them to UTF16 characters here:
https://github.com/dotnet/corefxlab/pull/1493/files#diff-93ccc6c3a377223009399d0d86387848R80. For example, you can end up with invalid UTF16 characters that will trigger error handling paths.

Having fast methods to convert between any of UTF8/UTF16/UTF32 sounds reasonable to me. Some of the combination are likely going to be used less frequently, but it does not mean that they should not exist.

ahsonkhan:
On a related note, what about UTF16 to UTF32 conversion? Will [this](https://github.com/dotnet/coreclr/blob/master/src/mscorlib/shared/System/Text/UTF32Encoding.cs#L521) work? And is it optimized?
```C#
Encoding utf32 = Encoding.UTF32;
int byteCount = utf32.GetByteCount(unicodeString);
byte[] byteArray = new byte[byteCount];
utf32 .GetBytes(charArray, 0, characters.Length, byteArray, 0);
```

jkotas:
>  Will this work? 

I think so.

> And is it optimized?

I do not think so. But it should not prevent you from measuring it.

ahsonkhan:
> But it should not prevent you from measuring it.

I measured, and it seems slow (even with sub-optimal Utf16LittleEndianEncoder.TryDecode).
image

ahsonkhan:
I am closing this PR and we will work on porting the coreclr impl where possible.
We will try a similar approach for UTF8 to UTF32.

PR #1498, opened 2017-04-24
Abstraced OwnedBuffer lifetime management

KrzysztofCwalina:
Abstracted lifetime management (reference counting) out of OwnedBuffer\<T\>.
Added a subclass ReferenceCountedBuffer for interlocked-based reference counting.
Other subclasses, can use whatever they want for lifetime management.

After this change, OwnedBuffer has just one field left (_disposed) which can be removed quite easily (at the cost of making it more tedious to implement IDisposable on subclasses.  

cc: @mjp41, @davidfowl, @ahsonkhan, @shiftylogic


davidfowl:
@pakrym can you see if this will break the Kestrel mirroring?

pakrym:
@davidfowl Nothing obvious

PR #1499, opened 2017-04-25
Moved ReferenceCounter to System.Buffers.Experimental

KrzysztofCwalina:


KrzysztofCwalina:
cc: @shiftylogic, @ahsonkhan, @davidfowl, @joshfree

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
```diff
@@ -167,59 +167,6 @@ private static IEnumerable<object[]> MakePermutations(params object[][] argss)
         }
         #endregion
 
-        public static IEnumerable<object[]> ReservationPerformanceData =>
```
shiftylogic: Did you mean to move these into experimental tests?

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
```diff
@@ -1,6 +1,8 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System.Buffers;
+using System.Collections.Generic;
+using System.Runtime;
```
ahsonkhan: Why did we need to add these?

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
KrzysztofCwalina: Initially I wanted to, but they I started to wonder if the tests are that useful given the API is now op-in. But I can add this test back, if you think it's useful.

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
KrzysztofCwalina: oops. i will remove them. This is a leftover from when I wanted to add the perf tests there.

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
KrzysztofCwalina: One thing I forgot to explain: the tests don't work as it because AutoPooledMemory does not use ReferenceCounter

shiftylogic: If you don't think they are valuable anymore, then that is fine. I was just curious why it was removed. I thought maybe you just forgot to save the file on the other side after moving them. :)

ahsonkhan: Removing unused helpers in #1526 since this (now removed) test was the only one using them.

Issue #1437, opened 2017-04-06
Buffer<T> wrapping OwnedBuffer<T> is problematic for the BCL


stephentoub:
I started looking more at the Buffer<T> / OwnedBuffer<T> in corefxlab, and I’m concerned about the design.

Specifically, for the BCL we want to add APIs in various places that work with buffers, in order to:
- Support passing in arbitrary pointers, e.g. Stream.Write
- Support passing in array subsets, e.g. int.Parse

etc. We could add individual overloads to these methods for native pointers, for offsets+counts, but part of the goal of this whole effort has been to unify all of that.

My hope had been that we'd be able to add an individual new overload in many of these situations, taking a single type that unified all of this and that didn't pull in other dependencies.  I thought that was `Buffer<T>`.  But in looking at it today, creating a `Buffer<T>` from a pointer or an array allocates an `OwnedBuffer<T>`, which is bad.  For example, I'd want someone to be able to do a stackalloc and then call `fileStream.Read(new Buffer<byte>(ptr, len))` to read into that memory, but with the current design, that allocates.  And even if we pooled OwnedBuffers under the covers, it's still allocating and/or potentially a source of contention on a pool.  And folks shouldn't need to be maintain their own pool of OwnedBuffes to avoid such things.

In my mind, there's the fast `Span<T>`, and then `Buffer<T>` simply be the heap-able version of `Span<T>`, essentially an `ArraySegment<T>` that can also work with pointers in addition to arrays.  The BCL could operate in terms of those two simple exchange types and not need to be concerned in these APIs with types like `OwnedBuffer<T>`; it would be up to callers to use them as they needed in their system:
```C#
public struct Span<T>
{
    private ByReference<T> _ptr;
    private int _length;
    ...
}
public struct Buffer<T>
{
    private T[] _array;
    private IntPtr _ptr;
    private int _offset;
    private int _count;
    ...
}
```

[...]

PR #1500, opened 2017-04-25
Removed _disposed field from OwnedBuffer

KrzysztofCwalina:
After this change, OwnedBuffer has no fields and can be turned into an interface (if we want and perf is right)


src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
```diff
@@ -33,29 +28,22 @@ public virtual BufferHandle Pin(int index = 0)
         internal protected abstract unsafe bool TryGetPointerInternal(out void* pointer);
 
         #region Lifetime Management
```
ahsonkhan: I think we can remove this region. right?

ahsonkhan:
LGTM

KodrAus:
@KrzysztofCwalina Are these changes to `OwnedBuffer` part of solving #1437?

PR #1502, opened 2017-04-26
Simplified BufferHandle and Pinning

KrzysztofCwalina:
This change allows the called of BufferHandle ctor to decide how GCHandle is constructed. For example, permamently pinned buffers can use default GCHandle and array based buffers can create single GCHandle per OwnedBuffer which gets released when reference count goes to zero.

The main changes are in BufferHandle.Create/ctor

cc: @davidfowl, @shiftylogic

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/IRetainable.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/IRetainable.cs
@@ -3,9 +3,10 @@
 
 namespace System.Buffers
 {
-    internal interface IKnown
+    public interface IRetainable
     {
-        void AddReference();
+        void Retain();
         void Release();
+        bool IsRetained { get; }
     }
 }
--- a/src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
@@ -2,9 +2,13 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
 namespace System.Buffers
 {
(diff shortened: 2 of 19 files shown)


KrzysztofCwalina:
@dotnet-bot test this please

src/System.Buffers.Primitives/System/Buffers/IRetainable.cs
```diff
@@ -3,9 +3,10 @@
 
 namespace System.Buffers
 {
-    internal interface IKnown
+    public interface IRetainable
     {
-        void AddReference();
+        void Retain();
```
ahsonkhan: What motivated this rename?

ahsonkhan:
@dotnet-bot test this please

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
```diff
@@ -20,27 +24,50 @@ public abstract class OwnedBuffer<T> : IDisposable, IKnown
 
         public virtual BufferHandle Pin(int index = 0)
```
ahsonkhan: Should we make the method unsafe? Will a class that overrides this method ever be safe?

ahsonkhan:
LGTM

src/System.Buffers.Primitives/System/Buffers/IRetainable.cs
KrzysztofCwalina: The method is abstract. Some implementations don't actually do reference counting.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
KrzysztofCwalina: method signature unsafe is simply a shortcut for unsafe block in the implementation, i.e. it's a property of implementation despite the fact that it appears in the signature.

PR #1372, opened 2017-03-25
Added support for ReadOnlyBytes to HttpParser

KrzysztofCwalina:
After I added inner itterations, the results are more realistic:

     Test Name                               | Metric                | Iterations |    AVERAGE |    STDEV.S |        MIN |        MAX
    :--------------------------------------- |:--------------------- |:----------:| ----------:| ----------:| ----------:| ----------:
     HttpParserBench.HttpParserReadableBytes | Duration              |    101     |      0.159 |      0.743 |      0.081 |      7.555
     HttpParserBench.HttpParserReadableBytes | Instructions Retired  |    101     | 1.009E+006 | 3.768E+006 | 6.000E+005 | 3.850E+007
     HttpParserBench.HttpParserReadableBytes | Branch Mispredictions |    101     |   4542.099 |  35810.853 |      0.000 | 3.604E+005
     HttpParserBench.HttpParserReadableBytes | Cache Misses          |    101     |   1581.624 |  15895.122 |      0.000 | 1.597E+005
     HttpParserBench.HttpParserRob           | Duration              |    101     |      0.106 |      0.187 |      0.081 |      1.967
     HttpParserBench.HttpParserRob           | Instructions Retired  |    101     | 8.069E+005 | 8.850E+005 | 7.000E+005 | 9.600E+006
     HttpParserBench.HttpParserRob           | Branch Mispredictions |    101     |   1824.950 |  10641.330 |      0.000 | 1.065E+005
     HttpParserBench.HttpParserRob           | Cache Misses          |    101     |    324.436 |   2877.898 |      0.000 |  28672.000
     HttpParserBench.HttpRequestParser       | Duration              |    101     |      0.398 |      1.654 |      0.229 |     16.854
     HttpParserBench.HttpRequestParser       | Instructions Retired  |    101     | 2.563E+006 | 7.863E+006 | 1.700E+006 | 8.080E+007
     HttpParserBench.HttpRequestParser       | Branch Mispredictions |    101     |   9286.970 |  92100.618 |      0.000 | 9.257E+005
     HttpParserBench.HttpRequestParser       | Cache Misses          |    101     |   2838.812 |  27303.720 |      0.000 | 2.744E+005

HttpParserReadableBytes : uses ReadableBuffer
HttpParserRob : uses ReadOnlyBytes
HttpRequestParser : toy parser in System.Text.Http implements purely in safe code (i.e. no pointers)

PR #1503, opened 2017-04-27
Build in VS: Including some package references in csproj and clean up.

ahsonkhan:
cc @davidfowl, @shiftylogic, @KrzysztofCwalina, @crummel

I can build in VS with these changes.
I first have to restore via command line using `dotnet\dotnet.exe restore corefxlab.sln` and then build in VS.

`========== Build: 61 succeeded or up-to-date, 0 failed, 0 skipped ==========`


samples/System.IO.Pipelines.Samples/System.IO.Pipelines.Samples.csproj
```diff
@@ -21,5 +21,6 @@
   <ItemGroup>
     <PackageReference Include="Microsoft.AspNetCore.Hosting.Server.Abstractions" Version="1.1.0" />
     <PackageReference Include="Microsoft.AspNetCore.Server.Kestrel" Version="1.1.0" />
+    <PackageReference Include="System.Net.NameResolution" Version="4.4.0-beta-24913-02" />
```
davidfowl: This is pretty random. Why does it need to be version 4.4.0-beta-24913-92? Doesn't 4.3.0 work? Also, what is it required for?

tests/System.IO.Pipelines.Performance.Tests/columns/RpsColumn.cs
```diff
@@ -21,10 +21,22 @@ public string GetValue(Summary summary, Benchmark benchmark)
 
         public bool IsDefault(Summary summary, Benchmark benchmark) => false;
         public bool IsAvailable(Summary summary) => true;
+
+        public string GetValue(Summary summary, Benchmark benchmark, ISummaryStyle style)
```
davidfowl: Since you have to change this, lets get rid of it altogether and replace it with this https://github.com/aspnet/KestrelHttpServer/blob/c2f15fcac374328d8256b8345bfb5eb22e8add75/test/Microsoft.AspNetCore.Server.Kestrel.Performance/configs/CoreConfig.cs#L19

ahsonkhan: What exactly should I replace with `Add(StatisticColumn.OperationsPerSecond);`?

samples/System.IO.Pipelines.Samples/System.IO.Pipelines.Samples.csproj
ahsonkhan: It is for Dns [here](https://github.com/dotnet/corefxlab/blob/00f62cb733560d6667e897d34c54df5cae205b1c/samples/System.IO.Pipelines.Samples/HttpClient/LibuvHttpClientHandler.cs#L277).

This resolves the build error: The name 'Dns' does not exist in the current context.

4.3.0 should work: https://www.nuget.org/packages/System.Net.NameResolution/

shiftylogic: If we foresee any other projects using this assembly, we might also want to move this into the props file. It allows us to manage it with Maestro if / when we need a newer version.

tests/Benchmarks/HttpParserBench.cs
```diff
@@ -271,11 +271,11 @@ static int FullRequestLegacy()
 
 class Request : IHttpHeadersHandler, IHttpRequestLineHandler
 {
-    public Http.Method Method;
-    public Http.Version Version;
-    public string Path;
-    public string Query;
-    public string Target;
+    //public Http.Method Method;
```
shiftylogic: Why comment these out and leave them rather than completely removing them? This goes for the code that is commented out within this class (and the one underneath it). It adds no real value to leave it there.

tests/System.IO.Pipelines.Performance.Tests/columns/RpsColumn.cs
shiftylogic: I believe he is suggesting to remove RpsColumn completely and do [this](https://github.com/ahsonkhan/corefxlab/blob/cbb539ce791872e63f3af1a2e2f916078e41eb7d/tests/System.IO.Pipelines.Performance.Tests/configs/CoreConfig.cs#L16).

tests/Benchmarks/HttpParserBench.cs
ahsonkhan: See: https://github.com/dotnet/corefxlab/pull/1372#discussion_r108022459

tests/System.IO.Pipelines.Performance.Tests/columns/RpsColumn.cs
ahsonkhan: I am confused. 'Add(new RpsColumn());' is already there.

shiftylogic: Yes. Remove it, and use the (I assume) built-in ops stats column.

ahsonkhan:
@shiftylogic, @KrzysztofCwalina: Good to go?

ahsonkhan:
FYI: https://github.com/dotnet/cli/issues/6433

PR #1504, opened 2017-04-27
Port CLR utf8-to-utf16 decode implementation

shiftylogic:
* We don't support fallback like the CLR does, so that code was deleted.
* We decode as much as we can and return what we consumed, so there are minor tweaks.
* Passes all existing tests, which cover scenarios like output buffer too small, partial decode and retry from point we stopped, etc.
* Added some random decoding tests for utf-8 => utf-16 scenarios across several code point ranges.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
@@ -28,11 +28,6 @@ namespace System.Text.Utf8
         private const byte b1111_0000U = 0xF0; //240
         private const byte b1111_1000U = 0xF8; //248
 
-        // bits 30 & 31 are used for pending bits fixup
-        private const int FinalByte = 1 << 29;
-        private const int SupplimentarySeq = 1 << 28;
-        private const int ThreeByteSeq = 1 << 27;
-
         #endregion Constants
 
         #region Decoding implementation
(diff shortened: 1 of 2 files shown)


shiftylogic:
FYI, the performance tests show this to be at or slightly faster than the built-in CLR implementation now (within noise; and we don't support fallback).

@KrzysztofCwalina @ahsonkhan

ahsonkhan:
Should we add a comment like this: https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L333?

ahsonkhan:
cc @jkotas

shiftylogic:
I think that comment is already out-of-sync. The implementations there are not the same (due to different requirements; no fallback, etc).

Also, without the comment on the other side, there is no guarantee that changes on the CLR side will make it to this code.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -28,19 +28,24 @@ internal static class Utf8Encoder
         private const byte b1111_0000U = 0xF0; //240
         private const byte b1111_1000U = 0xF8; //248
 
+        // bits 30 & 31 are used for pending bits fixup
```
jkotas: This is used to deal with partially encoded characters.

You should not need this because of the corefx decoders do not ever stop at half character. Instead, they stop will not fully consume the input buffer.

tests/System.Text.Primitives.Tests/Encoding/Utf8Utf16ConversionTests.cs
```diff
@@ -77,5 +77,67 @@ public void EnumerateAndEnsureCodePointsOfTheSameUtf8AndUtf16StringsAreTheSame(s
...
         }
+
+        [Theory]
+        [InlineData(0, 0x7f)]
+        [InlineData(0x80, 0x7ff)]
+        [InlineData(0x800, 0x7fff)]
+        [InlineData(0, 0xffff)]
+        [InlineData(0x10000, 0x10ffff)]
+        [InlineData(0, 0x10ffff)]
+        public void RandomUtf8ToUtf16DecodingTests(int minCodePoint, int maxCodePoint)
```
ahsonkhan: Thoughts on brute forcing like these (using GenerateAllCharString):
https://github.com/dotnet/corefxlab/blob/master/tests/System.Text.Primitives.Tests/Encoding/Utf8EncoderTests.cs#L212
https://github.com/dotnet/corefxlab/blob/master/tests/System.Text.Primitives.Tests/Encoding/Utf8EncoderTests.cs#L576

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -51,39 +56,497 @@ internal static class Utf8Encoder
...
+                            *(pTarget+1) = (char)((ch >> 16) & 0x7F);
+                            *(pTarget+2) = (char)((ch >> 8) & 0x7F);
+                            *(pTarget+3) = (char)(ch & 0x7F);
+                            pSrc += 8;
+                            *(pTarget+4) = (char)((chb >> 24) & 0x7F);
+                            *(pTarget+5) = (char)((chb >> 16) & 0x7F);
+                            *(pTarget+6) = (char)((chb >> 8) & 0x7F);
+                            *(pTarget+7) = (char)(chb & 0x7F);
+                            pTarget += 8;
+    #else // BIGENDIAN
```
KrzysztofCwalina: The TexEncoder is UTF8 to UTF16LE, and so this section could be removed. We might want to create a separate subclass of TextEncoder to do UTF8 to UTF16BE

jkotas: This ifdef is for endianness of the underlying platform, not for endianness of the encoding...

KrzysztofCwalina:
Do we have perf tests comparing the perfromance of the existing encoders to this?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
KrzysztofCwalina: Ah, I just realized the output is Span<char> (i.e. underlying platform dependent).

shiftylogic:
I will be adding more tests for corner cases we do not appear to catch right now around incomplete sequences at the tail of spans. Also, I will add the performance tests to the benchmarks.

However, my results are below. Note that the currently checked in implementation is slower by factors of 21x, 6x, 4x for ranges (0-7f, 80-7ff, 800-ffff) respectively.

| Test | New | CLR | Ratio |
|---|---|---|---|
| UTF-8 => UTF-16 (1000, 0x0000, 0x007f) | 33 | 34 | 0.971 |
| UTF-8 => UTF-16 (1000, 0x0080, 0x07ff) | 147 | 152 | 0.967 |
| UTF-8 => UTF-16 (1000, 0x0800, 0xffff) | 262 | 270 | 0.970 |
| UTF-8 => UTF-16 (1000, 0x10000, 0x10ffff) | 370 | 371 | 0.997 |
| UTF-8 => UTF-16 (1000, 0x0000, 0x10ffff) | 417 | 387 | 1.078 |

ahsonkhan:
> I will be adding more tests for corner cases we do not appear to catch right now around incomplete sequences at the tail of spans. 

@shiftylogic, hopefully [this](https://github.com/dotnet/corefxlab/pull/1508) work will help you with that.

KrzysztofCwalina:
Looks good

PR #1511, opened 2017-05-01
Invalid sequence tests for UTF-8 to UTF-16; plus 2-byte long code bug fix.

shiftylogic:


shiftylogic:
@KrzysztofCwalina @ahsonkhan

tests/System.Text.Primitives.Tests/Encoding/Utf8Utf16ConversionTests.cs
```diff
@@ -78,6 +78,127 @@ public void EnumerateAndEnsureCodePointsOfTheSameUtf8AndUtf16StringsAreTheSame(s
             }
         }
 
+        public static object[][] InvalidUtf8ToUtf16SequenceData = {
```
ahsonkhan: Should we add this to set of Encode/Decode tests for all encodings? And are the test values specific to this one implementation?

tests/System.Text.Primitives.Tests/Encoding/Utf8Utf16ConversionTests.cs
```diff
@@ -78,6 +78,127 @@ public void EnumerateAndEnsureCodePointsOfTheSameUtf8AndUtf16StringsAreTheSame(s
...
+            Span<char> actualOutput = new Span<char>(new char[expectedOutput.Length + 10]);
+            Assert.False(TextEncoder.Utf8.TryDecode(input, actualOutput, out int consumed, out int written));
+            Assert.Equal(expectedConsumed, consumed);
+            Assert.Equal(expectedOutput.Length, written);
+
+            actualOutput = actualOutput.Slice(0, written);
+            //Assert.True(actualOutput.SequenceEqual(expectedOutput));
+
+            Assert.Equal(expectedOutput.Length, actualOutput.Length);
+            for (var i = 0; i < Math.Min(expectedOutput.Length, actualOutput.Length); i++)
```
ahsonkhan: Math.Min isn't necessary since they are both equal (from previous Assert.Equal check).

tests/System.Text.Primitives.Tests/Encoding/Utf8Utf16ConversionTests.cs
shiftylogic: These tests are specific to going from UTF-8 to UTF-16. You'd have to craft different tests for other conversions.

Most of these tests are generic to benefit any implementation. However, the "long" versions know that the buffer needs to be greater than 13 bytes to get into the "fast" path. So, there are some tests specific to ensure we exercise all the code.

tests/System.Text.Primitives.Tests/Encoding/Utf8Utf16ConversionTests.cs
shiftylogic: Yeah... that should be there. That was me debugging. I'll fix it.

tests/System.Text.Primitives.Tests/Encoding/Utf8Utf16ConversionTests.cs
ahsonkhan: Do you think there is a way for us to systematically catch all the corner cases without having implementation specific cases? Given most of these rely on values of characters consumed and bytes written, I would think, no.

shiftylogic: I don't think these specific tests can be generalized more because they know about the underlying implementation. If we were to find an implementation with similar performance characteristics that didn't have to have various special paths through the code, then we wouldn't need some of these specialized tests.

PR #1516, opened 2017-05-02
Renaming charactersWritten/Consumed to codePointsWritten/Consumed.

ahsonkhan:
Specifically for encoding/decoding `ReadOnlySpan<uint> utf32`

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
@@ -101,23 +101,23 @@ namespace System.Text.Utf16
         /// 
         /// This method will consume as many of the input characters as possible.
         ///
-        /// On successful exit, the entire input was consumed and encoded successfully. In this case, <paramref name="charactersConsumed"/> will be
+        /// On successful exit, the entire input was consumed and encoded successfully. In this case, <paramref name="codePointsConsumed"/> will be
         /// equal to the length of the <paramref name="utf32"/> and <paramref name="charactersWritten"/> will equal the total number of bytes written to
         /// the <paramref name="utf16"/>.
         /// 
         /// On unsuccessful exit, the following conditions can exist.
-        ///  1) If the output buffer has been filled and no more input characters can be encoded, another call to this method with the input sliced to
-        ///     exclude the already encoded characters (using <paramref name="charactersConsumed"/>) and a new output buffer will continue the encoding.
+        ///  1) If the output buffer has been filled and no more input code points can be encoded, another call to this method with the input sliced to
+        ///     exclude the already encoded code points (using <paramref name="codePointsConsumed"/>) and a new output buffer will continue the encoding.
         ///  2) Encoding may have also stopped because the input buffer contains an invalid sequence.
         /// </summary>
         /// <param name="utf32">A span containing a sequence of UTF-32 characters.</param>
         /// <param name="utf16">A span to write the UTF-16 encoded data into.</param>
-        /// <param name="charactersConsumed">On exit, contains the number of code points that were consumed from the UTF-32 character span.</param>
+        /// <param name="codePointsConsumed">On exit, contains the number of code points that were consumed from the UTF-32 character span.</param>
         /// <param name="charactersWritten">An output parameter to store the number of bytes written to <paramref name="utf8"/></param>
         /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
-        public static bool TryEncode(ReadOnlySpan<uint> utf32, Span<char> utf16, out int charactersConsumed, out int charactersWritten)
+        public static bool TryEncode(ReadOnlySpan<uint> utf32, Span<char> utf16, out int codePointsConsumed, out int charactersWritten)
         {
-            charactersConsumed = 0;
+            codePointsConsumed = 0;
             charactersWritten = 0;
 
             for (int i = 0; i < utf32.Length; i++)
@@ -126,7 +126,7 @@ namespace System.Text.Utf16
                 if (!TryEncode(utf32[i], utf16, charactersWritten, out written))
                     return false;
 
-                charactersConsumed++;
+                codePointsConsumed++;
                 charactersWritten += written;
             }
 
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf16/Utf16LittleEndianEncoder.cs
@@ -40,7 +40,7 @@ namespace System.Text.Utf16
         /// This method will consume as many of the input characters as possible.
         ///
         /// On successful exit, the entire input was consumed and encoded successfully. In this case, <paramref name="charactersConsumed"/> will be
-        /// equal to the length of the <paramref name="utf16"/> and <paramref name="bytesWritten"/> will equal the total number of bytes written to
+        /// equal to the length of the <paramref name="utf16"/> and <paramref name="codeUnitsWritten"/> will equal the total number of code units written to
         /// the <paramref name="utf32"/>.
         /// 
         /// On unsuccessful exit, the following conditions can exist.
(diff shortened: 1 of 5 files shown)


KrzysztofCwalina:
I am not sure this is the right terminology. I am under the impression that "code point" is a Unicode concept independent of encoding, i.e. there are no code points in encodings, like UTF32. The units of encoding in encodings (like the UTF32 encoding) are called "code units".

tarekgh:
@KrzysztofCwalina you definition is not correct. please have a look at the definition in the following link:

https://en.wikipedia.org/wiki/Code_point

```
Definition
The notion of a code point is used for abstraction, to distinguish both:

the number from an encoding as a sequence of bits, and
the abstract character from a particular graphical representation (glyph).
This is because one may wish to make these distinctions:

encode a particular code space in different ways, or
display a character via different glyphs.
```

ahsonkhan:
Hmm, it also states the following:
> For Unicode, the particular sequence of bits is called a **code unit** – for the UCS-4 encoding, any code point is encoded as 4-byte (octet) binary numbers

And: https://en.wikipedia.org/wiki/Character_encoding#Terminology
> A code point of a coded character set is any legal value in the character set.
A code unit is a bit sequence used to encode each character of a repertoire within a given encoding form.

So, @KrzysztofCwalina, is the suggestion to rename as codeUnitsConsumed instead of codePoints?

tarekgh:
Code Unit is not popular term and I am not seeing anyone is using it. And I don't think it will be a good idea to use this term.

KrzysztofCwalina:
Is this wikipedia entry what we should base our naming on? Especially that I find the particular section you quoted badly written/worded. 

When I go to the Unicode spec, this is what they say: http://www.unicode.org/glossary/#code_unit

And in general, I worry that calling the encoded output "code points" does not scale for other encodings. For example, it does not make any sense for UTF16. Not surprisingly, the UTF16 wikipedia page clearly makse the distinction in the opening pharagraph: https://en.wikipedia.org/wiki/UTF-16

ahsonkhan:
> Is this wikipedia entry what we should base our naming on?

The quotes I provided from wikipedia lean towards using code units for UTF-32, not code points anyway.

> And in general, I worry that calling the encoded output "code points" does not scale for other encodings. For example, it does not make any sense for UTF16.

Why does this rename affect the UTF-16 case? The rename to code points/code units is only for UTF-32. That said, I am leaning towards code unit as well (keeping characters consumed for UTF-16).

```C#
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, Span<byte> utf8, out int bytesConsumed, out int bytesWritten);
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, Span<char> utf16, out int bytesConsumed, out int charactersWritten);
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, Span<uint> utf32, out int bytesConsumed, out int codeUnitsConsumed);
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, out string text, out int bytesConsumed);

public abstract bool TryEncode(ReadOnlySpan<byte> utf8, Span<byte> encodedBytes, out int bytesConsumed, out int bytesWritten);
public abstract bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> encodedBytes, out int charactersConsumed, out int bytesWritten);
public abstract bool TryEncode(ReadOnlySpan<uint> utf32, Span<byte> encodedBytes, out int codeUnitsConsumed, out int bytesWritten);
public abstract bool TryEncode(string text, Span<byte> encodedBytes, out int bytesWritten);
```

KrzysztofCwalina:
```c#
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, Span<uint> utf32, out int bytesConsumed, out int codeUnitsConsumed);
```
the second parameter should be codeUnitsWritten

tarekgh:
> the second parameter should be codeUnitsWritten

I am not familiar with the proposed API but I am wondering are you always emitting in utf32? or it can be in other encoding? code units is specific to the encoding (like one byte in UTF8, 2 bytes for UTF16...etc.) so the name will be tied to the encoding. code point name is abstracted from the encoding and represent just one real char regardless of the encoding. again I am not familiar with the API so I am not really suggesting any name but I wanted to ensure we are not confusing with the terms here

ahsonkhan:
> I am not familiar with the proposed API but I am wondering are you always emitting in utf32?

You emit whatever you are decoding to. It could be utf8 bytes, utf16 chars, utf32 uints, or string.

Below are the TextEncoder APIs for encoding/decoding:

You would call them like:
`TextEncoder.UTF8.TryDecode(encodedBytes, utf32, out int bytesConsumed, out int codeUnitsWritten);`
OR
`TextEncoder.UTF16.TryDecode(encodedBytes, utf32, out int bytesConsumed, out int codeUnitsWritten);`
...

```C#
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, Span<byte> utf8, out int bytesConsumed, out int bytesWritten);
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, Span<char> utf16, out int bytesConsumed, out int charactersWritten);
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, Span<uint> utf32, out int bytesConsumed, out int codeUnitsWritten);
public abstract bool TryDecode(ReadOnlySpan<byte> encodedBytes, out string text, out int bytesConsumed);

public abstract bool TryEncode(ReadOnlySpan<byte> utf8, Span<byte> encodedBytes, out int bytesConsumed, out int bytesWritten);
public abstract bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> encodedBytes, out int charactersConsumed, out int bytesWritten);
public abstract bool TryEncode(ReadOnlySpan<uint> utf32, Span<byte> encodedBytes, out int codeUnitsConsumed, out int bytesWritten);
public abstract bool TryEncode(string text, Span<byte> encodedBytes, out int bytesWritten);
```

tarekgh:
if for example you are emitting in UTF16 and you have surrogate pair. in your API this will be counted as one or 2 units?

ahsonkhan:
> if for example you are emitting in UTF16 and you have surrogate pair. in your API this will be counted as one or 2 units?

2 units (which in this case would be characters). We count whatever the unit we are emitting is.

tarekgh:
and if we have UTF8, we'll have 4 units. right? 
if this is the case then Code Units name is the right term to use here as @KrzysztofCwalina suggested.

ahsonkhan:
> if this is the case then Code Units name is the right term to use here as @KrzysztofCwalina suggested.

For all?

So, no bytesWritten for utf8 and charactersWritten for utf16? codeUnitsWritten for all utf8/16/32?

tarekgh:
> So, no bytesWritten for utf8 and charactersWritten for utf16? codeUnitsWritten for all utf8/16/32?

I think this will be the right thing to apply it on all encodings

ahsonkhan:
> I think this will be the right thing to apply it on all encodings

I would imagine that if you are emitting a Span of bytes, it might be clearer for the user to see bytesWritten (not codeUnitsWritten) and similarly for Span of chars. That is why I was specifically only proposing to change Span of uints to codeUnitsWritten. That way the parameters make sense together.

tarekgh:
what is the difference between the Span of units and Span of bytes for instance? why we don't have just Span of Units which cover all cases (bytes, chars...etc)

ahsonkhan:
> what is the difference between the Span of units and Span of bytes for instance? why we don't have just Span of Units which cover all cases (bytes, chars...etc)

Not Span of units, uints (unsigned ints). `Span<uint>` (primitive type)

tarekgh:
> Not Span of units, uints (unsigned ints). Span<uint> (primitive type)

Thanks for the correction. 
@KrzysztofCwalina will be better for naming here but I am wondering if we used bytesWritten and characterWritten, why we need to use codeUnitWritten here and not used someing like (intWritten/uintWritten)? 
my idea is either to use codeUnitWritten all the way or you use the actual primitive type name all the way.

ahsonkhan:
To summarize, here are the options:
@KrzysztofCwalina, @shiftylogic - what do you think?

**Option 1:** @ahsonkhan
`TryDecode(... Span<byte> utf8 ... )` - out bytesWritten
`TryDecode(... Span<char> utf16 ... )` - out charactersWritten
`TryDecode(... Span<uint> utf32 ... )` - out codeUnitsWritten

**Option 2:**  @tarekgh, @ahsonkhan
`TryDecode(... Span<byte> utf8 ... )` - out bytesWritten
`TryDecode(... Span<char> utf16 ... )` - out charactersWritten
`TryDecode(... Span<uint> utf32 ... )` - out uintsWritten

**Option 3:**  @tarekgh
`TryDecode(... Span<byte> utf8 ... )` - out codeUnitsWritten
`TryDecode(... Span<char> utf16 ... )` - out codeUnitsWritten
`TryDecode(... Span<uint> utf32 ... )` - out codeUnitsWritten

KrzysztofCwalina:
I think the overloads that write bytes and chars will be most common. The current names of the parameter in these overloads are very clear. I would prefer not to rename them. Many users won't know what code unit is. The uint overload will be almost never used. It's mainly to support two-hop (exotic) encodings, e.g. 8859-7 -> UTF32 then UTF32 -> UTF8

I think option #2 would be interesting, but bytes and characters are English words; uints is not. Maybe it's a small enough problem that we can live with it. 

cc: @terrajobst

ahsonkhan:
> I think the overloads that write bytes and chars will be most common. The current names of the parameter in these overloads are very clear. I would prefer not to rename them.

+1

> I think option 2 would be interesting, but bytes and characters are English words; uints is not. 

It seems like you are leaning towards 1. I do prefer option 1 over 2 (i.e. just use codeUnitsWritten for the UTF-32 case).

> We could also rename all the parameters to simply "written"?

I think written doesn't give enough info.

KrzysztofCwalina:
I am ok with either option 1 or 2. Your call.

ahsonkhan:
@shiftylogic, good to go?

PR #1519, opened 2017-05-03
Fix formatting of titles.

JimBobSquarePants:
Simple fix for titles that were not rendering correctly due to malformed markup syntax.


joshfree:
thanks @JimBobSquarePants

Issue #1514, opened 2017-05-02
UTF8 Encoder doesn't do validation if input is UTF8 bytes.


ahsonkhan:
From: https://github.com/dotnet/corefxlab/pull/1508

https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8TextEncoder.cs#L104

This returns true even if inputUtf8 contains invalid UTF8 bytes.
```C#
utf8.TryEncode(inputUtf8, encodedBytes, out int charactersConsumed, out bytesWritten)
```
Another example is if "InputBufferLargerThanOutputBuffer". The CopyTo throws an exception:
System.ArgumentException : Destination is too short

This isn't the expected behaviour for other encodings. We should be allowed to restart with a larger buffer to complete:
```C#
Assert.False(utf8.TryEncode(inputUtf8, encodedBytes1, out int charactersConsumed1, out bytesWritten1));
encodedBytes2 = new Span<byte>(new byte[expectedBytesWritten - bytesWritten1]);
Assert.True(utf8.TryEncode(inputUtf8.Slice(charactersConsumed1), encodedBytes2, out int charactersConsumed2, out bytesWritten2));
```

cc @KrzysztofCwalina, @shiftylogic

ahsonkhan:
Assigning to @GrabYourPitchforks, as this overlaps with the UTF-8 work he has been doing.

GrabYourPitchforks:
@ahsonkhan, I don't see the type `Utf8TextEncoder` across any of our repos, and the earlier link seems dead. Can you point me to the method we should investigate? Thanks!

ahsonkhan:
TextEncoders were re-designed here: https://github.com/dotnet/corefxlab/pull/1622

Here is the method in question, now:
https://github.com/dotnet/corefxlab/blob/4f23c82316b5e14ef8379ab6e4653bb43ab92dd8/src/System.Text.Primitives/System/Text/SymbolTable/SymbolTable_utf8.cs#L51

PR #1508, opened 2017-04-28
Adding TextEncoder encode/decode tests for UTF-8 & 16.

ahsonkhan:
Issues:
https://github.com/dotnet/corefxlab/issues/1512
https://github.com/dotnet/corefxlab/issues/1513
https://github.com/dotnet/corefxlab/issues/1514
https://github.com/dotnet/corefxlab/issues/1515
https://github.com/dotnet/corefxlab/issues/1524
https://github.com/dotnet/corefxlab/issues/1525

cc @shiftylogic, @KrzysztofCwalina

PR #1520, opened 2017-05-03
Adding performance tests for all encoding combinations.

ahsonkhan:
These test all encoding combinations and compares the CoreCLR implementations with the Corefxlab implementations:
- From UTF-8/UTF-16/UTF-32 to UTF-8
- From UTF-8/UTF-16/UTF-32 to UTF-16

Seven different test strings were measured to cover some interesting scenarios. Additional test strings that might be interesting can easily be added.

cc @shiftylogic, @KrzysztofCwalina, @jkotas

**TO DO:**
- Decoding Tests #1523
- ~Run tests for different string lengths (and not just 999).~
- ~Add inner iteration count or encode larger strings to reduce instability in results.~

**Note:** This relies on some helper functions/constants that will become available after https://github.com/dotnet/corefxlab/pull/1508 is merged. For now, they are duplicated here.

**INSIGHTS ABOUT/FROM THE DATA:**
- AllASCII/MostlyASCII => Surprisingly, the CoreCLR implementation is on par (if not faster) for strings that mostly contain ASCII compared to strings that are ASCII only.
- ALLSurrogates/AlternatingASCII => However, alternating between ASCII and non-ASCII characters severely degrades performance for encoding to UTF-16 (and somewhat for encoding to UTF-8 as well).
- Some corefxlab encoders don't validate and just do a copy (example UTF-8 to UTF-8) #1514
- Some encoding combinations don't exist for CoreCLR (example UTF-8 to UTF-32), so we use Convert which allocates/is slow

**RESULTS:**

image

image

image

image


KrzysztofCwalina:
Does it make sense to test UTF8 to UTF8 conversion? Same with UTF16 to UTF16. Seems to me like these tests just add noise.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
```diff
@@ -0,0 +1,180 @@
...
+        [InlineData(DataLength, 0x0, Utf8ThreeBytesLastCodePoint)]
+        [InlineData(DataLength, 0, 0, SpecialTestCases.AlternatingASCIIAndNonASCII)]
+        [InlineData(DataLength, 0, 0, SpecialTestCases.MostlyASCIIAndSomeNonASCII)]
+        public void EncodeFromUtf16toUtf8CoreCLR(int length, int minCodePoint, int maxCodePoint, SpecialTestCases special = SpecialTestCases.None)
+        {
+            string inputString = GenerateStringData(length, minCodePoint, maxCodePoint, special);
+            char[] characters = inputString.AsSpan().ToArray();
+            Encoding utf8 = Encoding.UTF8;
+            int utf8Length = utf8.GetByteCount(characters);
+            var utf8Buffer = new byte[utf8Length];
```
KrzysztofCwalina: Why not call Encoding.UTF8.GetBytes(inputString) to get an array of UTF8 bytes?

tests/System.Text.Primitives.Tests/EncodingPerfTestsCorefxlab.cs
```diff
@@ -0,0 +1,341 @@
...
+
+        [Benchmark]
+        [InlineData(DataLength, 0x0, Utf8OneByteLastCodePoint)]
+        [InlineData(DataLength, Utf8OneByteLastCodePoint + 1, Utf8TwoBytesLastCodePoint)]
+        [InlineData(DataLength, Utf8TwoBytesLastCodePoint + 1, Utf8ThreeBytesLastCodePoint)]
+        [InlineData(DataLength, Utf16HighSurrogateFirstCodePoint, Utf16LowSurrogateLastCodePoint)]
+        [InlineData(DataLength, 0x0, Utf8ThreeBytesLastCodePoint)]
+        [InlineData(DataLength, 0, 0, SpecialTestCases.AlternatingASCIIAndNonASCII)]
+        [InlineData(DataLength, 0, 0, SpecialTestCases.MostlyASCIIAndSomeNonASCII)]
+        public void EncodeFromUtf8toUtf8Corefxlab(int length, int minCodePoint, int maxCodePoint, SpecialTestCases special = SpecialTestCases.None)
```
KrzysztofCwalina: Would it be better to use type names as opposed to "corfxlab"? e.g. EncodeFromUtf8ToUtf16TextEncoder and EncodeFromUtf8ToUtf16Encoding

tests/System.Text.Primitives.Tests/EncodingPerfTestsCorefxlab.cs
```diff
@@ -0,0 +1,341 @@
...
+            }
+        }
+
+        // TODO
+        public void DecodeFromUtf8toUtf8Corefxlab() { }
+        public void DecodeFromUtf8toUtf16Corefxlab() { }
+        public void DecodeFromUtf8toUtf32Corefxlab() { }
+        public void DecodeFromUtf16toUtf8Corefxlab() { }
+        public void DecodeFromUtf16toUtf16Corefxlab() { }
+        public void DecodeFromUtf16toUtf32Corefxlab() { }
```
KrzysztofCwalina: I would rather add an issue to keep track of tests that still need to be added.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCorefxlab.cs
```diff
@@ -0,0 +1,341 @@
...
+
+        public static string GenerateStringData(int length, int minCodePoint, int maxCodePoint, SpecialTestCases special = SpecialTestCases.None)
```
KrzysztofCwalina: Do these need to be public?

ahsonkhan: No. Will fix.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCorefxlab.cs
ahsonkhan: Sure.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
```diff
@@ -0,0 +1,180 @@
...
+            var utf32Buffer = new byte[utf32Length];
+            utf32.GetBytes(characters, 0, characters.Length, utf32Buffer, 0);
+
+            int utf8Length = utf8.GetByteCount(characters);
+            var output = new byte[utf8Length];
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                    output = Encoding.Convert(utf32, utf8, utf32Buffer);
```
ahsonkhan: So, are you suggesting go from UTF-32 to UTF-16 and then from UTF-16 to UTF-8 instead?
Should I do the same thing for the Corefxlab tests even though we have a (potentially) faster UTF-32 to UTF-8 direct encoder? I was trying to keep the comparison consistent. I think chaining the decoders will make the CoreCLR test slower.

ahsonkhan:
> Does it make sense to test UTF8 to UTF8 conversion? Same with UTF16 to UTF16. Seems to me like these tests just add noise.

It isn't really noise since the coreclr encoders still do meaningful work and validate/fall back for invalid characters. We do not do the validation and we do provide such an API. So why not test it?

http://referencesource.microsoft.com/#mscorlib/system/text/encoding.cs,388

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
ahsonkhan: The test is measuring the performance of that exact function call and I do that within the loop being measured. Using  Encoding.UTF8.GetBytes(inputString) causes an allocation of the byte[] being returned.

```C#
            int utf8Length = utf8.GetByteCount(characters);
            var utf8Buffer = new byte[utf8Length];

            {
                    utf8.GetBytes(characters, 0, characters.Length, utf8Buffer, 0);
            }
```

tests/System.Text.Primitives.Tests/EncodingPerfTestsCorefxlab.cs
ahsonkhan: #1523

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
shiftylogic: This implementation is not really consistent with what we do in the primitives. In order for the performance tests to be meaningful, I would pick the fastest way via the CoreCLR implementation to compare to our direct implementation. The most efficient way in CoreCLR would be to pre-allocate the buffers (not part of the performance test since we do the same with Span), then call the necessary decoders / encoders to get where you want. If there is no direct path, use as few links in the chain as you can. Most cases,  you can go through UTF-16.

ahsonkhan: > I would pick the fastest way via the CoreCLR implementation to compare to our direct implementation.

I am not certain if chaining is actually faster but I will update and test.

ahsonkhan:
**Question:** Is it worth it to add encoding tests of varying lengths or is it unnecessarily increasing test duration & noise?

I choose 99, 999, 9999, 99999. The highest priority encoders behave very similarly to each other.
The larger the input length, the better the encoding gets (per character). For short input length (< 100), it is significantly more expensive.

image

image

image

image

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
KrzysztofCwalina: Ah, sorry. I thought you are creating the input buffer. Still you may try to simplify the code to:
```c#
sring inputString = GenerateStringData(length, minCodePoint, maxCodePoint, special);
char[] characters = inputString.ToCharArray();
int utf8Length = Encoding.UTF8.GetByteCount(characters);
var utf8Buffer = new byte[utf8Length];
```
up to you

KrzysztofCwalina:
Maybe it would be sufficient to skip the longest one, i.e. 99999

KrzysztofCwalina:
looks good otherwise

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
```diff
@@ -0,0 +1,146 @@
...
+            Encoding utf32 = Encoding.UTF32;
+            Encoding utf8 = Encoding.UTF8;
+            int utf32Length = utf32.GetByteCount(characters);
+            var utf32Buffer = new byte[utf32Length];
+            utf32.GetBytes(characters, 0, characters.Length, utf32Buffer, 0);
+
+            int utf8Length = utf8.GetByteCount(characters);
+
+            var temp = utf32.GetChars(utf32Buffer);
+            byte[] output = Encoding.Convert(utf32, utf8, utf32Buffer);
```
shiftylogic: Are you doing this as a way to pre-allocate memory for the calls below? If that is true, you have the necessary value to allocate directly via the utf8Length above, which is otherwise never used.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
```diff
@@ -0,0 +1,146 @@
...
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                        utf8.GetBytes(utf32.GetChars(utf32Buffer), 0, temp.Length, output, 0);
```
shiftylogic: The GetChars call is still allocating here. You need to call the versions for both GetBytes and GetChars that use your input and output pre-allocated buffers.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
```diff
@@ -0,0 +1,146 @@
...
+            utf16.GetBytes(characters, 0, characters.Length, utf16Buffer, 0);
+
+            var temp = utf16.GetChars(utf16Buffer);
+            byte[] output = Encoding.Convert(utf16, utf16, utf16Buffer);
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                        utf16.GetBytes(utf16.GetChars(utf16Buffer), 0, temp.Length, output, 0);
```
shiftylogic: Same here. GetChars is allocating.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
```diff
@@ -0,0 +1,146 @@
...
+            utf32.GetBytes(characters, 0, characters.Length, utf32Buffer, 0);
+
+            var temp = utf32.GetChars(utf32Buffer);
+            byte[] output = Encoding.Convert(utf32, utf16, utf32Buffer);
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                        utf16.GetBytes(utf32.GetChars(utf32Buffer), 0, temp.Length, output, 0);
```
shiftylogic: And here again.

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
ahsonkhan: Ah yes, I missed that. Will fix :)

tests/System.Text.Primitives.Tests/EncodingPerfTestsCoreCLR.cs
ahsonkhan: image

ahsonkhan:
image

ahsonkhan:
image

ahsonkhan:
Good to go?

KrzysztofCwalina:
yeah, looks good

jkotas:
> Why is all ASCII encoding significantly slower than mostly ASCII?

This is not expected, is it? I do not see anything obvious. Run it under perfview, look at the disassembly and figure it out.

> Alternating ASCII and surrogate characters severely degrades encoder performance (especially for UTF-32 to UTF-16). Is this a rare scenario?

Depends on which part of the world you are in. I expect that mix of ASCII and surrogates to be relatively common in China. Also, I would expect a long runs of surrogates to be very rare; surrogates should be typically mixed with non-surrogates. 

The encoders/decoders implementations tend to have switching cost between the more common tighter loop (ASCII, non-surrogates, etc.) and the less common loop (non-ASCII, surrogates, etc.). In good implementations, this switching cost is balanced with the performance for other inputs.

tests/System.Text.Primitives.Tests/EncodingPerfTestsUsingTextEncoder.cs
```diff
@@ -0,0 +1,316 @@
...
+            TextEncoder utf16 = TextEncoder.Utf16;
+            Assert.True(utf16.TryComputeEncodedBytes(utf8Span, out int encodedBytes));
+            var utf16Buffer = new byte[encodedBytes];
+            var span = new Span<byte>(utf16Buffer);
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                        if (!utf16.TryEncode(utf8Span, span, out int consumed, out int written)) { throw new Exception(); }
```
shiftylogic: This is the equivalent of calling utf8.TryDecode, but goes through an extra indirection and a bit more code. That explains a bit of the discrepancy against CoreCLR implementation.

tests/System.Text.Primitives.Tests/EncodingPerfTestsUsingEncoding.cs
```diff
@@ -0,0 +1,158 @@
...
+
+            Encoding utf16 = Encoding.Unicode;
+            int utf16Length = utf16.GetCharCount(utf8Buffer);
+            var utf16Buffer = new char[utf16Length];
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                        utf16.GetChars(utf8Buffer, 0, utf8Buffer.Length, utf16Buffer, 0);
```
shiftylogic: This is the wrong comparison. You are passing a buffer of UTF8 bytes and UnicodeEncoding will assume that byte buffer is filled with UTF-16 chars and process them. So, this is effectively equivalent to utf-16 to utf-16 conversion with validation.

You need to call Utf8.GetBytes instead. I suspect if you do that, you will see that the performance is equivalent to the Span-based TextEncoder.

tests/System.Text.Primitives.Tests/EncodingPerfTestsUsingTextEncoder.cs
shiftylogic: Note, the biggest cause of the difference is a result of the below comment since we aren't comparing the same thing.

ahsonkhan:
> Also, I would expect a long runs of surrogates to be very rare; surrogates should be typically mixed with non-surrogates.

Don't most (if not all) emoji translate to surrogate characters in utf-16 since the code point for emoji is larger than 0xFFFF?

jkotas:
> > Also, I would expect a long runs of surrogates to be very rare; surrogates should be typically mixed with non-surrogates.
> Don't most (if not all) emoji translate to surrogate characters in utf-16 since the code point for emoji is larger than 0xFFFF?

This does not contradict what I have said. Where do you see a problem?

ahsonkhan:
> This does not contradict what I have said. Where do you see a problem?

I have seen communication from some like the following. I am not sure about the exact data (so I could be wrong), but I am guessing this isn't very rare.

image

jkotas:
There are ASCII characters before and after emojis in the web page payload so this is still the surrogates mixed with non-surrogates case.

ahsonkhan:
> There are ASCII characters before and after emojis in the web page payload so this is still the surrogates mixed with non-surrogates case.

Even if the web page itself is non-English?

shiftylogic:
Can you post perf numbers for the latest code?

ahsonkhan:
> Can you post perf numbers for the latest code?

Yes, doing that now.

ahsonkhan:
image

ahsonkhan:
image

ahsonkhan:
image

ahsonkhan:
image

jkotas:
> Even if the web page itself is non-English?

Yes. HTML tags are always ASCII.

ahsonkhan:
> Yes. HTML tags are always ASCII.

OK, but wouldn't that mean that in those scenarios we have long runs of surrogate pairs with some ASCII characters in between rather than long runs of ASCII with some surrogate characters in between?

jkotas:
Depends on how you define a "long run" and how you define "likely". Go to a several random Chinese webpages and run "view source" on them. It will give you a pretty good feel what a typical Chinese HTML payload looks like.

ahsonkhan:
> Go to a several random Chinese webpages and run "view source" on them. It will give you a pretty good feel what a typical Chinese HTML payload looks like.

Good point!

```html
<meta http-equiv="Content-Type" content="text/html; charset=utf-8">
<link href="http://www.gov.cn/govweb/xhtml/favicon.ico" rel="shortcut icon" type="image/x-icon">
<title>中国政府网_中央人民政府门户网站</title>
<meta name="others" content="页面生成时间 2017-04-01 17:39:07" />
<meta http-equiv="X-UA-Compatible" content="IE=8" />
<meta name="keywords" content="中央政府门户网站,中国政府网，国务院,总理,李克强，政府，GOV，我向总理说句话,双创,副总理，国务委员，国办,国务院办公厅,政策，政府信息公开,国务院公报,国务院文件,国务院常务会,政府权责清单,网上办事，国务院督查,政务联播,部门大厅，放假安排">
<meta name="description" content="中国政府网由国务院办公厅主办，中国政府网运行中心负责运行维护,是国务院和国务院各部门，以及各省、自治区、直辖市人民政府在国际互联网上发布政府信息和提供在线服务的综合平台,第一时间权威发布国务院重大决策部署和重要政策文件，国务院领导同志重要会议、考察、出访活动等政务信息，面向社会提供与政府业务相关的服务，建设基于互联网的政府与公众互动交流新渠道。">
  <meta name="lanmu" content="中国政府网2016版首页">
<meta name="filepath" content="/index.htm">
<META name="catalog" content="2016govs">
<META name="filetype" content="0">
<META name="publishedtype" content="1">
<META name="pagetype" content="2">

```

```html
      
   </div>
   <!--<div style="width:1100px;height:51px;margin:0px auto;clear:both;"><iframe id="ifr_zhaopin.htm" src="http://www.gov.cn/zhuanti/2016zhaopinwenzi/zhaopin.html" frameborder="0" scrolling="no" height="51" width="1100" height="51"></iframe></div>-->
<!--第二部分banner结束-->
<div class="clear"></div>
<!--第三部分开始-->
   <div class="column3">
       <!--左侧-->
       <div class="column3lp">
       <!--要闻-->
          <div class="column3lp_1 fl">
            <div class="lmtit"><a href="/xinwen/yaowen.htm" target="_blank">要闻</a></div>
            <ul class="h206">
```

ahsonkhan:
@shiftylogic - For reference:
http://blogs.adobe.com/CCJKType/files/2015/02/beyond-bmp-top10-2015.pdf
https://en.wikipedia.org/wiki/CJK_Unified_Ideographs_Extension_B

ahsonkhan:
> This is not expected, is it? I do not see anything obvious. Run it under perfview, look at the disassembly and figure it out.

I stepped through the Encoding code and found a bug in my input string. I wasn't actually adding surrogate characters. The results make more sense now.

**Encoding only ASCII is faster than encoding mostly ASCII by quite a bit (as expected) for:**
- UTF-8 to UTF-16
- UTF-16 to UTF-8

**However, it isn't faster for:**
- UTF-32 to UTF-8
- UTF-32 to UTF-16

**The TextEncoder implementation is also not faster for (which is expected given the implementation):**
- UTF-8 to UTF-8
- UTF-16 to UTF-16

image

KrzysztofCwalina:
@ahsonkhan, are we merging it in?

PR #1521, opened 2017-05-03
Removed DisposableReservation

KrzysztofCwalina:
BufferHandle and DisposableReservation become almost identical. 
This change removes DisposableReservation and uses BufferHandle for both pinning and retaining
If a BufferHandle is created with Buffer<T>.Pin, BufferHandle.PinnedPointer will be not null. Otherwise it will be null and the calling code is responsible for not dereferencing it (and using the handle solely to Retain/Release). I could add a null check to throw when null PinnedPointer is accessed, but I am not sure it's worth it.


KrzysztofCwalina:
@mjp41, @ahsonkhan, @shiftylogic, @davidfowl

mjp41:
LGTM

PR #1527, opened 2017-05-04
Simplified OwnedBuffer Pinning Support

KrzysztofCwalina:
- Removed TryGetPointer. Pin should now be used to get to buffer pointers. 
- Pin got faster. The method used to be implemented in the base class and just caledl through TryGetPointer. Now OwnedBufer subtypes have specialized logic for pinning that is often simpler/faster than the base class implementation. Compare MemoryPoolBlock.Pin after this PR to OwnedBuffer.Pin before.
- Getting Span from OwnedBuffer is now done through AsSpan method. This allows to avoid slicing. See Buffer<T>.get_Span.

The perf (elapsed time and instructions retired) of E2EPipelineTests benchmark improved very slightly 0.5%, but consistently.

cc: @davidfowl, @mjp41, @ahsonkhan, @shiftylogic


KrzysztofCwalina:
Before
```
 Test Name                                                                                       | Metric                                        | Iterations |    AVERAGE |    STDEV.S |        MIN |        MAX
:----------------------------------------------------------------------------------------------- |:--------------------------------------------- |:----------:| ----------:| ----------:| ----------:| ----------:
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Duration                                      |     11     |    986.483 |     24.952 |    958.166 |   1039.723
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Instructions Retired                          |     11     | 7.220E+009 | 4.629E+007 | 7.131E+009 | 7.274E+009
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Branch Mispredictions                         |     11     | 9.290E+005 |  70012.088 | 7.987E+005 | 1.049E+006
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Cache Misses                                  |     11     | 6.639E+006 | 1.358E+006 | 4.387E+006 | 9.671E+006
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Allocation Size on Benchmark Execution Thread |     11     | 5.043E+007 |   3695.813 | 5.043E+007 | 5.044E+007
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Duration                                      |     11     |    972.915 |      8.116 |    962.184 |    986.075
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Instructions Retired                          |     11     | 7.223E+009 | 4.696E+007 | 7.121E+009 | 7.276E+009
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Branch Mispredictions                         |     11     | 9.447E+005 |  44315.340 | 8.929E+005 | 1.028E+006
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Cache Misses                                  |     11     | 6.436E+006 | 1.033E+006 | 4.317E+006 | 7.487E+006
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Allocation Size on Benchmark Execution Thread |     11     | 5.044E+007 |   4608.166 | 5.043E+007 | 5.044E+007
```

After
```
 Test Name                                                                                       | Metric                                        | Iterations |    AVERAGE |    STDEV.S |        MIN |        MAX
:----------------------------------------------------------------------------------------------- |:--------------------------------------------- |:----------:| ----------:| ----------:| ----------:| ----------:
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Duration                                      |     11     |    982.122 |     20.869 |    954.812 |   1029.571
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Instructions Retired                          |     11     | 7.214E+009 | 4.710E+007 | 7.126E+009 | 7.268E+009
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Branch Mispredictions                         |     11     | 9.000E+005 |  83564.241 | 7.823E+005 | 1.085E+006
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Cache Misses                                  |     11     | 6.725E+006 | 1.397E+006 | 4.202E+006 | 9.572E+006
 E2EPipelineTests.TechEmpowerHelloWorldNoIO(numberOfRequests: 10000, concurrentConnections: 256) | Allocation Size on Benchmark Execution Thread |     11     | 5.043E+007 |   2853.642 | 5.043E+007 | 5.044E+007
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Duration                                      |     11     |    970.666 |      7.329 |    953.240 |    982.167
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Instructions Retired                          |     11     | 7.217E+009 | 4.575E+007 | 7.115E+009 | 7.269E+009
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Branch Mispredictions                         |     11     | 9.168E+005 |  36673.187 | 8.520E+005 | 9.748E+005
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Cache Misses                                  |     11     | 6.567E+006 | 1.066E+006 | 4.063E+006 | 7.676E+006
 E2EPipelineTests.TechEmpowerJsonNoIO(numberOfRequests: 10000, concurrentConnections: 256)       | Allocation Size on Benchmark Execution Thread |     11     | 5.043E+007 |   3951.940 | 5.043E+007 | 5.044E+007
```

KrzysztofCwalina:
If Retain is called more than Release, Buffer.Span succeed. This requires reference counting for unmanaged buffers and pooled arrays. It does not require reference counting for array based buffers.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
```diff
@@ -69,7 +40,11 @@ public void Dispose()
         public abstract void Retain();
 
         public abstract void Release();
-
         #endregion
+
+        protected static unsafe void* Add(void* pointer, int offset)
```
ahsonkhan: Does this Add method belong here?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -57,18 +57,11 @@ public BufferHandle Retain()
...
-   
-        public unsafe bool TryGetPointer(out void* pointer)
-        {
-            if (!_owner.TryGetPointerAt(_index, out pointer)) {
-                return false;
-            }
-            return true;
-        }
 
         public unsafe bool TryGetArray(out ArraySegment<T> buffer)
```
ahsonkhan: Does this method still need to be unsafe?

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
```diff
@@ -13,45 +12,17 @@ public abstract class OwnedBuffer<T> : IDisposable, IRetainable
 
         public abstract int Length { get; }
 
-        public abstract Span<T> Span { get; }
+        public abstract Span<T> AsSpan(int index, int length);
```
ahsonkhan: Do we need this overload? Why not just use AsSpan().Slice(index, length)?

src/System.Buffers.Primitives/System/Buffers/OwnedArray.cs
```diff
@@ -0,0 +1,83 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Runtime;
+using System.Runtime.InteropServices;
+using System.Threading;
+
+namespace System.Buffers
+{
+    public class OwnedArray<T> : OwnedBuffer<T>
```
ahsonkhan: Is the only change here moving OwnedArrray from System/Buffers/Internals to System/Buffers or are there other changes as well?

src/System.Buffers.Primitives/System/Internal/OwnedArray.cs
```diff
@@ -1,63 +0,0 @@
...
-        {
-            _array = new T[length];
-        }
-
-        public OwnedArray(T[] array)
-        {
-            _array = array;
-        }
-
-        public OwnedArray(ArraySegment<T> segment)
```
ahsonkhan: What is the motivation to remove the OwnedArray constructor that takes an ArraySegment?

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
```diff
@@ -38,13 +39,10 @@ static void Free(byte[] array)
...
-        public override Span<byte> Span
+        public override Span<byte> AsSpan(int index, int length)
         {
-            get
-            {
-                if (IsDisposed) BufferPrimitivesThrowHelper.ThrowObjectDisposedException(nameof(OwnedBuffer));
-                return _array;
-            }
+            if (IsDisposed) BufferPrimitivesThrowHelper.ThrowObjectDisposedException(nameof(OwnedBuffer));
+            return _array.Slice(index, length);
```
ahsonkhan: `_array.AsSpan().Slice(index, length);` instead?

src/System.Buffers.Primitives/System/Buffers/BufferHandle.cs
```diff
@@ -21,7 +21,12 @@ public BufferHandle(IRetainable owner, void* pinnedPointer, GCHandle handle = de
 
         public BufferHandle(IRetainable owner) : this(owner, null) { }
 
-        public void* PinnedPointer => _pointer;
+        public void* PinnedPointer {
+            get {
+                if (_pointer == null) throw new InvalidOperationException();
```
ahsonkhan: Use ThrowHelper?
https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System/Runtime/BufferPrimitivesThrowHelper.cs#L37

src/System.Buffers.Primitives/System/Buffers/ReferenceCountedBuffer.cs
```diff
@@ -13,11 +14,13 @@ public abstract class ReferenceCountedBuffer<T> : OwnedBuffer<T>
 
         public override void Retain()
         {
+            if (IsDisposed) throw new InvalidOperationException();
```
ahsonkhan: Use BufferPrimitivesThrowHelper?

src/System.Buffers.Primitives/System/Buffers/OwnedArray.cs
```diff
@@ -0,0 +1,83 @@
...
+        int _referenceCount;
+
+        public OwnedArray(int length)
+        {
+            _array = new T[length];
+        }
+
+        public OwnedArray(T[] array)
+        {
+            if (array == null) throw new ArgumentNullException(nameof(array));
```
ahsonkhan: Use BufferPrimitivesThrowHelper?

src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -39,13 +42,11 @@ protected MemoryPoolBlock(MemoryPool pool, MemoryPoolSlab slab, int offset, int
...
-        public override Span<byte> Span
+        public override Span<byte> AsSpan(int index, int length)
         {
-            get
-            {
-                if (IsDisposed) PipelinesThrowHelper.ThrowObjectDisposedException(nameof(MemoryPoolBlock));
-                return new Span<byte>(Slab.Array, _offset, _length);
-            }
+            if (IsDisposed) PipelinesThrowHelper.ThrowObjectDisposedException(nameof(MemoryPoolBlock));
+            if (length > _length - index) throw new ArgumentOutOfRangeException();
```
ahsonkhan: Is this check redundant? https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Span.cs#L91

src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -94,39 +95,49 @@ public override string ToString()
...
-#if KESTREL_BY_SOURCE
-        internal
-#endif
-        protected override unsafe bool TryGetPointerAt(int index, out void* pointer)
+        public override BufferHandle Pin(int index = 0)
         {
             if (IsDisposed) PipelinesThrowHelper.ThrowObjectDisposedException(nameof(MemoryPoolBlock));
-            pointer = (Slab.NativePointer + _offset + index).ToPointer();
-            return true;
+            if (index > _length) throw new ArgumentException(nameof(index));
```
ahsonkhan: Use ThrowHelper.

src/System.IO.Pipelines/UnownedBuffer.cs
```diff
@@ -33,32 +33,58 @@ public OwnedBuffer<byte> MakeCopy(int offset, int length, out int newStart, out
             global::System.Buffer.BlockCopy(_buffer.Array, _buffer.Offset + offset, buffer, 0, length);
             newStart = 0;
             newEnd = length;
-            return buffer;
+            return (OwnedArray<byte>)buffer;
```
ahsonkhan: Why is this cast necessary?

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
KrzysztofCwalina: It's used in many subtypes. I would like to move it somewhere, but I am not sure where.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
KrzysztofCwalina: This is faster. The method is virtual, so the same optimizations we were able to do in Span probably will not work here. Moreover, it allows subclasses to materialize only part of the buffer representing this range.

src/System.Buffers.Primitives/System/Internal/OwnedArray.cs
KrzysztofCwalina: It requires fields for offset and length to be stored in OwnedArray.

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/OwnedBuffer.cs
KrzysztofCwalina: Yeah, I was wondering why this Slice method is still here. I will fix.

src/System.Buffers.Primitives/System/Buffers/BufferHandle.cs
KrzysztofCwalina: will do.

src/System.IO.Pipelines/MemoryPoolBlock.cs
KrzysztofCwalina: It's not. The slab array is very large and the span ctor will often work

src/System.IO.Pipelines/UnownedBuffer.cs
KrzysztofCwalina: implicit cast caannot go through two hops (array -> OwnedArray -> OwnedBuffer)

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
ahsonkhan: > This is faster. The method is virtual, so the same optimizations we were able to do in Span probably will not work here.

Does this mean that the [IndexOf that takes an index/count vs Slice](https://github.com/dotnet/corefxlab/pull/1278/) analysis doesn't apply to this? If so, and it is faster, then it is good to keep.

> Moreover, it allows subclasses to materialize only part of the buffer representing this range.

Why can't the subtypes just slice the span?

src/System.IO.Pipelines/MemoryPoolBlock.cs
ahsonkhan: Should it then be `if (length > _length - (_offset + index))`?

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
ahsonkhan: What about using [Unsafe.Add](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/ref/System.Runtime.CompilerServices.Unsafe.cs#L13)? We use it in Primitive Formatters a lot.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
KrzysztofCwalina: >Does this mean that the IndexOf that takes an index/count vs Slice analysis doesn't apply to this? If so, and it is faster, then it is good to keep.

It does not apply because IndexOf and Slice on Span are non-virtual, can be inlined, and code optimized out.

>Why exactly can't the subtypes just slice the span?

Let's say the buffer wraps memory mapped file. AsSpan(0, 10) needs to read only first ten bytes of the file. Span.Slice(0, 10) would first have to read the whole file and then create a span of just first ten bytes.

mjp41:
@KrzysztofCwalina I have been experimenting with your refactor today partly to enable `ReferenceCounter`.

I wonder if it makes sense to create a method that is called when a BufferHandle is released:
```csharp
   protected internal abstract void ReleaseHandle(); 
```

I tried to refactor the code a little to remove the lifetime management from the surface that `Buffer` sees, introducing a `BufferSource` that is what `Buffer` uses, which doesn't have the `Retain`/`Release` that are likely to map into reference counting.  This would enable making subclasses that don't use reference counting for `Pin`, while still having reference counting for lifetime management, which may please @davidfowl.  Then `OwnedBuffer` is a subclass which implements lifetime management.

I have pushed the first part of my experiment  https://github.com/mjp41/corefxlab/commit/944933ee55021bcb9c273035fbd87e6a91e3611d

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
KrzysztofCwalina: Unsafe.Add takes ref T. I need byte*

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
KrzysztofCwalina: fixed

src/System.IO.Pipelines/MemoryPoolBlock.cs
KrzysztofCwalina: No, because _length is not the length of the whole buffer, but rather the length of the section represented by this OwnedBuffer.

KrzysztofCwalina:
@mj1856, how is ReleaseHandle different from Release?

Also, BufferSource has RetainHandle and ReleaseHandle. How are these different from Retain/Release?

mjp41:
So `RetainHandle` constructs the Handle for accessing, but it is not required to be hooked up to any lifetime management.

`ReleaseHandle` is called when a `BufferHandle` is disposed.  But again it is not required to be hooked up with any lifetime management.  

At the moment, I hooked them up to calling the `OwnedBuffer` lifetime management routines, but they don't have to.  This allows for `BufferHandle` not to be lifetime managed even if the underlying `BufferSource` is by implementing the `OwnedBuffer`.

KrzysztofCwalina:
Sorry, I still don't understand the difference. OwnedBuffer.Retain/Release can do whatever the subclasses want, i.e. don't have to do lifetime management.

mjp41:
So for `BufferSegment`, the implementation requires `Retain` and `Release` to do lifetime management otherwise it won't work.  The `BufferHandle` in this case, will be forced to follow the same lifetime management strategy.  By adding this extra part, then it can do a different/non-existant strategy for the `BufferHandle`.  This would allow say to use `ReferenceCounter` for `BufferHandle`s and interlocked for the `Retain` and `Release` calls coming from `BufferSegments`.

KrzysztofCwalina:
Ah, I get it. You want to have different lifetime management strategy for stack bound handles and heap ownedbuffer.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
shiftylogic: Unsafe.AsPointer(Unsafe.Add(ref Unsafe.AsRef<T>(myPointer), offset))

shiftylogic: Might not be exactly that since I'm going from memory. But then you don't have to do the sizeof math and it all gets inlined anyways.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
ahsonkhan: >  Span.Slice(0, 10) would first have to read the whole file and then create a span of just first ten bytes.

I don't completely understand this. The cost of creating a span isn't proportional to the input buffer size it is wrapping, correct? So why does Span.Slice(0, 10) read the whole file whereas AsSpan(0, 10) does not?

KrzysztofCwalina: If you buffer is in memory, then you are right. But OwnedBuffer is an abstraction and not all it's data have to be in memory.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
KrzysztofCwalina: This is code that implementers of OwnedBuffers will have to write. I would prefer this be simpler than going back and forth between references and pointers. Let me leave the PR as-is and we can think separately what's the best way to provide convenient API for this scenario. I filed issue #1529

ahsonkhan: Is adding it to [S.R.CS.U](https://github.com/dotnet/corefx/blob/master/src/System.Runtime.CompilerServices.Unsafe/ref/System.Runtime.CompilerServices.Unsafe.cs) a valid option?

cc @jkotas

jkotas: Sounds good to me. S.R.CS.Unsafe has both `void*` and `ref T` overloads for most methods. `void*` overloads for Add and Subtract are missing for no good reason.

KrzysztofCwalina:
@davidfowl, any more feedback on this?

src/System.IO.Pipelines/UnownedBuffer.cs
```diff
@@ -33,32 +33,58 @@ public OwnedBuffer<byte> MakeCopy(int offset, int length, out int newStart, out
...
+            unsafe
+            {
+                Retain();
+                var handle = GCHandle.Alloc(_buffer.Array, GCHandleType.Pinned);
+                var pointer = Add((void*)handle.AddrOfPinnedObject(), index + _buffer.Offset);
+                return new BufferHandle(this, pointer, handle);
+            }
         }
 
         private ArraySegment<byte> _buffer;
```
davidfowl: Private fields belong at the top of the class.

KodrAus:
@KrzysztofCwalina @mjp41 Is there a reason `OwnedBuffer.Retain`/`OwnedBuffer.Release` isn't idempotent but `Buffer.Retain`/`Buffer.Release` over that owner is?

KrzysztofCwalina:
@KodrAus, they seem to be equivalent. Buffer.Retain, just calls OwnedBuffer.Retain.

KodrAus:
@KrzysztofCwalina I guess the difference is `OwnedBuffer.Retain` returns `void`, whereas `Buffer.Retain` returns a `BufferHandle`, which can only be `Release`d once when disposed.

Issue #1525, opened 2017-05-04
EncodeIntoUtf8Tests.OutputBufferEmpty(from: FromUtf16 & FromString) throws NullReferenceException


ahsonkhan:
Test (from https://github.com/dotnet/corefxlab/pull/1508):
```C#
string inputString = TextEncoderTestHelper.GenerateValidString(TextEncoderConstants.DataLength, 0, TextEncoderConstants.Utf8ThreeBytesLastCodePoint);

byte[] expectedBytes;
Span<byte> encodedBytes = Span<byte>.Empty;

int bytesWritten;
int charactersConsumed;
...
case TextEncoderTestHelper.SupportedEncoding.FromUtf16:
    temp = testEncoderUnicode.GetBytes(inputString);
    expectedBytes = Text.Encoding.Convert(testEncoderUnicode, testEncoder, temp);
    ReadOnlySpan<char> inputUtf16 = temp.AsSpan().NonPortableCast<byte, char>();
    Assert.False(utf8.TryEncode(inputUtf16, encodedBytes, out charactersConsumed, out bytesWritten));
    break;

case TextEncoderTestHelper.SupportedEncoding.FromString:
    temp = testEncoderUnicode.GetBytes(inputString);
    expectedBytes = Text.Encoding.Convert(testEncoderUnicode, testEncoder, temp);
    string inputStr = inputString;
    Assert.False(utf8.TryEncode(inputStr, encodedBytes, out bytesWritten));
    charactersConsumed = 0;
    break;
```

```C#
Failed   System.Text.Primitives.Tests.Encoding.EncodeIntoUtf8Tests.OutputBufferEmpty(from: FromUtf16)
Error Message:
 System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
   at System.Text.Utf8.Utf8Encoder.TryEncode(ReadOnlySpan`1 utf16, Span`1 utf8, Int32& charactersConsumed, Int32& bytesWritten)
   at System.Text.Utf8.Utf8TextEncoder.TryEncode(ReadOnlySpan`1 utf16, Span`1 data, Int32& charactersConsumed, Int32& bytesWritten)
   at System.Text.Primitives.Tests.Encoding.EncodeIntoUtf8Tests.OutputBufferEmpty(SupportedEncoding from)
```
```C#
Failed   System.Text.Primitives.Tests.Encoding.EncodeIntoUtf8Tests.OutputBufferEmpty(from: FromString)
Error Message:
[...]

PR #1542, opened 2017-05-13
Adding span.isempty checks to encoders. Fixes #1524, #1525.

### LINKED ISSUE #1524: EncodeIntoUtf8Tests.InputBufferEmpty(from: FromUtf16) throws NullReferenceException
Test (from https://github.com/dotnet/corefxlab/pull/1508):
```C#
temp = testEncoderUnicode.GetBytes(inputString);
expectedBytes = Text.Encoding.Convert(testEncoderUnicode, testEncoder, temp);
encodedBytes = Span<byte>.Empty;
Assert.True(utf8.TryEncode(ReadOnlySpan<char>.Empty, encodedBytes, out charactersConsumed, out bytesWritten));
Assert.Equal(0, charactersConsumed);
encodedBytes = new Span<byte>(new byte[1]);
Assert.True(utf8.TryEncode(ReadOnlySpan<char>.Empty, encodedBytes, out charactersConsumed, out bytesWritten));
Assert.Equal(0, charactersConsumed);
```
```C#
Failed   System.Text.Primitives.Tests.Encoding.EncodeIntoUtf8Tests.InputBufferEmpty(from: FromUtf16)
Error Message:
 System.NullReferenceException : Object reference not set to an instance of an object.
Stack Trace:
   at System.Text.Utf8.Utf8Encoder.TryEncode(ReadOnlySpan`1 utf16, Span`1 utf8, Int32& charactersConsumed, Int32& bytesWritten)
   at System.Text.Utf8.Utf8TextEncoder.TryEncode(ReadOnlySpan`1 utf16, Span`1 data, Int32& charactersConsumed, Int32& bytesWritten)
   at System.Text.Primitives.Tests.Encoding.EncodeIntoUtf8Tests.InputBufferEmpty(SupportedEncoding from)
```

cc @shiftylogic, @KrzysztofCwalina, @jkotas 

Related bug: https://github.com/dotnet/corefxlab/issues/1525
> The method should check for number of bytes on input. When they get zero bytes on input, they should not ever read anything from it. Thus it does not matter whether the pointer is null or not – you are not supposed to ever dereference it.
The failure suggests that TryEncode has buffer overrun bug and reading beyond the end of the buffer.

> Yes, it appears we are doing math on pointers and ~overflow~ underflow causes the checks to be incorrect when we have a null (i.e. zero) address. That results in us thinking we have room to write data.

ahsonkhan:
cc @shiftylogic, @jkotas, @KrzysztofCwalina.

Fixes issue #1524 and #1525.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
@@ -644,12 +644,6 @@ namespace System.Text.Utf8
         /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
         public static unsafe bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, out int charactersConsumed, out int bytesWritten)
         {
-            if (utf8.IsEmpty)
-            {
-                charactersConsumed = 0;
-                bytesWritten = 0;
-                return utf16.IsEmpty;
-            }
             //
             //
             // KEEP THIS IMPLEMENTATION IN SYNC WITH https://github.com/dotnet/corert/blob/master/src/System.Private.CoreLib/src/System/Text/UTF8Encoding.cs
@@ -842,7 +836,7 @@ namespace System.Text.Utf8
 
                     if (ch <= 0x7F)
                     {
-                        if (pTarget >= pAllocatedBufferEnd)
+                        if (pAllocatedBufferEnd - pTarget <= 0)
                             goto NeedMore;
 
                         *pTarget = (byte)ch;
@@ -853,7 +847,7 @@ namespace System.Text.Utf8
                     int chd;
                     if (ch <= 0x7FF)
                     {
-                        if (pTarget >= pAllocatedBufferEnd - 1)
+                        if (pAllocatedBufferEnd - pTarget <= 1)
                             goto NeedMore;
 
                         // 2 byte encoding
@@ -864,7 +858,7 @@ namespace System.Text.Utf8
                         // if (!IsLowSurrogate(ch) && !IsHighSurrogate(ch))
                         if (!InRange(ch, HIGH_SURROGATE_START, LOW_SURROGATE_END))
                         {
-                            if (pTarget >= pAllocatedBufferEnd - 2)
+                            if (pAllocatedBufferEnd - pTarget <= 2)
                                 goto NeedMore;
 
                             // 3 byte encoding
@@ -872,7 +866,7 @@ namespace System.Text.Utf8
                         }
                         else
                         {
-                            if (pTarget >= pAllocatedBufferEnd - 3)
+                            if (pAllocatedBufferEnd - pTarget <= 3)
                                 goto NeedMore;
 
                             // 4 byte encoding - high surrogate + low surrogate
--- a/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
@@ -52,19 +52,6 @@ namespace System.Text.Utf8
         /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
         public static unsafe bool TryDecode(ReadOnlySpan<byte> utf8, Span<char> utf16, out int bytesConsumed, out int charactersWritten)
         {
-            if (utf8.IsEmpty)
-            {
-                bytesConsumed = 0;
-                charactersWritten = 0;
-                return true;
-            }
-            if (utf16.IsEmpty)
-            {
-                bytesConsumed = 0;
-                charactersWritten = 0;
-                return false;
-            }
-
             fixed (byte* pUtf8 = &utf8.DangerousGetPinnableReference())
             fixed (char* pUtf16 = &utf16.DangerousGetPinnableReference())
             {
@@ -657,12 +644,6 @@ namespace System.Text.Utf8
         /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
         public static unsafe bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, out int charactersConsumed, out int bytesWritten)
         {
-            if (utf16.IsEmpty)
-            {
-                charactersConsumed = 0;
-                bytesWritten = 0;
-                return true;
-            }
             if (utf8.IsEmpty)
             {
                 charactersConsumed = 0;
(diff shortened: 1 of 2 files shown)


src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -52,6 +52,19 @@ internal static class Utf8Encoder
         /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
         public static unsafe bool TryDecode(ReadOnlySpan<byte> utf8, Span<char> utf16, out int bytesConsumed, out int charactersWritten)
         {
+            if (utf8.IsEmpty)
```
jkotas: It should not be needed here. The rest of the implementation seems to handle it fine.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -644,6 +657,18 @@ private static bool InRange(int ch, int start, int end)
         /// <returns>True if the input buffer was fully encoded into the output buffer, otherwise false.</returns>
         public static unsafe bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, out int charactersConsumed, out int bytesWritten)
         {
+            if (utf16.IsEmpty)
```
jkotas: I think it would be better to fix the loop below to subtract the pointers instead - like `while (pEnd - pSrc < 13) instead of this upfront check.

ahsonkhan: What about pSrc < pEnd?
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L828

ahsonkhan: Also, (0 - 0 < 13) == true and hence we still enter this while loop:https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L669

jkotas: I got the loop condition wrong - it should be `while (pEnd - pSrc > 13)`.

ahsonkhan: What about if the output buffer is empty (and input buffer is not)? pTarget is null and we don't return from the method if available = 0 (we just break from the fast loop and we enter the slow loop), which causes this to execute.
```C#
*pTarget = (byte)chd;
```
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L901

jkotas: These two checks

https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L861

https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L869

should be rewritten to subtract the pointers, similar to how to the fast loop condition was changed.

ahsonkhan: And this one?
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L839

ahsonkhan:

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
jkotas: I think it looks fine how you have made it look same as the other cases.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -847,7 +847,7 @@ public static unsafe bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, o
                     int chd;
                     if (ch <= 0x7FF)
                     {
-                        if (pTarget >= pAllocatedBufferEnd - 1)
+                        if (pAllocatedBufferEnd - pTarget <= 1)
```
shiftylogic: Now this requires 2 or more, but it should be 1 or more.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -858,15 +858,15 @@ public static unsafe bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, o
                         // if (!IsLowSurrogate(ch) && !IsHighSurrogate(ch))
                         if (!InRange(ch, HIGH_SURROGATE_START, LOW_SURROGATE_END))
                         {
-                            if (pTarget >= pAllocatedBufferEnd - 2)
+                            if (pAllocatedBufferEnd - pTarget <= 2)
```
shiftylogic: Now this expects 3 or more. I believe it should be 2 or more.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
```diff
@@ -858,15 +858,15 @@ public static unsafe bool TryEncode(ReadOnlySpan<char> utf16, Span<byte> utf8, o
...
                                 goto NeedMore;
 
                             // 3 byte encoding
                             chd = unchecked((sbyte)0xE0) | (ch >> 12);
                         }
                         else
                         {
-                            if (pTarget >= pAllocatedBufferEnd - 3)
+                            if (pAllocatedBufferEnd - pTarget <= 3)
```
shiftylogic: Requires 4+ now, should be 3+.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: x >= y -2
x - y >= -2
-(x-y) <= 2
-x+y <= 2
y - x <= 2

[x >= y - 2] == [y - x <= 2]
Where is the error?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
ahsonkhan: Are you suggesting it was wrong originally?

shiftylogic: Ignore this. I was thinking that we already wrote a byte and only needed space for 1 more, but we actually need space for 2 more.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
shiftylogic: Same as above.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs
shiftylogic: Ignore this. I was thinking that we already wrote a byte and only needed space for 2 more, but we actually need space for 3 more.

Issue #1540, opened 2017-05-12
Issue when advancing pipe reader when start and end are at a block boundary


cesarbs:
The code below enters an infinite loop if `result.Buffer.Start` and `result.Buffer.End` are at a block boundary i.e. `Start` is at the end of a block and `End` is at the beginning of the next block:

```csharp
private static async Task<ArraySegment<byte>> PeekAsyncAwaited(this IPipeReader pipelineReader,
    ReadableBufferAwaitable readingTask)
{
    while (true)
    {
        var result = await readingTask;

        try
        {
            if (!result.Buffer.IsEmpty)
            {
                var segment = result.Buffer.First;
                return segment.GetArray();
            }
            else if (result.IsCompleted)
            {
                return default(ArraySegment<byte>);
            }
        }
        finally
        {
            pipelineReader.Advance(result.Buffer.Start, result.Buffer.Start);
        }

        readingTask = pipelineReader.ReadAsync();
    }
}
```

The problem is in the line

```csharp
pipelineReader.Advance(result.Buffer.Start, result.Buffer.Start);
```

Even though the buffer IS empty, the next `ReadAsync` will return immediately as if more was available, even though there's none.

This can be mitigated by changing the call to

```csharp
pipelineReader.Advance(result.Buffer.Start, result.Buffer.IsEmpty
    ? result.Buffer.End
    : result.Buffer.Start);
```

cc @pakrym

pakrym:
Fixed by https://github.com/dotnet/corefxlab/pull/1546

PR #1546, opened 2017-05-16
Return all blocks when reader catches up to writer

pakrym:


src/System.IO.Pipelines/PipelinesThrowHelper.cs
```diff
@@ -153,6 +153,9 @@ private static string GetResourceString(ExceptionResource argument, string locat
                 case ExceptionResource.BackpressureDeadlock:
                     resourceString = "Advancing examined to the end would cause pipe to deadlock because FlushAsync is waiting";
                     break;
+                case ExceptionResource.AdvanceToInvalidCursor:
+                    resourceString = "Trying to advance to and invalid cursor";
```
davidfowl: This doesn't help me understand the problem (and it has typos 😉 )

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -412,35 +412,60 @@ void IPipeReader.Advance(ReadCursor consumed, ReadCursor examined)
...
+                        _length < _maximumSizeLow)
+                    {
+                        continuation = _writerAwaitable.Complete();
+                    }
+
+                    // Check if we consumed entire last segment
+                    // if we are going to return commit head
+                    // we need to check that there is no writing operation that
+                    // might be using tailspace
+                    if (consumed.Index == returnEnd.End &&
```
halter73: So this only returns consumed.Segment if the consumer reads the entire block? I see how this fixes #1540 (though even then, it would cause ReadAsync/Advance to be called one more time unnecessarily).

The way I'm reading this, I don't see how this would "Return all blocks when reader catches up to writer" when the writer partially fills the last block which I think would be most of the time. If that's a case, that's a wasted 8KB of request/response buffer space used by Kestrel per idle connection over 1.x.

Am I misunderstanding how this code works?

halter73: Discussed this is person. `returnEnd.End` is not the end of the block, just the end of segment.

PR #1553, opened 2017-05-17
New implementation of JsonWriter with encoder / formatting optimizations

shiftylogic:
Performance metrics:

| Test | New | Old | JSON.NET | JSON.NET + Span | Old / New | JSON.NET / New | JSON.NET / Span |
|---|---|---|---|---|---|---|---|
| Json Writer (Formatted, Utf8) | 1792 | 4484 | 2021 | 2556 | 0.400 | 0.887 | 1.265 |
| Json Writer (Non-formatted, Utf8) | 1495 | 1856 | 1364 | 1641 | 0.805 | 1.096 | 1.203 |
| Json Writer (Formatted, Utf16) | 1854 | 4517 | 1991 | 2739 | 0.410 | 0.931 | 1.376 |
| Json Writer (Non-formatted, Utf16) | 1538 | 1827 | 1352 | 1670 | 0.842 | 1.138 | 1.235 |


KrzysztofCwalina:
what are the units of the metrics? is higher number good or bad?

shiftylogic:
Elapsed time, in milliseconds, for 100,000 iterations of writing an object that ends up being approximately 6kb.

ahsonkhan:
What does the last column mean (i.e. JSON.NET / Span)?

src/System.Text.Json/System/Text/Json/JsonConstants.cs
```diff
@@ -0,0 +1,40 @@
...
+        public const char Quote = '"';
+
+        #endregion Control characters
+
+        #region Common values
+
+        public static readonly char[] TrueValue = { 't', 'r', 'u', 'e' };
+        public static readonly char[] FalseValue = { 'f', 'a', 'l', 's', 'e' };
+        public static readonly char[] NullValue = { 'n', 'u', 'l', 'l' };
+        public static readonly char[] UndefinedValue = { 'u', 'n', 'd', 'e', 'f', 'i', 'n', 'e', 'd' };
```
ahsonkhan: Are we sure we want to support undefined even though it is outside of the JSON spec? Can we instead define this behavior on the layer above the Reader/Writer which transforms it to null internally?

shiftylogic:
Ignore the "JSON.NET / Span" column. That was an experiment to create a JsonWriter implementation that could swap into the JSON.NET stack as a replacement for Newtonsoft.Json.JsonWriter.

src/System.Text.Json/System/Text/Json/JsonConstants.cs
shiftylogic: The processing of "undefined" has to happen at this layer even if we don't support it at a higher layer. Otherwise the parse fails because the undefined value is not quoted.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
         }
 
-        public void WriteArrayEnd()
+        /// <summary>
+        /// Write the starting tag of an array. This is used for adding an array to a
+        /// nested object. If this is used while inside a nested array, the property
+        /// name will be written and result in invalid JSON.
+        /// </summary>
+        /// <param name="name">The name of the property (i.e. key) within the containing object.</param>
+        public void WriteArrayStart(string name)
```
ahsonkhan: Is this overload necessary? What motivated the change where we use this instead of WriteAttribute + WriteArrayStart?

src/System.Text.Json/System/Text/Json/JsonConstants.cs
ahsonkhan: Shouldn't it fail? I am suggesting the higher layer transform undefined to null first before using the underlying Reader/Writer.

ahsonkhan:
> Performance metrics

Can you check-in the performance tests?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
         {
-            if (PrettyPrint) {
-                _formatter.Append('\n');
+            var output = _output;
+            if (UseFastUtf8)
+            {
+                if (output.Buffer.Length < 1)
+                    output.Enlarge();
+
+                output.Buffer.DangerousGetPinnableReference() = (byte)value;
```
ahsonkhan: Question: Why does this down casting from char to byte work if UseFastUtf8 is true?

src/System.Text.Json/System/Text/Json/JsonConstants.cs
shiftylogic: For strict conformance, I can remove it for now. If anyone makes a request for it, we can having this debate at that time.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: It seems cleaner. It seemed odd (and definitely not as efficient) to have to call WritePropertyName + WriteValue / WriteObjectStart / WriteArrayStart for every property. Combining them into one call was a performance bump and made the resulting code easier to read / parse in my head.

If we have a burning desire to keep the pattern from JSON.NET, then I can be convinced to change it. However, considering this is supposed to be an efficient / low-level layer, I liked this pattern.

shiftylogic:
Formatted vs Non-formatted is "pretty print" vs "raw" (i.e. no spaces, tabs, new lines).

I will be submitting a separate PR for the performance tests.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: Because this function is specifically internal and only used for ASCII control characters. I'll add a comment to this function to make that clear.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
 
 namespace System.Text.Json
 {
-    public struct JsonWriter<TFormatter> where TFormatter: ITextOutput
+    public struct JsonWriter
```
KrzysztofCwalina: Why did you decide to remove the type parameter?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
KrzysztofCwalina: Would be be better for this method to take byte and so force the caller to do the cast?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
+
+                output.Advance(2);
+            }
+            else
+            {
+                unsafe
+                {
+                    // Slow path, if we are dealing with non-invariant.
+                    ReadOnlySpan<char> chSpan = new ReadOnlySpan<char>(&value, 1);
+                    Write(chSpan);
```
KrzysztofCwalina: It would be good to add per benchmarks for non-invariant case. It's going to be slower, but we cannot be ridiculously slow.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
+            Write(value.AsSpan());
+            WriteChar(JsonConstants.Quote);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void WriteNumber(long value)
+        {
+            int written;
+            while (!value.TryFormat(_output.Buffer, out written, JsonConstants.NumberFormat, _output.Encoder))
+                _output.Enlarge();
```
KrzysztofCwalina: Would it make sense to check after enlarge that the buffer got actually larger. I worry about bad implementations of IOutput causing infinite loops that are hard to debug.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
+        private void WriteQuotedString(string value)
+        {
+            WriteChar(JsonConstants.Quote);
+            // TODO: We need to handle escaping.
+            Write(value.AsSpan());
+            WriteChar(JsonConstants.Quote);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void WriteNumber(long value)
```
KrzysztofCwalina: All these write methods implement a loop that is very similar to what System.Text.Formatting component does. Why cannot these methods be reusable? Can we somehow make them reusable so that every developer who wants to create a writer does not have to write the same boilerplate.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
KrzysztofCwalina: Enlarge takes a parameter. Should something be passed here?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
+                _output.Enlarge();
+
+            _output.Advance(written);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void WriteNumber(ulong value)
+        {
+            int written;
+            while (!value.TryFormat(_output.Buffer, out written, JsonConstants.NumberFormat, _output.Encoder))
```
KrzysztofCwalina: In parsing, we have inner classes for UTF8 and UTF16 invariant parsing that provide specialized fast parsing methods. Would it make sense to do the same here?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
+                _output.Enlarge();
+
+            _output.Advance(written);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void WriteDateTime(DateTime value)
```
KrzysztofCwalina: How do we chose the set of data types that this writer supports? i.e. Why DateTime is here? Also, how does a third party plug into this? For example, if I create my UtcTime types, how do I write it to the JSON output?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void WriteAscii(ReadOnlySpan<char> values)
         {
-            if (_prettyPrint) {
-                for (int i = 0; i < _indent; i++) {
-                    _formatter.Append("    ");
+            if (UseFastUtf8)
+            {
+                int len = values.Length;
+                Span<byte> buffer = _output.Buffer;
```
KrzysztofCwalina: _output.Buffer will be called many times. I wonder how much the perf would improve if we make this writer to be a Span-holding struct and cache the span.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -6,193 +6,592 @@
...
+            int indent = _indent;
+            while (indent-- >= 0)
+            {
+                WriteChar(JsonConstants.Space);
+                WriteChar(JsonConstants.Space);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void WriteSpacingUtf8(bool newline)
```
KrzysztofCwalina: This seems like a very long method to force its inlining. Also, is it even that hot?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: Not sure it makes a difference one way or the other. If byte makes it clear that this only supports ASCII, I'm fine with that. Note, this is a private / internal method, so only implementation methods will be calling it.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: It isn't needed anymore. I directly use the primitive formatters now.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: I'll give that a try to see how it performs.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: So, throw after the enlarge if the size doesn't change?

shiftylogic: Also, that means an extra check on every write.

shiftylogic: I could put a value in here based on the largest possible number (as a string of characters).

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: It is doing that, but the check happens in the TryFormat method.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: I tried to stay consistent with JSON.NET as much as I could. This is just a helper that writes DateTime as a string. The person parsing the JSON has to know the field is a DateTime and process accordingly. If you had a custom "time" class, you could just call your own method to convert to a string and call the WriteValue(string) version.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: I thought about this. Though, as it is, this is not a stack-only struct right now. Caching the buffer would make this stack-only.

FYI, there are other optimizations to be had in this code. I'll create a new work item to do more performance tweaks.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
ahsonkhan: Consider also making the constants that get passed in as args to this method bytes instead of char.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: This method gets called on every single write if you are formatting the JSON text. I can remove the inlining and see how it affects performance.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: I would have to change the constants to byte to make this change work. Otherwise, I'd be casting everywhere, which is ugly.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
KrzysztofCwalina: I was used to avoid boxing implementations of ITextOutut and to devirtualize calls to members of the interface.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
KrzysztofCwalina: The check would happen only on resize, not on every write.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
KrzysztofCwalina: So, it's not doing that :-)
The parser used to take encoding, did the check internally, and had fast path for known encodings. Ahson measured perf and discovered that just passing the parameter and doing checks was expensive given how cheap some of the actual work is. I think the ration of the check cost to the work cost is even higher here as writing a single char as UTF16 is super inexpensive.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
KrzysztofCwalina: That would allocate. There is an interface called IBufferFormattable to avoid allocations when writing 3rd party types to ITextOutput. Can the interface be used with this JSON writer?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: Without the forced inline attribute, it had a run-time of ~1.4x of what the inlined version has.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: None of the types I added overloads for actual allocate. They use the primitive formatter for all of this. So, though I said "writes DateTime as a string", it doesn't actually convert it to a string then write it.

The write methods are all there to stay consistent with JSON.NET. Enough people write dates and guids to JSON that these make sense. They have the benefit of being non-allocating. If they weren't there, most people would likely do a string convert and write the string. In the "efficient" case, they would still need a temporary byte buffer / Span to use the primitive formatter first.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: I tried to switch back to using the extension methods that exist for ITextOutput (i.e. the way it was before). This adds about a ~15% overhead.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: For reference, I disabled the shortcuts and used the encoder I was given, effectively taking all the slow paths for both UTF-8 and UTF-16. Below you can see the comparison. In the formatted scenarios, it makes a big difference. In the non-formatted scenarios, it makes much less difference.

| Test | New | JSON.NET | JSON.NET / New |
|---|---|---|---|
| Json Writer (True, Utf8) | 1825 | 1951 | 0.935 |
| Json Writer (False, Utf8) | 1499 | 1357 | 1.105 |
| Json Writer (True, Utf16) | 1845 | 1914 | 0.964 |
| Json Writer (False, Utf16) | 1511 | 1349 | 1.120 |
| Json Writer (True, Utf8-Slow) | 4197 | 2066 | 2.031 |
| Json Writer (False, Utf8-Slow) | 1608 | 1456 | 1.104 |
| Json Writer (True, Utf16-Slow) | 4415 | 2057 | 2.146 |
| Json Writer (False, Utf16-Slow) | 1646 | 1431 | 1.150 |

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: I made this change, which seems like the right thing to do. The way I did it, I ended up getting a small perf bump from what I had.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: See comment above. Are we willing to sacrifice the overhead of using these?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: I ended up doing a constant bump anytime we need to resize (currently set to 1024 bytes).

shiftylogic: New numbers based on this change:

| Test | New | JSON.NET | Ratio |
|---|---|---|---|
| Json Writer (True, Utf8) | 1716 | 2025 | 0.847 |
| Json Writer (False, Utf8) | 1586 | 1452 | 1.092 |
| Json Writer (True, Utf16) | 1767 | 2011 | 0.879 |
| Json Writer (False, Utf16) | 1518 | 1459 | 1.040 |
| Json Writer (True, Utf8-Slow) | 4180 | 2074 | 2.015 |
| Json Writer (False, Utf8-Slow) | 1620 | 1495 | 1.084 |
| Json Writer (True, Utf16-Slow) | 4239 | 2018 | 2.101 |
| Json Writer (False, Utf16-Slow) | 1648 | 1470 | 1.121 |

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: So, doing this gains me ~5-6%.

However, it is currently not possible without exposing some methods that are currently private on the primitive formatter. If we want to do this, I'll file an issue and do it as a separate check-in since it will require some changes / re-structuring of the primitive formatter(s).

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: Filed #1556.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: Covered under #1556

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: Filed #1557

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -342,42 +321,36 @@ private void WriteStartAttribute(string name)
...
             else if (UseFastUtf16)
             {
-                if (output.Buffer.Length < 2)
-                    output.Enlarge(2);
-
                 unsafe
                 {
-                    Unsafe.AsRef<char>(Unsafe.AsPointer<byte>(ref output.Buffer.DangerousGetPinnableReference())) = value;
+                    var buffer = EnsureBuffer(2);
+                    Unsafe.AsRef<char>(Unsafe.AsPointer<byte>(ref buffer.DangerousGetPinnableReference())) = (char)value;
```
ahsonkhan: Is there a way to use Unsafe.Write/WriteUnaligned instead of having this unsafe block and getting a pointer?

shiftylogic: Yes, I can change this to a different construct. Unsafe.As<byte, char> also does not require unsafe block and is good for writing a single char.

PR #1558, opened 2017-05-18
Initial optimizations for the JsonReader

shiftylogic:
NOTE: This change builds on top of #1553. So you can safely ignore JsonWriter, JsonConstants, JsonEncoderState, and any tests related to the writer. Once #1553 gets merged, I'll do a final rebase of this PR. Just wanted to get the PR started so people could start looking at it.

* Now supports UTF-16 and UTF-8.
* Properly supports nested objects and arrays in any combination.
* Due to above fix from previous implementation, the current implementation only supports 64 levels of nesting at the moment.
* Still work to do for the "slow" (non-invariant) cases.
* I think we need to add some helper methods similar to JSON.NET that convert the resulting Span Value properties to a desired type with a single call (instead of a Read then conversion). Though, these would make sense to be extension methods on the reader.

Current Performance Results:

| Test | New | Old | JSON.NET | New / Old | New / JSON.NET |
|---|---|---|---|---|---|
| Json Reader (Utf8) | 57 | 75 | 262 | 0.760 | 0.218 |
| Json Reader (Utf16) | 67 | 76 | 255 | 0.882 | 0.263 |


src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
+        /// Gets the encoder instance used when the reader was constructed.
+        /// </summary>
+        public TextEncoder Encoder => _encoder;
+
+        /// <summary>
+        /// Constructs a new JsonReader instance. This is a stack-only type.
+        /// </summary>
+        /// <param name="data">The <see cref="Span{byte}"/> value to consume. </param>
+        /// <param name="encoder">An encoder used for decoding bytes from <paramref name="data"/> into characters.</param>
+        public JsonReader(ReadOnlySpan<byte> data, TextEncoder encoder)
```
davidfowl: Does this assume the entire JSON payload exists in a single buffer?

shiftylogic: This version, yes, just like the previous one. However, this is a work-in-progress and I want to start using ReadOnlyBytes / ReadableBuffer once we solidify on whatever that design is. I'm pretty confident that this should be relatively easy to modify for that concept.

davidfowl: My fear is that it'll end up throwing away a bunch of things you did here. Designing for incrementally I think fundamentally changes things.

KrzysztofCwalina:
The last collumn of the table should be New / JSON.NET, shouldn't it? Same with the second to last.

src/System.Text.Json/System/Text/Json/JsonReader.cs
KrzysztofCwalina: Since the PR is ready and done, we would be throwing away work anyway, and so I think we should just merge. Possibly not that much will be thrown away, and we will be able to observe how support for multi-span buffers impacts perf.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -25,54 +25,54 @@ public static JsonDynamicObject Parse(ReadOnlySpan<byte> utf8, int expectedNumbe
...
             while (reader.Read())
             {
                 switch (reader.TokenType)
                 {
-                    case JsonReader.JsonTokenType.Property:
-                        var name = reader.GetName();
-                        var type = reader.GetJsonValueType();
-                        var value = reader.GetValue();
+                    case JsonTokenType.PropertyName:
+                        var name = new Utf8String(reader.Value);
```
KrzysztofCwalina: I think we should avoid taking dependencies on Utf8String. I worry that we will ship the JSON APIs before we finish the design of Utf8String.

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
-        private bool _jsonStartIsObject;
-
-        public enum JsonTokenType
+        // We are using a ulong to represent our nested state, so we can only go 64 levels deep.
+        private const int MaxDepth = sizeof(ulong) * 8;
+
+        private readonly JsonEncoderState _encoderState;
+        private readonly TextEncoder _encoder;
+
+        private ReadOnlySpan<byte> _buffer;
```
KrzysztofCwalina: So, writer is not a ref-like type, but reader is?

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
-            TokenType = 0;
-            _jsonStartIsObject = (byte)_str[0] == '{';
+            ref byte bytes = ref _buffer.DangerousGetPinnableReference();
+            int count = _buffer.Length;
+            int skip = SkipWhiteSpace(ref bytes, count);
+
+            ref byte next = ref Unsafe.Add(ref bytes, skip);
+            count -= skip;
+
+            int step = GetNextCharAscii(ref next, count, out char ch);
```
KrzysztofCwalina: Why just ASCII? What happens if the property name contains non-ascii characters? Can property names be non-ASCII?

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
+        private void StartObject()
         {
-            _str = new Utf8String(str).Trim();
-            _index = 0;
-            _insideObject = 0;
-            _insideArray = 0;
-            TokenType = 0;
-            _jsonStartIsObject = (byte)_str[0] == '{';
+            if (_depth > MaxDepth)
+                throw new JsonReaderException();
```
KrzysztofCwalina: Would it  make sense to use throw helpers so that the method can be inlined?

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
+        private int ConsumeNumber(ref byte src, int count, bool negative)
+        {
+            if (UseFastUtf8)
             {
-                return JsonValueType.Number;
+                unsafe
+                {
+                    fixed (byte* pSrc = &src)
+                    {
+                        if (!PrimitiveParser.InvariantUtf8.TryParseDecimal(pSrc, count, out decimal value, out int consumed))
```
KrzysztofCwalina: Do you know what's the hit if you'd use safe (span) parser?

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
+                    }
+                }
+            }
+            else
+                throw new NotImplementedException();
         }
 
-        public Utf8String GetValue()
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int ConsumeNaN(ref byte src, int count)
```
KrzysztofCwalina: I think "count" should be "length", i.e. we use "length" for sizes of buffers.

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int ConsumeNull(ref byte src, int count)
         {
-            _index++;
-            var count = _index;
-            do
+            Value = ReadOnlySpan<byte>.Empty;
+            ValueType = JsonValueType.Null;
+
+            if (UseFastUtf8)
```
KrzysztofCwalina: Also, I wonder if instead of all the branches, it would not be better to have two parsers and do the check at the beginning, i.e. in Json.Parse(...)

src/System.Text.Json/System/Text/Json/JsonReaderException.cs
```diff
@@ -0,0 +1,12 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System;
+
+namespace System.Text.Json
+{
+    public class JsonReaderException : Exception
+    {
+        public JsonReaderException() : base() { }
```
KrzysztofCwalina: Why do we need this type? i.e. when would this exception be handled programatically?

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
shiftylogic: This was just patching existing code that already assumed Utf8String.

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: That is currently true. I'm wondering if the right strategy here is to create something like an ITextInput similar to ITextOutput but for reading that would break this issue.

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: This function is only used for reading characters that result in state changes (same as Control characters in the writer). I could probably rename this or add a comment to reflect this.

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: Yes. The JsonReaderException was actually a placeholder for something else. I just hadn't figured out what that was yet.

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: It is similar to the test I did in the writer (~3-5% if I recall correctly).

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: Yes, I thought about that. Ahson and I were talking about it yesterday. That has its own challenges though.

src/System.Text.Json/System/Text/Json/JsonReaderException.cs
shiftylogic: This was a placeholder for something that I haven't really figured out yet. Right now, we fail on a bad / unparsable stream. Long-term, I think it might be wiser to expose the errors in a different way.

src/System.Text.Json/System/Text/Json/JsonReader.cs
davidfowl: I'm not saying it shouldn't be merged, I'm just saying the design might completely change is all so i can't really look at the perf results until then.

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: FYI, after preparing the performance tests for a PR, I discovered that the decimal parser in the primitive parser class is allocating memory. It is building up a string using basic string concatenation. I've opened #1560 to track it.

tests/System.Text.Json.Tests/JsonReaderTests.cs
```diff
@@ -91,15 +92,15 @@ public void ReadFullJsonSchemaAndGetValue()
...
+        [Fact(Skip = "This test is injecting invalid characters into the stream and needs to be re-visited")]
         public void ReadJsonSpecialStrings()
         {
             var readJson = ReadJson(TestJson.JsonWithSpecialStrings);
             var json = readJson.ToString();
             Assert.Equal(json, TestJson.ExpectedJsonWithSpecialStrings);
         }
 
-        [Fact]
+        [Fact(Skip = "The current primitive parsers do not support E-notation for numbers.")]
```
ahsonkhan: Do we already have an issue for this?

src/System.Text.Json/System/Text/Json/JsonTokenType.cs
```diff
@@ -0,0 +1,19 @@
...
+        None,
+        StartObject,
+        EndObject,
+        StartArray,
+        EndArray,
+        PropertyName,
+        Comment,
+        Value,
+        Null,
+        Undefined,
```
ahsonkhan: Should we add support to undefined only after it gets requested rather than preemptively?

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
-        private Utf8String _str;
-        private int _index;
-        private int _insideObject;
-        private int _insideArray;
-        public JsonTokenType TokenType;
-        private bool _jsonStartIsObject;
-
-        public enum JsonTokenType
+        // We are using a ulong to represent our nested state, so we can only go 64 levels deep.
+        private const int MaxDepth = sizeof(ulong) * 8;
```
ahsonkhan: What is our plan to work around this limit or is this a constraint we want to stick with?

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
+            TokenType = JsonTokenType.EndArray;
         }
 
-        public JsonValueType GetJsonValueType()
+        /// <summary>
+        /// This method consumes the next token regardless of whether we are inside an object or an array.
+        /// For an object, it reads the next property name token. For an array, it just reads the next value.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int ConsumeNext(char marker, int markerBytes, ref byte src, int count)
```
ahsonkhan: Change marker from char to byte?

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -1,383 +1,836 @@
...
+                unsafe
+                {
+                    fixed (byte* pSrc = &src)
+                    {
+                        if (!PrimitiveParser.InvariantUtf8.TryParseDecimal(pSrc, count, out decimal value, out int consumed))
+                            throw new JsonReaderException();
+
+                        // NOTE: For now, we are throwing away the resulting value. It was only used to consume all the
+                        //       numbers and discover the slice of the buffer we needed. We may want to figure out how
+                        //       to retain this value for the cases where the caller actually wants the parsed value.
```
ahsonkhan: Instead of using primitive parser and actually parsing, is there a way to just move the cursor (and validate) but not actually get the number (which we throw away)? I would imagine the ReadInt/etc extension methods would do this instead.

src/System.Text.Json/System/Text/Json/JsonValueType.cs
```diff
@@ -0,0 +1,21 @@
...
+        Unknown,
+        Object,
+        Array,
+        Number,
+        String,
+        True,
+        False,
+        Null,
+        Undefined,
+        NaN,
```
ahsonkhan: Interesting discussion on having the Reader be more lenient than the JSON spec:
http://stackoverflow.com/questions/15228651/how-to-parse-json-string-containing-nan-in-node-js

http://static.javadoc.io/com.google.code.gson/gson/2.8.0/com/google/gson/GsonBuilder.html#serializeSpecialFloatingPointValues--

> Section 2.4 of JSON specification disallows special double values (NaN, Infinity, -Infinity). However, Javascript specification (see section 4.3.20, 4.3.22, 4.3.23) allows these values as valid Javascript values. Moreover, most JavaScript engines will accept these special values in JSON without problem. So, at a practical level, it makes sense to accept these values as valid JSON even though JSON specification disallows them.

> Gson always accepts these special values during deserialization. However, it outputs strictly compliant JSON. Hence, if it encounters a float value Float.NaN, Float.POSITIVE_INFINITY, Float.NEGATIVE_INFINITY, or a double value Double.NaN, Double.POSITIVE_INFINITY, Double.NEGATIVE_INFINITY, it will throw an IllegalArgumentException. This method provides a way to override the default behavior when you know that the JSON receiver will be able to handle these special values.

ahsonkhan: Should we make this an option that the user can toggle? Or the inverse, have a more strict reader that only allows valid JSON?

tests/System.Text.Json.Tests/JsonReaderTests.cs
shiftylogic: Not sure. I'll check and file if necessary.

src/System.Text.Json/System/Text/Json/JsonTokenType.cs
shiftylogic: I'm leaving this in. It is implemented and adds value. Not worth deleting unless we decide we aren't doing it when we move to release.

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: For now, a 64 level limit is fine. We will re-visit this later.

src/System.Text.Json/System/Text/Json/JsonReader.cs
shiftylogic: I'm removing the parsing and making the caller do this.

src/System.Text.Json/System/Text/Json/JsonValueType.cs
shiftylogic: For this PR, I'm leaving as it is.

PR #1561, opened 2017-05-18
Added JsonObject.ArrayLength property

Bobris:
fixed bug in counting nested arrays
added and improved tests

### DIFF
--- a/src/System.Text.Json/System/Text/Json/JsonParseObject.cs
+++ b/src/System.Text.Json/System/Text/Json/JsonParseObject.cs
@@ -45,7 +45,7 @@ namespace System.Text.Json
                 throw new KeyNotFoundException();
             }
             if (record.Type != JsonValueType.Object) {
-                throw new NullReferenceException();
+                throw new InvalidOperationException();
             }
 
             for (int i = DbRow.Size; i <= _db.Length; i += DbRow.Size) {
@@ -89,7 +89,7 @@ namespace System.Text.Json
             }
 
             if (record.Type != JsonValueType.Object) {
-                throw new NullReferenceException();
+                throw new InvalidOperationException();
             }
 
             for (int i = DbRow.Size; i <= _db.Length; i += DbRow.Size) {
@@ -154,7 +154,7 @@ namespace System.Text.Json
                 }
 
                 if (record.Type != JsonValueType.Array) {
-                    throw new NullReferenceException();
+                    throw new InvalidOperationException();
                 }
 
                 int counter = 0;
@@ -182,14 +182,14 @@ namespace System.Text.Json
             }
         }
 
-        public int Length
+        public int ArrayLength
         {
             get
             {
                 var record = Record;
                 if (record.Type != JsonValueType.Array)
                 {
-                    throw new NullReferenceException();
+                    throw new InvalidOperationException();
                 }
                 return record.Length; 
             }
--- a/src/System.Text.Json/System/Text/Json/JsonParseObject.cs
+++ b/src/System.Text.Json/System/Text/Json/JsonParseObject.cs
@@ -182,6 +182,19 @@ namespace System.Text.Json
             }
         }
 
+        public int Length
+        {
+            get
+            {
+                var record = Record;
+                if (record.Type != JsonValueType.Array)
+                {
+                    throw new NullReferenceException();
+                }
+                return record.Length; 
+            }
+        }
+
         public static explicit operator string(JsonObject json)
         {
             var utf8 = (Utf8String)json;
--- a/src/System.Text.Json/System/Text/Json/JsonParser.cs
+++ b/src/System.Text.Json/System/Text/Json/JsonParser.cs
@@ -190,7 +190,7 @@ namespace System.Text.Json
                         break;
                     case JsonTokenType.ArrayEnd:
                         _db.Span.Slice(FindLocation(_stack.ArrayStackCount - 1, false)).Write<int>(arrayItemsCount);
-                        arrayItemsCount += _stack.PopArray();
+                        arrayItemsCount = _stack.PopArray();
                         break;
                     case JsonTokenType.Property:
                         ParsePropertyName();
--- a/tests/System.Text.Json.Tests/JsonObjectTests.cs
+++ b/tests/System.Text.Json.Tests/JsonObjectTests.cs
@@ -15,12 +15,13 @@ namespace System.Text.Json.Tests
         {
             var buffer = StringToUtf8BufferWithEmptySpace(TestJson.SimpleArrayJson, 60);
             using (var parsedObject = JsonObject.Parse(buffer.AsSpan())) {
+                Assert.Equal(2, parsedObject.ArrayLength);
+
                 var phoneNumber = (string)parsedObject[0];
                 var age = (int)parsedObject[1];
 
-                Assert.Equal(phoneNumber, "425-214-3151");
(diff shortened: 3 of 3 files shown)


KrzysztofCwalina:
Thanks! I think it's a nice addition.

src/System.Text.Json/System/Text/Json/JsonParseObject.cs
```diff
@@ -182,6 +182,19 @@ public bool TryGetValue(string propertyName, out JsonObject value)
             }
         }
 
+        public int Length
```
KrzysztofCwalina: I wonder if it should not be called ArrayLength given it works only for arrays.

Bobris: No problem, will rename.

src/System.Text.Json/System/Text/Json/JsonParseObject.cs
```diff
@@ -182,6 +182,19 @@ public bool TryGetValue(string propertyName, out JsonObject value)
...
+        {
+            get
+            {
+                var record = Record;
+                if (record.Type != JsonValueType.Array)
+                {
+                    throw new NullReferenceException();
```
KrzysztofCwalina: This should throw InvalidOperationException, which indicates that the call could work if the object was in some other state, i.e. if the JsonObject was an array.

src/System.Text.Json/System/Text/Json/JsonParser.cs
```diff
@@ -190,7 +190,7 @@ public JsonObject Parse(ReadOnlySpan<byte> utf8Json, BufferPool pool = null)
                         break;
                     case JsonTokenType.ArrayEnd:
                         _db.Span.Slice(FindLocation(_stack.ArrayStackCount - 1, false)).Write<int>(arrayItemsCount);
-                        arrayItemsCount += _stack.PopArray();
+                        arrayItemsCount = _stack.PopArray();
```
ahsonkhan: What motivated this change? If this is a bug, can you please add a test that exposes it?

tests/System.Text.Json.Tests/JsonObjectTests.cs
```diff
@@ -20,6 +20,7 @@ public void ParseArray()
 
                 Assert.Equal(phoneNumber, "425-214-3151");
                 Assert.Equal(age, 25);
+                Assert.Equal(parsedObject.Length, 2);
```
KrzysztofCwalina: You fixed the order issue below, i.e. expected first, then actual. Could you please fix the order here as well? I realize you did not write all these asserts, but it would be good to be consistent (and correct)

src/System.Text.Json/System/Text/Json/JsonParseObject.cs
Bobris: I copied this from TryGetValue just to be consistent. Should I change it also there?

tests/System.Text.Json.Tests/JsonObjectTests.cs
ahsonkhan: I would move this check before `var phoneNumber = (string)parsedObject[0];` because if this fails, we don't want to get out of range exception.

src/System.Text.Json/System/Text/Json/JsonParser.cs
Bobris: I did added just new assert to existing tests.

tests/System.Text.Json.Tests/JsonObjectTests.cs
```diff
@@ -70,14 +71,34 @@ public void ParseNestedJson()
...
 
-                Assert.Equal(age, 30);
-                Assert.Equal(first, "John");
-                Assert.Equal(last, "Smith");
-                Assert.Equal(phoneNum1, "425-000-1212");
-                Assert.Equal(phoneNum2, "425-000-1213");
-                Assert.Equal(street, "1 Microsoft Way");
-                Assert.Equal(city, "Redmond");
-                Assert.Equal(zipCode, 98052);
+                Assert.Equal(1, parsedObject.Length);
```
ahsonkhan: Same as above, move this to before the line `var person = parsedObject[0];`

tests/System.Text.Json.Tests/JsonObjectTests.cs
```diff
@@ -70,14 +71,34 @@ public void ParseNestedJson()
...
+                Assert.Equal(30, age);
+                Assert.Equal("John", first);
+                Assert.Equal("Smith", last);
+                Assert.Equal(2, phoneNums.Length);
```
ahsonkhan: Move this check to right after `var phoneNums = person["phoneNumbers"];`

src/System.Text.Json/System/Text/Json/JsonParseObject.cs
KrzysztofCwalina: Yeah, I just noticed TryGetValue does the same thing. @ahsonkhan, is there a reason for NullRefException here? If not I would appreciate if @Bobris changed TryGetValue too.

tests/System.Text.Json.Tests/JsonObjectTests.cs
```diff
@@ -70,14 +71,34 @@ public void ParseNestedJson()
...
+                Assert.Equal("Redmond", city);
+                Assert.Equal(98052, zipCode);
+                try
+                {
+                    var _ = person.Length;
+                    throw new Exception("Never get here");
+                }
+                catch (Exception ex)
+                {
+                    Assert.IsType<NullReferenceException>(ex);
```
ahsonkhan: Nit: Use Assert.Throws instead of the try/catch similar to https://github.com/dotnet/corefxlab/blob/master/tests/System.Buffers.Experimental.Tests/MemoryTests.cs#L83-L86

tests/System.Text.Json.Tests/JsonObjectTests.cs
```diff
@@ -70,14 +71,34 @@ public void ParseNestedJson()
...
+                }
+                try
+                {
+                    var _ = phoneNums[2];
+                    throw new Exception("Never get here");
+                }
+                catch(Exception ex)
+                {
+                    Assert.IsType<IndexOutOfRangeException>(ex);
```
ahsonkhan: Same as above.

tests/System.Text.Json.Tests/JsonObjectTests.cs
Bobris: Unfortunately it is not possible because it would move Span<T> to heap. I did try of course.

tests/System.Text.Json.Tests/JsonObjectTests.cs
Bobris: Same as above.

src/System.Text.Json/System/Text/Json/JsonParseObject.cs
ahsonkhan: InvalidOperationException is more appropriate. Feel free to change the exception type. Tyvm :)

tests/System.Text.Json.Tests/JsonObjectTests.cs
ahsonkhan: Ah yes. Of course. My mistake.

KrzysztofCwalina:
Thanks!

PR #1569, opened 2017-05-23
Cleaned up System.Buffers.Primitives

KrzysztofCwalina:
Various API cleanup
Moved as much as possible out from System.Buffers.Primitives into System.Buffers.Experimental
Made as many types internal as possible/reasonable
Added reference tests for Buffer<T>

cc: @ahsonkhan, @shiftylogic, @joshfree


src/System.Buffers.Primitives/System/Buffers/Buffer.cs
```diff
@@ -106,14 +106,20 @@ public BufferHandle Pin()
 
         public bool TryGetArray(out ArraySegment<T> buffer)
         {
-            if (!_owner.TryGetArray(out buffer))
+            if (_owner!=null && _owner.TryGetArray(out var segment))
```
shiftylogic: nit: Spacing around _owner comparison

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
```diff
@@ -63,5 +69,7 @@ public void Dispose()
         {
             return (byte*)pointer + ((ulong)Unsafe.SizeOf<T>() * (ulong)offset);
         }
+
+        internal static readonly T[] EmptyArray = new T[0];
```
shiftylogic: Should this be Array.Empty?

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
```diff
@@ -106,14 +106,20 @@ public BufferHandle Pin()
...
+            if (_array != null)
             {
-                if (_array == null) return false;
                 buffer = new ArraySegment<T>(_array, _index, _length);
                 return true;
             }
-            buffer = new ArraySegment<T>(buffer.Array, buffer.Offset + _index, _length);
-            return true;
+
+            buffer = default(ArraySegment<T>);
```
ahsonkhan: Will there ever be a scenario where both _array and _owner are both null causing us to get here?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -98,18 +98,6 @@ public BufferHandle Pin()
             }
         }
 
-        public bool TryGetArray(out ArraySegment<T> buffer)
```
ahsonkhan: How come we don't need TryGetArray on ReadOnlyBuffer anymore?

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
```diff
@@ -20,36 +22,44 @@ public static ManagedBufferPool Shared
...
             public override Span<byte> AsSpan(int index, int length)
             {
-                if (IsDisposed) BufferPrimitivesThrowHelper.ThrowObjectDisposedException(nameof(ManagedBufferPool));
+                if (IsDisposed) BufferPrimitivesThrowHelper.ThrowObjectDisposedException(nameof(ArrayPoolBuffer));
                 return new Span<byte>(_array, index, length);
             }
 
             protected override void Dispose(bool disposing)
             {
+                _disposed = true;
```
ahsonkhan: We don't use the disposing argument here.

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
```diff
@@ -70,7 +80,17 @@ public override BufferHandle Pin(int index = 0)
...
-            byte[] _array;
+            public override void Retain()
+            {
+                if (IsDisposed) BufferPrimitivesThrowHelper.ThrowObjectDisposedException(nameof(ArrayPoolBuffer));
+                Interlocked.Increment(ref _referenceCount);
+            }
+
+            public override void Release()
+            {
+                Debug.Assert(!IsDisposed);
```
ahsonkhan: Should this be a Debug.Assert? What is preventing the caller to call Release multiple times?

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
```diff
@@ -85,5 +89,233 @@ static void TestLifetime(OwnedBuffer<byte> owned)
...
+            {
+                Assert.Equal(0, owned.OnNoRefencesCalledCount);
+                Assert.True(owned.IsRetained);
+            }
+            Assert.Equal(1, owned.OnNoRefencesCalledCount);
+            Assert.False(owned.IsRetained);
+        }
+
+        [Fact(Skip = "This needs to be fixed and re-enabled or removed.")]
+        public void RacyAccess()
```
ahsonkhan: Is this test still broken?

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
```diff
@@ -85,5 +89,233 @@ static void TestLifetime(OwnedBuffer<byte> owned)
...
+            Assert.Equal(false, owned.IsDisposed);
+            var reservation = memory.Retain();
+            Assert.Equal(false, owned.IsDisposed);
+            owned.Release();
+            Assert.Equal(false, owned.IsDisposed);
+            reservation.Dispose();
+            Assert.Equal(true, owned.IsDisposed);
+        }
+
+
```
ahsonkhan: nit: extra spacing.

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
```diff
@@ -1,14 +1,18 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 using System.Buffers;
+using System.Diagnostics;
+using System.Runtime.InteropServices;
+using System.Threading;
+using System.Threading.Tasks;
 using Xunit;
 
-namespace System.Slices.Tests
+namespace System.Buffers.Tests
```
ahsonkhan: Should this be in namespace System.Buffers.Experimental.Tests?

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
```diff
@@ -10,176 +10,10 @@ namespace System.Buffers.Tests
...
-            owned.Release();
-            Assert.Equal(false, owned.IsDisposed);
-            reservation.Dispose();
-            Assert.Equal(true, owned.IsDisposed);
-        }
-
-        [Fact]
         public void MemoryHandleFreeUninitialized()
         {
             var h = default(BufferHandle);
```
ahsonkhan: nit: Rename h to handle.

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
ahsonkhan: change var to ArraySegment<T>?

KrzysztofCwalina: fixed

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
KrzysztofCwalina: We don't have it in NS 1.1, at least I don't see it.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
KrzysztofCwalina: You can get here if _owner.TryGetArray returns false (because the buffer is not array bases).

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
KrzysztofCwalina: I don't think there is a difference between the disposing and finalization code path.

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
KrzysztofCwalina: It is Debug.Assert.

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
KrzysztofCwalina: I am pretty sure it it.

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
KrzysztofCwalina: will fix.

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
KrzysztofCwalina: The namespaces in this assembly are very inconsitent. I changed them all to System.Buffers.Tests. I don't think tests need to say "Experimental"

tests/System.Buffers.Primitives.Tests/MemoryTests.cs
KrzysztofCwalina: will do

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
KrzysztofCwalina: You can get here if _owner.TryGetArray returns false (because the buffer is not array bases).

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
KrzysztofCwalina: I removed it because it is a safe API that gives r/w access to read-only buffer.

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
ahsonkhan: In that case, should buffer = segment (i.e. what you get as the out parameter of TryGetArray) since we already have it?

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
ahsonkhan: I don't follow. What is the behavior if someone calls Dispose(false) vs Dispose(true)?

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
ahsonkhan: Yes, but in release builds, what happens if IsDisposed is true?

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
ahsonkhan: Is it ok to have the same namespace and class name in different assemblies?

- System.Buffers.Experimental.Tests contains System.Buffers.Tests.MemoryTests
- System.Buffers.Primitives.Tests contains System.Buffers.Tests.MemoryTests

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
KrzysztofCwalina: What would it buy us?

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
KrzysztofCwalina: Same behavior

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
KrzysztofCwalina: The ref count will possibly get below zero. I could do a release check, but it does not really address the problem, i.e. an invalid double release could suceed, and then valid single release would throw, e.g.
```c#
void Bar() {
    OwnedBuffer<byte> buffer = ...
    buffer.Retain();
    Foo(buffer);
    buffer.Release(); // this will throw despite the fact that this method is correct 
}
void Foo(OwnedBuffer<T> buffer){
    buffer.Retain();
    buffer.Relase();
    buffer.Release();
}
```

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
KrzysztofCwalina: Yes, we have System.Uri in System.dll and System.Object in mscorlib.dll

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
KrzysztofCwalina: After chatting, I added checks as follows:
```c#
            if (!IsRetained) BuffersExperimentalThrowHelper.ThrowInvalidOperationException();
```

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
KrzysztofCwalina: But we should not have more than one type with the same name and the same namespace. So, I fixed.

src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -106,11 +106,13 @@ protected override void Dispose(bool disposing)
 
         public override void Retain()
         {
+            if (IsDisposed) PipelinesThrowHelper.ThrowObjectDisposedException(nameof(UnownedBuffer));
```
ahsonkhan: nit: nameof(MemoryPoolBlock), not UnownedBuffer

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
```diff
@@ -246,7 +243,8 @@ public override void Retain()
 
         public override void Release()
         {
-            Debug.Assert(!IsDisposed);
+            if (!IsRetained) new InvalidOperationException();
```
ahsonkhan: missing throw keyword.

KrzysztofCwalina: will fix

src/System.IO.Pipelines/MemoryPoolBlock.cs
KrzysztofCwalina: will fix

PR #1572, opened 2017-05-24
Added some more reference tests for Buffer<T> and ROBuffer<T>

KrzysztofCwalina:
@ahsonkhan, @shiftylogic, @joshfree

### DIFF
--- a/tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
+++ b/tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
@@ -190,6 +190,12 @@ namespace System.Buffers.Tests
             BufferBasics(create());
             BufferLifetime(create());
         }
+        
+        public static void TestBuffer(Func<ReadOnlyBuffer<byte>> create)
+        {
+            BufferBasics(create());
+            BufferLifetime(create());
+        }
 
         static void BufferBasics(Buffer<byte> buffer)
         {
@@ -236,6 +242,38 @@ namespace System.Buffers.Tests
             //using(var handle = buffer.Retain())
             //{
             //}
+        }
+
+        static void BufferBasics(ReadOnlyBuffer<byte> buffer)
+        {
+            var span = buffer.Span;
+            Assert.Equal(buffer.Length, span.Length);
+            Assert.True(buffer.IsEmpty || buffer.Length != 0);
+
+            var slice = buffer.Slice(1);
+
+            for (int i = 0; i < slice.Length; i++) Assert.Equal(slice.Span[i], span[i+1]);
+
+            var array = buffer.ToArray();
+            for (int i = 0; i < array.Length; i++) Assert.Equal(array[i], span[i]);
+        }
+
+        static void BufferLifetime(ReadOnlyBuffer<byte> buffer)
+        {
+            var array = buffer.ToArray();
+            using(var handle = buffer.Retain())
+            using (var pinned = buffer.Pin())
+            {
+                unsafe
+                {
+                    var p = (byte*)pinned.PinnedPointer;
+                    Assert.True(null != p);
+                    for(int i=0; i<buffer.Length; i++)
+                    {
+                        Assert.Equal(array[i], p[i]);
+                    }
+                }
+            }
         }
     }
 }
--- a/tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
+++ b/tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
@@ -59,11 +59,13 @@ namespace System.Buffers.Tests
                 Assert.Equal(span[i + 5], slice[i]);
             }
 
-            Assert.Throws<ArgumentOutOfRangeException>(() => {
+            Assert.Throws<ArgumentOutOfRangeException>(() =>
+            {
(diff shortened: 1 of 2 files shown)


tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -237,5 +243,37 @@ static void BufferLifetime(Buffer<byte> buffer)
             //{
             //}
         }
+
+        static void BufferBasics(ReadOnlyBuffer<byte> buffer)
+        {
+            var span = buffer.Span;
+            Assert.Equal(buffer.Length, span.Length);
+            Assert.True(buffer.IsEmpty || buffer.Length != 0);
```
ahsonkhan: Why do we need this check?

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -237,5 +243,37 @@ static void BufferLifetime(Buffer<byte> buffer)
...
+
+            var slice = buffer.Slice(1);
```
ahsonkhan: if buffer.IsEmpty is true, won't this result in a buffer of length -1?

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -237,5 +243,37 @@ static void BufferLifetime(Buffer<byte> buffer)
...
+        {
+            var array = buffer.ToArray();
+            using(var handle = buffer.Retain())
+            using (var pinned = buffer.Pin())
+            {
+                unsafe
+                {
+                    var p = (byte*)pinned.PinnedPointer;
+                    Assert.True(null != p);
+                    for(int i=0; i<buffer.Length; i++)
```
ahsonkhan: nit: spacing.

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -237,5 +243,37 @@ static void BufferLifetime(Buffer<byte> buffer)
...
+            for (int i = 0; i < slice.Length; i++) Assert.Equal(slice.Span[i], span[i+1]);
+
+            var array = buffer.ToArray();
+            for (int i = 0; i < array.Length; i++) Assert.Equal(array[i], span[i]);
+        }
+
+        static void BufferLifetime(ReadOnlyBuffer<byte> buffer)
+        {
+            var array = buffer.ToArray();
+            using(var handle = buffer.Retain())
```
ahsonkhan: nit: spacing.

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
KrzysztofCwalina: But the test can run on zero length buffers. This test ensures that is buffer is not empty, then Length !=0

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
KrzysztofCwalina: good point. I wonder why the tests don't fail. I will investigate.

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
KrzysztofCwalina: will fix

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
KrzysztofCwalina: will fix

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
ahsonkhan: Why are we checking both buffer.IsEmpty and buffer.Length?
IsEmpty already does the check if Length == 0.
```C#
public bool IsEmpty => Length == 0;
```

https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System/Buffers/Buffer.cs#L61

> This test ensures that is buffer is not empty, then Length !=0

How does this test ensure that?

Assert.True(buffer.IsEmpty || X) => passes regardless of what X returns. If we have buffer.Length != 0, buffer.Empty should be false, No? Am I missing something here?

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -202,35 +210,39 @@ static void BufferBasics(Buffer<byte> buffer)
             var span = buffer.Span;
             Assert.Equal(buffer.Length, span.Length);
             Assert.True(buffer.IsEmpty || buffer.Length != 0);
+            Assert.True(!buffer.IsEmpty || buffer.Length == 0);
```
ahsonkhan: Is the following the intention of the two Asserts?
```C#
if (buffer.Empty)
{
    Assert.Equal(0, buffer.Length);
}
else
{
    Assert.True(buffer.Length != 0)
}
```

PR #1574, opened 2017-05-25
Adding ITransformation and changing Base64 to implement it

ahsonkhan:
Ready for review.
- perf tests
- adding ITransformation
- adding test helper and minor bug fixes

**Edit:** See below for updated results that are more accurate, along with conclusions.

### DIFF
--- a/src/System.Binary.Base64/System/Binary/Base64.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64.cs
@@ -375,6 +375,8 @@ namespace System.Binary.Base64
             int sourceIndex = 0;
             int destIndex = 0;
 
+            if (source.Length == 0) goto TrueExit;
+
             int result = 0;
 
             while (sourceIndex < srcLength - 4)
@@ -426,6 +428,7 @@ namespace System.Binary.Base64
 
             if (srcLength != source.Length) goto FalseExit;
 
+            TrueExit:
             bytesConsumed = sourceIndex;
             bytesWritten = destIndex;
             return true;
--- a/src/System.Binary.Base64/System/Binary/Base64.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64.cs
@@ -5,7 +5,7 @@ namespace System.Binary.Base64
 {
     public sealed class Base64
     {
-        public static readonly Base64Encoder Encoder = Base64Encoder.Encoder;
-        public static readonly Base64Decoder Decoder = Base64Decoder.Decoder;
+        public static readonly Base64Encoder Encoder = Base64Encoder.Instance;
+        public static readonly Base64Decoder Decoder = Base64Decoder.Instance;
     }
 }
--- a/src/System.Binary.Base64/System/Binary/Base64.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64.cs
@@ -473,5 +473,10 @@ namespace System.Binary.Base64
         }
 
         #endregion
+
+        public enum ReturnState
+        {
+            SUCCESS, OUTPUT_TOO_SMALL, NEED_MORE_INPUT, INPUT_INVALID
+        }
     }
 }
--- a/src/System.Binary.Base64/System/Binary/Base64.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64.cs
@@ -1,0 +1,11 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Binary.Base64
+{
+    public sealed class Base64
+    {
+        public static readonly Base64Encoder Encoder = Base64Encoder.Encoder;
+        public static readonly Base64Decoder Decoder = Base64Decoder.Decoder;
+    }
+}
--- a/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
@@ -8,13 +8,13 @@ namespace System.Binary.Base64
 {
     public sealed class Base64Decoder : ITransformation
     {
-        private static readonly Base64Decoder s_decoder = new Base64Decoder();
+        private static readonly Base64Decoder s_instance = new Base64Decoder();
 
-        public static Base64Decoder Decoder
+        public static Base64Decoder Instance
         {
             get
             {
-                return s_decoder;
+                return s_instance;
             }
         }
 
--- a/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
@@ -6,8 +6,20 @@ using System.Runtime.CompilerServices;
 
 namespace System.Binary.Base64
 {
-    public class Base64Decoder : ITransformation
+    public sealed class Base64Decoder : ITransformation
     {
+        private static readonly Base64Decoder s_decoder = new Base64Decoder();
+
+        public static Base64Decoder Decoder
+        {
+            get
+            {
+                return s_decoder;
+            }
+        }
+
+        private Base64Decoder() { }
+
         // Pre-computing this table using a custom string(s_characters) and GenerateEncodingMapAndVerify (found in tests)
         static readonly byte[] s_encodingMap = {
             65, 66, 67, 68, 69, 70, 71, 72,         //A..H
--- a/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
@@ -1,11 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Buffers;
 using System.Runtime.CompilerServices;
 
 namespace System.Binary.Base64
 {
-    public static class Base64Encoder
+    public class Base64Decoder : ITransformation
     {
         // Pre-computing this table using a custom string(s_characters) and GenerateEncodingMapAndVerify (found in tests)
         static readonly byte[] s_encodingMap = {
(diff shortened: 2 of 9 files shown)


KrzysztofCwalina:
How do you interpret the data?

Also, the "no three calls" test should use APIs what take two input buffers.

ahsonkhan:
> How do you interpret the data?

Once I am confident in the correctness of the perf test, I will add some conclusions/insights.

> Also, the "no three calls" test should use APIs what take two input buffers.

I will add that case as well.

ahsonkhan:
**Data:**
image
**Graph:**
image
**Conclusion:**
Two TryDecode Calls (where you copy the entire second buffer + leftover fully into a stack alloc'd span) is ~2.5x faster than three TryDecode Calls (where you use a large stack alloc'd span to stitch the two buffers).
Surprisingly, Two TryDecode Calls (where you copy the entire second buffer + leftover fully into a stack alloc'd span) is ~2.5x faster than even having two TryDecode Calls where you don't use need to stitch at all and there is no stack allo'd span.

**Data:**
image
**Graph:**
image
**Conclusion:**
The stack size chosen for stitching the leftover bytes with the next buffer (between 10 and 500 bytes and even 60% of the input buffer) has no impact on performance of decoding for input buffers between 1k and 50k bytes.

**Data:**
image
**Graph:**
image
**Conclusion:**
As expected, the larger the input buffer, the longer it takes to decode. However, we get slightly faster (per byte), as we process larger buffers.

KrzysztofCwalina:
> Two TryDecode Calls (where you copy the entire second buffer + leftover fully into a stack alloc'd span) is ~2.5x faster than even having two TryDecode Calls where you don't use need to stitch at all and there is no stack allo'd span.

Are you sure your tests are correct? This seems very strange.

ahsonkhan:
> Are you sure your tests are correct? This seems very strange.

Yes, I caught and fixed the bug and the results make more sense now. I have updated the results.

tests/System.Binary.Base64.Tests/PerformanceTests.cs
```diff
@@ -96,5 +119,176 @@ private static void Base64DecodeBaseline(int numberOfBytes)
...
+
+            foreach (var iteration in Benchmark.Iterations)
+            {
+                using (iteration.StartMeasurement())
+                {
+                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                    {
+                        if (!Base64Encoder.TryDecode(source1, destination, out int bytesConsumed, out int bytesWritten))
+                        {
+                            // this shouldn't happen!
```
KrzysztofCwalina: Should you add Assert.True(false); here?

tests/System.Binary.Base64.Tests/PerformanceTests.cs
```diff
@@ -96,5 +119,176 @@ private static void Base64DecodeBaseline(int numberOfBytes)
...
+                    for (int i = 0; i < Benchmark.InnerIterationCount; i++)
+                    {
+                        if (!Base64Encoder.TryDecode(source1, destination, out int bytesConsumed, out int bytesWritten))
+                        {
+                            int leftOverBytes = source1.Length - bytesConsumed;
+                            if (leftOverBytes < 4)
+                            {
+                                source1.Slice(bytesConsumed).CopyTo(stackSpan);
+                                int amountToCopy = Math.Min(source2.Length, stackSpan.Length - leftOverBytes);
+                                source2.Slice(0, amountToCopy).CopyTo(stackSpan.Slice(leftOverBytes));
```
KrzysztofCwalina: Since you slice source2 here, where does the program decode the sliced off tail of source2?

src/System.Binary.Base64/System/Binary/Base64.cs
```diff
@@ -473,5 +473,10 @@ public static int DecodeInPlace(Span<byte> buffer)
         }
 
         #endregion
+
+        public enum ReturnState
+        {
+            SUCCESS, OUTPUT_TOO_SMALL, NEED_MORE_INPUT, INPUT_INVALID
```
KrzysztofCwalina: Enum values use PascalCasing.

src/System.Binary.Base64/System/Binary/Base64.cs
```diff
@@ -473,5 +473,10 @@ public static int DecodeInPlace(Span<byte> buffer)
         }
 
         #endregion
+
+        public enum ReturnState
```
KrzysztofCwalina: Why do you add this enum here without changing the APIs to return it?

tests/System.Binary.Base64.Tests/BasicUnitTests.cs
```diff
@@ -186,6 +304,256 @@ public void DecodingOutputTooSmall()
...
+                numBytes = rnd.Next(100, 251);
+                decodedBytes = new byte[numBytes];
+            } while (!success);
+
+            Assert.Equal(source.Length, bytesConsumed);
+            Assert.Equal(Base64Encoder.ComputeDecodedLength(source), bytesWritten);
+        }
+
+        [Fact]
+        public void SampleUsage3()
```
KrzysztofCwalina: I find it hard to follow the tests. There are several of them and I am not sure what they test. I think it would be better to have:
1. One test that tests that a single buffer is encoded properly 
2. One test that tests that a single buffer is decoded properly
3. One tests for various error conditions (if any)
4. One tests that implements the piping loop with InlineData for various combinations of buffers.

src/System.Binary.Base64/System/Binary/Base64.cs
ahsonkhan: This is a WIP and I do not intend to merge this PR. My intention was to just share the perf tests and data. There is a lot of leftover code here where I was trying things (which doesn't need to be reviewed).

src/System.Binary.Base64/System/Binary/Base64.cs
ahsonkhan: I will fix when I update the APIs and submit an actual PR.

tests/System.Binary.Base64.Tests/BasicUnitTests.cs
ahsonkhan: > Also, i don't think you need to unit tests all the stitching strategies. Just the one we currently think is the best one. The perf tests should validate after they run that the full data set was properly encoded/decoded.

Of course. I will add the validation within the perf test itself once we settle on what we want.

> One tests that implements the piping loop with InlineData for various combinations of buffers. Ideally the pipeline loop would be a reusable routine that can be called from both unit tests and performance tests.

Agreed. The SampleUsage tests will be removed and can be ignored for this PR/discussion.

tests/System.Binary.Base64.Tests/PerformanceTests.cs
ahsonkhan: The intent was to just measure the cost of stitching compared to a baseline. I will have better perf tests that measure the full piping loop.

tests/System.Binary.Base64.Tests/BasicUnitTests.cs
ahsonkhan: > I find it hard to follow the tests. There are several of them and I am not sure what they test. 

Yes, there is unnecessary stuff in this PR that can be ignored. The currently checked in tests are cleaner and BasicEncoding / BasicDecoding do cover scenario 1/2 that you mentioned.

ahsonkhan:
The Span Copy is only around 1% of the execution time of the Decode which is why it has no noticeable impact on the performance.

image

image
image

ahsonkhan:
@KrzysztofCwalina, please review.

I added ITransformation and split Base64 into encoder and decoder.

ahsonkhan:
**Changing signature from:**
```C#
public static bool TryDecode(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
```
**To:**
```C#
public TransformationStatus Transform(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
```

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -1,11 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Buffers;
 using System.Runtime.CompilerServices;
 
 namespace System.Binary.Base64
 {
-    public static class Base64Encoder
+    public class Base64Decoder : ITransformation
```
KrzysztofCwalina: this needs to be a singleton. Make the default ctor private, create a static singleton instance, and expose the instance as a static property. See https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System/Buffers/BufferPool.cs#L8

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+using System.Collections.Sequences;
+
+// This should most likely be in System.Text.Primitives and not here. Will move once Base64 moves.
+namespace System.Binary.Base64
+{
+    public static class BufferExtensions
+    {
+        const int stackLength = 32;
+
+        public static void Pipe(this ITransformation transformation, ReadOnlyBytes source, IOutput destination)
```
KrzysztofCwalina: I think IOutput and ROB are fine for now. But we need to close on whether these are the primitives we want.
ROB needs to be unified with ReadableBuffer
IOutput: we need to validate that it works for chaining transformers.

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+        {
+            int afterMergeSlice = 0;
+            ReadOnlySpan<byte> remainder;
+            Span<byte> stackSpan;
+
+            unsafe
+            {
+                byte* stackBytes = stackalloc byte[stackLength];
```
KrzysztofCwalina: Do we have to do this here? What if the stack buffer does not endup being needed?

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
```diff
@@ -0,0 +1,116 @@
...
+                stackSpan = new Span<byte>(stackBytes, stackLength);
+            }
+
+            var poisition = Position.First;
+            while (source.TryGet(ref poisition, out var sourceBuffer, true))
```
KrzysztofCwalina: You might want to check if it's not worth optimizing this for a single segment ROB. i.e. if ROB has one segment, there is no need to TryGet and instead just access the First property.

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
ahsonkhan: If we do it in the loop, don't we open up the potential for stack overflow?

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -0,0 +1,331 @@
...
+using System.Buffers;
+using System.Runtime.CompilerServices;
+
+namespace System.Binary.Base64
+{
+    public sealed class Base64Decoder : ITransformation
+    {
+        private static readonly Base64Decoder s_decoder = new Base64Decoder();
+
+        public static Base64Decoder Decoder
```
shiftylogic: Call this "Instance"? That is the typical pattern for a singleton.

src/System.Binary.Base64/System/Binary/Base64Encoder.cs
```diff
@@ -0,0 +1,202 @@
...
+using System.Buffers;
+using System.Runtime.CompilerServices;
+
+namespace System.Binary.Base64
+{
+    public sealed class Base64Encoder : ITransformation
+    {
+        private static readonly Base64Encoder s_encoder = new Base64Encoder();
+
+        public static Base64Encoder Encoder
```
shiftylogic: Instance?

ahsonkhan: OK.

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
ahsonkhan: Can that be addressed outside this PR?

I will file an issue and work on this investigation and optimization a bit later.

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
ahsonkhan: > Unless, if we add a check like if stackSpan.IsEmpty, then we do the allocation.

Can that be addressed outside this PR? I will investigate if it is worth moving it inside the `if (!remainder.IsEmpty)` block within the loop.

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
ahsonkhan: Issue #1584

src/System.Binary.Base64/System/Binary/BufferExtensions.cs
ahsonkhan: Issue #1585

PR #1581, opened 2017-05-31
Native Brotli Libraries for Windows

shiftylogic:


shiftylogic:
@KrzysztofCwalina @joshfree @ahsonkhan

shiftylogic:

KrzysztofCwalina:
looks good.

.gitignore
```diff
@@ -264,3 +264,5 @@ BenchmarkDotNet.Artifacts/
 BDN.Generated/
 binaries/
 
+# Do not exclude any external files
+!external/**/*
```
ahsonkhan: Is this required to override the gitignore of .dll/etc?
https://github.com/dotnet/corefxlab/blob/master/.gitignore#L10
https://github.com/dotnet/corefxlab/blob/master/.gitignore#L51

Do we want to keep this exclusion outside of this single change?

shiftylogic: Yes, it is required and will be required each time we update the binaries. I intentionally put the binaries in an external folder so that the exclusion was contained.

PR #1583, opened 2017-05-31
Propagate channel exception from {WaitFor}Read/WriteAsync methods

stephentoub:
cc: @davidfowl, @moozzyk, @anurse 

Discussed in https://github.com/aspnet/SignalR/pull/496

### DIFF
--- a/src/System.Threading.Tasks.Channels/System/Threading/Tasks/Channels/BoundedChannel.cs
+++ b/src/System.Threading.Tasks.Channels/System/Threading/Tasks/Channels/BoundedChannel.cs
@@ -142,10 +142,10 @@ namespace System.Threading.Tasks.Channels
             // We also know that only one thread (this one) will ever get here, as only that thread
             // will be the one to transition from _doneWriting false to true.  As such, we can
             // freely manipulate _blockedReaders and _waitingReaders without any concurrency concerns.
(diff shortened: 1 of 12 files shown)


davidfowl:
@stephentoub The Wait* methods never throw `ClosedChannelException` right?

stephentoub:
They do with this change, wrapping the original. I could change that, but it a) seems like you want to indicate that the failure is because it was closed and then provide the details, and b) seemed odd to differ from Read/WriteAsync in whether the exception was wrapped. Don't feel strongly about it, though. I can see arguments either way, so if you'd prefer non-wrapped, I can change it.

davidfowl:
I'd prefer if it didn't wrap the exception.

anurse:
To be clear, given a normal `.TryComplete()` call (with no exception), the matching `.WaitToReadAsync()` calls will still **not** throw, right? It looks like that's still the case, just confirming.

Given that, I don't have a problem with this. I was just awaiting `.Completion` after my loops generally to avoid this, but it does seem a little counter-intuitive.

stephentoub:
> given a normal .TryComplete() call (with no exception), the matching .WaitToReadAsync() calls will still not throw, right?

Correct

ahsonkhan:
@davidfowl, @stephentoub, is this change good to go?

stephentoub:
> is this change good to go?

No, I haven't re-updated it in response to the feedback.

Issue #1580, opened 2017-05-31
Base64Encoder.TryDecode can be much faster


jkotas:
The current Base64.Base64Encoder needs ~18 instructions to decode a single byte. The path length of this loop is 73 instructions:

```
            while (sourceIndex < srcLength - 4)
            {
                result = Decode(ref Unsafe.Add(ref srcBytes, sourceIndex));
                if (result == -1) throw new FormatException();  // invalid bytes
                if (destIndex > destLength - 3) goto FalseExit;
                Unsafe.WriteUnaligned(ref Unsafe.Add(ref destBytes, destIndex), result);
                destIndex += 3;
                sourceIndex += 4;
            }
```

There is a lot of overhead  in the piece meal split into a bunch of little functions with a ton of arguments. I think you should be able to get under 10 instructions to decode a single byte by streamlining the implementation (about 2x better than the current state), and proportional improvement in throughput.

ahsonkhan:
>  proportional improvement in throughput

I did not see any perf improvement when I made this change initially, and hence I left it as is. I will try again without the helper functions and see if we get a performance improvement.

ahsonkhan:
> There is a lot of overhead in the piece meal split into a bunch of little functions with a ton of arguments.

I removed the use of the functions and in-lined the code:
```C#
while (sourceIndex < srcLength - 4)
{
    byte b0 = Unsafe.Add(ref srcBytes, sourceIndex);
    byte b1 = Unsafe.Add(ref srcBytes, sourceIndex + 1);
    byte b2 = Unsafe.Add(ref srcBytes, sourceIndex + 2);
    byte b3 = Unsafe.Add(ref srcBytes, sourceIndex + 3);

    if (b0 == s_encodingPad || b1 == s_encodingPad || b2 == s_encodingPad || b3 == s_encodingPad) goto InvalidExit;

    byte i0 = s_decodingMap[b0];
    byte i1 = s_decodingMap[b1];
    byte i2 = s_decodingMap[b2];
    byte i3 = s_decodingMap[b3];

    if (i0 == s_invalidByte || i1 == s_invalidByte || i2 == s_invalidByte || i3 == s_invalidByte) goto InvalidExit;

    byte r0 = (byte)(i0 << 2 | i1 >> 4);
    byte r1 = (byte)(i1 << 4 | i2 >> 2);
    byte r2 = (byte)(i2 << 6 | i3);

    result = r2 << 16 | r1 << 8 | r0;

[...]

Issue #1579, opened 2017-05-31
Buffer overrun in Base64Encoder.TryDecode


jkotas:
This line https://github.com/dotnet/corefxlab/blob/604ebd8ae3ad303f44317673f3e5a7ca2c39ed07/src/System.Binary.Base64/System/Binary/Base64.cs#L385 writes 4 bytes, but the previous line only ensures 3 bytes are available in the buffer.

```
    while (sourceIndex < srcLength - 4)
    {
        result = Decode(ref Unsafe.Add(ref srcBytes, sourceIndex));
        if (result == -1) throw new FormatException();  // invalid bytes
        if (destIndex > destLength - 3) goto FalseExit;
        Unsafe.WriteUnaligned(ref Unsafe.Add(ref destBytes, destIndex), result); <--- bug
        destIndex += 3;
        sourceIndex += 4;
    }
```

ahsonkhan:
> but the previous line only ensures 3 bytes are available in the buffer.

Base 64 Decoding goes from 4 bytes to 3 bytes. So the check is correct.

Does WriteUnaligned write 4 bytes because we are writing an int, which is 4 bytes? Therefore, should we just write one byte at a time instead since we only need to write 3 bytes?

jkotas:
You either need to write one byte at a time (or two bytes + one byte). Or you need to ensure that you use the fast loop only when there is spare byte available in the output buffer - this is likely faster (related to the other issue that I have just opened).

ahsonkhan:
> Or you need to ensure that you use the fast loop only when there is spare byte available in the output buffer

Is writing 3 bytes, one byte at a time, slower than writing 4 bytes (all at once, as an int)?

> You either need to write one byte at a time (or two bytes + one byte). 

I wonder if this gets us any perf wins. And like you said, writing 4 bytes at a time instead might be faster.

jkotas:
> Is writing 3 bytes, one byte at a time, slower than writing 4 bytes (all at once, as an int)?

Measure it...

ahsonkhan:
Fixed in https://github.com/dotnet/corefxlab/pull/1586.

PR #1586, opened 2017-06-01
Optimizing Base64Decoder.Transform using an sbyte decoding map

ahsonkhan:
Fixes https://github.com/dotnet/corefxlab/issues/1580 & https://github.com/dotnet/corefxlab/issues/1579

cc @KrzysztofCwalina. @jkotas, @shiftylogic 

Ignore all commits, except for the last (most of these are part of https://github.com/dotnet/corefxlab/pull/1574/):
**This is the commit that matters** [Optimizing Base64Decoder.Transform using an sbyte decoding map](https://github.com/ahsonkhan/corefxlab/commit/8f5ca1f2c865071d632a76656fe3d9bf6ae96d6a).

image

**Here are the key optimizations:**
- Changed decoding map to sbyte which enables negative values for invalid bytes instead of positive bytes that are out of range (like 255).
   - This allows us to remove several invalid byte checks on individual bytes. Instead, once we fetch all the bytes and do the appropriate 'shift' and 'or' operations, the error is maintained. If any of the four bytes map to a negative value, the 'or' operation (or bit shift) will maintain the negative sign and hence the final result will be negative. Hence, this cuts down on a lot of unnecessary checks per byte an enables us to replace those checks with a single check for every 4 bytes.
- Fetching the bytes into ints rather than bytes.
   - We reduced the number of bit shift operators because of this. Instead of first shifting the individual bytes into 6 bits and then combining them into the first 24 bits of an int, we shift the bytes in place within the int. As an aside, I also re-used the variables, so we use 4 ints (input and output) rather than 7 bytes (4 input bytes, 3 output bytes)+ 1 int (final result).
- Using Unsafe.Add to index the decoding map instead of array indexer.
   - Array indexer has range check. This one does not.
- Treat the padding character as invalid within the decoding map
   - This allows us to treat is as invalid for most of the decoding until the last 4 bytes, which we already special case anyway. This way, the main loop becomes much faster. The only time the padding character becomes valid is within the final 2 bytes of the last 4 bytes of the input.
- Removing some duplicate code/effort from PR review
   - Mainly removing the redundant work to check the padding count first and then using it later. The same conditions were used directly.


ahsonkhan:
This diff is easier to see and review:
https://github.com/ahsonkhan/corefxlab/commit/8f5ca1f2c865071d632a76656fe3d9bf6ae96d6a

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -0,0 +1,266 @@
...
+            int destLength = destination.Length;
+
+            int sourceIndex = 0;
+            int destIndex = 0;
+
+            if (source.Length == 0) goto DoneExit;
+
+            ref sbyte decodingMap = ref s_decodingMap[0];
+
+            while (sourceIndex < srcLength - 4) // TODO: you can make this more efficient by using unsafe pointer as the loop limit
```
jkotas: Can you fix these TODOs while you are on it?

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -0,0 +1,266 @@
...
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,         // Hence, padding the map with 255, which indicates invalid input
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+            -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
+        };
+
+        static readonly byte s_encodingPad = (byte)'=';              // '=', for padding
```
jkotas: This should be const.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -0,0 +1,266 @@
...
+                WriteThreeBytes(ref Unsafe.Add(ref destBytes, destIndex), i0);
+                destIndex += 3;
+                sourceIndex += 4;
+            }
+
+            if (sourceIndex >= srcLength) goto NeedMoreExit;
+
+            int padding = 0;
+
+            if (Unsafe.Add(ref srcBytes, srcLength - 1) == s_encodingPad)
```
jkotas: It is not necessary to compute the padding first and the branch based on it. You can directly branch directly on the values:

```
if (Unsafe.Add(ref srcBytes, srcLength - 1) != s_encodingPad)
{
    ... no pad case ...
}
else if (Unsafe.Add(ref srcBytes, srcLength - 2) != s_encodingPad)
{
    ... one pad case ...
}
else
{
    ... two pad case ...
}
```

src/System.Binary.Base64/System/Binary/Base64Encoder.cs
```diff
@@ -0,0 +1,201 @@
...
+
+            Encode(b0, b1, b2, out byte r0, out byte r1, out byte r2, out byte r3);
+
+            int result = r3 << 24 | r2 << 16 | r1 << 8 | r0;
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static void Encode(byte b0, byte b1, byte b2, out byte r0, out byte r1, out byte r2)
+        {
```
jkotas: This can use same set of optimizations as the decoder.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
jkotas: (Or delete them)

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -0,0 +1,266 @@
...
+                sourceIndex += 4;
+            }
+            else
+            {
+                int i0 = Decode(ref Unsafe.Add(ref srcBytes, sourceIndex), ref decodingMap);
+                if (i0 < 0) goto InvalidExit;
+                if (destIndex > destLength - 3) goto DestinationSmallExit;
+                WriteThreeBytes(ref Unsafe.Add(ref destBytes, destIndex), i0);
+                destIndex += 3;
+                sourceIndex += 4;
```
jkotas: `sourceIndex += 4;` is same in every if case, so you can move it after the if chain to save a bit code size.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
ahsonkhan: > Can you fix these TODOs while you are on it?

I will tackle them tomorrow.

src/System.Binary.Base64/System/Binary/Base64Encoder.cs
ahsonkhan: Yes, of course. Will tackle that outside of this PR.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
ahsonkhan: > `sourceIndex += 4;` is same in every if case, so you can move it after the if chain to save a bit code size.

Will fix.

I think I accidentally deleted your comment.

ahsonkhan: .

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -0,0 +1,266 @@
...
+            int padding = 0;
+
+            if (Unsafe.Add(ref srcBytes, srcLength - 1) == s_encodingPad)
+            {
+                padding = (Unsafe.Add(ref srcBytes, srcLength - 2) == s_encodingPad) ? 2 : 1;
+            }
+
+            if (padding == 1)
+            {
+                int i0 = DecodePadByOne(ref Unsafe.Add(ref srcBytes, sourceIndex), ref decodingMap);
```
jkotas: Note that each of these if-else cases will end up having a lot of common code - each of them will fetch all 4 bytes (the last byte will be likely fetched twice - first for the comparison, and second time to do the actual work), map and shift the first two at least, and or them together. Lean and mean implementation would avoid this duplicated code.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -0,0 +1,266 @@
...
+            Unsafe.Add(ref destBytes, 1) = (byte)(i0 >> 8);
+            Unsafe.Add(ref destBytes, 2) = (byte)i0;
+        }
+
+        public TransformationStatus Transform(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
+        {
+            ref byte srcBytes = ref source.DangerousGetPinnableReference();
+            ref byte destBytes = ref destination.DangerousGetPinnableReference();
+
+            int srcLength = source.Length / 4 * 4;  // only decode input up to the closest multiple of 4.
```
Tornhoof: `` /4 * 4``  is the same as ``& ~0x3``. It might be slightly faster because you don't idiv

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
ahsonkhan: > you can make this more efficient by using unsafe pointer as the loop limit

Can you show me a code sample on how to do that?

> you can get rid off this line by including it in loop limit

If I move this check into the loop limit, won't I have to add an extra if check after the loop to see if we need to go to DestinationSmallExit?
`if (destIndex > destLength - 3) goto DestinationSmallExit;`

jkotas: E.g. it is done in the UTF8 encoders/decoders: https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf8Encoder.cs#L713

>  I have to add an extra if check after the loop to see if we need to go to DestinationSmallExit?

Correct.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -294,9 +211,10 @@ public TransformationStatus Transform(ReadOnlySpan<byte> source, Span<byte> dest
         /// <returns>Number of bytes written to the buffer.</returns>
         public static int DecodeInPlace(Span<byte> buffer)
         {
+            // TODO: Fix
```
KrzysztofCwalina: What is this fix?

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -294,9 +211,10 @@ public TransformationStatus Transform(ReadOnlySpan<byte> source, Span<byte> dest
...
             int di = 0;
             int si = 0;
-            byte r0, r1, r2;
+            //byte r0, r1, r2;
```
KrzysztofCwalina: Why is that commented out?

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -307,21 +225,21 @@ public static int DecodeInPlace(Span<byte> buffer)
 
             for (; si < buffer.Length - (padding != 0 ? 4 : 0);)
             {
-                Decode(buffer[si++], buffer[si++], buffer[si++], buffer[si++], out r0, out r1, out r2);
+                /*Decode(buffer[si++], buffer[si++], buffer[si++], buffer[si++], out r0, out r1, out r2);
```
KrzysztofCwalina: Why is that commented out?

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -307,21 +225,21 @@ public static int DecodeInPlace(Span<byte> buffer)
...
                 buffer[di++] = r0;
                 buffer[di++] = r1;
-                buffer[di++] = r2;
+                buffer[di++] = r2;*/
             }
 
             if (padding != 0)
             {
-                Decode(buffer[si++], buffer[si++], buffer[si++], buffer[si++], out r0, out r1, out r2);
+                /*Decode(buffer[si++], buffer[si++], buffer[si++], buffer[si++], out r0, out r1, out r2);
```
KrzysztofCwalina: Same?

KrzysztofCwalina:
Looks good. There are some nits about commented out code. Feel free to merge and then fix in a subsequent PR.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
ahsonkhan: I removed the following API which DecodeInPlace relied on:
```C#
private static bool Decode(byte b0, byte b1, byte b2, byte b3, out byte r0, out byte r1, out byte r2)
```

I essentially have to make similar changes that I made to the implementation of the Transform API to the DecodeInPlace, which I will do in subsequent PR.

Also, since I changed the decoding map, it means the implementation needs to be updated.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
ahsonkhan: The Decode API that takes individual bytes is gone. I will re-write this method in future PRs which works with the new decoding map.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -82,188 +67,120 @@ public static int ComputeDecodedLength(ReadOnlySpan<byte> source)
...
+            int i1 = Unsafe.Add(ref srcBytes, srcLength - 3);
+            int i2 = Unsafe.Add(ref srcBytes, srcLength - 2);
+            int i3 = Unsafe.Add(ref srcBytes, srcLength - 1);
+
+            i0 = Unsafe.Add(ref decodingMap, i0);
+            i1 = Unsafe.Add(ref decodingMap, i1);
 
-            if (Unsafe.Add(ref srcBytes, srcLength - 1) == s_encodingPad)
+            i0 <<= 18;
+            i1 <<= 12;
```
jkotas: You can move `i0 |= i1` to this common part as well.

ahsonkhan: I tried doing that, but i1 is different for some cases (it needs to get or'd with i2 first).

```C#
i0 |= i3;
i1 |= i2;

i0 |= i1;
```
vs. just
```C#
i0 |= i1;
```

jkotas: The order does not matter. You can or with i1 first, and i2 and i3 second.

ahsonkhan: Ah yes, I have to change `i1 |= i2;` to `i0 |= i2;` as well :)

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
KrzysztofCwalina: It would be great if the TO described what the work is. I want to maximize the bus number ;-)

ahsonkhan: Definitely, and in general, I agree with the principle. However, I will make an effort to re-implement and resolve the TODO ASAP.

Otherwise, I will instead open an issue and detail what is required there. I prefer those over TODO comments within the code anyway.

PR #1588, opened 2017-06-01
Update Windows Brotli libraries to expose a couple functions.

shiftylogic:
Functions are already exposed in Linux and OSX binaries, but were not being exported on Windows.


ahsonkhan:
Is this just rebuilding the binaries and updating them?

KrzysztofCwalina:
looks good.

shiftylogic:
Yes, just new Windows binaries with 2 functions exported that were not before (they are already exported on Linux and OSX).

PR #1590, opened 2017-06-03
Fixing DecodeInPlace and optimizing it.

ahsonkhan:
image

The performance gain mainly comes from the optimizations that were done for the Decoder **(see https://github.com/dotnet/corefxlab/pull/1586#issue-232772292, and https://github.com/dotnet/corefxlab/pull/1555#issue-229548969)**.
- indexing into the span by ref instead of the indexer (for read and write)
- saving operations on the redundant padding calc/branch
- using an optimized Decode helper function that does fewer bit shift operations
- this change is actually even faster than the previous (relatively speaking) since the previous one wouldn't check for invalid input bytes but this one does

**Open question:**
Should we change the signature of DecodeInPlace to follow Transform?
i.e.
```C#
public TransformationStatus DecodeInPlace(Span<byte> buffer, out int bytesConsumed, out int bytesWritten)
```
The way it is defined atm, it returns an int (for the number of bytes written to the buffer), but we don't know error/state information for invalid bytes, etc. We also do not currently know how much was consumed which could be useful to the caller.

cc @KrzysztofCwalina, @jkotas 

As an aside:
I removed the case `[InlineData(1000 * 1000 * 50)]` from the perf tests. Testing the Encode/Decode APIs for such a large span (while inner iteration count is 1k) takes too long and doesn't provide much value/insight.

### DIFF
--- a/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
@@ -207,39 +207,79 @@ namespace System.Binary.Base64
         /// <returns>Number of bytes written to the buffer.</returns>
         public static int DecodeInPlace(Span<byte> buffer)
         {
-            // TODO: Fix
-            int di = 0;
-            int si = 0;
-            //byte r0, r1, r2;
-            int padding = 0;
+            ref byte bufferBytes = ref buffer.DangerousGetPinnableReference();
 
-            if (buffer[buffer.Length - 1] == s_encodingPad)
+            int bufferLength = buffer.Length & ~0x3;  // only decode input up to the closest multiple of 4.
+
+            int sourceIndex = 0;
+            int destIndex = 0;
+
+            if (buffer.Length < 4) return 0;
+
+            ref sbyte decodingMap = ref s_decodingMap[0];
+
+            while (sourceIndex < bufferLength - 4)
             {
-                padding = 1;
-                if (buffer[buffer.Length - 2] == s_encodingPad) padding = 2;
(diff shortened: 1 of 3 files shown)


KrzysztofCwalina:
> indexing into the span by ref instead of the indexer (for read and write)
@russellhadley, how close are we to being able to stop doing this?

KrzysztofCwalina:
looks good

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -207,39 +207,79 @@ public TransformationStatus Transform(ReadOnlySpan<byte> source, Span<byte> dest
...
-            int padding = 0;
+            ref byte bufferBytes = ref buffer.DangerousGetPinnableReference();
 
-            if (buffer[buffer.Length - 1] == s_encodingPad)
+            int bufferLength = buffer.Length & ~0x3;  // only decode input up to the closest multiple of 4.
+
+            int sourceIndex = 0;
+            int destIndex = 0;
+
+            if (buffer.Length < 4) return 0;
```
shiftylogic: Can you use bufferLength here?

ahsonkhan: Hmm... bufferLength is buffer.Length /4 which rounds down. So if buffer.Length is < 4, bufferLength would be 0 and I could just instead change the check to that. However, I will wait until we decide on the return value of the method. We would need to have the state "NeedMoreExit", for buffer lengths between 1 and 3 (inclusive), so I will wait for that before making this change.

PR #1591, opened 2017-06-05
Made BufferPool buffers 'auto-dispose'

KrzysztofCwalina:
I changed BufferPool's OwnedBuffer to auto-dispose after refcount drops to zero.

When a buffer is renter, its ref count is zero, but unless somebody increments it and then decrements it, the buffer won't be disposed. And so the proper usage is:

```c#
var owned = _pool.Rent();
using(var handle = owned.Buffer.Retain()){
    PassToConsumer(owned.Buffer);
} 
...
void PassToConsumer(Buffer<byte> buffer){
    _handle = buffer.Retain(); // this needs to be disposed when PassToConsumer is done with the buffer
   ...
}
```

@davidfowl, @ahsonkhan, @shiftylogic, @joshfree

### DIFF
--- a/src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
+++ b/src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
@@ -22,7 +22,8 @@ namespace System.Buffers.Internal
 
         public override OwnedBuffer<byte> Rent(int minimumBufferSize)
         {
-            return new ArrayPoolBuffer(minimumBufferSize);
+            var buffer = new ArrayPoolBuffer(minimumBufferSize);
+            return buffer;
         }
 
         protected override void Dispose(bool disposing)
@@ -89,7 +90,10 @@ namespace System.Buffers.Internal
             public override void Release()
             {
                 if (!IsRetained) BufferPrimitivesThrowHelper.ThrowInvalidOperationException();
-                Interlocked.Decrement(ref _referenceCount);
+                if(Interlocked.Decrement(ref _referenceCount) <= 0)
+                {
+                    Dispose();
+                }
             }
         }
     }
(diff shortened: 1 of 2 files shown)


src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
```diff
@@ -22,7 +22,8 @@ public static ManagedBufferPool Shared
 
         public override OwnedBuffer<byte> Rent(int minimumBufferSize)
         {
-            return new ArrayPoolBuffer(minimumBufferSize);
+            var buffer = new ArrayPoolBuffer(minimumBufferSize);
+            return buffer;
```
ahsonkhan: Why did you decide to make this change?

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
```diff
@@ -89,7 +90,10 @@ public override void Retain()
             public override void Release()
             {
                 if (!IsRetained) BufferPrimitivesThrowHelper.ThrowInvalidOperationException();
-                Interlocked.Decrement(ref _referenceCount);
+                if(Interlocked.Decrement(ref _referenceCount) <= 0)
```
shiftylogic: If there was ever a case where _referenceCount could be < 0, it means that someone called Release without having previously called Retain. Assume that bug exists, what happens when your dispose method calling the ArrayPool return method twice with the same array reference?

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
KrzysztofCwalina: I was increment the refcount in Rent. Then I removed the increment. I could rollback this change now, but it makes it easier to experiment with refcounting. If you feel like we should roll back, I will do.

src/System.Buffers.Primitives/System/Internal/ManagedBufferPool.cs
KrzysztofCwalina: Yeah, you are right. I was trying to avoid some checks, but in the absence of synchronization around IsRetianed, the optimizations won't work. I will fix.

PR #1507, opened 2017-04-27
Removing all unused/extra using directives.

ahsonkhan:
cc @davidfowl, @KrzysztofCwalina, @shiftylogic

PR #1597, opened 2017-06-07
Fix conditionally compiled Debug.Assert

halter73:


shiftylogic:

ahsonkhan:
Are there other using directives that we may need back because of conditionally compiled code?

**Some were cleaned up here:**
https://github.com/dotnet/corefxlab/pull/1281
https://github.com/dotnet/corefxlab/pull/1507

pakrym:
@halter73 why do we care about master? We consume code from aspnetcore branch and nobody except use merges there. What broke and when?

halter73:
Kestrel.Transport.Abstractions doesn't compile with BLOCK_LEASE_TRACKING. With this change, it does. I didn't look into which branch we clone for our CoreFxLab.Sources package. I can rebase this.

halter73:
Also, if we're worried that the usings might just get removed again, we can instead use `global::System.Diagnostics...` inside of conditionally compiled sections.

halter73:
This is subsumed by #1601

PR #1601, opened 2017-06-09
Fix race in BufferSegment and MemoryPoolBlock

halter73:
- Assert MemoryPoolBlocks don't get finalized in all debug builds

I found this issue when enabling BLOCK_LEASE_TRACKING and running Kestrel's functional test project.

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
@@ -2,9 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using System.Diagnostics;
 using System.Runtime;
-using System.Runtime.CompilerServices;
 using System.Threading;
 
 namespace System.Buffers
@@ -41,7 +39,7 @@ namespace System.Buffers
         {
             get
             {
-                return _referenceCount > 0
+                return Volatile.Read(ref _referenceCount) > 0
                         || (ReferenceCountingSettings.OwnedMemory == ReferenceCountingMethod.ReferenceCounter
                             && ReferenceCounter.HasReference(this));
             }
@@ -54,13 +52,13 @@ namespace System.Buffers
 
         public void Release()
         {
-            if(Interlocked.Decrement(ref _referenceCount) == 0)
+            if (Interlocked.Decrement(ref _referenceCount) == 0)
                 OnZeroReferences();
         }
 
         protected virtual void OnZeroReferences()
         { }
-        
+
         public virtual BufferHandle Pin(int index = 0)
         {
             return BufferHandle.Create(this, index);
--- a/src/System.IO.Pipelines/BufferSegment.cs
+++ b/src/System.IO.Pipelines/BufferSegment.cs
@@ -91,11 +91,6 @@ namespace System.IO.Pipelines
             Debug.Assert(_owned.HasOutstandingReferences);
 
             _owned.Release();
-
-            if (!_owned.HasOutstandingReferences)
-            {
-                _owned.Dispose();
-            }
         }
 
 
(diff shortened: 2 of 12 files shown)


src/System.IO.Pipelines/BufferSegment.cs
```diff
@@ -91,11 +91,6 @@ public void Dispose()
             Debug.Assert(_owned.HasOutstandingReferences);
 
             _owned.Release();
-
-            if (!_owned.HasOutstandingReferences)
```
halter73: Here is the race. If BufferHandle.Free() releases the last reference to the _owned buffer, _owned would never get disposed. This is fixed by always returning MemoryPoolBlocks to the pool in OnZeroReferences.

src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -71,17 +77,11 @@ public override Span<byte> Span
             MemoryPool pool,
             MemoryPoolSlab slab)
         {
-            return new MemoryPoolBlock(pool, slab, offset, length)
-            {
-#if BLOCK_LEASE_TRACKING
-                Leaser = Environment.StackTrace,
```
halter73: This is always set by MemoryPool anyway.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
```diff
@@ -41,7 +39,7 @@ public bool HasOutstandingReferences
         {
             get
             {
-                return _referenceCount > 0
+                return Volatile.Read(ref _referenceCount) > 0
```
pakrym: Is this required?

src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -55,11 +56,16 @@ public override Span<byte> Span
 
         ~MemoryPoolBlock()
         {
-#if BLOCK_LEASE_TRACKING
-            Debug.Assert(Slab == null || !Slab.IsActive, $"{Environment.NewLine}{Environment.NewLine}*** Block being garbage collected instead of returned to pool: {Leaser} ***{Environment.NewLine}");
-#endif
             if (Slab != null && Slab.IsActive)
             {
+#if DEBUG
+                Debug.Assert(false, $"{Environment.NewLine}{Environment.NewLine}*** Block being garbage collected instead of returned to pool" +
```
pakrym: 👍

src/System.IO.Pipelines/BufferSegment.cs
davidfowl: /cc @KrzysztofCwalina @msftdata This was causing a memory leak in kestrel.

src/System.Buffers.Primitives/System/Buffers/OwnedBuffer.cs
halter73: I'm not sure, but I added this because I'm worried about the Debug.Assert(_owned.HasOutstandingReferences) in BufferSegment.Dispose().

Isn't possible that without this volatile read, a BufferSegment could be created and then disposed on a different thread that sees a cashed value of _owned._referenceCount from before the BufferSegment called _owned.AddReference() in the ctor?

src/System.IO.Pipelines/System.IO.Pipelines.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>An abstraction for doing efficient asynchronous IO</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
halter73: This is to support Environment.StackTrace when BLOCK_LEASE_TRACKING is defined.

PR #1602, opened 2017-06-09
Adding Buffer<T> constructors that takes an array

ahsonkhan:
- Making `internal Buffer(T[] array, int index, int length)` public
- Adding constructors (similar to Span):
   - public Buffer(T[] array)
   - public Buffer(T[] array, int start)
   - public Buffer(T[] array, int start, int length)
- Adding tests for all scenarios.
- Same with ReadOnlyBuffer\<T\>

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/Buffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Buffer.cs
@@ -26,36 +26,17 @@ namespace System.Buffers
             _length = length;
         }
 
-        public Buffer(T[] array)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public Buffer(T[] array) : this (array, 0, array.Length)
         {
-            if (array == null)
-                BufferPrimitivesThrowHelper.ThrowArgumentNullException(ExceptionArgument.array);
-            if (default(T) == null && array.GetType() != typeof(T[]))
-                BufferPrimitivesThrowHelper.ThrowArrayTypeMismatchException(typeof(T));
-
-            _array = array;
-            _owner = null;
-            _index = 0;
-            _length = array.Length;
         }
 
-        public Buffer(T[] array, int start)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public Buffer(T[] array, int start) : this(array, start, array.Length - start)
         {
-            if (array == null)
-                BufferPrimitivesThrowHelper.ThrowArgumentNullException(ExceptionArgument.array);
-            if (default(T) == null && array.GetType() != typeof(T[]))
-                BufferPrimitivesThrowHelper.ThrowArrayTypeMismatchException(typeof(T));
-
-            int arrayLength = array.Length;
-            if ((uint)start > (uint)arrayLength)
-                BufferPrimitivesThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.start);
-
-            _array = array;
-            _owner = null;
-            _index = start;
-            _length = arrayLength - start;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public Buffer(T[] array, int start, int length)
         {
             if (array == null)
--- a/src/System.Buffers.Primitives/System/Buffers/Buffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Buffer.cs
@@ -26,11 +26,48 @@ namespace System.Buffers
             _length = length;
         }
 
-        internal Buffer(T[] array, int index, int length)
+        public Buffer(T[] array)
         {
+            if (array == null)
+                BufferPrimitivesThrowHelper.ThrowArgumentNullException(ExceptionArgument.array);
+            if (default(T) == null && array.GetType() != typeof(T[]))
+                BufferPrimitivesThrowHelper.ThrowArrayTypeMismatchException(typeof(T));
+
             _array = array;
             _owner = null;
-            _index = index;
+            _index = 0;
+            _length = array.Length;
+        }
+
+        public Buffer(T[] array, int start)
+        {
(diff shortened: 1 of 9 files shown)


src/System.Buffers.Primitives/System/Buffers/Buffer.cs
```diff
@@ -26,11 +26,48 @@ internal Buffer(OwnedBuffer<T> owner, int index, int length)
             _length = length;
         }
 
-        internal Buffer(T[] array, int index, int length)
+        public Buffer(T[] array)
```
KrzysztofCwalina: Why don't the short ctor just call the longet one? i.e. Buffer(T[] array) : this(array, 0, array.Length)?

tests/System.Buffers.Primitives.Tests/CtorArray.cs
```diff
@@ -0,0 +1,216 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Xunit;
+
+namespace System.Buffers.Tests
+{
+    public class CtorArray
+    {
+        [Fact]
+        public static void CtorArray1()
```
KrzysztofCwalina: Should it be called CtorArrayInt and the one below CtorArrayLong?

tests/System.Buffers.Primitives.Tests/TestHelpers.cs
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+using Xunit;
+
+namespace System.Buffers.Tests
+{
+    internal static class TestHelpers
+    {
+        public static void Validate<T>(Buffer<T> buffer, params T[] expected)
```
KrzysztofCwalina: Should it be called SequenceEqual? or something like that?

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
shiftylogic: Agreed. Same for ReadOnlyBuffer.

tests/System.Buffers.Primitives.Tests/CtorArray.cs
```diff
@@ -0,0 +1,216 @@
...
+            TestHelpers.Validate(buffer, 42, -1);
+
+            buffer = new Buffer<int>(aAsIntArray, 0, aAsIntArray.Length);
+            TestHelpers.Validate(buffer, 42, -1);
+        }
+
+        [Fact]
+        public static void CtorArrayInt1()
+        {
+            int[] a = { 90, 91, 92, 93, 94, 95, 96, 97, 98 };
```
shiftylogic: Many of these functions could be merged into a single test with test data for running the same test several times.

Your call though.

tests/System.Buffers.Primitives.Tests/TestHelpers.cs
shiftylogic: You could write 2 versions of this function. One has a "where T : struct, IEquatable<T>" which handles all your value types and can be routed almost directly to Span.SequenceEqual. The other would need to do the iteration and reference comparison.

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
ahsonkhan: I followed the [Span\<T\> implementation](https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Span.cs#L30). Does doing this have a performance cost? If not, should we also make the same change to Span\<T\>?

The longer one has this extra check:
```C#
if ((uint)start > (uint)array.Length || (uint)length > (uint)(array.Length - start))
    BufferPrimitivesThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.start);
```

tests/System.Buffers.Primitives.Tests/CtorArray.cs
ahsonkhan: Can you explain how they could be merged? I borrowed these from Span tests.

Do you mean tests like these can be combined?
- CtorArrayInt
- CtorArrayWithStartInt
- CtorArrayWithStartAndLengthInt

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
ahsonkhan: I can't do this since array could be null and then array.Length would throw null exception.

ahsonkhan:
cc @KrzysztofCwalina, I made this constructor internal instead of private since it is required for the implicit cast from Buffer\<T\> to ReadOnlyBuffer\<T\>. Is that fine?
```C#
internal ReadOnlyBuffer(OwnedBuffer<T> owner, T[] array, int index, int length)
```

There was a bug in the implicit cast since it was using the owner only constructor which could be null if Buffer was constructed using an array... leading to both owner and array to be null after the cast.

ahsonkhan:
Fixing Buffer\<T\>.Slice implementation to catch errors and renaming index to start to be consistent with Span\<T\>.

ahsonkhan:
@KrzysztofCwalina, good to go?

Tornhoof:
Maybe rename ``index`` in ``internal ReadOnlyBuffer(OwnedBuffer<T> owner, T[] array, int index, int length)`` to ``start`` like in the other ctors?

ahsonkhan:
> Maybe rename ``index`` in ``internal ReadOnlyBuffer(OwnedBuffer<T> owner, T[] array, int index, int length)`` to ``start`` like in the other ctors?

It is internal which is why I left it for now. I will address this instance and other inconsistencies of naming index vs start in the future.

PR #1603, opened 2017-06-09
Move UTF-8 string to own assembly.

shiftylogic:
* Pulled all relevant parts out of System.Text.Primitives.
* Removed methods from Encoders that were internal and only used by Utf8String. They now exist in a helper class in the Utf8String assembly.
* Extracted the Utf8String tests into their own test assembly.
* Had to patch a few other assemblies (dependencies and small surgical fixes).
* There is more cleanup needed in the encoders, but it will be happening in a next wave of changes as we shift around the encoder design.

@KrzysztofCwalina @ahsonkhan


tests/System.Text.Primitives.Tests/RandomTests.cs
```diff
@@ -4,727 +4,12 @@
...
 using System.Collections.Generic;
 using System.Linq;
-using System.Text.Utf8;
-using System.Text.Utf16;
 using Xunit;
-using Xunit.Abstractions;
 
 namespace System.Text.Primitives.Tests
 {
     public class Utf8StringTests
```
ahsonkhan: Rename class?

tests/System.Text.Utf8String.Tests/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,22 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
ahsonkhan: Is this AssemblyInfo file necessary?

src/System.Text.Utf8String/Properties/AssemblyInfo.cs
```diff
@@ -0,0 +1,27 @@
...
+using System.Reflection;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+// General Information about an assembly is controlled through the following
+// set of attributes. Change these attribute values to modify the information
+// associated with an assembly.
+[assembly: AssemblyTrademark("")]
+
+[assembly: InternalsVisibleTo("System.Text.Utf8String.Tests,PublicKey="+
```
ahsonkhan: Can we make the following classes public and then remove this AssemblyInfo file?
- Utf16LittleEndianCodePointEnumerable
- Utf16LittleEndianCodePointEnumerator

Other than these, what else does the Utf8String Test project need that is internal?

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf16LittleEndianStringExtensions.cs
```diff
@@ -1,15 +0,0 @@
-// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-namespace System.Text.Utf8
-{
-    internal static class Utf16LittleEndianStringExtensions
-    {
-        // TODO: Naming it Equals causes picking up wrong overload when compiling (Equals(object))
-        public static bool EqualsUtf8String(this string left, Utf8String right)
```
ahsonkhan: Did you intentionally leave this out from the move?

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -35,25 +35,6 @@ public Utf8String(byte[] utf8bytes, int index, int length)
             _buffer = new ReadOnlySpan<byte>(utf8bytes, index, length);
         }
 
-        // TODO: reevaluate implementation
-        public Utf8String(IEnumerable<uint> codePoints)
```
ahsonkhan: intentional exclusion?

tests/System.Text.Utf8String.Tests/Properties/AssemblyInfo.cs
shiftylogic: For shipping libraries, it is necessary to get versioning, description, etc. However, it is not really necessary in this particular assembly since we aren't doing any friend assembly stuff. I'll remove it. We can always add one later if we need it.

src/System.Text.Utf8String/Properties/AssemblyInfo.cs
shiftylogic: Making these classes public is probably not the right thing to do right now. This entire assembly is going to be re-designed and possibly removed at some point in the near future. So, I'm leaving it as is for now.

src/System.Text.Primitives/System/Text/Encoding/Utf8/Utf16LittleEndianStringExtensions.cs
shiftylogic: I removed it because it is not being used at all. Given it was internal only, nobody external could have accessed it either.

src/System.Text.Utf8String/System/Text/Utf8String.cs
shiftylogic: Yes. The chunk of code that was using it was written slightly different and it allowed me to remove an extra encoder method.

Issue #1144, opened 2017-01-21
Add Tests for Memory<T> to validate the != and == operators


KrzysztofCwalina:
Related to PR #1143 
https://github.com/dotnet/corefxlab/pull/1143#issuecomment-274281682

PR #1607, opened 2017-06-13
Adding Buffer Equals tests and changing API based on design review

ahsonkhan:
Fixes https://github.com/dotnet/corefxlab/issues/1144

- Removed unnecessary (redundant) ==, !=, and Equals overloads that take permutations of ReadOnlyBuffer and Buffer.
- Combined Retain and Pin by adding a bool pin argument to Retain.
- Buffer and ReadOnlyBuffer no longer implement the IEquatable interface.
- Added SequenceEqual extension method on Buffer\<T\> and ReadOnlyBuffer\<T\>.
- Fixed bug in implementation of `operator !=(ReadOnlyBuffer<T> left, ReadOnlyBuffer<T> right)`
- Fixed bug in implementation of `bool Equals(object obj)` for ReadOnlyBuffer\<T\>.
- Added tests for == and != for Buffer\<T\> and ReadOnlyBuffer\<T\>.


src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
```diff
@@ -124,7 +124,7 @@ public WritableDeflateTransform(CompressionLevel compressionLevel, int bits)
 
                     unsafe
                     {
-                        var handle = buffer.Pin();
+                        var handle = buffer.Retain(true);
```
KrzysztofCwalina: It might be better to use named parameter here, i.e. buffer.Retain(pin: true);

KrzysztofCwalina:
Looks good.

src/System.Buffers.Primitives/System/Buffers/Buffer.cs
```diff
@@ -126,28 +126,38 @@ public Span<T> Span
...
+            else
             {
-                return _owner.Pin(_index);
-            }
-            var handle = GCHandle.Alloc(_array, GCHandleType.Pinned);
-            unsafe
-            {
-                var pointer = OwnedBuffer<T>.Add((void*)handle.AddrOfPinnedObject(), _index);
-                return new BufferHandle(null, pointer, handle);
+                if (_owner != null)
```
shiftylogic: This could be:
if (_owner != null) _owner.Retain();
bufferHandle = new BufferHandle(_owner);

since _owner being null in the ctor is acceptable.

ahsonkhan: Is that likely to change or is that the correct behaviour? @KrzysztofCwalina 

If using the default constructor of BufferHandle is acceptable (where void* _pointer == null and IRetainable _owner == null), then I will make this change.

src/System.IO.Pipelines.Compression/CompressionPipelineExtensions.cs
ahsonkhan: Specifically here only or in most/all places? What is your rationale for adding it here specifically?

Issue #1434, opened 2017-04-06
JsonDynamicObject and JsonLazyDynamicObject need to be re-designed


ahsonkhan:
JsonDynamicObject wraps around `Dictionary<JsonProperty, JsonValue>` and JsonProperty & JsonValue both contain Utf8String fields, which are stack-only, and hence cannot be used as a generic type in `Dictionary<T>`.

Also JsonProperty implements `IEquatable<JsonProperty>` which is not allowed for the same reason.

JsonLazyDynamicObject is a class and has a JsonObject field which contains a `ReadOnlySpan<byte>` field. A value type containing a by-ref instance field, such as Span<T>, cannot be used as the type for a class instance field.

**Edit:** With the new compiler that enforces correct Span usage, incorrect uses have been commented out. From https://github.com/dotnet/corefxlab/pull/1608: 
https://github.com/dotnet/corefxlab/pull/1608/files#diff-d7c8e1ff012098966b3cdaa243f77ae0R229
```C#
//TODO: no spans on the heap
```

These should be resolved as part of this issue.

ahsonkhan:
cc @KrzysztofCwalina, @shiftylogic

KodrAus:
I'd be interested in getting on board with this.

ahsonkhan:
> I'd be interested in getting on board with this.

Awesome :)

KodrAus:
So currently `JsonObject` doesn't own its values, which is reasonable, but comes with those caveats. Maybe we could layer it so there's a lower-level borrowed layer that requires an input `Span` for building up the db and for reading values etc, and a wrapper around that that requires owned buffers and has an ergonomic API for clients?

KodrAus:
So I guess the design here could depend on the design for `Buffer` that comes out of #1437, but in the case where `Buffer` is just a 'fat span' I think changing `Span`s in `JsonObject` to `Buffer`s is a bit sketchy. Storing `Buffer`s in fields would be problematic because you can't guarantee their lifetime and if that storage is opaque to API users then they might not even know you're doing it and leave you with a dangling pointer.

#1447 may also affect this if the design for utf8-strings changes.

I'm still keen to play with the idea of an owned and borrowed variant for Jason objects.

KodrAus:
Does anybody else have any thoughts on this? @shiftylogic?

shiftylogic:
I am definitely interested in exploring this with other folks. However, it is lower on my priority list at the moment.

[...]

PR #1608, opened 2017-06-13
New toolset compiler

VSadov:
The new version is: 2.6.0-rdonly-ref-61810-01
The corresponding VSIX (if need VS support) can be obtained at: 
https://dotnet.myget.org/feed/roslyn/package/vsix/0b48e25b-9903-4d8b-ad39-d4cca196e3c7/2.6.0.6181001

### DIFF
--- a/src/System.Text.Http/System/Text/Http/Obsolete/HttpRequestParser.cs
+++ b/src/System.Text.Http/System/Text/Http/Obsolete/HttpRequestParser.cs
@@ -6,7 +6,7 @@ using System.Text.Utf8;
 namespace System.Text.Http.SingleSegment
 {
     [Obsolete("Use System.Text.Http.HttpRequest")]
-    public struct HttpRequestSingleSegment
+    public ref struct HttpRequestSingleSegment
     {
         private HttpRequestLine _requestLine;
         private HttpHeadersSingleSegment _headers;
@@ -65,11 +65,15 @@ namespace System.Text.Http.SingleSegment
         }
     }
 
-    public struct HttpRequestReader
+    public ref struct HttpRequestReader
     {
-        static readonly Utf8String s_Http1_0 = new Utf8String("HTTP/1.0");
-        static readonly Utf8String s_Http1_1 = new Utf8String("HTTP/1.1");
-        static readonly Utf8String s_Http2_0 = new Utf8String("HTTP/2.0");
+        static readonly byte[] b_Http1_0 = new Utf8String("HTTP/1.0").CopyBytes();
+        static readonly byte[] b_Http1_1 = new Utf8String("HTTP/1.1").CopyBytes();
+        static readonly byte[] b_Http2_0 = new Utf8String("HTTP/2.0").CopyBytes();
+
+        static Utf8String s_Http1_0 => new Utf8String(b_Http1_0);
+        static Utf8String s_Http1_1 => new Utf8String(b_Http1_1);
+        static Utf8String s_Http2_0 => new Utf8String(b_Http2_0);
 
         internal static readonly byte[] Eol = new byte[] { s_CR, s_LF };
         internal static readonly byte[] Eoh = new byte[] { s_CR, s_LF, s_CR, s_LF }; // End of Headers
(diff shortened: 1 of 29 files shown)


VSadov:
There were some major changes in the way we create the compiler package compatible with CoreCLR.
I wanted to look at test results - to make sure the package is indeed used, etc...

ahsonkhan:
> I wanted to look at test results - to make sure the package is indeed used, etc...

Good idea.
Are we confident that the right compiler and the new features are being used (and correctly)?

VSadov:
I think it uses the default compiler. Probably since that technically has a higher version.

VSadov:
I was referring to a package that does not exist.   `2.3`  should be `2.6`

VSadov:
Ok, now we see violations when span contained structs are not `ref` structs.
Definitely a new toolset.  :-)

ahsonkhan:
> System/Text/Utf8String.cs(15,17): error CS0610: Field or property cannot be of type 'ReadOnlySpan<byte>' [/Users/dotnet-bot/j/workspace/dotnet_corefxlab/master/osx10.12_debug_prtest/src/System.Text.Utf8String/System.Text.Utf8String.csproj]

If the failures all have to do with UTf8String, I wonder if we should just exclude that project from the build until the re-design.

VSadov:
I think we are ok to merge now.

ahsonkhan:
> We should file issues for the TODO comments that resulted from this change. 

I updated the JSON issue to include this: https://github.com/dotnet/corefxlab/issues/1434

tests/System.Text.Http.Tests/GivenAnHttpHeaders.cs
```diff
@@ -54,11 +54,13 @@ public void Its_enumerator_Current_returns_the_same_item_until_MoveNext_gets_cal
             enumerator.MoveNext();
 
             var current = enumerator.Current;
-            current.Should().Be(enumerator.Current);
+            //TODO: no spans on the heap
```
ahsonkhan: @shiftylogic, I don't know what we should do about this. It is part of a skipped test already. Thoughts?

VSadov: I can take care of this by comparing the pair elements individually. This TODO will go away.
I was not sure if that is interesting considering that the test is disabled.

VSadov:
test Innerloop Ubuntu16.04 Debug Build and Test please

ahsonkhan:
> some testcase flakiness?

Yes. Specific to Linux, there is a test in Pipelines that is flaky which we need to resolve.

VSadov:
I think all new TODOs are matched with bugs now

src/System.Text.Http/System/Text/Http/Obsolete/HttpRequestParser.cs
```diff
@@ -65,11 +65,15 @@ public static HttpRequestSingleSegment Parse(ReadOnlySpan<byte> bytes)
...
+    public ref struct HttpRequestReader
     {
-        static readonly Utf8String s_Http1_0 = new Utf8String("HTTP/1.0");
-        static readonly Utf8String s_Http1_1 = new Utf8String("HTTP/1.1");
-        static readonly Utf8String s_Http2_0 = new Utf8String("HTTP/2.0");
+        static readonly byte[] b_Http1_0 = new Utf8String("HTTP/1.0").CopyBytes();
+        static readonly byte[] b_Http1_1 = new Utf8String("HTTP/1.1").CopyBytes();
+        static readonly byte[] b_Http2_0 = new Utf8String("HTTP/2.0").CopyBytes();
+
+        static Utf8String s_Http1_0 => new Utf8String(b_Http1_0);
```
ahsonkhan: To be explicit, this change works because:
- Fields cannot contain structs that contain span (like Utf8String)
- Properties can contain structs that contain span

@VSadov, is that correct?

cc @KrzysztofCwalina

VSadov: yes, properties/methods can return spans or ref structs in general. 
The fields are not allowed, unless they are instance fields in a ref struct.

PR #1609, opened 2017-06-14
Changing ITransformation to abstract class and TryGetArray arg rename

ahsonkhan:
- TryGetArray(out ArraySegment<byte> buffer) => TryGetArray(out ArraySegment<byte> arraySegment)
- Moving Buffer\<T\> & ReadOnlyBuffer\<T\> to System namespace (from System.Buffers)
- interface ITransformation => abstract class Transformation

cc @shiftylogic, @KrzysztofCwalina

### DIFF
--- a/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
@@ -282,9 +282,9 @@ namespace System.Binary.Base64
             return TransformationStatus.InvalidData;
         }
 
-        class FromBase64 : ITransformation
+        class FromBase64 : Transformation
         {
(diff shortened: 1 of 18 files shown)


src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -282,9 +282,9 @@ public static TransformationStatus DecodeInPlace(Span<byte> buffer, out int byte
             return TransformationStatus.InvalidData;
         }
 
-        class FromBase64 : ITransformation
+        class FromBase64 : Transformation
```
shiftylogic: These classes should be sealed. Based on conversations with @russellhadley, this will give JIT a better chance at de-virtualizing the calls.

KrzysztofCwalina:
looks good.

PR #1615, opened 2017-06-16
Priority queue proposal

pgolebiowski:
@safern @ianhays 
Adding API proposal. Link to the discussion (and specific request for this PR): [#574](https://github.com/dotnet/corefx/issues/574#issuecomment-308609496).

### DIFF
--- a/docs/specs/priority-queue.md
+++ b/docs/specs/priority-queue.md
@@ -1,0 +1,219 @@
+# Priority queue proposal
+
+## User data
+
+Before anything is proposed, let's start with what can be prioritized...
+
+### Input types
+
+We want to be able to support all of the following configurations:
+
+1. User data *is separate* from the priority (two physical instances).
+2. User data *contains* the priority (as one or more properties).
+3. User data *is* the priority (implements `IComparable<T>`).
+4. Rare case: priority is obtainable via some other logic (resides in an object different from the user data).
+
+In this entire document, I will refer to the cases above with **(1)**, **(2)**, **(3)**, and **(4)**.
+
+### Ideal solution
+
+Obviously, our solution should be flexible enough to (respectively):
+
+1. Simply accept two separate instances. The user should not be forced to create a wrapper class for the two types only because of our API limitations.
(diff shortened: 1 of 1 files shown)


KrzysztofCwalina:
I really like the separation between the single type parameter type and the two generic parameter type. I have seen earlier proposals with just one two parameter type, and I think this proposal is much better.

In addition, I wonder if we should not start with just one type, and for cases where the priority is external, use ValueTuple, e.g.

            var queue = new PriorityQueue<(int priority, string value)>(comparer);

If this is not continent enough, a set of static methods (some extensions) could be prototype to make it easier to create such queues and to enqueue/dequeue items to them.

svick:
@KrzysztofCwalina 

> If this is not continent enough, a set of static methods (some extensions) could be prototype to make it easier to create such queues and to enqueue/dequeue items to them.

Would that actually work? Consider this API fragment:

```c#
class PriorityQueue<T>
{
    public T Dequeue();
}

static class TuplePriorityQueue
{
    public static PriorityQueue<(TPriority priority, TElement element)> Create<TPriority, TElement>();

    public static TElement Dequeue<TPriority, TElement>(
        this PriorityQueue<(TPriority priority, TElement element)> queue);
    public static TElement Dequeue<TPriority, TElement>(
        this PriorityQueue<(TPriority priority, TElement element)> queue,
        out TPriority priority);
}
```

I think for the "two-generic" version, having a version of `Dequeue` that returns just the element (but not the priority) would be useful. But the above API wouldn't work, because you couldn't call that version of `Dequeue` as extension method, the instance method would always take priority. There are workarounds (like using a different name for the extension method), but I'm not convinced it could be good enough.

KrzysztofCwalina:
Yeah, the method would have to be called DequeueFirst, or something like that. I am not saying that it would be the best API we can get, but rather that maybe it's worth starting with this approach and it it turns out it's very commonly used and does not feel right, we could add the second type. 

BTW, I am not strongly opposed to adding two types, just throwing an idea here.

PR #1616, opened 2017-06-16
Optimizing Int32 Primitive Parsers and clean up

ahsonkhan:
- Also adding some performance tests and additional test cases.
- Optimized Invariant UTF-8 and Non-Invariant (for Int32). I would imagine similar optimizations can be applied to the rest.

**Here are the key optimizations:**
- TBD

image

image

**Edit:** Updated the results with latest changes


ahsonkhan:
cc @jkotas 

Any thoughts on how I can improve the performance for small integers/strings for Invariant UTF8 TryParse?
```C#
public static bool TryParseInt32(ReadOnlySpan<byte> text, out int value, out int bytesConsumed)
```
I think we can get some more savings to avoid the ~10% perf regression for a single digit.

jkotas:
First, you should fix the buffer overruns in your unsafe code before trying to figure out how to optimize it.

I do not think you should be using unsafe code for number parsing. The overhead of the bounds check should be neglible - assuming that the JIT works as expected.

jkotas:
And yes ... I can be optimized. You can start by reading the first character from memory no more than once.

jkotas:
@alexandrnikitin is working on improving Int32 parser in CoreLib: https://github.com/dotnet/coreclr/pull/12196. Some of the tricks can be shared.

ahsonkhan:
The PR is ready for review. I have updated the performance results in the original post.

src/System.Text.Primitives/System/Text/Parsing/InvariantSigned.cs
```diff
@@ -899,176 +899,89 @@ public unsafe static bool TryParseInt32(byte* text, int length, out int value, o
...
-                }
+                int textLength = text.Length;
+                if (textLength < 1) goto FalseExit;
 
-                int indexOfFirstDigit = 0;
-                int sign = 1;
-                if (text[0] == '-')
+                sbyte sign = 1;
+                int index = 0;
+                byte num = text[index];
```
jkotas: It will likely work better for this to be `int`, so that you pay for byte->int zero extension just once.

src/System.Text.Primitives/System/Text/Parsing/InvariantSigned.cs
```diff
@@ -899,176 +899,89 @@ public unsafe static bool TryParseInt32(byte* text, int length, out int value, o
...
+                    num = text[index];
                 }
 
-                int overflowLength = Int32OverflowLength + indexOfFirstDigit;
+                int answer = 0;
 
-                // Parse the first digit separately. If invalid here, we need to return false.
-                int firstDigit = text[indexOfFirstDigit] - 48; // '0'
-                if (firstDigit < 0 || firstDigit > 9)
+                if (num >= '0' && num <= '9')
```
jkotas: You can use a trick like `(num - '0') <= ('9' '- '0')` to save an extra conditional branch in these.

stephentoub: It'd be nice if the JIT would do this (can't remember if there's an issue tracking it).  Better yet, the C# compiler or an IL optimizer.

jkotas: > if there's an issue tracking it

I am not able to find one (it was discussed in https://github.com/dotnet/coreclr/pull/4881 and other PRs). @ahsonkhan Could you please make one?

VSadov: Logged https://github.com/dotnet/roslyn/issues/20375 on the C# side.

src/System.Text.Primitives/System/Text/Parsing/InvariantSigned.cs
```diff
@@ -2314,8 +2227,14 @@ public unsafe static bool TryParseInt32(char* text, int length, out int value, o
...
             public static bool TryParseInt32(ReadOnlySpan<char> text, out int value)
             {
+                return TryParseInt32(text, out value, out int charsConsumed);
+            }
+
+            public static bool TryParseInt32(ReadOnlySpan<char> text, out int value, out int charsConsumed)
+            {
                 if (text.Length < 1)
                 {
+                    charsConsumed = 0;
```
VSadov: `default(int)` ?  is that just `0` ?

benaadams: yes, though `default` implies it isn't necessarily set vs explicitly setting to `0` which suggests a chosen value? So semantically using `value = default(int);` makes more sense?

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
```diff
@@ -4,10 +4,69 @@
...
+
+        private const sbyte maxValueSbyteDiv10 = sbyte.MaxValue / 10;
+        private const short maxValueShortDiv10 = short.MaxValue / 10;
+        private const int maxValueIntDiv10 = int.MaxValue / 10;
+        private const long maxValueLongDiv10 = long.MaxValue / 10;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool IsDigit(int i)
+        {
+            return i >= 0 && i <= 9;
```
VSadov: `(uint)i <= 9`

tannergooding: Why? The first check should already have guaranteed the value is not negative.

src/System.Text.Primitives/System/Text/Parsing/InvariantSigned.cs
ahsonkhan: Yes, of course. This code will likely change quite a bit, so I will address it then.

src/System.Text.Primitives/System/Text/Parsing/InvariantSigned.cs
ahsonkhan: Tyvm :)

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
ahsonkhan:
```C#
private static bool IsDigitA(int i)
{
    return (uint)i <= 9;
}

private static bool IsDigitB(int i)
{
    return i >= 0 && i <= 9;
}
```

C.IsDigitA(Int32)
    L0000: cmp ecx, 0x9
    L0003: setbe al
    L0006: movzx eax, al
    L0009: ret

C.IsDigitB(Int32)
    L0000: test ecx, ecx
    L0002: jl L000e
    L0004: cmp ecx, 0x9
    L0007: setle al
    L000a: movzx eax, al
    L000d: ret
    L000e: xor eax, eax
    L0010: ret

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
```diff
@@ -4,10 +4,69 @@
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool IsValid(uint i)
+        {
+            return i <= (uint)TextEncoder.Symbol.D9;
+        }
+
+        // If parsedValue > (sbyte.MaxValue / 10), any more appended digits will cause overflow.
+        // if parsedValue == (sbyte.MaxValue / 10), any nextDigit greater than 7 or 8 (depending on sign) implies overflow.
```
VSadov: the second line of the comment does not match the implementation. not sure which is right. the comment?

ahsonkhan: Is this the part you are referring to? 
```C#
// if parsedValue == (sbyte.MaxValue / 10)
```

Are you suggesting something like this:
```C#
// If parsedValue > (sbyte.MaxValue / 10), any more appended digits will cause overflow.
// Else, any nextDigit greater than 7 or 8 (depending on sign) implies overflow.
```

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
tannergooding: Ah, I misread as `return i >=0 && (uint)i <= 9` rather than just replace with `return (uint)i <= 9`.

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
ahsonkhan: Hmm, just realized we need still need this check (i.e. the comment is correct):
```C#
if parsedValue == (sbyte.MaxValue / 10)
```

Will fix.

src/System.Text.Primitives/System/Text/Parsing/InvariantSigned.cs
```diff
@@ -899,176 +899,89 @@ public unsafe static bool TryParseInt32(byte* text, int length, out int value, o
...
-                        parsedValue = parsedValue * 10 + nextDigit;
+                            answer = answer * 10 + num - '0';
+                            index++;
+                            if (index - firstNonZeroDigitIndex == Int32OverflowLength - 1)
+                            {
+                                num = text[index];
+                                if (num >= '0' && num <= '9')
+                                {
+                                    num -= (byte)'0';
+                                    if (WillOverFlow(answer, num, sign)) goto FalseExit;
```
VSadov: This seems very complicated. My suggestion would be 
- separate parsing of input longer than Int32OverflowLength into a separate method. Shorter numbers are much more common.
- WillOverFlow seems pretty complex for what it does.   `a * 10 + b`  is not so expensive nowdays. Perhaps it is cheaper to just add and see if sign flips ?

VSadov: Here is basically the idea for parsing numbers longer than Int32OverflowLength, I wonder if it will work better:

```C#
        {
            for (long i = int.MaxValue - 1000; i < int.MaxValue + 10L; i++)
            {
                // lets parse "chars" 
                string chars = i.ToString();

                // this will need to come from parsing leading '-'
                int sign = 1;

                // here is our result
                int accumulator = 0;

                foreach (var c in chars)
                {
                    // will need to check if 'c' is a digit. Assume it is..
                    var d = c - '0';

                    // skip leading 0s
                    if ((d | accumulator) == 0)
                        continue;

                    // try add a digit
                    if (!TryAddNoOverflow(ref accumulator, d, sign))
                    {
                        System.Console.WriteLine("not representable " + (i));
                    }
                }
            }
        }

        // accumulate the value and ensure the sign
        // returns false when number is no longer representable in 32bit
        static bool TryAddNoOverflow(ref int value, int nextDigit, int sign)
        {
            var newValue = value * 10 + (nextDigit * sign);

            // opposite sign check for {newValue, sign}
            if ((newValue ^ sign) < 0)
            {
                return false;
            }

            value = newValue;
            return true;
        }
```

ahsonkhan: > // opposite sign check for {newValue, sign}

This won't always give you the right result.

If value becomes > 2* int.MaxValue when a new digit gets added, the sign doesn't change.

TryAddNoOverflow(214748364, 7, 1); => Returns True, correct.
TryAddNoOverflow(514748364, 7, 1); => Returns True, wrong.

VSadov: right, i did not realize that. Most of this solution would not be applicable then.

src/System.Text.Primitives/System/Text/Parsing/InvariantSigned.cs
mikedn: > Can't remember if there's an issue tracking it

AFAIK there's no such issue. Though I happen to have an experimental JIT change that does this optimization. It remains to be seen what will come out of it.

ahsonkhan:

ahsonkhan:

ahsonkhan:
> Here is what I have in mind (other suggestions are incorporated as well:

```C#
index++;
if (index < textLength) goto Done;
if (!IsDigit(text[index])) goto Done;
```
Should line 2 above be `if (index <= textLength)`? Otherwise, there is an argument out of range exception.
Also, after changing line 2, a test like this would fail: Parse("21474836471") - should return false, it returns true since `if (lAnswer > Int32.MaxValue)` is false.

ahsonkhan:
@jkotas, @VSadov 
Thoughts on a loop unrolled version? It seems quite promising:
https://gist.github.com/ahsonkhan/dba0afb3304de15be844111d0f8a3ea6

jkotas:
> Should line 2 above be if (index <= textLength)? 

Yes, it should be `if (index >= textLength)`.

jkotas:
> Thoughts on a loop unrolled version? It seems quite promising:

The switch works great for microbenchmarks that are not sensitive to code size (the manual unrolling is like 10x code bloat); and that are always parsing numbers with exact same number of digits - the indirect jump that switch compiles into is well predicted. Once you throw more realistic mix on it - where the number of digits is not always the same and there is a bunch of other code running - it becomes not so good. We used to have similar switch in the memcpy implementation in CoreLib. https://github.com/dotnet/coreclr/pull/9786 got rid of it because of it did not work for real workloads. 

So it is tricky to see whether the switch is a good thing on microbenchmarks.

jkotas:
Loop unrolled version without a switch may be interesting though. Something like:

```
   if (index >= textLength) goto Done;
   num = text[index];
   if (!IsDigit(num)) goto Done;
   index++;
   answer = 10 * answer + num - '0';

   ... same block repeated ...

   if (index >= textLength) goto Done;
   num = text[index];
   if (!IsDigit(num)) goto Done;
   index++;
   answer = 10 * answer + num - '0';

   // Potential overflow
   if (index >= textLength) goto Done;
   num = text[index];
   if (!IsDigit(num)) goto Done;
   long lAnswer = (long)answer * 10 + num - '0';
   if (sign < 0)
   {
       if (lAnswer > (long)Int32.MaxInt + 1) goto FalseExit;
   }
   else
   {
       if (lAnswer > (long)Int32.MaxInt) goto FalseExit;
   }
   answer = (int)lAnswer;
   index++;

   if (index >= textLength) goto Done;
   if (!IsDigit(text[index])) goto Done;

   // Guaranteed overflow
   goto FalseExit;
```

You can also tune the code bloat factor by unrolling just the first few digits - where it improves the microbenchmark significantly, and run a regular loop for the rest - where the unrolling benefits starts to diminish.

src/System.Text.Primitives/System/Text/Parsing/Signed.cs
```diff
@@ -237,122 +295,128 @@ public static bool TryParseInt16(ReadOnlySpan<byte> text, out short value, out i
...
-                value = default;
-                bytesConsumed = 0;
-                return false;
-            }
+            int textLength = text.Length;
+            if (textLength < 1) goto FalseExit;
 
-            int sign = 1;
-            if ((TextEncoder.Symbol)nextSymbol == TextEncoder.Symbol.MinusSign)
+            ref byte textByte = ref text.DangerousGetPinnableReference();
```
jkotas: This should not be using unsafe code either.

ahsonkhan:
> Loop unrolled version without a switch may be interesting though.

Can you explain how that can potentially be an optimization? With the switch, you avoid the `(index >= textLength)` checks, but I don't see how loop unrolling by repeating the blocks (as per your suggestion) saves any checks/instructions.

Are the savings coming from removing the initial if/else condition (i.e. `if (textLength - index < Int32OverflowLength - 1)`)?

jkotas:
>  With the switch, you avoid the (index >= textLength) checks,

I do not think you will avoid them. The JIT will still generate them for the bounds checks when accessing the Span.

ahsonkhan:
> So it is tricky to see whether the switch is a good thing on microbenchmarks.

I would imagine that this test case accounts for this, no?
PrimitiveParserByteSpanToInt32_BytesConsumed_VariableLength

In the gist, you can see the switch version degrades 30% in this scenario (whereas other tests showed a 10-20% improvement).

```C#
for (int i = 0; i < Benchmark.InnerIterationCount; i++)
{
    int.TryParse(s_Int32TextArray[i % textLength], out int value);
    DoNotIgnore(value, 0);
}
```

> You can also tune the code bloat factor by unrolling just the first few digits - where it improves the microbenchmark significantly, and run a regular loop for the rest - where the unrolling benefits starts to diminish.

I unrolled for all and it seems to do really well in most scenarios. For 1-4 digit case, where there is slight regression, is not too bad given scenarios like this would be more common (especially for JSON parsing):
"-2abcdefghijklmnop"

ahsonkhan:
> I think that this WillOverFlow method is hard to get right and reason about. It is also becoming pretty complex, so it is probably not very performant. I think it would be better to just perform the multiplication - since you have to do it anyway - and then check whether the result overflowed.

Any ideas on how such an optimization would work for Int64 parsing? For ints, we can take advantage of the long primitive, but that strategy doesn't work when parsing longs. For longs, do we keep the WillOverFlow method?

ahsonkhan:
Is this optimization in the right direction? We save one conditional and substitute with some arithmetic instructions.
**From:**
```C#
if (sign< 0)
{
    if (lAnswer > (long) Int32.MaxInt + 1) goto FalseExit;
}
else
{
    if (lAnswer > (long) Int32.MaxInt) goto FalseExit;
}
```
**To:**
```C#
if (lAnswer > (long)Int32.MaxValue + (-1 * sign + 1) / 2) goto FalseExit;
```

image

jkotas:
> Any ideas on how such an optimization would work for Int64 parsing

The following modification should work equally well for both Int32 and Int64.

```
// Potential overflow 
if (index >= textLength) goto Done;
num = text[index];
if (!IsDigit(num)) goto Done; 
if (answer > Int32.MaxValue/10 + 1) goto FalseExit; // Overflow
answer = answer * 10 + num - '0'; 

if ((uint)answer > (uint)Int32.MaxValue + (-1 * sign + 1) / 2) goto FalseExit; // Overflow
index++;
if (index >= textLength) goto Done;
if (!IsDigit(text[index])) goto Done; 

// Guaranteed overflow 
goto FalseExit;
```

jkotas:
> Is this optimization in the right direction? We save one conditional and substitute with some arithmetic instructions

It looks better - do you see any measurable perf difference? It will be different if you take the unification for Int32/Int64 above. And for 32-bit vs. 64-bit platforms.

It would be interesting to see the code for the full parse method - for both x86 and x64. Could you please post it into a gist once you get something you are happy with?

ahsonkhan:
> The following modification should work equally well for both Int32 and Int64.

Ah, lol, leveraging unsigned int works :)

What about unsigned long parsing?

jkotas:
> What about unsigned long parsing?

Check that the answer did become smaller (same for uint32 parsing):

```
if (answer > UInt64.MaxValue/10 + 1) goto FalseExit; // Overflow
ulong answer2 = answer * 10 + num - '0'; 

if (answers2 < answer) goto FalseExit; // Overflow
answer = answer2;
index++;
```

BTW: long/ulong should have the loop fully unrolled - having the same bit repeated ~18x would be too much, I think.

ahsonkhan:
> BTW: long/ulong should have the loop fully unrolled - having the same bit repeated ~18x would be too much, I think.

should **NOT** have the loop fully unrolled?

jkotas:
right

ahsonkhan:
> It will be different if you take the unification for Int32/Int64 above. And for 32-bit vs. 64-bit platforms.
It would be interesting to see the code for the full parse method - for both x86 and x64. Could you please post it into a gist once you get something you are happy with?

@jkotas, here is the data and disassembly:
x64: https://gist.github.com/ahsonkhan/dba0afb3304de15be844111d0f8a3ea6#gistcomment-2131566
x86: https://gist.github.com/ahsonkhan/dba0afb3304de15be844111d0f8a3ea6#gistcomment-2131567

ahsonkhan:
Is the PR good to go or should I update the overflow checks to one of the following options?
**(This is currently in the PR)**
```C#
if (lAnswer > (long)Int32.MaxValue + (-1 * sign + 1) / 2) goto FalseExit;
```
**(Option 1)**
```C#
if (sign< 0)
{
    if (lAnswer > (long) Int32.MaxValue + 1) goto FalseExit;
}
else
{
    if (lAnswer > (long) Int32.MaxValue) goto FalseExit;
}
```
**(Option 2)**
```C#
if ((uint)answer > (uint)Int32.MaxValue + (-1 * sign + 1) / 2) goto FalseExit;
```
**(Option 3)**
```C#
if (sign< 0)
{
    if ((uint)answer > (uint)Int32.MaxValue + 1) goto FalseExit;
}
else
{
    if ((uint)answer > (uint)Int32.MaxValue) goto FalseExit;
}
```

jkotas:
> update the overflow checks to one of the following options?

I do not think that it matters a lot which option you pick. Option 2 is probably smallest/fastest code.

jkotas:
```
04BB16A4  cmp         edx,ebx  
04BB16A6  jge         04BB1923  
04BB16AC  cmp         edx,edi  
04BB16AE  jae         04BB1943  
```

This looks like that the caching of length in textLength local is hurting because of the JIT just burns a register for it. I think you should use `text.Length` instead everywhere, and get rid of the local.

Also, these bounds checks should not be duplicated. I know it was discussed before, but I do not remember the conclusion ... you may need to help the JIT to eliminate the redundant check by using unsigned comparison like `if ((uint)index >= (uint)text.Length)`.

ahsonkhan:
>  you may need to help the JIT to eliminate the redundant check by using unsigned comparison

That did not help reduce the duplicated bounds check.

ahsonkhan:
@shiftylogic - good to go?

shiftylogic:
This code makes assumptions that the values in the SymbolTable.Symbol enum are very specific. I suggest adding a comment to that enum stating that the parsing code depends on these specific value settings and that they should not be touched without being very careful of breaking the parsing code.

ahsonkhan:
> I suggest adding a comment to that enum stating that the parsing code depends on these specific value settings and that they should not be touched without being very careful of breaking the parsing code.

I believe the tests would fail if the enums changed. I will definitely add a comment though.
However, this PR did not introduce this dependency on the SymbolTable, it already existed.

PR #1617, opened 2017-06-16
Using new C# 7.1 language features - cleaning use of default

ahsonkhan:
cc @shiftylogic, @KrzysztofCwalina, @Vedin


shiftylogic:
Just curious why this change is necessary. This is a lot of churn for a trivial benefit.

ahsonkhan:
> Just curious why this change is necessary. This is a lot of churn for a trivial benefit.

Agreed. The change isn't really necessary, but we have new language features, so might as well use and try them, no?

Just as an example, is use of var necessary?

Otherwise, do you suggest we continue to use default expressions with type specified or just start using the shorter form from now on?

shiftylogic:
I'm not against using new language features. I'm just commenting that this is a lot of churn. My opinion would be to use new features for new code or when you go an update an area / fix a bug. However, going and wholesale changing the code every time we add a new language features seems like a lot of work.

ahsonkhan:
> However, going and wholesale changing the code every time we add a new language features seems like a lot of work.

Visual Studio was the beast of burden here so it wasn't very difficult. In general, I agree though.

KrzysztofCwalina:
I like cleanup work :-). Keeps sources from deteriorating over time.

davidfowl:
Does any of this work in the IDE?

ahsonkhan:
> Does any of this work in the IDE?

Yes. You have to install the VSIX linked here:
https://github.com/dotnet/corefxlab/pull/1608#issue-235443369

ahsonkhan:
cc @KrzysztofCwalina, shall I resolve the conflict and merge?

PR #1618, opened 2017-06-16
BrotliCanterburyPerf

Vedin:
As Canterbury Corpus set of files don't contains compressed brotli files, I just create own based on original files. The result below is in how many times GzipStream is faster than Brotli now:
Decompress:
image

Compress:
image

As we can see now Brotli is slower. But I don't know what really means CompressionLevel.Optimal in GzipStream, because in Brotli now it means the best compression ratio. @ianhays May be you can explain?
Also it is a specific files to test GZip, so may be in general difference is different. 

Thus, 
1. I need to optimize Brotli and BrotliStream to make it faster.
2. Compare a compression ratio


ianhays:
>But I don't know what really means CompressionLevel.Optimal in GzipStream, because in Brotli now it means the best compression ratio. @ianhays May be you can explain?

There was some contention around a year or two ago. Currently, CompressonLevel.Optimal maps to the "default" zlib compressionlevel which is a balanced middle in the speed vs space tradeoff. There is not, however, a *better* CompressionLevel option that represents the optimal compression ratio regardless of the speed at which the object is compressed. There was some testing done that showed that the "default" was almost always preferable over MaxRatio option because of diminishing returns in decreased output size at a high time cost, so we didn't bother with adding a MaxRatio.

The takeaway here is that yes, CompressionLevel.Optimal is the best compression ratio offered for GZipStream/DeflateStream.

ahsonkhan:
@Vedin, what do the four columns of data mean? Can you add a table header?

Vedin:

Vedin:

Vedin:
@ahsonkhan, @shiftylogic Can you please check and approve/comment changes?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -36,11 +36,20 @@ public partial class BrotliStream : Stream
...
 
-        public override int ReadTimeout { get ; set; }
+        public override int ReadTimeout { get; set; }
 
         public override int WriteTimeout { get; set; }
 
-        public BrotliStream(Stream baseStream, CompressionMode mode, bool leaveOpen, int bufferSize, uint windowSize, CompressionLevel quality) : this(baseStream, mode, leaveOpen, bufferSize)
+        public BrotliStream(Stream baseStream, CompressionMode mode, bool leaveOpen, int bufferSize, CompressionLevel quality) : this(baseStream, mode, leaveOpen, bufferSize)
+        {
+            if (_mode == CompressionMode.Decompress) throw new System.IO.IOException(BrotliEx.QualityAndWinSize);
```
shiftylogic: nit: format. There are braces on else but not if. Also, newline for throw.

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
```diff
@@ -0,0 +1,98 @@
...
+        /// <summary>
+        /// Benchmark tests to measure the performance of individually compressing each file in the
+        /// Canterbury Corpus
+        /// </summary>
+        [Benchmark]
+        [MemberData(nameof(CanterburyCorpus))]
+        public void Compress_Canterbury(int innerIterations, string fileName, CompressionLevel compressLevel)
+        {
+            byte[] bytes = File.ReadAllBytes(Path.Combine("BrotliTestData", "Canterbury", fileName));
+            int bufferSize = 8096;
```
shiftylogic: nit: 8kb == 8192.

Also, I'd consider moving this to a constant at class level.

ahsonkhan:
> The result below is in how many times GzipStream is faster than Brotli now

So, if I am reading the data correctly, currently:
GzipStream is 1.5-2x faster than Brotli for decompression and several factors faster for compression.

Is that right?

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
```diff
@@ -0,0 +1,98 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+using Microsoft.Xunit.Performance;
```
ahsonkhan: nit: add new line

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
```diff
@@ -0,0 +1,98 @@
...
+            }
+        }
+
+        /// <summary>
+        /// Benchmark tests to measure the performance of individually compressing each file in the
+        /// Canterbury Corpus
+        /// </summary>
+        [Benchmark]
+        [MemberData(nameof(CanterburyCorpus))]
+        public void Compress_Canterbury(int innerIterations, string fileName, CompressionLevel compressLevel)
```
ahsonkhan: What does innerIteration mean here? This is more about getting test data. I would avoid using this name sine it is already overloaded with benchmark inner iteration count

tests/System.IO.Compression.Tests/System.IO.Compression.Tests.csproj
```diff
@@ -22,7 +22,7 @@
     <ProjectReference Include="..\..\src\System.Buffers.Primitives\System.Buffers.Primitives.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <Content Include="BrotliTestData\*">
+    <Content Include="BrotliTestData\**">
```
ahsonkhan: I am curious, is this how wildcard works? What was the behavior before this change?

shiftylogic: ** is a GLOB... and works recursively.

tests/System.IO.Compression.Tests/Util.cs
```diff
@@ -0,0 +1,105 @@
...
+    {
+        public enum CompressionType
+        {
+            CryptoRandom,
+            RepeatedSegments,
+            VeryRepetitive,
+            NormalData
+        }
+
+        public static string GetTestFilePath(int? index = null, string memberName = null, int lineNumber = 0)
```
ahsonkhan: This is this method public while the others are internal?

tests/System.IO.Compression.Tests/System.IO.Compression.Tests.csproj
Vedin: The folder inside BrotliTestData didn't copy to the Appointment.

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
Vedin: Just use such name to be consistent with https://github.com/dotnet/corefx/blob/master/src/System.IO.Compression/tests/Performance/Perf.GZipStream.cs#L41
Should I change it?

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
Vedin: Do you mean new line after license?

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
ahsonkhan: > Should I change it?

I would say yes since I think that code was written prior to us having the inner iteration count feature and hence this naming conflict wasn't there. However, I don't have a strong opinion either way.

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
ahsonkhan: Yes.

Vedin:
Yes, Ahson. You are right.

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
Vedin: fixed

tests/System.IO.Compression.Tests/Util.cs
Vedin: Yes, make sense, thanks

PR #1621, opened 2017-06-21
Api brotli changes

Vedin:
Changes to the Brotli API. What's new:

1. Brotli (Primitives) was rewrote and now return all possible TransformationStatus.

2. BroltiStream now base on Brotli class. Also it uses twice less memory as before and doesn't use marshaling. 

3. @KrzysztofCwalina Finally, I remove everything from State and rewrite all in way where we need only BrotliNativeState and LastDecoderResult, So how to Dispose it know (there are 2 commented methods for compression/decompression mode? May be add some flag to detect in what mode state now?

4. BrotliPrimitivesTests needs an update.


src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -15,8 +15,67 @@ namespace System.IO.Compression
...
+        private const int MaxQuality = 11;
+
+        public struct State : IDisposable
+        {
+            internal IntPtr BrotliNativeState { get; private set; }
+
+            public void Dispose()
+            {
+                //    BrotliNative.BrotliDecoderDestroyInstance(BrotliNativeState);
+                //   BrotliNative.BrotliEncoderDestroyInstance(BrotliNativeState);
```
Vedin: @KrzysztofCwalina This 2 methods, which I mentioned above

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -15,8 +15,67 @@ namespace System.IO.Compression
...
+                BrotliNativeState = BrotliNative.BrotliEncoderCreateInstance();
+                if (BrotliNativeState == IntPtr.Zero)
+                {
+                    throw new System.IO.IOException(BrotliEx.EncoderInstanceCreate);
+                }
+            }
+
+            public void SetQuality(uint quality)
+            {
+                if (quality < MinQuality || quality > MaxQuality)
```
ahsonkhan: MinQuality is 0 and quality is an unsigned int. Quality will never be < MinQuality and hence this check is unnecessary.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -15,8 +15,67 @@ namespace System.IO.Compression
...
+            }
+
+            public void SetQuality()
+            {
+                SetQuality(MaxQuality);
+            }
+
+            public void SetWindow(uint window)
+            {
+                if (window < MinWindowBits || window > MaxWindowBits)
```
ahsonkhan: `if (window - MinWindowBits > MaxWindowBits - MinWindowBits)` saves an extra conditional branch.

See: https://github.com/dotnet/corefxlab/pull/1616#discussion_r123245176

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -44,16 +103,13 @@ private static TransformationStatus GetTransformationStatusFromBrotliDecoderResu
             return TransformationStatus.InvalidData;
         }
 
-        public static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, CompressionLevel quality = (CompressionLevel)DefaultQuality, int windowSize = DefaultWindowSize, BrotliEncoderMode encMode = BrotliEncoderMode.Generic)
-        {
-            return Compress(source, destination, out bytesConsumed, out bytesWritten, GetQualityFromCompressionLevel(quality), windowSize, encMode);
-        }
-
-        internal static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, int quality = DefaultQuality, int windowSize = DefaultWindowSize, BrotliEncoderMode encMode = BrotliEncoderMode.Generic)
+        public static TransformationStatus FlushEncoder(Span<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref State state, bool is_finished = true)
```
ahsonkhan: nit: argument name doesn't follow coding style

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -44,16 +103,13 @@ private static TransformationStatus GetTransformationStatusFromBrotliDecoderResu
...
         {
-            if (quality > DefaultQuality || quality < 0) throw new System.ArgumentOutOfRangeException(BrotliEx.WrongQuality);
-            if (windowSize > DefaultWindowSize || windowSize <= 0) throw new System.ArgumentOutOfRangeException(BrotliEx.WrongWindowSize);
-            bytesConsumed = bytesWritten = 0;
+            BrotliEncoderOperation operation = is_finished ? BrotliEncoderOperation.Finish : BrotliEncoderOperation.Flush;
+            bytesConsumed = source.Length;
+            bytesWritten = destination.Length;
```
ahsonkhan: These should not be set to source.Length and destination.Length in the beginning.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -64,20 +120,61 @@ internal static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<by
...
+                    if (!BrotliNative.BrotliEncoderCompressStream(state.BrotliNativeState, operation, ref consumed, ref bufIn, ref written, ref bufOut, out nuint totalOut))
                     {
                         return TransformationStatus.InvalidData;
-                    };
+                    }
                     bytesConsumed = (int)consumed;
                     bytesWritten = (int)written;
-                    return TransformationStatus.Done;
                 }
+                bytesWritten = destination.Length - bytesWritten;
```
ahsonkhan: I am curious, why do we do this?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -64,20 +120,61 @@ internal static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<by
...
+                        bufIn = new IntPtr(inBytes);
+                        bufOut = new IntPtr(outBytes);
+                        nuint written = (nuint)bytesWritten;
+                        nuint consumed = (nuint)bytesConsumed;
+                        if (!BrotliNative.BrotliEncoderCompressStream(state.BrotliNativeState, BrotliEncoderOperation.Process, ref consumed, ref bufIn, ref written, ref bufOut, out nuint totalOut))
+                        {
+                            return TransformationStatus.InvalidData;
+                        };
+                        bytesConsumed = (int)consumed;
+                        bytesWritten = (int)((nuint)destination.Length - written);
```
ahsonkhan: Won't this set bytesWritten to 0 in the success/done case, if written == destination.Length?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -64,20 +120,61 @@ internal static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<by
...
 
-        public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
+        public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref State state)
         {
-            bytesConsumed = bytesWritten = 0;
+            bool endOfStream = false;
+            bool errorDetected = false;
+            bytesConsumed = source.Length;
+            bytesWritten = destination.Length;
+            BrotliDecoderResult LastDecoderResult = BrotliDecoderResult.NeedsMoreInput;
```
ahsonkhan: Why is NeedsMoreInput the starting state of LastDecoderResult?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -86,17 +183,42 @@ public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<by
...
+                {
+                    return TransformationStatus.NeedMoreSourceData;
+                }
+                else if (LastDecoderResult == BrotliDecoderResult.NeedsMoreOutput)
+                {
+                    return TransformationStatus.DestinationTooSmall;
+                }
+                else
+                {
+                    endOfStream = true;
```
ahsonkhan: Is this bool necessary? The other two conditions return, so reaching here automatically implies endOfStream is true. No?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -86,17 +183,42 @@ public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<by
...
                 }
+
+                if (endOfStream && !BrotliNative.BrotliDecoderIsFinished(state.BrotliNativeState))
+                {
+                    errorDetected = true;
```
ahsonkhan: Similarly, I don't think we need this. You can just use `!BrotliNative.BrotliDecoderIsFinished(state.BrotliNativeState))` at line 209. Let me know if I am mistaken.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -86,17 +183,42 @@ public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<by
...
+                }
+                if (LastDecoderResult == BrotliDecoderResult.Error || errorDetected)
+                {
+                    var error = BrotliNative.BrotliDecoderGetErrorCode(state.BrotliNativeState);
+                    var text = BrotliNative.BrotliDecoderErrorString(error);
+                    throw new System.IO.IOException(text + BrotliEx.unableDecode);
+                }
+                if (endOfStream && !BrotliNative.BrotliDecoderIsFinished(state.BrotliNativeState) && LastDecoderResult == BrotliDecoderResult.NeedsMoreInput)
```
ahsonkhan: I don't know if this will ever be true.
if (x && y && LastDecoderResult == BrotliDecoderResult.NeedsMoreInput) => false if `LastDecoderResult == BrotliDecoderResult.NeedsMoreInput` is false, and we only get here if that is the case.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -42,13 +43,10 @@ public partial class BrotliStream : Stream
 
         public BrotliStream(Stream baseStream, CompressionMode mode, bool leaveOpen, int bufferSize, CompressionLevel quality) : this(baseStream, mode, leaveOpen, bufferSize)
         {
-            if (_mode == CompressionMode.Decompress)
-            {
-                throw new System.IO.IOException(BrotliEx.QualityAndWinSize);
-            }
+            if (_mode == CompressionMode.Decompress) throw new System.IO.IOException(BrotliEx.QualityAndWinSize);
```
ahsonkhan: nit: inconsistent use of brace brackets.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -74,24 +72,24 @@ public BrotliStream(Stream baseStream, CompressionMode mode, bool leaveOpen = fa
             _mode = mode;
             _stream = baseStream;
             _leaveOpen = leaveOpen;
+            _state = new Brotli.State();
             if (_mode == CompressionMode.Compress)
             {
-                _encoder = new Encoder();
-                _encoder.SetQuality();
-                _encoder.SetWindow();
+                _state.InitializeEncoder();
```
ahsonkhan: InitializeEncoder() can throw. Should we catch the exception somewhere?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -229,16 +207,29 @@ private void EnsureNotDisposed()
                 throw new ObjectDisposedException(BrotliEx.StreamDisposed);
         }
 
+        internal void RemoveBytes(int numberOfBytes)
+        {
+            ArraySegment<byte> buf;
+            if (_bufferStream.TryGetBuffer(out buf))
```
ahsonkhan: nit: you can inline the type, i.e. `out ArraySegment<byte> buf`

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -229,16 +207,29 @@ private void EnsureNotDisposed()
...
+            {
+                Buffer.BlockCopy(buf.Array, numberOfBytes, buf.Array, 0, (int)_bufferStream.Length - numberOfBytes);
```
ahsonkhan: nit: is the int cast necessary?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -229,16 +207,29 @@ private void EnsureNotDisposed()
...
+                _bufferStream.SetLength(BufferStream.Length - numberOfBytes);
+            }
+            else
+            {
+                throw new UnauthorizedAccessException();
```
ahsonkhan: Why do we throw this exception here?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -247,68 +238,42 @@ public override int Read(byte[] buffer, int offset, int count)
...
-                        ref _availableOutput, ref _nextOutput, out totalCount);
+                    transformationResult = Brotli.Decompress(_nextInput, _bufferOutput, out _availableInput, out _availableOutput, ref _state);
                     if (bytesRead >= count) break;
                 }
-                if (endOfStream && !BrotliNative.BrotliDecoderIsFinished(_decoder.State))
-                {
-                    errorDetected = true;
-                }
-                if (_decoder.LastDecoderResult == BrotliDecoderResult.Error || errorDetected)
+                if (transformationResult==TransformationStatus.Done)
```
ahsonkhan: nit: spacing between operator/operands.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -346,27 +310,19 @@ public override void Write(byte[] buffer, int offset, int count)
                     throw new TimeoutException(BrotliEx.TimeoutWrite);
                 }
                 copyLen = bytesRemain > _bufferSize ? _bufferSize : bytesRemain;
-                Marshal.Copy(buffer, currentOffset, _bufferInput, copyLen);
+                byte[] _bufferInput = new byte[copyLen];
```
ahsonkhan: nit: coding style - variable name.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: really it should be because we send this variables to the native library (where they means array size) 
(see written and consumed)

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: because bytesWritten after calling a native func is: "how many bytes have already free", and we convert it to how many bytes have already written

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: Yes, it's true and it should works like this

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: Because at start we haven't send anything for decompress, so we need more input

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: it is no need to inline it. Do you mean that inline type will work faster?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: Because it means something wrong with memory allocation or smth else now. I think it is not necessary  exception, but useful

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: thanks

Vedin: @shiftylogic @ahsonkhan Also I need an advice about this variable. I now think about remove this bufferInput and just use buffer instead, but in this case offset and count make no sense.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: endOfStream was deleted

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: changed

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: make sense, nice optimization

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: removed

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: fixed

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: Hmm... Yes I think

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: fixed here and in some other places

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
shiftylogic: I think you are going to have to hold enough information that allows you to distinguish between whether you are doing compress or decompress, then call the correct de-allocation function.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -15,8 +15,67 @@ namespace System.IO.Compression
...
+        {
+            internal IntPtr BrotliNativeState { get; private set; }
+
+            public void Dispose()
+            {
+                //    BrotliNative.BrotliDecoderDestroyInstance(BrotliNativeState);
+                //   BrotliNative.BrotliEncoderDestroyInstance(BrotliNativeState);
+            }
+
+            public void InitializeDecoder()
```
shiftylogic: I suggest a single initialize function with a flag that says whether they want compress or decompress. Also, I'd make the initialization protect against a second call. If you don't you will leak native memory.

shiftylogic: You might also consider making these internal. State is a struct, so have the default object be "uninitialized" then have the state initialize itself on first use. That makes the developer have to so less work and won't get it wrong.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
shiftylogic: In the stream case, I would say that an exception bubbling up is probably a good thing, unless you plan on wrapping and re-throwing.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -138,26 +139,20 @@ public override long Position
...
-            if (_encoder.State == IntPtr.Zero) return;
-            if (BrotliNative.BrotliEncoderIsFinished(_encoder.State)) return;
-            BrotliEncoderOperation op = finished ? BrotliEncoderOperation.Finish : BrotliEncoderOperation.Flush;
-            nuint totalOut = 0;
-            while (true)
+            if (_state.BrotliNativeState == IntPtr.Zero) return;
+            if (BrotliNative.BrotliEncoderIsFinished(_state.BrotliNativeState)) return;
+            _nextInput = new byte[0];
+            _availableInput = 0;
+            TransformationStatus ts = TransformationStatus.DestinationTooSmall;
```
shiftylogic: Rename ts to something more meaningful. Even "status" would be better.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
shiftylogic: That exception is used for security checks. It is not appropriate here.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -346,27 +315,19 @@ public override void Write(byte[] buffer, int offset, int count)
                     throw new TimeoutException(BrotliEx.TimeoutWrite);
                 }
                 copyLen = bytesRemain > _bufferSize ? _bufferSize : bytesRemain;
-                Marshal.Copy(buffer, currentOffset, _bufferInput, copyLen);
+                byte[] bufferInput = new byte[copyLen];
```
shiftylogic: Why is this extra buffer allocation and copy necessary? If you are going to need an extra buffer on each write call, you might want to rent a buffer from a pool, have your own buffer that you re-use, or use a stack buffer (if it isn't too large).

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: This method will be fully deleted))

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: Not necessary It is temporary. What the problem is that signature of Compress takes out parameter (consumed) and input is exactly as source size. I can change consumed to ref parameter, but it is inconsistent with other Encode/Decode methods, which using Span

Vedin: Is something like this ok?
```
Span<byte> bufferInput = new Span<byte>(buffer);
bufferInput.Slice(0, copyLen);
transformationResult = Brotli.Compress(bufferInput, _bufferOutput, out _availableInput, out _availableOutput, ref _state);
```

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -108,8 +136,9 @@ public static TransformationStatus FlushEncoder(Span<byte> source, Span<byte> de
             BrotliEncoderOperation operation = isFinished ? BrotliEncoderOperation.Finish : BrotliEncoderOperation.Flush;
             bytesConsumed = source.Length;
             bytesWritten = destination.Length;
+            bytesConsumed = 0;
```
ahsonkhan: Why set bytesConsumed to source.Length first and then to 0?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -317,17 +308,18 @@ public override void Write(byte[] buffer, int offset, int count)
...
                 byte[] bufferInput = new byte[copyLen];
                 Array.Copy(buffer, currentOffset, bufferInput, 0, copyLen);
-                bytesRemain -= copyLen;
-                currentOffset += copyLen;
-                _availableInput = copyLen;
-                _availableOutput = _bufferSize;
                 TransformationStatus transformationResult = TransformationStatus.DestinationTooSmall;
-                while (transformationResult == TransformationStatus.DestinationTooSmall)
+                /*Span<byte> bufferInput = new Span<byte>(buffer);
+                bufferInput.Slice(0, copyLen);*/
```
ahsonkhan: nit: remove commented out code?

tests/System.IO.Compression.Tests/BrotliPrimitivesTests.cs
```diff
@@ -24,24 +24,33 @@ public void RoundtripCompressDecompress(int totalSize)
...
-            compressed = compressed.Slice(0, written);
-            ValidateCompressedData(compressed, data);
+            Assert.Equal(consumed, 0);
+            byte[] resultCompressed = new byte[written];
+            Array.Copy(compressed, resultCompressed, written);
+            ValidateCompressedData(resultCompressed, data);
         }
 
-        private void ValidateCompressedData(Span<byte> data, byte[] expected)
+        private void ValidateCompressedData(byte[] data, byte[] expected)
```
ahsonkhan: Why was this change required?

Vedin: Just to understand that I send byte array not span, in fact it won't change anything

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: It was a question to @shiftylogic Can I use smth like this instead of allocate and copy array?

tests/System.IO.Compression.Tests/BrotliPerfomanceTests.cs
```diff
@@ -75,9 +75,17 @@ public void Decompress(Util.CompressionType type)
             byte[] data = File.ReadAllBytes(testFilePath);
             var bytes = new byte[bufferSize];
             foreach (var iteration in Benchmark.Iterations)
+            {
```
ahsonkhan: FYI, this filename has a typo:
BrotliPerfomanceTests.cs => BrotliPerformanceTests.cs

Vedin: wow, really

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: The problem is that in encoder mode developer also may want to set quality and window size, in this case He needs to have a state before first call. Possible solution is add another Compress method with quality and window size, but it doesn't look obvious for user to call one method at first iteration and other on next iterations, does it?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: deleted

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: renamed to flushStatus

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -15,8 +15,95 @@ namespace System.IO.Compression
...
+        private const int MinWindowBits = 10;
+        private const int MaxWindowBits = 24;
+        private const int MinQuality = 0;
+        private const int MaxQuality = 11;
+
+        public struct State : IDisposable
+        {
+            internal IntPtr BrotliNativeState { get; private set; }
+            internal BrotliDecoderResult LastDecoderResult;
+            public bool CompressMode { get; set; }
```
shiftylogic: This should probably be readonly and set internally when InitializeDecoder or InitializeEncoder is called.

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -346,27 +292,20 @@ public override void Write(byte[] buffer, int offset, int count)
                     throw new TimeoutException(BrotliEx.TimeoutWrite);
                 }
                 copyLen = bytesRemain > _bufferSize ? _bufferSize : bytesRemain;
-                Marshal.Copy(buffer, currentOffset, _bufferInput, copyLen);
-                bytesRemain -= copyLen;
-                currentOffset += copyLen;
-                _availableInput = (nuint)copyLen;
-                _nextInput = _bufferInput;
-                while ((int)_availableInput > 0)
+                byte[] bufferInput = new byte[copyLen];
```
shiftylogic: Why is this extra array and copy needed? Why can't you use the input directly for the compress call?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -138,26 +154,21 @@ public override long Position
...
         protected virtual void FlushEncoder(bool finished)
         {
-            if (_encoder.State == IntPtr.Zero) return;
-            if (BrotliNative.BrotliEncoderIsFinished(_encoder.State)) return;
-            BrotliEncoderOperation op = finished ? BrotliEncoderOperation.Finish : BrotliEncoderOperation.Flush;
-            nuint totalOut = 0;
-            while (true)
+            if (_state.BrotliNativeState == IntPtr.Zero) return;
+            if (BrotliNative.BrotliEncoderIsFinished(_state.BrotliNativeState)) return;
+            _nextInput = new byte[0];
```
shiftylogic: Why allocate an empty array and throw away the one you allocated in the constructor?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -167,7 +181,14 @@ public static TransformationStatus FlushEncoder(Span<byte> source, Span<byte> de
 
         public static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref State state)
         {
-            EnsureInitialized(ref state, true);
+            try
```
shiftylogic: This try catch and throw is not necessary if all you are doing it re-throwing the same exception. Just let it bubble up to the caller.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -200,7 +221,14 @@ public static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<byte
 
         public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref State state)
         {
-            EnsureInitialized(ref state, false);
+            try
```
shiftylogic: Same here. Just let this bubble up to the caller.

Vedin:

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: yes, fixed it

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: I rewrote the logic of detecting errors, but you are right

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -4,15 +4,116 @@
...
+                BrotliNative.BrotliEncoderSetParameter(BrotliNativeState, BrotliEncoderParameter.LGWin, window);
+            }
+
+            public void SetWindow()
+            {
+                SetWindow(MaxWindowBits);
+            }
+        }
+
+        public static void EnsureInitialized(ref State state, bool compress)
```
shiftylogic: Probably want this internal, unless you think someone needs this for external use.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -40,16 +141,14 @@ private static TransformationStatus GetTransformationStatusFromBrotliDecoderResu
             return TransformationStatus.InvalidData;
         }
 
-        public static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, CompressionLevel quality = (CompressionLevel)DefaultQuality, int windowSize = DefaultWindowSize, BrotliEncoderMode encMode = BrotliEncoderMode.Generic)
+        public static TransformationStatus FlushEncoder(Span<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref State state, bool isFinished = true)
```
ahsonkhan: Why change the input source to Span instead of ReadOnlySpan?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -82,17 +223,34 @@ public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<by
...
-                    if (res == BrotliDecoderResult.Success)
-                    {
-                        bytesWritten = (int)written;
-                        bytesConsumed = (int)consumed;
-                    }
-                    return GetTransformationStatusFromBrotliDecoderResult(res);
+                    nuint availableOutput = (nuint)bytesWritten;
+                    nuint consumed = (nuint)bytesConsumed;
+                    state.LastDecoderResult = BrotliNative.BrotliDecoderDecompressStream(state.BrotliNativeState, ref consumed, ref bufIn, ref availableOutput, ref bufOut, out nuint totalOut);
+                    bytesWritten = (int)((nuint)destination.Length - availableOutput);
```
ahsonkhan: This is simpler. Would it work?
`bytesWritten = destination.Length - (int)availableOutput;`

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -1,34 +1,34 @@
...
-using nuint = System.UInt32;
+    using nuint = System.UInt32;
 #endif 
 
 namespace System.IO.Compression
 {
     public partial class BrotliStream : Stream
     {
-        private const int DefaultBufferSize = (1 << 16) - 1;
+        private const int DefaultBufferSize = (1 << 16) - 16;
```
ahsonkhan: Can you add the decimal value of DefaultBufferSize as a comment?

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
```diff
@@ -134,26 +123,19 @@ public override long Position
...
+            if (_state.BrotliNativeState == IntPtr.Zero) return;
+            if (BrotliNative.BrotliEncoderIsFinished(_state.BrotliNativeState)) return;
+            TransformationStatus flushStatus = TransformationStatus.DestinationTooSmall;
+            while (flushStatus == TransformationStatus.DestinationTooSmall)
             {
-                if (!BrotliNative.BrotliEncoderCompressStream(_encoder.State, op, ref _availableInput, ref _nextInput, ref _availableOutput, ref _nextOutput, out totalOut))
-                    throw new System.IO.IOException(BrotliEx.unableEncode);
-                var extraData = (nuint)_availableOutput != (nuint)_bufferSize;
-                if (extraData)
+                flushStatus = Brotli.FlushEncoder(Array.Empty<byte>(), _buffer, out _availableInput, out _availableOutput, ref _state, finished);
```
ahsonkhan: Out of curiosity, why do we pass an empty array here?

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: It is different methods.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
Vedin: Yes, I should change it, I just think that it is better to do arithmetic operations in the narrowed type, than expand it to int and than do minus

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
Vedin: Because, when we start flush stream, we shouldn't send any more date to the original native lib

Vedin:

Vedin:

src/System.IO.Compression.Brotli/System/IO/Compression/BrotliStream.cs
ahsonkhan: Why not Span.Empty?

Vedin: Hmm...Does Span.Empty work faster? I just thought it is the same

ahsonkhan: > Hmm...Does Span.Empty work faster? I just thought it is the same

I am not certain (will have to look at assembly). However, FlushEncoder takes a span, and if Span.Empty works, I would use that. If there is a specific reason to keep Array.Empty instead and use the implicit cast to Span, then use array. I believe you avoid the implicit cast: https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Span.cs#L376

shiftylogic: Span<byte>.Empty should be faster because it avoids the implicit cast and thus the extra ctor call.

PR #1622, opened 2017-06-21
Encoder Design Change

shiftylogic:
NOTE: This does not change any of the names from our proposal based on the review that happened. The class / method names are not finalized, so I left them as they were in our proposal until we lock them down.


shiftylogic:
Couple notes:
* The actual encoder implementations are not that different from what we had. However, getting them to line up so there was less noise in the review was difficult. The core difference is that they return a status enum instead of bool and that means the code needed to be tweaked some.
* The Transformation classes will go internal (or be removed) based on what we decide.
* Doing this change, I found some usages of SymbolTable are a bit more code than should be necessary. I added an extension method for instance to handle "string" for the user, but it involves an allocation and extra conversion in some cases. We have to find somewhere that these helper extensions will live.

@KrzysztofCwalina @ahsonkhan

shiftylogic:
Also, there are some ASCII specific functions that need to be moved still. Still trying to figure out where they will go.

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -184,47 +184,48 @@ public bool TryFormat(Span<byte> buffer, out int written, TextFormat format, Tex
...
-            ReadOnlySpan<char> charSpan = new ReadOnlySpan<char>(&value, 1);
-            
-            int consumed;
-            return formattingData.TryEncode(charSpan, buffer, out consumed, out written);
+            return symbolTable.TryEncode(value, buffer, out written);
         }
 
         struct JsonValue : IBufferFormattable
         {
+            static readonly byte[] NullBytes = { (byte)'n', (byte)'u', (byte)'l', (byte)'l' };
```
KrzysztofCwalina: I think we use s_xxxYyy as the naming pattern for private static fileds

src/System.Text.Json/System/Text/Json/JsonConstants.cs
```diff
@@ -30,9 +30,9 @@ static class JsonConstants
 
         #region Common values
 
-        public static readonly char[] TrueValue = { 't', 'r', 'u', 'e' };
-        public static readonly char[] FalseValue = { 'f', 'a', 'l', 's', 'e' };
-        public static readonly char[] NullValue = { 'n', 'u', 'l', 'l' };
+        public static readonly byte[] TrueValue = { (byte)'t', (byte)'r', (byte)'u', (byte)'e' };
+        public static readonly byte[] FalseValue = { (byte)'f', (byte)'a', (byte)'l', (byte)'s', (byte)'e' };
+        public static readonly byte[] NullValue = { (byte)'n', (byte)'u', (byte)'l', (byte)'l' };
```
KrzysztofCwalina: It seems like we are repeating these constants from above. Can we have just one copy?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -420,42 +419,59 @@ private void WriteGuid(Guid value)
...
+                    if (status == Buffers.TransformationStatus.Done)
+                    {
+                        _output.Advance(written);
+                        return;
+                    }
+
+                    if (status == Buffers.TransformationStatus.DestinationTooSmall)
+                    {
+                        destination = EnsureBuffer();
+                        continue;
```
KrzysztofCwalina: The loop just wraps here. But I think we partially encode when the destination is too small. Can we somehow not completly redo the encoding? Would it be perf positive?

src/System.Text.Json/System/Text/Json/JsonWriter.cs
```diff
@@ -420,42 +419,59 @@ private void WriteGuid(Guid value)
...
+                }
             }
             else if (UseFastUtf16)
             {
-                int needed = values.Length * sizeof(char);
-                var buffer = EnsureBuffer(needed);
-
-                Span<char> span = buffer.NonPortableCast<byte, char>();
-                values.CopyTo(span);
+                Span<byte> destination = EnsureBuffer(source.Length);
```
KrzysztofCwalina: I think there are some buffers that might not be able to resize to an arbitrary size. i.e. are we sure that EnsureBuffer guarantees the size? I though most of the Ensure calls try the best. Anyway, we need to decide on the contract of "resize" and add tests for it.

src/System.Text.Primitives/System/Text/Encoding/Encoders.Utf16.cs
```diff
@@ -0,0 +1,790 @@
...
+                            availableBytes = availableChars;
+
+                        // don't fall into the fast decoding loop if we don't have enough bytes
+                        if (availableBytes <= 13)
+                        {
+                            // try to get over the remainder of the ascii characters fast though
+                            byte* pLocalEnd = pSrc + availableBytes;
+                            while (pSrc < pLocalEnd)
+                            {
+                                ch = *pSrc;
```
KrzysztofCwalina: Please file an issue to explore using SIMD widening operations to implement the ASCII to UTF16 part. Thanks!

src/System.Text.Primitives/System/Text/Encoding/Encoders.Utf8.cs
```diff
@@ -0,0 +1,481 @@
...
+                    }
+
+                    if ((utf16Length << 1) != source.Length)
+                        return TransformationStatus.NeedMoreSourceData;
+
+                    return TransformationStatus.Done;
+                }
+                catch (ArgumentOutOfRangeException)
+                {
+                    return TransformationStatus.InvalidData;
```
KrzysztofCwalina: BTW, yesterday I was discussing some File APIs operating on spans and buffers with @stephentoub. It seems like we will want to use something like TranstromationStatus for such file i/o APIs. But file transfer is not really a transformation. Maybe we could come up with a better name for the enum. "OperationStatus" seems to generic. I would love a name that says "an operation on data in a buffer", but not a super long name.

src/System.Text.Primitives/System/Text/Encoding/EncodingHelper.cs
```diff
@@ -0,0 +1,151 @@
...
+        public const byte b1110_0000U = 0xE0; //224
+        public const byte b1111_0000U = 0xF0; //240
+        public const byte b1111_1000U = 0xF8; //248
+
+        #endregion Constants
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static bool IsWhitespace(uint codePoint)
+        {
+            return Array.BinarySearch<uint>(SortedWhitespaceCodePoints, codePoint) >= 0;
```
KrzysztofCwalina: I wonder is SIMD would not be good here.

src/System.Text.Primitives/System/Text/Encoding/SymbolTable.Symbol.cs
```diff
@@ -0,0 +1,28 @@
...
+            D7 = (ushort)7,
+            D8 = (ushort)8,
+            D9 = (ushort)9,
+            DecimalSeparator = (ushort)10,
+            Exponent = (ushort)16,
+            GroupSeparator = (ushort)11,
+            InfinitySign = (ushort)12,
+            MinusSign = (ushort)13,
+            NaN = (ushort)15,
+            PlusSign = (ushort)14,
```
KrzysztofCwalina: Should we add some common JSON symbols to the table? e.g. '}", '[", etc.? Would it speed up our JSON writer?

src/System.Text.Primitives/System/Text/Encoding/SymbolTable.Utf8.cs
```diff
@@ -0,0 +1,100 @@
...
+                new byte[] { 78, 97, 78, }, // NaN
+                new byte[] { 69, }, // E
+                new byte[] { 101, }, // e
+            };
+
+            public Utf8InvariantSymbolTable() : base(Utf8DigitsAndSymbols) {}
+
+            public override bool TryEncode(byte utf8, Span<byte> destination, out int bytesWritten)
+            {
+                if (destination.Length < 1)
```
KrzysztofCwalina: I wonder if we would not get better codegen if:
1. First, we did a check for the utf8 range.
2. Then we did a check for length, and immediately after wrote to the span.

The reason is that span does the exact same length check, which might be easier to eliminate if the checks endup being one after the other.

src/System.Text.Primitives/System/Text/Formatting/FloatFormatter.cs
```diff
@@ -44,7 +44,42 @@ public static bool TryFormatNumber(double value, bool isSingle, Span<byte> buffe
...
+                // TODO: This is currently pretty expensive. Can this be done more efficiently?
+                //       Note: removing the hack might solve this problem a very different way.
+                var status = Encoders.Utf8.ComputeEncodedBytesFromUtf16(utf16Bytes, out int needed);
+                if (status != Buffers.TransformationStatus.Done)
+                {
+                    bytesWritten = 0;
+                    return false;
+                }
+
+                Span<byte> temp = new byte[needed];
```
KrzysztofCwalina: Should we stack alloc this?

KrzysztofCwalina:
Thanks! It looks good. Feel free to merge and just file issues for the comments. Some of the correctness issues (resizing buffers) should be tracked carefully and fixed soon. The rest can probably be done later.

src/System.Text.Json/System/Text/Json/JsonConstants.cs
shiftylogic: Difference projects. One is further along than the other. I assume if we were ever to productize, we would merge the 2 projects together and fix the issue at that time.

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: Issue #1625

src/System.Text.Json/System/Text/Json/JsonWriter.cs
shiftylogic: The EnsureBuffer function validates that the resize expanded by at least what we asked for. It throws if not. Possible we want to re-visit this if/when we finalize the JSON story.

shiftylogic: #1626

src/System.Text.Primitives/System/Text/Encoding/Encoders.Utf16.cs
shiftylogic: #1627

src/System.Text.Primitives/System/Text/Encoding/EncodingHelper.cs
shiftylogic: #1628

src/System.Text.Primitives/System/Text/Encoding/SymbolTable.Symbol.cs
shiftylogic: The JSON processor is not using the symbol table for these scenarios. The branching checks for determining where we are in a JSON sequence is done directly by byte / char comparisons. I'm pretty sure that going through the SymbolTable would actually slow it down quite a bit.

src/System.Text.Primitives/System/Text/Encoding/SymbolTable.Utf8.cs
shiftylogic: These functions need optimization work. This was just a re-structure. We can address this during an optimization pass. Filed #1629 to capture general symbol table optimizations.

src/System.Text.Primitives/System/Text/Formatting/FloatFormatter.cs
shiftylogic: Good call. Given this is guaranteed to be small (under 64 bytes), I made that change.

PR #1623, opened 2017-06-22
Fixed a small bug in Trie creation logic

KrzysztofCwalina:
List.Sort seems to be comparing the same item (as this and other) sometimes during List.Sort
This means that the tie validation logic (which was done in the comparer) could sometimes fail.
This PR moves the validation after the array is sorted with a comparer that never fails.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoding/Internal/ParsingTrie.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Internal/ParsingTrie.cs
@@ -304,7 +304,7 @@ namespace System.Text
             //   a) each symbol must be unique
             //   b) a symbol cannot be a prefix of another symbol
             //   c) symbols cannot be empty
-            for(int i=1; i<symbolList.Count; i++)
+            for(int i = 1; i < symbolList.Count; i++)
             {
                 var first = symbolList[i - 1];
                 var second = symbolList[i];
--- a/src/System.Text.Primitives/System/Text/Encoding/Internal/ParsingTrie.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Internal/ParsingTrie.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Collections;
 using System.Collections.Generic;
 
 namespace System.Text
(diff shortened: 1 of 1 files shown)


KrzysztofCwalina:
cc: @shiftylogic

src/System.Text.Primitives/System/Text/Encoding/Internal/ParsingTrie.cs
```diff
@@ -95,32 +102,14 @@ public Suffix(int symbolIndex, ReadOnlySpan<byte> bytes)
 
             public int CompareTo(Suffix other)
             {
-                int index = 0;
-                if (Bytes.Length == 0 || other.Bytes.Length == 0)
+                var shorter = Math.Min(other.Bytes.Length, Bytes.Length);
+                for(int index=0; index<shorter; index++)
```
ahsonkhan: nit: spacing

src/System.Text.Primitives/System/Text/Encoding/Internal/ParsingTrie.cs
```diff
@@ -311,6 +300,40 @@ public static Node[] Create(byte[][] symbols)
...
+                var second = symbolList[i];
+
+                if(first.Bytes.Length == 0 || second.Bytes.Length == 0)
+                {
+                    throw new ArgumentException("Symbol cannot be zero bytes long");
+                }
+                var firstSpan = first.Bytes.AsSpan();
+                if (firstSpan.SequenceEqual(second.Bytes))
+                {
+                    throw new ArgumentException(String.Format("Symbols cannot be identical"));
```
ahsonkhan: Is the String.Format necessary here?

src/System.Text.Primitives/System/Text/Encoding/Internal/ParsingTrie.cs
```diff
@@ -311,10 +300,44 @@ public static Node[] Create(byte[][] symbols)
             // counting the number of children a node has.
             symbolList.Sort();
 
+            // validate symbol consistemcy:
+            //   a) each symbol must be unique
+            //   b) a symbol cannot be a prefix of another symbol
+            //   c) symbols cannot be empty
+            for(int i=1; i<symbolList.Count; i++)
```
ahsonkhan: nit: I forgot to comment here, same spacing.

Also, should we add a test to confirm the fix?

PR #1633, opened 2017-06-27
Brotli web tests

Vedin:
Retest performance on Canterbury set of files after API rewriting and add set of different web files. 

As in previous performance tests (https://github.com/dotnet/corefxlab/pull/1618) The result below is in how many times GzipStream is faster than Brotl (red(GZIP faster)  to green(Brotli faster)). Light Blue color of test name means that it is web file. 

Compress:
image

Decompress
image

And Compression Ratio(green means Brotli better):
image

Also we need to do something with CompressionLevel, because now we can't have clear result instead of difference of it. In GZip NoCompression is literally no compression in my case it is the lowest possible compression. So know we can really compare Fastest and Optimal level. 

So, as expected compression ratio is higher, decompress speed is equal on small files and higher on large, compress speed is slower in most cases. But I really don't understand how Brotli Compress can work faster on NoCompression level.


tests/System.IO.Compression.Tests/System.IO.Compression.Tests.csproj
```diff
@@ -27,15 +27,16 @@
     </Content>
   </ItemGroup>
   <ItemGroup>
+    <Compile Remove="Resources\**" />
```
ahsonkhan: Why do we need this itemgroup?

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
```diff
@@ -19,19 +19,36 @@ public static IEnumerable<object[]> CanterburyCorpus()
         {
             foreach (CompressionLevel compressionLevel in Enum.GetValues(typeof(CompressionLevel)))
             {
+                string folder = "Canterbury";
```
ahsonkhan: nit: folder doesn't change in this loop. Declare two separate string variables (one for Canterbury folder and one for WebFiles) and move the declarations outside the foreach?

tests/System.IO.Compression.Tests/System.IO.Compression.Tests.csproj
Vedin: You said me to delete Resources, so VS adds it automatically, when I do it

tests/System.IO.Compression.Tests/BrotliPerformanceCanterburyCorpus.cs
Vedin: done

ahsonkhan:

PR #1635, opened 2017-06-27
Fixed the build

KrzysztofCwalina:
The most recent CLI tools update started to cause some problems with our project structure.
This fixes the issues and makes it possible to build cofxlab from the command line.

### DIFF
--- a/corefxlab.sln
+++ b/corefxlab.sln
@@ -1,6 +1,6 @@
 Microsoft Visual Studio Solution File, Format Version 12.00
 # Visual Studio 15
-VisualStudioVersion = 15.0.26507.0
+VisualStudioVersion = 15.0.26602.0
 MinimumVisualStudioVersion = 10.0.40219.1
 Project("{2150E333-8FDC-42A3-9474-1A3956D46DE8}") = "Solution Items", "Solution Items", "{5E7EB061-B9BC-4DA2-B5E5-859AA7C67695}"
 	ProjectSection(SolutionItems) = preProject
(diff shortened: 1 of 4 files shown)


shiftylogic:
Aside from adding 2 new projects to the solution, this change appears to be benign. I'm curious how the projects before were causing problems building. Seems like a bug in the CLI that should be reported.

KrzysztofCwalina:
@piotrpMSFT, it seems like the CLI tools we picked up sometimes don't properly handle package reference in separate item groups.

src/System.IO.Pipelines.Networking.Sockets/System.IO.Pipelines.Networking.Sockets.csproj
```diff
@@ -7,8 +7,6 @@
   </PropertyGroup>
   <ItemGroup>
     <ProjectReference Include="..\System.IO.Pipelines\System.IO.Pipelines.csproj" />
-  </ItemGroup>
```
ahsonkhan: I am surprised that these changes are necessary. If they are, shouldn't they be applied to all projects? For example, Buffers.Experimental.csproj?

corefxlab.sln
```diff
@@ -144,6 +144,10 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Text.Utf8String", "s
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Text.Utf8String.Tests", "tests\System.Text.Utf8String.Tests\System.Text.Utf8String.Tests.csproj", "{C6F061D6-18D6-41E3-A692-C0603A994F30}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Server", "samples\SimpleHttp\Server.csproj", "{5B4ACAF4-9ECB-431E-A37B-960A224A9B40}"
```
ahsonkhan: If we are adding these projects to the solution, we should add
`<Import Project="..\..\tools\common.props" />`
to the csproj files to bring in the new compiler/etc.

KrzysztofCwalina:
It seems like this is not actually fixing anything. It just happened to built on my machine after these changes. some race is happening in the toolchain.

PR #1636, opened 2017-06-27
Refining text API shape

shiftylogic:
**NOTE: Add w=1 to the address bar as a query parameter and you will save yourself a lot of reviewing. The indentation / spacing changes made this review look way worse than it actually is.**

* Moving UTF-16 extension methods for SymbolTable directly onto base class.
* Removing Transforms.
* Changing Encoders static class to just be a namespace.

@ahsonkhan @KrzysztofCwalina


shiftylogic:

ahsonkhan:
> NOTE: Add w=1 to the address bar as a query parameter and you will save yourself a lot of reviewing. 

What does adding w=1 do exactly?

https://github.com/dotnet/corefxlab/pull/1636/files?w=1

ahsonkhan:
> Changing Encoders static class to just be a namespace.

Would Base64 APIs go under this namespace as well?

shiftylogic:
@ahsonkhan w=1 makes the diff ignore whitespace changes.

ahsonkhan:
> makes the diff ignore whitespace changes.

shiftylogic:
@ahsonkhan Re: Base64... I think that makes sense.

ahsonkhan:
> Removing Transforms.

Do we want to completely remove them? Can we instead move them to experimental or hide them?

shiftylogic:
@ahsonkhan Re: Transforms => There was no real code there, so removing these for now is fine. They were also not being used anywhere that mattered. The one test was a few character patch to remove.

We can trivially add them back once we figure out whether we want them and what the interface will look like.

src/System.Text.Primitives/System/Text/Encoding/SymbolTable.cs
```diff
@@ -107,6 +107,77 @@ public bool TryParse(ReadOnlySpan<byte> source, out Symbol symbol, out int bytes
 
         public abstract bool TryParse(ReadOnlySpan<byte> source, Span<byte> utf8, out int bytesConsumed, out int bytesWritten);
 
+        #region Public UTF-16 to UTF-8 helpers
+
+        public bool TryEncode(ReadOnlySpan<char> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
```
ahsonkhan: I am curious, why does this API take RoS\<char\> and not byte?

shiftylogic: That would conflict with the UTF-8 version... and this is specifically for UTF-16 strings.

ahsonkhan: OK.
Also, we changed this method from an extension method to an instance method on SymbolTable. What is the motivation for this?

shiftylogic: There was a single extension method. It made the extension class overkill. This is a common enough function that we decided it made sense to put it directly on the class.

shiftylogic:

PR #1638, opened 2017-06-27
downgrading compiler version temporarily

VSadov:
This will still have all the 7.2 features, but will not match the most recent VSIX

It seems like we brought in a race condition via the last integration and it affects this repo more than any other.


VSadov:
looks like this has no race. 
It really seems to have come with the merge from the upstream branch then. Feature-wise this build is the same as the one with problems.

PR #1639, opened 2017-06-28
Brotli base tests2

Vedin:
Small changes and some test for Brotli base class.

### DIFF
--- a/src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
+++ b/src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
@@ -13,7 +13,6 @@ namespace System.IO.Compression
     {
         private const int MinWindowBits = 10;
         private const int MaxWindowBits = 24;
-        private const int MinQuality = 0;
         private const int MaxQuality = 11;
 
         public struct State : IDisposable
@@ -24,6 +23,7 @@ namespace System.IO.Compression
 
             public void Dispose()
             {
+                if (BrotliNativeState == IntPtr.Zero) return;
                 if (CompressMode)
                 {
                     BrotliNative.BrotliEncoderDestroyInstance(BrotliNativeState);
@@ -32,6 +32,7 @@ namespace System.IO.Compression
                 {
                     BrotliNative.BrotliDecoderDestroyInstance(BrotliNativeState);
                 }
+                BrotliNativeState = IntPtr.Zero;
             }
 
             internal void InitializeDecoder()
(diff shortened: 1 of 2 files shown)


src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -117,6 +117,10 @@ internal static void EnsureInitialized(ref State state, bool compress)
 
         public static int GetMaximumCompressedSize(int inputSize)
         {
+            if (inputSize < 0 || inputSize > 2147483132) // 2^32 - 1 - 515 (max compressed extra bytes)
```
Vedin: should it be a const variable or "magic number" with comment is ok?

ahsonkhan: It should be a const with the comment as well, imo.

What would the math operation return if inputSize > 2147483132 (i.e., do we even need this check)?

Vedin: the result will be wrong, because in fact true number is bigger than int. We need this check because of 2 reasons:
1. We return int, so result should be less than int
2. Span size can't be more than MaxInt and we expected that after calling this method user will create span or array with result size;

Vedin:

tests/System.IO.Compression.Tests/BrotliPrimitivesTests.cs
```diff
@@ -10,6 +10,125 @@ public class BrotliTests
     {
         static string brTestFile(string fileName) => Path.Combine("BrotliTestData", fileName);
 
+        [Fact(Skip = "Fails in VS - System.BadImageFormatException : An attempt was made to load a program with an incorrect format.")]
```
ahsonkhan: Any ideas how to resolve these test failures in VS? We should enable these tests if possible.

src/System.IO.Compression.Brotli/System/IO/Compression/Brotli.cs
```diff
@@ -117,6 +118,10 @@ internal static void EnsureInitialized(ref State state, bool compress)
 
         public static int GetMaximumCompressedSize(int inputSize)
         {
+            if (inputSize < 0 || inputSize > MaxInputSize)
```
ahsonkhan: Aside from integer overflow when doing:
```C#
int result = inputSize + overhead;
```
when will `(result < inputSize)` ever be true in `return (result < inputSize) ? inputSize : result;`?

And since you are now catching the overflow case here (`if (inputSize < 0 || inputSize > MaxInputSize)`), do we need to check if result will be less than inputSize?

tests/System.IO.Compression.Tests/BrotliPrimitivesTests.cs
Vedin: Yes, I want to try one thing, but you know I have some trouble with VS and VS team haven't answered me yet.

PR #1641, opened 2017-06-28
Added 'l' format to DateTime formatting

KrzysztofCwalina:
This format is needed to produce Azure authentication tokens. When the tokens are created, they need lowercase RFC1123 representation of the current time.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.Time.cs
+++ b/src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.Time.cs
@@ -19,14 +19,15 @@ namespace System.Text
                 offset = value.Offset;
             }
 
-            Precondition.Require(format.Symbol == 'R' || format.Symbol == 'O' || format.Symbol == 'G');
-
             symbolTable = symbolTable ?? SymbolTable.InvariantUtf8;
 
             switch (format.Symbol)
             {
                 case 'R':
                     return TryFormatDateTimeRfc1123(value.UtcDateTime, buffer, out bytesWritten, symbolTable);
+
+                case 'l':
+                    return TryFormatDateTimeRfc1123Lowercase(value.UtcDateTime, buffer, out bytesWritten, symbolTable);
 
                 case 'O':
                     return TryFormatDateTimeFormatO(value.DateTime, value.Offset, buffer, out bytesWritten, symbolTable);
@@ -46,14 +47,15 @@ namespace System.Text
                 format.Symbol = 'G';
             }
 
-            Precondition.Require(format.Symbol == 'R' || format.Symbol == 'O' || format.Symbol == 'G');
-
             symbolTable = symbolTable ?? SymbolTable.InvariantUtf8;
 
             switch (format.Symbol)
             {
                 case 'R':
                     return TryFormatDateTimeRfc1123(value, buffer, out bytesWritten, symbolTable);
+
+                case 'l':
+                    return TryFormatDateTimeRfc1123Lowercase(value, buffer, out bytesWritten, symbolTable);
 
                 case 'O':
                     return TryFormatDateTimeFormatO(value, NullOffset, buffer, out bytesWritten, symbolTable);
@@ -112,6 +114,18 @@ namespace System.Text
                 return InvariantUtf8TimeFormatter.TryFormatRfc1123(value, buffer, out bytesWritten);
             else if (symbolTable == SymbolTable.InvariantUtf16)
                 return InvariantUtf16TimeFormatter.TryFormatRfc1123(value, buffer, out bytesWritten);
+            else
+                throw new NotImplementedException();
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        static bool TryFormatDateTimeRfc1123Lowercase(DateTime value, Span<byte> buffer, out int bytesWritten, SymbolTable symbolTable)
+        {
+            // for now it only works for invariant culture
+            if (symbolTable == SymbolTable.InvariantUtf8)
+                return InvariantUtf8TimeFormatter.TryFormatRfc1123Lowercase(value, buffer, out bytesWritten);
+            else if (symbolTable == SymbolTable.InvariantUtf16)
+                return InvariantUtf16TimeFormatter.TryFormatRfc1123Lowercase(value, buffer, out bytesWritten);
             else
                 throw new NotImplementedException();
         }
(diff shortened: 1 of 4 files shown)


KrzysztofCwalina:

tests/System.Text.Primitives.Tests/Formatting/TimeTests.cs
```diff
@@ -51,6 +53,7 @@ public void RandomDateTimeTests(char format)
         static void TestDateTimeFormat(DateTime dt, char format, SymbolTable symbolTable)
         {
             var expected = dt.ToString(format.ToString(), CultureInfo.InvariantCulture);
+            if (format == 'r') expected = expected.ToLowerInvariant();
```
ahsonkhan: Is that the difference between 'r' and 'R'?
I thought both 'r' and 'R' produce the same result.

```C#
DateTime date1 = new DateTime(2010, 8, 18);
Console.WriteLine(date1.ToString("r", CultureInfo.InvariantCulture)); \\ Output: Wed, 18 Aug 2010 00:00:00 GMT
Console.WriteLine(date1.ToString("R", CultureInfo.InvariantCulture)); \\ Output: Wed, 18 Aug 2010 00:00:00 GMT
```

src/System.Text.Primitives/System/Text/Formatting/PrimitiveFormatter.Time.cs
```diff
@@ -19,15 +19,16 @@ public static bool TryFormat(this DateTimeOffset value, Span<byte> buffer, out i
...
-            Precondition.Require(format.Symbol == 'R' || format.Symbol == 'O' || format.Symbol == 'G');
-
             symbolTable = symbolTable ?? SymbolTable.InvariantUtf8;
 
             switch (format.Symbol)
             {
                 case 'R':
                     return TryFormatDateTimeRfc1123(value.UtcDateTime, buffer, out bytesWritten, symbolTable);
 
+                case 'r':
```
ahsonkhan: Should this instead just be as follows (without needing the new method TryFormatDateTimeRfc1123Lowercase):

```C#
case 'R':
case 'r':
    return TryFormatDateTimeRfc1123(value.UtcDateTime, buffer, out bytesWritten, symbolTable); 
```

If not, can you clarify my misunderstanding?

KrzysztofCwalina: The whole point of this addition/PR is that for Azure authentication token I need to have a lowercased datetime. i.e. this PR adds 'r' to mean: "just like R but lowercase"

tests/System.Text.Primitives.Tests/Formatting/TimeTests.cs
KrzysztofCwalina: Hmm, I am not sure what to do here. I could use some other format char for "lowercassed R", but 'r' seems to be perfect for it. But you are right, it creates an incompatibility with what ToString does.

@tarekgh, any feedback on this? Do you think it would be reasonable to use 'r' for lowercased format? Also, why ToString('r') does not produce lowercases output?

tarekgh: 'r' and 'R' are exactly same things. both of them should produce the exact same results. why you think 'r' should produce lowercase results? these specifiers has nothing to do with upper or lower casing.

KrzysztofCwalina: 'x' and 'X' both produce hex with lower and upper case. I am not saying that ToString should change what it does with 'r'. What I am saying is that in System.Text.Primitives we need the ability to produce lowercased R format, and I thought 'r' would be best for it.

tarekgh: Lowecasing the date format is not a good idea anyway.

benaadams:
Are there any docs on it being lowercase? Even js doesn't lowercase RFC1123

tests/System.Text.Primitives.Tests/Formatting/TimeTests.cs
KrzysztofCwalina: Lowercase is required. https://docs.microsoft.com/en-us/rest/api/documentdb/access-control-on-documentdb-resources?redirectedfrom=MSDN

benaadams: From that link, it seems to want it both in regular RFC1123 (for **x-ms-date** header) and lowercase for **Authorization** header. So might be better to lowercase the string rather than generate it from date twice?

KrzysztofCwalina: Maybe, but we don't have UTF8 to lower conversion, so I cannot even try it. Moreover, I am not sure if all UTF8 to lower would be always non-resizing, so implementing it as an in-place conversion might not be possible.

benaadams: Headers and url are in practice 7-bit ascii (single byte); so it should always be a non-resizing conversion for date for `x-ms-date` to `Authorization` header. Maybe need an `asciiString` type? 😉

Tornhoof: You need that for that ``StringToSign`` right, the one with the line breaks? If Performance and allocations are important for that, you probably want to incrementally build the hash with Byte arrays directly, bypassing all the String formatting

KrzysztofCwalina: @Tornhoof, this whole package is about doing formatting directly to byte buffers (i.e. bypassing string formatting).

KrzysztofCwalina: Anyway, I changed the format to 'l' to avoid the controversy.

Tornhoof: @KrzysztofCwalina Yes, I understand that, what I was referring to:
```csharp
    string payLoad = string.Format(System.Globalization.CultureInfo.InvariantCulture, "{0}\n{1}\n{2}\n{3}\n{4}\n",  
            verb.ToLowerInvariant(),  
            resourceType.ToLowerInvariant(),  
            resourceLink,  
            date.ToLowerInvariant(),  
            ""  
    );  

    byte[] hashPayLoad = hmacSha256.ComputeHash(System.Text.Encoding.UTF8.GetBytes(payLoad));  
```
If the above piece is performance critical, you should use the incremental Hash function, as the http verbs and the type are static (can be static byte arrays), the line break is a char/byte etc. Then you don't need ``ToString("R")``, but ``ToFormattedBytes("R")``. If your use case is really driven by the snippet above, then you'll probably gain way more by actually bypassing string formatting (as you said) instead of adding a new string formatting function.

KrzysztofCwalina:
BTW, this improved the header generation routine perf by 20%

PR #1648, opened 2017-06-30
Changing IRetainable.Release return type to bool & removing IsRetainable

ahsonkhan:
- Also removing the IsRetainable property

cc @KrzysztofCwalina, @davidfowl, @shiftylogic, @stephentoub


src/System.Buffers.Experimental/ReferenceCountedBuffer.cs
```diff
@@ -17,15 +17,16 @@ public override void Retain()
...
         {
             if (!IsRetained) BuffersExperimentalThrowHelper.ThrowInvalidOperationException();
             if (Interlocked.Decrement(ref _referenceCount) <= 0) {
                 OnNoReferences();
             }
+            return IsRetained;
         }
 
-        public override bool IsRetained => _referenceCount > 0;
+        protected override bool IsRetained => _referenceCount > 0;
```
davidfowl: This change isn't right. The idea was to return the result of the atomic operation.

```C#
public override bool Release()
{
  var result = Interlocked.Decrement(ref _referenceCount);
  if (result < 0) BuffersExperimentalThrowHelper.ThrowInvalidOperationException();
  if (result == 0) {
      OnNoReferences();
      return true;
  }
  return false;
}
```

KrzysztofCwalina: @davidfowl, what do you mean by that? Are you proposing that we return the refcount instead of bool? If yes, we cannot do it. There are some lifetime management strategies/algorithms that don't relly know the exact refcount; they just know if the refcount is positive or zero.

davidfowl: I'm proposing that the ones that do ref count need to be written like what I have above so that it's correct. Ones that don't ref count I have no idea what it would look like (if we have any implementations like that besides the managed array).

KrzysztofCwalina: OK, I think I know what you mean. I was confused cause your comment is below the IsRetained property. Yes, we need to make the implementation of Release better.

shiftylogic: I see now you removed it from the IRetainable interface. What value is the property here?

src/System.Buffers.Primitives/System/Internal/OwnedArray.cs
```diff
@@ -68,20 +68,21 @@ public override void Retain()
...
 
-        public override void Release()
+        public override bool Release()
         {
             if (!IsRetained) BufferPrimitivesThrowHelper.ThrowInvalidOperationException();
             if (Interlocked.Decrement(ref _referenceCount) == 0)
             {
                 OnNoReferences();
             }
+            return IsRetained;
```
shiftylogic: This is racy and has the potential for conflict. The interlocked operation above could bring the ref count to a value above 0 (i.e. this code does not do the disposed) but by the time this line of code runs, someone else did bring it down to 0. Then, both callers get a result of true. If the one that didn't end up releasing it makes some decision based on the belief that they did (via the return value), that could cause an issue.

ahsonkhan: Yep, agreed. This needs to be addressed.

src/System.Buffers.Experimental/ReferenceCountedBuffer.cs
KrzysztofCwalina: The property is on OwnedBuffer. It needs to know its own refcount, but it does not call Release on itself.

davidfowl: Why is the property there at all. It shouldn't be protected either.

ahsonkhan: The property is used in the OwnedBuffer.Dispose method. We want to throw an exception on Dispose when ref count is greater than 0.

ahsonkhan:
@davidfowl, @shiftylogic - please review

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -106,15 +106,12 @@ static void Buffer(OwnedBuffer<byte> buffer)
         static void Pin(OwnedBuffer<byte> buffer)
         {
             var memory = buffer.Buffer;
-            Assert.False(buffer.IsRetained);
             var handle = memory.Retain(pin: true);
             unsafe
             {
                 Assert.NotEqual(0L, new IntPtr(handle.PinnedPointer).ToInt64());
             }
-            Assert.True(buffer.IsRetained);
```
KrzysztofCwalina: Why is this line removed?

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -106,15 +106,12 @@ static void Buffer(OwnedBuffer<byte> buffer)
         static void Pin(OwnedBuffer<byte> buffer)
         {
             var memory = buffer.Buffer;
-            Assert.False(buffer.IsRetained);
```
KrzysztofCwalina: After all the checks to IsRetained are gone, it seems like the tests are not testing anything.

ahsonkhan: Are you talking about this particular test or others as well, like MemoryHandleDoubleFree?

Should I remove the Pin test? It is still testing that pinning returns a non-zero pointer address and that Dispose doesn't throw.

ahsonkhan:
> Is anyone even using the IsRetained property anymore? Why do we need it?

The property is used in the OwnedBuffer.Dispose method. We want to throw an exception on Dispose when ref count is greater than 0. How else can we check for double dispose?

PR #1654, opened 2017-07-05
Made Ascii 'encoding' APIs consistent with other encodings

KrzysztofCwalina:
Renamed PrimitiveEncoding to Ascii and placed the type in System.Text.Encoders.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoding/Ascii/PrimitiveEncoder_ascii.cs
+++ b/src/System.Text.Primitives/System/Text/Encoding/Ascii/PrimitiveEncoder_ascii.cs
@@ -9,7 +9,7 @@ namespace System.Text.Encoders
     public static partial class Ascii
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static string ToUtf16String(this ReadOnlySpan<byte> bytes)
+        public static string ToUtf16String(ReadOnlySpan<byte> bytes)
         {
             var len = bytes.Length;
             if (len == 0) {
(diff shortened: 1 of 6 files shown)


src/System.Text.Primitives/System/Text/Encoding/Ascii/PrimitiveEncoder_ascii.cs
```diff
@@ -4,12 +4,12 @@
 
 using System.Runtime.CompilerServices;
 
-namespace System.Text
+namespace System.Text.Encodings
```
shiftylogic: The other encoders are in System.Text.Encoders right now. Was the usage of "Encodings" intentional?

src/System.Text.Primitives/System/Text/Encoding/Ascii/PrimitiveEncoder_ascii.cs
```diff
@@ -4,12 +4,12 @@
...
 {
-    public static partial class PrimitiveEncoder
+    public static partial class Ascii
     {
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static string DecodeAscii(this ReadOnlySpan<byte> bytes)
+        public static string ToUtf16String(this ReadOnlySpan<byte> bytes)
```
shiftylogic: This seems like an odd extension. This adds a ToUtf16String extension to any ROS<byte> instance, however, it will clearly only function if the data in that ROS<byte> happens to be ASCII. The naming (to me) would make me believe that this should also work for some other data. For example, I know my ROS<byte> is UTF-8 data, so I call this method thinking I will get a valid UTF-16 string.

shiftylogic: Maybe these shouldn't be extension methods on Span<byte> / ROS<byte>

KrzysztofCwalina: Yeah, I will remove "this"

PR #1655, opened 2017-07-05
Added Ascii.ToUpper/ToLower

KrzysztofCwalina:


shiftylogic:
LGTM

omariom:
`bytesChanged` and `processedBytes` are references. I suspect each access will be a dereference.
I would suggest making a local copy and updating the source before returning. Though not sure how much it will improve perf

PR #1658, opened 2017-07-07
Add Readme to Brotli

Vedin:


src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+            public void SetWindow();
+            public void SetWindow(uint window);
+        }
+    }
+```
+```out bytesConsumed``` - number of bytes from source, which used in this call
+
+```out bytesWritten``` - number of written to destination bytes 
+
+The ```Compress``` performs data compression. It sets ```source``` data should be compressed and return compressed data in ```destination```. ```state.CompressMode``` should be ```true``` .
```
shiftylogic: The word "set" in the context of source seems off since source is readonly.

Also, state should be moved up and described as an opaque state structure.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+
+The ```FlushEncoder``` return compressed data, which have sent to ```state``` using Compress. 
+
+The ```Decompress``` Decompresses the data in ```source``` into ```destination```. ```state.CompressMode``` should be ```false``` .
+
+```State``` uses in both compress and decompress mode to save a temporary status and data of process.
```
shiftylogic: uses => used.

Also, state should be moved up and described as an opaque state structured that cannot be re-used once you start compressing / decompression data with it.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+```CompressMode =  true``` shows that ```State``` in compress mode, ```false``` in decompress mode. 
+
+```SetQuality``` alows you to set quality of compression ```0 to 11```. The higher quality means higher compression ratio.
```
shiftylogic: alows => allows

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+```C#
+public static class Brotli {
+        public static TransformationStatus Compress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref Brotli.State state);
+        public static TransformationStatus Decompress(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref Brotli.State state);
+        public static TransformationStatus FlushEncoder(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten, ref Brotli.State state, bool isFinished=true);
+        public static int GetMaximumCompressedSize(int inputSize);
+        public struct State : IDisposable {
+            public bool CompressMode { get; }
+            public void Dispose();
+            public void SetQuality();
```
shiftylogic: A SetXxxx method without a parameter seems odd. I think this should be removed, along with the other one below.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+    Brotli.State state = new Brotli.State();
+    byte[] compressed = new byte[Brotli.GetMaximumCompressedSize(bytes.Length)];
+    state.SetQuality();
+    TransformationStatus result = Brotli.Compress(bytes, compressed, out int consumed, out int written, ref state);
+    while (result != TransformationStatus.Done)
+    {
+        result = Brotli.Compress(bytes, compressed, out consumed, out written, ref state);
+        bytes = bytes.AsSpan().Slice(consumed).ToArray();
+    }
+    result = Brotli.FlushEncoder(Array.Empty<byte>(), compressed, out consumed, out written, ref state);
```
shiftylogic: Do you always call FlushEncoder with an empty array as the first argument? If so, having that argument at all doesn't make sense.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+Simple method to decompress bytes to file. If out data is larger than destination ```result``` be ```TransformationStatus.DestinationTooSmall``` and it's available to call Decompress again to collect rest of data.
+
+```C#
+ static void DecompSimple(byte[] bytes, string outFile, int decompressedLength)
+ {
+    Brotli.State state = new Brotli.State();
+    byte[] data = ms.ToArray();
+    byte[] decompressed = new byte[decLen];
+    TransformationStatus result = Brotli.Decompress(data, decompressed, out int consumed, out int written, ref state);
```
shiftylogic: Why is there a loop necessary for compression above, but not decompression here?

shiftylogic: Also, is FlushEncoder only for compression?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+    using (BrotliStream decompressBrotli = new BrotliStream(compressed, CompressionMode.Decompress))
+    {
+        decompressBrotli.CopyTo(fs);
+    }
+    fs.Dispose();
+}
+```
+
+#### Dynamic Web Compress
+You can add code like this to you Global.asax.cs in ASP.NET aplication to support dynamic compress
```
shiftylogic: second 'you' => your

shiftylogic: aplication => application

shiftylogic: compress => compression?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,219 @@
...
+
+public static void BrotliEncodePage()
+{
+    HttpResponse Response = HttpContext.Current.Response;
+
+    if (IsBrotliSupported())
+    {
+        string AcceptEncoding = HttpContext.Current.Request.Headers["Accept-Encoding"];
+        Response.Filter = new System.IO.Compression.BrotliStream(Response.Filter,
+        System.IO.Compression.CompressionMode.Compress);
```
shiftylogic: remove line-break here.

src/System.IO.Compression.Brotli/README.md
Vedin: FlushEncode is only for compression, because we should say that we won't sent any more data for compress and want to get compressed data.

src/System.IO.Compression.Brotli/README.md
Vedin: As for me I always do in this way, but it is possible to send the last part of data to flush instead of compress, the result will be the same.

src/System.IO.Compression.Brotli/README.md
Vedin: +

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+            public void SetQuality(uint quality);
+            public void SetWindow(uint window);
+        }
+    }
+```
+```out bytesConsumed``` - number of bytes used from source data while executing the operation
+
+```out bytesWritten``` - number of bytes are written in destination while executing the operation
+
+```State``` is used in both compress and decompress mode to save a temporary status and data of process.  State will set automaticaly at first call of Compress or Decompress. Once it is initialized and being used for one data stream, it shouldn't be used for another one and in other mode.
```
shiftylogic: * automaticaly => automatically
* "at first call" => "on initial call"
* Remove sentence "Once it is initialized..." and replace with "The state cannot be re-used for multiple data streams and cannot switch modes (compress vs. decompress)."

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The ```Compress``` performs data compression. Taking data from ```source``` and writing already compressed data to ```destination```.
```
shiftylogic: Remove "already"

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The ```FlushEncoder``` returns compressed data, which was sent to ```state``` using Compress. This method always should be called after all Compress executions (```isFinished = true```) or when you want to get compressed data immediately (```isFinished = false```).
```
shiftylogic: "always should" => "should always"

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The ```Decompress``` Decompresses the data from ```source``` into ```destination```. 
+
+```SetQuality``` allows you to set quality of compression ```0 to 11```. The higher quality means the higher compression ratio.
```
shiftylogic: Higher quality means higher compression ratios but more compute time.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The ```Compress``` performs data compression. Taking data from ```source``` and writing already compressed data to ```destination```.
+
+The ```FlushEncoder``` returns compressed data, which was sent to ```state``` using Compress. This method always should be called after all Compress executions (```isFinished = true```) or when you want to get compressed data immediately (```isFinished = false```).
+
+The ```Decompress``` Decompresses the data from ```source``` into ```destination```. 
+
+```SetQuality``` allows you to set quality of compression ```0 to 11```. The higher quality means the higher compression ratio. 
+
+```SetWindow``` - Logarithm of Recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is (1 << value) - 16. Possible values: ```11 to 24```
```
shiftylogic: Recommended => lowercase R

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
+﻿# System.IO.Compression.Brotli
+
+## Introduction
+
+Brotli is a generic-purpose lossless compression algorithm that compresses data
```
ahsonkhan: nit: general-purpose instead of generic purpose?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+using a combination of a modern variant of the LZ77 algorithm, Huffman coding
+and 2nd order context modeling, with a compression ratio comparable to the best
+currently available general-purpose compression methods. It is similar in speed
+to deflate but offers more dense compression.
+
+The specification of the Brotli Compressed Data Format is defined in [RFC 7932](https://www.ietf.org/rfc/rfc7932.txt).
+
+Brotli encoding is supported by most web browsers, major web servers, and some CDNs.
```
ahsonkhan: Can you define what CDN stands for as this is the first time it is used in the README?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+            public bool CompressMode { get; }
+            public void Dispose();
+            public void SetQuality(uint quality);
+            public void SetWindow(uint window);
+        }
+    }
+```
+```out bytesConsumed``` - number of bytes used from source data while executing the operation
+
+```out bytesWritten``` - number of bytes are written in destination while executing the operation
```
ahsonkhan: "number of bytes that are written..."

src/System.IO.Compression.Brotli/README.md
shiftylogic: That was my reaction too. However, this text was lifted from the official Google Brotli repo, so I decided to ignore it.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+```State``` is used in both compress and decompress mode to save a temporary status and data of process.  State will set automatically on initial call of Compress or Decompress. The state cannot be re-used for multiple data streams and cannot switch modes (compress vs. decompress).
```
ahsonkhan: "State will be set automatically..."

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The ```Compress``` performs data compression. Taking data from ```source``` and writing compressed data to ```destination```.
```
ahsonkhan: "It takes data from source..."?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+The ```FlushEncoder``` returns compressed data, which was sent to ```state``` using Compress. This method should always be called after all Compress executions (```isFinished = true```) or when you want to get compressed data immediately (```isFinished = false```).
+
+The ```Decompress``` Decompresses the data from ```source``` into ```destination```.
```
ahsonkhan: nit: lowercase D in Decompresses

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+```State``` is used in both compress and decompress mode to save a temporary status and data of process.  State will set automatically on initial call of Compress or Decompress. The state cannot be re-used for multiple data streams and cannot switch modes (compress vs. decompress).
+
+The ```Compress``` performs data compression. Taking data from ```source``` and writing compressed data to ```destination```.
+
+The ```FlushEncoder``` returns compressed data, which was sent to ```state``` using Compress. This method should always be called after all Compress executions (```isFinished = true```) or when you want to get compressed data immediately (```isFinished = false```).
+
+The ```Decompress``` Decompresses the data from ```source``` into ```destination```. 
+
+```SetQuality``` allows you to set quality of compression ```0 to 11```. The higher quality means the higher compression ratio, but more compute time.
```
ahsonkhan: "...you to set the quality of compression between 0 and 11 (inclusive)."

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+```SetWindow``` - Logarithm of recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is (1 << value) - 16. Possible values: ```11 to 24```
```
ahsonkhan: put (1 << value) - 16 between inline code tags?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+### Examples
+Simple method to compress bytes to file.
+```C#
+
+ static void CompressSimple(byte[] bytes, string OutFile)
+ {
+    Brotli.State state = new Brotli.State();
+    byte[] compressed = new byte[Brotli.GetMaximumCompressedSize(bytes.Length)];
+    state.SetQuality();
```
ahsonkhan: SetQuality doesn't have an argument passed to it.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+    TransformationStatus result = Brotli.Compress(bytes, compressed, out int consumed, out int written, ref state);
+    while (result != TransformationStatus.Done)
+    {
+        result = Brotli.Compress(bytes, compressed, out consumed, out written, ref state);
+        bytes = bytes.AsSpan().Slice(consumed).ToArray();
```
ahsonkhan: Why cast it to array when Brotli.Compress takes a span?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+    }
+    result = Brotli.FlushEncoder(Array.Empty<byte>(), compressed, out consumed, out written, ref state);
+    byte[] ans = new byte[written];
+    Array.Copy(compressed, ans, written);
```
ahsonkhan: I am assuming this array.copy is necessary? If so, can you explain why? Do we have a plan to remove the need to copy into an array? Would it be resolved by having a span version of WriteAllBytes?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+    File.WriteAllBytes(OutFile, ans);
+ }
+```
+
+Simple method to decompress bytes to file. If out data is larger than destination ```result``` will be ```TransformationStatus.DestinationTooSmall``` and it's available to call Decompress again to collect rest of data.
```
ahsonkhan: nit: "If out data is larger than destination, ..." (add comma)

Also, I don't know how to read "and it's available to call Decompress again...". Can you please re-word it?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+    public override void Flush();
+    protected virtual void FlushEncoder(bool finished);
+    public override int Read(byte[] buffer, int offset, int count);
+    public override long Seek(long offset, SeekOrigin origin);
+    public override void SetLength(long value);
+    public override void Write(byte[] buffer, int offset, int count);
+}
+```
+
+```Write``` write and compress bytes to stream, can only be called in Compress mode
```
ahsonkhan: Add hyphen?
"Write - write and compress..."

Same with Read.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+{
+    using (var compressor = new BrotliStream(compressed, CompressionMode.Compress, true))
+    {
+        compressor.Write(data, i, chunkSize);
+        compressor.Dispose();
+    }
+}
+```
+
+After calling method compressed data will be in ```MemoryStream compressed``` .
```
ahsonkhan: "After calling the method, the compressed data..."

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+    using (BrotliStream decompressBrotli = new BrotliStream(compressed, CompressionMode.Decompress))
+    {
+        decompressBrotli.CopyTo(fs);
+    }
+    fs.Dispose();
+}
+```
+
+#### Dynamic Web Compression
+You can add code like this to your Global.asax.cs in ASP.NET application to support dynamic compress
```
ahsonkhan: "...dynamic compression"

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+}
+
+public static void BrotliEncodePage()
+{
+    HttpResponse Response = HttpContext.Current.Response;
+
+    if (IsBrotliSupported())
+    {
+        string AcceptEncoding = HttpContext.Current.Request.Headers["Accept-Encoding"];
+        Response.Filter = new System.IO.Compression.BrotliStream(Response.Filter,        System.IO.Compression.CompressionMode.Compress);
```
ahsonkhan: Please remove the extra spacing between Response.Filter and System.IO.Compression.CompressionMode.Compress.

src/System.IO.Compression.Brotli/README.md
```diff
@@ -0,0 +1,214 @@
...
+
+```SetQuality``` allows you to set quality of compression ```0 to 11```. The higher quality means the higher compression ratio, but more compute time.
+
+```SetWindow``` - Logarithm of recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is (1 << value) - 16. Possible values: ```11 to 24```
+
+### Examples
+Simple method to compress bytes to file.
+```C#
+
+ static void CompressSimple(byte[] bytes, string OutFile)
```
ahsonkhan: string outFile (lowercase O)

src/System.IO.Compression.Brotli/README.md
Vedin: @ahsonkhan because bytes is byte[] array, I can change the CompressSimple signature or add other variable. what looks better for you?

src/System.IO.Compression.Brotli/README.md
Vedin: <<Would it be resolved by having a span version of WriteAllBytes?
Yes,

src/System.IO.Compression.Brotli/README.md
ahsonkhan: Why not wrap bytes as a Span\<byte\> local and pass that to Brotli.Compress? You can do something like this:
`result = Brotli.Compress(spanBytes.Slice(consumed), compressed, out consumed, out written, ref state);`

src/System.IO.Compression.Brotli/README.md
ahsonkhan: Should we add a source/link to that repo page as a reference?

src/System.IO.Compression.Brotli/README.md
ahsonkhan: If compressed was a span, wouldn't passing compressed.Slice(0, written).ToArray() to WriteAllBytes work?

Vedin: I'll change it to     File.WriteAllBytes(outFile, compressed.AsSpan().Slice(0, written).ToArray());

ahsonkhan: That works. Any reason why you aren't wrapping compressed into a Span first before passing it to FlushEncoder? You would save the additional AsSpan() call (or implicit conversion from array to span).

Vedin: I don't want to create additional variable, because the main goal of examples is to be very simple and understandable, not fully optimized.

ahsonkhan: I understand. However, imo, we should make sure that sample code showcases the patterns we expect users to write (which includes using spans where possible to avoid unnecessary work). Up to you :)

PR #1663, opened 2017-07-11
Brotli Update Readme

Vedin:
@ahsonkhan we can continue conversation here. I think that I resolved all comments from previous pr.

### DIFF
--- a/src/System.IO.Compression.Brotli/README.md
+++ b/src/System.IO.Compression.Brotli/README.md
@@ -10,7 +10,7 @@ to deflate but offers more dense compression.
 
 The specification of the Brotli Compressed Data Format is defined in [RFC 7932](https://www.ietf.org/rfc/rfc7932.txt).
 
-Brotli encoding is supported by most web browsers, major web servers, and some CDNs (Content delivery network).
+Brotli encoding is supported by most web browsers, major web servers, and some CDNs (Content Delivery Networks).
 
 ## Brotli Class
 ```C#
@@ -39,14 +39,13 @@ The ```FlushEncoder``` returns compressed data, which was sent to ```state``` us
 
 The ```Decompress``` decompresses the data from ```source``` into ```destination```. 
 
-```SetQuality``` allows you to set quality of compression ```between 0 to 11```. The higher quality means the higher compression ratio, but more compute time.
+```SetQuality``` allows you to set quality of compression ```between 0 to 11``` inclusive. The higher quality means the higher compression ratio, but more compute time.
 
-```SetWindow``` - Logarithm of recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is ```(1 << value) - 16```. Possible values: ```11 to 24```
+```SetWindow``` - Logarithm of recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is ```(1 << value) - 16```. Possible values: ```betwwen 11 to 24``` inclusive
 
 ### Examples
 Simple method to compress bytes to file.
 ```C#
-
  static void CompressSimple(byte[] bytes, string outFile)
  {
     Brotli.State state = new Brotli.State();
--- a/src/System.IO.Compression.Brotli/README.md
+++ b/src/System.IO.Compression.Brotli/README.md
@@ -39,7 +39,7 @@ The ```FlushEncoder``` returns compressed data, which was sent to ```state``` us
 
 The ```Decompress``` decompresses the data from ```source``` into ```destination```. 
 
-```SetQuality``` allows you to set quality of compression ```between 0 to 11``` inclusive. The higher quality means the higher compression ratio, but more compute time.
+```SetQuality``` allows you to set the quality of compression ```between 0 to 11``` inclusive. The higher quality means the higher compression ratio, but more compute time.
 
 ```SetWindow``` - Logarithm of recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is ```(1 << value) - 16```. Possible values: ```betwwen 11 to 24``` inclusive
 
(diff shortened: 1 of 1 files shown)


Vedin:

src/System.IO.Compression.Brotli/README.md
```diff
@@ -10,7 +10,7 @@ to deflate but offers more dense compression.
 
 The specification of the Brotli Compressed Data Format is defined in [RFC 7932](https://www.ietf.org/rfc/rfc7932.txt).
 
-Brotli encoding is supported by most web browsers, major web servers, and some CDNs.
+Brotli encoding is supported by most web browsers, major web servers, and some CDNs (Content delivery network).
```
ahsonkhan: nit: Content Delivery Networks?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -29,43 +29,41 @@ public static class Brotli {
...
-The ```Compress``` performs data compression. Taking data from ```source``` and writing compressed data to ```destination```.
+The ```Compress``` performs data compression. It takes data from ```source``` and writes compressed data to ```destination```.
 
 The ```FlushEncoder``` returns compressed data, which was sent to ```state``` using Compress. This method should always be called after all Compress executions (```isFinished = true```) or when you want to get compressed data immediately (```isFinished = false```).
 
-The ```Decompress``` Decompresses the data from ```source``` into ```destination```. 
+The ```Decompress``` decompresses the data from ```source``` into ```destination```. 
 
-```SetQuality``` allows you to set quality of compression ```0 to 11```. The higher quality means the higher compression ratio, but more compute time.
+```SetQuality``` allows you to set quality of compression ```between 0 to 11```. The higher quality means the higher compression ratio, but more compute time.
```
ahsonkhan: "allows you to set the quality of compression between 0 and 11"
I would also add "inclusive" to be less ambiguous. Between 0 to 11 could mean [0, 11] or (0, 11).

src/System.IO.Compression.Brotli/README.md
```diff
@@ -29,43 +29,41 @@ public static class Brotli {
...
 
-```SetWindow``` - Logarithm of recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is (1 << value) - 16. Possible values: ```11 to 24```
+```SetWindow``` - Logarithm of recommended sliding LZ77 window size. Encoder may reduce this value, e.g. if input is much smaller than window size. Window size is ```(1 << value) - 16```. Possible values: ```11 to 24```
```
ahsonkhan: I am curious, is there any difference between having three \` and just a single \`?

i.e. 
\```(1 << value) - 16\```
vs.
\`(1 << value) - 16\`

Doesn't seem like it:
```(1 << value) - 16```
`(1 << value) - 16`

File: src/System.IO.Compression.Brotli/README.md [hunk 4]
```diff
@@ -29,43 +29,41 @@ public static class Brotli {
...
 Simple method to compress bytes to file.
 ```C#
 
- static void CompressSimple(byte[] bytes, string OutFile)
+ static void CompressSimple(byte[] bytes, string outFile)
  {
     Brotli.State state = new Brotli.State();
+    Span<byte> spanBytes = new Span<byte>(bytes);
     byte[] compressed = new byte[Brotli.GetMaximumCompressedSize(bytes.Length)];
-    state.SetQuality();
```
ahsonkhan: Was the SetQuality call unnecessary?

src/System.IO.Compression.Brotli/README.md
```diff
@@ -29,43 +29,41 @@ public static class Brotli {
...
-    TransformationStatus result = Brotli.Compress(bytes, compressed, out int consumed, out int written, ref state);
+    TransformationStatus result = Brotli.Compress(spanBytes, compressed, out int consumed, out int written, ref state);
     while (result != TransformationStatus.Done)
     {
-        result = Brotli.Compress(bytes, compressed, out consumed, out written, ref state);
-        bytes = bytes.AsSpan().Slice(consumed).ToArray();
+        result = Brotli.Compress(spanBytes, compressed, out consumed, out written, ref state);
+        spanBytes = spanBytes.Slice(consumed);
```
ahsonkhan: Is this assignment necessary?
Why not pass `spanBytes.Slice(consumed)` to Brotli.Compress?

src/System.IO.Compression.Brotli/README.md
Vedin: Yes, setXXX is no sense method, so I removed it.

Vedin:

src/System.IO.Compression.Brotli/README.md
Vedin: as I remember Slice return Span and don't change argument. So if we have multiply iterations of this loop we had or create a variable and do ```allConsumed += consumed``` and than can do ```spanBytes.Slice(allConsumed)``` or just update spanBytes every time.

ahsonkhan: That is correct. Good point.

src/System.IO.Compression.Brotli/README.md
Vedin: Yes, really. Should I change this everywhere?

ahsonkhan: It seems unnecessary to have three \`. I would change it to single \` but your call.

src/System.IO.Compression.Brotli/README.md
ahsonkhan: to clarify: "allows you to set **the** quality of compression between 0 **and** 11"

ahsonkhan:
@Vedin, thoughts on adding a reference link?

https://github.com/dotnet/corefxlab/pull/1658#discussion_r126804291

Vedin:
@ahsonkhan , Do you mean a reference to a google brotli repo?
In fact, I don't know. Can we do it or should we do it?

ahsonkhan:
@Vedin, @shiftylogic good to go?

PR #1664, opened 2017-07-11
Decompress Invalid Data test for Brotli

Vedin:


tests/System.IO.Compression.Tests/BrotliPrimitivesTests.cs
```diff
@@ -131,6 +133,34 @@ public void WrongWindowSize(uint windowSize)
...
         [Theory]
         [InlineData(1)]
+        [InlineData(10)]
+        [InlineData(1024)]
+        public void DecompressRandomInvalidData(int totalSize)
+        {
+            Brotli.State state = new Brotli.State();
+            byte[] data = new byte[totalSize];
+            new Random(42).NextBytes(data);
+            Assert.Throws<System.IO.IOException>(delegate { Brotli.Decompress(data, Array.Empty<byte>(), out int consumed, out int written, ref state);  });
```
ahsonkhan: Why do we need a delegate here?

Vedin: It's just way to transform call to System.Action and handle and check exception. I'm really interested in other ways to do it

ahsonkhan: Ah yes. Thanks for clarifying. You could use lambda expressions which are similar (or the same?). But a delegate is fine.
https://github.com/dotnet/corefxlab/blob/master/tests/System.Buffers.Primitives.Tests/CtorArray.cs#L200
```C#
Assert.Throws<System.IO.IOException>(() => Brotli.Decompress(data, Array.Empty<byte>(), out int consumed, out int written, ref state)); 
```

ahsonkhan:
@Vedin, @shiftylogic good to go?

PR #1666, opened 2017-07-12
Fixed a bug in the IOutput implementation of WritableBuffer

KrzysztofCwalina:
The current implementation often results in very small segments. The new implementation will ensure that segments are never smaller than 256 bytes.

@davidfowl


KrzysztofCwalina:
@dotnet-bot build Ubuntu16.04

PR #1668, opened 2017-07-18
Fast length and bounds checks, use long for Length's

pakrym:


tests/System.IO.Pipelines.Tests/PipeLengthTests.cs
```diff
@@ -102,6 +104,10 @@ public void LengthNotChangeAfterReadAdvanceExamine()
         [Fact]
         public void ByteByByteTest()
         {
+            //while (!Debugger.IsAttached)
```
ahsonkhan: Is this intended to be checked in?

tests/System.IO.Pipelines.Tests/UnownedBufferReaderFacts.cs
```diff
@@ -4,6 +4,7 @@
 
 using System;
 using System.Buffers;
+using System.Diagnostics;
```
ahsonkhan: Why is this addition necessary?

ahsonkhan:
> Fast length and bounds checks, use long for Length's

Is this actually faster? By how much? Can you explain why that is?

pakrym:
@ahsonkhan profiling kestrel with this change ATM.

This change allows us to calculate length of readable buffer and run bounds checks without traversing segment linked list when there are more then one segment.

src/System.IO.Pipelines/BufferSegment.cs
```diff
@@ -26,13 +26,18 @@ internal class BufferSegment : IDisposable
...
+        /// leased Next is guaranteed to be null. Start, End, and Next are used together in order to create a linked-list of discontiguous
+        /// working memory. The "active" memory is grown when bytes are copied in, End is increased, and Next is assigned. The "active"
         /// memory is shrunk when bytes are consumed, Start is increased, and blocks are returned to the pool.
         /// </summary>
         public BufferSegment Next;
 
         /// <summary>
+        /// Combined length of all segments before this
+        /// </summary>
+        public long RunningLength;
```
davidfowl: This makes cursors huge.

pakrym: It's not a cursor.

davidfowl: Ah wait this is a class so it won't bloat the cursor size 👍

src/System.IO.Pipelines/ReadCursor.cs
```diff
@@ -68,73 +68,38 @@ private bool IsEndMultiSegment()
...
-                    }
-                    else
-                    {
-                        length += segment.End - index;
-                        segment = segment.Next;
-                        index = segment.Start;
-                    }
-                }
-            }
+            return (endSegment.RunningLength + endIndex) - (start.RunningLength + startIndex);
```
halter73: Woohoo!!🎉🎈🎆

src/System.IO.Pipelines.Extensions/PipelineReaderExtensions.cs
```diff
@@ -22,15 +22,16 @@ public static ValueTask<int> ReadAsync(this IPipeReader input, ArraySegment<byte
                 var result = awaiter.GetResult();
                 var inputBuffer = result.Buffer;
 
-                var sliced = inputBuffer.Slice(0, Math.Min(inputBuffer.Length, destination.Count));
+                var length = (int) Math.Min(inputBuffer.Length, destination.Count);
```
benaadams: This `(int)` cast safe? Why is it needed?

pakrym: `destination.Count` is `int`, needed because `inputBuffer.Length` is `long`

benaadams: Ah so `destination.Count` will always limit to `int` range. Makes sense.

src/System.IO.Pipelines/ReadableBuffer.cs
```diff
@@ -133,19 +134,25 @@ public ReadableBuffer Slice(ReadCursor start)
...
 
             var begin = BufferStart.Seek(start, BufferEnd, false);
-            return new ReadableBuffer(begin, BufferEnd);
+            return SliceImpl(begin, BufferEnd);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private ReadableBuffer SliceImpl(ReadCursor start, ReadCursor end)
+        {
+            return new ReadableBuffer(start, end);
```
davidfowl: ?

pakrym:
:up::date:

src/System.IO.Pipelines/BufferSegment.cs
```diff
@@ -128,16 +133,37 @@ public static BufferSegment Clone(ReadCursor beginBuffer, ReadCursor endBuffer,
...
+        public void SetNext(BufferSegment segment)
+        {
+            if (segment == null)
+            {
+                throw new ArgumentNullException(nameof(segment));
+            }
+            if (Next != null)
+            {
+                throw new InvalidOperationException($"{nameof(Next)} is already set.");
+            }
```
halter73: This is only called internally. When we were just calling `segment.Next = nextSegment` there weren't any of these silly safety checks. I'm fine with making these asserts.

@benaadams Is this a good candidate for the aggressive inlining attribute?

src/System.IO.Pipelines/ReadCursor.cs
```diff
@@ -89,52 +90,23 @@ internal int GetLength(ReadCursor end)
...
             {
                 return this;
             }
 
             ReadCursor cursor;
-            if (Segment == end.Segment && end.Index - Index >= bytes)
+            if (Segment == end.Segment && end.Index >= bytes + Index)
             {
-                cursor = new ReadCursor(Segment, Index + bytes);
+                cursor = new ReadCursor(Segment, Index + (int)bytes);
```
halter73: Maybe add a comment saying how that if `end.Index - Index >= bytes`, bytes must be safely castable to an int. I think this one is a little bit less obvious than the other uncommented int casts.

benaadams:
>  Is this SetNext method a good candidate for the aggressive inlining attribute?

It currently won't inline by default because of the while; how common would be the while loop executing? e.g. appending a chain of BufferSegments

Could do something like (not verified correct code):
```csharp
public void SetNext(BufferSegment segment)
{
    Debug.Assert(segment != null);
    Debug.Assert(Next == null);

    Next = segment;

    segment.RunningLength = this.RunningLength + this.ReadableBytes;

    if (segment.Next != null)
    {
        SetRunningLengthChain(segment);
    }
}

private static void SetRunningLengthChain(BufferSegment startSegment)
{
    Debug.Assert(startSegment.Next != null);

    var runningLength = startSegment.RunningLength + startSegment.ReadableBytes;
    var segment = startSegment.Next;

    do
    {
        segment.RunningLength = runningLength;
        runningLength += segment.ReadableBytes;
        segment = segment.Next;
    } while (segment != null);
}
```

PR #1670, opened 2017-07-24
Int16 UTF8 parser

Vedin:
Add Int16 parser and test to it. 
Fix bug in Utf16 Int16 Parser, because it failed on new tests.


ahsonkhan:
@Vedin , it looks good overall.
Can you please also add performance test results so we can see the impact of this optimization to the previous implementation?

tests/System.Text.Primitives.Tests/Parsing/PrimitiveParserInt16PerfTests.cs
```diff
@@ -0,0 +1,308 @@
...
+        [InlineData("+๒๑๔๗๖abcdefghijklmnop")]
+        [InlineData("+ลบ๒๑abcdefghijklmnop")]
+        [InlineData("+๒๑๔๗abcdefghijklmnop")]
+        [InlineData("+ลบ๒๑abcdefghijklmnop")]
+        [InlineData("+ลบ๒abcdefghijklmnop")]
+        [InlineData("+๒๑๔abcdefghijklmnop")]
+        [InlineData("+ลบ๒๑abcdefghijklmnop")]
+        [InlineData("+๒๑abcdefghijklmnop")]
+        [InlineData("+ลบ๑๔abcdefghijklmnop")]
+        public unsafe void ParseInt16Thai(string text)
```
ahsonkhan: Is there any particular reason for this to be an unsafe method? If not, I would remove it here and elsewhere.

Vedin:
image

PR #1674, opened 2017-07-27
Merge aspnetcore branch back into master

pakrym:


src/System.IO.Pipelines/MemoryPoolBlock.cs
```diff
@@ -106,7 +111,10 @@ public override void Retain()
         public override void Release()
         {
             // TODO: should it check IsRetained?
-            Interlocked.Decrement(ref _referenceCount);
+            if (Interlocked.Decrement(ref _referenceCount) <= 0)
```
halter73: I would make this `==` instead.

src/System.IO.Pipelines/System.IO.Pipelines.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>An abstraction for doing efficient asynchronous IO</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
ahsonkhan: I am curious, what motivated this version bump?

src/System.Text.Http.Parser/System.Text.Http.Parser.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>HTTP Parser</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
ahsonkhan: Why did we bump up the version?

pakrym: Was done by accident as part of different change.

pakrym:
ping @halter73 @ahsonkhan

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
```diff
@@ -220,7 +223,7 @@ public static unsafe string GetUtf8String(this ReadableBuffer buffer)
 
                 buffer.CopyTo(destination);
 
-                textSpan = destination.Slice(0, buffer.Length);
+                textSpan = destination.Slice(0, (int) buffer.Length);
```
ahsonkhan: Can you please add a comment here: // We are able to cast because buffer.Length < 128?

src/System.IO.Pipelines/Pipe.cs
```diff
@@ -388,59 +390,95 @@ void IPipeWriter.Complete(Exception exception)
...
             TrySchedule(_readerScheduler, awaitable);
 
-            if (_readerCompletion.IsCompleted)
+            if (readerCompleted)
             {
                 Dispose();
             }
         }
 
+
```
ahsonkhan: nit: remove extra line

src/System.IO.Pipelines/ReadCursor.cs
```diff
@@ -88,52 +89,24 @@ internal int GetLength(ReadCursor end)
...
+        internal ReadCursor Seek(long bytes, ReadCursor end, bool checkEndReachable = true)
         {
             if (IsEnd)
             {
                 return this;
             }
 
             ReadCursor cursor;
-            if (Segment == end.Segment && end.Index - Index >= bytes)
+            if (Segment == end.Segment && end.Index >= bytes + Index)
```
ahsonkhan: Is it possible for `bytes + Index` to overflow?

src/System.IO.Pipelines/ReadCursor.cs
```diff
@@ -157,7 +130,7 @@ private ReadCursor SeekMultiSegment(int bytes, ReadCursor end, bool checkEndReac
                 {
                     if (segmentPart.Length >= bytes)
                     {
-                        result = new ReadCursor(segmentPart.Segment, segmentPart.Start + bytes);
+                        result = new ReadCursor(segmentPart.Segment, segmentPart.Start + (int)bytes);
```
ahsonkhan: What happens if `(segmentPart.Start + (int)bytes > segmentPart.Length)`?

src/System.IO.Pipelines/ReadCursor.cs
```diff
@@ -314,25 +287,7 @@ public override int GetHashCode()
...
-            while (current != null)
-            {
-                if (current == Segment)
-                {
-                    return true;
-                }
-                current = current.Next;
-            }
-            return false;
+            return other.Segment.RunningLength + other.Index <= Segment.RunningLength + Index;
```
ahsonkhan: Is overflow possible?

tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
```diff
@@ -298,12 +298,12 @@ private unsafe void TestValue(ref ReadableBuffer readBuffer, ulong value)
         {
             fixed (byte* ptr = &readBuffer.First.Span.DangerousGetPinnableReference())
             {
-                Assert.True(ptr != null);
                 string s = value.ToString(CultureInfo.InvariantCulture);
                 int written;
                 fixed (char* c = s)
                 {
-                    written = Encoding.ASCII.GetBytes(c, s.Length, ptr, readBuffer.Length);
+                    // Sizes smaller then max int
```
ahsonkhan: Can you re-write the comment. I am not sure how to read that.

tests/System.IO.Pipelines.Tests/WritableBufferFacts.cs
```diff
@@ -92,14 +92,14 @@ public void ThrowsOnAdvanceOverMemorySize()
...
-                int offset = 0;
+                long offset = 0;
                 while (true)
                 {
                     var result = await pipe.Reader.ReadAsync();
                     var input = result.Buffer;
                     if (input.Length == 0) break;
-
-                    Assert.True(input.EqualsTo(new Span<byte>(data, offset, input.Length)));
+                    // Sizes less then max int
```
ahsonkhan: nit: same here

tests/System.Buffers.Primitives.Tests/BufferReferenceTesting.cs
```diff
@@ -6,17 +6,30 @@ namespace System.Buffers.Tests
 {
     public class BufferReferenceTests
     {
-        public static void TestOwnedBuffer(Func<OwnedBuffer<byte>> create)
+        public static void TestOwnedBuffer(Func<OwnedBuffer<byte>> create, Action<OwnedBuffer<byte>> dispose = null)
```
ahsonkhan: I am curious, what motivated this change?

pakrym: When block is allocated from MemoryPool it has to be properly returned, otherwise `Assert` would fire during finalization crashing test process.

src/System.IO.Pipelines/ReadCursor.cs
pakrym: It's fine, `segmentPart.Length == segmentPart.End - segmentPart.Start`

ahsonkhan: I do not follow.
'bytes' can be as large as `segmentPart.End - segmentPart.Start`, (given the condition `if (segmentPart.Length >= bytes)`). Let's say it is.

Then:
`segmentPart.Start + (int)bytes`
= `segmentPart.Start + segmentPart.End - segmentPart.Start` => `segmentPart.End`

`segmentPart.End` can be > `segmentPart.Length`

What happens in that case?

pakrym: `ReadCursor` stores absolute position inside segment. And it's not `=> segmentPart.End` it's maximum value is `segmentPart.End` which is supported.

pakrym:
:up::date:

src/System.IO.Pipelines/ReadCursor.cs
ahsonkhan: > it's maximum value is `segmentPart.End` which is supported.

Sounds good.

PR #1676, opened 2017-07-31
Add System.Numerics.Tensors library

ericstj:
Add a library to represent tensors with some basic mathematical and resize operations.  Tensors are stored as a single-dimensional array with C-style ordering of elements to aide in easy interop with native libraries without copying the array.  Currently a prototype and primarily a means for experimentation.

/cc @mellinoe @migueldeicaza


src/System.Numerics.Tensors/System/Numerics/TensorTemplate.ttinclude
```diff
@@ -0,0 +1,253 @@
...
+    {
+        new MethodConfiguration("Add", MethodType.Binary, "+", isNumeric:true),
+        new MethodConfiguration("Add", MethodType.BinaryScalar, "+", isNumeric:true),
+        new MethodConfiguration("UnaryPlus", MethodType.Unary, "+", isNumeric:true),
+        new MethodConfiguration("Subtract", MethodType.Binary, "-", isNumeric:true),
+        new MethodConfiguration("Subtract", MethodType.BinaryScalar, "-", isNumeric:true),
+        new MethodConfiguration("UnaryMinus", MethodType.Unary, "-", isNumeric:true),
+        new MethodConfiguration("Increment", MethodType.UnaryInPlace, "++", isNumeric:true),
+        new MethodConfiguration("Decrement", MethodType.UnaryInPlace, "--", isNumeric:true),
+        new MethodConfiguration("Multiply", MethodType.Binary, "*", isNumeric:true),  // element-wise product, not matrix product
```
ericstj: I'm working on Tensor contraction.

ericstj:
Tests failed with 
```
The test source file "D:\j\workspace\windows_nt_re---4abe806d\tests\System.Numerics.Tensors.Tests\bin\Release\netcoreapp1.1\System.Numerics.Tensors.Tests.dll" provided was not found.
```

Something about my project configuration appears busted.  I'll see what I need to fix.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -0,0 +1,1197 @@
...
+                throw new ArgumentException("Operands must have matching dimensions");
+            }
+
+            for (int i = 0; i < left.Rank; i++)
+            {
+                if (left.Dimensions[i] != right.Dimensions[i])
+                {
+                    throw new ArgumentException("Operands must have matching dimensions");
+                }
+            }
```
stephentoub: Seems like a non-trivial amount of validation that happens for each operation.  Are the operations generally expensive enough that this overhead won't matter?

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -0,0 +1,1197 @@
...
+    // When we cross-compile for frameworks that expose ICloneable this must implement ICloneable as well.
+    public class Tensor<T>: IList, ICollection, IEnumerable, IStructuralComparable, IStructuralEquatable
+    {
+        private readonly T[] backingArray;
+        private readonly int[] dimensions;
+        private IReadOnlyList<int> readOnlyDimensions;
+
+        private static ITensorArithmetic<T> arithmetic => TensorArithmetic.GetArithmetic<T>();
+
+        public Tensor(Array fromArray)
```
stephentoub: Presumably we'll want span-based ctors, too?

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -0,0 +1,1197 @@
...
+        /// </summary>
+        public T[] Buffer => backingArray;
+
+        /// <summary>
+        /// Sets all elements in Tensor to <paramref name="value"/>.
+        /// </summary>
+        /// <param name="value">Value to fill</param>
+        public void Fill(T value)
+        {
+            // JIT look here, lend us a hand
```
stephentoub: Array.Fill?  Or a Vector-based implementation?

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -0,0 +1,1197 @@
...
+        }
+        
+    }
+
+    // When we cross-compile for frameworks that expose ICloneable this must implement ICloneable as well.
+    public class Tensor<T>: IList, ICollection, IEnumerable, IStructuralComparable, IStructuralEquatable
+    {
+        private readonly T[] backingArray;
+        private readonly int[] dimensions;
+        private IReadOnlyList<int> readOnlyDimensions;
```
stephentoub: If everything's readonly and it's just these few fields, could `Tensor<T>` be an immutable struct instead of a class?

ericstj: I suspect so, we'll see as it grows, but for now I can switch it.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
ericstj: This is O(numDimensions), whereas the operations are O(numElements) at best.  I don't expect this to make much difference but if it does we can try to come up with some better fast paths which "pack" the dimensions into a single value (rather than array) and allow for a fast comparison for smaller numbers of dimensions.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
ericstj: Yeah, I haven't done too much work on the different CTors but I was thinking span is a definite candidate.  I'd also like some T[], T[,], etc CTORs or static factory methods for convenience.  I'll file an issue tracking that.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
ericstj: According to a post by the JIT team a while back, they recognize this pattern explicitly and generate better code for it.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
ericstj: I've given it a try.

corefxlab.sln
```diff
@@ -1004,5 +1032,10 @@ Global
 		{74EEC686-FB85-43D5-95EA-10048C4A4567} = {3079E458-D0E6-4F99-8CAB-80011D35C7DA}
 		{07CC9F28-FC67-453B-9057-AC9025E5CA67} = {4B000021-5278-4F2A-B734-DE49F55D4024}
 		{C6F061D6-18D6-41E3-A692-C0603A994F30} = {3079E458-D0E6-4F99-8CAB-80011D35C7DA}
+		{662968A7-5BCC-467B-A814-38584D0184CD} = {4B000021-5278-4F2A-B734-DE49F55D4024}
+		{C11F232A-ED36-4DCE-A18D-BF90CED8D1B4} = {3079E458-D0E6-4F99-8CAB-80011D35C7DA}
+	EndGlobalSection
+	GlobalSection(ExtensibilityGlobals) = postSolution
+		SolutionGuid = {C08AC393-4718-423C-8FFA-F97E16E9B9EE}
```
ahsonkhan: I am curious, what does this postSolution GlobalSection do?

ericstj: VS added it, no idea what its for.  Looks like its related to source control extensions in TFS.

ericstj: If there is some concern about this I can revert that portion of the change.

mellinoe:
High-level commentary:

* Have you tested the performance impact of this `IArithmetic` abstraction? Going through a couple levels of indirection, and through interface dispatch, seems like a lot of overhead just to start with for every single arithmetic operation. Given that most of the code is generated, and all internal details, it could be inlined.
* Having some of the IArithmetic implementations throw NotSupportedException seems very bad. Is there a different way to express the concepts without these problems leaking out?
* You mention interop with native libraries. Have you seen how well this works with CUDA/OpenCL/whatever?

vadimkantorov:
Also, a high-level commentary. If the tensor support gets more developed and aims to target machine learning needs, please take a glance over PyTorch / TensorFlow / NumPy designs. These communities have been exploring most efficient platform-dependent ways to accelerate tensor operations (CUDA / OpenCL / BLAS / OpenMP / SSE) and tensor type polymorphism (including float16 types).

ericstj:
@mellinoe right now this is really just an experiment focused on the functional behavior, once that's in place we can play with different ways to represent the arithmetic ops.  I agree I want this stuff inlined where possible.  I haven't spent a ton of time looking into perf yet.  If you have a suggestion file an issue and tag it for the tensors area path.

As far as native libraries, we've experimented with directly passing these to our own native libraries using signatures like double* and int* and it works as expected without copying.  It'll create a memory layout that's consistent with a C-style multi-dimensional array so it should be usable by most native libraries.  We aren't intending for it to work directly with existing libraries, but be easy to use with folks who are developing libraries and want them to interop with C# with minimal work.  Taking a brief look at CUDA and OpenCL I see they use C-style multi-dimensional arrays so I believe this should work.

mellinoe:
It may help to describe the storage layout as "row-major", or use some other appropriate term for higher-dimensional arrays. That would be more explicit than describing them as "C-style multidimensional arrays".

It looks like other similar libraries actually let you control the storage layout explicitly, noting that some algorithms may favor one storage order or the other. We could consider adding that flexibility here, too.

> We aren't intending for it to work directly with existing libraries

Hm, I thought part of the motivation for this was to interface with ML frameworks like TensorFlow, etc. I guess what you mean is that we don't necessarily want to constrain ourselves to work with any particular existing framework at the expense of the API itself?

ericstj:
@mellinoe right now this is prototyping we'll do what is necessary once we discover what our real requirements are.  We don't want to go engineering a bunch of knobs yet until we know what folks want to use.  Plumbing different layout/indexing schemes is a significant amount of work, and it isn't clear to me if there is a real need for differentiation.  I'm refactoring this stuff daily as I understand more about requirements and we may bring in the different layout schemes as needed.  Right now I'd like to provide something and see if its useful for folks and get feedback.  WRT to naming I use the *C-style* term as that is what python uses as well, which differentiates C-Style (first-to-last-dimension-order, aka row-major) vs Fortran-style (last-to-first-dimension - aka column major).  I have some existing folks willing to use this that are fine with C-style ordering which is why I started there.

My meaning behind not making this work with existing libraries is we aren't trying to make this look like an existing technology-specific exchange type.  We're trying to create a single .NET exchange type for tensors that is compatible with as many libraries as possible.  As such we'll start with a feature set that we think is a good starting point and get feedback from the adopting libraries around what we need to add.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
```diff
@@ -0,0 +1,1490 @@
...
+            Assert.Throws<ArgumentException>(() => defaultTensor % 1);
+            Assert.Throws<ArgumentException>(() => defaultTensor % defaultTensor);
+            Assert.Throws<ArgumentException>(() => defaultTensor & defaultTensor);
+            Assert.Throws<ArgumentException>(() => defaultTensor & -1);
+            Assert.Throws<ArgumentException>(() => defaultTensor | defaultTensor);
+            Assert.Throws<ArgumentException>(() => defaultTensor | -1);
+            Assert.Throws<ArgumentException>(() => defaultTensor ^ defaultTensor);
+            Assert.Throws<ArgumentException>(() => defaultTensor ^ -1);
+            Assert.Throws<ArgumentException>(() => defaultTensor >> 1);
+            Assert.Throws<ArgumentException>(() => defaultTensor << 1);
```
asbjornu: Shouldn't `ArgumentException.ParamName` be verified so you know the right kind of `ArgumentException` is thrown? In theory, the exception can be thrown by any code involved in the code path here, not necessarily the validation code being tested here.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
```diff
@@ -0,0 +1,1490 @@
...
+            Assert.Equal(true, StructuralComparisons.StructuralEqualityComparer.Equals(diag, new[] { 7 }));
+            Assert.Throws<ArgumentException>(() => tensor.GetDiagonal(5));
+
+            diag = tensor.GetDiagonal(-1);
+            Assert.Equal(true, StructuralComparisons.StructuralEqualityComparer.Equals(diag, new[] { 8, 7 }));
+            diag = tensor.GetDiagonal(-2);
+            Assert.Equal(true, StructuralComparisons.StructuralEqualityComparer.Equals(diag, new[] { 1 }));
+            Assert.Throws<ArgumentException>(() => tensor.GetDiagonal(-3));
+            Assert.Throws<ArgumentException>(() => tensor.GetDiagonal(-4));
+            Assert.Throws<ArgumentException>(() => tensor.GetDiagonal(-5));
```
asbjornu: Same as above.

ericstj:
Folks I'm not expecting an in-depth code review of this stuff.  Most of it is all experiment/prototype quality.  The things I do want folks to review is the project setup and infrastructure.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
ericstj: Sure I can polish these up.  These aren't really meant to be a full suite of tests yet but merely a start.

corefxlab.sln
ahsonkhan: Imo, we should remove it unless it is intentional or we have a reason for it.

src/System.Numerics.Tensors/System.Numerics.Tensors.csproj
```diff
@@ -0,0 +1,48 @@
...
+  <Import Project="..\..\tools\common.props" />
+  <PropertyGroup>
+    <Description>Tensor class which represents and extends multi-dimensional arrays</Description>
+    <TargetFramework>netstandard1.1</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <PackageTags>.NET Tensor Array corefxlab</PackageTags>
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="System\Numerics\ArrayUtilities.cs" />
```
ahsonkhan: Do we need to explicitly include the source files in the csproj? Other projects in the repo do not require it, for example:
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System.Text.Primitives.csproj
https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System.Buffers.Primitives.csproj

src/System.Numerics.Tensors/System.Numerics.Tensors.csproj
```diff
@@ -0,0 +1,48 @@
...
+      <LastGenOutput>TensorArithmetic.cs</LastGenOutput>
+    </None>
+    <None Update="System\Numerics\Tensor.Operations.tt">
+      <Generator>TextTemplatingFileGenerator</Generator>
+      <LastGenOutput>Tensor.Operations.cs</LastGenOutput>
+    </None>
+    <None Update="System\Numerics\TensorTemplate.ttinclude" />
+  </ItemGroup>
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
```
ahsonkhan: Can you please explain what this is?

src/System.Numerics.Tensors/System.Numerics.Tensors.csproj
ericstj: Yes, we do.  I've disable the wildcard include because I want to apply the code-generator metadata to the TT-generated items.  See: EnableDefaultCompileItems

src/System.Numerics.Tensors/System.Numerics.Tensors.csproj
ericstj: Used by VS for the T4 code generator.

ericstj: Added a comment.

src/System.Numerics.Tensors/System.Numerics.Tensors.csproj
ericstj: Added a comment.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
ericstj: Fixed these.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
ericstj: Fixed.

corefxlab.sln
ericstj: done

src/System.Numerics.Tensors/System.Numerics.Tensors.csproj
ahsonkhan: Ah yes, hadn't noticed that property. Thanks for clarifying.

ericstj:
Ubuntu failing with `Reason: Unhandled Exception: System.IO.Pipelines.Networking.Libuv.Interop.UvException: Error -16 EBUSY resource busy or locked`

ahsonkhan:
> Ubuntu failing

This failure has been recurring (and intermittent). I have filed an issue: https://github.com/dotnet/corefxlab/issues/1687

Thanks.

ericstj:
Unless there is further feedback I plan to merge this so that I can do further work: https://github.com/dotnet/corefxlab/issues?q=is%3Aopen+is%3Aissue+label%3Aarea-System.Numerics.Tensors

src/System.Numerics.Tensors/System.Numerics.Tensors.sln
```diff
@@ -0,0 +1,28 @@
+﻿
```
ahsonkhan: Do we need to have a separate solution for this project?

ericstj: I prefer a single solution for my development since I only care about this library.

ahsonkhan: > FWIW this is what we do in CoreFx as well

That is true. Good point.

> I prefer a single solution for my development since I only care about this library. 

Fair enough :)

PR #1690, opened 2017-08-04
Tensor Contraction

### LINKED ISSUE #1681: Implement Tensor Contraction
We need an operation
Contract(Tensor<T> left, Tensor<T> right, int[] leftAxes, int[] rightAxes)

See: https://en.wikipedia.org/wiki/Tensor_contraction

TL;DR
This operation produces a resulting tensor with dimensions of the axes not listed in leftAxes and rightAxes where the value of the element at an index in the result is the sum of the products of all the elements in the  indices listed in left/rightAxes.

For two dimensional tensors this can be seen as [matrix multiplication](https://en.wikipedia.org/wiki/Matrix_multiplication).

ericstj:
Fixes #1681 

This adds Tensor contraction (the generic form of Matrix multiplication).

In doing this I came up with a simpler concept to encapsulate the offset calculation for the layout: *strides* for each dimension.  I refactored other non-arithmetic methods on Tensor to use the same *strides* concept.

/cc @mellinoe @migueldeicaza

### DIFF
--- a/src/System.Numerics.Tensors/System/Numerics/ArrayUtilities.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/ArrayUtilities.cs
@@ -1,5 +1,7 @@
 ﻿using System;
 using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
 using System.Text;
 
 namespace System.Numerics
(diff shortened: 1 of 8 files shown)


src/System.Numerics.Tensors/System/Numerics/ArrayUtilities.cs
```diff
@@ -15,5 +17,181 @@ public static T[] Empty<T>()
...
+
+            return true;
+        }
+
+        /// <summary>
+        /// Gets the set of strides that can be used to calculate the offset of n-dimensions in a 1-dimensional layout
+        /// </summary>
+        /// <param name="dimensions"></param>
+        /// <returns></returns>
+        public static int[] GetStrides(IReadOnlyList<int> dimensions)
```
KrzysztofCwalina: Why do you use IReadOnlyList here, but int[] in GetProduct to represent dimensions?

ericstj: These are just internal helpers and those happened to be the types I was passing to them.  Is there a reason to favor one vs the other?

KrzysztofCwalina: int[] does not allocate when you iterate over it and there is no interface dispatch.

benaadams: Also `ReadOnlySpan<int>` may be an option if you want to be explicit in RO access

ericstj: @KrzysztofCwalina yep: that's why I opted for `[]` initially, but then didn't have a good way to expose the `[]` without giving people the ability to mutate.  I'll make the `[]` internal then and switch these to all use the `[]`.

ericstj: @benaadams not yet sure about usage of System.Memory in this library.  Also it seems like ReadOnlySpan still lets you get at the memory via `DangerousGetPinnableReference`.  I really don't want someone mucking with the dimensions.  I currently have a copy-on-first-read/cache semantic for the public API and I think I'll stick with that unless we can prove we have a safe alternative with benefits for the public API.

Drawaes: The dangerous get pinnable is like saying you don't want someone taking a fix and getting a pointer to the inside of the ireadonly list and changing the values. It's all bets are off at that point, but most devs just won't go there.

ericstj: My IReadOnlyList is a copy, so even if they do it they're just modifying the copy not the one that's used by my internals.  I just don't see the benefits I get using ReadOnlySpan over the current ReadOnlyCollection.  Perhaps @stephentoub can you cover this is some API guidelines around use of Span/ReadOnlySpan.  Code in question is here: https://github.com/dotnet/corefxlab/blob/661c2b2d92a43903818e01e301846f65b0f1ee15/src/System.Numerics.Tensors/System/Numerics/Tensor.cs#L241-L253

ericstj:
Last call for feedback on this, plan to merge it and move on to more issues.

PR #1693, opened 2017-08-10
Add the Owner property to BufferHandle

Drawaes:
This would stop having to carry around the buffer with the pinned handle. This is already carried around by the Handle anyway, so it seems a waste to have to have "BufferHandle" + "Buffer" be passed through when calling functions.

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/BufferHandle.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/BufferHandle.cs
@@ -22,6 +22,8 @@ namespace System.Buffers
 
         public BufferHandle(IRetainable owner) : this(owner, null) { }
 
+        public IRetainable Owner => _owner;
+
         public void* PinnedPointer {
             get {
                 if (_pointer == null) BufferPrimitivesThrowHelper.ThrowInvalidOperationException();


KrzysztofCwalina:
But now callers can get OwnedBuffers from Buffer\<T\>. This is not good. The idea behind the architecture is that OwnedBuffer\<T\> is never shared around and instead Buffer\<T\> is shared.

benaadams:
Would mean this could go away https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines.Networking.Windows.RIO/Internal/RioThread.cs#L94-L135

Drawaes:
So here is the scenario (psuedo code)
So I want to avoid pinning for each of the NativeCalls but instead take the hit once (imagine there is more than 1 native call). I could wrap the pointer into a span, but then I need to fix the span in each call as well. So I want to pass the handle, that means either
1. fixing a span that contains a pointer that is already on a fixed object is very close to free? If so forget this I don't care
1. I have to pass both around, which just seems very ugly
1. There is a size put on the bufferHandle?
1. There is an owner property on the bufferHandle
1. I have to pass both around which is pretty ugly

```
async Task UpperMethod()
{
    var pooledBuffer = FromSomePool();
    var bufferHandle = pooledBuffer.Pin();
    do(true)
    {
            var workNotDone = SomeNativeWorkThingThatWillNeedToPinBuffer(pooledBuffer);
            if(workNotDone)
            {
                 await SomeThingToGetMoreWork(pooledBuffer);
            }
            else
            {
                 break;
             }
     }
     bufferHandle.Release();
     //return buffer
}

```

ahsonkhan:
@dotnet-bot test this please

> But now callers can get OwnedBuffers from Buffer<T>. This is not good. The idea behind the architecture is that OwnedBuffer<T> is never shared around and instead Buffer<T> is shared.

@Drawaes, can you address @KrzysztofCwalina's point?

Drawaes:
I guess my view on it is, if you are pinning stuff and handing around a pin, it's got a pointer on it anyway, so I don't see it as an added risk that you could get access to the owned buffer....?

benaadams:
As long as its not directly the type e.g. `OwnedBuffer`; or a suggestion for people to as cast it e.g. `object`

Can it be non browsable/non-intellisense? (or `DangerousOwner`...)

ahsonkhan:
@Drawaes, is this change still relevant? If so, can you please open an issue in corefx for API review? Ty.

Closing this PR given the type has moved to corefx.

cc @benaadams, @KrzysztofCwalina

PR #1694, opened 2017-08-10
Add cancellation token to stream.CopyToAsync for IPipeWriter

BrennanConroy:
@pakrym @davidfowl 
Need this for SSE in SignalR

### DIFF
--- a/src/System.IO.Pipelines.Extensions/StreamExtensions.cs
+++ b/src/System.IO.Pipelines.Extensions/StreamExtensions.cs
@@ -62,11 +62,12 @@ namespace System.IO.Pipelines
         /// </summary>
         /// <param name="stream"></param>
         /// <param name="writer"></param>
+        /// <param name="cancellationToken"></param>
         /// <returns></returns>
         public static Task CopyToAsync(this Stream stream, IPipeWriter writer, CancellationToken cancellationToken = default(CancellationToken))
         {
             // 81920 is the default bufferSize, there is not stream.CopyToAsync overload that takes only a cancellationToken
-            return stream.CopyToAsync(new PipelineWriterStream(writer, cancellationToken), bufferSize: 81920, cancellationToken: cancellationToken);
+            return stream.CopyToAsync(new PipelineWriterStream(writer), bufferSize: 81920, cancellationToken: cancellationToken);
         }
 
         public static async Task CopyToEndAsync(this Stream stream, IPipeWriter writer, CancellationToken cancellationToken = default(CancellationToken))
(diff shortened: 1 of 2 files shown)


src/System.IO.Pipelines.Extensions/StreamExtensions.cs
```diff
@@ -218,11 +219,13 @@ public override int Read(byte[] buffer, int offset, int count)
 
         private class PipelineWriterStream : Stream
         {
-            private IPipeWriter _writer;
+            private readonly IPipeWriter _writer;
+            private readonly CancellationToken _cancellationToken;
 
-            public PipelineWriterStream(IPipeWriter writer)
+            public PipelineWriterStream(IPipeWriter writer, CancellationToken cancellationToken)
```
davidfowl: What is this token for?

src/System.IO.Pipelines.Extensions/StreamExtensions.cs
```diff
@@ -63,16 +63,17 @@ public static IPipeReader AsPipelineReader(this Stream stream, CancellationToken
         /// <param name="stream"></param>
         /// <param name="writer"></param>
         /// <returns></returns>
-        public static Task CopyToAsync(this Stream stream, IPipeWriter writer)
+        public static Task CopyToAsync(this Stream stream, IPipeWriter writer, CancellationToken cancellationToken = default(CancellationToken))
         {
-            return stream.CopyToAsync(new PipelineWriterStream(writer));
+            // 81920 is the default bufferSize, there is not stream.CopyToAsync overload that takes only a cancellationToken
+            return stream.CopyToAsync(new PipelineWriterStream(writer, cancellationToken), bufferSize: 81920, cancellationToken: cancellationToken);
```
davidfowl: Why are you passing the token twice?

BrennanConroy: Because streams are awful and the one passed to CopyToAsync is only used up until write is called internally. Then one passed to PipelineWriterStream is so you can actually cancel the write operation with the cancellation we pass to CopyToAsync

src/System.IO.Pipelines.Extensions/StreamExtensions.cs
BrennanConroy: Token is the one from `CopyToAsync`

BrennanConroy:
Ping

pakrym:
Test?

pakrym:
Create Pipe with size limit and try to copy stream that's longer then the limit so `FlushAsync` blocks. Then check if firing cancellation token works.

BrennanConroy:
🆙 📅

PR #1696, opened 2017-08-10
System.IO.Compression.Brotli: .NET Standard 1.4 -> 1.1

Taritsyn:
It would be nice in the `System.IO.Compression.Brotli` library to downgrade .NET Standard version from 1.4 to 1.1. In general, class libraries which target a lower [.NET Platform Standard](https://github.com/dotnet/standard/blob/master/docs/versions.md) version, can be loaded by the largest number of existing platforms. Key libraries are still targeted on .NET Standard 1.1  (e.g., [`System.IO.Compression`](https://www.nuget.org/packages/System.IO.Compression/)).


danmosemsft:
Ha- I didn't realize there were no code changes. @ianhays ? 

This isn't my library and I don't know the status of it. I assume it's on myget but not nuget.

shiftylogic:
This change looks fine to me.

PR #1698, opened 2017-08-10
Add support for column-major layout to tensor

### LINKED ISSUE #1695: Add support for column-major Tensor layout
This should be easier now that we're computing in terms of strides.

We should just be able to change the order of the stride calculation to support this and fix a few asserts.

ericstj:
Fixes #1695 

Still working through the details / design on this.  Submitting as a means for test coverage and feedback.


ericstj:
Merging this so that I can move on to refactor API for dense and sparse tensors.

PR #1692, opened 2017-08-09
Added System.Azure.Experimental Project

KrzysztofCwalina:
This adds a new corfxlab project where we want to experiment with building blocks for Azure SDK.

The initial commit contains an API to general CosmosDB authentication header without allocations.
The currently recommended way to generate the headers is described at https://docs.microsoft.com/en-us/rest/api/documentdb/access-control-on-documentdb-resources

The following is a result of a micro benchmark run comparing the currently recommended methods and the experimental corfxlab API:

| Method            | Mean    | Error      | StdDev   | Scaled | Gen 0 | Allocated |
|-------------------|---------:|----------:|----------:|-------:|-------:|----------:|
| SignatureGenerate | 2.236 us | 0.0009 us | 0.0007 us |   0.36 |      - |       0 B |
| MsdnSample        | 6.292 us | 0.0051 us | 0.0045 us |   1.00 | 0.7248 |    3056 B |

The improvements are mainly due to using non-allocating System.Text.Primitives APIs for text manipulation, Spans for slicing, non-allocating SHA256 generation, and native UTF8 text generation.

Note: Do not merge the PR. It does not work on Linux. I will have an update to make it work.

@AlexGhiondea, @joshfree, @shiftylogic, @ahsonkhan

PR #1702, opened 2017-08-11
Updating to a new CLI + CoreFx assemblies

shiftylogic:
* Had to patch all string.AsSpan calls to use string.AsReadOnlySpan now.
* We are forced to include a specific runtime version as a temporary workaround to out-of-sync issues with CLI, core-setup, and CoreFx (see bugs dotnet/cli#7157 and dotnet/corefx#23116).
* A side-effect of the above workaround is that the 2 projects that targeted netcoreapp1.1 needed to change to 2.0 to fix a run-time issue.
* All projects should be re-targeted to 2.1 once the above bugs are resolved.
* I also updated the "update-dependencies" tool so that Maestro will keep the runtime version in-sync with the CoreFx assemblies and CLI. In theory, PRs posted by Maestro should merge and build clean now.


shiftylogic:
@ahsonkhan @KrzysztofCwalina @joshfree

KrzysztofCwalina:
looks good

ahsonkhan:
> A side-effect of the above workaround is that the 2 projects that targeted netcoreapp1.1 needed to change to 2.0 to fix a run-time issue.

What about VS test discovery?
https://github.com/dotnet/corefxlab/pull/1692#discussion_r132338457

shiftylogic:
No idea. I don't tend to use VS. I suspect they disappear like our other tests. I could have solved this with extra work and flags in the project files. I'm hoping this workaround is only temporary (on the order of a couple weeks).

At a macro level, test discovery needs to be debugged and solved anyways.

PR #1708, opened 2017-08-17
Added CosmosDbAuthorizationHeader

KrzysztofCwalina:
Added CosmosDbAuthorizationHeader implementing IBufferFormattable, which makes it easy to write to various formattable buffers (as one of the tests illustrates).

@joshfree, @ahsonkhan, @shiftylogic

### DIFF
--- a/src/System.Azure.Experimental/System.Azure.Experimental.csproj
+++ b/src/System.Azure.Experimental/System.Azure.Experimental.csproj
@@ -8,7 +8,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Description>Tensor class which represents and extends multi-dimensional arrays</Description>
+    <Description>Set of building blocks for Azure SDK</Description>
     <TargetFramework>netstandard1.3</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageTags>BCL for Azure SDK </PackageTags>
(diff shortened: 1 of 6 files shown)


src/System.Azure.Experimental/System/Azure/Authentication.cs
```diff
@@ -210,24 +174,84 @@ public static bool TryWriteCosmosDbAuthorizationHeader(Span<byte> output, Sha256
...
+    public static class StorageAccessSignature
+    {
+        public static bool TryWrite(Span<byte> output, Sha256 hash, string verb, string canonicalizedResource, DateTime utc, out int bytesWritten)
+        {
             int written, consumed;
-            var buffer = new Span<byte>(pBuffer, bufferLength);
-            if (Utf16.ToUtf8(key.AsReadOnlySpan().AsBytes(), buffer, out consumed, out written) != TransformationStatus.Done)
+            bytesWritten = 0;
+
+            if (verb.Equals("GET", StringComparison.Ordinal))
```
joshfree: nit: can you handle the hard coded literals "GET" and its length 3 another way (I'm assuming you'll need to handle GET / PUT in several places)

joshfree: I see you fixed this in a new change that got pushed right when I wrote this.  :)

src/System.Azure.Experimental/System.Azure.Experimental.csproj
```diff
@@ -8,15 +8,17 @@
   </PropertyGroup>
 
   <PropertyGroup>
-    <Description>Tensor class which represents and extends multi-dimensional arrays</Description>
+    <Description>Set of building blocks for Azure SDK</Description>
     <TargetFramework>netstandard1.3</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageTags>BCL for Azure SDK </PackageTags>
     <!-- disable automatic compile item inclusion so that we can set the correct metadata on TextTemplate-generated sources -->
     <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
```
ahsonkhan: We don't need this attribute and hence can get rid of this and the explicit list of the *.cs files to include below.

PR #1712, opened 2017-08-22
UTF8String TrimStart with prefix

A-And:
Function and test implementation

### DIFF
--- a/src/System.Text.Utf8String/System/Text/Utf8String.cs
+++ b/src/System.Text.Utf8String/System/Text/Utf8String.cs
@@ -587,16 +587,16 @@ namespace System.Text.Utf8
             throw new NotImplementedException();
         }
 
-        public Utf8String TrimStart(Utf8String prefix)
+        public Utf8String TrimStart(Utf8String trimCharacters)
         {
-            if(prefix == Empty)
+            if (trimCharacters == Empty)
             {
                 // Trim Whitespace
                 return TrimStart();
             }
 
             CodePointEnumerator it = GetCodePointEnumerator();
-            CodePointEnumerator itPrefix = prefix.GetCodePointEnumerator();
+            CodePointEnumerator itPrefix = trimCharacters.GetCodePointEnumerator();
             
             while (it.MoveNext())
             {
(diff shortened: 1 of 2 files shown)


KrzysztofCwalina:

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -587,6 +587,43 @@ public Utf8String TrimStart(byte[] trimCodeUnits)
             throw new NotImplementedException();
         }
 
+        public Utf8String TrimStart(Utf8String prefix)
+        {
+            if(prefix == Empty)
```
ahsonkhan: nit: spacing between `if` and `(prefix == Empty)`

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -587,6 +587,43 @@ public Utf8String TrimStart(byte[] trimCodeUnits)
             throw new NotImplementedException();
         }
 
+        public Utf8String TrimStart(Utf8String prefix)
```
KrzysztofCwalina: This implementation removes any "char" in prefix from the start of the original string. There is an alternative implementation that removes the prefix if found at the beginning of the string, i.e.
a) "abad".TrimStart("ab") => "ad"
b) "abad".TrimStart("ab") => "d"

Which of these two should we implement? It might be worth asking the person who requested this API.

If the they wanted the currently implemented semantics (case b), then the parameter should probably not be called prefix.

A-And: "Prefix" isn't very semantically correct, true. 
I considered a managed string's TrimStart() as the canonical intended functionality. Wouldn't a) cause potential confusion between `string` and `Utf8String` 's use?

KrzysztofCwalina:

src/System.Text.Utf8String/System/Text/Utf8String.cs
ahsonkhan: Can you please add a test like "abad".TrimStart("ab") returns "d" to make it explicit that we are intentionally following the same behavior as string.TrimStart() and not going with option a)?

KrzysztofCwalina:
At some point, we should try to optimize it to not use the enumerators. I suspect the enumerators add too much overhead. But for now, the PR looks good to me. @shiftylogic, if you have no comments, please merge.

shiftylogic:
Agree with the need for these to be optimized. Not necessary now.

Issue #1711, opened 2017-08-21
Move Buffer<T> and friends to corlib, netstandard, system.runtime


KrzysztofCwalina:

KrzysztofCwalina:
Before the move, we also agreed to do the following cleanup:

1.	Cleanup span/buffer equality APIs (as discussed in review) <= Done (https://github.com/dotnet/corefxlab/pull/1646)
2.	Rename Buffer<T> to Memory<T>
3.	Remove offset from OwnedBuffer.Pin
4.	Remove OwnedBuffer.Add (will submit PR for this soon – using Unsafe.Add - https://github.com/dotnet/corefx/pull/20886)
5.	Remote IsRetainable and change Release return type to bool (PR needs to be updated - https://github.com/dotnet/corefxlab/pull/1648)

ahsonkhan:
> Span and Buffer ctros should be the same. See dotnet/corefx#23471

Is this work item solely about removing Memory(T[], int) ctor or are there other constructors we want to add/remove?

ahsonkhan:
@KrzysztofCwalina, can we consider this work/issue complete? The extension work can be tracked by the api review issue - https://github.com/dotnet/corefx/issues/23862

KrzysztofCwalina:
Yeah, go ahead and close it. Just make sure the extensions issue does not get lost.

PR #1715, opened 2017-08-22
Remove offset from OwnedBuffer.Pin

ahsonkhan:
Remove offset from OwnedBuffer.Pin
Addresses part of https://github.com/dotnet/corefxlab/issues/1711

cc @davidfowl, @KrzysztofCwalina, @shiftylogic

### DIFF
--- a/tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
+++ b/tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
@@ -20,6 +20,8 @@ namespace System.IO.Pipelines.Tests
 {
     public class ReadableBufferFacts
     {
+        const int BlockSize = 4032;
+
         [Fact]
         public void EmptyIsCorrect()
         {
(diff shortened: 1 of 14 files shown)


tests/System.IO.Pipelines.Tests/ReadableBufferFacts.cs
```diff
@@ -262,7 +262,7 @@ private unsafe void TestIndexOfWorksForAllLocations(ref ReadableBuffer readBuffe
                 var ptr = (byte*)handle.PinnedPointer;
                 for (int i = 0; i < memory.Length; i++)
                 {
-                    addresses[index++] = ptr++;
+                    addresses[index++] = ptr;
```
davidfowl: What's the deal with this? Doesn't it invalidate the test?

ahsonkhan: That was a mistake which I have reverted.
With Pin no longer taking an index, the test needs to be modified.

PR #1719, opened 2017-08-24
Structural changes to text formatting primitive APIs.

shiftylogic:
* Part 1 of the effort to finalizing the text primitive APIs.
* Moving and renamed several files.
* Moved primitive formatting APIs into the System.Text.Formatters namespace.
* Direct access to encoding specific APIs via Utf8 and Utf16 classes.
* Generic encoding usage now done through a "Custom" class that allows swapping in a specific symbol table.
* TryFormat method extensions still exist for supported types via an "Extensions" class by including just the "System.Text" namespace. This probably isn't final. However, I wanted to avoid a huge change to other assemblies for this PR.

NOTE: There should be little (zero?) logic changes in this PR.

cc: @ahsonkhan @KrzysztofCwalina


shiftylogic:

src/System.Text.Primitives/System/Text/Extensions.Float.cs
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Text
+{
+    public static partial class Extensions
+    {
+        public static bool TryFormat(this double value, Span<byte> buffer, out int bytesWritten, ParsedFormat format = default, SymbolTable symbolTable = null)
```
ahsonkhan: Split this to Extensions.Double.cs?

shiftylogic: Float implies "floating-point". That would encapsulate double, float and possible decimal at some point.

src/System.Text.Primitives/System/Text/Formatters/Utf16.cs
```diff
@@ -0,0 +1,192 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Text.Formatters
+{
+    public static partial class Utf16
+    {
+        #region Common constants
+
+        private const char Colon = ':';
```
ahsonkhan: Can you please explain how you decide what constants get moved to "common constants" here? Why is Colon here and why isn't Separator in Utf16.Integer?

shiftylogic: For this change, because I used partial classes, I moved anything that was "common" and used in multiple places into here... plus ones that I foresee as probably being needed in more than 1 type format.

shiftylogic: That's not to say that it was the "right" way to break it up. They are all private anyways, so we can move them however we want.

ahsonkhan: Fair enough. That makes sense to me.

PR #1718, opened 2017-08-24
Utf8 String TrimEnd 

A-And:
Follow-up to PR #1712 . Testing and implementation of TrimEnd.

Issue #1510, opened 2017-05-01
Utf8String: desired features


migueldeicaza:
Would like to have the following features, ran into these while porting some Go code:

* `LastIndexOf (uint codepoint)`
* `TrimStart (Utf8String prefix)`
* `TrimEnd (Utf8String suffix)`

A-And:
@migueldeicaza 
Could you clarify the `TrimStart ` and `TrimEnd ` functionality? 

Were you thinking of Go's `TrimPrefix` and `TrimSuffix` functions (i.e. removing a specified prefix/suffix if present, string is unchanged if not) or more of a managed string's `TrimStart ` and `TrimEnd` (i.e. remove any occurrence of a character in a set passed as an argument)?

KrzysztofCwalina:
This APIs were added by @A-And and merged into master.

PR #1721, opened 2017-08-25
Utf8String LastIndexOf() 

A-And:
Related to #1712 and #1718 for Feature request #1510 
Implemented for substrings, codepoints and code units.
Apologies for the larger PR.


tests/System.Text.Utf8String.Tests/RandomTests.cs
```diff
@@ -631,6 +631,37 @@ public void IndexOfTests(int expected, string s, string substring)
         }
 
         [Theory]
+        [InlineData(0, "", "")]
+        [InlineData(0, "abc", "")]
```
KrzysztofCwalina: I tried it on .NET ( ```c# Console.WriteLine("abc".LastIndexOf(""));``` ) and it prints 2. I don't think it makes sense, but maybe we should be compatible? @stephentoub, @terrajobst, @joshfree, any opinions on this?

tests/System.Text.Utf8String.Tests/RandomTests.cs
```diff
@@ -631,6 +631,37 @@ public void IndexOfTests(int expected, string s, string substring)
...
+        [InlineData(38, "aaaaaaabaaaaaaaabaaaaaaaabaaaaaaaabaaaaaaaaaaab", "aaaaaaaab")]
+        [InlineData(16, "abbbbbbbabbbbbbbabbbbbbbbbbbbbbbbbbbbbbbbbbbb", "abbbbbbbb")]
+        [InlineData(-1, "aaabaaa", "aabaaaa")]
+        [InlineData(-1, "aaabaaa", "aadaaaa")]
+        [InlineData(3, "ababababa", "bababa")]
+        [InlineData(-1, "abababab", "bbababa")]
+        [InlineData(-1, "abababab", "bababaa")]
+        [InlineData(7, "aaabaacaaac", "aaac")]
+        [InlineData(9, "baaaabaaabaaaa", "baaaa")]
+        [InlineData(0, "\uABC0\uABC1\uABC2", "\uABC0\uABC1")]
```
KrzysztofCwalina: Could you add a case that uses codeponts from outside of the BMP, but one that does not return 0? Thanks!

tests/System.Text.Utf8String.Tests/RandomTests.cs
danmosemsft: Unless there's a good reason to not be compatible (eg significant perf impact) why would we not be intentionally compatible...

KrzysztofCwalina: Because I am not sure how returning 2 makes any sense here. :-)

KrzysztofCwalina:

KrzysztofCwalina:
Thanks!

PR #1725, opened 2017-08-29
Remove use of Span ctor(T[], int) which no longer exists

ahsonkhan:
Also remove duplicate test cases

cc @KrzysztofCwalina

### DIFF
--- a/tests/System.Text.Primitives.Tests/Parsing/PrimitiveParserIntegerTests.cs
+++ b/tests/System.Text.Primitives.Tests/Parsing/PrimitiveParserIntegerTests.cs
@@ -685,15 +685,10 @@ namespace System.Text.Primitives.Tests
         [InlineData("-0128", true, -128, 5)] // min
         [InlineData("0128", false, 0, 0)] // positive overflow test
         [InlineData("-0129", false, 0, 0)] // negative overflow test
-        [InlineData("111", true, 111, 3)]
         [InlineData("49abcdefg", true, 49, 2)]
         [InlineData("127", true, 127, 3)] // max
         [InlineData("-128", true, -128, 4)] // min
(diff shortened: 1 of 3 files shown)


ahsonkhan:
@KrzysztofCwalina, this is a fairly benign PR. Should I go ahead and merge it?

tests/System.Text.Primitives.Tests/Parsing/PrimitiveParserIntegerTests.cs
```diff
@@ -685,15 +685,10 @@ public unsafe void ParseUInt64Hex(string text, bool expectSuccess, UInt64 expect
         [InlineData("-0128", true, -128, 5)] // min
         [InlineData("0128", false, 0, 0)] // positive overflow test
         [InlineData("-0129", false, 0, 0)] // negative overflow test
-        [InlineData("111", true, 111, 3)]
```
KrzysztofCwalina: Why are these changes here?

KrzysztofCwalina:
looks good, if the tests are really duplicate.

tests/System.Text.Primitives.Tests/Parsing/PrimitiveParserIntegerTests.cs
ahsonkhan: Yes, these test cases are duplicate. It was observed when VS test discovery stated explicitly that the tests have already been added to the test window and hence ignoring it.

Line 688 is: `[InlineData("111", true, 111, 3)]`
Which is the same as line 649 - https://github.com/dotnet/corefxlab/pull/1725/files/d6f7d417dcd26ee92a6f13795ee58273fe24bf26#diff-de2a92cba24dadec8a492537ca278d1bL649

PR #1737, opened 2017-08-30
Correct check direction.

Clockwork-Muse:


KrzysztofCwalina:
Thanks! The change looks good.

ahsonkhan:
We should add unit tests around the equality operators to catch such issues. @Clockwork-Muse, please feel free to do so, if you like :)

PR #1739, opened 2017-09-01
Updating to new toolset compiler.

VSadov:
Version 2.6.0-rdonly-ref-62106-01

Now includes:
- support for stackallocated spans and 
- escape analysis to make sure spans that refer to local data do not outlive the referenced data.

The corresponding VSIX is at 
https://dotnet.myget.org/feed/roslyn/package/vsix/0b48e25b-9903-4d8b-ad39-d4cca196e3c7/2.6.0.6210601

### DIFF
--- a/src/System.Text.Http.Parser/HttpParser.cs
+++ b/src/System.Text.Http.Parser/HttpParser.cs
@@ -23,7 +23,7 @@ namespace System.Text.Http.Parser
         private const byte ByteQuestionMark = (byte)'?';
         private const byte BytePercentage = (byte)'%';
 
-        public unsafe bool ParseRequestLine<T>(T handler, ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined) where T : IHttpRequestLineHandler
+        public unsafe bool ParseRequestLine<T>(T handler, in ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined) where T : IHttpRequestLineHandler
         {
             consumed = buffer.Start;
             examined = buffer.End;
@@ -42,7 +42,7 @@ namespace System.Text.Http.Parser
             }
             else
             {
-                span = TryGetNewLineSpan(ref buffer, out consumed);
+                span = TryGetNewLineSpan(buffer, out consumed);
                 if (span.Length == 0)
                 {
                     // No request line end
@@ -61,7 +61,7 @@ namespace System.Text.Http.Parser
         }
 
         static readonly byte[] s_Eol = Encoding.UTF8.GetBytes("\r\n");
-        public unsafe bool ParseRequestLine<T>(ref T handler, ReadOnlyBytes buffer, out int consumed) where T : IHttpRequestLineHandler
+        public unsafe bool ParseRequestLine<T>(ref T handler, in ReadOnlyBytes buffer, out int consumed) where T : IHttpRequestLineHandler
         {
             // Prepare the first span
             var span = buffer.First.Span;
@@ -95,14 +95,13 @@ namespace System.Text.Http.Parser
 
         private unsafe void ParseRequestLine<T>(T handler, byte* data, int length) where T : IHttpRequestLineHandler
         {
-            int offset;
-            ReadOnlySpan<byte> customMethod;
             // Get Method and set the offset
-            var method = HttpUtilities.GetKnownMethod(data, length, out offset);
-            if (method == Http.Method.Custom)
-            {
-                customMethod = GetUnknownMethod(data, length, out offset);
-            }
+            var method = HttpUtilities.GetKnownMethod(data, length, out var offset);
+
+            ReadOnlySpan<byte> customMethod =
+                method == Http.Method.Custom ?
+                    GetUnknownMethod(data, length, out offset) :
+                    default;
 
             // Skip space
             offset++;
@@ -212,7 +211,7 @@ namespace System.Text.Http.Parser
             handler.OnStartLine(method, httpVersion, targetBuffer, pathBuffer, query, customMethod, pathEncoded);
         }
 
-        public unsafe bool ParseHeaders<T>(T handler, ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined, out int consumedBytes) where T : IHttpHeadersHandler
+        public unsafe bool ParseHeaders<T>(T handler, in ReadableBuffer buffer, out ReadCursor consumed, out ReadCursor examined, out int consumedBytes) where T : IHttpHeadersHandler
         {
             consumed = buffer.Start;
             examined = buffer.End;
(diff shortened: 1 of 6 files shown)


VSadov:
@stephentoub @jaredpar   -   FYI

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -42,7 +42,7 @@ public class HttpParser : IHttpParser
             }
             else
             {
-                span = TryGetNewLineSpan(ref buffer, out consumed);
+                span = TryGetNewLineSpan(buffer, out consumed);
                 if (span.Length == 0)
```
VSadov: if buffer is an ordinary parameter, passing it by ref results in a span that cannot be returned.
I made it an "in" in the containing method to widen its escape.
Made it "in" in couple more places. 
ReadableBuffer is a huge struct, I guess that was a reason why it was passed by ref to TryGetNewLineSpan, since otherwise it is no being mutated there.

src/System.Text.Http/System/Text/Http/Obsolete/HttpHeaderBuffer.cs
```diff
@@ -39,64 +39,62 @@ public static void SetFromRestOfSpanToEmpty(this Span<byte> span, int startingFr
             }
         }
 
-        internal static ReadOnlySpan<byte> SliceTo(this ReadOnlySpan<byte> buffer, char terminator, out int consumedBytes)
```
VSadov: Here I am mostly replacing pattern  
` Span Method(..., out int) `
into 
` int Method(..., out Span) `

The first form easily gets into trouble when a ordinary local is passed to the out variable since that leads to unreturnable result.  (since we must assume the resulting span is possibly referring to the local that we passed in).

src/System.Text.Http/System/Text/Http/Obsolete/HttpRequestParser.cs
```diff
@@ -188,26 +188,24 @@ static class HttpRequestParser
         static Utf8String s_Delete => new Utf8String(b_Delete);
 
 
-        internal static bool TryParseRequestLine(ReadOnlySpan<byte> buffer, out HttpRequestLine requestLine, out int totalParsedBytes)
```
VSadov: Here we have a form ` bool TryParse(..., out Span, out int) `
It easily gets in trouble when we pass a returnable span (which we actually intend to return) and also a ref to an ordinary int local. Callee may stash the ref into the span and that would be bad, so this triggers an error.

I also noticed that in all cases returning `false` is matched with bytes being 0, so I just changed this to a form

` int TryParse(..., out Span) ` where returning 0 indicates a failure.

src/System.Text.Json/System/Text/Json/JsonReader.cs
```diff
@@ -97,7 +97,7 @@ public bool Read()
             ref byte next = ref Unsafe.Add(ref bytes, skip);
             length -= skip;
 
-            int step = GetNextCharAscii(ref next, length, out char ch);
+            int step = GetNextCharAscii(this, ref next, length, out char ch);
             if (step == 0) return false;
```
VSadov: This one is interesting.
`JsonReader` is a mutable struct and `GetNextCharAscii` is a formally mutating method, even though it does not mutate.
No-mixing rules assume here that the callee may wrap the `ch` in a span, directly or indirectly and "poison" `this`, with the method can do, since this is writable.

I could solve this by making `GetNextCharAscii` a readonly method, but we do not support methods that are individually readonly.

I could not make it a 'in this' extension either since it reads private members and I did not want to add another class, although that could be the least disruptive solution if we had a lot of callers.

So I made it a static method that takes the struct as `in` parameter.

ahsonkhan:
cc @KrzysztofCwalina, @shiftylogic, @davidfowl

ahsonkhan:
@dotnet-bot test this please

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -95,14 +95,13 @@ public class HttpParser : IHttpParser
...
-            ReadOnlySpan<byte> customMethod;
             // Get Method and set the offset
-            var method = HttpUtilities.GetKnownMethod(data, length, out offset);
-            if (method == Http.Method.Custom)
-            {
-                customMethod = GetUnknownMethod(data, length, out offset);
-            }
+            var method = HttpUtilities.GetKnownMethod(data, length, out var offset);
+
+            ReadOnlySpan<byte> customMethod =
```
ahsonkhan: Was there a particular reason for this change or just code clean up?

VSadov: GetUnknownMethod was taking a local by ref/out and returning a ref-like. That needed to be reversed to make result returnable.  
GetKnownMethod returned ordinary enum so did not need the change. I just made it for consistency with Unknown.

src/System.Text.Http/System/Text/Http/Obsolete/HttpRequestParser.cs
KrzysztofCwalina: How can the ref be stashed into the span? The span is of bytes. If you stash anything into it, it will just copy the byte from the ref location.

Also, what's the error's message? If we have to keep the error, we should make sure it explains what's going on.

KrzysztofCwalina:
Some of the tricks you had to use are scary, i.e. I wonder how many developers will understand what's going on and how to deal with the compiler errors. Also, I hope we won't have many/any of such issues once we build corfx with this new compiler.

@stephentoub, @MadsTorgersen, @jaredpar

src/System.Text.Http/System/Text/Http/Obsolete/HttpRequestParser.cs
VSadov: Copying the bytes is not a problem. Creating a one-element span that wraps the given reference is a problem though. 
We do not know what happens inside TryParseRequestLine and must assume the worst case scenario. The worst case scenario is that TryParseRequestLine does:

```C#
bool TryParseRequestLine(ReadOnlySpan<byte> buffer, out HttpRequestLine requestLine, out int totalParsedBytes)
{
    totalParsedBytes = 42;
    requestLine = WrapInOneElementSpan(ref totalParsedBytes);
    return true;
}
```
As a result if you pass a returnable `out requestLine` together with a local-referring `out totalParsedBytes` you get an error about mixing references that have different escape scopes.

If an error is not given, you may end up returning `requestLine` that refers to stack memory, which is basically a GC hole.

The errors could be improved, but it is correct to give an error here. 
Ref-like types are considered as containing references and when they are ref/out parameters they are also assignable by the callee.
As a result mixing a ref/out ref-like argument with references that have narrower life time in the same invocation is the same as assigning one into another and that is illegal.

VSadov:
@KrzysztofCwalina - the problem arises when you have a method that:    
- returns a ref-like type (directly or via ref/out parameter) and also 
- takes a ref/out/in parameter of an ordinary type. 

Then you can get in trouble with escaping rules when passing a local to the ref/out/in.

Basically - If you are producing spans or ref-likes in general, do not take ordinary byrefs.  
(unless you actually intend to wrap them like `Span<int> Wrap(ref int element)` )

FWIW we had just 1 error because of this in Kestrel, so it is hard to tell if this is common.   
see: https://github.com/aspnet/KestrelHttpServer/pull/2043

VSadov:
I think this is ready to be merged.
I will reset it couple more times just to make sure it consistently passes.

VSadov:
test this please

VSadov:
test this please

VSadov:
@ahsonkhan @KrzysztofCwalina - please review

ahsonkhan:
cc @shiftylogic

PR #1742, opened 2017-09-07
Use safe stackalloc where possible

VSadov:


src/System.Text.Formatting/System/Text/Parsing/SequenceParser.cs
```diff
@@ -37,47 +37,48 @@ public static class TextSequenceExtensions
...
+                    {
+                        return true;
                     }
                 }
             }
 
             // for invariant culture, we should never reach this point, as invariant uint text is never longer than 127 bytes. 
             // I left this code here, as we will need it for custom cultures and possibly when we shrink the stack allocated buffer.
-            combinedSpan = bufferSequence.ToSpan();
-            if (!PrimitiveParser.InvariantUtf8.TryParseUInt64(first.Span, out value, out consumed)) {
```
VSadov: Was 'first.Span' a bug here?

VSadov:
@KrzysztofCwalina @ahsonkhan @shiftylogic  - please review.

VSadov:
@stephentoub @jaredpar  - FYI. Using safe `stackalloc` in CoreFxLab.

src/System.Text.Formatting/System/Text/Parsing/SequenceParser.cs
ahsonkhan: I am not sure but it seems like it given combinedSpan is assigned and then not used.
cc @KrzysztofCwalina

ahsonkhan:

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -16,8 +16,11 @@ public static void Pipe(this Transformation transformation, ReadOnlyBytes source
         {
             int afterMergeSlice = 0;
 
+            // TODO: "ReadOnlySpan<byte> remainder = stackalloc byte[0]" would fit better here, 
+            //       but it emits substandard IL, see https://github.com/dotnet/roslyn/issues/21952
+            //
             // make 'remainder' formally stack-referring or we won't be able to reference stack data later
-            ReadOnlySpan<byte> remainder = stackalloc byte[0]; 
+            var remainder = true ? new ReadOnlySpan<byte>() : stackalloc byte[0];
```
ahsonkhan: nit: change from var to ReadOnlySpan<byte>, esp since in the future, the r.h.s would look like `stackalloc byte[0];`

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
```diff
@@ -210,7 +210,11 @@ public static string GetUtf8String(this ReadableBuffer buffer)
                 return null;
             }
 
-            ReadOnlySpan<byte> textSpan = stackalloc byte[0];
+            // TODO: "ReadOnlySpan<byte> textSpan = stackalloc byte[0]" would fit better here, 
+            //       but it emits substandard IL, see https://github.com/dotnet/roslyn/issues/21952
+            //
+            // make 'textSpan' formally stack-referring or we won't be able to reference stack data later
```
ahsonkhan: I know you didn't write the comment @VSadov (copy/pasted), but I don't understand it.

What does it mean by 'formally stack-referring'. Also, shouldn't it read:
make 'textSpan' formally stack-referring or **we will be able to** reference stack data later 

If textSpan is a stackalloc variable, we can't reference it outside the method, correct? The comment seems to suggest that if we don't make it 'stack-referring', we won't be able to reference it later, which has the opposite meaning.

davidfowl:
This is pretty nice cleanup

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -325,21 +324,11 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyBu
...
-                }
-                else
-                {
-                    // TODO (pri 3): I think this could be eliminated by chunking values
-                    combined = new byte[combinedBufferLength];
-                }
+                Span<byte> combined = combinedBufferLength < 128 ?
+                                        stackalloc byte[combinedBufferLength] :
+                                        // TODO (pri 3): I think this could be eliminated by chunking values
+                                        combined = new byte[combinedBufferLength];
```
stephentoub: Is this "combined =" on purpose or a mistake?

VSadov: typo, it is a redundant assignment

src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs
VSadov: If we do not assign anything, 'textSpan' will default to be a returnable variable. We will be able to return it from the method, but will not be able to assign stackalloc'd spans. That is what you normally want, but not in this case. Here we need to initialize to something stackallocated to intentionally "taint" the variable as "possibly referring to stack data". 

'stackalloc byte[0]' would be a nice choice since it is basically the same as 'default(Span<T>)', but formally stack-allocated. However we do a poor job with optimizing such trivial case, so I am using a more verbose, but reliable, way to essentially assign a default span, while forcing the variable be classified as stack-referring.

I will try to change the comment to be more clear.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
VSadov: Will do. I was not sure if mentioning the type twice would be too verbose.

src/System.Azure.Experimental/System/Azure/CosmosDbAuthorizationHeader.cs
```diff
@@ -35,12 +35,7 @@ public static bool TryWrite(Span<byte> output, Sha256 hash, string keyType, stri
             int written, consumed, totalWritten = 0;
             bytesWritten = 0;
 
-            Span<byte> buffer;
-            unsafe
-            {
-                var pBuffer = stackalloc byte[AuthenticationHeaderBufferSize];
-                buffer = new Span<byte>(pBuffer, AuthenticationHeaderBufferSize);
-            }
+            Span<byte> buffer = stackalloc byte[AuthenticationHeaderBufferSize];
```
KrzysztofCwalina: Nice!

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -325,21 +325,11 @@ internal static int IndexOfStraddling(this ReadOnlySpan<byte> first, IReadOnlyBu
...
-                        byte* temp = stackalloc byte[combinedBufferLength];
-                        combined = new Span<byte>(temp, combinedBufferLength);
-                    }
-                }
-                else
-                {
-                    // TODO (pri 3): I think this could be eliminated by chunking values
-                    combined = new byte[combinedBufferLength];
-                }
+                var combined = combinedBufferLength < 128 ?
```
KrzysztofCwalina: I think we should settle on the following idiom:
```c#
Span<byte> buffer = size < 128 ? stackalloc byte[size] : new byte[size];
```

.. as opposed to assigning to var and then having to cast.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -15,14 +15,14 @@ public static class BufferExtensions
...
-                byte* stackBytes = stackalloc byte[stackLength];
-                stackSpan = new Span<byte>(stackBytes, stackLength);
-            }
+            // TODO: "ReadOnlySpan<byte> remainder = stackalloc byte[0]" would fit better here, 
+            //       but it emits substandard IL, see https://github.com/dotnet/roslyn/issues/21952
+            //
+            // Assign 'remainder' to something formally stack-referring.
+            // The default classification is "returnable, not referring to stack", we want the opposite in this case.
+            ReadOnlySpan<byte> remainder = true ? new ReadOnlySpan<byte>() : stackalloc byte[0];
+            Span<byte> stackSpan = stackalloc byte[stackLength];
```
KrzysztofCwalina: Will this be fixed before we ship?

src/System.Text.Formatting/System/Text/Parsing/SequenceParser.cs
KrzysztofCwalina: I would not be shocked if this codepath never executed (as the comment implies)

KrzysztofCwalina:
Wow, this is such a groundbreaking change to our platform where features in the run-time, libraries, and compilers come together so nicely.

@raja-krishnan, @MeiChin-Tsai

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
shiftylogic: Can we just use ReadOnlySpan<byte>.Empty here?

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
VSadov: As it stands now the dominant type of a ternary expression is selected from one of consequence or alternative branches.  

If you have a span as a type of either of the branches, it will dominate. Otherwise compiler faces an ambiguity since no dominant type can be chosen.   

I will bring this up to language folks.  
There are some tradeoffs that we can have around the natural type of `stackalloc`.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
VSadov: @KrzysztofCwalina - yes, will fix this before shipping. The trivial `stackalloc byte[0]` is emitted "correctly", but the IL is embarrassingly redundant, even worse after JIT-ting.  

Optimizing this case in C# compiler backend will benefit existing code as well, so the the value vs. effort is very favorable here.

VSadov: @shiftylogic - if you use ReadOnlySpan.Empty or any other _not_ stack allocated data to initialize the span you are telling the compiler - "I want to be able to return this from the method or mix with other returnables, please make sure I do not refer to stack data by accident".  
That is generally the case.

Here we have a relatively rare opposite - We are slicing through stack allocated data and store in this variable, but have no intent to return it. 
 `stackalloc byte[0]` will convey to the compiler - "please allow referring to stack data, make sure I can never return this from the method"

omariom:
@KrzysztofCwalina 

>Wow, this is such a groundbreaking change to our platform where features in the run-time, libraries, and compilers come together so nicely.

The most impressive was how quickly the combined efforts of all three teams made it a real thing. A year ago it was a  prototype in CorefxLab.

Issue #1723, opened 2017-08-27
Add CancellationToken to ReadableChannel<T>


davidfowl:
So that I don't have to convert this task (https://github.com/dotnet/corefxlab/blob/1d690999ac3d4010b282def0cb946bea72a214a3/src/System.Threading.Tasks.Channels/System/Threading/Tasks/Channels/ReadableChannel.cs#L22) to a `CancellationToken` (related https://github.com/dotnet/corefx/issues/23590)

stephentoub:
@davidfowl, can you be more specific?  You want to pass a CancellationToken to a ReadableChannel?  You want ReadableChannel to expose a CancellationToken that's canceled when it completes?  Something else?  And specific to ReadableChannel?

davidfowl:
> You want ReadableChannel to expose a CancellationToken that's canceled when it completes?

This. Specific to ReadableChannel because that's where the Task that represents completion exists today.

stephentoub:
Thanks.  I'm hesitant to add another "this is completed property", but I'll think about it some more.  In the meantime, it's of course possible to add one's own Task->CancellationToken conversion: https://github.com/dotnet/corefx/issues/23590#issuecomment-325975594

Issue #1127, opened 2017-01-15
Add BoundedChannelFullMode.Ignore for Channel Writing


shaggygi:
It would be nice to ignore items written to a channel if it has already reached its buffered capacity.  Would by adding another enum of BoundedChannelFullMode.Ignore help this scenario?

```csharp
Channel<int> myChannel = Channel.CreateBounded<int>(3, BoundedChannelFullMode.Ignore);
myChannel.Out.WriteAsync(1);
myChannel.Out.WriteAsync(2);
myChannel.Out.WriteAsync(3);
myChannel.Out.WriteAsync(4);  // This should be ignored and never added to queue.
var value = myChannel.In.ReadAsync().Result; // value = 1.
// myChannel should include 2 items (2, 3).
```

Using the Wait enum in example above will still add 4 to queue after ReadAsync() and myChannel includes 3 items (2, 3, 4).

I was comparing this to TDF BufferBlock like below:

```csharp
BufferBlock<int> bufferBlock = new BufferBlock<int>(new DataflowBlockOptions { BoundedCapacity = 3 });
bufferBlock.SendAsync(1);
bufferBlock.SendAsync(2);
bufferBlock.SendAsync(3);
bufferBlock.SendAsync(4);  // This is ignored and never added to queue.
var bufferBlockValue = bufferBlock.Receive();
// bufferBlockValue includes 2 items (2, 3).
```

stephentoub:
This is a reasonable thing to add; I'll do so. 

However, the comment about dataflow isn't correct. That last SendAsync's value will be added once room is available, at which point is returned task will complete successfully, just like with WriteAsync with the default options on the bounded channel.

shaggygi:
Gotcha.  I missed this by breakpointing and didn't give a chance to add last SendAsync.  Thanks for reviewing.

shaggygi:
Thanks @stephentoub

Issue #1731, opened 2017-08-29
Productize System.Threading.Tasks.Channels


KrzysztofCwalina:
- [x] Clean up API
- [x] Conduct API review
- [ ] Move the package to corfx repo and publish package on nuget
- [ ] Performance work
- [ ] More concurrency-focused testing
- [x] Delete corfxlab implementation

PR #1743, opened 2017-09-07
Initial clean up of System.Threading.Tasks.Channels

stephentoub:
In support of https://github.com/dotnet/corefxlab/issues/1731, some clean up work on Channels, including removing experimental API surface area (e.g. CaseBuilder, IAsyncEnumerator, etc.), which can be added back later when the support is ready, and removing from the public surface area internal dependencies that would be better off in a lower-level library (though some of the public surface area still there should likely end up in System.Threading.Tasks.Extensions, or something like that).  Also includes some perf work, but there's more of that to be done separately.

Fixes https://github.com/dotnet/corefxlab/issues/1127
Fixes https://github.com/dotnet/corefxlab/issues/1723
Contributes to https://github.com/dotnet/corefxlab/issues/1731


src/System.Threading.Tasks.Channels/System/Threading/Tasks/Channels/WritableChannel.cs
```diff
@@ -18,9 +18,9 @@ public abstract class WritableChannel<T>
         /// <param name="error">An <see cref="Exception"/> indicating the failure causing no more data to be written, or null for success.</param>
         /// <returns>
         /// true if this operation successfully completes the channel; otherwise, false if the channel could not be marked for completion,
-        /// for example due to having already been marked as such.
+        /// for example due to having already been marked as such, or due to not supporting completion.
         /// </returns>
-        public abstract bool TryComplete(Exception error = null);
+        public virtual bool TryComplete(Exception error = null) => false;
```
KrzysztofCwalina: Why this change?

stephentoub: It's essentially saying that the default is an infinite channel; Completion returns a never-completing task by default, and TryComplete returns false to say you can't complete it.  If a derived channel wants to override it, it of course can, and do so in a way that appropriately coordinates between the read side and the write side.  We could go back to having them both be abstract, but it seemed like an unnecessary additional thing to be forced to override in order to implement a custom channel.

I don't feel strongly about it, though.  Happy to change it back if you disagree.

KrzysztofCwalina: Ah, thanks. The change makes sense.

src/System.Threading.Tasks.Channels/System/Threading/Tasks/Channels/ReadableChannel.cs
```diff
@@ -192,52 +223,5 @@ public void Dispose()
                 }
             }
         }
-
-        /// <summary>Gets an async enumerator of the data in this channel.</summary>
-        /// <param name="cancellationToken">The cancellation token to use to cancel the asynchronous enumeration.</param>
-        /// <returns>The async enumerator.</returns>
-        public virtual IAsyncEnumerator<T> GetAsyncEnumerator(CancellationToken cancellationToken = default) =>
```
BrennanConroy: @davidfowl @anurse We use this for streaming in SignalR. Looks easy enough to make into an extension method for our internal use until a public API is added back.

@stephentoub Is there an ETA for this support, like 2.1?

stephentoub: I removed it because I don't want it to conflict with the real `IAsyncEnumerable<T>/IAsyncEnumerator<T>` we intend to deliver along with C# 8.  I do not expect this to be in 2.1.
https://github.com/dotnet/csharplang/blob/master/proposals/async-streams.md

anurse: Ok, I think we'll probably just copy this extension method into our code base as internal code (and plan to replace it with the real thing when that comes out).

stephentoub: Sounds good.  Thanks.

davidfowl: ZOMG 😄

PR #1748, opened 2017-09-10
Use ReferenceEquals not Equals for delegate ref equals

benaadams:
[MulticastDelegate.Equals](https://github.com/dotnet/coreclr/blob/888b7a94f7b684f988110b9f0b8f3f6ea49746e1/src/mscorlib/src/System/MulticastDelegate.cs#L57-L132) is a chunky non-inlinable function to call just to test for reference equality


benaadams:
Exciting change

/cc @pakrym @davidfowl

PR #1749, opened 2017-09-11
Fix some ill-formatted XML comments.

AtsushiKan:
Kestrel actually treats this a fatal build break which makes it
annoying to copy code over for testing.

Then again, so does corefx...


KrzysztofCwalina:
looks good.

PR #1750, opened 2017-09-11
Changing Base64 APIs to follow same pattern as text encoders

KrzysztofCwalina:
The new APIs are as follows:

```c#
    public static class Base64 {
        public static readonly BufferDecoder Utf8ToBytesDecoder;
        public static readonly BufferEncoder BytesToUtf8Encoder;

        public static OperationStatus BytesToUtf8(ReadOnlySpan<byte> bytes, Span<byte> utf8, out int consumed, out int written);
        public static OperationStatus BytesToUtf8InPlace(Span<byte> buffer, int bytesLength, out int written);
        public static int BytesToUtf8Length(int bytesLength);

        public static OperationStatus Utf8ToBytes(ReadOnlySpan<byte> utf8, Span<byte> bytes, out int consumed, out int written);
        public static OperationStatus Utf8ToBytesInPlace(Span<byte> buffer, out int consumed, out int written);
        public static int Utf8ToBytesLength(ReadOnlySpan<byte> utf8);
    }
```

@ahsonkhan, @shiftylogic, @terrajobst, @joshfree

### DIFF
--- a/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Buffers;
+using System.ComponentModel;
 using System.Runtime.CompilerServices;
 
 namespace System.Binary.Base64
@@ -28,8 +29,9 @@ namespace System.Binary.Base64
             -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
         };
 
+        [Obsolete("Use Base64.Utf8ToBytesLength")]
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static int ComputeDecodedLength(ReadOnlySpan<byte> source)
+        public static int ComputeDecodedUtf8Length(ReadOnlySpan<byte> source)
         {
             int srcLength = source.Length;
 
@@ -84,6 +86,8 @@ namespace System.Binary.Base64
             Unsafe.Add(ref destBytes, 2) = (byte)i0;
         }
 
+        [Obsolete("Use Base64.Utf8ToBytes")]
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public static OperationStatus Decode(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
         {
             ref byte srcBytes = ref source.DangerousGetPinnableReference();
@@ -191,6 +195,8 @@ namespace System.Binary.Base64
         /// </summary>
         /// <param name="buffer"></param>
         /// <returns>Number of bytes written to the buffer.</returns>
+        [Obsolete("Use Base64.Utf8ToBytesInPlace")]
+        [EditorBrowsable(EditorBrowsableState.Never)]
         public static OperationStatus DecodeInPlace(Span<byte> buffer, out int bytesConsumed, out int bytesWritten)
         {
             ref byte bufferBytes = ref buffer.DangerousGetPinnableReference();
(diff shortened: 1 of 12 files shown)


src/System.Binary.Base64/System/Binary/Base64Encoder.cs
```diff
@@ -103,6 +105,8 @@ private static int EncodePadByTwo(ref byte srcBytes)
         /// <param name="source"></param>
         /// <param name="destination"></param>
         /// <returns>Number of bytes written to the destination.</returns>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        [Obsolete("Use Base64.BytesToUtf8InPlace")]
```
ahsonkhan: Question: Why are these methods marked as Obsolete/hidden rather than them being internal or even private?

KrzysztofCwalina: I don't want to break Signal-R. Once they move to the new APIs, I will remove these members.

ahsonkhan:
> public static int BytesToUtf8Length(int bytesLength);

Thoughts on Length being a prefix instead of a suffix for easier discoverability and differentiation from the "operation" APIs (I.e. LengthOfBytesToUtf8)?

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
```diff
@@ -282,10 +288,15 @@ public static OperationStatus DecodeInPlace(Span<byte> buffer, out int bytesCons
             return OperationStatus.InvalidData;
         }
 
-        sealed class FromBase64 : BufferDecoder
+        sealed class FromBase64Utf8 : BufferDecoder
```
ahsonkhan: FromBase64ToUtf8?

src/System.Binary.Base64/System/Binary/Base64Encoder.cs
```diff
@@ -192,10 +198,15 @@ public static bool EncodeInPlace(Span<byte> buffer, int sourceLength, out int by
...
         }
 
-        sealed class ToBase64 : BufferEncoder
+        sealed class ToBase64Utf8 : BufferEncoder
         {
             public override OperationStatus Encode(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
                 => Base64.Encode(source, destination, out bytesConsumed, out bytesWritten);
+
+            public override OperationStatus EncodeInPlace(Span<byte> buffer, int inputLength, out int written)
+                => Base64.EncodeInPlace(buffer, inputLength, out written)?OperationStatus.Done:OperationStatus.DestinationTooSmall;
```
ahsonkhan: Why not change EncodeInPlace to return OperationStatus instead of bool just like DecodeInPlace?

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
KrzysztofCwalina: At some point we will add UTF16 version

src/System.Binary.Base64/System/Binary/Base64Encoder.cs
KrzysztofCwalina: EncodeInPlace will go away. It's obsolete. I don't want to change it because I don't want to break Signal-R.

src/System.Binary.Base64/System/Binary/Base64Decoder.cs
ahsonkhan: Got it. Ty.

KrzysztofCwalina:
@ahsonkhan, Length postfix is consistent with text encodings. See: https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoders/Utf8.cs#L22

Besides, postfix sorts nicely (in documentation, intellisense, etc.) with the related APIs

PR #1753, opened 2017-09-13
Removed multiple Utf8 classes

KrzysztofCwalina:
We cannot have multiple types called e.g. Utf8, even if they are in different namespaces. Our ecosystem of tools gets confused with such APIs). This moves Utf8, Utf16, and Custom encoders, parsers, formatters to be nested types in Encodings, Parsers, Formatters outer classes.

Also, I moved all the APIs into System.Buffers* namesapces in preparation for the move to System.Buffers.dll

cc: @ahsonkhan, @shiftylogic, @joshfree

### DIFF
--- a/src/System.Text.Primitives/System/Text/Extensions.Integer.cs
+++ b/src/System.Text.Primitives/System/Text/Extensions.Integer.cs
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace System.Text
+using System.Buffers;
+
+namespace System.Buffers.Text
 {
     /// <summary>
     /// Pseudo-implementations of IBufferFormattable interface for primitive types
--- a/samples/LibuvWithNonAllocatingFormatters/Program.cs
+++ b/samples/LibuvWithNonAllocatingFormatters/Program.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Buffers;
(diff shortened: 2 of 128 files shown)


src/System.Text.Primitives/System/Text/Extensions.Integer.cs
```diff
@@ -1,7 +1,9 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace System.Text
+using System.Buffers;
+
+namespace System.Buffers.Text
```
ahsonkhan: Why are the TryFormat extension methods in the System.Buffers.Text namespace when the Custom Formatter TryFormat are in System.Buffers namespace? Should the Extensions class also be in System.Buffers namespace?

ahsonkhan:
> Also, I moved all the APIs into System.Buffers* namesapces in preparation for the move to System.Buffers.dll


Can you summarize what is in the System.Buffers.Text namespace?

KrzysztofCwalina:
I left Extensions and SymbolTable in System.Buffers.Text.
I think the Extensions class should simply go away. It does not add much value above the Formatters APIs. It just defaults some parameters, which are mostly not needed in the Utf8 and Utf16 inner class APIs. See https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Extensions.Float.cs

SymbolTable could be moved somewhere else, but it's so specific to Text. Having said that I don't feel strongly either way.

Issue #1714, opened 2017-08-22
API Review Cleanup for System.Text.Primitives


KrzysztofCwalina:
These are cleanup workitems based on API reviews, and other discussions that we would like to do before moving the component to corefx repo:

- [x] hide byte* and char* parsing APIs. Span APIs should be used instead
- [ ] More datatypes can be formatted than parsed. We should make the parse/format API symmetrical. #1755
- [x] PrimitiveParser has inner types for efficient invariant and hex parsing. This is also inconsistent with the formatting APIs. I would hide the hex inner types and add invariant inner types to the formatter. e.g. public static class Formatter { public static class Ut8 {…}  public static class Ut16 {…}  public static class Custom {…}  }.
- [x] Namespace should be System.Buffers (in a separate workitem we will place the APIs in System.Buffers.dll)
- [x] Move ParsedFormat to System namespace (we will use it for more than just text)
- [x] Rename TransformationStatus to OperationStatus (we will rename the interface later)
- [x] TransformationStatus.NeedMoreSourceData to TransformationStatus.NeedMoreData 
- [x] Consider renaming PrimitiveFormatter to Formatter and PrimitiveParser to Parser

AtsushiKan:
Official api review is done.

PR #1754, opened 2017-09-13
More cleanup based on API review

KrzysztofCwalina:
Working through items in https://github.com/dotnet/corefxlab/issues/1714

cc @ahsonkhan, @shiftylogic, @joshfree


joshfree:
Thank you @KrzysztofCwalina for picking this up off the backlog

src/System.Text.Primitives/System/Text/Parsers/InvariantBool.cs
```diff
@@ -12,7 +12,7 @@ public static partial class Parsers
     {
         public static partial class Utf8
         {
-            public unsafe static bool TryParseBoolean(byte* text, int length, out bool value)
+            unsafe static bool TryParseBoolean(byte* text, int length, out bool value)
```
ahsonkhan: We should get rid of these APIs instead of hiding them (at least that was the conclusion from our API review - which didn't include the unsafe APIs that take pointers).
These were the APIs:
https://github.com/dotnet/apireviews/tree/master/2017/06-16-System.Text.Buffers
https://gist.github.com/ahsonkhan/6665d60cd41950dcbcb71f10a15886c9

KrzysztofCwalina: When we move the feature to corfx, we will just not move these APIs. I wanted to somehow keep them in case we need them in the future.

ahsonkhan: That sounds good. In that case, should we mark it as internal, make internals visible to the test and keep the tests uncommented?

PR #1756, opened 2017-09-13
Use {ReadOnly}Memory and related types from System.Memory

ahsonkhan:
Part of https://github.com/dotnet/corefxlab/issues/1711

Removing {RO}Buffer and OwnedBuffer and updating all uses to use {RO}Memory and OwnedMemory from the System.Memory package.

Made OwnedArray public since Pipelines (ReadableBuffer) requires a way to cast array to OwnedMemory (and we no longer have the implicit cast that called OwnedArray internally).

cc @shiftylogic, @KrzysztofCwalina, @davidfowl, @pakrym


src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -264,7 +264,7 @@ public int CopyTo(Span<byte> buffer)
                     }
                     else
                     {
-                        segment.CopyTo(free);
+                        segment.Span.CopyTo(free);
```
KrzysztofCwalina: Why don't we have CopyTo on Memory<T>?

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
```diff
@@ -63,12 +62,12 @@ public unsafe void PinnedArrayMemoryLifetime()
...
 
-        static void TestLifetime(OwnedBuffer<byte> owned)
+        static void TestLifetime(OwnedMemory<byte> owned)
         {
-            Buffer<byte> copyStoredForLater;
+            Memory<byte> copyStoredForLater;
             try {
-                Buffer<byte> memory = owned.Buffer;
-                Buffer<byte> memorySlice = memory.Slice(10);
+                Memory<byte> memory = owned.AsMemory;
```
KrzysztofCwalina: AsMemory should not be a property. It should be either a method or should be called Memory

KrzysztofCwalina:
I just had two comments related but not part of this PR. Otherwise looks good. Thanks!

tests/System.Buffers.Experimental.Tests/MemoryTests.cs
ahsonkhan: I agree.
I can make this change, but I was waiting for API review:
https://github.com/dotnet/coreclr/issues/13801#issuecomment-327525698
> Before we do this change, can we discuss (at the API review) when we intend to use properties and when methods for span/memory related conversions? I

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
ahsonkhan: Waiting for API review from this issue which adds a few instance and extension methods:
https://github.com/dotnet/corefx/issues/23862

Issue #1691, opened 2017-08-09
Support sparse layouts


ericstj:
Currently we only use row-major (aka, C-style) layout of Tensor.

We should also consider some sparse layout format, either from a create/convert perspective or potentially an internal layout perspective.

There are many different formats where each is catered towards the type of access.  These involve layered dictionaries which map coordinates to values, lists of lists contain tuples of index and value, CSC/CSS and CSR storage techniques extrapolated to n-dimensions (where some dimensions are compressed by providing a "count" for the number of non-zero values when iterating across the stride of that dimension), etc.

Supporting more formats internally can be done but the various operations on Tensor would need to be special cased to provide ideal performance (like they are special cased for a single-dimension row-major backing buffer today).  This likely will grow with each format to get ideal perf.  If we don't care about perf we could fall-back to an implementation that translates to n-coordinates and let each format deal with the coordinate mapping.

If we support sparse layout internally we'd need to decide what to do about the Buffer property since the its not relevant for a sparse format.  

Supporting formats for creation/conversion is simpler, but negates the memory benefits of the sparse format.  This may be feasible if merely compatibility with these formats is a goal.

ericstj:
This turns out to be a requirement for dealing with real data sets.  We should support at a minimum both CSR and CSC.  These are defined in the following terms:

**values** - one entry for each non-zero value in the tensor, ordered
**cumulative-count**  - for the compressed dimension, gives the count of non-zero values that occur before that index in the dimension
**value-index** - corresponding index for the non-compressed dimensions for each value.  In the case of n-dimensions we can pack this into a single value via the stride calculations, however its unclear if this is the *standard* representation for an n-dimensional CSR/CSC sparse array.  I'll need to do some research.

ericstj:
I plan to change the class heirarchy as follows:

public abstract class Tensor<T> 
public class DenseTensor<T> : Tensor<T>
public class CompressedSparseTensor<T> : Tensor<T>

DenseTensor will support either row-major (last dimension with stride of 1) or column-major (first dimension with stride of 1).
CompressedSparseTensor will support either compressed row (CSR) or compressed column (CSC).

We should also consider what it looks like to build a sparse tensor.  The CSR and CSC formats are very fixed and insertions are expensive (arrays need to be resized, items need to be shifted).

PR #1757, opened 2017-09-13
Add compressed sparse tensor layout

ericstj:
Adds a layout for CompressedSparseTensors supporting both CSR and CSC formats.

/cc @mellinoe


ericstj:
Fixes #1691

PR #1758, opened 2017-09-14
Add detailed parse error strings.

AtsushiKan:
This is part of converging the Kestrel and CoreFxLab
parsers. Kestrel expects this information for logging
and also expects it to be pay-for-play.

### DIFF
--- a/src/System.Text.Http.Parser/HttpParser.cs
+++ b/src/System.Text.Http.Parser/HttpParser.cs
@@ -23,6 +23,18 @@ namespace System.Text.Http.Parser
         private const byte ByteQuestionMark = (byte)'?';
         private const byte BytePercentage = (byte)'%';
 
+        private readonly bool _showErrorDetails;
+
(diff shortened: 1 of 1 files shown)


src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -23,6 +23,18 @@ public class HttpParser : IHttpParser
         private const byte ByteQuestionMark = (byte)'?';
         private const byte BytePercentage = (byte)'%';
 
+        private readonly bool _showErrorDetails;
+
+        public HttpParser()
+            : this(showErrorDetails: true)
```
KrzysztofCwalina: Should we default this to false?

AtsushiKan: I think we should get to a state first where Kestrel is consuming our packages before we start fine tuning the api surface. With this change, we're just one test away from passing all their tests.

PR #1764, opened 2017-09-18
Added Decimal formatting

KrzysztofCwalina:
The implementation is quite naive; we will improve it overtime. Getting this particular overload to be fast/non-allocating is not lower priority. I filed #1765 to track this.

cc: @ahsonkhan, @joshfree, @shiftylogic

### DIFF
--- a/src/System.Text.Primitives/System/Text/Formatters/Utf16.cs
+++ b/src/System.Text.Primitives/System/Text/Formatters/Utf16.cs
@@ -190,11 +190,50 @@ namespace System.Buffers
 
             #region Floating-point APIs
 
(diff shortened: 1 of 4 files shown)


joshfree:
The boilerplate code looks fine to me.  Is this the plan to share some common code and reuse tests with the existing UTF16 string routines?

KrzysztofCwalina:
>The boilerplate code looks fine to me. Is this the plan to share some common code and reuse tests with the existing UTF16 string routines?

If the routines get larger, then we should think about reuse. For now (given the tests are 4 lines), trying to reuse would make the code hard to read, IMO.

PR #1771, opened 2017-09-20
Trying to use Binary Reader on struct with new Reverse Endian method.

ahsonkhan:
Related to: https://github.com/dotnet/corefx/issues/24144#issuecomment-330638602
> - Consider making the Read/Write Apis use the default endianness and just have a flip endianness method for just the single value primitives. 

For structs containing types of of varying lengths, is the above suggestion possible? I don't know how we can call Read and then correctly set the struct values by calling primitive type ReverseEndianness methods.

This is experimental (and won't work as is) and hence **do not merge**.

```C#
int ReverseEndianness(this int num)
long ReverseEndianness(this long num)
```

Note that the test BufferReaderReadHeterogeneousStructWILLFAIL() will fail.
```C#
            HeterogenousTestStruct readStruct = spanBE.Read<HeterogenousTestStruct>();

            // This does not work!
            if (BitConverter.IsLittleEndian)
            {
                readStruct.I0 = readStruct.I0.ReverseEndianness();
                readStruct.L1 = readStruct.L1.ReverseEndianness();
                readStruct.I2 = readStruct.I2.ReverseEndianness();
                readStruct.L3 = readStruct.L3.ReverseEndianness();
                readStruct.I4 = readStruct.I4.ReverseEndianness();
                readStruct.L5 = readStruct.L5.ReverseEndianness();
            }
```

cc @KrzysztofCwalina

### DIFF
--- a/src/System.Binary/System/Binary/BufferReader.cs
+++ b/src/System.Binary/System/Binary/BufferReader.cs
@@ -105,5 +105,33 @@ namespace System.Binary
                 throw new ArgumentOutOfRangeException();
             }
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe static int ReverseEndianness(this int num)
+        {
+            uint val = 0;
+            Unsafe.Write(&val, num);
+            val = (val << 24)
+                | ((val & 0xFF00) << 8)
+                | ((val & 0xFF0000) >> 8)
+                | (val >> 24);
+            return Unsafe.Read<int>(&val);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe static long ReverseEndianness(this long num)
+        {
+            ulong val = 0;
+            Unsafe.Write(&val, num);
+            val = (val << 56)
+                | ((val & 0xFF00) << 40)
+                | ((val & 0xFF0000) << 24)
+                | ((val & 0xFF000000) << 8)
+                | ((val & 0xFF00000000) >> 8)
+                | ((val & 0xFF0000000000) >> 24)
+                | ((val & 0xFF000000000000) >> 40)
+                | (val >> 56);
+            return Unsafe.Read<long>(&val);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


src/System.Binary/System/Binary/BufferReader.cs
```diff
@@ -105,5 +105,33 @@ public static void RequiresInInclusiveRange(int start, uint length)
                 throw new ArgumentOutOfRangeException();
             }
         }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe static int ReverseEndianness(this int num)
+        {
+            uint val = 0;
```
KrzysztofCwalina: BTW, now that Reverse methodfs are non-generic, you don't need to Read and Write.

KrzysztofCwalina:
Why doesn't it work?

ahsonkhan:
> Why doesn't it work?

It seems like this fails to correctly assign the right bits to the appropriate struct field.
```C#
HeterogenousTestStruct readStruct = spanBE.Read<HeterogenousTestStruct>();
```
Below you can see that the order of the struct fields is not being maintained (I0, I2, I4 first rather than I0, L1, I2, ...)
image

HeterogenousTestStruct = [ I0, L1, I2, L3, I4, L5 ]
spanBE = [ 00 00 48 57, 00 00 00 00 247 53 147 255, 127 255 255 255, 128 00 00 00 00 00 00 00, 128 00 00 00, 127 255 255 255 255 255 255 255, 00 00 00 00 00 00 00 00 00 00 00 00 ]

spanBE.Read<> should map the first 32 bits to I0, the next 64 bits to L2, and so on, but that doesn't seem to be happening.

As an aside, for current generic Reverse method: there is no way to know what bytes correspond to the right type within the struct.

Currently for arbitrary struct, Reverse is doing a reverse of the entire byte array and this won't work for structs containing variable length types when we try to read the values.
```C#
                // default implementation
                int len = Unsafe.SizeOf<T>();
                var val = stackalloc byte[len];
                Unsafe.Write(val, value);
                int to = len >> 1, dest = len - 1;
                for (int i = 0; i < to; i++) {
                    var tmp = val[i];
                    val[i] = val[dest];
                    val[dest--] = tmp;
                }
                return Unsafe.Read<T>(val);
```

I don't think we should have the else and just throw an exception in that case.

KrzysztofCwalina:
>I don't think we should have the else and just throw an exception in that case.

Yeah, I talked with @jkotas about it. We need to either remove the else, or make reverse non generic and allow it only on primitve types.

PR #1772, opened 2017-09-20
Adding non-generic {Try}Read/Write BE & LE methods

ahsonkhan:
Related to: https://github.com/dotnet/corefx/issues/24144#issuecomment-330638602
> - BigEndian/LittleEndian the T should be constrained to our single value primitive types, perhaps just use an overload for each of those instead of a T. 

- Adding safe ReverseEndianness primitive APIs to UnsafeUtilities (should move to helper class)
- Adding non-generic {Try}ReadBE & LE and {Try}WriteBE & LE
- Removed generic TryReadBE & LE and TryWriteBE & LE

- Fixed implementation of ReverseEndianness to no longer use Unsafe.Write and Unsafe.Read which improves performance significantly:
image

**Note:** Piplines uses the generic [ReadBE & LE and WriteBE & LE](https://github.com/dotnet/corefxlab/blob/master/src/System.IO.Pipelines.Extensions/ReadWriteExtensions.cs#L39). Should we move them there?
cc @davidfowl 

cc @KrzysztofCwalina

### DIFF
--- a/src/System.Binary/System/Binary/BufferReader_BE.cs
+++ b/src/System.Binary/System/Binary/BufferReader_BE.cs
@@ -22,8 +22,280 @@ namespace System.Buffers
         public static T ReadBigEndian<[Primitive]T>(this Span<byte> buffer) where T : struct
             => BitConverter.IsLittleEndian ? UnsafeUtilities.Reverse(buffer.Read<T>()) : buffer.Read<T>();
 
+        #region ReadBigEndianROSpan
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool TryReadBigEndiann<[Primitive]T>(this ReadOnlySpan<byte> buffer, out T value) where T : struct
+        public static short ReadBigEndianInt16(this ReadOnlySpan<byte> buffer)
+        {
+            short result = buffer.Read<short>();
+            if (BitConverter.IsLittleEndian)
+            {
+                result = UnsafeUtilities.ReverseEndianness(result);
+            }
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int ReadBigEndianInt32(this ReadOnlySpan<byte> buffer)
+        {
+            int result = buffer.Read<int>();
+            if (BitConverter.IsLittleEndian)
+            {
+                result = UnsafeUtilities.ReverseEndianness(result);
+            }
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static long ReadBigEndianInt64(this ReadOnlySpan<byte> buffer)
+        {
+            long result = buffer.Read<long>();
+            if (BitConverter.IsLittleEndian)
+            {
+                result = UnsafeUtilities.ReverseEndianness(result);
+            }
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ushort ReadBigEndianUInt16(this ReadOnlySpan<byte> buffer)
+        {
+            ushort result = buffer.Read<ushort>();
+            if (BitConverter.IsLittleEndian)
+            {
+                result = UnsafeUtilities.ReverseEndianness(result);
+            }
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static uint ReadBigEndianUInt32(this ReadOnlySpan<byte> buffer)
+        {
+            uint result = buffer.Read<uint>();
+            if (BitConverter.IsLittleEndian)
+            {
+                result = UnsafeUtilities.ReverseEndianness(result);
+            }
+            return result;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static ulong ReadBigEndianUInt64(this ReadOnlySpan<byte> buffer)
+        {
+            ulong result = buffer.Read<ulong>();
+            if (BitConverter.IsLittleEndian)
+            {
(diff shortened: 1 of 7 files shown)


src/System.Binary/System/Binary/BufferReader_BE.cs
```diff
@@ -22,22 +22,276 @@ public static partial class Binary
         public static T ReadBigEndian<[Primitive]T>(this Span<byte> buffer) where T : struct
             => BitConverter.IsLittleEndian ? UnsafeUtilities.Reverse(buffer.Read<T>()) : buffer.Read<T>();
 
+        #region ReadBigEndianROSpan
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool TryReadBigEndiann<[Primitive]T>(this ReadOnlySpan<byte> buffer, out T value) where T : struct
+        public static short ReadBigEndianInt16(this ReadOnlySpan<byte> buffer)
```
KrzysztofCwalina: I think the name should be ReadInt16BigEndian as opposed to ReadBigEndian16. The former is better for intellisense completion, as you have to type less before you are down to one matching API.

src/System.Binary/System/Binary/UnsafeUtilities.cs
```diff
@@ -64,5 +64,55 @@ internal static class UnsafeUtilities
...
+        {
+            return (value & 0x000000FF) << 24 |
+                (value & 0x0000FF00) << 8 |
+                (value & 0x00FF0000) >> 8 |
+                (int)(((uint)value & 0xFF000000) >> 24);
+        }
+
+        public static long ReverseEndianness(long value)
+        {
+            return (value & 0x00000000000000FFL) << 56 |
```
KrzysztofCwalina: Is there a bit perf hit if we share the implementation between ulong and long? (inlining the shared part)

ahsonkhan: I will check.

ahsonkhan: There is a ~10% performance difference (if I change it to something like what's shown below):
image

```C#
       public static ulong ReverseEndianness(ulong value)
        {
            return SharedHelper(value) | (value & 0xFF00000000000000UL) >> 56;
        }

        public static long ReverseEndianness(long value)
        {
            return (long)SharedHelper((ulong)value) | (long)(((ulong)value & 0xFF00000000000000L) >> 56);
        }

        private static ulong SharedHelper(ulong value)
        {
            return (value & 0x00000000000000FFUL) << 56 |
                (value & 0x000000000000FF00UL) << 40 |
                (value & 0x0000000000FF0000UL) << 24 |
                (value & 0x00000000FF000000UL) << 8 |
                (value & 0x000000FF00000000UL) >> 8 |
                (value & 0x0000FF0000000000UL) >> 24 |
                (value & 0x00FF000000000000UL) >> 40;
        }
```

ahsonkhan: Thinking about it some more and looking at the disassembly, theoretically there shouldn't be any performance difference.

The only difference I see is the use of shr vs sar instructions (signed divide vs unsigned divide) for the `ReverseEndianness(long value)`.

image

I don't know why the results consistently differ though:
image

@jkotas, any ideas why we would see a performance difference here?

@jorive, here is what my benchmark looks like. Is there something I can do to improve how I write the benchmark?
```C#
private const int TestArrayLength = 1000000;
private static byte[] myArray = new byte[TestArrayLength * 8];
volatile static int tempVal;

[Benchmark]
private static void MyTestAOld()
{
    Span<byte> source = GetArray();
            
    {
        tempVal = 0;
        {
            for (int j = 0; j < TestArrayLength; j++)
            {
                tempVal ^= (int)source.Slice(j * 8).ReadInt64BigEndianOld();
            }
        }
        Assert.Equal(tempVal, 1173436863);
    }
}

[Benchmark]
private static void MyTestBNew()
{
    Span<byte> source = GetArray();
            
    {
        tempVal = 0;
        {
            for (int j = 0; j < TestArrayLength; j++)
            {
                tempVal ^= (int)source.Slice(j * 8).ReadInt64BigEndian();
            }
        }
        Assert.Equal(tempVal, 1173436863);
    }
}

private static Span<byte> GetArray()
{
    Span<byte> source = myArray;
    Random random = new Random(42);
    for (int i = 0; i < TestArrayLength; i++)
    {
        int temp = random.Next();
        source.Slice(i * 8).WriteInt64BigEndian(temp);
    }
    return source;
}
```

jkotas: > private const int TestArrayLength = 1000000;

Iterating over 4MB array is processor cache bound. 4MBs will definitely not fit into L1. It may fit into L2 or L3 depending on what kind of machine you are doing the measurements on.

My guess is that the exact array location in the memory may be influencing the benchmark results because of iteractions with the cache. You may want to try to run both benchmarks on the exact same array.

jkotas: @erozenfeld It may be nice to pattern match this as `bswap`.

ahsonkhan: >  You may want to try to run both benchmarks on the exact same array.

I am using the same array. I wrap a span around `private static byte[] myArray = new byte[TestArrayLength * 8];` that I pass to the methods being measured.

I can try a smaller array (with a high inner iteration count) that would fit into L1 would (4-16 KB be fine)?

erozenfeld: @jkotas I opened https://github.com/dotnet/coreclr/issues/14122 for bswap support in the jit.

PR #1773, opened 2017-09-20
Update to new nuget.exe and add logging and process exit code check.

ahsonkhan:
The nuget.exe at the previous url is really old (2.8.60717.93) and package publishing is failing incorrectly due to "File contains corrupted data." error.

This updates the nuget.exe to a newer version.
Also added an exit code check and the -NoNewWindow flag so that logs can be visible and the CI will catch the failures in the future.

cc @dagood, @shiftylogic, @KrzysztofCwalina

### DIFF
--- a/scripts/package.ps1
+++ b/scripts/package.ps1
@@ -15,7 +15,7 @@ Function Ensure-Nuget-Exists {
             New-Item -ItemType directory -Path "$repoRoot\nuget"
         }
         Write-Host "nuget.exe not found. Downloading to $nugetPath"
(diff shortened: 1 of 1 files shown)


ahsonkhan:
FYI: The task has "Fail on Standard Error" selected.
> If this is true, this task will fail if any errors are written to the error pipeline, or if any data is written to the Standard Error stream. Otherwise the task will rely solely on $LASTEXITCODE and the exit code to determine failure.

ahsonkhan:

PR #1777, opened 2017-09-21
replace "in" with "ref readonly"

VSadov:
"ref readonly" is the only variant that will be supported once we switch to the new toolset compiler.

We had two ways to express the same thing, some people find it confusing, so support for "in" is being removed.

### DIFF
--- a/src/System.Text.Http.Parser/HttpParser.cs
+++ b/src/System.Text.Http.Parser/HttpParser.cs
@@ -35,7 +35,7 @@ namespace System.Text.Http.Parser
             _showErrorDetails = showErrorDetails;
         }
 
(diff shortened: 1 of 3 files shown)


VSadov:
@KrzysztofCwalina - PTAL, trivial change. Thanks!

VSadov:
Thanks!!

KrzysztofCwalina:
why is it "ref readonly" and not "readonly ref"?

VSadov:
Long time ago it was considered that we may one day need `readonly ref readonly` to mean "a readonly ref variable that does not support ref assignments after declaration".   
The leading `readonly` is reserved for such case.

PR #1779, opened 2017-09-21
Adding Binary Reader perf test for reading structs and reversing endianness

ahsonkhan:
It is almost 5x better to Read once and Reverse each field rather than reading one field at a time for a 12 field struct.

image

- Renamed ReverseEndianness to Reverse and made them extension methods.
- Removed UnsafeUtilities.

cc @dotnet/corefxlab-contrib 

Number of   Fields | ReadAndReverse | ReadFieldByField | Ratio
-- | -- | -- | --
3 | 0.9011 | 1.3390 | 1.486
6 | 1.5694 | 2.0663 | 1.317
12 | 3.0530 | 16.9419 | 5.549


KrzysztofCwalina:
All the bounds checks cost, don't they? :-) Reverse does not have any bounds checks. Read does.

KrzysztofCwalina:
If you want to merge the test, go ahead. It looks good. But it might be worth adding a comparison to BitConverter.

ahsonkhan:
> All the bounds checks cost, don't they? :-) Reverse does not have any bounds checks. Read does.

It is also the `BitConvert.LittleEndian` if/else check which only occurs once if you use Read\<T\> and then Reverse each field rather than ReadEndian... on each field.

ahsonkhan:
> If you want to merge the test, go ahead. It looks good. But it might be worth adding a comparison to BitConverter.

image

GrabYourPitchforks:
For the endianness swapping routines, does it make sense to call the BSWAP x86 intrinsic if available? I'm not sure what optimizations the JITter is able to do on our behalf.

ahsonkhan:
> For the endianness swapping routines, does it make sense to call the BSWAP x86 intrinsic if available? I'm not sure what optimizations the JITter is able to do on our behalf.

There is already an issue opened related to this: https://github.com/dotnet/coreclr/issues/14122
From previous PR: https://github.com/dotnet/corefxlab/pull/1772#discussion_r140376081

ahsonkhan:
@KrzysztofCwalina, good to go?

jkotas:
> It is interesting though that these Reverse APIs are 2-3x faster than the NetworkToHostOrder APIs (for Int32 and Int64

They are using naive layered cake implementation (Int32 method calls Int16 method). It is easy to see that there is a bunch of unnecessary extra operations with this approach.

Could you please submit PR to corefx to change these methods to the streamlined implementation you have here?

jkotas:
Also, the mask on the highest order part is not necessary - the bits will be shifted out for free. This:

`return (value & 0x000000FF) << 24 |`

should be just:

return `(value << 24) |`

jkotas:
And this:

`(int)(((uint)value & 0xFF000000) >> 24);`

should be just:

`((uint)value >> 24);`

for the same reason.

Drawaes:
I think we can do faster here. So for a ulong we can swap the first 4 bytes, then swap each short inside the int and finally flip the bytes.  So say for ulong I get (measure reverse is your function run in corefxlabs in release mode, and the ulong is the same but with a ulong instead of an int).

|Test|AVERAGE | STDEV.S |     MIN |     MAX|
|---|---|---|----|----|
|MeasureReverse|196.745 |   8.810 | 187.388 | 232.469|
|MeasureReverse2|177.338 |   8.354 | 169.954 | 211.146|
|MeasureUlongReverse| 358.954 |  11.486 | 343.930 | 394.239|
|MeasureUlongReverse2|284.049 |   9.004 | 271.679 | 304.954|

The code for the ulong looks something like this

``` csharp
public static ulong Reverse2(this ulong value)
{
    value = (value << 32) | (value >> 32);
    value = (value & 0x0000FFFF0000FFFF) << 16 | (value & 0xFFFF0000FFFF0000) >> 16;
    value = (value & 0x00FF00FF00FF00FF) << 8 | (value & 0xFF00FF00FF00FF00) >> 8;
    return value;
}
```

``` csharp
public static uint Reverse2(this uint value)
{
            value = (value << 16) | (value >> 16);
            value = (value & 0x00FF00FF) << 8 | (value & 0xFF00FF00) >> 8;
            return value;
}
```

Finally for the "reverse" test I was lazy and just did this for int and long

``` csharp
public static int Reverse2(this int value) => (int)Reverse2((uint)value);

public static long Reverse2(this long value) => (long)Reverse2((ulong)value);
```

Drawaes:
I put in a PR to change corefx at the same time.

ahsonkhan:
Intermittent Pipelines.Extensions test failure:
https://ci.dot.net/job/dotnet_corefxlab/job/master/job/windows_nt_debug_prtest/763/console
14:42:07 The active test run was aborted. Reason: Unhandled Exception: System.IO.Pipelines.Networking.Libuv.Interop.UvException: Error -4082 EBUSY resource busy or locked
14:42:07    at System.IO.Pipelines.Networking.Libuv.Interop.Uv.ThrowError(Int32 statusCode) in D:\j\workspace\windows_nt_de---4b678204\src\System.IO.Pipelines.Networking.Libuv\Interop\Uv.cs:line 79
14:42:07    at System.IO.Pipelines.Networking.Libuv.Interop.Uv.ThrowIfErrored(Int32 statusCode) in D:\j\workspace\windows_nt_de---4b678204\src\System.IO.Pipelines.Networking.Libuv\Interop\Uv.cs:line 70
14:42:07    at System.IO.Pipelines.Networking.Libuv.Interop.Uv.loop_close(UvLoopHandle handle) in D:\j\workspace\windows_nt_de---4b678204\src\System.IO.Pipelines.Networking.Libuv\Interop\Uv.cs:line 108
14:42:07    at System.IO.Pipelines.Networking.Libuv.Interop.UvLoopHandle.ReleaseHandle() in D:\j\workspace\windows_nt_de---4b678204\src\System.IO.Pipelines.Networking.Libuv\Interop\UvLoopHandle.cs:line 42
14:42:07    at System.Runtime.InteropServices.SafeHandle.InternalDispose()
14:42:07    at System.IO.Pipelines.Networking.Libuv.UvThread.RunLoop() in D:\j\workspace\windows_nt_de---4b678204\src\System.IO.Pipelines.Networking.Libuv\UvThread.cs:line 85
14:42:07    at System.IO.Pipelines.Networking.Libuv.UvThread.OnStart(Object state) in D:\j\workspace\windows_nt_de---4b678204\src\System.IO.Pipelines.Networking.Libuv\UvThread.cs:line 63
14:42:07    at System.Threading.Thread.ThreadMain_ParameterizedThreadStart(Object parameter)
14:42:07    at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)

cc @pakrym 

PR #1780, opened 2017-09-22
Fix errors on selfcontained publish on HttpParser tests.

AtsushiKan:
Since nuget apparently needs me to do its job for it,
I'll add the package references to humor it so I can
do a --self-contained "dotnet publish" without the
scads of NU1605 errors.


KrzysztofCwalina:
ha ha, nice PR comment :-)

ahsonkhan:
I don't really understand the issue or how adding package references resolves it.
@AtsushiKan, can you please explain for my benefit?

AtsushiKan:
This is the explicit guidance given for NU1605:

https://docs.microsoft.com/en-us/nuget/reference/errors-and-warnings#unexpected-package-version-warnings

If you're asking why it's our job to do that instead of NuGet to do its job... you and me both. I just know this makes it work.

PR #1781, opened 2017-09-22
Add Utf8Decoder and tests

GrabYourPitchforks:
This is basic functionality to perform UTF8 sequence validation and to extract Unicode scalar values from a UTF8 sequence. This class meets the requirements as specified in the [Unicode Standard 10.0, Sec. 3.9](http://www.unicode.org/versions/Unicode10.0.0/ch03.pdf), and [Unicode TR 36](http://www.unicode.org/reports/tr36/). The purpose of this class is to provide a single secure decoding component that can consumed by our web encoders, Utf8String, and other types.

Seeking feedback on the API shape and whether this type should be made public. Among the open questions: should this use the existing `OperationStatus` type? This would make APIs like "read next scalar" have yet another _out_ parameter, but maybe the consistency is worth it. We can also nix the `ErrorStatus` type entirely if needed since we really only need to return -1 to indicate error.

One other issue is that running unit tests takes 2 minutes from the command line but only seconds from within VS. Still investigating this.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
+++ b/src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
@@ -2,49 +2,58 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Buffers;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
-namespace System.Text.Primitives.System.Text.Encoders
+namespace System.Text.Encoders
 {
     /// <summary>
     /// Contains facilities for validating and extracting Unicode scalar values from UTF8 byte sequences.
     /// </summary>
     internal static class Utf8Decoder
     {
-        // Assumption: bytesAvailable > 0
+        private const int ReplacementChar = 0xFFFD; // U+FFFD Unicode Replacement Character
+
         /// <summary>
-        /// Reads a Unicode scalar value from the provided UTF8 sequence.
+        /// Reads the next Unicode scalar value from the provided UTF8 sequence.
         /// </summary>
-        /// <param name="data">A reference to the first byte of the UTF8 sequence from which to read the scalar value.</param>
-        /// <param name="bytesAvailable">The number of elements available in the buffer referenced by <paramref name="data"/>.</param>
-        /// <param name="bytesConsumed">When this method returns, contains the number of bytes
-        /// from <paramref name="data"/> which were consumed while reading the scalar value.</param>
-        /// <returns>On success, a non-negative integer which represents a Unicode scalar value.
-        /// On failure, a negative integer which represents an <see cref="ErrorStatus"/> failure code.</returns>
+        /// <param name="data">The UTF8 sequence from which to read the scalar value.</param>
+        /// <param name="bytesAvailable">The number of bytes available in the buffer referenced by <paramref name="data"/>.</param>
+        /// <param name="scalarValue">
+        /// If this method returns <see cref="OperationStatus.Done"/>, contains the next Unicode scalar value that appears in <paramref name="data"/>.
+        /// Otherwise the value is undefined.
+        /// </param>
+        /// <param name="bytesConsumed">
+        /// If this method returns <see cref="OperationStatus.Done"/>, contains the number of bytes that were consumed from <paramref name="data"/> to read the scalar value.
+        /// If this method returns <see cref="OperationStatus.InvalidData"/>, contains the number of bytes in <paramref name="data"/> which form the invalid sequence.
+        /// Otherwise the value is undefined.
+        /// </param>
+        /// <returns><see cref="OperationStatus.Done"/> on success, or <see cref="OperationStatus.InvalidData"/> or <see cref="OperationStatus.NeedMoreData"/> on failure.</returns>
         /// <remarks>
         /// The caller must use extreme caution to avoid passing a value for <paramref name="bytesAvailable"/> that is
         /// larger than the actual buffer referenced by <paramref name="data"/>, else a buffer overrun could occur.
         /// See http://www.unicode.org/glossary/#unicode_scalar_value for the definition of Unicode scalar value.
         /// </remarks>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static int DangerousReadUnicodeScalarValue(ref byte data, int bytesAvailable, out int bytesConsumed)
+        public static OperationStatus DangerousReadNextUnicodeScalarValue(ref byte data, int bytesAvailable, out int scalarValue, out int bytesConsumed)
         {
-            if (bytesAvailable == 0)
+            if (bytesAvailable <= 0)
             {
                 // Quick check: no data to consume
                 bytesConsumed = 0;
-                return ErrorStatus.InsufficientData;
+                scalarValue = ReplacementChar;
+                return OperationStatus.NeedMoreData;
             }
             else
             {
-                return DangerousReadUnicodeScalarValueWithoutNullCheck(ref data, bytesAvailable, out bytesConsumed);
+                return DangerousReadNextUnicodeScalarValueWithoutNullCheck(ref data, bytesAvailable, out scalarValue, out bytesConsumed);
             }
         }
 
         // Assumption: bytesAvailable > 0
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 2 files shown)


GrabYourPitchforks:
Turns out I misread the test data. The 2 min test execution time was due to existing tests in System.Text.Primitives.Tests and isn't a regression caused by this PR. So that's good news. :)

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// from <paramref name="data"/> which were consumed while reading the scalar value.</param>
+        /// <returns>On success, a non-negative integer which represents a Unicode scalar value.
+        /// On failure, a negative integer which represents an <see cref="ErrorStatus"/> failure code.</returns>
+        /// <remarks>
+        /// The caller must use extreme caution to avoid passing a value for <paramref name="bytesAvailable"/> that is
+        /// larger than the actual buffer referenced by <paramref name="data"/>, else a buffer overrun could occur.
+        /// See http://www.unicode.org/glossary/#unicode_scalar_value for the definition of Unicode scalar value.
+        /// </remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int DangerousReadUnicodeScalarValue(ref byte data, int bytesAvailable, out int bytesConsumed)
```
KrzysztofCwalina: Do you think this API will ever be public? If yes, I worry about returning error codes mixed with return values. It's not super .net like,

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
...
+            if ((data & 0x80) == 0)
+            {
+                // Fast case: ASCII character => the code point is the byte itself
+                bytesConsumed = 1;
+                return data;
+            }
+            else
+            {
+                // Slow case: multi-byte sequence
+                bytesConsumed = 0;
```
KrzysztofCwalina: Why do you set it to zero here?

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// Reads a Unicode scalar value from the provided UTF8 sequence.
+        /// </summary>
+        /// <param name="data">The UTF8 sequence from which to read the scalar value.</param>
+        /// <param name="bytesConsumed">When this method returns, contains the number of bytes
+        /// from <paramref name="data"/> which were consumed while reading the scalar value.</param>
+        /// <returns>On success, a non-negative integer which represents a Unicode scalar value.
+        /// On failure, a negative integer which represents an <see cref="ErrorStatus"/> failure code.</returns>
+        /// <remarks>See http://www.unicode.org/glossary/#unicode_scalar_value for the definition of Unicode scalar value.</remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int ReadUnicodeScalarValue(ReadOnlySpan<byte> data, out int bytesConsumed)
```
KrzysztofCwalina: It would be great to make some of these routines public and reusable.

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.Primitives.System.Text.Encoders
```
ahsonkhan: That is a verbose namespace. Maybe System.Text.Primitives.Encoders? I thought we had a namespace with "Web" in the name somewhere.

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
...
+{
+    /// <summary>
+    /// Contains facilities for validating and extracting Unicode scalar values from UTF8 byte sequences.
+    /// </summary>
+    internal static class Utf8Decoder
+    {
+        // Assumption: bytesAvailable > 0
```
ahsonkhan: The comment suggests that we have an assumption that bytesAvailable > 0 but then we check for == 0 in code. Update the comment to >= 0?

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
...
+
+            return -1; // no invalid bytes
+        }
+
+        /// <summary>
+        /// Given the first byte of a UTF8 multi-byte sequence, returns the number of expected trailing
+        /// bytes and the minimum allowed code point value which is encoded by this sequence. Returns
+        /// 0 on invalid leading byte.
+        /// </summary>
+        private static int GetNumberOfTrailingBytes(byte firstByte, out int minAllowedCodePointValue)
```
ahsonkhan: Would any of the helpers in here be a helpful substitute for this, particularly:
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoders/EncodingHelper.cs#L116?

If not, maybe we should add this method there.

ahsonkhan:
cc @shiftylogic

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
...
+        /// <remarks>See http://www.unicode.org/glossary/#unicode_scalar_value for the definition of Unicode scalar value.</remarks>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static int ReadUnicodeScalarValue(ReadOnlySpan<byte> data, out int bytesConsumed)
+        {
+            return DangerousReadUnicodeScalarValue(ref data.DangerousGetPinnableReference(), data.Length, out bytesConsumed);
+        }
+
+        // Assumption: bytesAvailable > 0
+        // Assumption: bytesConsumed = 0
+        private static int ReadUnicodeScalarValueFromNonAscii(ref byte data, int bytesAvailable, ref int bytesConsumed)
```
shiftylogic: If bytesConsumed is effectively required to be set to 0 on input, why is this not just an out parameter?

src/System.Text.Primitives/System/Text/Encoders/Utf8Decoder.cs
```diff
@@ -0,0 +1,264 @@
...
+        {
+            if (bytesAvailable == 0)
+            {
+                // Quick check: no data to consume
+                bytesConsumed = 0;
+                return ErrorStatus.InsufficientData;
+            }
+            else
+            {
+                return DangerousReadUnicodeScalarValueWithoutNullCheck(ref data, bytesAvailable, out bytesConsumed);
```
shiftylogic: bytesAvailable is an input here, and cannot be assumed to be positive. The Debug.Assert in the below private method won't catch issues for the public layer for release builds.

KrzysztofCwalina:
@shiftylogic, for context, all this code is to support System.Text.Encodings.Web.Utf8. But I agree (and added comment along the lines) that some of the unicode related routines can be reused/shared between S.T.Primitives, S.T.Utf8String, and System.Text.Encodings.Web.Utf8

GrabYourPitchforks:
@shiftylogic This type is specifically for consumers who need to iterate over the code points within a UTF8 sequence and perform some operation on those individual code points. In the case of the web encoders, each code point must be checked against a known allow list, and if the code point fails the check then the web encoder performs a replacement on the fly. In particular, this PR is _not_ intended for consumers who need to perform bulk transcoding of textual data from one representation to another, e.g., UTF8 to UTF16. Those bulk transcoding operations can be highly optimized precisely because the caller doesn't need to inspect each individual code point flying by.

KrzysztofCwalina:
@GrabYourPitchforks, what do you want to do with this PR?

GrabYourPitchforks:
@KrzysztofCwalina I have been working on this in another branch to address some of the concerns, especially re: performance and compatibility with the existing System.Text.Encoding types. Hopefully will have a new iteration available within a few days.

GrabYourPitchforks:
Abandoning this PR in favor of https://github.com/dotnet/corefxlab/pull/1881.

PR #1785, opened 2017-09-25
IsSupportedCodePoint shouldn't reject private use or nonchars

### LINKED ISSUE #1784: EncodingHelpers.IsSupportedCodePoint incorrectly disallows noncharacters
`EncodingHelpers.IsSupportedCodePoint` returns _false_ for values outside the valid Unicode range, for surrogates, and for private-use or noncharacters in the BMP. [Per the Unicode Consortium](http://www.unicode.org/faq/private_use.html) this behavior is incorrect, as private use and noncharacters are perfectly valid in well-formed UTF strings.

> Q: Are noncharacters invalid in Unicode strings and UTFs?
> A: Absolutely not. Noncharacters do not cause a Unicode string to be ill-formed in any UTF. This can be seen explicitly in the table above, where every noncharacter code point has a well-formed representation in UTF-32, in UTF-16, and in UTF-8. An implementation which converts noncharacter code points between one UTF representation and another must preserve these values correctly. The fact that they are called "noncharacters" and are not intended for open interchange does not mean that they are somehow illegal or invalid code points which make strings containing them invalid.

If the intent of this method is instead to determine if a particular value has a valid encoding in Unicode (i.e., "is valid scalar?"), then the only check it should perform is that the value is in the range U+0000..U+10FFFF and not within range U+D800..U+DFFF.

GrabYourPitchforks:
Fixes #1784.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoders/EncodingHelper.cs
+++ b/src/System.Text.Primitives/System/Text/Encoders/EncodingHelper.cs
@@ -56,18 +56,9 @@ namespace System.Buffers
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static bool IsSupportedCodePoint(uint codePoint)
+        public static bool IsValidScalarValue(uint codePoint)
         {
-            if (codePoint >= FirstNotSupportedCodePoint)
-                return false;
-            if (codePoint >= HighSurrogateStart && codePoint <= LowSurrogateEnd)
-                return false;
-            if (codePoint >= 0xFDD0 && codePoint <= 0xFDEF)
-                return false;
-            if (codePoint == 0xFFFE || codePoint == 0xFFFF)
-                return false;
-
-            return true;
+            return (codePoint < FirstNotSupportedCodePoint) && !IsSurrogate(codePoint);
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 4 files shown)


ahsonkhan:
cc @shiftylogic

src/System.Text.Primitives/System/Text/Encoders/EncodingHelper.cs
```diff
@@ -56,18 +56,9 @@ public static bool IsWhitespace(uint codePoint)
...
-                return false;
-            if (codePoint >= HighSurrogateStart && codePoint <= LowSurrogateEnd)
-                return false;
-            if (codePoint >= 0xFDD0 && codePoint <= 0xFDEF)
-                return false;
-            if (codePoint == 0xFFFE || codePoint == 0xFFFF)
-                return false;
-
-            return true;
+            return (codePoint < FirstNotSupportedCodePoint) && !IsSurrogate(codePoint);
```
shiftylogic: This is an incorrect change for the way that this function is being used in our encoders. It doesn't consider the small extra holes we check in the code you removed.

shiftylogic: I'd have to dig back through emails (changes) on why we did this, but I vaguely remember there being a reason.

tests/System.Text.Primitives.Tests/Encoding/TextEncoderTestHelper.cs
```diff
@@ -526,20 +524,6 @@ public static bool IsSupportedCodePoint(uint codePoint)
                 return false;
             if (codePoint >= TextEncoderConstants.Utf16HighSurrogateFirstCodePoint && codePoint <= TextEncoderConstants.Utf16LowSurrogateLastCodePoint)
                 return false;
-            if (codePoint >= 0xFDD0 && codePoint <= 0xFDEF)
```
shiftylogic: Again, I vaguely recall that if you generate strings with these characters, you end up getting intermittent crashes in the tests.

PR #1792, opened 2017-09-27
Add a -SkipTests:true option to build.cmd

AtsushiKan:
Shaves five minutes off my morning "sync and refresh
5 different repos") routine.

Defaults to "false" so this is opt-in.


KrzysztofCwalina:
looks good

PR #1800, opened 2017-09-29
Add per pipe segment pool

pakrym:


davidfowl:
Show me the numbers.

/cc @benaadams

pakrym:
Same speed less allocations: 
```

------------------------------------------------ |---------:|----------:|----------:|------------:|----------:|
                           ParseLiveAspNetInline | 225.2 ns |  7.020 ns | 10.507 ns | 4,440,049.0 |       0 B |
                                 LongWriteInline | 842.7 ns |  8.907 ns | 13.331 ns | 1,186,701.3 |       0 B |
                          WritePlaintextResponse | 547.3 ns | 18.352 ns | 27.469 ns | 1,827,117.7 |       0 B |
                      WriteSpanPlaintextResponse | 725.3 ns |  8.725 ns | 13.059 ns | 1,378,758.0 |       0 B |
 WriteableBufferWriterWriteFastPlaintextResponse | 322.1 ns |  3.936 ns |  5.891 ns | 3,104,647.4 |       0 B |

                                           Method |     Mean |    Error |    StdDev |        Op/s |  Gen 0 | Allocated |
------------------------------------------------ |---------:|---------:|----------:|------------:|-------:|----------:|
                           ParseLiveAspNetInline | 226.4 ns | 2.527 ns |  3.782 ns | 4,416,144.1 | 0.0002 |      72 B |
                                 LongWriteInline | 839.6 ns | 7.668 ns | 11.477 ns | 1,191,049.0 |      - |      72 B |
                          WritePlaintextResponse | 533.2 ns | 5.943 ns |  8.895 ns | 1,875,336.4 |      - |      72 B |
                      WriteSpanPlaintextResponse | 719.2 ns | 5.356 ns |  8.017 ns | 1,390,467.0 |      - |      72 B |
 WriteableBufferWriterWriteFastPlaintextResponse | 322.7 ns | 4.989 ns |  7.467 ns | 3,099,028.2 |      - |      72 B |

```

davidfowl:
@benadams can you give this a try and see if it helps your allocation profile?

benaadams:
Yeah, though is it still a frankenbuild to get it working in Kestrel? or compile corefxlabs and copy over?

pakrym:
@davidfowl was too optimistic, you won't be able to test it, kestrel was build against older version of pipes and many breaking changes we made since then. I doubt you can easily test it.

src/System.IO.Pipelines.Testing/BufferUtilities.cs
```diff
@@ -35,7 +35,8 @@ public static ReadableBuffer CreateBuffer(params byte[][] inputs)
 
                 // Create a segment that has offset relative to the OwnedMemory and OwnedMemory itself has offset relative to array
                 var ownedBuffer = new OwnedArray<byte>(chars);
-                var current = new BufferSegment(ownedBuffer, length, length * 2);
```
halter73: I kinda like keeping to old ctors, but I could be convinced otherwise.

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -167,7 +174,8 @@ internal void Ensure(int count)
             if (bytesLeftInBuffer == 0 || bytesLeftInBuffer < count || segment.ReadOnly)
             {
                 var nextBuffer = _pool.Rent(count);
-                var nextSegment = new BufferSegment(nextBuffer);
+                var nextSegment = CreateSegmentUnsynchronized();
```
halter73: This needs the lock, right?

tests/System.IO.Pipelines.Performance.Tests/configs/CoreConfig.cs
```diff
@@ -15,9 +17,11 @@ public CoreConfig()
         {
             Add(JitOptimizationsValidator.FailOnError);
             Add(StatisticColumn.OperationsPerSecond);
+            Add(MemoryDiagnoser.Default);
 
             Add(Job.Default
                 .With(BenchmarkDotNet.Environments.Runtime.Core)
+                .With(CsProjCoreToolchain.NetCoreApp20)
```
halter73: 👍

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -14,6 +14,8 @@ namespace System.IO.Pipelines
     /// </summary>
     public class Pipe : IPipe, IPipeReader, IPipeWriter, IReadableBufferAwaiter, IWritableBufferAwaiter
     {
+        private const int SegmentPoolSize = 16;
```
halter73: How did you come up with this?

For Kestrel, the 16 pooled segments could reference upwards of 64KB. This is just enough for the response buffer, but it's possible the request buffer could need unpooled segments if the app falls too far behind reading it.

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -215,6 +224,28 @@ private BufferSegment AllocateWriteHeadUnsynchronized(int count)
...
+                return _bufferSegmentPool[_pooledSegmentCount];
+            }
+            else
+            {
+                return new BufferSegment();
+            }
+        }
+
+        private void ReturnSegmentUnsynchronized(BufferSegment segment)
+        {
```
halter73: Should the call to ResetMemory happen here to be safe?

src/System.IO.Pipelines.Testing/BufferUtilities.cs
pakrym: Old constructors + Set/Reset methods?

halter73: Yeah, but only do it if *you* think it's better too.

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
pakrym: Selected randomly not to be too large. I think optimizing this allocation matters more in high performance scenarios where reader would either would be fast enough to keep up or request/response sizes would not be too large.

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -35,12 +37,15 @@ public class Pipe : IPipe, IPipeReader, IPipeWriter, IReadableBufferAwaiter, IWr
         private long _length;
         private long _currentWriteLength;
 
+        private byte _pooledSegmentCount;
```
davidfowl: Change this to int

PR #1802, opened 2017-09-30
Removing the generic Reverse API and adding no-op byte overloads.

ahsonkhan:
Notes from System.Binary API review:

1)	Add Reverse overload for byte & sbyte (which is a no-op).
    - This is a no-op and added only for consistency. This allows the caller to read a struct of numeric primitives and reverse each field rather than having to skip byte/sbyte fields.
2)	Remove the generic Reverse API. **Edit**: Moved to Pipelines ReadWriteExtensions
3)	The rest of the APIs are fine to move to corefx as is to System.Buffers dll (and namespace)

cc @dotnet/corefxlab-contrib, @pakrym, @davidfowl

### DIFF
--- a/src/System.Binary/System/Binary/BufferReader.cs
+++ b/src/System.Binary/System/Binary/BufferReader.cs
@@ -16,83 +16,60 @@ namespace System.Buffers
     public static partial class Binary
     {
         /// <summary>
-        /// Reverses a primitive value - performs an endianness swap
+        /// This is a no-op and added only for consistency.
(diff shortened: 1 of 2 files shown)


KrzysztofCwalina:
I think it would be good to describe why reverse on byte makes sense. Otherwise looks good.

ahsonkhan:
> I think it would be good to describe why reverse on byte makes sense (in the PR description).

Done

PR #1805, opened 2017-10-02
Update channels library with API review feedback/requests

stephentoub:
- Change namespace from System.Threading.Tasks.Channels to System.IO.Channels
- Rename Readable/WritableChannel to ChannelReader/Writer
- Rename Channel.In/Out to Reader/Writer
- Remove virtual ReadAsync, GetAwaiter, AsObserver, AsObservable methods
- Rename ClosedChannelException to ChannelClosedException
- Rename BoundedChannelFullMode.Ignore to DropWrite
- Add per-channel type options bags for parameters

I still have reservations about:
- The System.IO.Channels namespace.  There's nothing in this library that's specific to I/O.  I think it'd be better as System.Threading.Channels.
- Removing ReadAsync.  I get that we don't like using exceptions for control flow.  But a) that's only relevant if you use the channel closing functionality, such that you may get an exception once there's no more data to read, b) it hampers the ability to make Channel.CreateUnbuffered useful, c) it's a very useful convenience method for writing concise reading code, and d) implementations can provide a more efficient implementation than can be done on top of the rest of the surface area.

I've still incorporated the feedback, but we should revisit both of those.

I also deleted a bunch of tests along with deleting the various methods mentioned; we should rerun code coverage analysis and fill any gaps that may have been left.

https://github.com/dotnet/corefxlab/issues/1731


src/System.Threading.Tasks.Channels/System/IO/Channels/BoundedChannelFullMode.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-namespace System.Threading.Tasks.Channels
+namespace System.IO.Channels
 {
     /// <summary>Specifies the behavior to use when writing to a bounded channel that is already full.</summary>
     public enum BoundedChannelFullMode
```
KrzysztofCwalina: The enum name is strange. I am not sure if we discussed it at the API review, but maybe:
1. We nest the enum in the type
2. We call the enum OverflowOptions
thoughts?

KrzysztofCwalina: Also, is there a channle or mode where writers can get a callback when the channel overflows?

src/System.Threading.Tasks.Channels/System/IO/Channels/Channel.cs
```diff
@@ -0,0 +1,76 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace System.IO.Channels
+{
+    /// <summary>Provides static methods for creating channels.</summary>
+    public static class Channel
+    {
+        /// <summary>Creates an unbounded channel usable by any number of readers and writers concurrently.</summary>
+        /// <returns>The created channel.</returns>
+        public static Channel<T> CreateUnbounded<T>() =>
```
KrzysztofCwalina: I still wish we chose one channel type as a default and called the factory method simply "Create"

src/System.Threading.Tasks.Channels/System/IO/Channels/BoundedChannelFullMode.cs
stephentoub: > I am not sure if we discussed it at the API review

We discussed it, but didn't come to any conclusions about better names.

> We nest the enum in the type

In which type?  BoundedChannelOptions?

> We call the enum OverflowOptions

I see the appeal, but personally that suggests to me it's about responding to an overflow having already occurred, whereas this is about what happens when one would occur if we didn't take action.

src/System.Threading.Tasks.Channels/System/IO/Channels/ChannelOptions.cs
```diff
@@ -0,0 +1,106 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.IO.Channels
+{
+    /// <summary>Provides options that control the behavior of channel instances.</summary>
+    public abstract class ChannelOptions
```
KrzysztofCwalina: Is there a reason for the hierarchy? Would it be better to have no hierarchy and make the options types structs?

src/System.Threading.Tasks.Channels/System/IO/Channels/Channel.cs
stephentoub: There wasn't any agreement on what that default type would be.  For users coming from Go, the default there is "Unbuffered".  In terms of usefulness, I expect "Unbounded" is the most useful (it's the direct equivalent of List/Queue/Dictionary/etc. which are all "unbounded").  But devs wanting to avoid OOM issues might want the default to be "Bounded".  Seems like the best option is to force the dev to make a decision about this up-front and to make the choice obvious.

src/System.Threading.Tasks.Channels/System/IO/Channels/ChannelReader.cs
```diff
@@ -0,0 +1,35 @@
...
+    /// Provides a base class for reading from a channel.
+    /// </summary>
+    /// <typeparam name="T">Specifies the type of data that may be read from the channel.</typeparam>
+    public abstract class ChannelReader<T>
+    {
+        /// <summary>
+        /// Gets a <see cref="Task"/> that completes when no more data will ever
+        /// be available to be read from this channel.
+        /// </summary>
+        public virtual Task Completion => ChannelUtilities.s_neverCompletingTask;
```
KrzysztofCwalina: Should it be called "Completed"?

src/System.Threading.Tasks.Channels/System/IO/Channels/ChannelOptions.cs
stephentoub: I don't have a strong opinion.  I thought this is what we'd decided on, but I'm fine changing it if that's preferred.

src/System.Threading.Tasks.Channels/System/IO/Channels/ChannelReader.cs
stephentoub: To me personally "Completed" makes it sound like it's a Boolean.  But I don't have a strong preference.

src/System.Threading.Tasks.Channels/System/IO/Channels/BoundedChannelFullMode.cs
KrzysztofCwalina: Yeah, I was thinking about nested in BoundedChannelOptions.

src/System.Threading.Tasks.Channels/System/IO/Channels/ChannelReader.cs
KrzysztofCwalina: Boolean property would be called "IsCompleted", wouldn't it?

stephentoub: That's what we'd name it... I'm just saying I see "Completed" and think "bool".

KrzysztofCwalina: OK, your call.

omariom:
@stephentoub @KrzysztofCwalina 
Have you considered making  channels and readers/writers public types?
Given the recent changes in JIT calling sealed methods directly instead of going via virtual dispatch can be much faster, at least for `SingleConsumerUnboundedChannel`. 
_Though I haven't tested the performance yet_

stephentoub:
> Have you considered making channels and readers/writers public types?

I've gone back and forth on that.  It can avoid some overhead.  It also complicates things, e.g. currently `Channel.CreateUnbounded(...)` might return one of a few different concrete types.  We couldn't continue with this pattern if we wanted the methods to return the strong-typed channels in the public API.  A developer could of course cast to them if it knew which it was getting, but we could also do that later by exposing the concrete types if we cared to.

KrzysztofCwalina:
@stephentoub, should we merge?

src/System.Threading.Tasks.Channels/System/IO/Channels/BoundedChannelFullMode.cs
stephentoub: I've been playing around with it.  Nesting it ends up being odd, as then the name you want for the type ends up being the same as the name you want for the property, but you can't name them the same.  I also think "full mode" is a fine description.  I'm going to leave it as-is for now; we can choose to rename it / reorganize it if there's feedback to that effect before it ships.

PR #1806, opened 2017-10-02
API changes remaining from API review

ericstj:
Merging a number of API changes from FxDC review.

/cc @mellinoe @eerhardt

### DIFF
--- a/src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
@@ -1,4 +1,7 @@
-﻿namespace System.Numerics
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Numerics
 {
     public static class ArrayTensorExtensions
     {
--- a/src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
@@ -1,0 +1,66 @@
+﻿namespace System.Numerics
+{
+    public static class ArrayTensorExtensions
+    {
+        public static DenseTensor<T> ToTensor<T>(this T[] array, bool reverseStride = false)
+        {
+            return new DenseTensor<T>(array, reverseStride);
+        }
+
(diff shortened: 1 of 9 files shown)


src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
```diff
@@ -0,0 +1,69 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Numerics
+{
+    public static class ArrayTensorExtensions
+    {
+        public static DenseTensor<T> ToTensor<T>(this T[] array, bool reverseStride = false)
```
eerhardt: This was probably covered in the FxDC review, but why is this `ToTensor` and not `ToDenseTensor`?

ericstj: Because an array is already dense, so its most likely that folks will want to make it a dense tensor.  Sparse tensors are more common when data is too large to represent as a dense tensor: in which case you're not likely to have a (dense) array representation.

src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
```diff
@@ -342,7 +349,63 @@ public override Tensor<T> Reshape(params int[] dimensions)
...
+            return new CompressedSparseTensor<T>(newValues, (int[])compressedCounts.Clone(), newIndicies, nonZeroCount, dimensions, IsReversedStride);
+        }
+
+        public override SparseTensor<T> ToSparseTensor()
+        {
+            var sparseTensor = new SparseTensor<T>(dimensions, capacity: NonZeroCount, reverseStride: IsReversedStride);
+
+            var compressedIndex = 0;
+
+            for (int valueIndex = 0; valueIndex < nonZeroCount; valueIndex++)
```
eerhardt: (minor) Not sure if this algorithm would ever change, but there is some duplicated logic between this and `ToDenseTensor` above.  Would it make sense to refactor this logic in a single spot?

ericstj: I would hope so, once we get https://github.com/dotnet/corefxlab/issues/1763.  

For something as small as this I wouldn't try and factor it out since the body of the loop is subtly different and I don't want to inject a delegate for that difference.

PR #1809, opened 2017-10-02
Add single-dimensional accessors to Tensor

ericstj:
Tensor represents a multi-dimensional API over something that's backed by single dimensional
data (memory).  Most implementations deal in the single-dimension and need to convert to
it (and back) often.  To avoid this add single-dimensional accessors and prefer them.

/cc @eerhardt

### DIFF
--- a/src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
@@ -2,6 +2,7 @@
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
 using System.Diagnostics;
+using System.Linq;
 
 namespace System.Numerics
 {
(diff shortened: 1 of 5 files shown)


src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
```diff
@@ -400,9 +431,8 @@ public override SparseTensor<T> ToSparseTensor()
                 }
 
                 var index = indices[valueIndex] + compressedIndex * strides[compressedDimension];
-
-                // interals access :( remove if we expose layout-based indexing
```
eerhardt: Can we make that field `private` again?

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1042,12 +1020,23 @@ private bool Equals(Tensor<T> other, IEqualityComparer comparer)
 
             for (int i = 0; i < Length; i++)
             {
-                ArrayUtilities.GetIndices(strides, IsReversedStride, i, indices);
-
-                if (!comparer.Equals(this[indices], other[indices]))
+                if (IsReversedStride == other.IsReversedStride)
```
eerhardt: (minor) would it make sense to move this branch out of the loop?  Checking it each time in the loop seems unnecessary.

ericstj: Yeah, I considered it but opted not to duplicate the loop.  I was hoping that the compiler would be smart enough to lift it since both of those are backed by readonly bools, but maybe not.

ericstj: I went ahead and lifted it out, did the same for CompareTo

ericstj:
Fixes #1809

Issue #1751, opened 2017-09-11
New UTF8String APIs


A-And:
## Rationale
As it stands UTF8String lacks full feature parity with System.String. To help alleviate this and taking Issue #358 as a guideline, below are a number of proposed methods to be added to UTF8String.

## API Proposal

```csharp
public partial ref struct Utf8String {

    public static Utf8String operator+(Utf8String s1, Utf8String s2);

    public bool IsNullOrWhiteSpace();

    public Utf8String Remove(int startIndex);
    public Utf8String Remove(int startIndex, int count);

    public Utf8String Replace(Utf8String oldValue, Utf8String newValue);

}

partial static class Utf8Helper {

    public static bool IsDigit(uint codePoint);

}
```

## Discussion

In the original proposal @krwq mentioned that `+` operator as well as `Remove` and `Replace` methods may cause issues with views of strings. However, currently functionality referring to substrings create new Utf8String instances instead of referring to the original. It's entirely possible I may have overlooked something, but as it stands the methods add a lot of convenience without an obvious clear detriment.

A-And:
@KrzysztofCwalina @ahsonkhan @shiftylogic

KrzysztofCwalina:
Thanks. Some questions/comments:

1. System.String is immutable. Do we want Utf8String to be immutable too?
2. System.String can be put on the heap. Utf8String currently cannot. Should Utf8String be heapable?
    - BTW, I think concatenation only makes sense for heapable strings.
3. Do we have APIs to do ToUpper/ToLower?
4. We should consider adding something like String.Split. It would be great if the API was non allocating, i.e. it would return struct collection like type.

VSadov:
I think Utf8String should be as close as possible to String.  Strings are units of data - want to store/compose them, use as keys in hashtables and so on.

Also it would be useful to have Utf8Extensions that work with ReadOnlySpan<byte> and Span<byte>, if In-place mutating like Replace.  

Maybe add similar Utf16Extensions over spans of Char for completeness.

A-And:
@KrzysztofCwalina Answers below: 
 
1. It probably should be. I think deviating behavior between Utf8String and System.String is unnecessary and confusing. To that end it would be worth it to change `Remove` and `Replace`  to fit with the System.String API, i.e.

```csharp
    public Utf8String Remove(int startIndex);
    public Utf8String Remove(int startIndex, int count);

    public Utf8String  Replace(Utf8String oldValue, Utf8String newValue);
```

[...]

Issue #1799, opened 2017-09-29
Introduce new primitive type System.Utf8Char


GrabYourPitchforks:
While looking at the encoding and transformation APIs I noticed that all data is represented by `Span<byte>`. I believe this behavior to be incorrect because it doesn't draw a distinction between textual data (data that consists of letters, symbols, and graphemes) and binary data (data that consists of arbitrary octets). I believe that if we just expose all data - both textual and binary - as `Span<byte>` and require developers to know from context how the data should be treated then this will lead to a pit of failure.

I propose a new type `System.Utf8Char` which is an 8-bit analog to `System.Char`. The general idea is that if the developer sees `Span<byte>` in his code, he should treat it as unstructed binary data; but if he sees `Span<Utf8Char>` or `Span<char>` in his code, then he knows that it's meaningful textual data.

One key difference between binary data and textual data is that binary data can interact directly with i/o. Textual data is not interchangeable outside the application unless it is first somehow converted to binary data. We would provide methods for developers to convert between `Span<byte>` and `Span<Utf8Char>`. The conversion routine would basically be a glorified validation routine and memcpy. If a developer really wanted to avoid the memcpy and knew ahead of time that the incoming binary data was well-formed UTF8 then he could bitblt from `Span<byte>` to `Span<Utf8Char>` directly, but we should take care not to encourage this as a common practice since bypassing validation could lead to security issues at runtime.

Ideally this new type would be an intrinsic, but it can be mocked at the moment by using a custom struct.

```csharp
using System.Runtime.InteropServices;

namespace System
{
    // An 8-bit type similar to but distinct from System.Byte.
    // Utf8Char is not integral (no arithmetic operations) or binary (no bitwise operations)
    // but is comparable (allow ==, <, etc.).
    [StructLayout(LayoutKind.Auto, Size = 1)]
    public struct Utf8Char : IComparable<Utf8Char>, IEquatable<Utf8Char>
    {
        private readonly byte _value;

        public Utf8Char(Utf8Char other)
        {
            _value = other._value;
        }

        public static bool operator ==(Utf8Char a, Utf8Char b) => a._value == b._value;

        public static bool operator !=(Utf8Char a, Utf8Char b) => a._value != b._value;

        public static bool operator <(Utf8Char a, Utf8Char b) => a._value < b._value;

        public static bool operator <=(Utf8Char a, Utf8Char b) => a._value <= b._value;

[...]

PR #1810, opened 2017-10-03
Heapable Utf8String Implementattion

A-And:
I apologize for the large PR. Changing some method names could be unnecessary (i.e. GetUtf8String to GetUtf8Span) and wouldn't be backwards compatible. 
Based on the discussion in #1751, the byte[] and constructor implementations could have to be changed.


joshfree:
Thanks @A-And!  Could someone on @dotnet/corefxlab-contrib please take a look?

KrzysztofCwalina:

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerator.cs
```diff
@@ -0,0 +1,120 @@
...
+                    if (IsOnResetPosition())
+                    {
+                        return -1;
+                    }
+
+                    return _index;
+                }
+            }
+
+            public unsafe uint Current
```
KrzysztofCwalina: Why is this unsafe?

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerator.cs
```diff
@@ -0,0 +1,120 @@
...
+                }
+
+                _index += _currentLenCache;
+                _currentLenCache = 0;
+
+                return HasValue();
+            }
+
+            // This is different than Reset, it goes to the first element not before first
+            private void MoveToFirstPosition()
```
KrzysztofCwalina: For short method bodies like that, you might consider using => syntax, i.e.
```c#
private void MoveToFirstPosition() => _index = 0;
```

src/System.Text.Utf8String/System/Text/Utf8Span.Enumerator.cs
```diff
@@ -0,0 +1,41 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+
+namespace System.Text.Utf8
+{
+    partial struct Utf8Span
+    {
+        public ref struct Enumerator
```
KrzysztofCwalina: Why do we need this type? i.e. what are the scenarios for enumerating byte by byte? If people need to access to the bytes, should they use Utf8Span.Bytes property which would return ReadOnlySpan<byte>

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+        public Utf8Span(byte[] utf8bytes, int index, int length)
+        {
+            _buffer = new ReadOnlySpan<byte>(utf8bytes, index, length);
+        }
+
+        public Utf8Span(string s)
+        {
+            if (s == null)
+            {
+                throw new ArgumentNullException("s", "String cannot be null");
```
KrzysztofCwalina: instream of hardcoding "s", you should use nameof(s).

Also, maybe we shoudl call the parameter utf16String

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+            {
+                _buffer = new ReadOnlySpan<byte>(GetUtf8BytesFromString(s));
+            }
+        }
+
+        /// <summary>
+        /// This constructor is for use by the compiler.
+        /// </summary>
+        [EditorBrowsable(EditorBrowsableState.Never)]
+        public Utf8Span(RuntimeFieldHandle utf8Data, int length) : this(CreateArrayFromFieldHandle(utf8Data, length))
```
KrzysztofCwalina: Two questions:
1. Do we need this ctor on Utf8Span, or we just need it on Utf8String?
2. Is the ctor even useful for the compiler? i.e. will the compiler benefit from it or it will just create byte[] and pass into the byte[] ctor?

@VSadov, @jaredpar

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+            RuntimeHelpers.InitializeArray(array, utf8Data);
+            return array;
+        }
+
+        public static Utf8Span Empty { get { return s_empty; } }
+
+        /// <summary>
+        /// Returns length of the string in UTF-8 code units (bytes)
+        /// </summary>
+        public int Length
```
KrzysztofCwalina: You can use:
```c#
public int Length => _buffer.Length;
```

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+
+        public CodePointEnumerable CodePoints
+        {
+            get
+            {
+                return new CodePointEnumerable(_buffer);
+            }
+        }
+
+        public byte this[int i]
```
KrzysztofCwalina: I would not expose it. Instead expose ReadOnlySpan<byte> Bytes property;

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+        {
+            return utf8.Bytes;
+        }
+
+        public static explicit operator Utf8Span(string s)
+        {
+            return new Utf8Span(s);
+        }
+
+        public static explicit operator string(Utf8Span s)
```
KrzysztofCwalina: Maybe we should remove this. It will be more explicit and discoverable to simply call ToString

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+                return string.Empty;
+
+            // UTF-16 is 2 bytes per char
+            var chars = new char[needed >> 1];
+            var utf16 = new Span<char>(chars).AsBytes();
+            status = Encodings.Utf8.ToUtf16(this.Bytes, utf16, out int consumed, out int written);
+            if (status != Buffers.OperationStatus.Done)
+                return string.Empty;
+
+            return new string(chars);
```
KrzysztofCwalina: The way this method should be implemented is similar to how Encodings.Ascii.ToUtf16String is implemented, i.e. you should compute how many bytes would need, then create System.String of appropriate length, then pin the string, and then write the transcoded bytes directly to the pinned string.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        /// <summary>
+        ///
+        /// </summary>
+        /// <param name="index">Index in UTF-8 code units (bytes)</param>
+        /// <returns>Length in UTF-8 code units (bytes)</returns>
+        public Utf8Span Substring(int index)
```
KrzysztofCwalina: I think it's ok to merge this for now, but I think we need to decide what to do with all the index-based APIs on Utf8 strings. I am not sure if indexing based on code units is useful. Maybe only deal with code points in Utf8Strings.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+                {
+                    return i;
+                }
+            }
+
+            return StringNotFound;
+        }
+
+        // TODO: Should this be public?
+        public int IndexOf(byte codeUnit)
```
KrzysztofCwalina: I don't think so. People can always do utf8.Bytes.IndexOf(...)

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+            {
+                result = default;
+                return false;
+            }
+
+            result = Substring(idx);
+            return true;
+        }
+
+        public bool TrySubstringFrom(byte codeUnit, out Utf8Span result)
```
KrzysztofCwalina: Yeah, the code unit APIs like that would be very buggy, i.e. a random code unit could match.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+            foreach (var codePoint in codePoints)
+            {
+                len += Utf8Helper.GetNumberOfEncodedBytes(codePoint);
+            }
+
+            return len;
+        }
+
+        // TODO: This should return Utf16CodeUnits which should wrap byte[]/Span<byte>, same for other encoders
+        private static byte[] GetUtf8BytesFromString(string str)
```
KrzysztofCwalina: Is this method better than Encoding.UTF8.GetBytes(str)?

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+            throw new NotImplementedException();
+        }
+
+        public Utf8Span Trim(byte[] trimCodeUnits)
+        {
+            throw new NotImplementedException();
+        }
+
+        // TODO: Name TBD, CopyArray? GetBytes?
+        public byte[] CopyBytes()
```
KrzysztofCwalina: Maybe utf8.Bytes.ToArray() is enough?

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+        {
+            return _buffer.ToArray();
+        }
+
+        public byte[] CopyCodeUnits()
```
KrzysztofCwalina: This is the same as utf8.Bytes.ToArray() isn't it? I would remove.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public static bool IsWhiteSpace(byte codePoint)
```
KrzysztofCwalina: This should not be public. If we have code point APIs they should be in general Unicode type and should be 4 byte int.

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -11,29 +11,48 @@
...
         {
-            _buffer = new ReadOnlySpan<byte>(utf8bytes, index, length);
+            if (index < 0)
+            {
+                throw new ArgumentOutOfRangeException("index");
+            }
+
+            if (length < 0)
+            {
+                // TODO: Should we support that?
```
KrzysztofCwalina: I don't think so.

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -11,29 +11,48 @@
...
+
+            if (index + length > utf8bytes.Length)
+            {
+                throw new ArgumentOutOfRangeException("index");
+            }
+
+            // Length checking
+            _buffer = new byte[length];
+            for (int i = 0; i < length; i++)
+                _buffer[i] = utf8bytes[i + index];
```
KrzysztofCwalina: It's strange that this ctor would copy, but the one taking byte[] would not. 

Also, the code should be written as follows:
```c#
_buffer = new Span<byte>(utf8Bytes, index, length).ToArray();
```

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -45,14 +64,15 @@ public Utf8String(string s)
 
             if (s == string.Empty)
             {
-                _buffer = ReadOnlySpan<byte>.Empty;
+                // Might not be the correct approach
+                _buffer = new byte[0];
```
KrzysztofCwalina: Yeah, why did you change it from the ReadOnlySpan\<byte\>.Empty?

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -124,7 +144,7 @@ public CodePointEnumerable CodePoints
             return s.ToString();
         }
 
-        public ReadOnlySpan<byte> Bytes => _buffer;
+        public byte[] Bytes => _buffer;
```
KrzysztofCwalina: This makes the type mutable. The property should return ReadOnlySpan<bytes.

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -149,7 +169,25 @@ public bool ReferenceEquals(Utf8String other)
 
         public bool Equals(Utf8String other)
         {
-            return _buffer.SequenceEqual(other._buffer);
+            if(other is null)
```
KrzysztofCwalina: Why not keep using SquenceEqual on Bytes property? i.e. 
```c#
return Bytes.SequanceEqual(other.Bytes); // assuming Bytes returns ReadOnlySpan
```

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -766,7 +804,7 @@ public Utf8String Trim(byte[] trimCodeUnits)
         // TODO: Name TBD, CopyArray? GetBytes?
         public byte[] CopyBytes()
         {
-            return _buffer.ToArray();
+            return _buffer;
```
KrzysztofCwalina: This does not copy. 

But as I commented above, I think this should be acomplished with utf8.Bytes.ToArray();

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -11,29 +11,48 @@
...
+        private readonly byte[] _buffer;
 
         private const int StringNotFound = -1;
 
-        static Utf8String s_empty => default;
+        static Utf8String s_empty = new Utf8String(string.Empty);
 
         // TODO: Validate constructors, When should we copy? When should we just use the underlying array?
         // TODO: Should we be immutable/readonly?
         public Utf8String(ReadOnlySpan<byte> buffer)
```
benaadams: Also constructor that takes `Utf8Span`?

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -11,29 +11,48 @@
 namespace System.Text.Utf8
 {
     [DebuggerDisplay("{ToString()}u8")]
-    public partial ref struct Utf8String
+    public partial class Utf8String
```
benaadams: Can still be a `struct`?

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/TcpConnectionFormatter.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Net.Http
 {
     public class TcpConnectionFormatter : ITextOutput, IDisposable
     {
-        static byte[] s_terminator = new Utf8String("0\r\n\r\n").Bytes.ToArray();
+        static byte[] s_terminator = new Utf8Span("0\r\n\r\n").Bytes.ToArray();
```
benaadams:
```csharp
static Utf8String s_terminator = new Utf8String(new Utf8Span("0\r\n\r\n");
```

blowdart:
Are we seriously committed to treating a byte array, a binary structure as text, with no validation? This is a security pit of failure. @GrabYourPitchforks suggested a more secure approach in https://github.com/dotnet/corefxlab/issues/1799

Right now this strikes me as way way too dangerous.

GrabYourPitchforks:
We need to define the behavior of what validation (if any) occurs when an arbitrary byte array is provided to the Utf8String constructor. FWIW, taking arbitrary data from the wire and bitblitting it directly as a UTF8 string was something I was trying to avoid doing, which is why I created the issue https://github.com/dotnet/corefxlab/issues/1799.

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerator.cs
```diff
@@ -0,0 +1,120 @@
...
+namespace System.Text.Utf8
+{
+    partial struct Utf8Span
+    {
+        public ref struct CodePointEnumerator
+        {
+            private ReadOnlySpan<byte> _buffer;
+            private int _index;
+            private int _currentLenCache;
+            private const int ResetIndex = -Utf8Helper.MaxCodeUnitsPerCodePoint - 1;
```
ahsonkhan: Can you please add a comment explaining why ResetIndex is set to this value?

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerator.cs
```diff
@@ -0,0 +1,120 @@
...
+
+            public unsafe CodePointEnumerator(ReadOnlySpan<byte> buffer) : this()
```
ahsonkhan: Does this need to be unsafe?

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerator.cs
```diff
@@ -0,0 +1,120 @@
...
+                        throw new InvalidOperationException("MoveNext() needs to be called at least once");
+                    }
+
+                    if (!HasValue())
+                    {
+                        throw new InvalidOperationException("Current does not exist");
+                    }
+
+                    uint codePoint;
+                    bool succeeded = Utf8Helper.TryDecodeCodePoint(_buffer, _index, out codePoint, out _currentLenCache);
```
ahsonkhan: nit: inline the codePoint delcaration
```C#
bool succeeded = Utf8Helper.TryDecodeCodePoint(_buffer, _index, out uint codePoint, out _currentLenCache);
```

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointReverseEnumerator.cs
```diff
@@ -0,0 +1,122 @@
...
+                    }
+
+                    if (!HasValue())
+                    {
+                        throw new InvalidOperationException("Current does not exist");
+                    }
+
+                    ReadOnlySpan<byte> buffer = _buffer.Slice(0, _index);
+                    uint ret;
+                    bool succeeded = Utf8Helper.TryDecodeCodePointBackwards(buffer, out ret, out _currentLenCache);
```
ahsonkhan: nit: same here, inline the declaration of ret.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+            {
+                return Empty;
+            }
+
+            if (length == Length)
+            {
+                return this;
+            }
+
+            if (index + length > Length)
```
ahsonkhan: This check should happen before
`if (length == Length) return this;` (which I don't think is necessary).

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,782 @@
...
+
+        // TODO: Naive algorithm, reimplement faster
+        // TODO: Should this be public?
+        public int IndexOf(Utf8Span value)
+        {
+            if (value.Length == 0)
+            {
+                // TODO: Is this the right answer?
+                // TODO: Does this even make sense?
+                return 0;
```
ahsonkhan: This is correct if we want to match the behavior with string index of:
https://msdn.microsoft.com/en-us/library/k8b1470s(v=vs.110).aspx
> The zero-based index position of value if that string is found, or -1 if it is not. If value is String.Empty, the return value is 0.

src/System.Text.Utf8String/System/Text/Utf8String.CodePointEnumerable.cs
```diff
@@ -4,18 +4,22 @@
...
     {
         public ref struct CodePointEnumerable 
         {
-            private ReadOnlySpan<byte> _buffer;
+            private byte[] _buffer;
 
             public CodePointEnumerable(byte[] bytes, int index, int length)
             {
-                _buffer = new ReadOnlySpan<byte>(bytes, index, length);
+                _buffer = new byte[length];
```
ahsonkhan: Should this be `new byte[index - length]` and then we copy only the subset of the bytes from the input array?

src/System.Text.Utf8String/System/Text/Utf8String.CodePointEnumerable.cs
```diff
@@ -4,18 +4,22 @@
...
+                for(int i = index; i < length; i++)
```
ahsonkhan: `for (int i = 0; i < _buffer.Length; i++)`

src/System.Text.Utf8String/System/Text/Utf8String.CodePointReverseEnumerator.cs
```diff
@@ -49,7 +49,8 @@ public unsafe uint Current
                         throw new InvalidOperationException("Current does not exist");
                     }
 
-                    ReadOnlySpan<byte> buffer = _buffer.Slice(0, _index);
+                    // New allocation ?
+                    ReadOnlySpan<byte> buffer = new ReadOnlySpan<byte>(_buffer, 0, _index);
```
ahsonkhan: This won't allocate. You would have to call ToArray() on the Span to allocate.

tests/System.Text.Utf8String.Tests/RandomTests.cs
```diff
@@ -961,4 +954,955 @@ public void EnsureCodeUnitsOfStringByIndexingBytes(byte[] expectedBytes, string
...
+        public static object[][] LengthInCodePointsTestCases = {
+            new object[] { 0, ""},
+            new object[] { 4, "1258"},
+            new object[] { 1, "\uABCD"},
+            new object[] { 2, "a\uABEE"},
+            new object[] { 3, "a\uABEEa"},
+            new object[] { 4, "a\uABEE\uABCDa"}
+        };
+
+        //[Theory(Skip = "System.InvalidProgramException : Common Language Runtime detected an invalid program."), MemberData("LengthInCodePointsTestCases")]
```
ahsonkhan: Why is this commented out?

tests/System.Text.Utf8String.Tests/RandomTests.cs
```diff
@@ -961,4 +954,955 @@ public void EnsureCodeUnitsOfStringByIndexingBytes(byte[] expectedBytes, string
             }
         }
     }
+
+    public class Utf8SpanTests
```
ahsonkhan: nit: split into separate file

src/System.Text.Utf8String/System/Text/Utf8Span.cs
benaadams: Potential overflow in addition; should subtract 
```
index > Length - length 
```

src/System.Text.Utf8String/System/Text/Utf8String.cs
KrzysztofCwalina: The idea was that we keep it as a class. It avoids accidental boxing, especially that this type will probably implement some interfaces, e.g. IEquatable.

We will ultimately have two utf8string: fast and slow.
This one is slow. The fast one will be length + data inline, just like System.String.
See https://github.com/dotnet/corefxlab/issues/1751#issuecomment-329936409

The only reason for making it a struct (that I see) is that it would allow fast and slow string to have the same perfromance characteristics, but I am not sure this is the right tradeoff.

KrzysztofCwalina:
I think we definitely should add APIs to validate that Span<byte> is actually valid UTF8.

@blowdart, could you explain how it's a security pit of failure to not validation in this primitive type? Validating on every ctor call would literally kill the Utf8Span type (as it's stack only and needs to be created very often).

GrabYourPitchforks:
@KrzysztofCwalina What is the result of `(u"admin\xFF" == u"admin\xFE")`? What about `((u"admin\xFF").ToString() == (u"admin\xFE").ToString())`? What happens if you try creating an entry in a database where the primary key is u"admin\xFF"? I've seen attacks where non-compliant input makes its way through the system and causes undefined behavior when the application tries processing it, even resulting in DoS or EoP as a result.

KrzysztofCwalina:
@GrabYourPitchforks, given how spans work, we cannot prevent malicious code creating u"admin\xFF" or u"admin\xFE". 

Also, u"admin\xFF" == u"admin\xFE" returns false.

KrzysztofCwalina:
Sorry, I keep clicking close on PRs, when I wanted to just close changes to a comment.

GrabYourPitchforks:
@KrzysztofCwalina I'm not talking about malicious code, I'm talking about trusted code operating on malicious input. We're all in agreement that validation needs to be performed, we're just seemingly trying to figure out when. I think my proposal offers the best performance while maintaining existing behaviors but I'd be interested in seeing alternatives.

KrzysztofCwalina:
@A-And, are you planing to update the PR so we can merge?

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -333,9 +333,9 @@ public bool Equals(JsonProperty other)
             public override int GetHashCode()
             {
                 int result = Object.GetHashCode();
-                result = result * 19 + (byte)_name[0];
-                result = result * 19 + (byte)_name[_name.Length - 1];
-                result = result * 19 + (byte)_name[_name.Length>>2];
+                result = result * 19 + (byte)_name.Bytes[0];
```
KrzysztofCwalina: In general, you should get Bytes once, store it in a local and then use it (if you need to use it more than once).

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerable.cs
```diff
@@ -0,0 +1,45 @@
...
+            {
+                return new CodePointEnumerator(_buffer);
+            }
+
+            public CodePointReverseEnumerator GetReverseEnumerator()
+            {
+                return new CodePointReverseEnumerator(_buffer);
+            }
+
+            public int Count()
```
KrzysztofCwalina: Why do we need it? If we do need it, can it be non-public?

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerator.cs
```diff
@@ -0,0 +1,116 @@
...
+
+            public unsafe CodePointEnumerator(ReadOnlySpan<byte> buffer) : this()
+            {
+                _buffer = buffer;
+
+                Reset();
+            }
+
+            // TODO: Name TBD
+            public int PositionInCodeUnits
```
KrzysztofCwalina: This could be written as
```c#
public int PositionInCodeUnits => IsOnResetPosition()? -1 : _index;
```

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -125,14 +120,18 @@ public CodePointEnumerable CodePoints
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get
             {
-                // there is no need to check the boundaries -> Span is going to do this on it's own
-                return (byte)_buffer[i];
+                // Check bounds
+                if( i < 0 || i > _buffer.Length)
```
ahsonkhan: Shouldn't we change the bounds check to i >= _buffer.Length?

tests/System.Text.Utf8String.Tests/TestHelper.cs
```diff
@@ -22,7 +22,7 @@ public static void Validate(Utf8Span str1, Utf8Span str2)
             Assert.Equal(str1.Length, str2.Length);
             for (int i = 0; i < str1.Length; i++)
             {
-                Assert.Equal(str1[i], str2[i]);
+                Assert.Equal(str1.Bytes[i], str2.Bytes[i]);
```
ahsonkhan: Can you use SequenceEqual instead of a for loop (element by element) comparison?

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,753 @@
...
+        {
+            get
+            {
+                return new CodePointEnumerable(_buffer);
+            }
+        }
+
+        private byte this[int i]
+        {
+                        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ahsonkhan: nit: spacing in this method is off

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerable.cs
A-And: This is the implementation of Utf8String's previous Enumerable. I have left it as it was implemented initially. Would it be worth changing it?

src/System.Text.Utf8String/System/Text/Utf8String.cs
A-And: Wouldn't that lead to the index being outside of the bounds of the array? (i.e. the buffer runs from 0 to Length - 1)

ahsonkhan: > Wouldn't that lead to the index being outside of the bounds of the array? (i.e. the buffer runs from 0 to Length - 1)

Currently, the check accounts for buffer from 0 to Length.

I might be mistaken, but if `i = _buffer.Length` and you don't throw (during bounds check), we have the following:

`return _buffer[_buffer.Length];` <= isn't this out of range?

src/System.Text.Json.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -333,9 +333,10 @@ public bool Equals(JsonProperty other)
             public override int GetHashCode()
             {
                 int result = Object.GetHashCode();
-                result = result * 19 + (byte)_name.Bytes[0];
-                result = result * 19 + (byte)_name.Bytes[_name.Length - 1];
-                result = result * 19 + (byte)_name.Bytes[_name.Length>>2];
+                ReadOnlySpan<byte> nameBytes = _name.Bytes;
+                result = result * 19 + (byte)nameBytes[0];
```
ahsonkhan: nit: Are the explicit casts to byte necessary? You are accessing an element of a ROSpan\<byte\>.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -101,7 +101,7 @@ public CodePointEnumerable CodePoints
 
         private byte this[int i]
         {
-                        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
              get
                        {
```
ahsonkhan: nit: spacing in this whole method is off.

src/System.Text.Utf8String/System/Text/Utf8String.cs
A-And: It is! Sorry and thanks.

src/System.Text.Utf8String/System/Text/Utf8Span.Enumerator.cs
KrzysztofCwalina: Let's just remove this type.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
KrzysztofCwalina: Let's leave it for now, since it exists in Utf8String.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
KrzysztofCwalina: Let's just remove this ctor.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
KrzysztofCwalina: Could you add a TODO in the source to optimize it later? Thanks.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
KrzysztofCwalina: I think it's ok to leave it for now and do a holistic cleanup of the index based APIs.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
KrzysztofCwalina: Let's just use the Encoding.UTF8.GetBytes.

src/System.Text.Utf8String/System/Text/Utf8Span.cs
KrzysztofCwalina: I think it's ok to leave it for now and do a holistic cleanup of the index based APIs.

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/TcpConnectionFormatter.cs
KrzysztofCwalina: Or even better:
```c#
static Utf8String s_terminator = new Utf8String("0\r\n\r\n");
```

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointEnumerable.cs
KrzysztofCwalina: Enumerating coding units (bytes) does not make much sense. Enumerating code points does make sense, but it's not this enumerator, is it? And so, I would just remove/hide all the enumerators that enumerate bytes (code units)

src/System.Text.Utf8String/System/Text/Utf8String.cs
KrzysztofCwalina: @A-And, just adding comment so that this change suggestion is not lost in the comments.

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointReverseEnumerator.cs
```diff
@@ -0,0 +1,121 @@
...
+    {
+        // TODO: Name TBD
+        public ref struct CodePointReverseEnumerator
+        {
+            private ReadOnlySpan<byte> _buffer;
+            private int _index;
+            private int _currentLenCache;
+            private const int ResetIndex = -Utf8Helper.MaxCodeUnitsPerCodePoint - 1;
+
+            public unsafe CodePointReverseEnumerator(ReadOnlySpan<byte> buffer) : this()
```
ahsonkhan: This method (amongst others) doesn't need to be unsafe.

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointReverseEnumerator.cs
```diff
@@ -0,0 +1,121 @@
...
+                    if (IsOnResetPosition())
+                    {
+                        return -1;
+                    }
+
+                    return _index;
+                }
+            }
+
+            public unsafe uint Current
```
ahsonkhan: This method (amongst others) doesn't need to be unsafe.

src/System.Text.Utf8String/System/Text/Utf8Span.Enumerator.cs
```diff
@@ -0,0 +1,41 @@
...
+            private int _index;
+
+            internal Enumerator(ReadOnlySpan<byte> buffer)
+            {
+                _buffer = buffer;
+                _length = buffer.Length;
+                _index = -1;
+            }
+
+            public byte Current
```
ahsonkhan: nit:
```C#
public byte Current => _buffer[_index];
```

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,753 @@
...
+        /// Returns length of the string in UTF-8 code units (bytes)
+        /// </summary>
+        public int Length => _buffer.Length;
+
+        public Enumerator GetEnumerator()
+        {
+            return new Enumerator(_buffer);
+        }
+
+        public CodePointEnumerable CodePoints
```
ahsonkhan: nit: 
```C#
public CodePointEnumerable CodePoints => new CodePointEnumerable(_buffer); 
```

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -0,0 +1,753 @@
...
+        public Utf8Span(RuntimeFieldHandle utf8Data, int length) : this(CreateArrayFromFieldHandle(utf8Data, length))
+        {
+        }
+
+        public static explicit operator Utf8Span(ArraySegment<byte> utf8Bytes)
+        {
+            return new Utf8Span(utf8Bytes);
+        }
+
+        static byte[] CreateArrayFromFieldHandle(RuntimeFieldHandle utf8Data, int length)
```
ahsonkhan: nit: can you please mark private methods explicitly with private keyword?

src/System.Text.Utf8String/System/Text/Utf8String.CodePointEnumerable.cs
```diff
@@ -4,18 +4,22 @@
...
     {
         public ref struct CodePointEnumerable 
         {
-            private ReadOnlySpan<byte> _buffer;
+            private byte[] _buffer;
 
             public CodePointEnumerable(byte[] bytes, int index, int length)
             {
-                _buffer = new ReadOnlySpan<byte>(bytes, index, length);
+                _buffer = new byte[length - index];
```
ahsonkhan: Should we add bounds checks on index and length and throw if args are out of range?

src/System.Text.Utf8String/System/Text/Utf8String.cs
```diff
@@ -12,29 +12,45 @@
...
+            {
+                throw new ArgumentOutOfRangeException("index");
+            }
+
+            if (length < 0)
+            {
+                throw new ArgumentOutOfRangeException("length");
+            }
+
+            if (index + length > utf8bytes.Length)
```
ahsonkhan: index + length could result in integer overflow

Consider something like the following for the bounds checks:
https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Span.cs#L58-L63
```C#
            if (utf8bytes == null)
                // throw null exception
            if ((uint)index > (uint)utf8bytes.Length || (uint)length > (uint)(utf8bytes.Length - index))
                // throw out of range exception
```

tests/System.Text.Utf8String.Tests/Utf8Span.RandomTests.cs
```diff
@@ -0,0 +1,967 @@
...
+            new object[] { 0, ""},
+            new object[] { 4, "1258"},
+            new object[] { 1, "\uABCD"},
+            new object[] { 2, "a\uABEE"},
+            new object[] { 3, "a\uABEEa"},
+            new object[] { 4, "a\uABEE\uABCDa"}
+        };
+
+        //[Theory(Skip = "System.InvalidProgramException : Common Language Runtime detected an invalid program."), MemberData("LengthInCodePointsTestCases")]
+        public void LengthInCodePoints(int expectedLength, string str)
```
ahsonkhan: Why are we skipping this test?

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointReverseEnumerator.cs
KrzysztofCwalina: @A-And, is that something you still want to change?

src/System.Text.Utf8String/System/Text/Utf8Span.CodePointReverseEnumerator.cs
KrzysztofCwalina: @A-And, is that something you still want to change?

src/System.Text.Utf8String/System/Text/Utf8String.CodePointEnumerable.cs
A-And: Certainly. Added!

KrzysztofCwalina:
I will merge to avoid getting the branch too stale. We can cleanup the remaining issues in subsequent PRs.

PR #1812, opened 2017-10-03
Make Tensors use Span/Memory instead of arrays

### LINKED ISSUE #1680: Should Tensor<T> be backed by Buffer<T> instead of T[]?
The idea behind Tensor is to provide multi-dimensional array semantics on top of a segment of memory.   Today that is done using `T[]` but `Span<T>` *may* be a better fit. 

One thought about this is around how well interop works with Span<T>.  @stephentoub is it possible to pass a Span<T> to a pinvoke without copy?

ericstj:
This permits Tensors to be backed with native memory, a slice of a managed array, or an
array (or really anything that can implement OwnedMemory abstraction).

Additionally where we exposed an array or a ReadOnlyCollection, expose a ReadOnlySpan
instead (and avoid a copy).

Fixes #1680 /cc @eerhardt @ahsonkhan @KrzysztofCwalina


src/System.Numerics.Tensors/System/Numerics/ArrayUtilities.cs
```diff
@@ -69,12 +69,17 @@ public static bool IsDescending(int[] values)
             return true;
         }
 
+        public static int[] GetStrides(int[] dimensions, bool reverseStride = false)
```
KrzysztofCwalina: You know there is an implicit cast from int[] to ReadOnlySpan<int>?

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
```diff
@@ -5,12 +5,12 @@ namespace System.Numerics
 {
     public class DenseTensor<T> : Tensor<T>
     {
-        private readonly T[] backingArray;
+        private readonly Memory<T> memory;
```
KrzysztofCwalina: Maybe we should have two dense tensor subclasses: array backed and Memory backed. The array backed would be faster for indexing operations (as it does not have to fish out the span) and the Memory backed would allow the flexibility of overlaying native buffers at some performance cost related to fishing the span out of Memory.

KrzysztofCwalina: BTW, I think it might be worth doing it just for the dense tensor, where the indexer is very fast. The sparse one is doing so much in the indexer that I think the difference does not matter.

src/System.Numerics.Tensors/System/Numerics/TensorArithmetic.cs
```diff
@@ -12072,28 +12716,33 @@ public void Contract(DenseTensor<uint> left, DenseTensor<uint> right, int[] left
...
-                result.Buffer[resultIndex] = sum;
+                resultSpan[resultIndex] = sum;
             }
         }
         public void Decrement(DenseTensor<uint> tensor, DenseTensor<uint> result)
         {
 
+            var resultSpan = result.Buffer.Span;
+            var tensorSpan = tensor.Buffer.Span;
             for(int i = 0; i < result.Length; i++)
```
KrzysztofCwalina: It might be worth changing the loop to use resultSpan.Length. The JIT is doing some bounds checks optimizations that I am not sure would kick in if you are not using the length of the same thing as you loop over.

src/System.Numerics.Tensors/System/Numerics/ArrayUtilities.cs
ericstj: Funny enough, it was calling `GetStrides(Array array)` rather than `GetStrides(ReadOnlySpan<int> dimensions, bool reverseStride = false)` when passing in an int[] (previously it would call the latter when it had an int[] in the signature), I renamed the array overload so I can delete this one now.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
```diff
@@ -986,7 +1104,7 @@ public void Reshape(TensorConstructor tensorConstructor)
             };
 
             var tensor = tensorConstructor.CreateFromArray<int>(arr);
-            var actual = tensor.Reshape(3, 2);
+            var actual = tensor.Reshape(new[] { 3, 2 });
```
KrzysztofCwalina: BTW, do you know that you can now allocate spans on the stack in safe context? Maybe even the following would work:
```c#
var actual = tensor.Reshape(stackalloc int[] { 3, 2 });
```

cc: @VSadov

KrzysztofCwalina: Speaking of which: @VSadov / @jaredpar, are you guys still thinking about non allocating params Span<T>?

src/System.Numerics.Tensors/System/Numerics/ArrayUtilities.cs
KrzysztofCwalina: Why do we even have the Array taking method?

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -35,7 +35,8 @@ public static Tensor<T> CreateIdentity<T>(int size, bool columMajor)
         /// <returns></returns>
         public static Tensor<T> CreateIdentity<T>(int size, bool columMajor, T oneValue)
         {
-            var result = new DenseTensor<T>(new[] { size, size }, columMajor);
+            var dimensions = new[] { size, size }.AsReadOnlySpan();
```
ahsonkhan: Consider re-writing as follows since it is clearer that dimensions is a Span of int.
```C#
ReadOnlySpan<int> dimensions = new[] {size, size};
```

src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
```diff
@@ -385,35 +397,35 @@ public override Tensor<T> Reshape(params int[] dimensions)
...
         }
 
         public override CompressedSparseTensor<T> ToCompressedSparseTensor()
         {
-            var newValues = new T[NonZeroCount];
-            Array.Copy(values, newValues, NonZeroCount);
-
-            var newIndicies = new int[NonZeroCount];
-            Array.Copy(indices, newIndicies, NonZeroCount);
+            var newValues = values.Span.Slice(0, nonZeroCount).ToArray();
```
ahsonkhan: Re-write as follows?
```C#
var newValues = values.Slice(0, nonZeroCount).ToArray();
```

src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
```diff
@@ -385,35 +397,35 @@ public override Tensor<T> Reshape(params int[] dimensions)
...
+            var newIndicies = indices.Span.Slice(0, nonZeroCount).ToArray();
```
ahsonkhan: same here.

src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
jkotas: Slicing of Span is cheaper than slicing of Memory. But it does not matter much here because `ToArray()` will dominate the cost in either case.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
jaredpar: Definitely thinking about it and something i want to do. Not convinced that we should use it Span directly. Been looking at a few ideas here.

src/System.Numerics.Tensors/System/Numerics/ArrayUtilities.cs
ericstj: Yeah, dead code, its been deleted.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
ericstj: I'll use @KrzysztofCwalina's stackalloc suggestion.

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
ericstj: This is what I was trying to tease apart in the issue discussion.  Perhaps a different field on DenseTensor?  I might roll this into an issue to do a separate investigation.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
ericstj: Actually, can you point me at the safe syntax that uses stackalloc and initializes inline?  I've tried a number of things and get messages that tells me the compiler doesn't know what I'm trying to do.

src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
```diff
@@ -202,14 +205,23 @@ private void EnsureCapacity(int min, int allocateIndex = -1)
...
                         {
-                            Array.Copy(values, allocateIndex, newValues, allocateIndex + 1, nonZeroCount - allocateIndex);
-                            Array.Copy(indices, allocateIndex, newIndices, allocateIndex + 1, nonZeroCount - allocateIndex);
+                            var valuesSpan = values.Span.Slice(allocateIndex, nonZeroCount - allocateIndex);
+                            var indicesSpan = indices.Span.Slice(allocateIndex, nonZeroCount - allocateIndex);
+
+                            var newValuesSpan = newValues.Span.Slice(allocateIndex + 1, nonZeroCount - allocateIndex);
+                            var newIndicesSpan = newIndices.Span.Slice(allocateIndex + 1, nonZeroCount - allocateIndex);
+
+                            valuesSpan.CopyTo(newValuesSpan);
```
ericstj: This felt like a lot of ceremony around just shifting bytes to the right.  Potential opportunity for API improvements on Span or Memory here.

One thing that comes to mind: a method to slice a span directly from memory, instead of accessing a span and slicing the span.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
KrzysztofCwalina: I checked, the syntax does not work now, though it would be nice if it did.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
KrzysztofCwalina: You can stackalloc, but you cannot currently initialize the data inline.

src/System.Numerics.Tensors/System/Numerics/TensorArithmetic.cs
ericstj: Typically I'll be accessing more than one span inside the loop and all are constrained by the same bounds check.  Any way to make the JIT aware of this?

tests/System.Numerics.Tensors.Tests/TensorTests.cs
ericstj: I couldn't find any syntax that works for stackalloc outside unsafe right now, not even https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md#language-support 4.

KrzysztofCwalina: There are numerous examples of this in corfxlab. See https://github.com/dotnet/corefxlab/blob/74ca8904031daf2d734104b90a6a5dadd22285c0/src/System.IO.Pipelines.Text.Primitives/ReadableBufferExtensions.cs#L115

ahsonkhan: > I couldn't find any syntax that works for stackalloc outside unsafe right now, not even https://github.com/dotnet/corefxlab/blob/master/docs/specs/span.md#language-support 4.

We should update the span design doc to show the correct stackalloc syntax.
Change:
```C#
void SafeMethod() {
    var buffer = stackalloc Span<byte>(128);
    PrimitiveFormatter.TryFormat(buffer, DateTime.Now, ...);
}
```
to:
```C#
void SafeMethod() {
    Span<byte> buffer = stackalloc byte[128];
    PrimitiveFormatter.TryFormat(buffer, DateTime.Now, ...);
}
```

ericstj: Do you suppose there will be any syntax for inline initialization, or is that out of scope?

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
KrzysztofCwalina: Since you have a hierarchy of types, why would you want a single class with fields for two classes (two modes). Why not have two subclasses of Tensor?

tests/System.Numerics.Tensors.Tests/TensorTests.cs
KrzysztofCwalina: @VSadov would have to answer this one.

KrzysztofCwalina: BTW, I updated the Span spec with the correct syntax.

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
ericstj: Right now the heirarchy is meant to represent different layouts, not necessarily different backing storage of a single layout, but it can be made to mean both.  I'm not thrilled about the idea of having to know to use one tensor with managed arrays but a different one with native memory.  I would imagine that if we cared a lot about perf we'd have that split everywhere and have the native ones work directly with pointers so that they wouldn't incur span overhead for access.

tests/System.Numerics.Tensors.Tests/TensorTests.cs
ericstj: Thanks, I've updated the product code to use stackalloc where I can.  I'm not going to bother for these tests until we have good inline initializer syntax as I think it makes things less readable.

KrzysztofCwalina: Makes sense

VSadov:
What is “inline initialization” in this context?  
I have some guesses, but not sure...

ericstj:
> What is “inline initialization” in this context?

The ability to do something like 
```c#
ReadOnlySpan<int> = stackalloc int[] { 1, 2, 3 };
```

PR #1823, opened 2017-10-05
Cut STP.Tests run time from 2.33 minutes to 0.5 seconds

AtsushiKan:
Finally took the time to track down why a test exercising
these high performance apis was taking so much longer
than any other test.

I don't think parsing a 2gb long string is a common enough
scenario to justify a 28,000% tax on every test cycle.
Regressions are caught much faster when people are incented
to run tests regularly.


KrzysztofCwalina:
Thanks! Looks good.

PR #1840, opened 2017-10-12
Removing the {RO}BufferDebuggerView as it is no longer needed.

ahsonkhan:
Memory (previously Buffer) was moved to corefx: https://github.com/dotnet/corefx/pull/24203 and https://github.com/dotnet/coreclr/pull/14130 along with the [DebuggerTypeProxy](https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/MemoryDebugView.cs).

Therefore, removing it from corefxlab.

Also making minor tweaks to the Parser performance tests.

cc @dotnet/corefxlab-contrib


KrzysztofCwalina:
looks good.

Issue #1841, opened 2017-10-12
priority queue spec should define decrease-key or some other priority update functionality


TimLovellSmith:
Re: https://github.com/dotnet/corefxlab/edit/master/docs/specs/priority-queue.md

The spec currently leaves as an open questions whether to support updating keys in PriorityQueue class. To close on this question... Yes! Updating priorities, especially decreasing them is important!

It is a generally useful operation, and it is possible to optimize special cases (mainly decrease key) enough, that it should be a first-class operation instead of always 'implemented' by the end user using remove + push.

If this is a question on whether to have it on the interface, why not? Some implementations of the interface might be no more efficient than calling remove element and then push element, but other implementations can be as efficient as possible.

It requires some thought to decide what the interface for collections where the elements contain the priority, or the queue works via `IComparable<T>` should be. Ideally (we can dream!) the collection would guarantee that the priority on the element is always consistent with the state of the priority queue. But a practical problem exists that the user can update his/her objects to be a new state but forget to notify the queue that priority has changed.

/cc @derekbjohnson 
/cc @pgolebiowski 
/cc @ebickle

pgolebiowski:
Please go through the frightfully long discussion over here: https://github.com/dotnet/corefx/issues/574. You'll see voices for and against. I've been fighting for this feature for a long time. Feel free to contribute to the discussion (better post your thoughts **here** rather than in that long thread though [just reference it]). 😄

TimLovellSmith:
Recapping, I think some key contentious points related update are:

-worry about how update should work when there are duplicate elements (which we think sometimes we want to support). Is updateFirst weird?
-somewhat unclear where the argument for/against supporting duplicates landed in the first place
-worry that it might be inefficient to have an additional lookup data structure for finding elements in the backing data structure, or give handles out, just to update priority of items. (Especially for scenarios where update is never performed!)

Now as a result, here is my take.

Look at the popular applications of PriorityQueues courtesy of Wikipedia.

Q. Do any of them require duplicate elements, in the sense that the 'element' stored in the queue can be identical, but stored with a different priority? (Note, that this is only possible if the priority is stored externally to the item in the queue.)

Bandwidth management: **no**, you could either be queuing at the packet level or the connection level, but they don't need to be duplicated.

Djikstra's algorithm, and other 'best-first' algorithms: **no**, each vertex only needs to appear once (and you would also be happy with 'deduplicate insert/update')

[...]

PR #1850, opened 2017-10-16
Solved priority queue design!

TimLovellSmith:
This is intended to address issue #1841 and all other design problems for priority queue. Now can .net finally have a priority queue please? :)

### DIFF
--- a/docs/specs/priority-queue.md
+++ b/docs/specs/priority-queue.md
@@ -94,7 +94,7 @@ public class PriorityQueue<TKey, TPriority> :
     public IEnumerator<KeyValuePair<TKey key, TPriority priority>> GetEnumerator();
     public bool Remove(KeyValuePair<TKey, TPriority> item);
 
-    public struct Enumerator : IEnumerator<KeyValuePair<TKey key, TPriority priority>> {} // enumerates the collection in arbitrary order, but with the least element first
+    public struct Enumerator : IEnumerator<KeyValuePair<TKey key, TPriority priority>> {} // enumerates the collection in some arbitrary order
     {
         // member declarations omitted for brevity
     }
--- a/docs/specs/priority-queue.md
+++ b/docs/specs/priority-queue.md
@@ -1,6 +1,6 @@
 ﻿# Priority queue proposal
 
-[Revised 8/22/20]
+[Revised 8/24/20]
 
 # Scenario and key assumptions.
 
@@ -69,12 +69,11 @@ public class PriorityQueue<TKey, TPriority> :
     public TKey Peek(); // throws InvalidOperationException if the queue is empty (like Queue<>)
     public void Enqueue(TKey key, TPriority priority); // throws ArgumentException if the key was already in the collection, like IDictionary<TKey,TPriority>.Add(), since  adding a distinct item twice with different priorities may be a programming error
     public void Update(TKey key, TPriority priority); // throws ArgumentException if the key was not in the collection
-    public void EnqueueOrUpdate(TKey key, TPriority priority); // doesn't throw, always updates priority of keys already in the collection
+    public bool EnqueueOrUpdate(TKey key, TPriority priority); // doesn't throw, always updates priority of keys already in the collection. Returns true if an item was newly enqueued.
     public bool TryDequeue(out TKey key, out TPriority priority); // returns false if the queue is empty
     public bool TryPeek(out TKey key, out TPriority priority); // returns false if the queue is empty
     
-    public void EnqueueOrDeccreasePriority(TKey key, TPriority priority); // doesn't throw, only updates priority of keys already in the collection if new priority is lesser
-    public void EnqueueOrIncreasePriority(TKey key, TPriority priority); // doesn't throw, only updates priority of keys already in the collection if new priority is greater
+    public bool EnqueueOrDecreasePriority(TKey key, TPriority priority); // doesn't throw, only updates priority of keys already in the collection if new priority is lesser. Returns true if an item was newly enqueued.
 
     // inherited from ICollection<KeyValuePair<TKey, TPriority>>
     public int Count { get; }
--- a/docs/specs/priority-queue.md
+++ b/docs/specs/priority-queue.md
@@ -1,6 +1,6 @@
 ﻿# Priority queue proposal
 
-[Revised 8/17/20]
+[Revised 8/19/20]
 
 # Scenario and key assumptions.
 
@@ -63,9 +63,9 @@ public class PriorityQueue<TKey, TPriority> :
     
     public TPriority Item[TKey] { get; set; } // can get priorities, update priorities, or add items and their priorities
 
-    public KeyValuePair<TKey, TPriority> Dequeue();  // throws InvalidOperationException if the queue is empty (like Queue<>)
+    public TKey Dequeue();  // throws InvalidOperationException if the queue is empty (like Queue<>)
+    public TKey Peek(); // throws InvalidOperationException if the queue is empty (like Queue<>)
     public void Enqueue(TKey key, TPriority priority); // throws ArgumentException if the key was already in the collection, like IDictionary<TKey,TPriority>.Add(), since  adding a distinct item twice with different priorities may be a programming error
-    public KeyValuePair<TKey, TPriority> Peek(); // throws InvalidOperationException if the queue is empty (like Queue<>)
     public bool TryDequeue(out TKey key, out TPriority priority); // returns false if the queue is empty
     public bool TryPeek(out TKey key, out TPriority priority); // returns false if the queue is empty
 
--- a/docs/specs/priority-queue.md
+++ b/docs/specs/priority-queue.md
@@ -1,281 +1,198 @@
-# Priority queue proposal
+﻿# Priority queue proposal
 
-## User data
+Before proposing the classes, let's explain the scenarios they meet...
 
-Before anything is proposed, let's start with what can be prioritized...
+## Scenario 1 - Priority queue - data items *are* intrinsically priorities
 
-### Input types
+They can be numbers, or any other `IComparable<>` type. Importantly, comparing
+any two items must *always return the same result*. In other words, relative item 
+priorities never change. So there is no 'update priority' operation.
 
-We want to be able to support all of the following configurations:
+## Scenario 2 - Priority queue - data items have *updatable* priorities
 
-1. User data *is separate* from the priority (two physical instances).
-2. User data *contains* the priority (as one or more properties).
-3. User data *is* the priority (implements `IComparable<T>`).
-4. Rare case: priority is obtainable via some other logic (resides in an object
-different from the user data).
+Sometimes a priority queue is used to track *changing* priority of a *set* of 
+unique objects (or *keys*). Priorities can change, and items may be added and 
+removed at any time. A priority is an *attached value* which just prioritizes 
+which object (or key) will be removed first.
 
-In this entire document, I will refer to the cases above with **(1)**, **(2)**,
-**(3)**, and **(4)**.
+This is like a dictionary where the objects are the unique *keys*, and the 
+priorities are their *values* - with the addition of being able to easily 
+dequeue items in priority order.
 
-### Ideal solution
+## Note: smallest priority comes first
 
-Obviously, our solution should be flexible enough to (respectively):
+The queues are *smallest-priority-first*, i.e. small priorities
+are the 'highest' priority.
 
-1. Simply accept two separate instances. The user should not be forced to create
-a wrapper class for the two types only because of our API limitations.
-2. Accept an element that already has the priority in it, without duplication
-(no copying).
-3. Be able to use `IComparable<T>` and don't expect an additional priority.
-4. Be able to execute some additional logic that retrieves the priority for a
-given element.
+## Prerequisites
 
-### Our approach
+### Priorities must have an IComparer 
 
-In order to be able to consume all of that, we need two types of priority
-queues:
+Priorities have to function like numbers. There needs to be a 
+'total ordering' on the set, which can optionally be supplied via the 
+`IComparer<>` abstraction. If you use a numeric type like `int`, 
+there is a default comparer.
 
-* `PriorityQueue<T>`,
-* `PriorityQueue<TElement, TPriority>`.
+### Keys in a priority dictionary are equality comparable (and hashable)
 
-## `PriorityQueue<T>`
+This can be optionally supplied via the `IEqualityComparer<>` abstraction. 
+But by default it uses `Equals()` and  `GetHashCode()`.
+
+# `PriorityQueue<T>`
+
+`PriorityQueue<T>` is a queue of items, where the least item (according to 
+the comparer) will be dequeued first. Duplicates are allowed.
 
 ```csharp
-public class PriorityQueue<T> : IQueue<T>
+public class PriorityQueue<T>:
+    ICollection<T>
+    IEnumerable<T> // inherited from ICollection<T>
+    IEnumerable
 {
     public PriorityQueue();
-    public PriorityQueue(IComparer<T> comparer);
-    public PriorityQueue(IEnumerable<T> collection);
-    public PriorityQueue(IEnumerable<T> collection, IComparer<T> comparer);
+    public PriorityQueue(IComparer<T> priorityComparer);
 
-    public IComparer<T> Comparer { get; }
+    public PriorityQueue(IEnumerable<T> items);
+    public PriorityQueue(IEnumerable<T> items, IComparer<T> priorityComparer);
+
+    // (inherited from ICollection<T>)
     public int Count { get; }
+    public bool IsReadOnly { get; } // returns false
 
-    public bool IsEmpty { get; }
+    // new
+    public IComparer<T> Comparer { get; } // compares priorities using .CompareTo(T)
+    
+    // (inherited from IEnumerable<T>)
+    public IEnumerator<T> GetEnumerator(); // enumeration is not in strict priority order, so that it is O(n) (inherited from IEnumerable<T>)
+    public IEnumerator GetEnumerator();
+
+    // (inherited from ICollection<T>)
+    public void Add(T item); // synonym for Enqueue
     public void Clear();
-    public bool Contains(T element);
+    public bool Contains(T item); // performance is worst-case O(n)
+    public void CopyTo (T[] array, int arrayIndex); // performance is O(n)
+    public bool Remove(T item); // removes the object from the collection. Returns true if successfully removed.
 
-    public void Enqueue(T element);
+    // new
+    public void Enqueue(T item); // performance is O(log n)
+    public T Dequeue();  // performance is O(log n) throws InvalidOperationException if the queue is empty (like Queue<>)
+    public T Peek(); // performance is O(1) throws InvalidOperationException if the queue is empty (like Queue<>)
+    public bool TryPeek(out T item); // returns false if the queue is empty (like Queue<>)
+    public bool TryDequeue(out T item); // returns false if the queue is empty (like Queue<>)
 
-    public T Peek();
-    public T Dequeue();
-    public bool Remove(T element);
-
-    public bool TryPeek(out T element);
-    public bool TryDequeue(out T element);
-
-    public IEnumerator<T> GetEnumerator();
-    IEnumerator IEnumerable.GetEnumerator();
-
-    public struct Enumerator : IEnumerator<T>
+    public struct Enumerator : IEnumerator<T> {} // enumerates the collection in priority order
     {
-        public T Current { get; }
-        object IEnumerator.Current { get; }
-        public bool MoveNext() => throw new NotImplementedException();
-        public void Reset() => throw new NotImplementedException();
-        public void Dispose() => throw new NotImplementedException();
+        // member declarations omitted for brevity
     }
 }
 ```
 
-### Scenarios
-
-#### (2)
-
-Custom class with a priority inside:
+### Example usage
 
 ```csharp
-public class MyClass
+var queue = new PriorityQueue<int>();
+queue.Enqueue(3);
+queue.Enqueue(2);
+queue.Enqueue(5);
+queue.Enqueue(3);
+Assert.Equal(2, queue.Dequeue());
+Assert.Equal(3, queue.Dequeue());
+Assert.Equal(3, queue.Dequeue());
+Assert.Equal(5, queue.Dequeue());
+Assert.True(queue.Empty());
+```
+
+# `PriorityDictionary<TKey, TPriority>`
+
+`PriorityDictionary<TKey, TPriority>` is a dictionary of items and their 
+priorities. If not otherwise  specified, it will use a 'default equality 
+comparer' for comparing objects (of type `TKey`) and a 'default priority comparer' 
+for comparing priorities (of type `TPriority`). Items or keys in the 
+dictionary must be unique.
+
+```csharp
+public class PriorityDictionary<TKey, TPriority> :
+    IDictionary<TKey, TPriority>,
+    ICollection<KeyValuePair<TKey key, TPriority priority>>, // inherited from IDictionary<TKey,TPriority>
+    IEnumerable<KeyValuePair<TKey key, TPriority priority>>, // inherited from ICollection<KeyValuePair<TKey,TPriority>>
+    IEnumerable // inherited from IEnumerable<KeyValuePair<TKey, TPriority>>
 {
-	public double Priority { get; }
-}
-```
+    public PriorityDictionary();
+    public PriorityDictionary(IComparer<TPriority> priorityComparer);
+    public PriorityDictionary(IComparer<TPriority> priorityComparer, IEqualityComparer<TKey> equalityComparer);
 
-The user defines their own comparer, for example:
+    public PriorityDictionary(IEnumerable<TKey, TPriority> dictionary);
+    public PriorityDictionary(IEnumerable<TKey, TPriority> dictionary, IComparer<TPriority> priorityComparer);
+    public PriorityDictionary(IEnumerable<TKey, TPriority> dictionary, IComparer<TPriority> priorityComparer, IEqualityComparer<TKey> equalityComparer);
 
-```csharp
-var comparer = Comparer<MyClass>.Create((a, b) =>
-{
-    return a.Priority.CompareTo(b.Priority);
-});
-```
+    // new
+    public IComparer<TPriority> PriorityComparer { get; }
+    public IEqualityComparer<TKey> EqualityComparer { get; }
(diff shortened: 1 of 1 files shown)


ahsonkhan:
cc @safern, @ianhays, @KrzysztofCwalina 
Link to original discussion (and specific request for this PR): [#574](https://github.com/dotnet/corefx/issues/574#issuecomment-308609496).

docs/specs/priority-queue.md
```diff
@@ -1,71 +1,73 @@
...
-We want to be able to support all of the following configurations:
+There are two possible scenarios in which to use a priority queue:
 
-1. User data *is separate* from the priority (two physical instances).
-2. User data *contains* the priority (as one or more properties).
-3. User data *is* the priority (implements `IComparable<T>`).
-4. Rare case: priority is obtainable via some other logic (resides in an object
-different from the user data).
+1. Each item of user data is *assigned* a *separate* priority (priority is a separate object from the elements in the collection.) We will sometimes call this the *extrinsic priority* scenario.
+2. User data either *is itself a priority*, or has some other *natural* priority which can be derived from the element. (It might have a priority property, a it might be a *computed* priority using some function mapping elements to priorities.) We will sometimes call this the *intrinsic priority* scenario.
```
ahsonkhan: > (It might have a priority property, a it might be a *computed* priority using some function mapping elements to priorities.)

This needs to be reworded since "...property, a it might be a ..." doesn't make sense.

docs/specs/priority-queue.md
```diff
@@ -77,9 +79,9 @@ public class PriorityQueue<T> : IQueue<T>
 }
 ```
 
-### Scenarios
+### Examples
 
-#### (2)
+#### (1)
```
ahsonkhan: Similar to this change (changing heading from Scenario to Examples and updating the numbers), consider changing it here too: https://github.com/TimLovellSmith/corefxlab/blob/d61eda595a030aee4f0563bebb400052d8ab0c13/docs/specs/priority-queue.md#scenario

docs/specs/priority-queue.md
```diff
@@ -1,71 +1,73 @@
...
-**(3)**, and **(4)**.
+### Uniqueness and changes in priority.
 
-### Ideal solution
+This implementation of priority queue is actually designed to provide a DICTIONARY-like or SET-like interface, where each element is unique (like a key), and priorities are mutable. This allows you to easily use the data structure for many popular scenarios requiring priority queues. 
+All elements are equality comparable, and each distinct element (according to equality comparison) can have at most a unique occurence in the queue, and be assigned a single priority value.
+That priority value may be updated, over time.
 
-Obviously, our solution should be flexible enough to (respectively):
+This means there are really two types of comparators (comparison operators) involved in the definition of a priority queue class - one for comparing *elements* pairwise for equality, and one for comparing *element priorities* pairwise to see which is higher priority. It is unnecessary, and usually undesirable for the same comparator to be used for both priority comparisons and equality comparisons.
```
safern: I don't really understand why we need 2 different comparers if one comparer should be enough to define if two elements are equal and if an element is greater than the other one based on its priority.

TimLovellSmith: @safern there's two completely different types of comparison here. 
Comparison type #1: is this element the same element which was already inserted in the queue? (Used for finding an element to update its priority)
Comparison type #2: is this element a higher priority than that element? (Used for sorting elements so they get dequeued in priority order)

TimLovellSmith: @safern This is also why the type parameters of the comparers are different. (One is a comparer of of `<TElement>` the other is of `<TPriority>`.)

TimLovellSmith:
Open question: is there really any need to define TPriorityQueue<TElement>? It is pretty much the same as SortedSet<TElement>, although I tried to add some update semantics for it.

docs/specs/priority-queue.md
ianhays: Wait, nevermind. That's a different thing.

We need the second comparer for both cases - for the second case it is because just because two items have the same priority doesn't mean they're necessarily equal.

safern: I see. That way makes sense.

docs/specs/priority-queue.md
```diff
@@ -1,71 +1,73 @@
...
     public T Dequeue();
-    public bool Remove(T element);
 
     public bool TryPeek(out T element);
     public bool TryDequeue(out T element);
 
     public IEnumerator<T> GetEnumerator();
     IEnumerator IEnumerable.GetEnumerator();
 
+    // + various other ISet interface methods (ExceptWith, Overlaps, IsProperSubsetOf, IsSubsetOf, IsSuperSetOf, IntersectWith, UnionWith, etc)
```
ianhays: nit: what's the plus sign doing in there?

docs/specs/priority-queue.md
safern: > We need the second comparer for both cases - for the second case it is because just because two items have the same priority doesn't mean they're necessarily equal.

That makes sense. I was not taking that in consideration either.

docs/specs/priority-queue.md
```diff
@@ -202,80 +214,13 @@ queue.Enqueue(userData, priority);
 
 To both priority queues:
 
-* If the `IComparer<T>` is not delivered, `Comparer<T>.Default` is summoned.
+* If the `IComparer<T>` is not delivered, `Comparer<T>.Default` is used.
```
ianhays: "delivered" is a weird word choice here. "Provided" would be better. 

Same comment for EqualityComparer

docs/specs/priority-queue.md
ianhays: > I was not taking that in consideration either.

+1. We should add a comment or reword it so it's a bit more clear that the reason for two comparers is that two elements can have the same priority but be unequal.

docs/specs/priority-queue.md
```diff
@@ -124,46 +129,53 @@ Then simply call the default constructor (`Comparer<T>.Default` is assumed):
...
-the comparer.
-
 ## `PriorityQueue<TElement, TPriority>`
 
 ```csharp
 public class PriorityQueue<TElement, TPriority> :
-    IEnumerable,
-    IEnumerable<(TElement element, TPriority priority)>,
-    IReadOnlyCollection<(TElement element, TPriority priority)>
+    IDictionary<TElement, TPriority>,
```
KrzysztofCwalina: What happens when two elements that are equal end up in the queue?

docs/specs/priority-queue.md
safern: > We should add a comment or reword it so it's a bit more clear that the reason for two comparers is that two elements can have the same priority but be unequal.

Agree.

docs/specs/priority-queue.md
TimLovellSmith: @KrzysztofCwalina Hm. I thought I understood this, but expanding it out I realize I need to make a tweak. The guideline I think we should follow here is that the classes should follow the contracted behavior for `.Add()` of the interfaces they implement. Specifically `PriorityQueue<T>` should just 'noop' (at least I believe that is what ISet<T>.Add() does - or does it 'replace' the element with itself?) and return false to indicate the element was already present. I need to fix the return type of `PriorityQueue<TElement,TPriority>.Add()` to void to match `IDictionary<TElement,TPriotity>`and it should throw `ArgumentException`.

nietras:
I can concur that we really need a PriorityQueue, but preferable one that allows updates of the priority based on the element. E.g. add a method (which does not seem to be part of proposal):
```csharp
public class PriorityQueue<TElement, TPriority>
{
    public void Update(TElement element, TPriority priority);
}
```
And preferable with O(1) complexity. This allows this to be used for say Dijkstra's shortest path for example.

Also consider that it might be beneficial to also be able to query in constant time also the "tail" of the priority queue. `SortedSet<T>` has `Min/Max` for that. Here it seems `Peek` if for getting the "head".

TimLovellSmith:
Hi @nietras 
I think you didn't understand that the design's current intention is to allow updates of the priority based on the element, in the 'external priority' case - it is done simply like this: 

`priorityQueue[elementKey] = newValue;`

And yes, this is definitely intended to have O(1) amortized complexity and be usable for Dijkstra's algorithm.

The suggestion to have constant time query of back of the priority queue is an interesting one. The main drawback I foresee is that it is not easy to implement efficiently unless your priority queue is in fact a sorted set underneath.

nietras:
@TimLovellSmith ah I missed that. My mistake. Great that sounds good.

> it is not easy to implement efficiently unless your priority queue is in fact a sorted set underneath.

No you are right, SortedSet can be used for that but it doesn't handle some of the use cases of priority queue. Min-max can be handled by two priority queues too, so better to be really good at the normal use case which is either min or max-heap.

Will there be an experimental implementation of this soon?

TimLovellSmith:
Here's a few updates + fixes based on the discussion so far, thanks for feedback! I think it might be clear enough for someone to have a go at implementing now. 
@nietras I'm now not sure the interface itself would really guarantee O(1) update key, I think it might only guarantee something a little easier to achieve like O(log(n)). It is intended to support increase key as well as decrease key for completeness.

nietras:
> not sure the interface itself would really guarantee O(1) update key,

@TimLovellSmith  that was my concern. Should it not be considered then to add a say `IncreasePriority(TKey key, TPriority priority)` or something named like that to the interface to ensure relaxation can be done optimally?

TimLovellSmith:
@nietras Well, I also have some doubts about whether an implementation would actually be the fastest for typical everyday use cases even though its O(1). But I figure that kind of detail can perhaps best be decided by implementors over time as opposed to trying to bake it into the interface. 
Unless you are asking for something which is really a different operation from update-priority, like a conditional-update-priority?

docs/specs/priority-queue.md
```diff
@@ -1,70 +1,99 @@
...
+
+`PrioritySet<T>` is a set of items with intrinsic priorities. It must be initialized with a 'priority comparer' which understands how to compare the intrinsic priorities of each item.
+
+It differs from `SortedSet<T>` because it allows separate items to have equal priorities, and because it does not maintain strict ordering at all times, and because it allows that priorities are mutable.
+
+If you provide `DefaultComparer<T>` as your priority comparer, and never mutate priorities, you will get similar behavior to `SortedSet<T>`, where priorities are the items and priorities are all distinct.
 
 ```csharp
-public class PriorityQueue<T> : IQueue<T>
+public class PrioritySet<T> : ISet<T>, ICollection<T>, IEnumerable<T>, IEnumerable,
```
nietras: nit: class name is now `PrioritySet` but the ctor names are not. I would prefer queue and follow c++ `priority_queue` inspired design http://en.cppreference.com/w/cpp/container/priority_queue

nietras: And I don't think `ISet<T>` is appropriate, what wil `IsSubsetOf` mean? etc.

docs/specs/priority-queue.md
```diff
@@ -1,70 +1,99 @@
...
     public T Peek();
     public T Dequeue();
-    public bool Remove(T element);
 
-    public bool TryPeek(out T element);
-    public bool TryDequeue(out T element);
+    public bool TryPeek(out T item);
+    public bool TryDequeue(out T item);
+
+    public void NotifyPriorityChanged(T item);
```
nietras: not sure what the `NotifyPriorityChanged` methods are for?

docs/specs/priority-queue.md
```diff
@@ -1,70 +1,99 @@
...
+    public void NotifyPrioritiesChanged(IEnumerable<T> items);
 
     public IEnumerator<T> GetEnumerator();
-    IEnumerator IEnumerable.GetEnumerator();
+
+    // ommitted: declaration of various other ICollection, IEnumerable, and ISet interface members (ExceptWith, Overlaps, IsProperSubsetOf, IsSubsetOf, IsSuperSetOf, IntersectWith, UnionWith, etc.)
```
nietras: exactly, how would these work in the context of a priority queue?

nietras:
> figure that kind of detail can perhaps best be decided by implementors over time as opposed to trying to bake it into the interface

@TimLovellSmith I agree 😄 

> Unless you are asking for something which is really a different operation from update-priority, like a conditional-update-priority?

No basically, just "increase-key" (or "decrease-key" depending on comparer), which I assume can be detected when the index operator is used.

One thing, I am not absolutely happy about is the fact that one cannot use a "value type" comparer, to avoid potential virtual call costs on compares. But no other type in .NET does that so that is probably too involved.

docs/specs/priority-queue.md
TimLovellSmith: @nietras I imagine that they would work just as if the priority queue is consider as 'a set of elements... which just happen to have priorities' e.g. IntersectWith removes all elements from the priority queue which are not part of the target collection.

docs/specs/priority-queue.md
TimLovellSmith: @nietras It is the simplest way I could think for PrioritySet is to support elements whose priority mutates over time.
I am still conflicted on whether it is a good feature, as it can let in consistency bugs - maybe it is better to force people to model mutable priority scenarios using PriorityQueue and not PrioritySet...

docs/specs/priority-queue.md
nietras: @TimLovellSmith right like `SortedSet` and `HashSet`, I understand that, but does that make sense in this case? There might not be an as efficient way to do those? Depending on implementation. I guess the question is what use case are there for set operations on a priority queue? Perhaps there are valid ones, I just wondered why? :)

docs/specs/priority-queue.md
nietras: @TimLovellSmith I guess my main objection here is the name, it is really poor in my view. It is not notifying anything. If anything it sounds more like you want it to do an `Update` but how would that work if the item itself is both priority and key? How can it then update the old item? So this kind of implicitly assumes the `item` has **two** properties a key and a priority, which does not seem to be a good assumption.

If that is I understand the purpose here. It is like the `priorityQueue[key] = priority` just with `item` being both right? So basically changing an item should perhaps involve first deleting the old and then adding new? Or with a proper `Replace(T old, T new)` method.

docs/specs/priority-queue.md
TimLovellSmith: @nietras Maybe the best answer here is that we don't really need PrioritySet at all. There's no PrioritySet scenario that can't instead be implemented with external priority PriorityQueue, and the downsides of having mutable intrinsic priorities are a huge pit of failure.

ianhays:
@TimLovellSmith this has been open for a while, are you happy with the current state or do you have more changes to make?

TimLovellSmith:
@ianhays @nietras 
Thanks for all the feedback. My thinking now about PrioritySet has since become that it was a cute idea but not a useful and good idea. Its not useful because you can just do every PrioritySet scenario using PriorityQueue instead. And its not good because it is a pit of failure with inconsistent data structure issues, I'm sure that's not the kind of thing standard library maintainers will want to help people debug. 

I still think it is nice for PriorityQueue to be a dictionary, since it should be a 1-1 mapping - every queue item should have a unique priority.

I'm going to do one more revision to remove PrioritySet from the design.

nietras:
A quick note, I really don't want a 1-1 mapping allowing only unique keys,
there's SortedDictionary for that. This proposal loses most it's value
then.

TimLovellSmith:
@nietras What scenario would you need to give the same item multiple priorities for? And would you use update key in that scenario? 
IMO PriorityQueue's main value is its different algorithmic complexity guarantees compared to sorted dictionary.

TimLovellSmith:
@nietras 
I should probably write some more about my thinking on why no duplicates. Then I might realize whether it is wrong or not...

I think it is ideal for priority queue to support both reference and value types.
e.g. you can just do a priority queue of numbers as well as objects.
This tends to be a pattern of 'the object is the priority'.
This scenario tends to not require any priority key updates, it instead works by removal / adding to the queue.

And actually this proposal is doing a pretty poor job of supporting that scenario, because what you have to do to have numbers in the queue with duplicates is either provide an equality comparer where objects are never equal to each other, or just provide dummy unique 'items' (e.g. increment an int64) with the priority values.

I feel another proposal version coming on...

TimLovellSmith:
I see a test failure logged from CI on this [xUnit.net 00:00:30.4812956] System.IO.Pipelines.Tests.ReadableBufferReaderFacts+SegmentPerByte.CopyToLargerBufferWorks [FAIL]
but I don't think I broke that. Flakey test?

ahsonkhan:
> but I don't think I broke that. Flakey test?

Yes, it is similar to https://github.com/dotnet/corefxlab/issues/2062

TimLovellSmith:
@safern, @ianhays, @KrzysztofCwalina, @ebickle, @nietras 
/cc @pgolebiowski @derekbjohnson @jnm2 @karelz @svick @benaadams 

I've updated my proposal to modify the priority queue proposal to where I feel happy with it again, so please take a look.

My thinking circled back to being two classes very similar in spirit to the current proposal at https://github.com/dotnet/corefx/issues/574 (`PriorityQueue<T>` and `PriorityQueue<TElement, TPriority>`).

The main differences are:
-no IQueue<T> interface
-update priority operation decision should now be clear: it is definitely supported by PriorityDictionary, and definitely not supported by PriorityQueue, which resolves some open issues
-yes ICollection<T> interface for PriorityQueue<T>
-adds IDictionary<TItem,TPriority> interface for PriorityDictionary
-naming of various things, noticeably the PriorityDictionary name. Naming feedback is appreciated! (We could go with 'TElement', 'TItem', or 'TKey'. TKey would helpfully suggest they are unique values, is that goodness? I'll change it to TKey instead of TItem...)
-PriorityDictionary can also allow specifying a custom EqualityComparer for flexibility

And re the other open issues:
-I don't think TryRemove is needed, Remove already returns a bool
-I think the proposed 'PriorityQueue' has such simple characteristics, it doesn't justify an interface to allow competing implementations
-PriorityDictionary could potentially be standardized to an interface (IHeap?) allowing for different implementations. And then people would be able to write other algorithms (shortest path, all pairs shortest path, etc.) in terms of the interface, and swap out the best known implementation for something else that works better for their one-off scenario. But I haven't convinced myself yet that people will really want/benefit from that kind of flexibility. Do you think its worth it?
-I don't think item 'handles' are needed any more

docs/specs/priority-queue.md
```diff
@@ -1,281 +1,194 @@
...
+    public PriorityDictionary(IEnumerable<TKey, TPriority> dictionary);
+    public PriorityDictionary(IEnumerable<TKey, TPriority> dictionary, IComparer<TPriority> priorityComparer);
+    public PriorityDictionary(IEnumerable<TKey, TPriority> dictionary, IComparer<TPriority> priorityComparer, IEqualityComparer<TKey> equalityComparer);
 
-```csharp
-var queue = new PriorityQueue<MyClass>();
-```
+    // new
+    public IComparer<TPriority> PriorityComparer { get; }
+    public IComparer<TPriority> EqualityComparer { get; }
```
safern: This should return `IEqualityComparer<TKey>`

docs/specs/priority-queue.md
```diff
@@ -1,281 +1,194 @@
...
+    public void Add(KeyValuePair<TKey, TPriority> item);
+    public bool Contains(KeyValuePair<TKey, TPriority> item);
+    public void CopyTo(KeyValuePair<TKey key, TPriority>[] array, int arrayIndex);
+    public IEnumerator<KeyValuePair<TKey key, TPriority priority>> GetEnumerator();
+    public bool Remove(KeyValuePair<TKey, TPriority> item);
+
+    // new
+    public TKey Dequeue();  // throws InvalidOperationException if the queue is empty (like Queue<>)
+    public void Enqueue(TKey key, TPriority priority); // an alias for IDictionary<TKey,TPriority>.Add()
+    public TKey Peek(TKey key); // throws InvalidOperationException if the queue is empty (like Queue<>)
```
safern: `Peek` should be parameter less since it returns the object at the beginning of the Queue. Also I think it should return `KeyValuePair<TKey, TPriority>`

docs/specs/priority-queue.md
```diff
@@ -1,281 +1,194 @@
...
+
+    // inherited from ICollection<KeyValuePair<TKey, TPriority>>
+    public void Add(KeyValuePair<TKey, TPriority> item);
+    public bool Contains(KeyValuePair<TKey, TPriority> item);
+    public void CopyTo(KeyValuePair<TKey key, TPriority>[] array, int arrayIndex);
+    public IEnumerator<KeyValuePair<TKey key, TPriority priority>> GetEnumerator();
+    public bool Remove(KeyValuePair<TKey, TPriority> item);
+
+    // new
+    public TKey Dequeue();  // throws InvalidOperationException if the queue is empty (like Queue<>)
```
safern: Should return `KeyValuePair<TKey, TPriority>`

docs/specs/priority-queue.md
```diff
@@ -1,281 +1,194 @@
...
+    public void Enqueue(TKey key, TPriority priority); // an alias for IDictionary<TKey,TPriority>.Add()
+    public TKey Peek(TKey key); // throws InvalidOperationException if the queue is empty (like Queue<>)
+    public bool TryDequeue(out TKey key);
```
safern: Instead of having two overrides for TryDequeue should we just have one that returns a `KeyValuePair`?

TimLovellSmith: Sure, or thinking of C# 7 and out vars + ignored out parameters, we could choose to go the two out-param route... I'm not sure which is ideally preferred, but personally I think we want this to be a bool return type, so maybe the two out parameters route is best?

safern: Yes I think the two out parameters would be fine as well. As with that route people will have access to both the key and priority after calling the method, without having to access properties inside the KeyValuePair object.

docs/specs/priority-queue.md
```diff
@@ -153,12 +153,10 @@ public class PriorityDictionary<TKey, TPriority> :
     public bool Remove(KeyValuePair<TKey, TPriority> item);
 
     // new
-    public TKey Dequeue();  // throws InvalidOperationException if the queue is empty (like Queue<>)
+    public KeyValuePair<TKey, TPriority> Dequeue();  // throws InvalidOperationException if the queue is empty (like Queue<>)
     public void Enqueue(TKey key, TPriority priority); // an alias for IDictionary<TKey,TPriority>.Add()
-    public TKey Peek(TKey key); // throws InvalidOperationException if the queue is empty (like Queue<>)
-    public bool TryDequeue(out TKey key);
+    public KeyValuePair<TKey, TPriority> Peek(TKey key); // throws InvalidOperationException if the queue is empty (like Queue<>)
```
safern: Peek needs to be parameterless :)

pgolebiowski:
could we have a short summary of this updated proposal and rationale behind it?

- what are the changes introduced?
- why are they being introduced? how is it better than the original one?

please be data-driven on these answers.

TimLovellSmith:
@pgolebiowski Hi, sure, would be happy to explain.

I came along one day and saw .Net *still* doesn't have a priority queue, and stumbling on this I figured one reason that nobody has implemented the design as is, is that it is leaving too many open issues undecided. So my primary goal as making these changes was to definitively resolve all these open issues, one way or another, so that someone can actually start implementing!

The open issues:
-Open Issue - Do we need IQueue?
-Open Issue - Do we need Remove and TryRemove?
-Open Issue - Do we want to be able to remove and update priorities of elements in O(log n)? 
-Open Issue - Do we need some sort of a handle to items, replicating the problems of the Java implementation? (to update priorities of elements in O(log n))
-Open Issue - Do we need an IPriorityQueue interface?

Additionally, I noticed something that seemed funny in the provided interfaces:
-It only implements IReadOnlyCollection<T> and not ICollection<T> why? Surely this is a [specially ordered!] collection of items and you can do all the operations that you can do on a collection of items (Add, Count, Clear, Contains, Enumerate, Remove)

Additionally, I felt the existing specification was unclear upon whether the priorities of items could change over time, and the specification needed to be very clear upon that point,
because this is a significant implementation concern. As I thought about the problem I decided that this was really a fundamental part of the design when it came to decomposing the design space into cases, and that really the description in terms of '4' scenarios to support was confusing and could be replaced with a description in terms of mutability.

Additionally, due to the related discussions, I felt the specification needed to be explicit on whether supporting duplicate priorities is a design goal.

Andbeing explicit about whether a single item could have multiple priorities seemed like it might be a good idea.

Finally, as part of the solution, I had a realization which became a slightly opinionated design choice: PriorityDictionary<TItem,TPriority> is the best way to describe items with priorities that can change over time, because it is essentially a collection of object keys and [modifiable] value pairs, and we all know a collection of key value pairs is a dictionary. And so it can potentially be used as a drop in replacement in code for other dictionaries, but with the enhancement of 'gaining' priority ordering feature almost for free when we implement the IDictionary<TKey,TValue> interface.

So, to summarize the main changes I made they are:
-Decision: IQueue is no longer part of the proposal (Although I am not strongly opinionated about it being a good or bad idea, conceptually it seems ok, but practically does anyone really want it? Does anyone want to write algorithms over an abstract queue without knowing whether it is FIFO or priority queue?)
-Decision: Remove is supported since it is an ICollection
-Decision: O(log n) is definitely a goal (hopefully I remembered to say that somewhere?)
-Design change outcome: handles are definitely not required
-Decision: IPriorityQueue might be nice to have, but it also is not part of the proposal because I am not sure there is real demand for it.
-Design change: implement ICollection<T>
-Specification change: PriorityQueue<T> is explicitly only usable for types which have immutable priority [comparison] behavior
-Specification clarification: duplicate priority values are explicitly allowed (but will just be dequeued in 'some' order, I didn't specify any order...)
-Naming change: PriorityQueue<TKey,TValue> is named PriorityDictionary<TKey,TValue>
-Added feature: PriorityDictionary<TKey,TValue> can implement IDictionary.
-Cut/clarification: a single distinct item (according to the equality comparer) cannot be in the dictionary multiple times with different priorities.
-And also as part of the dictionary change, the PriorityDictionary concept needed to explicitly have an IEqualityComparer on TKey.

Bottom line - I hope the new design or whatever we can agree on will be positive progress on achieving a spec that can be implemented - and then used by everyone.

danmosemsft:
Note to self 

https://github.com/Microsoft/Trill/blob/master/Sources/Core/Microsoft.StreamProcessing/Collections/PriorityQueue.cs

ahsonkhan:
What's our plan for this PR?

pgolebiowski:
https://github.com/dotnet/corefx/issues/35380

docs/specs/priority-queue.md
TimLovellSmith: Revised a couple times, hopefully its clear now.

feliwir:
What's missing to get the PriorityQueue merged?

PR #1851, opened 2017-10-17
Add Base64 encoder with ParsedFormat and isFinalBlock bool

ahsonkhan:
Related to https://github.com/dotnet/corefx/issues/24568

The new BytesToUtf8 API that takes in a ParsedFormat was added as a proof of concept to showcase that an encoder with line break capabilities can be built on top of the existing API.

cc @dotnet/corefxlab-contrib

### DIFF
--- a/src/System.Binary.Base64/System/Binary/Base64Encoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Encoder.cs
@@ -37,6 +37,7 @@ namespace System.Binary.Base64
                 sourceIndex += 3;
             }
 
+            // TODO: Implement logic for when isFinalBlock is false
             if (isFinalBlock != true) throw new NotImplementedException();
 
             if (sourceIndex == srcLength - 1)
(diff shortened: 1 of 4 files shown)


src/System.Binary.Base64/System/Binary/Base64Encoder.cs
```diff
@@ -61,10 +67,101 @@ public static OperationStatus BytesToUtf8(ReadOnlySpan<byte> bytes, Span<byte> u
...
+                utf8[lineLength] = (byte)'\r';
+                utf8[lineLength + 1] = (byte)'\n';
+                bytesWrittenLoop += 2;
+                bytes = bytes.Slice(bytesConsumedLoop);
+                utf8 = utf8.Slice(bytesWrittenLoop);
+                consumed += bytesConsumedLoop;
+                written += bytesWrittenLoop;
+            }
+
+            if (isFinalBlock)
```
KrzysztofCwalina: Why do we need isFinalBlock in Base64 encoding? I understand that some encodings might need it (and so our abstractions might need it), but the static method is specific  to Base64, which does not need to ever know that something is a final block.

bartonjs: The last segment of Base64 data doesn't need to be 3 bytes, and padding gets added to indicate how long it was (so that the output is always a multiple of 4 characters)

ahsonkhan: > Why do we need isFinalBlock in Base64 encoding?

The isFinalBlock is required to know when to add padding (for input length that is not a multiple of 3). Currently, we always add padding which is incorrect when partial input is passed in.

ahsonkhan:
Is this change good to go?

PR #1854, opened 2017-10-17
Tensors should only depend on System.Memory (not Buffers.Primitives)

ericstj:
We don't actually use Buffers.Primitives now that Memory (previously Buffer) has been moved into System.Memory.


ahsonkhan:

Issue #1821, opened 2017-10-05
Add interop samples to System.Numerics.Tensors


ericstj:
We should add some samples that demonstrate interop use cases for Tensors.

I haven't done this yet because I haven't looked in to how to build a native DLL in corefxlab.  We could just fake it and write some managed code that behaves like native (using pointers calling VirtualAlloc and what not).

The scenarios we should demonstrate are as follows:
1. Pass managed heap allocated tensor to native code, various layouts and strides.
2. Have native code fill in a managed heap-allocated tensor of known size that is passed in.
3. Have native code allocate its own memory, fill it, pass it back to managed code, and wrap it in a tensor. This should be similar to these tests: https://github.com/dotnet/corefxlab/commit/0e5bbc3e4ad38185fa7e6e2f556b6e17499707ba#diff-f769120a2575d4bfd7f070f8adfff09aR258, https://github.com/dotnet/corefxlab/commit/0e5bbc3e4ad38185fa7e6e2f556b6e17499707ba#diff-f769120a2575d4bfd7f070f8adfff09aR364
4. Have native code call back into managed code to allocate memory, fill it in native, pass it back to managed and wrap it in a tensor.  I have this sample here: https://github.com/ericstj/scratch/commit/d1f18a23982e2aa81f55ff0581d32190c6d52b3b.

I'd like to set this up as a set of test cases so that we can keep the sample code running and make sure it continues to work.

PR #1857, opened 2017-10-17
Adding interop samples for Tensor

ericstj:
Add some samples that illustrate how to interop with Tensor.

These are only part of the Tensor sln at the moment because I don't want the official build to take a dependency on having VC++ installed (nor can I for x-plat builds).

#1821

### DIFF
--- a/samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
+++ b/samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
@@ -29,7 +29,7 @@ namespace TensorSamples
         }
         
         // The following represent what .NET bindings would look like for a native library that wishes
-        // to deal take a sparse tensor and mutate its values        
+        // to deal take a sparse tensor and mutate its values     
 
         public static unsafe double ScalarPowerSparse(CompressedSparseTensor<double> tensor, int exponent)
         {
--- a/samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
+++ b/samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
@@ -1,0 +1,47 @@
+﻿using System;
+using System.Numerics;
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace TensorSamples
+{
+    class MutateSparse
+    {
+        public static void RunSample()
+        {
+            Console.WriteLine($"*** {nameof(MutateSparse)} ***");
+            CompressedSparseTensor<double> tensor = new CompressedSparseTensor<double>(new[] { 3, 5, 4 }, 10);
+
+            tensor[0, 2, 0] = 1.6;
+            tensor[1, 0, 3] = 5.3;
+            tensor[1, 1, 0] = Math.PI;
+            tensor[2, 3, 3] = 867.5309;
+            tensor[2, 4, 0] = 19.4231;
+            tensor[2, 4, 3] = 10000.1;
+
+            Console.WriteLine("Original tensor:");
+            Console.WriteLine(tensor.GetArrayString());
+
+            ScalarPowerSparse(tensor, 3);
+
+            Console.WriteLine("Cubed tensor:");
+            Console.WriteLine(tensor.GetArrayString());
+        }
+        
+        // The following represent what .NET bindings would look like for a native library that wishes
+        // to deal take a sparse tensor and mutate its values        
+
+        public static unsafe double ScalarPowerSparse(CompressedSparseTensor<double> tensor, int exponent)
+        {
+            fixed (double* valuesPtr = &tensor.Values.Span.DangerousGetPinnableReference())
+            fixed (int* compressedCountsPtr = &tensor.CompressedCounts.Span.DangerousGetPinnableReference())
+            fixed (int* indicesPtr = &tensor.Indices.Span.DangerousGetPinnableReference())
+            {
+                return ScalarPowerSparse(valuesPtr, compressedCountsPtr, indicesPtr, tensor.Dimensions.ToArray(), tensor.Rank, tensor.NonZeroCount, exponent);
+            }
+        }
+
+        [DllImport("TensorSamples.native.dll")]
+        extern private static unsafe double ScalarPowerSparse(double* values, int* compressedCounts, int* indices, int[] dimensions, int rank, int valueCount, int power);
+    }
+}
--- a/samples/System.Numerics.Tensors.Samples/TensorSamples/PreAllocate.cs
+++ b/samples/System.Numerics.Tensors.Samples/TensorSamples/PreAllocate.cs
@@ -5,11 +5,11 @@ using System.Text;
 
 namespace TensorSamples
 {
-    partial class WrapNativeMemory
+    partial class PreAllocate
     {
         public static void RunSample()
         {
-            Console.WriteLine($"*** {nameof(WrapNativeMemory)} ***");
+            Console.WriteLine($"*** {nameof(PreAllocate)} ***");
             var multTable = GetMultiplicationTable(5);
 
             Console.WriteLine("Multiplication table:");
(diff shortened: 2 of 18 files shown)


samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
```diff
@@ -0,0 +1,47 @@
...
+            Console.WriteLine("Cubed tensor:");
+            Console.WriteLine(tensor.GetArrayString());
+        }
+        
+        // The following represent what .NET bindings would look like for an native library that wishes
+        // to deal take a sparse tensor and mutate its values        
+
+        public static unsafe double ScalarPowerSparse(CompressedSparseTensor<double> tensor, int exponent)
+        {
+            fixed (double* valuesPtr = &tensor.Values.Span.DangerousGetPinnableReference())
```
ericstj: These feel a bit clunky, not to mention the lack of intellisense.  Is this really the best way to interop with span?

samples/System.Numerics.Tensors.Samples/TensorSamples/ReversePInvoke.cs
```diff
@@ -0,0 +1,58 @@
...
+        {
+            fixed(double* dataPtr = &tensor.Buffer.Span.DangerousGetPinnableReference())
+            {
+                return GetRowSum(dataPtr, tensor.Dimensions.ToArray(), tensor.Rank, row);
+            }
+        }
+
+
+        delegate IntPtr AllocatorDelegate(int size);
+        [DllImport("native.dll")]
```
eerhardt: (nit) I know this is just a sample, but naming this `native.dll`  doesn't seem obvious to me.  Can it be renamed to `TensorSamples.native`?

eerhardt:
There's something wrong with all the C++ code files.  GitHub thinks they are binary files.

samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
```diff
@@ -0,0 +1,47 @@
...
+            Console.WriteLine("Original tensor:");
+            Console.WriteLine(tensor.GetArrayString());
+
+            ScalarPowerSparse(tensor, 3);
+
+            Console.WriteLine("Cubed tensor:");
+            Console.WriteLine(tensor.GetArrayString());
+        }
+        
+        // The following represent what .NET bindings would look like for an native library that wishes
```
eerhardt: (type-o)  `an native library`

eerhardt:
> extern "C"  __declspec(dllexport) void __stdcall ScalarPowerSparse(double* values, int* compressedCounts, int* indices, int* dimensions, int rank, int valueCount, int power)

Some of these parameters aren't used.  Can this be similified?

---
Refers to: samples/System.Numerics.Tensors.Samples/native/native.cpp:93 in 4cc1fc1. [](commit_id = 4cc1fc17e89af3f9ba0f03bb5131c04d88623327, deletion_comment = False)

ericstj:
> Some of these parameters aren't used. Can this be simplified?

I want to show how to pass all the relevant data for reading the sparse array.  Perhaps I'll come up with a better native-side method to illustrate using everything, but barring that they are providing value since in a real solution they will all be needed.

ericstj:
> There's something wrong with all the C++ code files. GitHub thinks they are binary files.

Perhaps because they are UTF-16 (apparently the default for the C++ templates)  I'll change the encoding and see if that helps.

samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
ericstj: /cc @KrzysztofCwalina @ahsonkhan @stephentoub

ericstj:
This now covers all the scenarios I listed in issue.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.cs
```diff
@@ -0,0 +1,60 @@
...
+using System.Text;
+
+namespace TensorSamples
+{
+    partial class WrapNativeMemory
+    {
+        public static void RunSample()
+        {
+            Console.WriteLine($"*** {nameof(WrapNativeMemory)} ***");
+            var multTable = GetMultiplicationTable(5);
```
KrzysztofCwalina: I think it would be better (idiomatice sample) to allocate NamtiveMemory explicitly, then wrap it in tensor, use the tensor, and then dispose the memory, as opposed to relying on the GC to finalize the memory, e.g.:
```c#
using(var memory = AllocateNative(...)){
    var tensor = new DenseTensor<double>(...);
    ...
}
```

ericstj: The point of this sample is that its supposed to represent the case where the .net app does not know what size will be needed for the native memory and the cost of computing that is equivalent to doing the work, so it cannot be allocated by the caller.  Granted in the case in the sample is pretty trivial, but imagine a case where the tensor might be constructed as a result of analyzing a block of text or some other non-trivial bit of quantification.

I agree that ideally folks would dispose of the memory explicitly, but we also don't want to force an application to treat a tensor as disposable.  In these cases the managed footprint of the tensor itself will be small so I wouldn't expect it to get on LOH.

I do have a better sample in the test code that allocates the native memory itself and then disposes it.  I could add another scenario for that.

samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
KrzysztofCwalina: @VSadov, long term we will get C# fixed statement to generate this code.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.cs
ericstj: I added the scenario for a pre-allocated tensor wrapping native memory.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.cs
```diff
@@ -0,0 +1,89 @@
...
+
+                GetMultiplicationTablePreallocated(5, tensor);
+
+                Console.WriteLine("Sums:");
+                for (int row = 0; row < tensor.Dimensions[0]; row++)
+                {
+                    Console.WriteLine(GetRowSum(tensor, row));
+                }
+            }
+        }
```
KrzysztofCwalina: nice!

ericstj:
FYI: @sethjuarez @liqunfu

samples/System.Numerics.Tensors.Samples/TensorSamples/MutateSparse.cs
VSadov: Will try enabling `fixed(double* valuesPtr = tensor.Values.Span)` in 7.3

Note that you can also do `fixed(double* valuesPtr = &tensor.Values.Span[0])`

ahsonkhan: > Note that you can also do `fixed(double* valuesPtr = &tensor.Values.Span[0])` 

If the span is empty, we can't do that.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.cs
```diff
@@ -0,0 +1,89 @@
...
+            GC.Collect();
+            GC.WaitForPendingFinalizers();
+
+            Console.WriteLine($"Allocated prior to the operation");
+            // in this case we'll assume we can know the size of the buffer up front and manage the lifetime explicitly
+            using (var nativeMemory = NativeMemory<double>.Allocate(5 * 5))
+            {
+                Span<int> dimensions = stackalloc int[2];
+                dimensions[0] = dimensions[1] = 5;
+                var tensor = new DenseTensor<double>(nativeMemory.Memory, dimensions);
```
nietras: Since `DenseTensor` is a reference type, this mean an allocation for every time native memory is "wrapped"... have value type `TensorView`s (ala `Span`) been considered to avoid this?

samples/System.Numerics.Tensors.Samples/TensorSamples/PreAllocate.cs
```diff
@@ -0,0 +1,58 @@
...
+            }
+
+            return result;
+        }
+
+        public static unsafe double GetRowSum(DenseTensor<double> tensor, int row)
+        {
+            fixed(double* dataPtr = &tensor.Buffer.Span.DangerousGetPinnableReference())
+            {
+                return GetRowSum(dataPtr, tensor.Dimensions.ToArray(), tensor.Rank, row);
```
nietras: Why `tensor.Dimensions.ToArray()` and not fix the `Dimensions` span? No need to alloc. `GetRowSum` should be changed accordingly.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.cs
```diff
@@ -0,0 +1,89 @@
...
+            var nativeMemory = new NativeMemory<double>(GetMultTableAllocateNative(maxNumber), maxNumber * maxNumber);
+            return new DenseTensor<double>(nativeMemory.Memory, dimensions);
+            
+        }
+
+        public static unsafe double GetRowSum(DenseTensor<double> tensor, int row)
+        {
+            fixed(double* dataPtr = &tensor.Buffer.Span.DangerousGetPinnableReference())
+            {
+                return GetRowSum(dataPtr, tensor.Dimensions.ToArray(), tensor.Rank, row);
```
nietras: Same here. Why `tensor.Dimensions.ToArray()` and not fix the `Dimensions` span? No need to `alloc. GetRowSum` should be changed accordingly.

samples/System.Numerics.Tensors.Samples/TensorSamples/ReversePInvoke.cs
```diff
@@ -0,0 +1,58 @@
...
+
+                return result;
+            }
+        }
+
+        public static unsafe double GetRowSum(DenseTensor<double> tensor, int row)
+        {
+            fixed(double* dataPtr = &tensor.Buffer.Span.DangerousGetPinnableReference())
+            {
+                return GetRowSum(dataPtr, tensor.Dimensions.ToArray(), tensor.Rank, row);
```
nietras: Same here. Why `tensor.Dimensions.ToArray()` and not fix the `Dimensions` span? No need to `alloc. GetRowSum` should be changed accordingly.

PR #1858, opened 2017-10-18
Preparing APIs for Span ReaderWriter

KrzysztofCwalina:
I am working on a prototype of self-advancing reader/writer APIs. 

This is just a preparatory work:
1. Adds IWritable interface similar to IBufferFormattable that allows custom types to write themselves to buffers as binary data.
2. Sha256 is an example of such self-writable type.
3. Changed Sha256 so it can write to buffers of equal or larger size. Before this PR the buffer had to be exact size.

CC: @ahsonkhan, @GrabYourPitchforks, @joshfree


joshfree:
👍

Drawaes:
I presume the reader/writer takes this interface as a generic constraint to avoid boxing of structs?

KrzysztofCwalina:
@Drawaes, yes. It will be something similar to https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Formatting/System/Text/Formatting/ITextOutputExtensions.cs#L11

Issue #1306, opened 2017-03-11
Slicing pattern with range


ahsonkhan:
@davidfowl 
Does slicing using a range refer to code like this (using '..')?
```C#
Span<byte> buffer = ...

Span<byte> view1 = buffer.Slice(3, 2);
Span<byte> view2 = buffer[3..4];
 
// view1 == view2
```

KrzysztofCwalina:
@ahsonkhan, do you still want to keep this open?

ahsonkhan:
That depends on whether we want to support range syntax or not.
This was from the initial discussion/backlog of span items where we were considering slicing using range. I presume that is what the syntax would look like.

Changing to future.

ahsonkhan:
@KrzysztofCwalina, if we are getting the Range syntax support by 2.1, we should consider adding support for it for Span. Should we change the milestone of this issue to 2.1? Let me know and I will move it over to corefx.

KrzysztofCwalina:
Yeah, once we settle on the design for Range, we can add APIs. I would not add them sooner.

PR #1859, opened 2017-10-18
Generic implementation of Range.

YohDeadfall:
At this moment the `Range` class isn't generic and stores `int Index` and `int Length` values, but probably it would be better to have a generic struct and move it to the `System.Collections.Sequences` library.

The main idea is that `Range<T>` could be used not only by buffers, but also in third party code to store a range of values (as inclusive `T From` and exclusive `T To` values), to manipulate collections and such related things.

Why the `To` field instead of `Length`? Because the `To` field can store floating point values without losing precision, and length can not be determined for some types like `string`. But length calculation can be made using extension methods for numeric types. Since we have the aggressively inlined and specialized `Length(this Range<Int>)` method, that change will not hurt performance.


KrzysztofCwalina:
This complicates the type significantly.  For example, the Slice method signature is currently Slice(Range range), but now it would be Slice(Range<int> range). Similarly on the caller side, it would complicate the callsite.

@jaredpar, I originally added the Range type to play with the idea we talked about where C# slice syntax would take a range type. Do you have any opinions on whether Range or Range<T> is closer to what C# could (would like to) use?

KrzysztofCwalina:
@benaadams, I keep forgetting the ticks. Updated now.

YohDeadfall:
@KrzysztofCwalina, not everywhere. You forgot about the second `Slice(Range range)`.

Drawaes:
I can confirm that reading that in email was confusing... range range

YohDeadfall:
@KrzysztofCwalina, could you provide a link to the discussion with @jaredpar? Is [it](https://github.com/dotnet/roslyn/issues/120)?

YohDeadfall:
Should the fields of the `Range<T>` struct be readonly or should they be replaced by readonly properties? Do we need implementation of the `ISerializable` and `IXmlSerializable` interfaces?

KrzysztofCwalina:
@YohDeadfall is was more of a hallway discussion.

YohDeadfall:
I can close this pull request and file an issue in the corefx repository to invite more people to discuss, but before that it would be good to know what is the slice syntax.

mj1856:
One thing - in order to be used as a range type, at some point you'll need to compare the extents of the range with other values - so shouldn't `T` be constrained to `IComparable<T>` ?

grant-d:
I am not sure the logic here is correct:
```C#
    public static bool IsEmpty(this Range<<#= type.Name #>> range)
    {
        return range.From == range.To;
    }
```
If I declare a range that, say, should only regard the 5th element:
```C#
    var myRange = new Range(5, 5);
```

Then that range is not `empty` per se, but perhaps `exact`.

Perhaps you could consider logic as follows?
```C#
    public static bool IsEmpty(this Range<<#= type.Name #>> range)
    {
        return range.From == range.To 
            && range.From.Equals(default(T)); // Or EqualityComparer.Default, etc
    }
```

In the past I have needed such a class myself and chose to support this semantic directly by having a `ctor` (or alternatively a factory) as follows:
```C#
    class Range
    {
        ...
        public Range(T exact) : this(exact, exact)
        { }
    }
```

src/System.Collections.Sequences/System/Range.cs
```diff
@@ -0,0 +1,71 @@
...
+        }
+
+        public void Deconstruct(out T from, out T to)
+        {
+            from = From;
+            to = To;
+        }
+
+        public override bool Equals(object obj) =>
+            obj is Range<T> && Equals((Range<T>)obj);
```
grant-d: Depending on your compiler version:
```C#
obj is Range<T> other && Equals(other); 
```

grant-d:
I posted (and deleted) some feedback based on my wrong assumption that the upper bound was inclusive. 
I am curious why this design was chosen. 
I understand `for` loops & their ilk are `[n, m)` boundaries but I wonder whether users would expect `ranges` to behave the same. In other words, my own (perhaps unique) expectation is:
```C#
//              Expected                         Actual (current design)
Range(0, 0); // ~equivalent to Empty             ~equivalent to Empty
Range(5, 5); // ~equivalent to Slice(5, 1)       ~equivalent to Empty (also)
Range(5, 6); // ~equivalent to Slice(5, 2)       ~equivalent to Slice(5, 1)
```

YohDeadfall:
@mj1856 Thanks. Don't know why I thought that the `string` class isn't `IComparable<T>`. Will fix it.

@grant-d They are already adapted for slices because exclusive ending ranges have length which equals to `to - from` (`end - start`) and slices get length as the second parameter. Also fully inclusive ranges cannot be empty, but exclusive ending ranges can therefore I chose them.
```csharp
public readonly ref struct Span<T>
{
    public Span<T> Slice(int start, int length) { throw null; }
}
public readonly struct ReadOnlyBytes
{
    public ReadWriteBytes Slice(Range<int> range)
    {
        return Slice(range.From, range.Length());
    }

    public ReadWriteBytes Slice(int index, int length)
    {
```
|       Range | Length |       Slice |                                               Description |
|-------------|--------|-------------|-----------------------------------------------------------|
| Range(0, 0) |       0| Slice(0, 0) | Take no values.                                           |
| Range(1, 1) |       0| Slice(1, 0) | Take no values.                                           |
| Range(1, 2) |       1| Slice(1, 1) | Take the first value starting 1, i.e. starting 1 until 2. |
| Range(1, 5) |       1| Slice(1, 4) | Take four values starting 1, i.e. starting 1 until 5.     |

An inclusive end makes sense because it allows to include the max value of `T`. So yes, ranges should be with exclusive ending and with inclusive ones as you suggested, but I don't think that everyone needs inclusive ranges. Adding a field to specify the type of the ending will increase the size of `Range<T>` and decrease performance. That's bad. Maybe developers should specify that in their API's by naming? For example:
```csharp
void DoSomething(Range<int> inclusiveRange, Range<int> exclusiveRange) { }
```
From our side we can provide additional extension methods:
```csharp
public static bool Contains<T>(this Range<T> range, T value); // exclusive by default *
public static bool ContainsInclusive<T>(this Range<T> range, T value); // inclusive by request
// or
public static bool Contains<T>(this Range<T> range, T value, bool inclusive = false);

public static double Length(this Range<double> range); // exclusive by default *
public static double LengthInclusive(this Range<double> range); // inclusive by request
// or
public static bool Contains<T>(this Range<T> range, bool inclusive = false);

// * in current implementation
```
From that point we have the two types of ranges with zero cost. With aggressive inlining the compiler will eliminate `inclusive` checks.

YohDeadfall:
The C# slice syntax was discussed in #1306.

YohDeadfall:
@mj1856 I remember why `T` isn't constrained to `IComparable<T>`. It's because `Nullable<T>` has no interfaces. Should the constraint be added anyway? Do we need to support nullable?

mj1856:
Just to chime in on the subject of the exclusive upper bound, that's really important if they are to be used generically, such as `Range<DateTime>`.  Consider two meetings back-to-back.  The first goes from 1:00 to 2:00.  The second from 2:00 to 3:00.  Which meeting is going on exactly at 2:00?  Most people would generally say the second one, because the first meeting has ended.  In other words, `[1:00, 2:00) [2:00, 3:00)` is the only rational way to think about ranges of time.

IMHO this should apply uniformly to all types.

mj1856:
WRT `Nullable<T>`,  I assume you mean for open-ended ranges like `new Range<int?>(null, 10)` or `new Range<int?>(10, null)` - but IMHO I don't think you need the `T` itself to be nullable to achieve that goal.  Couldn't you just do that on the constructor? As in:

```csharp
class Range<T>
    where T: struct, IComparable<T>
{
    public Range(T? from, T? to) { ... }
}
```

YohDeadfall:
# Design overview
## Type constraints

There are no constraints at all for two reasons:
1. To support reference types and value types.
2. To support `Nullable<T>` since it has no interfaces.

```csharp
var range1 = new Range<long>(0, 100);
var range2 = new Range<long?>(null, 100);
var range3 = new Range<string>("a", "z");
```

## Methods
I've implemented `IsEmpty`, `IsNormalized`, etc. as extension methods because:
1. This way allows to write specialized methods even for non-BCL types and optimize performance.
```csharp
public readonly struct TinyTime
{
    private readonly byte _ticks;
    // properties, operators and so on
}
public static class RangeExtensions
{
    // everything with MethodImplOptions.AggressiveInlining
    public static bool IsEmpty(this Range<TinyTime> range) => range.From == range.To;
    public static bool IsNormalized(this Range<TinyTime> range) => range.From <= range.To;
    // ...
}
```
2. It's possible to provide a custom comparer to the methods. For example it's useful for strings.
```csharp
var range = new Range<string>("A", "a");
var empty1 = range.IsEmpty(StringComparer.InvariantCulture); // false
var empty2 = range.IsEmpty(StringComparer.InvariantCultureIgnoreCase); // true
```

---

@mj1856 The `Range<T>` type could store the endings as `T?` but that will increase the size of `Range<T>`.

```csharp
// Marshal.SizeOf(new Range<int>()) == 16
struct Range<T> where T: struct, IComparable<T>
{
    public T? From;
    public T? To;
}
```
```csharp
// Marshal.SizeOf(new Range<int>()) == 8
struct Range<T> where T: IComparable<T>
{
    public T From;
    public T To;
}
```

jaredpar:
The language design at this time does not intend to support a generic Range type. Rather we are focusing on concrete types and operator extensibility to support additional ranges. 

Do not think we should take this change as it's going the opposite direction as the language / compiler.

YohDeadfall:
@jaredpar  Okay, then I'll close this pull request. But could you post your response to https://github.com/dotnet/roslyn/issues/23205 because not only I think about generics?

PR #1736, opened 2017-08-30
Change test projects to target netcoreapp2.1

ahsonkhan:
After validating that tests targeting netcoreapp2.0 will pass using the latest packages (https://github.com/dotnet/corefxlab/pull/1735), update the tests to target netcoreapp2.1.

**Do NOT merge until the above mentioned PR is green.**

We could update to 2.1 right away, but we should wait and make sure that the packages we depend on still work on netcoreapp2.0 first.

cc @shiftylogic, @KrzysztofCwalina

PR #1861, opened 2017-10-18
Change sample projects part of the solution to target netcoreapp2.1

ahsonkhan:
**Edit:**
Resolving based on https://github.com/dotnet/corefx/issues/24910#issuecomment-339832002
> The fix here is to do what nuget points out and deal with the downgrade issue by upgrading your version of System.Numerics.Vectors package.

Updating S.N.V package version to sync with what System.Memory uses.

Fixes restore failure here caused by package downgrade error (NU1605): https://github.com/dotnet/corefxlab/pull/1853

From: https://ci.dot.net/job/dotnet_corefxlab/job/master/job/windows_nt_release_prtest/881/consoleFull#-1454977910a82fefab-f698-416f-8fca-58544c94cd4e
> D:\j\workspace\windows_nt_re---4abe806d\samples\EchoService\EchoService.csproj : error NU1605: Detected package downgrade: System.Numerics.Vectors from 4.5.0-preview1-25818-02 to 4.3.0. Reference the package directly from the project to select a different version.  [D:\j\workspace\windows_nt_re---4abe806d\corefxlab.sln]
06:47:42 D:\j\workspace\windows_nt_re---4abe806d\samples\EchoService\EchoService.csproj : error NU1605:  EchoService -> System.Net.Libuv -> System.Buffers.Experimental -> System.Collections.Sequences -> System.Memory 4.5.0-preview1-25818-02 -> System.Numerics.Vectors (>= 4.5.0-preview1-25818-02)  [D:\j\workspace\windows_nt_re---4abe806d\corefxlab.sln]
06:47:42 D:\j\workspace\windows_nt_re---4abe806d\samples\EchoService\EchoService.csproj : error NU1605:  EchoService -> System.Net.Libuv -> System.Buffers.Experimental -> System.Numerics.Vectors (>= 4.3.0) [D:\j\workspace\windows_nt_re---4abe806d\corefxlab.sln]

Alternatively, we could explicitly reference the specific version of System.Numerics.Vectors rather than 4.3.0 (and update it in lock step with System.Memory/etc).

cc @dotnet/corefxlab-contrib, @ericstj


ahsonkhan:
**Question @ericstj:** Why would this error surface now, after upgrading System.Memory from `4.5.0-preview1-25816-01` to `4.5.0-preview1-25818-02`? What changed?

ericstj:
Shouldn't you be updating System.Buffers.Experimental?  That is the project that is actually doing the downgrade based on the log above.
```
System.Buffers.Experimental -> System.Collections.Sequences -> System.Memory 4.5.0-preview1-25818-02 -> System.Numerics.Vectors (>= 4.5.0-preview1-25818-02)
System.Buffers.Experimental -> System.Numerics.Vectors (>= 4.3.0)
```

ericstj:
> Why would this error surface now, after upgrading System.Memory from 4.5.0-preview1-25816-01 to 4.5.0-preview1-25818-02? What changed?

@weshaggard fixed https://github.com/dotnet/corefx/issues/24606 with https://github.com/dotnet/corefx/pull/24645.  One thing to consider with System.Memory: shouldn't the netcoreapp2.0 implementation use the *inbox* version of System.Numerics.Vectors?  If it did so we wouldn't emit a package dependency.

ahsonkhan:
> Shouldn't you be updating System.Buffers.Experimental?

System.Buffers.Experimental is a netstandard library and it doesn't need a version above 4.3.0. You mentioned previously the potential of package downgrade here: https://github.com/dotnet/corefxlab/pull/1736#discussion_r142217772

I don't fully understand why an application targeting netcoreapp which indirectly depends on a higher version of System.Numerics (through its project reference) results in the netstandard library to update it's dependencies.

Why don't we get the similar package downgrade error when building just System.Buffers.Experimental? It has a project reference to System.Collections.Sequences -> S.M (4.5.xxx) -> S.N.V (4.5.xxx) along with a package reference S.N.V (4.3.0).

ericstj:
> Why don't we get the similar package downgrade error when building just System.Buffers.Experimental

Because it doesn't target the specific TFM which references the latest version of System.Numerics.Vectors.  Its only targeting an older TFM so it never sees the downgrade.

System.Memory is doing some unusual things by setting a lower baseline: https://github.com/dotnet/corefx/pull/14271.  This made sense when it was a simple PCL, but now that its cross-compiling and moving types into the runtime folks probably need to take a closer look.  It ends up using the lower baseline for the downlevel implementations, but then the latest package due to the direct project-references and/or live builds for those TFMs.

I'd suggest filing an issue in CoreFx saying that the project configurations and depend
ckage due to the direct project-references and/or live builds for those TFMs.

I'd suggest filing an issue in CoreFx saying that the project configurations and dependencies of System.Memory need to be carefully reviewed to ensure that they make sense create a compatible portable story for the package as folks switch between TFMs.

ahsonkhan:
> One thing to consider with System.Memory: shouldn't the netcoreapp2.0 implementation use the *inbox* version of System.Numerics.Vectors?  If it did so we wouldn't emit a package dependency.

How would I do that? Would that require a specific build configuation for netcoreapp2.0 (rather than just netcoreapp)?

ericstj:
> How would I do that? Would that require a specific System.Memory build configuation for netcoreapp2.0 (rather than just netcoreapp)?

You need to investigate how System.Memory is building for netcoreapp2.0 and where it is getting the System.Numerics.ValueTuple that it is building against.  Let's not deal with it here, file an issue to track an investigation on CoreFx side.

src/System.IO.Pipelines.Extensions/System.IO.Pipelines.Extensions.csproj
```diff
@@ -12,6 +12,6 @@
   </ItemGroup>
   <ItemGroup>
     <PackageReference Include="System.Threading.Tasks.Extensions" Version="$(CoreFxStableVersion)" />
-    <PackageReference Include="System.Numerics.Vectors" Version="$(CoreFxStableVersion)" />
+    <PackageReference Include="System.Numerics.Vectors" Version="$(SystemNumericsVectorsVersion)" />
```
KrzysztofCwalina: @davidfowl, is that ok? I

ahsonkhan: We were already referencing the latest System.Numerics.Vectors package prior to https://github.com/dotnet/corefxlab/pull/1736.

davidfowl: Hmm, not sure yet.

PR #1862, opened 2017-10-18
Added an abstraction for in-place transformations

KrzysztofCwalina:
The idea is that Writer.Write method can take in IBufferTransformation to additional transform the date being written.  For example, the following code writes Base64 encoded hash to a buffer:
```c#
Span<byte> buffer = stackalloc byte[256];
var writer = new Utf8SpanWriter(buffer);

Sha256 hash = Sha256.Create(key, data);
writer.Write(hash, Base64.Encoder); // the encoder implements IBufferTransformation
```

CC: @joshfree, @ahsonkhan, @GrabYourPitchforks

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/Transformation.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Transformation.cs
@@ -8,6 +8,10 @@ namespace System.Buffers
     {
         OperationStatus Execute(ReadOnlySpan<byte> input, Span<byte> output, out int consumed, out int written);
     }
+    public interface IBufferTransformation
+    {
+        OperationStatus Transform(Span<byte> buffer, int dataLength, out int written);
+    }
 
     public enum OperationStatus
     {
--- a/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Decoder.cs
@@ -228,13 +228,13 @@ namespace System.Binary.Base64
             }
         }
 
-        sealed class FromBase64Utf8 : BufferDecoder
+        sealed class FromBase64Utf8 : BufferDecoder, IBufferTransformation
         {
             public override OperationStatus Decode(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
                 => Utf8ToBytes(source, destination, out bytesConsumed, out bytesWritten);
 
-            public override OperationStatus DecodeInPlace(Span<byte> buffer, int inputLength, out int written)
-                => Utf8ToBytesInPlace(buffer.Slice(0, inputLength), out var consumed, out written);
+            public override OperationStatus Transform(Span<byte> buffer, int dataLength, out int written)
+                => Utf8ToBytesInPlace(buffer.Slice(0, dataLength), out var consumed, out written);
 
             public override bool IsDecodeInPlaceSupported => true;
         }
--- a/src/System.Binary.Base64/System/Binary/Base64Encoder.cs
+++ b/src/System.Binary.Base64/System/Binary/Base64Encoder.cs
@@ -201,13 +201,13 @@ namespace System.Binary.Base64
             return OperationStatus.DestinationTooSmall;
         }
 
-        sealed class ToBase64Utf8 : BufferEncoder
+        sealed class ToBase64Utf8 : BufferEncoder, IBufferTransformation
         {
             public override OperationStatus Encode(ReadOnlySpan<byte> source, Span<byte> destination, out int bytesConsumed, out int bytesWritten)
                 => BytesToUtf8(source, destination, out bytesConsumed, out bytesWritten);
 
-            public override OperationStatus EncodeInPlace(Span<byte> buffer, int inputLength, out int written)
-                => BytesToUtf8InPlace(buffer, inputLength, out written);
+            public override OperationStatus Transform(Span<byte> buffer, int dataLength, out int written)
+                => BytesToUtf8InPlace(buffer, dataLength, out written);
 
             public override bool IsEncodeInPlaceSupported => true;
         }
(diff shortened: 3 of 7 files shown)


src/System.Buffers.Primitives/System/Buffers/Transformation.cs
```diff
@@ -8,6 +8,10 @@ public interface IBufferOperation
     {
         OperationStatus Execute(ReadOnlySpan<byte> input, Span<byte> output, out int consumed, out int written);
     }
+    public interface IBufferTransformation
```
ahsonkhan: Why do we need a separate interface? Can we add the in-place transform to IBufferOperation?

GrabYourPitchforks: @ahsonkhan Not all transforms would support in-place. IMO it makes sense to have a separate interface for this capability. (Though you'd also probably want a `CanTransformInPlace` API just in case you're subclassing a base class that implements this interface and you need to suppress this capability.)

src/System.Buffers.Primitives/System/Buffers/Transformation.cs
```diff
@@ -8,6 +8,10 @@ public interface IBufferOperation
...
+    {
+        OperationStatus Transform(Span<byte> buffer, int dataLength, out int written);
```
ahsonkhan: As part of updating the Base64 APIs, I will change the in-place methods to return bool instead of OperationStatus. That won't break your scenario, correct?

src/System.Buffers.Primitives/System/Buffers/Transformation.cs
KrzysztofCwalina: Some transformations cannot be done in-place

GrabYourPitchforks:
LGTM, though I'd also suggest a `bool CanTransformInPlace { get; }` property. That's what I did in my own sample at https://github.com/GrabYourPitchforks/transforms/blob/master/Transforms/ITransformInPlace.cs, and it seems to work in the samples I've played with.

src/System.Buffers.Primitives/System/Buffers/Transformation.cs
KrzysztofCwalina: I still need to know if: a) the buffer was too small, b) the data in the buffer was invalid

KrzysztofCwalina:
@GrabYourPitchforks, when would the implementation of your interface return false from CanTransfrormInPlace. If it cannot, then it should not implement the interface.

GrabYourPitchforks:
@KrzysztofCwalina If you're subclassing a base class which has implemented the interface but your own specific implementation doesn't support in-place transforms, you'd want the ability to signal to the caller that you don't support it even though the interface exists in your type hierarchy.

src/System.Buffers.Primitives/System/Buffers/Transformation.cs
ahsonkhan: So, if the in place statics methods returned bool, would the Transform implementation look something like this:
```C#
public override OperationStatus Transform(Span<byte> buffer, int dataLength, out int written)
{
   if (dataLength > buffer.Length) 
      return OperationStatus.DestinationTooSmall;
   if (Utf8ToBytesInPlace(buffer.Slice(0, dataLength), out var consumed, out written)) 
      return OperationStatus.Done;
   return OperationStatus.InvalidData;
}
```

KrzysztofCwalina:
@GrabYourPitchforks, then the "specific implementation" does not implement the base abstraction; it lies about implementing it. 

BTW, I am pushing back because if we did that, then the Write method in the example above would have to fail if it cannot do the transformation.

GrabYourPitchforks:
@KrzysztofCwalina Why would the Write method fail? Presumably it would just fall back to non-in-place-transform logic (temporary buffer), same as if you'd given it an `IBufferOperation` instance instead of an `IBufferTransformation` instance.

Please see also my comment above re: breaking changes. If an unsealed class implements `IBufferOperation`, it is a breaking change for that class to later implement `IBufferTransformation`.

src/System.Buffers.Primitives/System/Buffers/Transformation.cs
KrzysztofCwalina: But then I have to compute, which would cause the buffer to be walked twice. Why would we want to do it? So we can not return an emum? Well, I take performance APIs that return an enum over APIs that force you to walk the data twice and return a bool.

KrzysztofCwalina:
@GrabYourPitchforks, then Write would have to take the base class as a parameter so it can select the appropriate transformation. I am not opposed to it, but then it basically means we don't need two interfaces and instead one that can optionally transform in place.

src/System.Buffers.Primitives/System/Buffers/Transformation.cs
ahsonkhan: > But then I have to compute, which would cause the buffer to be walked twice

You wouldn't walk the buffer twice since getting the required length is just an O(1) computation based on the length. I meant to use GetMaxEncodedToUtf8Length in the sample above, not ComputeRequiredLength.

GrabYourPitchforks: But you’d have to compute *anyway* if your in-place transformation could possibly increase the amount of data in the buffer. That’s because the operation is necessarily destructive and you have no way to signal partial completion to the caller in a manner that he can resume since the data he would need to pass into you has been overwritten. The only way to avoid this double computation is to require the caller to give you a large enough buffer ahead of time or to require in-place transforms to never produce more output than it consumes, which by definition means you can never receive “output too small” as a status code.

KrzysztofCwalina: It's O(1) only in the case of some transformations

KrzysztofCwalina: Yeah, we have to decide what to do with the destructive nature of some of these transformations. It's a big problem.

PR #1863, opened 2017-10-19
Modify Base64 APIs based on API review and update uses

ahsonkhan:
Made some changes to the Base64 APIs based on https://github.com/dotnet/corefx/issues/24568#issuecomment-337427063.

**Open Questions:**
1. How can we incorporate isFinalBlock true/false to the "in place" methods? Since they are destructive, if someone passes isFinalBlock false to EncodeToUtf8InPlace, what are we expected to do?
2. I am not sure if we can change DecodeFromUtf8InPlace to return bool instead of OperationStatus. We lose the OperationStatus.NeedMoreData state which we need for partial processing. If we return bool, what do we do in the case where the input buffer contains data that is not a multiple of 4 (i.e. the following scenario)?
```C#
Span<byte> buffer; // data that is a multiple of 4
int splitLocation; //some arbitrary location, not at a multiple of 4
Span<byte> input1 = buffer.Slice(0, splitLocation);
DecodeFromUtf8InPlace(input1, out consumed, out written); // returns NeedMoreData
Span<byte> input2 = intput1.Slice(consumed) + buffer.Slice(splitLocation);
DecodeFromUtf8InPlace(input2.CopyTo(buffer.Slice(written)), out consumed, out written); // returns Done
```


GrabYourPitchforks:
IMO partial processing doesn't make sense for the in-place transforms. The point of partial processing it to support streaming scenarios, and in these cases you're shuttling data from a source to a sink. With in-place transformations you're only ever operating on a single buffer, so there's an implicit expectation that all of the data on which to operate is already self-contained within that buffer. So there's no loop mechanism, hence no "is final block?" parameter needed. It's an implicit one-shot operation.

GrabYourPitchforks:
Also, regarding the return type and behavior of in-place transforms, you have three potential outcomes.

First, the transform could *expand* the input data, such as in the case of Base64Encode. As stated earlier, you can't possibly perform a partial transform, since you're necessarily overwriting some of the input data with the output. This causes the "remainder" of the input data to be lost, which means the method cannot be called in a loop. "Expand" transforms are therefore necessarily one-shot.

Second, the transform could *shrink* the input data, such as in the case of Base64Decode. In theory it's possible to support partial transforms here if you introduce an `int bytesConsumed` out parameter. But now you run into the issue that on the next iteration of the loop, the "start reading from here" and "start writing to here" references point to different locations in the buffer. The way to address this is to have `ReadOnlySpan<byte> source` and `Span<byte> sink` parameters, but then you've reinvented the non-in-place signature. Alternatively, the caller could always memmove the input data so that on each iteration it's properly aligned, but... why? It just seems like extra work for the caller when there's already a perfectly good non-in-place transform API. So "shrink" transforms are also presumably one-shot in the sense that they're not adding significant value for streaming scenarios.

The final option is a transform whose output length matches its input length. This is the only possible in-place transform that could be called in a loop correctly, which means that it could be used for non-one-shot scenarios. But there are still two problems. First, we don't have any transforms under consideration that maintain the data length during transform, so we don't know of any practical use case for this. Second, if you do assume that this is used in a streaming scenario, then you're also assuming that you're operating on multiple disjointed buffers. The looping code still needs to take care of stitching if a single block of data exists between multiple buffers, which means that there's a temporary buffer being used for this purpose, which means data is being copied between the temporary buffer and the actual data buffers, so why not just skip the memcpy step and pass the two buffer references directly to the non-in-place transform API? Like the "shrink" transform, the "keep" transform would in practice end up being one-shot since it's not adding significant value for streaming scenarios.

If everything is one-shot, then the input buffer is presumed to contain *all* of the data, which means that `NeedsMoreData` and `OutputTooSmall` return codes are meaningless. The operation either completes entirely or fails entirely, which implies a `bool` return type.

KrzysztofCwalina:
My intuition tells me re-entrant in-place transformations can be done (even for the expanding ones), but the APIs might be very wacky, and the implementation tricky.

ahsonkhan:
> The operation either completes entirely or fails entirely, which implies a `bool` return type.

In the case you return false, the input buffer that the user provided most likely get partially trampled on(unless we validate the input buffer first, which means walking it twice). What is the user expected to do when we return false? Given there is no recourse for the user, other than the user try to get back the original input from its source, should we throw instead and have in place APIS that return void rather than a TryX?

PR #1865, opened 2017-10-19
Add XML docs to Tensor library

ericstj:
/cc @eerhardt @sethjuarez @liqunfu

### DIFF
--- a/src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
@@ -5,61 +5,145 @@ namespace System.Numerics
 {
     public static class ArrayTensorExtensions
     {
+        /// <summary>
+        /// Creates a copy of this single-dimensional array as a DenseTensor<T>
+        /// </summary>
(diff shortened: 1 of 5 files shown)


src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
```diff
@@ -5,61 +5,145 @@ namespace System.Numerics
 {
     public static class ArrayTensorExtensions
     {
+        /// <summary>
+        /// Creates a copy of this single-dimensional array as a DenseTensor<T>
+        /// </summary>
+        /// <typeparam name="T"></typeparam>
+        /// <param name="array"></param>
+        /// <param name="reverseStride"></param>
+        /// <returns></returns>
```
eerhardt: I'm confused at why we have these empty elements?  If we aren't going to fill them out, why have them?

ericstj: These are the auto-generated triple-slash comments.  I'll fill out the parameters I just wanted to get a first pass with all the method/type docs.

PR #1872, opened 2017-10-23
Added SpanWriter

KrzysztofCwalina:
Added a prototype of byref SpanWriter and rewrote CosmosDB authentication token generation routine using the new writer (see CosmosDbAuthorizationHeader.TryWrite2). The rewritten routine is much much simpler and the perfromance regression is pretty much insignificant (2% in instructions retired; no detectable change in time elapsed). 

Note that CosmosDbAuthorizationHeader.TryWrite2 is 3x faster than the [MSDN prescribed way of generating the header](https://docs.microsoft.com/en-us/rest/api/documentdb/access-control-on-documentdb-resources). Also, it does not allocate any GC heap memory; MSDN routine allocates 2KB per header.

See the following benchmark results. "Raw" is [the complicated generation routine using low level primitive APIs](https://github.com/dotnet/corefxlab/blob/master/src/System.Azure.Experimental/System/Azure/CosmosDbAuthorizationHeader.cs#L33), "Writer" is the routine using the new SpanWriter, and "Msdn" is an MSDN sample using allocating classic APIs.

 Test Name         | Metric                                        |  AVERAGE 
:----------------- |:--------------------------------------------- |:---------: 
 AzureBench.Msdn   | Duration                                      |     0.008 
 AzureBench.Msdn   | Instructions Retired                          | 56800 
 AzureBench.Msdn   | Branch Mispredictions                         |    24.576 
 AzureBench.Msdn   | Cache Misses                                  |    81.920 
 AzureBench.Msdn   | Allocation Size on Benchmark Execution Thread |  1920 
 AzureBench.Raw    | Duration                                      |     0.003 
 AzureBench.Raw    | Instructions Retired                          | 19300 
 AzureBench.Raw    | Branch Mispredictions                         |     0 
 AzureBench.Raw    | Cache Misses                                  |    12.288 
 AzureBench.Raw    | Allocation Size on Benchmark Execution Thread |     0 
 AzureBench.Writer | Duration                                      |     0.003 
 AzureBench.Writer | Instructions Retired                          | 19800
 AzureBench.Writer | Branch Mispredictions                         |     8.192 
 AzureBench.Writer | Cache Misses                                  |    16.384 
 AzureBench.Writer | Allocation Size on Benchmark Execution Thread |     0 

cc: @ahsonkhan, @GrabYourPitchforks, @joshfree, @ianhays, @jkotas

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
@@ -29,7 +29,6 @@ namespace System.Buffers
 
         public void WriteBytes<T>(T value, IBufferTransformation transformation = null) where T : IWritable
         {
-            var start = Index;
             int written;
             while (!value.TryWrite(Free, out written, default(ParsedFormat)))
             {
@@ -38,7 +37,7 @@ namespace System.Buffers
 
             if (transformation != null)
             {
-                var status = transformation.Transform(Buffer.Slice(start), written, out written);
+                var status = transformation.Transform(Buffer.Slice(Index), written, out written);
                 if (status == OperationStatus.Done)
                 {
                     _written += written;
@@ -54,7 +53,6 @@ namespace System.Buffers
 
         public void WriteText<T>(T value, IBufferTransformation transformation = null) where T : IBufferFormattable
         {
-            var start = Index;
             int written;
             while (!value.TryFormat(Free, out written, default(ParsedFormat), _symbols))
             {
@@ -63,7 +61,7 @@ namespace System.Buffers
 
             if (transformation != null)
             {
-                var status = transformation.Transform(Buffer.Slice(start), written, out written);
+                var status = transformation.Transform(Buffer.Slice(Index), written, out written);
                 if (status == OperationStatus.Done)
                 {
                     _written += written;
@@ -111,7 +109,6 @@ namespace System.Buffers
 
         public void Write(string text, IBufferTransformation transformation = null)
         {
-            int start = Index;
             int encoded;
             while (true)
             {
@@ -124,10 +121,10 @@ namespace System.Buffers
 
             if (transformation != null)
             {
-                var status = transformation.Transform(Buffer.Slice(start), encoded, out int transformed);
+                var status = transformation.Transform(Buffer.Slice(Index), encoded, out int transformed);
                 if (status == OperationStatus.Done)
                 {
-                    _written += transformed - encoded; // as this was an in-place transform;
+                    _written += transformed;
                     return;
                 }
 
@@ -135,7 +132,7 @@ namespace System.Buffers
 
                 // if OperationStatus.DestinationTooSmall
                 Resize();
-                WriteLine(text, transformation);
+                Write(text, transformation);
             }
         }
 
@@ -163,18 +160,14 @@ namespace System.Buffers
 
         public void WriteBytes(ReadOnlySpan<byte> value)
         {
-            if (value.Length == 0) return;
-            while (Free.Length < value.Length) Resize();
-            value.CopyTo(Free);
-            _written += value.Length;
+            while (!value.TryCopyTo(Free)) Resize();
+            _written += value.Length; 
         }
 
         public void WriteBytes(ReadOnlyMemory<byte> value)
         {
-            if (value.Length == 0) return;
-            while (Free.Length < value.Length) Resize();
-            value.Span.CopyTo(Free);
-            _written += value.Length;
+            while (!value.Span.TryCopyTo(Free)) Resize();
+            _written += value.Length; 
         }
 
         public Span<byte> Free => _buffer.Slice(_written);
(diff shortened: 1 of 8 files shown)


benaadams:
>  MSDN routine allocates 2KB per header.

wtf

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
```diff
@@ -0,0 +1,207 @@
...
+            _buffer = buffer;
+            _written = 0;
+            _symbols = symbols == default(SymbolTable) ? SymbolTable.InvariantUtf8 : symbols;
+            Newline = s_defaultNewline;
+            Enlarge = null;
+        }
+
+        public void WriteBytes<T>(T value, IBufferTransformation transformation = null) where T : IWritable
+        {
+            var start = Index;
```
ahsonkhan: Do we need `start`? Why not just use `_written` to slice the buffer?

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
```diff
@@ -0,0 +1,207 @@
...
+                if (status == OperationStatus.Done)
+                {
+                    _written += written;
+                    return;
+                }
+
+                if (status == OperationStatus.InvalidData || status == OperationStatus.NeedMoreData) throw new Exception("invalid value or transformation");
+                // if OperationStatus.DestinationTooSmall
+                Resize();
+                WriteBytes(value, transformation);
```
ahsonkhan: Instead of calling the function from the beginning again, why not go to line 41 (inside the `if (transformation != null)`) block)? We have already written the value into the buffer so there is no need to redo that.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
```diff
@@ -0,0 +1,207 @@
...
+            while (!date.TryFormat(Free, out written, format, _symbols)) Resize();
+            _written += written;
+            WriteBytes(Newline);
+        }
+
+        public void WriteBytes(ReadOnlySpan<byte> value)
+        {
+            if (value.Length == 0) return;
+            while (Free.Length < value.Length) Resize();
+            value.CopyTo(Free);
```
ahsonkhan: Use TryCopyTo

```C#
while (!value.TryCopyTo(Free)) Resize();
_written += value.Length; 
```

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
```diff
@@ -0,0 +1,207 @@
...
+            _written += value.Length;
+        }
+
+        public void WriteBytes(ReadOnlyMemory<byte> value)
+        {
+            if (value.Length == 0) return;
+            while (Free.Length < value.Length) Resize();
+            value.Span.CopyTo(Free);
```
ahsonkhan: Same here. Use TryCopyTo.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
```diff
@@ -0,0 +1,207 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
ahsonkhan: Move SpanWriter.cs to System.Binary?
https://github.com/dotnet/corefxlab/blob/master/src/System.Binary/System/Binary/SpanWriter.cs

KrzysztofCwalina: It's not binary. It mainly writes text for now. Besides, for now I want to keep it in experimental. I am not sure this whole idea of passing in transformations is sound.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
KrzysztofCwalina: Because the transformation is in-place and could have destroyed the written value.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
KrzysztofCwalina: I think you are right. I will fix

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
KrzysztofCwalina: good suggestion. thanks.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
KrzysztofCwalina: good suggestion. thanks.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
```diff
@@ -0,0 +1,201 @@
...
+
+                // if OperationStatus.DestinationTooSmall
+                Resize();
+                WriteLine(text, transformation);
+            }
+        }
+
+        public void Write(string text, IBufferTransformation transformation = null)
+        {
+            int start = Index;
```
ahsonkhan: We don't need start here either.

KrzysztofCwalina: yeah. thanks.

PR #1874, opened 2017-10-24
Cleaned up Utf8String

KrzysztofCwalina:
Cleaned up Utf8String

@A-And, @joshfree, @ahsonkhan, @GrabYourPitchforks


KrzysztofCwalina:

tests/System.Text.Utf8String.Tests/Utf8StringPerformanceTests.cs
```diff
@@ -47,304 +47,18 @@ public void ConstructFromString(int length, int minCodePoint, int maxCodePoint,
         [InlineData(5, 32, 0xD7FF, "Short string", true)]
         [InlineData(50000, 32, 126, "Long ASCII string")]
         [InlineData(50000, 32, 0xD7FF, "Long string")]
-        public void EnumerateCodeUnitsConstructFromByteArray(int length, int minCodePoint, int maxCodePoint, string description, bool useInnerLoop = false)
```
ahsonkhan: Why are we removing all these performance tests? Are they no longer necessary?

KrzysztofCwalina: I removed enumerating code units from Utf8String/Span. The only way to enumerate code units now is to get Bytes property, which returns span, and then party on the span.

PR #1875, opened 2017-10-24
Update Base64 impl to support isFinalBlock = false

ahsonkhan:
Made some changes to the Base64 APIs based on https://github.com/dotnet/corefx/issues/24568#issuecomment-338833370.

Also updated DecodeFromUtf8InPlace to take in `dataLength` as arg and no longer return `out consumed`. This also means it will never return NeedsMoreData.

**Note:**
Since our IBufferOperation hasn't been updated to include isFinalBlock (along with the encoder/decoders that implement it), I have disabled the Pipe test.


ahsonkhan:
Is this PR good to go?

PR #1877, opened 2017-10-24
More Utf8Span/String cleanup

KrzysztofCwalina:
More cleanup. Mainly, I removed methods that operate on code units, streamlined Utf8String implementations by delegating to Utf8Span, and optimized some methods, e.g. ToString, LastIndexOf

cc: @ahsonkhan, @GrabYourPitchforks, @joshfree, @ianhays

### DIFF
--- a/samples/QotdService/Program.cs
+++ b/samples/QotdService/Program.cs
@@ -12,7 +12,7 @@ namespace QotdService
         public static void Main(string[] args)
         {
             var buffer = new byte[1024];
-            var quoteBytes = new Utf8Span("Insanity: doing the same thing over and over again and expecting different results. - Albert Einstein").CopyBytes();
+            var quote = new Utf8String("Insanity: doing the same thing over and over again and expecting different results. - Albert Einstein");
 
             var loop = new UVLoop();
 
@@ -22,9 +22,8 @@ namespace QotdService
             {
                 connection.ReadCompleted += (data) =>
                 {
-                    var quote = new Utf8String(quoteBytes);
-                    quote.CopyTo(buffer);
-                    connection.TryWrite(buffer, quote.Length);
+                    quote.Bytes.CopyTo(buffer);
+                    connection.TryWrite(buffer, quote.Bytes.Length);
                 };
 
                 connection.ReadStart();
--- a/src/System.Text.Utf8String/System/Text/Utf8Span.cs
+++ b/src/System.Text.Utf8String/System/Text/Utf8Span.cs
@@ -22,20 +22,7 @@ namespace System.Text.Utf8
 
         // TODO: Validate constructors, When should we copy? When should we just use the underlying array?
         // TODO: Should we be immutable/readonly?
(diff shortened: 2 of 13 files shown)


samples/QotdService/Program.cs
```diff
@@ -22,9 +22,8 @@ public static void Main(string[] args)
             {
                 connection.ReadCompleted += (data) =>
                 {
-                    var quote = new Utf8String(quoteBytes);
-                    quote.CopyTo(buffer);
-                    connection.TryWrite(buffer, quote.Length);
+                    quote.Bytes.CopyTo(buffer);
```
ahsonkhan: Is there any benefit in caching quote.Bytes into a local? Are we allowed to do it given Bytes is a Span type?

src/System.Text.Utf8String/System/Text/Utf8Span.cs
```diff
@@ -54,89 +41,47 @@ public Utf8Span(string utf16String)
...
-            return _buffer.SequenceEqual(other._buffer);
-        }
+        public bool Equals(Utf8Span other) => Bytes.SequenceEqual(other._buffer);
+
+        public bool Equals(Utf8String other) => Equals(other.Span);
 
         public bool Equals(string other)
         {
             Utf8CodePointEnumerator thisEnumerator = GetEnumerator();
+            Debug.Assert(BitConverter.IsLittleEndian);
```
ahsonkhan: Why is this assert necessary?

tests/System.Text.Utf8String.Tests/Utf8String.RandomTests.cs
```diff
@@ -132,32 +132,6 @@ public unsafe void StringEqualsConvertedToUtf16String(bool expected, string str1
         };
 
         [Theory]
-        [InlineData("", 'a')]
-        [InlineData("abc", 'a')]
-        [InlineData("v", 'v')]
-        [InlineData("1", 'a')]
-        [InlineData("1a", 'a')]
-        public void StartsWithCodeUnit(string s, char c)
```
ahsonkhan: Should we substitute these Utf8String tests that are being removed with tests on the new Utf8Span APIs?

src/System.Text.Utf8String/System/Text/Utf8Span.cs
KrzysztofCwalina: If we ever run (and test) in environments where System.String is UTF16BE this assert will fail and will remind us to implement Utf16BECodePointEnumerator

tests/System.Text.Utf8String.Tests/Utf8String.RandomTests.cs
KrzysztofCwalina: THis test is removed because it tests code-unit based APIs, which I removed.

samples/QotdService/Program.cs
KrzysztofCwalina: In general yes. in this case, it's just a sample. It actually should be rewritten to not use the temporary buffer.

PR #1881, opened 2017-10-26
UTF8 processing: performance improvements, validation, and other useful APIs

GrabYourPitchforks:
This PR introduces a workhorse routine for validating UTF-8 strings without performing a proper to-UTF16 conversion. It is intended to be used as the backend to UTF-8 validation routines, including routines that perform substitution of invalid sequences to convert an ill-formed UTF-8 string into a well-formed UTF-8 string. There are friendly APIs on the `Utf8Utility` type that allow inspecting the UTF-8 string. These APIs are intended to be compatible with `System.Text.Encoding.UTF8` in that they agree on the length of bad sequences, etc.

There is a sister workhorse "convert UTF-8 to UTF-16" routine, but it's not part of this PR since I'm still adding unit tests for it.

The unit tests for the workhorse routine perform some limited self-fuzzing. Additionally, the unit test project introduces the `NativeMemory` type (for unit test use only), which grabs memory pages directly from the kernel's page allocator and bookends the returned range with poison pages, where accessing these pages will immediately AV the process. This type helps detect buffer overruns in code that works directly with memory buffers in an unsafe fashion.

Finally, this API introduces substantial performance improvements over the existing "how many UTF-16 / UTF-32 code units would there be if I were to convert this UTF-8 string to one of those representations?" framework API `System.Text.Encoding.UTF8Encoding.GetChars`. I've copied the performance improvement table below.

# Perf results

1,000,000 iterations on each lipsum

Testbed: Intel(R) Core(TM) i7-6700 CPU @ 3.40GHz, 3408 Mhz, 8 Core(s), 8 Logical Processor(s), Win10 RS3 amd64fre

## English (ASCII)

* `System.Text.Encoding.UTF8`: 0.286 sec
* `Utf8Util.GetCharCount`: 0.077 sec

73% reduction in runtime

## Hebrew (primarily 2-byte)

* `System.Text.Encoding.UTF8`: 5.22 sec
* `Utf8Util.GetCharCount`: 1.81 sec

65% reduction in runtime

## Cyrillic (primarily 2-byte)

* `System.Text.Encoding.UTF8`: 5.20 sec
* `Utf8Util.GetCharCount`: 1.78 sec

65% reduction in runtime

## Japanese (primarily 3-byte)

* `System.Text.Encoding.UTF8`: 3.06 sec
* `Utf8Util.GetCharCount`: 1.08 sec

65% reduction in runtime

## Chinese (primarily 3-byte)

* `System.Text.Encoding.UTF8`: 6.30 sec
* `Utf8Util.GetCharCount`: 2.25 sec

65% reduction in runtime

The corpus that was used for testing can be found at https://github.com/GrabYourPitchforks/fast-utf8/blob/master/FastUtf8Tester/Lipsum.cs. That same directory contains the test program and full results.


ahsonkhan:
>  I've copied the performance improvement table below.

Can you please check-in the performance tests to this repo (preferably as part of this PR) so that we can run them using xunit.performance and our perf harness?

GrabYourPitchforks:
@ahsonkhan I just added the perf tests and sample texts. These are different than the sample texts I had been using, so the numbers will be somewhat off.

Drawaes:
@GrabYourPitchforks  +++10 I would love this to become the start of more fuzzing ability for .net as it's something I have been wanting to get around to for a while...

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+        /// The integer value of this scalar.
+        /// </summary>
+        public readonly int Value; // = U+0000 if using default init
+
+        /// <summary>
+        /// Constructs a Unicode scalar from the given UTF-16 code point.
+        /// The code point must not be a surrogate.
+        /// </summary>
+        /// <param name="char"></param>
+        public UnicodeScalar(char @char)
```
ahsonkhan: What is the meaning behind the '@' before the char parameter?
**Edit:** Interesting... using reserved keywords as parameter name - https://stackoverflow.com/questions/1038674/c-sharp-prefixing-parameter-names-with

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+        private UnicodeScalar(uint value)
+        {
+            Value = (int)value;
+        }
+
+        /// <summary>
+        /// Converts a <see cref="char"/> to a <see cref="UnicodeScalar"/>.
+        /// Identical to calling <see cref="UnicodeScalar.UnicodeScalar(char)"/>.
+        /// </summary>
+        public static explicit operator UnicodeScalar(char @char) => new UnicodeScalar(@char);
```
ahsonkhan: Why do we need these explicit operators?

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+
+                if ((uint)Value < 0x80U)
+                {
+                    return 1;
+                }
+                else if ((uint)Value < 0x800U)
+                {
+                    return 2;
+                }
+                else if ((uint)Value < 0x10000U)
```
ahsonkhan: Save 0x10000U (and others) as constants or use https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoders/EncodingHelper.cs#L13?

This is a duplicate of: https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Primitives/System/Text/Encoders/EncodingHelper.cs#L107
What should we do about these methods?

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Helpers.cs
```diff
@@ -0,0 +1,441 @@
...
+        {
+            return ((value & 0x80808080U) == 0U);
+        }
+
+        /// <summary>
+        /// Given a UTF-8 buffer which has been read into a DWORD in machine endianness,
+        /// returns <see langword="true"/> iff the buffer contains two UTF-8 sequences
+        /// that match the mask [ 110yyyyy 10xxxxxx 110yyyyy 10xxxxxx ]. This method *does not*
+        /// validate that the sequences are well-formed; the caller must still perform
+        /// overlong form checking.
```
ahsonkhan: What do you mean by "overlong form checking"?

tests/System.Text.Primitives.Tests/MemoryProtection/NativeMemory.Default.cs
```diff
@@ -0,0 +1,44 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Text.Primitives.Tests.MemoryProtection
+{
+    public static partial class NativeMemory
+    {
+        // for non-Windows systems
+        private static INativeMemory AllocateWithoutDataPopulationDefault(int cb)
```
ahsonkhan: What is cb?

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
```diff
@@ -0,0 +1,728 @@
...
+
+            // Begin the main loop.
+
+#if DEBUG
+            long lastOffsetProcessed = -1; // used for invariant checking in debug builds
+#endif
+
+            while (inputBufferRemainingBytes >= sizeof(uint))
+            {
+                BeforeReadDWord:
```
benaadams: `BeforeReadUInt` rather than `BeforeReadDWord`? similar for other occurrences

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
```diff
@@ -0,0 +1,728 @@
...
+
+                    SuccessfullyProcessedThreeByteSequence:
+
+                    // Optimization: A three-byte character could indicate CJK text, which makes it likely
+                    // that the character following this one is also CJK. We'll try to process several
+                    // three-byte sequences at a time.
+
+                    if (IntPtr.Size >= 8 && BitConverter.IsLittleEndian && inputBufferRemainingBytes >= (sizeof(ulong) + 1))
+                    {
+                        ulong thisQWord = Unsafe.ReadUnaligned<ulong>(ref Unsafe.Add(ref inputBuffer, inputBufferCurrentOffset));
```
benaadams: `thisULong` instead of `thisQWord`? Similar for other occurrences.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
```diff
@@ -0,0 +1,728 @@
...
+                            comparand = (uint)(thisQWord >> 48) & 0x200FU;
+                            if ((comparand == 0U) || (comparand == 0x200DU))
+                            {
+                                thisDWord = (uint)thisQWord;
+                                goto ProcessSingleThreeByteSequenceSkipOverlongAndSurrogateChecks;
+                            }
+
+                            inputBufferCurrentOffset += 9;
+                            inputBufferRemainingBytes -= 9;
+                            tempRuneCount -= 6; // 9 bytes -> 3 runes
```
benaadams: A nod to @migueldeicaza? 😄

tests/System.Text.Primitives.Tests/MemoryProtection/NativeMemory.Default.cs
whoisj: "cb" is shorthand for "count of bytes", it's Win32 speak for "size".

benaadams: `byteCount`? Can't use a Win32ism with the comment "for non-Windows systems" above it - that's just trollin' :)

GrabYourPitchforks: Yup, similarly, `cch` = "count of chars". Though this parameter name wasn't chosen because I'm in love with Hungarian notation. :) It was simply to match the already existing parameter name on `Marshal.AllocHGlobal`.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Helpers.cs
GrabYourPitchforks: "Overlong" is the Unicode term for a UTF-8 sequence that consists of more than the minimum number of code units required to represent the scalar value. For instance, it is forbidden to use a 2-byte sequence to represent an ASCII character.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
GrabYourPitchforks: I thought they'd be more convenient and result in nicer looking code than calling the ctor, but TBH I don't have a use case for it right now. Maybe I should just delete?

src/System.Text.Primitives/System/Text/Encoders/SequenceValidity.cs
```diff
@@ -0,0 +1,54 @@
...
+{
+    /// <summary>
+    /// Represents the validity of a UTF code unit sequence.
+    /// </summary>
+    public enum SequenceValidity
+    {
+        /// <summary>
+        /// The sequence is empty.
+        /// </summary>
+        Empty = 0,
```
KrzysztofCwalina: Why do we need both Empty and WellFormed?

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+using System.Globalization;
+
+namespace System.Buffers.Text
```
KrzysztofCwalina: I wonder what's the best namespace for these types. These are more general purpose than Buffers.Text, IMO

src/System.Text.Primitives/System/Text/Encoders/SequenceValidity.cs
GrabYourPitchforks: This enum is only used for the "read the first sequence from the string and tell me about it" API. The caller needs to distinguish between empty string and valid string so that he doesn't try interpreting the [meaningless] first sequence in an empty string. I thought about making these APIs "TryPeek" instead, but having both a bool and an enum returned by a single method left a bad taste in my mouth.

BTW, the "is this string valid?" APIs simply return bool, not this enum. This enum is only used when the caller needs to perform deep inspection, such as if he's iterating or performing replacement.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+        /// returns the scalar value as the UTF-16 code unit. Otherwise throws <see cref="OverflowException"/>.
+        /// </summary>
+        public static explicit operator char(UnicodeScalar value) => checked((char)value.Value);
+
+        /// <summary>
+        /// Returns the scalar value as a 24-bit integer. Equivalent to reading the <see cref="Value"/> field.
+        /// </summary>
+        public static explicit operator int(UnicodeScalar value) => value.Value;
+
+        public static bool operator ==(UnicodeScalar a, UnicodeScalar b) => (a.Value == b.Value);
```
KrzysztofCwalina: We use left/right for parameter names of binary operators.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+        public static bool operator >=(UnicodeScalar a, UnicodeScalar b) => (a.Value >= b.Value);
+
+        private bool IsValid => IsValidScalar((uint)Value);
+
+        /// <summary>
+        /// Returns the number of UTF-16 code units (<see cref="char"/>s) required to represent this scalar.
+        /// Scalars in the range [U+0000..U+FFFF] require one UTF-16 code unit.
+        /// Scalars in the range [U+10000..U+10FFFF] require two UTF-16 code units.
+        /// </summary>
+        public int Utf16CodeUnitCount
```
KrzysztofCwalina: Should this API be on the Utf16 encoder?

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+        {
+            return this.Value.CompareTo(other.Value);
+        }
+
+        /// <summary>
+        /// Copies the UTF-16 code unit representation of this scalar to an output buffer.
+        /// The buffer must be large enough to hold the required number of <see cref="char"/>s.
+        /// The <see cref="Utf16CodeUnitCount"/> property gives the required output buffer length.
+        /// </summary>
+        public void CopyUtf16CodeUnitsTo(Span<char> utf16)
```
KrzysztofCwalina: Wouldn't it be useful if the routine returned the number of code units written?

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+    {
+        /// <summary>
+        /// The Unicode Replacement Character U+FFFD.
+        /// </summary>
+        public static readonly UnicodeScalar ReplacementChar = new UnicodeScalar(0xFFFD);
+
+        /// <summary>
+        /// The integer value of this scalar.
+        /// </summary>
+        public readonly int Value; // = U+0000 if using default init
```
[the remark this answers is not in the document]
YohDeadfall: `uint` is better because `char` is unsigned too and there are no way to convert an negative integer to `char` or `uint` without `unchecked`.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+
+        public override int GetHashCode() => Value;
+
+        private static bool IsValidScalar(uint value) =>
+            (value < 0xD800U) || Utf8Util.IsInRangeInclusive(value, 0xE000U, 0x10FFFFU);
+
+        public override string ToString()
+        {
+            // This can be made much more efficient.
+            return "U+" + ((uint)Value).ToString("X4", CultureInfo.InvariantCulture);
```
KrzysztofCwalina: It might be good to return both the numeric and textual representation of the scalar.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Helpers.cs
```diff
@@ -0,0 +1,441 @@
...
+            // if (BitConverter.IsLittleEndian)
+            // {
+            //     return ((value & 0x80U) == 0U);
+            // }
+            // else
+            // {
+            //     return ((int)value >= 0);
+            // }
+
+            return (BitConverter.IsLittleEndian && ((value & 0x80U) == 0U))
```
KrzysztofCwalina: I wonder if it would not be better to have two sets of checks like that: one for little endian machines and other for big endian ones. Higher level loops would do one endinanness check, and then call the appropriate API.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Helpers.cs
```diff
@@ -0,0 +1,441 @@
...
+            {
+                return (int)(long)value;
+            }
+        }
+
+        /// <summary>
+        /// Returns <see langword="true"/> iff all bytes in <paramref name="value"/> are ASCII.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private static bool DWordAllBytesAreAscii(uint value)
```
KrzysztofCwalina: This is an internal API, but in public API we would use UInt32, not DWord.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Helpers.cs
```diff
@@ -0,0 +1,441 @@
...
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+//
+// This file contains utility methods shared by the UTF-8 workhorse methods.
+//
+
+namespace System.Buffers.Text
+{
+    internal static partial class Utf8Util
```
KrzysztofCwalina: All the APIs seem to related to UTF8 validation. Should this type be called Utf8ValidationHelpers?

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
```diff
@@ -0,0 +1,728 @@
...
+        // This method will consume as many ASCII bytes as it can using fast vectorized processing, returning the number of
+        // consumed (ASCII) bytes. It's possible that the method exits early, perhaps because there is some non-ASCII byte
+        // later in the sequence or because we're running out of input to search. The intent is that the caller *skips over*
+        // the number of bytes returned by this method, then it continues data processing from the next byte.
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static unsafe int ConsumeAsciiBytesVectorized(ref byte buffer, int length)
+        {
+            // Only allow vectorization if vector operations are hardware-accelerated
+            // and vectors are at least 128 bits in length.
+            if (Vector.IsHardwareAccelerated && (Vector<byte>.Count >= 2 * sizeof(ulong)))
```
KrzysztofCwalina: Should you skip vectorized implementation if the length is small?

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
GrabYourPitchforks: Are there any immediate concerns with this method being inefficient? I didn’t really bother improving this because I thought it would be used more for debugging than for anything else.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
```diff
@@ -0,0 +1,728 @@
...
+            {
+                // Need to pin buffer so the GC doesn't cause it to be misaligned when we're
+                // reading vectors from it.
```
KrzysztofCwalina: I heard there is no reason to do it. The perf penalty is not significant.
cc: @benaadams

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
```diff
@@ -0,0 +1,728 @@
...
+                    // Consume QWORDs or DWORDs until we're aligned with where we can start vectorization.
+                    // It's ok if we duplicate a little bit of work by having a single QWORD or DWORD read
+                    // overlap with the vector read area. It's not worth the overhead of checking for this
+                    // condition and avoiding the duplicate work.
+
+                    if (IntPtr.Size >= sizeof(ulong))
+                    {
+                        while (numBytesConsumed < offsetAtWhichVectorIsAligned)
+                        {
+                            if (!QWordAllBytesAreAscii(Unsafe.ReadUnaligned<ulong>(&pbBuffer[numBytesConsumed])))
```
KrzysztofCwalina: You know you can just reinterpret cast Span<byte> to Vector<byte> without having to use unsafe code/pointers. See https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/SpanHelpers.byte.cs#L474 for example.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
GrabYourPitchforks: I saw that the JITter produced VMOVDQU, so unaligned access should be safe, but I wasn’t sure if this was an implementation detail or not. I should perf test VMOVDQU vs VMOVDQA when the intrinsics come online.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
KrzysztofCwalina: Or rather this UInt64

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
benaadams: The idiv will likely be more significant :) `(int)pbBuffer % Vector<byte>.Count`

Is a perfhit if you cross cache lines (worse if you cross pages); so could go for 64 byte alignment and play it safe for all vector lengths?

```csharp
int referenceMisalignment = (int)pbBuffer & 63;
```

Issue if they go bigger than 512 bytes, but will also be larger issues as most algos are optimized for cache lines of 64 bytes

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+    /// http://www.unicode.org/versions/Unicode10.0.0/ch03.pdf
+    /// </remarks>
+    public static class Utf8Utility
+    {
+        private static readonly byte[] _replacementCharAsUtf8 = new byte[] { 0xEF, 0xBF, 0xBD };
+
+        /// <summary>
+        /// The Unicode Replacement Character (U+FFFD) as the three-byte UTF-8 sequence [ EF BF BD ].
+        /// </summary>
+        public static ReadOnlySpan<byte> ReplacementCharacterByteSequence => _replacementCharAsUtf8;
```
KrzysztofCwalina: Why not return byte[]?

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
GrabYourPitchforks: Yes, this is checked both by the caller (to avoid the non-inlineable call) and again by this method on line 37.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
GrabYourPitchforks: Yup, JIT optimizes the modulus operation since the value is known at codegen time. Sounds like I’ll still need to pin the buffer though since the GC doesn’t guarantee alignment beyond the natural word size, right?

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Helpers.cs
GrabYourPitchforks: There will also be UTF8-to-UTF16 code in this same static class at some point. But you’re right, the naming is horrible. Perhaps simply Utf8Helpers (drop the “validation”) would work?

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
GrabYourPitchforks: There’s a property that gives you that information and the caller needs to query the property ahead of time, so it seems redundant. Is your suggestion a TryCopy method?

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+namespace System.Buffers.Text
+{
+    /// <summary>
+    /// Contains utility methods for inspecting UTF-8 sequences.
+    /// </summary>
+    /// <remarks>
+    /// The UTF-8 specification can be found in the Unicode Standard, Sec. 3.9.
+    /// http://www.unicode.org/versions/Unicode10.0.0/ch03.pdf
+    /// </remarks>
+    public static class Utf8Utility
```
KrzysztofCwalina: Could all these static methods be on our Utf8 static encoder type?

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+
+        /// <summary>
+        /// Given the first byte of a sequence, returns the expected number of continuation bytes
+        /// which should follow this byte.
+        /// </summary>
+        /// <remarks>
+        /// Returns 0 if <paramref name="firstByte"/> does not have any expected continuation bytes
+        /// or if <paramref name="firstByte"/> cannot begin a well-formed UTF-8 sequence.
+        /// </remarks>
+        public static int GetExpectedNumberOfContinuationBytes(byte firstByte)
```
KrzysztofCwalina: Do we need this to be public?

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+        /// Returns the index of the first byte of the first invalid UTF-8 sequence in <paramref name="data"/>,
+        /// or -1 if <paramref name="data"/> is a well-formed UTF-8 string.
+        /// </summary>
+        public static int GetIndexOfFirstInvalidUtf8Sequence(ReadOnlySpan<byte> data) =>
+            Utf8Util.GetIndexOfFirstInvalidUtf8Sequence(data, out _, out _);
+
+        /// <summary>
+        /// Return <see langword="true"/> iff <paramref name="value"/> is an ASCII value (within the range 0-127, inclusive).
+        /// </summary>
+        public static bool IsAsciiValue(byte value) => (value < (byte)0x80);
```
KrzysztofCwalina: I have seen this method before. We shoudl have one in one place.

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+        /// <summary>
+        /// Returns <see langword="true"/> iff <paramref name="value"/> is a UTF-8 continuation byte.
+        /// A UTF-8 continuation byte is a byte whose value is in the range 0x80-0xBF, inclusive.
+        /// </summary>
+        public static bool IsUtf8ContinuationByte(byte value) => ((value & (byte)0xC0) == (byte)0x80);
+
+        /// <summary>
+        /// Returns <see langword="true"/> iff <paramref name="data"/> represents a well-formed UTF-8 string.
+        /// </summary>
+        public static bool IsWellFormedUtf8String(ReadOnlySpan<byte> data) => (GetIndexOfFirstInvalidUtf8Sequence(data) < 0);
```
KrzysztofCwalina: Shoudl the "String" suffix be dropped?

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+        /// <summary>
+        /// Peeks at the first UTF-8 sequence in the input buffer and returns information about that
+        /// sequence. If the sequence is well-formed, returns <see cref="SequenceValidity.WellFormed"/>
+        /// and sets the <paramref name="scalarValue"/> output parameter to the scalar value encoded by
+        /// the sequence. If the return value is anything other than <see cref="SequenceValidity.WellFormed"/>,
+        /// sets the <paramref name="scalarValue"/> output parameter to <see cref="UnicodeScalar.ReplacementChar"/>.
+        /// In all cases, the <paramref name="numBytesConsumed"/> output parameter will contain the
+        /// number of UTF-8 code units read from the input buffer in order to make the determination.
+        /// </summary>
+        public static SequenceValidity PeekFirstSequence(ReadOnlySpan<byte> data, out int numBytesConsumed, out UnicodeScalar scalarValue)
```
KrzysztofCwalina: should data be called utf8?

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+        /// </summary>
+        /// <param name="inputBuffer">The buffer containing UTF-8 text.</param>
+        /// <param name="charCount">
+        /// If this method returns <see langword="true"/>, contains the equivalent <see cref="char"/> count of the input string.
+        /// If this method returns <see langword="false"/>, the value is undefined.
+        /// </param>
+        /// <returns>
+        /// <see langword="true"/> on success, <see langword="false"/> if the input is not a well-formed UTF-8 string.
+        /// </returns>
+        public static bool TryGetUtf16CharCount(ReadOnlySpan<byte> inputBuffer, out int charCount)
```
KrzysztofCwalina: It's strange to see this on Utf8Util type. Also, inputBuffer should be called utf8?

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
```diff
@@ -0,0 +1,357 @@
...
+        /// </param>
+        /// <param name="bytesConsumed">
+        /// If this method returns <see langword="true"/>, contains the number of bytes of the input buffer which were consumed in order to generate this scalar value.
+        /// If this method returns <see langword="false"/>, the value is undefined.
+        /// </param>
+        /// <returns>
+        /// <see langword="true"/> if a scalar value could be decoded from the beginning of the input buffer,
+        /// <see langword="false"/> if the input buffer did not begin with a valid UTF-8 sequence.
+        /// </returns>
+        public static bool TryReadFirstRune(ReadOnlySpan<byte> inputBuffer, out int rune, out int bytesConsumed)
```
KrzysztofCwalina: Why "Rune"? We should stick to "Unicode scalar"

tests/System.Text.Primitives.Tests/MemoryProtection/NativeMemory.Windows.cs
```diff
@@ -0,0 +1,277 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.ConstrainedExecution;
+using System.Runtime.InteropServices;
+using System.Security;
+
+namespace System.Text.Primitives.Tests.MemoryProtection
+{
+    public static partial class NativeMemory
```
KrzysztofCwalina: It might be worth creating a subclass of OwnedMemory<T> that implements the protection semantics. Such subclass could be in System.Buffers.Experimental and then used in many tests.

GrabYourPitchforks: I agree. However I still have some questions as to how `OwnedMemory<T>` works and want to sit down with you and Ahson and others to get a better understanding of it. IMO I shouldn't let this block the merge, and I can create a new PR when that change is ready to come in.

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
GrabYourPitchforks: I don't have a use case for this API right now; will just remove until it's needed.

src/System.Text.Primitives/System/Text/Encoders/Utf8Util.Validation.cs
GrabYourPitchforks: Yes. However, this is an unaligned read, and the JITter produces optimal assembly for `Unsafe.ReadUnaligned<T>(void*)` but _not_ for `Unsafe.ReadUnaligned<T>(ref Unsafe.Add(...))`. I'm taking advantage of the fact that the buffer is already pinned in order to let me call the optimal overload.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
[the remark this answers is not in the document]
KrzysztofCwalina: There are pros and cons. I don't have a well formed opinion on which type we should chose, I just commented that we should choose because today we are using both types in various APIs.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
KrzysztofCwalina: I see. Thanks.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
KrzysztofCwalina: Yeah, it's not about efficiency. It's about debug-ability.

GrabYourPitchforks:
Since the bulk of the logic isn't changing and we're discussing utility methods / renamings, I'm going to merge what I have now so that people can immediately start playing with it. Will send a new PR with additional performance improvements and changes.

src/System.Text.Primitives/System/Text/Encoders/Utf8Utility.cs
redknightlois: Make sure these guys are annotated with AggresiveInline, there are edge cases where this kind of small methods are not inlined while when decorated most of the time they are honored.

src/System.Text.Primitives/System/Text/Encoders/UnicodeScalar.cs
```diff
@@ -0,0 +1,246 @@
...
+using System.Diagnostics;
+using System.Globalization;
+
+namespace System.Buffers.Text
+{
+    /// <summary>
+    /// Represents a 24-bit Unicode scalar value.
+    /// A scalar value is any value in the range [U+0000..U+D7FF] or [U+E000..U+10FFFF].
+    /// </summary>
+    public struct UnicodeScalar : IComparable<UnicodeScalar>, IEquatable<UnicodeScalar>
```
whoisj: Q: is `UnicodeScalar` the correct name, or should this be `CodePoint`? The two concepts are rather similar, and I'm asking because the terms are defined as:

> Code Point. (1) Any value in the Unicode codespace; that is, the range of integers from 0 to 10FFFF16. (See definition D10 in Section 3.4, Characters and Encoding.) Not all code points are assigned to encoded characters. See code point type. (2) A value, or position, for a character, in any coded character set.

> Unicode Scalar Value. Any Unicode code point except high-surrogate and low-surrogate code points. In other words, the ranges of integers 0 to D7FF16 and E00016 to 10FFFF16 inclusive. (See definition D76 in Section 3.9, Unicode Encoding Forms.)

PR #1883, opened 2017-10-26
Added Utf8String APIs to SpanWriter

KrzysztofCwalina:
- Added Utf8Span/String APIs to SpanWriter
- Added Azure CosmosDb headers generation APIs using Utf8Strings
- Moved non-SpanWriter implementation of CosmosDb API to Benchmarks.dll (to serve as a baseline). The SpanWriter version is on par in terms of perfromance.
- Added perf tests comparing the Azure Utf8 and Utf16 APIs. Utf8 APIs result in lass instructions retired.
- Refactored some reusable routines from System.Text.Utf8String to other packages
- Removed System.Binary package. New binary API prototypes will go into Buffers.Experimental

cc: @ahsonkhan, @GrabYourPitchforks, @joshfree, @A-And


ahsonkhan:
> Removed System.Binary package. 

Can you please remove the project reference from the PerfHarness as well (along with updating the assembly list)?
https://github.com/dotnet/corefxlab/blob/master/scripts/PerfHarness/PerfHarness.csproj#L14
https://github.com/dotnet/corefxlab/blob/master/scripts/PerfHarness/PerfHarness.cs#L63

src/System.Azure.Experimental/System/Azure/StorageAccessSignature.cs
```diff
@@ -1,4 +1,5 @@
-﻿// Copyright (c) Microsoft. All rights reserved.
+﻿
```
ahsonkhan: nit: remove extra line

src/System.Azure.Experimental/System/Azure/StorageAccessSignature.cs
```diff
@@ -11,6 +12,75 @@ namespace System.Azure.Authentication
 {
     public static class StorageAccessSignature
     {
+        public static bool TryWrite2(Span<byte> output, Sha256 hash, string verb, string canonicalizedResource, DateTime utc, out int bytesWritten)
```
ahsonkhan: Rename TryWrite2 to something more meaningful.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
```diff
@@ -192,6 +257,42 @@ private void Resize()
...
+            
+            return true;
+        }
+
+        private bool TryApplyTransformations(int index, int length, ReadOnlySpan<IBufferTransformation> transformations)
+        {
+            var buffer = Buffer.Slice(index);
+            for (int i = 0; i < transformations.Length; i++)
+            {
+                var status = transformations[i].Transform(buffer, length, out length);
```
ahsonkhan: Use a differently named local for the out parameter instead of length.
What happens if transformations.Length == 0? We end up setting _written += length (length being what the user passed in). Is that expected?

src/System.Buffers.Primitives/System/Buffers/BufferExtensions.cs
```diff
@@ -15,5 +15,58 @@ public static class BufferExtensions
...
+            if (values.Length == 0) return 0;
+
+            int candidateLength = buffer.Length;
+            var firstByte = values[0];
+            while (true)
+            {
+                int index = LastIndexOf(buffer.Slice(0, candidateLength), firstByte);
+                if (index == -1) return -1;
+                var slice = buffer.Slice(index);
+                if (slice.StartsWith(values)) return index;
```
ahsonkhan: Should this be values.Slice(1) (along with some bounds checks) since we already found the `firstByte` at `index`?

src/System.Azure.Experimental/System/Azure/StorageAccessSignature.cs
KrzysztofCwalina: will do

src/System.Azure.Experimental/System/Azure/StorageAccessSignature.cs
KrzysztofCwalina: The Idea was that I will implement it using SpanWriter, but I discovered SpanWriter does not work for this routine. I will remove it for now, fix SpanWriter, and then implement this routine again.

src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs
KrzysztofCwalina: Yes, if the last transformation produced no bytes, then no bytes have been written. i.e. using length here is by design.

src/System.Buffers.Primitives/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: I could have sliced the buffer by index + 1 and then call StartsWith with values.Slice(1), as you said. But, I think all this additional work to slice values and do more arithmetic would be more expensive than searching one byte one more time.

KrzysztofCwalina:
@ahsonkhan, I now removed the references from perf tests to System.Binary

src/System.Buffers.Primitives/System/Buffers/BufferExtensions.cs
```diff
@@ -15,5 +15,58 @@ public static class BufferExtensions
...
+        public static int SequenceCompareTo(this ReadOnlySpan<byte> left, ReadOnlySpan<byte> right)
+        {
+            var minLength = left.Length;
+            if (minLength > right.Length) minLength = right.Length;
+            for (int i = 0; i < minLength; i++)
+            {
+                var result = left[i].CompareTo(right[i]);
+                if (result != 0) return result;
+            }
+            return left.Length.CompareTo(right.Length);
```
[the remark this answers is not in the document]
KrzysztofCwalina: This is less readable and not really performant either. We need to rewirite this whole routine to either use SIMD or loop unrolling,

tests/Benchmarks/AzureBench.cs
```diff
@@ -88,5 +112,135 @@ static string CosmosDbBaselineFromMsdn(string key, string keyType, string verb,
...
+    public static bool TryWritePrimitives(Span<byte> output, Sha256 hash, string keyType, string verb, string resourceId, string resourceType, string tokenVersion, DateTime utc, out int bytesWritten)
+    {
+        int written, consumed, totalWritten = 0;
+        bytesWritten = 0;
+
+        Span<byte> buffer = stackalloc byte[AuthenticationHeaderBufferSize];
+
+        s_type.CopyTo(buffer);
+        totalWritten += s_type.Length;
+        var bufferSlice = buffer.Slice(totalWritten);
```
ahsonkhan: Do we need bufferSlice?
It looks like you always just update totalWritten and do `bufferSlice = buffer.Slice(totalWritten);`
Why not just pass `buffer.Slice(totalWritten)` everywhere instead of setting it to bufferSlice?

tests/Benchmarks/AzureBench.cs
```diff
@@ -88,5 +112,135 @@ static string CosmosDbBaselineFromMsdn(string key, string keyType, string verb,
...
+
+        if (Encodings.Utf16.ToUtf8(keyType.AsReadOnlySpan().AsBytes(), bufferSlice, out consumed, out written) != OperationStatus.Done)
```
ahsonkhan: We never use consumed anywhere in this method. Remove the local and replace with `out _`?

tests/Benchmarks/Benchmarks.csproj
```diff
@@ -29,6 +29,8 @@
   <!-- Project references -->
   <ItemGroup>
     <ProjectReference Include="..\..\src\System.Azure.Experimental\System.Azure.Experimental.csproj" />
+    <ProjectReference Include="..\..\src\System.Binary.Base64\System.Binary.Base64.csproj" />
```
ahsonkhan: Why is this project reference necessary?

src/System.Buffers.Primitives/System/Buffers/BufferExtensions.cs
ahsonkhan: That makes sense.

tests/Benchmarks/AzureBench.cs
KrzysztofCwalina: There are multiple scoped in this code all using consumed. So each callsite would have to have it's own variable, which is probably inefficient.

tests/Benchmarks/AzureBench.cs
KrzysztofCwalina: I can remove some of these. Later in the code, the sliced buffer is reused.

tests/Benchmarks/AzureBench.cs
ahsonkhan: Regardless, you save a couple of instructions (an xor and mov) if you don't have the local and replace all function calls with `out _`:
image

Issue #1876, opened 2017-10-24
Implement generic interfaces on Tensor<T>


ericstj:
Also look at what's exposed by T[,] and make sure we have everything.

Originally I just looked at Array so I was missing all the generic interfaces.

PR #1884, opened 2017-10-27
Implement IList<T> on Tensor<T>

eerhardt:
Implementing `IList<T>` on `Tensor<T>`.

/cc @KrzysztofCwalina @ahsonkhan

### DIFF
--- a/src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
@@ -114,7 +114,7 @@ namespace System.Numerics
 
         protected override int IndexOf(T item)
         {
-            // TODO: use Span.IndexOf when/if it removes the struct type constraint
+            // TODO: use Span.IndexOf when/if it removes the IEquatable type constraint
             if (Buffer.TryGetArray(out var arraySegment))
             {
                 var result = Array.IndexOf(arraySegment.Array, item, arraySegment.Offset, arraySegment.Count);
--- a/src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
@@ -98,6 +98,29 @@ namespace System.Numerics
             Buffer.Span[index] = value;
         }
 
+        protected override void CopyTo(T[] array, int arrayIndex)
+        {
+            Buffer.Span.CopyTo(array.AsSpan().Slice(arrayIndex));
+        }
+
+        protected override int IndexOf(T item)
+        {
+            // TODO: use Span.IndexOf when/if it removes the struct type constraint
+            if (Buffer.TryGetArray(out var arraySegment))
+            {
+                var result = Array.IndexOf(arraySegment.Array, item, arraySegment.Offset, arraySegment.Count);
+                if (result != -1)
+                {
+                    result -= arraySegment.Offset;
+                }
+                return result;
+            }
+            else
+            {
+                return base.IndexOf(item);
+            }
+        }
+
         /// <summary>
         /// Creates a shallow copy of this tensor, with new backing storage.
         /// </summary>
--- a/src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
@@ -100,6 +100,15 @@ namespace System.Numerics
 
         protected override void CopyTo(T[] array, int arrayIndex)
         {
+            if (array == null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+            if (array.Length < arrayIndex + Length)
+            {
+                throw new ArgumentException("The number of elements in the Tensor is greater than the available space from index to the end of the destination array.", nameof(array));
+            }
+
             Buffer.Span.CopyTo(array.AsSpan().Slice(arrayIndex));
         }
 
--- a/src/System.Numerics.Tensors/System/Numerics/Tensor.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/Tensor.cs
@@ -339,7 +339,7 @@ namespace System.Numerics
     /// <typeparam name="T">type contained within the Tensor.  Typically a value type such as int, double, float, etc.</typeparam>
     [DebuggerDisplay("{GetArrayString(false)}")]
     // When we cross-compile for frameworks that expose ICloneable this must implement ICloneable as well.
-    public abstract class Tensor<T> : IList, IList<T>, IStructuralComparable, IStructuralEquatable
+    public abstract class Tensor<T> : IList, IList<T>, IReadOnlyList<T>, IStructuralComparable, IStructuralEquatable
     {
         internal static ITensorArithmetic<T> arithmetic => TensorArithmetic.GetArithmetic<T>();
 
@@ -934,7 +934,7 @@ namespace System.Numerics
         #region IEnumerable members
         IEnumerator IEnumerable.GetEnumerator()
         {
-            return GetEnumerator();
+            return ((IEnumerable<T>)this).GetEnumerator();
         }
         #endregion
 
@@ -947,25 +947,25 @@ namespace System.Numerics
 
         void ICollection.CopyTo(Array array, int index)
         {
-            if (array == null)
-            {
-                throw new ArgumentNullException(nameof(array));
-            }
-            if (array.Rank != 1)
-            {
-                throw new ArgumentException("Only single dimensional arrays are supported for the requested action.", nameof(array));
-            }
-            if (array.Length < index + Length)
-            {
-                throw new ArgumentException("The number of elements in the Tensor is greater than the available space from index to the end of the destination array.", nameof(array));
-            }
-
             if (array is T[] destinationArray)
             {
                 CopyTo(destinationArray, index);
             }
             else
             {
+                if (array == null)
+                {
+                    throw new ArgumentNullException(nameof(array));
+                }
+                if (array.Rank != 1)
+                {
+                    throw new ArgumentException("Only single dimensional arrays are supported for the requested action.", nameof(array));
+                }
+                if (array.Length < index + Length)
+                {
+                    throw new ArgumentException("The number of elements in the Tensor is greater than the available space from index to the end of the destination array.", nameof(array));
+                }
+
                 for (int i = 0; i < length; i++)
                 {
                     array.SetValue(GetValue(i), index + i);
@@ -1043,7 +1043,7 @@ namespace System.Numerics
         #endregion
 
         #region IEnumerable<T> members
-        public IEnumerator<T> GetEnumerator()
+        IEnumerator<T> IEnumerable<T>.GetEnumerator()
         {
             for (int i = 0; i < Length; i++)
             {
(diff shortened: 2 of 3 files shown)


src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
```diff
@@ -98,6 +98,29 @@ public override void SetValue(int index, T value)
             Buffer.Span[index] = value;
         }
 
+        protected override void CopyTo(T[] array, int arrayIndex)
+        {
```
KrzysztofCwalina: Do you want to do any argument validation here?

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
```diff
@@ -98,6 +98,29 @@ public override void SetValue(int index, T value)
...
+
+        protected override int IndexOf(T item)
+        {
+            // TODO: use Span.IndexOf when/if it removes the struct type constraint
+            if (Buffer.TryGetArray(out var arraySegment))
+            {
+                var result = Array.IndexOf(arraySegment.Array, item, arraySegment.Offset, arraySegment.Count);
+                if (result != -1)
+                {
+                    result -= arraySegment.Offset;
```
KrzysztofCwalina: You also need to ensure that the result is not > array.Length - offset - count

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1039,6 +1029,117 @@ void IList.RemoveAt(int index)
...
+            }
+        }
+        #endregion
+
+        #region ICollection<T> members
+        int ICollection<T>.Count => (int)Length;
+
+        void ICollection<T>.Add(T item)
+        {
+            throw new InvalidOperationException();
```
KrzysztofCwalina: It makes me wonder why Tensor even implements the collection interfaces. I know it's not part of the PR, and more of a question to @ericstj

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1039,6 +1029,117 @@ void IList.RemoveAt(int index)
...
+        /// </summary>
+        /// <param name="array">
+        /// The one-dimensional Array that is the destination of the elements copied from Tensor<T>. The Array must have zero-based indexing.
+        /// </param>
+        /// <param name="arrayIndex">
+        /// The zero-based index in array at which copying begins.
+        /// </param>
+        protected virtual void CopyTo(T[] array, int arrayIndex)
+        {
+            for (int i = 0; i < length; i++)
```
KrzysztofCwalina: Do you want to validate arguments?

KrzysztofCwalina:
The title of the PR is a bit strange.

benaadams:
Not sure what enumerating over a Tenor means; but should it have a struct GetEnumerator rather than just an `IEnumerator<T> GetEnumerator()` version?

eerhardt:
> The title of the PR is a bit strange.

It took the title from my branch name and I forgot to update it.

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
eerhardt: Yes, will add.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
eerhardt: @ericstj is on leave for the next 3 months and I'm taking this work over.

I would think this behavior satisfies the collection interfaces because Tensor always returns `IsFixedSize = true`.  Thus you can't add/insert/remove.

From a general perspective, it makes sense to me for these interfaces to be implemented.  Tensor is a collection of values.  It just happens to be n-dimensional.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
eerhardt: Yes, will do.

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
eerhardt: I'm not sure how that could happen, can you explain?  I'm calling Array.IndexOf passing in the offset and a count.  Array.IndexOf shouldn't return a result that is past the count, will it?

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
KrzysztofCwalina: But what are the scenarios for Tensors being operated on as IList or IList\<T\>?

eerhardt: I believe the original intention was to match the API surface of Array. See the issue I'm fixing https://github.com/dotnet/corefxlab/issues/1876.

> Also look at what's exposed by T[,] and make sure we have everything.
Originally I just looked at Array so I was missing all the generic interfaces.

For `IEnumerable<T>`, a concrete scenario is with the existing CNTK library.  When you call their evaluation library, they take an `IEnumerable<T>` for the data.  See [CreateBatch](https://github.com/Microsoft/CNTK/blob/73cae11ed6707c5bfe0adc7a2f238da79ab52342/bindings/csharp/CNTKLibraryManagedDll/ShimApiClasses/ValueShim.cs#L104). 

Of course we want to make that exact API better in the future, by not copying the data to a new array, and directly passing the Tensor's memory into their native code.  But my thinking is that there are probably other libraries that take IList/Collection/etc for their data, and implementing these interfaces on Tensor allow for it to work with those libraries.

eerhardt:
> should it have a struct GetEnumerator rather than just an IEnumerator<T> GetEnumerator() version?

I'm not sure myself.  @KrzysztofCwalina - thoughts?  Do struct Enumerators still have a benefit when using generics?

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
```diff
@@ -98,6 +98,29 @@ public override void SetValue(int index, T value)
...
         }
 
+        protected override void CopyTo(T[] array, int arrayIndex)
+        {
+            Buffer.Span.CopyTo(array.AsSpan().Slice(arrayIndex));
+        }
+
+        protected override int IndexOf(T item)
+        {
+            // TODO: use Span.IndexOf when/if it removes the struct type constraint
```
ahsonkhan: This should be complete soon: https://github.com/dotnet/corefx/pull/24911

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -956,9 +947,16 @@ private IEnumerable<T> Enumerate()
 
         void ICollection.CopyTo(Array array, int index)
         {
-            for(int i = 0; i < length; i++)
+            if (array is T[] destinationArray)
```
ahsonkhan: What is the difference between System.Array and T[]? When is array not a T[]?

eerhardt: The interface ICollection specifies that the parameter is `Array`, which can be any array. Someone could pass an `int[]` into a `Tensor<double>`.  Or even `Array` could be a multi-dimension array - `int[,]`.

As @KrzysztofCwalina points out, I need to add parameter validation here.

KrzysztofCwalina:
>> should it have a struct GetEnumerator rather than just an IEnumerator GetEnumerator() version?

> I'm not sure myself. @KrzysztofCwalina - thoughts? Do struct Enumerators still have a benefit when using generics?

As I mentioned above, I am not sure what the scenarios for enumerating tensors are. I can see how IEnumerable<T> is the only interface that might make sense. Struct based enumerator cannot be used by reusable routines, and is one needs to "enumerate" the concrete tensor, they should use the indexer.

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
KrzysztofCwalina: My bad. The line is so long that I did not see the end of it in the default browser view. It's all good.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
YohDeadfall: But what's about `IReadOnlyList<T>`? `IList<T>` doesn't inherit from it because of [that issue](https://social.msdn.microsoft.com/Forums/vstudio/en-US/b4fb991a-3f5c-4923-93d4-7cd5c004f859/new-interfaces-ireadonlylist-and-ireadonlydictionary?forum=netfxbcl), so classes should implement it themselves.

> **Why did we not change the existing interfaces to extend the read-only interfaces?**
> It looks like a reasonable assumption that it works because the read-only interfaces are purely a subset of the read-write interfaces. Unfortunately, it is incompatible because at the metadata level every method on every interface has its own slot (which makes explicit interface implementations work).

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -956,9 +947,29 @@ private IEnumerable<T> Enumerate()
...
+            if (array.Rank != 1)
+            {
+                throw new ArgumentException("Only single dimensional arrays are supported for the requested action.", nameof(array));
+            }
+            if (array.Length < index + Length)
+            {
+                throw new ArgumentException("The number of elements in the Tensor is greater than the available space from index to the end of the destination array.", nameof(array));
+            }
+
+            if (array is T[] destinationArray)
```
YohDeadfall: You already have not null and array size checks in the protected `CopyTo` method. Skip these checks if array is `T[]`.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1039,6 +1042,126 @@ void IList.RemoveAt(int index)
         }
         #endregion
 
+        #region IEnumerable<T> members
+        public IEnumerator<T> GetEnumerator()
```
benaadams: > As I mentioned above, I am not sure what the scenarios for enumerating tensors are. I can see how IEnumerable is the only interface that might make sense. 

Can this be explicit then
```csharp
IEnumerator<T> IEnumerable<T>.GetEnumerator()
```
So if it was determined in the future that a struct enumerator would be advantageous (no allocations; no virtual calls, inlining) - it wouldn't be a breaking change by changing the return type

eerhardt: I will make it explicit.  I've also opened #1893 to consider adding the struct Enumerator.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
eerhardt: Done.

src/System.Numerics.Tensors/System/Numerics/DenseTensor.cs
eerhardt: It seems I still can't use the method since it requires `where T : IEquatable<T>`.  We don't have that constraint either.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
eerhardt: Will add `IReadOnlyList<T>`.

PR #1887, opened 2017-10-27
Revert "Delete System.Threading.Tasks.Channels from corefxlab"

ahsonkhan:
This reverts commit 29db22ae8636cbe83944461a5fa0516799fc91fa.


stephentoub:
@ahsonkhan, we now have a copy of the code in corefx and this reverted copy in corefxlab, which will no longer be maintained.  When will you be re-deleting it from here?

ahsonkhan:
> we now have a copy of the code in corefx and this reverted copy in corefxlab

I reverted the deletion only for the signalr alpha 2 branch so that I can produce signed packages for their consumption. I did not see the package with these APIs in the latest build on myget.

PR #1886, opened 2017-10-27
Update use of MemoryHandle to react to property rename

ahsonkhan:
Supersedes https://github.com/dotnet/corefxlab/pull/1853

cc @dotnet/corefxlab-contrib

Issue #1729, opened 2017-08-29
Productize System.Binary.Base64


KrzysztofCwalina:
Probably in this order:

- [x] Move non-obsolete APIs to  ~System.Buffers.dll~ System.Memory.dll in corefx and namespace ~System.Buffers~ System.Buffers.Text
- [x] Port SignalR to the new APIs
- [x] Remove corfxlab project.
- [x] Ensure good test coverage

ahsonkhan:
Waiting for SignalR to start using these APIs for validation as the last work item for this issue, similar to https://github.com/dotnet/corefxlab/issues/1730#issuecomment-337632375.

@muratg, @moozzyk @BrennanConroy

moozzyk:
I will start looking into tomorrow.

ahsonkhan:
This is now complete. See https://github.com/aspnet/SignalR/pull/1127

PR #1896, opened 2017-10-31
Remove Base64 APIs and OperationStatus and update uses

ahsonkhan:
Part of https://github.com/dotnet/corefxlab/issues/1729

Supersedes https://github.com/dotnet/corefxlab/pull/1892 & https://github.com/dotnet/corefxlab/pull/1886

~Depends on https://github.com/dotnet/corefxlab/pull/1886 to fix the MemoryHandle CI failures.~

cc @dotnet/corefxlab-contrib, @pakrym, @davidfowl


src/System.Azure.Experimental/System/Azure/CosmosDbAuthorizationHeader.cs
```diff
@@ -42,7 +42,7 @@ public static bool TryWrite(Span<byte> output, Sha256 hash, string keyType, stri
             writer.Write(tokenVersion);
             writer.Write("&sig=");
 
-            writer.WriteBytes(hash, default, Base64.BytesToUtf8Encoder);
+            writer.WriteBytes(hash, default, Base64Experimental.BytesToUtf8Encoder);
```
KrzysztofCwalina: Why Base64Experimental? Why not call the corfxlab APIs?

ahsonkhan: I don't understand what you mean. `Base64Experimental.BytesToUtf8Encoder` is a type in corefxlab.
https://github.com/dotnet/corefxlab/pull/1896/files#diff-bdbcc9558ff685630e7887fe9c35bf95R14

The WriteBytes method takes an IBufferTransformation as argument so that is why I am passing a Utf8Encoder to it.

ahsonkhan:

src/System.Azure.Experimental/System/Azure/CosmosDbAuthorizationHeader.cs
KrzysztofCwalina: Why do we need Base64Experimental give now we have official Base64 APIs in corfx.

ahsonkhan: For the types that implement IBufferTransformation and IBufferOperation such as Utf8Encoder and Utf8Decoder.
https://github.com/ahsonkhan/corefxlab/blob/master/src/System.Binary.Base64/System/Binary/Base64Encoder.cs#L207

KrzysztofCwalina: Ah, thanks. We might need to come up with a better name/design for these, but for now it's ok.

PR #1897, opened 2017-11-01
Added ReadOnlyBytes.CursorOf

KrzysztofCwalina:
cc: @davidfowl, @pakrym, @ahsonkhan, @terrajobst, @joshfree

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -179,6 +179,24 @@ namespace System.Buffers
             return -1;
         }
 
+        public Cursor CursorOf(byte value)
+        {
+            var first = _first.Span;
+            var index = first.IndexOf(value);
+            if (index != -1) return new Cursor(this, index);
+            if (_rest == null) return default;
+            return CursorOf(_rest, value);
+        }
+
+        private static Cursor CursorOf(IReadOnlyBufferList<byte> list, byte value)
+        {
+            var first = list.First.Span;
+            var index = first.IndexOf(value);
+            if (index != -1) return new Cursor(list, index);
+            if (list.Rest == null) return default;
+            return CursorOf(list.Rest, value);
+        }
+
         public int CopyTo(Span<byte> buffer)
         {
             var first = First;
@@ -329,6 +347,18 @@ namespace System.Buffers
             else
             {
                 return new ReadOnlyBytes(first);
+            }
+        }
+
+        public struct Cursor
+        {
+            IReadOnlyBufferList<byte> _node;
+            int _index;
+
+            public Cursor(IReadOnlyBufferList<byte> node, int index) : this()
+            {
+                _node = node;
+                _index = index;
             }
         }
     }
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -95,6 +95,8 @@ namespace System.Buffers
             }
         }
 
+        public bool IsEmpty => _first.Length == 0 && _rest == null;
+
         public static ReadOnlyBytes Empty => s_empty;
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -181,8 +181,8 @@ namespace System.Buffers
 
         public Cursor CursorOf(byte value)
         {
-            var first = _first.Span;
-            var index = first.IndexOf(value);
+            ReadOnlySpan<byte> first = _first.Span;
+            int index = first.IndexOf(value);
             if (index != -1) return new Cursor(this, index);
             if (_rest == null) return default;
             return CursorOf(_rest, value);
@@ -190,8 +190,8 @@ namespace System.Buffers
 
         private static Cursor CursorOf(IReadOnlyBufferList<byte> list, byte value)
         {
-            var first = list.First.Span;
-            var index = first.IndexOf(value);
+            ReadOnlySpan<byte> first = list.First.Span;
+            int index = first.IndexOf(value);
             if (index != -1) return new Cursor(list, index);
             if (list.Rest == null) return default;
             return CursorOf(list.Rest, value);
@@ -355,7 +355,7 @@ namespace System.Buffers
             IReadOnlyBufferList<byte> _node;
             int _index;
 
-            public Cursor(IReadOnlyBufferList<byte> node, int index) : this()
+            public Cursor(IReadOnlyBufferList<byte> node, int index)
             {
                 _node = node;
                 _index = index;
--- a/src/System.Text.Http/System/Text/Http/HttpRequest.cs
+++ b/src/System.Text.Http/System/Text/Http/HttpRequest.cs
@@ -92,7 +92,7 @@ namespace System.Text.Http
         public bool TryGet(ref Position position, out HttpHeader value, bool advance = true)
         {
             var bytes = _headers.Slice(position.IntegerPosition);
-            if (bytes.Length == 0)
+            if (bytes.IsEmpty)
             {
                 value = default;
                 return false;


src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -177,6 +179,24 @@ int IndexOfRest(byte value, int firstLength)
             return -1;
         }
 
+        public Cursor CursorOf(byte value)
+        {
+            var first = _first.Span;
+            var index = first.IndexOf(value);
```
ahsonkhan: Is the use of `var` ok here?

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -329,5 +349,17 @@ public static ReadOnlyBytes Create(params byte[][] buffers)
                 return new ReadOnlyBytes(first);
             }
         }
+
+        public struct Cursor
+        {
+            IReadOnlyBufferList<byte> _node;
+            int _index;
+
+            public Cursor(IReadOnlyBufferList<byte> node, int index) : this()
```
ahsonkhan: Do we need `this()`?

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: fixed.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: refactoring. fixed.

davidfowl:
@KrzysztofCwalina I don't see the point to this change. How do you use the cursor?

KrzysztofCwalina:
What do you want to do with it? It give O(1) access to a location in ROB.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -177,6 +179,24 @@ int IndexOfRest(byte value, int firstLength)
             return -1;
         }
 
+        public Cursor CursorOf(byte value)
+        {
+            ReadOnlySpan<byte> first = _first.Span;
+            int index = first.IndexOf(value);
+            if (index != -1) return new Cursor(this, index);
```
pakrym: Allocation in most common case?

YohDeadfall: Where? `Cursor` is a struct.

YohDeadfall:
Ah, now I see. Thanks.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
YohDeadfall: Ah, now I see. Thanks.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -329,5 +349,17 @@ public static ReadOnlyBytes Create(params byte[][] buffers)
                 return new ReadOnlyBytes(first);
             }
         }
+
+        public struct Cursor
```
benaadams: Perhaps this should this be
```csharp
public struct Cursor<TBufferList> where TBufferList : IReadOnlyBufferList<byte>
{
    TBufferList _node;
```

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -329,5 +349,17 @@ public static ReadOnlyBytes Create(params byte[][] buffers)
...
+        {
+            IReadOnlyBufferList<byte> _node;
+            int _index;
+
+            public Cursor(IReadOnlyBufferList<byte> node, int index)
```
benaadams: And this?
```csharp
Cursor(TBufferList node, int index)
```

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
YohDeadfall: Then the `CursorOf` method will be generic too, and `this` should be convertible to `TBufferList` which is impossible.

```
        public Cursor<TBufferList> CursorOf<TBufferList>(byte value)
            where TBufferList : IReadOnlyBufferList<byte>
        {
            ReadOnlySpan<byte> first = _first.Span;
            int index = first.IndexOf(value);
            if (index != -1) return new Cursor<TBufferList>(this, index); // CS1503: cannot convert from 'ReadOnlyBytes' to 'T'
            if (_rest == null) return default;
            return CursorOf<TBufferList>((T)_rest, value); // runtime error if _rest is not T
        }
```

benaadams: Its a concrete type at this point not a generic; so would be:
```csharp
public Cursor<ReadOnlySpan<byte>> CursorOf(byte value)
{
    ReadOnlySpan<byte> first = _first.Span;
    int index = first.IndexOf(value);
    if (index != -1) return new Cursor(this, index);
    if (_rest == null) return default;
    return CursorOf(_rest, value); 
}

private static Cursor<ReadOnlySpan<byte>> CursorOf(IReadOnlyBufferList<byte> list, byte value)
{
    ReadOnlySpan<byte> first = list.First.Span;
    int index = first.IndexOf(value);
    if (index != -1) return new Cursor(new ReadOnlyBytes(first, list.Rest), index);
    if (list.Rest == null) return default;
    return CursorOf(list.Rest, value);
}
```
Probably could event drop the generic type from Cursor at that point?

YohDeadfall: Do you mean `ReadOnlyBytes` instead of `ReadOnlySpan<byte>`?

benaadams: Yes :)

benaadams: Coding on a phone is hard :(

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: I will fix it. Null list will mean first.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
YohDeadfall: So if I understand you, you want to move `Cursor` outside of `ReadOnlyBytes` and make it generic. Then a consumer method will accept `Cursor<ReadOnlyBytes>` as an argument like that:

```csharp
public ReadOnlyBytes Slice(Cursor<ReadOnlyBytes> from, Cursor<ReadOnlyBytes> to)
```

Yes? I think it could be simplified.

```csharp
public struct ReadOnlyBytes : IReadOnlyBufferList<byte>
{
    public Cursor CursorOf(byte value)
    {
        ReadOnlySpan<byte> first = _first.Span;
        int index = first.IndexOf(value);
        if (index != -1) return new Cursor(_first, _rest, index);
        if (_rest == null) return default;
        return CursorOf(_rest, value);
    }

    private static Cursor CursorOf(IReadOnlyBufferList<byte> list, byte value)
    {
        ReadOnlySpan<byte> first = list.First.Span;
        int index = first.IndexOf(value);
        if (index != -1) return new Cursor(list.First, list.Rest, index);
        if (list.Rest == null) return default;
        return CursorOf(list.Rest, value);
    }

    public ReadOnlyBytes Slice(Cursor from, Cursor to)
    {
        // if from.First equals to to.First
        //     return from.First.Slice(from.Index, to.Index)
        // first ROB = from.First.Slice(from.Index)
        // while current ROBList not equals to to.Rest
        //     current ROBs = current ROBList.First
        // if to.First is not empty
        //     last ROB = if to.First is not empty
    }
}

public readonly struct Cursor
{
    internal readonly ReadOnlyMemory<byte> First;
    internal readonly IReadOnlyBufferList<byte> Rest;
    internal readonly int Index;

    public Cursor(ReadOnlyMemory<byte> node, IReadOnlyBufferList<byte> rest, int index) { }
}
```

PR #1900, opened 2017-11-03
Removed allocation when creating cursors to first segment

KrzysztofCwalina:
Also, I added Slice that takes a cursor as @davidfowl was asking how the cursor would be used.

cc: @pakrym, @davidfowl, @benaadams, @ahsonkhan, @YohDeadfall, @joshfree

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -10,11 +10,11 @@ namespace System.Buffers
     /// <summary>
     ///  Multi-segment buffer
     /// </summary>
-    public struct ReadOnlyBytes : IReadOnlyBufferList<byte>
+    public readonly struct ReadOnlyBytes : IReadOnlyBufferList<byte>
     {
-        ReadOnlyMemory<byte> _first; // pointer + index:int + length:int
-        IReadOnlyBufferList<byte> _all;
-        long _totalLength;
+        readonly ReadOnlyMemory<byte> _first; // pointer + index:int + length:int
+        readonly IReadOnlyBufferList<byte> _all;
+        readonly long _totalLength;
 
         static readonly ReadOnlyBytes s_empty = new ReadOnlyBytes(ReadOnlyMemory<byte>.Empty);
 
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -148,11 +148,17 @@ namespace System.Buffers
             return -1;
         }
 
+        public ReadOnlyBytes Slice(Cursor from)
+        {
+            if (from._node == null) Slice(from._index);
+            return new ReadOnlyBytes(from._node).Slice(from._index);
+        }
+
         public Cursor CursorOf(byte value)
         {
             ReadOnlySpan<byte> first = _first.Span;
             int index = first.IndexOf(value);
-            if (index != -1) return new Cursor(this, index);
+            if (index != -1) return new Cursor(null, index);
             if (_rest == null) return default;
             return CursorOf(_rest, value);
         }
@@ -295,8 +301,8 @@ namespace System.Buffers
 
         public struct Cursor
         {
-            IReadOnlyBufferList<byte> _node;
-            int _index;
+            internal IReadOnlyBufferList<byte> _node;
+            internal int _index;
 
             public Cursor(IReadOnlyBufferList<byte> node, int index)
             {
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -13,32 +13,36 @@ namespace System.Buffers
     public struct ReadOnlyBytes : IReadOnlyBufferList<byte>
     {
         ReadOnlyMemory<byte> _first; // pointer + index:int + length:int
-        IReadOnlyBufferList<byte> _rest;
+        IReadOnlyBufferList<byte> _all;
         long _totalLength;
 
         static readonly ReadOnlyBytes s_empty = new ReadOnlyBytes(ReadOnlyMemory<byte>.Empty);
 
-        public ReadOnlyBytes(ReadOnlyMemory<byte> first, IReadOnlyBufferList<byte> rest, long length)
+        public ReadOnlyBytes(ReadOnlyMemory<byte> buffer)
         {
-            if(length < 0) throw new ArgumentOutOfRangeException(nameof(length));
-            if (first.Length + (rest == null ? 0 : rest.Length) < length)
-                throw new ArgumentOutOfRangeException(nameof(length), length, string.Format("{0}+{1}", first.Length, rest.Length));
-            _rest = rest;
+            _first = buffer;
+            _all = null;
+            _totalLength = _first.Length;
+        }
+
+        public ReadOnlyBytes(IReadOnlyBufferList<byte> segments, long length)
+        {
+            _first = segments.First;
+            _all = segments;
+            _totalLength = segments.Length;
+        }
+
+        private ReadOnlyBytes(ReadOnlyMemory<byte> first, IReadOnlyBufferList<byte> all, long length)
+        {
+            Debug.Assert(all != null);
+            // TODO: add assert that first overlaps all (once we have Overlap on Span)
             _first = first;
+            _all = all;
             _totalLength = length;
         }
 
-        public ReadOnlyBytes(ReadOnlyMemory<byte> first, IReadOnlyBufferList<byte> rest) :
-            this(first, rest, first.Length + (rest == null ? 0 : rest.Length)) { }
-
-        public ReadOnlyBytes(ReadOnlyMemory<byte> buffer) :
-            this(buffer, null, buffer.Length) { }
-
         public ReadOnlyBytes(IReadOnlyBufferList<byte> segments) :
-            this(segments == null ? ReadOnlyMemory<byte>.Empty : segments.First, segments?.Rest, segments == null ? 0 : segments.Length) { }
-
-        public ReadOnlyBytes(IReadOnlyBufferList<byte> segments, long length) :
-            this(segments == null ? ReadOnlyMemory<byte>.Empty : segments.First, segments?.Rest, length) { }
+            this(segments, segments == null ? 0 : segments.Length) { }
 
         public bool TryGet(ref Position position, out ReadOnlyMemory<byte> value, bool advance = true)
         {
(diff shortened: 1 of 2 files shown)


src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -295,8 +301,8 @@ public static ReadOnlyBytes Create(params byte[][] buffers)
 
         public struct Cursor
```
YohDeadfall: Should it be readonly?

YohDeadfall: Should it be readonly?

KrzysztofCwalina: Yeah, I made it readonly. Though, honestly I am trying to figure out how to use Position as the cursor. i.e. it can be easily done but it might be confusing that the same type would represent a position in the sequence (linked list node) and a position withing the sequence of bytes. I need to think about it.

pakrym:
This implementation has severe limitation: it can only be used with the exact slice it was produced from. This makes it impossible to use for things like `consumed/examined` and might causes very non obvious bugs in user code:

```
var rob = ReadOnlyBytes.Create(new [] {1,2,3,4,5,6,7,8});
var c = rob.CursorOf(5);

var slice1 = rob.Slice(2).Slice(c);
var slice2 = rob.Slice(c).Slice(2);

```
In example above both slices would be the same which is a bug that is very hard to spot. Too make things even worse above code would have different behavior in case when `rob` is multi segmented.

YohDeadfall:
Okay, let's think about `Cursor`'s design. There are three ways to implement it and each of the ways brings its own problems: make it absoulte (#1897), relative (just offset from thee start, i.e. `struct Cursor { long Index }`, or mixed (implemented in this pull request).

The first one has a bug:
```csharp
var rob1 = ReadOnlyBytes.Create(new [] { 1, 2, 3, 4 });
var rob2 = ReadOnlyBytes.Create(new [] { 5, 6, 7, 8 });

var cursor = rob1.CursorOf(3);
var slice = rob2.Slice(cursor);
```

And what do we have? A correct result from the incorrect input because `this` is unused in the `Slice` method.

The bug of the second approach was demonstrated by @pakrym. Just replace `CursorOf` with `IndexOf`. As you can see, the problem isn't in the `Cursor` implementation, but in the design of `ReadOnlyBuffer`. It's a single linked list without a collection which owns it.

So there are no bug-free version and you should choose which bugs are acceptable.

pakrym:
Absolutely positioned cursor has enough information to check that chain that is being sliced contains segment from cursor. So the example you gave just illustrates another bug of implementation that can be prevented and there is bug free version.

Relative cursor should not be called cursor it would be just `int` and customers wouldn't expect any special cursor properties from it.

YohDeadfall:
Is it allowed to iterate from the start of the current chain to check that its rest contains `from`?  Are the following situations acceptable?

```csharp
var rob = ReadOnlyBytes.Create(new [] { 1, 2, 3, 4, 5, 6, 7, 8 });
var c = rob.CursorOf(3);
var slice = rob.Slice(4).Slice(c);
```
```csharp
var rob1 = ReadOnlyBytes.Create(new [] { 1, 2 }, new [] { 3, 4 }, new [] { 5, 6 }, new [] { 7, 8 });
var rob2 = rob.Slice(5);
var c = rob1.CursorOf(6);
var slice = rob2.Slice(c);
```

I think that a lot of additional checks would hurt performance, but yep, then it's possible to eliminate all bugs.

YohDeadfall:
@pakrym Is it?

```csharp
public Cursor CursorOf(byte value)
{
    ReadOnlySpan<byte> first = _first.Span;
    int index = first.IndexOf(value);
    if (index != -1) return new Cursor(_first, _rest, index);
    if (_rest == null) return default;
    return CursorOf(_rest, value);
}

private static Cursor CursorOf(IReadOnlyBufferList<byte> list, byte value)
{
    ReadOnlySpan<byte> first = list.First.Span;
    int index = first.IndexOf(value);
    if (index != -1) return new Cursor(list.First, list.Rest, index);
    if (list.Rest == null) return default;
    return CursorOf(list.Rest, value);
}

public readonly struct Cursor
{
    internal readonly ReadOnlyMemory<byte> First;
    internal readonly IReadOnlyBufferList<byte> Rest;
    internal readonly int Index;

    public Cursor(ReadOnlyMemory<byte> node, IReadOnlyBufferList<byte> rest, int index) { }
}

public ReadOnlyBytes Slice(Cursor from)
{
    if (_first.Equals(from.First)) return Slice(from.Index);
    
    IReadOnlyBufferList current = _rest;
    while (true)
    {
        if (current == null) break;
        if (current.First.Equals(from.First)) return new ReadOnlyMemory(current.First.Slice(from.Index), current.Rest);
        current = _rest.Rest;
    }

    current = from.Rest;
    while (true)
    {
        if (current == null) throw new InvalidOperationException();
        if (current.First.Equals(_first)) return new ReadOnlyMemory(from.First.Slice(from.Index), from.Rest);
        current = _rest.Rest;
    }
}
```

pakrym:
@YohDeadfall there are better ways of implementing it but yes

KrzysztofCwalina:
I can change the design such that when ROB is multisegmented, even the first segment is in the list. I think it will fix the issues.

KrzysztofCwalina:
@pakrym, I made the change. I added bunch of tests for what I think are your scenarios, including the sample you provided above.

The perf results from HttpParserBench are still roughly the same, i.e. ROB parsing is 67% of RB parsing for TechEmpower plain-text request.

 Test Name                         | Metric                                        | Iterations |    AVERAGE 
:--------------------------------- |:--------------------------------------------- |:----------:| ----------:
 HttpParserBench.FullRequestRb     | Duration                                      |    100     |     16.658 
 HttpParserBench.FullRequestRb     | Cache Misses                                  |    100     |  27811.840 
 HttpParserBench.FullRequestRob    | Duration                                      |    100     |     11.070 
 HttpParserBench.FullRequestRob    | Cache Misses                                  |    100     |  19087.360

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -10,35 +10,39 @@ namespace System.Buffers
...
-            _first = first;
-            _totalLength = length;
+            _first = buffer;
+            _all = null;
+            _totalLength = _first.Length;
         }
 
-        public ReadOnlyBytes(ReadOnlyMemory<byte> first, IReadOnlyBufferList<byte> rest) :
-            this(first, rest, first.Length + (rest == null ? 0 : rest.Length)) { }
+        public ReadOnlyBytes(IReadOnlyBufferList<byte> segments, long length)
```
YohDeadfall: `length` is unused.

pakrym:
What abut `Slice(from, to)`? I'm interested in length calculation implementation there.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: Ah, thanks; copy and paste is evil :-)

pakrym:
Can you please change parser to return cursor for consumed and examined so they would be more comparable?

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
YohDeadfall: nit: probably `ReadOnlyBytes(IReadOnlyBufferList<byte>)` should be placed near `ReadOnlyBytes(IReadOnlyBufferList<byte>, long)` for maintainability.

KrzysztofCwalina:
@pakrym, I just pushed Slice(Cursor, Cursor).

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -142,19 +146,35 @@ public int IndexOf(byte value)
 
         int IndexOfRest(byte value, int firstLength)
         {
-            if (_rest == null) return -1;
-            var index = _rest.IndexOf(value);
+            if (Rest == null) return -1;
```
YohDeadfall: Wouldn't it be better to store the result of `Rest`? It isn't a const now.
```csharp
var rest = Rest;
if (rest == null) return -1;
var index = rest.IndexOf(value);
```

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: Yeah, will do.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: Yeah, there are many optimizations like this one that we could/should do. Though, I will do this one now.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -176,13 +213,13 @@ public int CopyTo(Span<byte> buffer)
                 return buffer.Length;
             }
             first.Span.CopyTo(buffer);
-            if (buffer.Length == firstLength || _rest == null) return firstLength;
-            return firstLength + _rest.CopyTo(buffer.Slice(firstLength));
+            if (buffer.Length == firstLength || Rest == null) return firstLength;
```
YohDeadfall: `Rest` isn't a const.

KrzysztofCwalina:
> Can you please change parser to return cursor for consumed and examined so they would be more comparable?

I will try to do it as a separate PR. 

Having said that, I would love to understand why we need the HTTP parser to deal with cursors at all. I think cursors might be useful for some extremely large ranges, but I think for HTTP (32KB max?) nothing will beat ints. 

In general, I wonder if we could make cursors optional when working with pipelines, i.e. there are both cursor and int based "Advance" APIs and in most cases people could/would use ints. In some very rare cases where they parse/consume very large payloads, they would use cursors.

Once we have such APIs, we could experimentally find out what's the threshold of size where cursors start paying off.

davidfowl:
> I will try to do it as a separate PR.

> Having said that, I would love to understand why we need the HTTP parser to deal with cursors at all. I think cursors might be useful for some extremely large ranges, but I think for HTTP (32KB max?) nothing will beat ints.
 
> In general, I wonder if we could make cursors optional when working with pipelines, i.e. there are both cursor and int based "Advance" APIs and in most cases people could/would use ints. In some very rare cases where they parse/consume very large payloads, they would use cursors.

> Once we have such APIs, we could experimentally find out what's the threshold of size where cursors start paying off.

If you can write something more complete that replacing a bunch of the existing pipelines samples it would be easier to discuss. The existing pattern didn't happen by mistake and I think if you converted more end to end examples it might help you understand the scenarios better.

YohDeadfall:
Is it okay that `Cursor` isn't fully absolute? :)

```csharp
var rob1 = new ReadOnlyBytes(new[] { 1, 2, 3, 4 });
var rob2 = new ReadOnlyBytes(new[] { 5, 6, 7, 8 });
var c = rob1.CursorOf(3);
var s = rob2.Slice(c);
```

KrzysztofCwalina:

PR #1903, opened 2017-11-04
Eliminated All Length Computations

KrzysztofCwalina:
Eliminated all length computations using the same "running index" trick used in ReadabaleBuffer.

cc: @pakrym, @davidfowl, @ahsonkhan, @joshfree

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/IReadOnlyBufferList.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/IReadOnlyBufferList.cs
@@ -12,6 +12,6 @@ namespace System.Buffers
 
         IReadOnlyBufferList<T> Rest { get; }
 
-        long Length { get; }
+        long Index { get; }
     }
 }
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -32,13 +32,8 @@ namespace System.Buffers
             _totalLength = length;
         }
 
-        public ReadOnlyBytes(IReadOnlyBufferList<byte> segments) :
-            this(segments, segments == null ? 0 : segments.Length)
-        { }
-
         private ReadOnlyBytes(ReadOnlyMemory<byte> first, IReadOnlyBufferList<byte> all, long length)
         {
-            Debug.Assert(all != null);
             // TODO: add assert that first overlaps all (once we have Overlap on Span)
             _first = first;
             _all = all;
@@ -89,6 +84,8 @@ namespace System.Buffers
 
         public long Length => _totalLength;
 
+        public long Index => _all == null ? 0 : _all.Index;
+
         public bool IsEmpty => _first.Length == 0 && _all == null;
 
         public static ReadOnlyBytes Empty => s_empty;
@@ -109,7 +106,7 @@ namespace System.Buffers
             {
                 var slice = first.Slice((int)index, (int)length);
                 if(slice.Length > 0) {
-                    return new ReadOnlyBytes(slice);
+                    return new ReadOnlyBytes(slice, _all, length);
                 }
                 return Empty;
             }
@@ -157,7 +154,11 @@ namespace System.Buffers
         public ReadOnlyBytes Slice(Cursor from)
         {
             if (from._node == null) return Slice(First.Length - from._index);
-            return new ReadOnlyBytes(from._node).Slice(from._index);
+            var headIndex = _all.Index + _all.First.Length - _first.Length;
+            var newHeadIndex = from._node.Index;
+            var diff = newHeadIndex - headIndex;
+            // TODO: this could be optimized to avoid the Slice
+            return new ReadOnlyBytes(from._node, Length - diff).Slice(from._index);
         }
 
         public ReadOnlyBytes Slice(Cursor from, Cursor to)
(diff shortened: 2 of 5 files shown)


src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -157,7 +154,11 @@ int IndexOfRest(byte value, int firstLength)
         public ReadOnlyBytes Slice(Cursor from)
         {
             if (from._node == null) return Slice(First.Length - from._index);
-            return new ReadOnlyBytes(from._node).Slice(from._index);
+            var headIndex = _all.Index + _all.First.Length - _first.Length;
```
ahsonkhan: Could this result in integer overflow?

long + int - int => overflow

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -169,10 +170,15 @@ public ReadOnlyBytes Slice(Cursor from, Cursor to)
                 return Slice(indexFrom, indexTo - indexFrom + 1);
             }
 
-            var fromIndex = from._node.Length - from._index;
-            var toIndex = to._node.Length - to._index;
-
-            return new ReadOnlyBytes(from._node).Slice(from._index, fromIndex - toIndex + 1);
+            var headIndex = _all.Index + _all.First.Length - _first.Length;
+            var newHeadIndex = from._node.Index + from._index;
+            var newEndIndex = to._node.Index + to._index;
```
ahsonkhan: Same here, all these integer additions could overflow.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -301,33 +331,17 @@ public bool TryGet(ref Position position, out ReadOnlyMemory<byte> item, bool ad
...
-            }
-
-            if (first == null) return Empty;
+            if (buffers.Length == 0) return Empty;
+            if (buffers.Length == 1) return new ReadOnlyBytes(buffers[0]);
 
-            if (first.Rest == null)
+            BufferListNode first = new BufferListNode(buffers[0]);
+            var last = first;
+            for(int i=1; i<buffers.Length; i++)
```
ahsonkhan: nit: spacing

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -301,33 +331,17 @@ public bool TryGet(ref Position position, out ReadOnlyMemory<byte> item, bool ad
...
             {
-                return new ReadOnlyBytes(first, first.First.Length);
-            }
-            else
-            {
-                return new ReadOnlyBytes(first);
+                last = last.Append(buffers[i]);
             }
+
+            return new ReadOnlyBytes(first, last.Index + last.Length);
```
ahsonkhan: possibility of overflow

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: you mean if index is approaching long.MaxValue?

ahsonkhan: Yes.

KrzysztofCwalina: I am pretty sure we would hit OOM before we allocate 4 billion linked list nodes :-)

src/System.Buffers.Experimental/System/Buffers/IReadOnlyBufferList.cs
```diff
@@ -12,6 +12,6 @@ public interface IReadOnlyBufferList<T> : ISequence<ReadOnlyMemory<T>>
 
         IReadOnlyBufferList<T> Rest { get; }
 
-        long Length { get; }
+        long Index { get; }
```
pakrym: Is this a useful public API to have?

pakrym:
Can we restrict sliced buffer so only it's length of data is visible via the `Rest`? It seems non intuitive and error prone that no matter what the slice is entire length of linked list would be accessible to consumer forcing them to write length tracking code all the time and producing a leaking abstraction.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -275,23 +305,23 @@ public bool TryGet(ref Position position, out ReadOnlyMemory<byte> item, bool ad
...
                     return true;
                 }
                 else if (position.ObjectPosition == null) { item = default; return false; }
 
                 var sequence = (BufferListNode)position.ObjectPosition;
-                item = sequence._first;
+                item = sequence._data;
                 if (advance)
                 {
                     if (position == Position.First)
```
YohDeadfall: What is the point to check `position` here? It's already compared to `Position.First` in the 306 line.

YohDeadfall: What is the point to check `position` here? It's already compared to `Position.First` in 306 line.
```csharp
position.ObjectPosition = sequence._next;
position.IntegerPosition++;
```

KrzysztofCwalina:
> Can we restrict sliced buffer so only it's length of data is visible via the Rest? 

Yes, if Rest returns ROB or allocates. At least I think these would be the implications.

pakrym:
@KrzysztofCwalina I think this is an important property of abstraction - to give limited access to a slice only. Allocating for Rest on other hand is not an option, it would be to expensive.

PR #1906, opened 2017-11-04
Update xunit.perf to newer version and update README

ahsonkhan:
We get capabilities to collect new metrics (and customize which metrics we want to see) with the latest package of xunit.performance.

Example:
> --perf:collect default+BranchMispredictions+CacheMisses+InstructionRetired+gcapi

cc @jorive, @KrzysztofCwalina

### DIFF
--- a/scripts/PerfHarness/README.md
+++ b/scripts/PerfHarness/README.md
@@ -4,27 +4,27 @@ This project contains a simple harness to run performance tests and measure the
 results.
 
 ## Run the Performance Tests
-**Pre-requisite:** The dotnet cli is available at the root in the dotnet directory. On a clean repo, build.cmd at the root installs the latest available dotnet.exe.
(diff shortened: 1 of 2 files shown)


KrzysztofCwalina:
Does it make sense to update the run.bat file too?

ahsonkhan:
> Does it make sense to update the run.bat file too?

Update the folder path to dotnetcli? That was done here: https://github.com/dotnet/corefxlab/pull/1872/files

Or are you suggesting adding the extra metrics to run.bat? If so, which ones would you like to see?

KrzysztofCwalina:
What are the default metrics we collect?

ahsonkhan:
> What are the default metrics we collect?

According to the README [here](https://github.com/Microsoft/xunit-performance), the default is Duration only (which is always enabled) along with any metrics specified in the test source.

> --perf:collect default
 Collect metrics specified in the test source code by using xUnit Performance API attributes

KrzysztofCwalina:
I added instructions retired and cache misses to run.bat in some other PR.
Thanks!

PR #1913, opened 2017-11-08
ROB now hides items past its length

KrzysztofCwalina:
ReadOnlyBytes stores both a linked list of buffers and total length. But it exposes Rest property which returns the linked list, regardless of its length. 

This change hides the ReadOnlyBytes.Rest property. The segments are accessed using the TryGet method, which will trim the last linked list segment and stop enumeration, if the enumeration moves past ReadOnlyBytes.Length.

cc: @ahsonkhan, @pakrym


scripts/PerfHarness/run.bat
```diff
@@ -1 +1 @@
-..\..\dotnetcli\dotnet.exe run -c Release -- --perf:outputdir results --perf:typenames %1
+..\..\dotnetcli\dotnet.exe run -c Release -- --perf:outputdir results --perf:typenames %1 --perf:collect default+CacheMisses+InstructionRetired
```
ahsonkhan: Do u care about allocations? If so, consider adding gcapi as well.

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -139,25 +139,28 @@ public static int IndexOf(this IReadOnlyBufferList<byte> sequence, ReadOnlySpan<
...
-            if (rest == null) return -1;
-
-            index = rest.IndexOf(value);
-            if (index != -1) return first.Length + index;
+            var p = Position.First;
+            int runningIndex = first.Length;
+            while(sequence.TryGet(ref p, out var memory, advance: true))
+            {
+                index = memory.Span.IndexOf(value);
+                if (index != -1) return index + runningIndex;
```
ahsonkhan: Could this cause integer overflow? `return index + runningIndex;`

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -370,35 +372,51 @@ public HttpParser(bool showErrorDetails)
...
-            var index = 0;        
-            var rest = buffer.Rest;
-            var span = buffer.First.Span;
-            var remaining = span.Length;
+            var index = 0;
             consumedBytes = 0;
-            while (true) {
-                fixed (byte* pBuffer = &span.DangerousGetPinnableReference()) {
-                    while (remaining > 0) {
+            var position = Position.First;
```
ahsonkhan: nit: use of var

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -440,52 +462,73 @@ public HttpParser(bool showErrorDetails)
...
+            }
+            else
+            {
+                ch1 = temp[0];
+                ch2 = temp[1];
             }
         }
 
+        // This is not needed, but I will leave it here for now, as we might need it later when optimizing
+        //private static unsafe void ReadNextTwoChars(byte* pBuffer, int remaining, int index, out int ch1, out int ch2, ReadOnlySpan<byte> next)
```
ahsonkhan: This is different from the original ReadNextTwoChars. Are the changes intentional?

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -306,23 +307,23 @@ public bool TryGet(ref Position position, out ReadOnlyMemory<byte> item, bool ad
...
                     item = _data;
-                    if (advance) { position.IntegerPosition++; position.ObjectPosition = _next; }
-                    return true;
+                    if (advance) {
+                        position.IntegerPosition++;
+                        position.ObjectPosition = _next;
+                    }
+                    return (!_data.IsEmpty || _next != null);
+                }
+                else if (position.ObjectPosition == null)
```
pakrym: Can we use `Cursor` instead of position? Seems to be the same concept.

pakrym:
Would be nice to have `End` cursor, it's often used in consumer code in kestrel especially for examined.

src/System.Buffers.Experimental/System/Buffers/IReadOnlyBufferList.cs
```diff
@@ -5,12 +5,15 @@
 
 namespace System.Buffers
 {
-    public interface IReadOnlyBufferList<T> : ISequence<ReadOnlyMemory<T>>
+    public interface IReadOnlyMemorySequence<T> : ISequence<ReadOnlyMemory<T>>
     {
         int CopyTo(Span<T> buffer);
```
pakrym: Is `CopyTo` required? Why can't it be implemented using `TryGet`?

scripts/PerfHarness/run.bat
KrzysztofCwalina: I was thinking about it, but many of our APIs simply don't allocate and so this stat shows zero

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
KrzysztofCwalina: good point. I will fix that.

src/System.Text.Http.Parser/HttpParser.cs
KrzysztofCwalina: will fix

src/System.Text.Http.Parser/HttpParser.cs
KrzysztofCwalina: Yeah, all the changes are intentional. I first changed this routine, then simplified at the callsite.

src/System.Buffers.Experimental/System/Buffers/IReadOnlyBufferList.cs
KrzysztofCwalina: This is both very convenient (for dealing with straddling segments) and faster than using TryGet.

pakrym: Why is it faster? First segment can be read from `Memory` and `TryGet` for other segments.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: Yeah, I was thinking about it, but there are two things to consider:
1. Curstors point to Ts, Position points to segments. If we combine them, people will have to keep track which Position/Cursor is which.
2. Position has (and must have) three fields, Cursors have two. If the combined type ends up with three fields, it will be a bit slower to pass around and will make types that contain it bigger.

None the less, very much something we should consider.

src/System.Buffers.Experimental/System/Buffers/IReadOnlyBufferList.cs
KrzysztofCwalina: Requires 3 calls and lots of passing of largish structs around.

KrzysztofCwalina:

KrzysztofCwalina:
> Would be nice to have End cursor, it's often used in consumer code in kestrel especially for examined.

Could you point me to code showing how End is used?

pakrym:
> Could you point me to code showing how End is used?

https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Core/Internal/Http/Http1MessageBody.cs
https://github.com/aspnet/KestrelHttpServer/blob/dev/src/Kestrel.Core/Internal/Http/HttpParser.cs

KrzysztofCwalina:
@pakrym, we talked about this in person, I thought you guys were changing the reader to assume all was examined if consumed is not advanced. If we make the change, the APIs will be simpler and there won't be a need for End.

pakrym:
@KrzysztofCwalina what about usages for `IndexOf`? What if you want to consume to the end?

KrzysztofCwalina:
@dotnet-build-bot test Innerloop OSX10.12 Release Build and Test

KrzysztofCwalina:
If consume, you keep advancing as you go till you are at end (and so you have the end cursor, or you have advanced using indexes).

pakrym:
@KrzysztofCwalina all API's would have to return cursors then, for example `CopyTo` would have to tell you where it stopped copying.

KrzysztofCwalina:
I don't understand why. In most cases these operations return very short ranges and for these int index/offset should be used to advance. I might be missing something but advancing 20 bytes using a cursor seems like a lot of overhead. For example, I don't understand why parsing request line uses cursors. It consumes very small range and so it seems like it should return an int. The code that calls the method can advance using either cursors or ints. If it calls a method that return consumed int, it calls Advance(int). If it calls a method that returns a cursor, it calls Advance(Cursor). Would that not work?

pakrym:
No, we need a single way to do things or customer would have to merge cursors returned from one set of methods with int's returned from another set.

And how do you decide something would return a short range if it's not a fixed size method like `ReadInt`. Should `CopyTo` return `int` or cursor? What about `IndexOf`, `ParseStartLine`, `ParseHeaders`, `ParseJson`?

We would also need some methods to move cursor by the `int` amount.

KrzysztofCwalina:
Many, if not all APIs today use ints as the index/offset/length currency, so we need a way to translate ints to cursors anyway. And I don't think we can change BCL APIs to use cursors instead of ints. And so, CopyTo, and many of the other APIs you listed would return int, not cursor. 

This is actually at the root of my position that we should be int first (as most consuming/processing APIs are int based). In fact, your APIs (e.g. buffer reader) do exactly that: they use ints internally and only once in a while the code/consumer translates the int the reader advanced to cursor position.

pakrym:
@KrzysztofCwalina difference is that in `ReadableBufferReader`  you don't have to walk the list to return a cursor. My goal is to avoid walking linked list in most situations and having `Advance(int)` is doing just that.

KrzysztofCwalina:
> in ReadableBufferReader you don't have to walk the list to return a cursor.

Only because it was already walked (during processing), correct?

pakrym:
Yes, same as `CopyTo` on `IReadOnlyMemorySequence` or `CursorOf` - we walk sequence there so based on your logic have to return the cursor to the point where we stopped.

KrzysztofCwalina:
Not sure where I said that all APIs that walk the sequence have to return cursors. 

CursorOf is a specific API for cases when the caller thinks they need a cursor
IReadOnlyMmeorySequence walks segments, not Ts, and so it uses "cursor"  (Position).
CopyTo is mainly used to stitch very small buffers at the boundary between two segments, so I definatelly don't propose it returns a cursor.

pakrym:
Rewriting parser to use cursors might help clarify things a little.

I don't think that re-walking segment list every time `Advance` is called is acceptable, we've seen large performance penalty when doing it.

KrzysztofCwalina:
I just added some perf tests that show that re-walking after ints are used does not seem to be a problem for short re walks like parsing headers and RL. I think the reason for it is that ints are super fast when doing small advances, the wins are so large in comparison to doing short advances using cursors, that the time to re walk later (sequentially) is not a big deal in comparison.

pakrym:
We've seen issues with large HTTP message bodies.

KrzysztofCwalina:
Can you provide info on the scenario? I would like to add a test and measure it.

PR #1916, opened 2017-11-08
Added HTTP parser benchmarks simulating parsing with cursors

KrzysztofCwalina:
Added benchmarks measuring the performance of calling int based ROB parser APIs and then translating the into to cursors. 

here are the perf results:

 Benchmark                       | Metric               | Unit  |    Average 
:------------------------------------- |:-------------------- |:-----:| ----------:
 HttpParserBench.FullRequestRb         | Duration             | msec  |     12.503 
 HttpParserBench.FullRequestRb         | Instructions Retired | count | 1.446E+008 
 HttpParserBench.FullRequestRb         | Cache Misses         | count |   7331.840 
 HttpParserBench.FullRequestRob        | Duration             | msec  |      8.324 
 HttpParserBench.FullRequestRob        | Instructions Retired | count | 1.058E+008 
 HttpParserBench.FullRequestRob        | Cache Misses         | count |   5529.600 
 HttpParserBench.FullRequestRobCursors | Duration             | msec  |      9.509 
 HttpParserBench.FullRequestRobCursors | Instructions Retired | count | 1.163E+008 
 HttpParserBench.FullRequestRobCursors | Cache Misses         | count |   6389.760 

cc: @ahsonkhan, @joshfree, @davidfowl, @pakrym 

Note, that since ROB does not store a pointer to the last segment, default(Cursor) means that the whole ROB was examined. We could change the int based parsers to return the index of where they stopped to parse in the case of failure (just like our transformation APIs) and then examined could return the cursor at the location where the parser ran out of data.

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
@@ -201,6 +201,27 @@ namespace System.Buffers
             return CursorOf(Rest, value);
         }
 
+        public Cursor CursorAt(int index)
+        {
+            ReadOnlySpan<byte> first = _first.Span;
+            int firstLength = first.Length;
+
+            if (index < firstLength)
+            {
+                if (_all == null)
+                {
+                    return new Cursor(null, firstLength - index);
+                }
+                else
+                {
+                    var allIndex = index + (_all.First.Length - firstLength);
+                    return new Cursor(_all, allIndex);
+                }
+            }
+            if (Rest == null) return default;
+            return CursorAt(Rest, index - firstLength);
+        }
+
         private static Cursor CursorOf(IReadOnlyMemoryList<byte> list, byte value)
         {
             ReadOnlySpan<byte> first = list.First.Span;
@@ -208,6 +229,19 @@ namespace System.Buffers
             if (index != -1) return new Cursor(list, index);
             if (list.Rest == null) return default;
             return CursorOf(list.Rest, value);
+        }
+
+        private static Cursor CursorAt(IReadOnlyMemoryList<byte> list, int index)
+        {
+            if (list == null) return default;
+            ReadOnlySpan<byte> first = list.First.Span;
+            int firstLength = first.Length;
+
+            if (index < firstLength)
+            {
+                return new Cursor(list, index);
+            }
+            return CursorAt(list.Rest, index - firstLength);
         }
 
         public int CopyTo(Span<byte> buffer)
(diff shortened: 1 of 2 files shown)


pakrym:
What does this show? It's obvious that creating a struct per iteration would not add considerable performance hit.

>default(Cursor) means that the whole ROB was examined
This isn't a good idea because of the same reason relative cursor wasn't, it's only useful in context of one ROB.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
```diff
@@ -201,6 +201,27 @@ public Cursor CursorOf(byte value)
...
+
+            if (index < firstLength)
+            {
+                if (_all == null)
+                {
+                    return new Cursor(null, firstLength - index);
+                }
+                else
+                {
+                    var allIndex = index + (_all.First.Length - firstLength);
```
ahsonkhan: nit: use of var

tests/Benchmarks/HttpParserBench.cs
```diff
@@ -269,6 +369,54 @@ static int FullRequestLegacy()
     }
 }
 
+static class HttpParserExtensions
```
ahsonkhan: move to separate file?

KrzysztofCwalina:
>What does this show? 

I thought you asked for perf results of parsing ROB but returning cursors.

>This isn't a good idea because of the same reason relative cursor wasn't, it's only useful in context of one ROB.

The caller of this method has the ROB.

src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs
KrzysztofCwalina: will fix

tests/Benchmarks/HttpParserBench.cs
KrzysztofCwalina: I kind of prefer to keep all Http parsing code in this file. All other benchmarks are in one file each.

pakrym:
@KrzysztofCwalina I said use (for things like IndexOf) not return.

The following would not work correctly:

``` C#
var rob = ...;
var slice = rob.Slice(100); // limit length
var jobj = ReadJson(slice, out var consumed);
var rest = rob.Slice(consumed); // get the rest of the buffer
```

KrzysztofCwalina:
I think it would work correctly, i.e. consumed is not relative.

pakrym:
@KrzysztofCwalina it would be if `ReadJson` would reach the end, right?

KrzysztofCwalina:
No, consumed always is updated, i.e. default is never returned.

PR #1917, opened 2017-11-08
Update many projects to netstandard1.3

GrabYourPitchforks:
I'll be making some changes to System.Text.Primitives that rely on `netstandard1.3` APIs, so proactively upgrading that project. This PR also upgrades all of the projects that directly or indirectly pull that in as a dependency.

### DIFF
--- a/src/System.Binary.Base64/System.Binary.Base64.csproj
+++ b/src/System.Binary.Base64/System.Binary.Base64.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>Non-allocating Base64 encoder and decoded</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
     <AllowUnsafeBlocks>True</AllowUnsafeBlocks>
     <PackageTags>.NET non-allocating Base64 encoder and decoder</PackageTags>
     <PackageIconUrl>http://go.microsoft.com/fwlink/?linkid=833199</PackageIconUrl>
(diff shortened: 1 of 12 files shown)


ahsonkhan:
cc @dotnet/corefxlab-contrib, @davidfowl

KrzysztofCwalina:
Why do you need 1.3?

GrabYourPitchforks:
I'll eventually need the EncoderFallback / DecoderFallback types when the conversion APIs come in. The fallback APIs are available beginning with 1.3. I'm also using FormattableString in a few places, and while a workaround exists right now, it seemed silly to implement the workaround if we'll need 1.3 eventually anyway.

I only upgraded System.Text.Primitives and the projects which reference it.

KrzysztofCwalina:
Would it be better to make the bump when it's actually needed so that we can better understand why we need it? In the PRs opened so far, it's not even clear we will endup using the existing fallback types.

GrabYourPitchforks:
There's an API coming with a signature along the lines of `OperationStatus ToUtf16(ReadOnlySpan<byte> utf8, Span<char> utf16, bool isFinalChunk, DecoderFallback fallback, out int numBytesConsumed, out int numCharsWritten)`. I don't have a PR for it yet because it's still under development. We can forego this by saying "we only allow replacement of invalid sequences by `U+FFFD`", but then we lose drop-in compatibility with `System.Text.Encoding`. I'm personally fine with this, but it caused some gnashing of teeth when I shopped it around internally.

However, _right now_ I'm trying to merge in code which uses `FormattableString` internally, which was introduced in netstandard 1.3. I can work around this for now by defining my own internal `FormattableString` type and letting the compiler work magic, but it felt weird to introduce a workaround which might not live longer than a few weeks.

GrabYourPitchforks:
Deleted previous comment. I have a pending PR that uses `Vector.Widen`, and this API isn't part of netstandard at all yet. I'll need to reconsider how to enable this, perhaps by cross-compiling?

GrabYourPitchforks:
Abandoning. We'll convert when we need it.

PR #1918, opened 2017-11-08
Add streaming UTF8 validator

GrabYourPitchforks:
Introduces a type that can perform streaming validation of UTF-8 byte sequences. This is intended to be a drop-in replacement for existing streaming validators such as the one that exists at https://github.com/dotnet/corefx/blob/5617d3996c8cb1c3d3332f6c03dd5f3bf785623f/src/System.Net.WebSockets/src/System/Net/WebSockets/ManagedWebSocket.cs#L1326. This type also demonstrates how some of the primitives (such as inspecting the first sequence or getting the index of the first invalid sequence) hanging off of `Utf8Utility` can be used.

The type is currently a mutable struct, 4 bytes in length. This should provide respectable performance on all platforms and avoids allocations. Alternative designs can be considered if it's too difficult to consume this type.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
+++ b/src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
@@ -1,0 +1,255 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Buffers.Binary;
+using System.Diagnostics;
+
+namespace System.Buffers.Text
+{
+    /// <summary>
+    /// Allows processing discontiguous buffers of UTF-8 data, letting the caller know whether
+    /// any data seen so far is invalid UTF-8. The caller should invoke <see cref="TryConsume(ReadOnlySpan{byte, bool})"/>
+    /// in a loop until all data has been consumed.
+    /// </summary>
+    /// <remarks>
+    /// This type is a mutable struct.
+    /// </remarks>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public struct Utf8Validator
+    {
+        /// <summary>
+        /// A sequence that represents a fresh no-partial-buffer validity checker.
+        /// </summary>
+        private const uint DefaultSequence = 0;
+
+        /// <summary>
+        /// A sequence that represents a validity checker which has seen invalid data.
+        /// </summary>
+        private const uint InvalidSequence = unchecked((uint)~0);
+
+        // Packed data that contains both the partial sequence seen and the expected number of bytes remaining.
+        // PS1B: partial sequence first byte, etc.
+        // LEN: BYTE stating how many bytes (0 .. 3) have been read so far into the partial sequence.
+        // If the high bit of LEN is set, an invalid sequence was seen earlier.
+        // Big-endian machine: [ PS1B, PS2B, PS3B, LEN ]
+        // Little-endian machine: [ PS3B, PS2B, PS1B, LEN ]
+        private uint _partialSequence;
+
+        private string DebuggerDisplay
+        {
+            get
+            {
+                if (_partialSequence == 0)
+                {
+                    return "Data VALID so far; no partial sequence consumed.";
+                }
+                else if (IsInvalid)
+                {
+                    return "Data INVALID.";
+                }
+                else
+                {
+                    switch ((byte)_partialSequence)
+                    {
+                        case 1:
+                            return (BitConverter.IsLittleEndian)
+                                ? $"Data VALID so far; partial sequence [ {(byte)(_partialSequence >> 8):X2} ] consumed."
+                                : $"Data VALID so far; partial sequence [ {(_partialSequence >> 24):X2} ] consumed.";
+                        case 2:
+                            return (BitConverter.IsLittleEndian)
+                                ? $"Data VALID so far; partial sequence [ {(byte)(_partialSequence >> 8):X2} {(byte)(_partialSequence >> 16):X2} ] consumed."
+                                : $"Data VALID so far; partial sequence [ {(_partialSequence >> 24):X2} {(_partialSequence >> 16):X2} ] consumed.";
+
+                        case 3:
+                            return (BitConverter.IsLittleEndian)
+                                ? $"Data VALID so far; partial sequence [ {(byte)(_partialSequence >> 8):X2} {(byte)(_partialSequence >> 16):X2} {(_partialSequence >> 24):X2} ] consumed."
+                                : $"Data VALID so far; partial sequence [ {(_partialSequence >> 24):X2} {(_partialSequence >> 16):X2} {(byte)(_partialSequence >> 8):X2} ] consumed.";
+
+                        default:
+                            return "** INTERNAL ERROR **";
+                    }
+                }
+            }
+        }
+
+        private bool IsInvalid => IsInvalidPartialSequence(_partialSequence);
+
+        private static uint ConsumeDataWithExistingPartialSequence(uint partialSequence, ReadOnlySpan<byte> utf8Bytes)
+        {
+            Debug.Assert(partialSequence != 0 && !IsInvalidPartialSequence(partialSequence));
+
+            int partialSequenceOriginalByteCount = (byte)partialSequence;
+
+            if (BitConverter.IsLittleEndian)
+            {
+                // When we turn this into a Span<byte>, want MSB to be the first byte of the partial sequence
+                partialSequence >>= 8;
+            }
+
+            // Copy as much data as we can from the input buffer to our partial sequence.
+
+            Span<byte> partialSequenceAsBytes = stackalloc byte[sizeof(uint)];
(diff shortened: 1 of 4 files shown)


src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
```diff
@@ -0,0 +1,252 @@
...
+
+            if (BitConverter.IsLittleEndian)
+            {
+                // When we turn this into a Span<byte>, want MSB to be the first byte of the partial sequence
+                partialSequence >>= 8;
+            }
+
+            // Copy as much data as we can from the input buffer to our partial sequence.
+
+            Span<byte> partialSequenceAsBytes = new Span<byte>(&partialSequence, sizeof(uint));
```
ahsonkhan: Can we use binary writer APIs instead of the unsafe constructor?
```C#
var span = new Span<byte>(new byte[4]);
WriteMachineEndian(span, value);
```

tests/System.Text.Primitives.Tests/Encoding/Utf8UtilTests.Validation.cs
```diff
@@ -274,7 +274,7 @@ private static void AssertIsInvalidTwoByteSequence(byte[] invalidSequence)
         {
             Assert.Equal(2, invalidSequence.Length);
 
-            byte[] KNOWN_GOOD_BYTES = DecodeHex(E_ACUTE);
+            byte[] KNOWN_GOOD_BYTES = TestHelper.DecodeHex(E_ACUTE);
```
ahsonkhan: nit: use camel casing

tests/System.Text.Primitives.Tests/Encoding/Utf8UtilTests.Validation.cs
```diff
@@ -296,7 +296,7 @@ private static void AssertIsInvalidThreeByteSequence(byte[] invalidSequence)
         {
             Assert.Equal(3, invalidSequence.Length);
 
-            byte[] KNOWN_GOOD_BYTES = DecodeHex(EURO_SYMBOL);
+            byte[] KNOWN_GOOD_BYTES = TestHelper.DecodeHex(EURO_SYMBOL);
```
ahsonkhan: nit for constants: use pascal casing

src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
```diff
@@ -0,0 +1,252 @@
...
+    {
+        /// <summary>
+        /// A sequence that represents a fresh no-partial-buffer validity checker.
+        /// </summary>
+        private const uint DefaultSequence = 0;
+
+        /// <summary>
+        /// A sequence that represents a validity checker which has seen invalid data.
+        /// </summary>
+        private const uint InvalidSequence = unchecked((uint)~0);
```
ahsonkhan: Why not just uint.MaxValue?

GrabYourPitchforks: `~0` says "give me an integer where every bit is set" more clearly than `uint.MaxValue` does.

src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
```diff
@@ -0,0 +1,252 @@
...
+
+                    Debug.Assert(1 <= numBytesConsumed && numBytesConsumed <= 3);
+
+                    // Put all partial data into the high 3 bytes, making room for us to
+                    // write the count of partial bytes in the buffer as the low byte.
+
+                    if (BitConverter.IsLittleEndian)
+                    {
+                        uint partialSequence = (uint)numBytesConsumed;
+                        utf8Bytes.Slice(0, numBytesConsumed).CopyTo(new Span<byte>(&partialSequence, sizeof(uint)).Slice(1));
```
ahsonkhan: Same here. This method could be safe if you use the binary APIs.

tests/System.Text.Primitives.Tests/Encoding/Utf8UtilTests.Validation.cs
GrabYourPitchforks: Existing unit tests are out of scope of this review. I'll handle it in a separate PR.

tests/System.Text.Primitives.Tests/Encoding/Utf8UtilTests.Validation.cs
GrabYourPitchforks: Existing unit tests are out of scope of this review. I'll handle it in a separate PR.

src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
GrabYourPitchforks: Fixed.

src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
GrabYourPitchforks: Fixed.

src/System.Binary.Base64/System.Binary.Base64.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>Non-allocating Base64 encoder and decoded</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFramework>netstandard1.3</TargetFramework>
```
KrzysztofCwalina: Why are we bumping the version?

src/System.Text.Primitives/System/Text/Encoders/Utf8Validator.cs
```diff
@@ -0,0 +1,255 @@
...
+    /// <summary>
+    /// Allows processing discontiguous buffers of UTF-8 data, letting the caller know whether
+    /// any data seen so far is invalid UTF-8. The caller should invoke <see cref="TryConsume(ReadOnlySpan{byte, bool})"/>
+    /// in a loop until all data has been consumed.
+    /// </summary>
+    /// <remarks>
+    /// This type is a mutable struct.
+    /// </remarks>
+    [DebuggerDisplay("{DebuggerDisplay,nq}")]
+    public struct Utf8Validator
```
KrzysztofCwalina: I feel like we have too many types. Developers will have to remember where each of the API lives. Could we have all Utf8 related methods in the Utf8 static class?

KrzysztofCwalina:
This is a different model than our other streaming APIs (i.e. it is a stateful multable struct). Why do we want two different models for APis operating on buffers? Why not stick to the current model, which allows for using [Pipe-like](https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs#L16) routines, and integrates well with [SpanWriter](https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Experimental/System/Buffers/SpanWriter.cs#L66), of example?

src/System.Binary.Base64/System.Binary.Base64.csproj
GrabYourPitchforks: This is from a separate PR. I'm working on that PR as a base instead of master.

GrabYourPitchforks:
@KrzysztofCwalina this doesn't produce any output other than true / false since it's a validation routine, not a conversion routine. I didn't see any existing API pattern that seemed applicable. Did I miss something?

KrzysztofCwalina:
>  this doesn't produce any output other than true / false
Yeah, I know they dont produce any output, but still, today we are telling people our streaming APIs work as follows:
1. The operations are stateless
2. You keep calling them to do work, while stitching some buffers using stacalloc spans
This introduces a new pattern and I am not sure there is a storn reason to do it.

Also, if we chose to go with Utf8CodeUnit, then I would imagine Validate would be a transformation from Span<byte> to Span<Utf8CodeUnit>

GrabYourPitchforks:
I've been giving this some thought, and I still think we should march forward with these APIs as proposed. Yes, this is a domain-specific _stateful_ validation API, but all of the other methods on `Utf8Utility` are also domain-specific and don't have easily generalizable analogs. I see the `System.Text.Primitives` project as being the One True API Surface(tm) for all UTF-8 conversion, parsing, validation, and inspection routines. My ultimate goal is to get the rest of the framework (including the `WebSocket` type mentioned earlier) to remove their custom UTF-8 logic and to instead use the appropriate API from this set. By keeping all of the UTF-8 related logic consolidated to a single project, we have a single place where we can make performance improvements, and we have a single place where we can perform security review, signoff, and patching.

Nothing here prevents us from in the future including a stateless validation / conversion API that follows a prescribed generalized pattern. Currently we have no consumers of such an API, but we _do_ have potential consumers of a stateful validation API like the one being proposed here.

KrzysztofCwalina:
I think it's fine to merge it, once/if we have a representation of a validated UTF8, we can adjust.

GrabYourPitchforks:
@dotnet-bot test Innerloop OSX10.12 Release

GrabYourPitchforks:

PR #1919, opened 2017-11-08
Removing System.Binary (and tests) from solution

ahsonkhan:
System.Binary project (along with the test project) was removed here and hence removing it from Cloud.sln.
https://github.com/dotnet/corefxlab/pull/1883

cc @dotnet/corefxlab-contrib


KrzysztofCwalina:
looks good

PR #1922, opened 2017-11-09
Update Utf8.ToUtf16Length and Utf8.ToUtf32Length to use new workhorse APIs

GrabYourPitchforks:
Consolidate the UTF-8 validation and character counting workhorse logic.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Encoders/Utf8.cs
+++ b/src/System.Text.Primitives/System/Text/Encoders/Utf8.cs
@@ -48,296 +48,29 @@ namespace System.Buffers.Text
             /// <param name="source">A span containing a sequence of UTF-8 bytes.</param>
             /// <param name="bytesNeeded">On exit, contains the number of bytes required for encoding from the <paramref name="source"/>.</param>
             /// <returns>A <see cref="OperationStatus"/> value representing the expected state of the conversion.</returns>
-            public unsafe static OperationStatus ToUtf16Length(ReadOnlySpan<byte> source, out int bytesNeeded)
+            public static OperationStatus ToUtf16Length(ReadOnlySpan<byte> source, out int bytesNeeded)
             {
-                fixed (byte* pUtf8 = &source.DangerousGetPinnableReference())
+                if (Utf8Util.GetIndexOfFirstInvalidUtf8Sequence(source, out int scalarCount, out int surrogatePairCount) < 0)
                 {
-                    byte* pSrc = pUtf8;
-                    byte* pSrcEnd = pSrc + source.Length;
+                    // Well-formed UTF-8 string.
 
+                    // 'scalarCount + surrogatePairCount' is guaranteed not to overflow because
+                    // the UTF-16 representation of a string will never have a greater number of
+                    // of code units than its UTF-8 representation.
+                    int numCodeUnits = scalarCount + surrogatePairCount;
+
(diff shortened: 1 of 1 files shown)


KrzysztofCwalina:
What's the perf impact of this change? Especially the Utf8/Utf16 lenght computation. It will be used quite often I suspect.

GrabYourPitchforks:
@KrzysztofCwalina The perf results are better than the original implementations for almost all scenarios. See https://github.com/GrabYourPitchforks/fast-utf8/blob/master/FastUtf8Tester/results.md for some runs I did earlier. There are also perf tests checked in at https://github.com/dotnet/corefxlab/blob/master/tests/System.Text.Primitives.Tests/Encoding/Performance/Utf8UtilityPerfTests.cs.

The only two cases I could find which are _worse_ are where (a) the text jumps back and forth between 1-byte, 2-byte, 3-byte, and 4-byte sequences, which defeats the optimizations in the verifier; or (b) the text contains many errors, which causes the verifier to go down slow error-handling code paths. These cases should be very uncommon.

KrzysztofCwalina:
Sounds good. 
Looks good to me!

src/System.Text.Primitives/System/Text/Encoders/Utf8.cs
```diff
@@ -48,296 +48,24 @@ public static OperationStatus FromUtf16(ReadOnlySpan<byte> source, Span<byte> de
...
-                            ch &= 0xFF;
-#endif // BIGENDIAN
-                            pSrc++;
-                            if (ch <= 0x7F)
-                            {
-                                bytesNeeded++;
-                                continue;
-                            }
+                int indexOfFirstInvalidSequence = Utf8Util.GetIndexOfFirstInvalidUtf8Sequence(source, out int scalarCount, out int surrogatePairCount);
+                bytesNeeded = checked((scalarCount + surrogatePairCount) * 2 /* UTF-16 code units are 2 bytes in length */);
```
ahsonkhan: nit: move the comment outside the code (address outside this PR). same with ToUTf32Length

src/System.Text.Primitives/System/Text/Encoders/Utf8.cs
```diff
@@ -768,30 +496,24 @@ public static OperationStatus FromUtf32(ReadOnlySpan<byte> source, Span<byte> de
             /// <returns>A <see cref="OperationStatus"/> value representing the expected state of the conversion.</returns>
             public static OperationStatus ToUtf32Length(ReadOnlySpan<byte> source, out int bytesNeeded)
             {
-                bytesNeeded = 0;
+                int indexOfFirstInvalidSequence = Utf8Util.GetIndexOfFirstInvalidUtf8Sequence(source, out int scalarCount, out _);
+                bytesNeeded = checked(scalarCount * 4 /* UTF-32 code units are 4 bytes in length */);
```
ahsonkhan: Should we return bytesNeeded = 0 for InvalidData?

GrabYourPitchforks: Fine by me. I'll also change it to return `InvalidData` even if the data is incomplete (implicit _isFinalBlock=true_). Since there's no out parameter that corresponds to how much data was consumed, the caller cannot recover from an "incomplete data" return code, so it should be treated identically to error.

PR #1923, opened 2017-11-09
Improvements to Date and Time

mj1856:
Several improvements per code review with @mconnew.

### DIFF
--- a/src/System.Time/System/Date.cs
+++ b/src/System.Time/System/Date.cs
@@ -1487,15 +1487,18 @@ namespace System
 
         /// <summary>
         /// Casts a <see cref="DateTime"/> object to a <see cref="Date"/> by returning a new
-        /// <see cref="Date"/> object that has the equivalent year, month, and day components.  This is useful when
-        /// using APIs that express a calendar date as a <see cref="DateTime"/> and expect the consumer to ignore
-        /// the time portion of the value.  This operator enables these values to be assigned to a variable having
-        /// a <see cref="Date"/> type.  However, since the time and kind of the <see cref="DateTime"/> are ignored,
-        /// the cast is required to be applied explicitly.
+        /// <see cref="Date"/> object that has the equivalent year, month, and day components. This is useful when
+        /// using APIs that express a calendar date as a <see cref="DateTime"/> with time set at midnight (00:00:00).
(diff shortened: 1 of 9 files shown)


mj1856:

src/System.Time/System/Date.cs
```diff
@@ -1721,33 +1728,40 @@ private static string NormalizeDateFormat(string format)
             // standard formats
             if (format.Length == 1)
             {
-                // pass-through formats
-                if ("DdMmYy".Contains(format))
+                switch (format)
                 {
```
mconnew: This will actually cause format.GetHashCode to be called. You should probably switch on format[0] to get a char and have the case statements be for chars.

src/System.Time/System/Time.cs
```diff
@@ -1681,40 +1662,37 @@ private static string NormalizeTimeFormat(string format)
             // standard formats
             if (format.Length == 1)
             {
-                // pass-through formats
-                if ("Tt".Contains(format))
+                switch (format)
                 {
```
mconnew: Same comment as for Date. switch(format[0])

tests/System.Time.Tests/DateTimeTests.cs
```diff
@@ -17,13 +17,13 @@ public class DateTimeTests
         [Fact]
         public void CanAssignDateTimeTodayToDate()
         {
-            Date dt = DateTime.Today;
+            Date dt = (Date) DateTime.Today;
         }
```
mconnew: Should this verify that the cast did the right thing? I.e. the year, date and month match the values from DateTime.Today?

src/System.Time/System/Date.cs
mj1856: Fixed.  Thanks.

src/System.Time/System/Time.cs
mj1856: Fixed.  Thanks.

tests/System.Time.Tests/DateTimeTests.cs
mj1856: Yes.  Rewrote these and more in two new files, `DateCastTests.cs` and `TimeCastTests.cs`.  Also changed the casting behavior. Thanks.

PR #1926, opened 2017-11-10
Cleaned up projects 

KrzysztofCwalina:
- Removed wrappers for formatting primitives
- Removed old HTTP parser
- Renamed some files to be more consistent with conventions we use in corefxlab

@ahsonkhan, @joshfree, @GrabYourPitchforks


ahsonkhan:
A lot of the file names were changed to Utf6xxx instead of Utf16xxx. Please update.

ahsonkhan:
Hmm... Jenkins is down so this did not run/pass CI. Hopefully it builds fine :)

GrabYourPitchforks:
Belated LGTM :)

KrzysztofCwalina:
@ahsonkhan, I will fix the filenames in the upcoming PR. Thanks for catching it.

"Custom" prefix for the type name is something I feel we cannot ship with, and so I did not want to name the filed with a name that won't stick. Maybe the CustomParser should be called simply Parser?

PR #1927, opened 2017-11-10
Implement Tensor Slicing

eerhardt:
Implement a `Slice` method on Tensor<T> that returns a Tensor containing the subset of the data.  For example, you can slice out one or two columns out of a matrix, or a smaller matrix out of a larger matrix, etc.

The backing data isn't copied unless it has to be in order to fulfill the request.


src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1431,6 +1430,54 @@ public virtual CompressedSparseTensor<T> ToCompressedSparseTensor()
...
+        }
+
+        public Tensor<T> Slice(ReadOnlySpan<Range> ranges)
+        {
+            if (ranges.Length != Rank)
+            {
+                throw new ArgumentException($"{nameof(ranges.Length)} of {nameof(ranges)} ({ranges.Length}) must match the {nameof(Rank)} of this {nameof(Tensor)} ({Rank}).");
+            }
+
+            Span<int> newDimensions = Rank < ArrayUtilities.StackallocMax ? stackalloc int[Rank] : new Span<int>(new int[Rank]);
```
ahsonkhan: You should be able to rewrite this as follows (using implicit cast):
```C#
Span<int> newDimensions = Rank < ArrayUtilities.StackallocMax ? stackalloc int[Rank] : new int[Rank];
```

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1431,6 +1430,54 @@ public virtual CompressedSparseTensor<T> ToCompressedSparseTensor()
...
+
+            // construct new dimensions and strides, compute offset as the linearization of the lower bounds
+            for (int i = 0; i < ranges.Length; i++)
+            {
+                var range = ranges[i];
+
+                // Range guarantees the Length is positive
+                Debug.Assert(range.Length > 0);
+
+                offset += range.Index * strides[i];
```
ahsonkhan: potential integer overflow?

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
```diff
@@ -0,0 +1,146 @@
...
+
+                _backingTensor.SetValue(_offset + ArrayUtilities.GetIndex(_stridesIntoBackingTensor, indices), value);
+            }
+        }
+
+        public override T GetValue(int index)
+        {
+            EnsureValidIndex(index);
+
+            Span<int> indices = Rank < ArrayUtilities.StackallocMax ? stackalloc int[Rank] : new Span<int>(new int[Rank]);
```
ahsonkhan: use implicit cast here and elsewhere

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
```diff
@@ -0,0 +1,146 @@
...
+        private void EnsureValidIndices(ReadOnlySpan<int> indices)
+        {
+            if (indices.Length != Rank)
+            {
+                throw new ArgumentException($"{nameof(indices)} Length '{indices.Length}' must match Rank '{Rank}' of the current Tensor.", nameof(indices));
+            }
+
+            for (int i = 0; i < indices.Length; i++)
+            {
+                if (indices[i] < 0 || indices[i] >= Dimensions[i])
```
ahsonkhan: Store indiced[i] in a local?

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
```diff
@@ -0,0 +1,146 @@
...
+                {
+                    throw new ArgumentOutOfRangeException($"Index at position '{i}' must be non-negative and less than the Dimension '{Dimensions[i]}' at that position. The index value is '{indices[i]}'.", nameof(indices));
+                }
+            }
+        }
+
+        private void EnsureValidIndex(int index)
+        {
+            if (index < 0 || index >= Length)
```
ahsonkhan: Change to:
```C#
if ((uint)index >= (uint)Length)
```

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
juliusfriedman: @ahsonkhan, your input here is misleading. What core of what processor in what NUMA domain will the code in question execute on? I have not even looked at the code which provides the logic associated with the property referenced and I will ask how will such property as `ArrayUtilities.StackallocMax` help if the stack size has been intentionally modified above or below auch limits as provided by the property in question... yet again and as always we thank you for your input..

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
juliusfriedman: Thats the least of my worries given the loop above uses signed integers...

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
juliusfriedman: See above...

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
juliusfriedman: Depending on how the clr is configured, e.g. what provisions have been alloted for by virtue of the fact the runtime exists and can be configured to prevent such... signed integer math is just unsigned  math with larger numbers....

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
ahsonkhan: I don't understand what you mean. See the following for what I am suggesting: https://github.com/dotnet/corefxlab/pull/1927#discussion_r150354779

tests/System.Numerics.Tensors.Tests/TensorSliceTests.cs
```diff
@@ -0,0 +1,310 @@
...
+            //{{{  0, 1, 2, 3},         {{{ 0, 6,12,18},
+            //  {  4, 5, 6, 7},           { 2, 8,14,20},
+            //  {  8, 9,10,11}},          { 4,10,16,22}},
+            //
+            // {{ 12,13,14*,15*},        {{ 1, 7,13*,19*},
+            //  { 16,17,18*,19*},         { 3, 9,15*,21*},
+            //  { 20,21,22*,23*}}}        { 5,11,17*,23*}}}
+
+            // get the *'d elements above
+            var slice = tensor.Slice(new Range(1, 1), new Range(0, 3), new Range(2, 2));
```
Anipik: shouldn't it be 
var slice = tensor.Slice(new Range(1, 1), new Range(0, 3), new Range(2, 4));

tests/System.Numerics.Tensors.Tests/TensorSliceTests.cs
```diff
@@ -0,0 +1,310 @@
...
+        {
+            var slice = Get3DSlice(constructor);
+            var diag = slice.GetDiagonal();
+
+            var expected = new[,]
+            {
+                { 2, 3 },
+                { 18, 19 },
+            };
+            Assert.Equal(true, StructuralComparisons.StructuralEqualityComparer.Equals(diag, expected));
```
Anipik: Can use assert.True()

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
juliusfriedman: Obviously or maybe only to me, it seems you do not understand the locality of the execution of the code which you are suggesting be utilized for some improvement when in reality it may actually not be an improvement and furthermore may introduce bugs; finally - for what I see at the link or lack thereof it will not chnage the nature of my statements which remain true.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1431,6 +1430,54 @@ public virtual CompressedSparseTensor<T> ToCompressedSparseTensor()
...
+
+            return Slice(newDimensions.Slice(0, newDimensionCount), newStrides.Slice(0, newDimensionCount), offset);
+        }
+
+        protected virtual Tensor<T> Slice(ReadOnlySpan<int> dimensions, ReadOnlySpan<int> strides, int offset)
+        {
+            return new TensorView<T>(this, dimensions, strides, offset);
+        }
+
+        #endregion
```
Anipik: where are we checking that range.index + range.length is not exceeding the value of that dimension ?

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
eerhardt: Fixed in all places I could find.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
eerhardt: I've added a `checked` block to this line.

tests/System.Numerics.Tensors.Tests/TensorSliceTests.cs
eerhardt: Fixed.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
eerhardt: Good catch.  I'll add the checks.

src/System.Numerics.Tensors/System/Numerics/TensorView.cs
eerhardt: Fixed.

KrzysztofCwalina:
Yeah, I needed to change Range to longs because of pipeline buffers (multi segmented) being longer than an int. Possibly this is not a good change and maybe we just don't use Range in pipeline scenarios. I worry about long based Range being to large (128 bits)

eerhardt:
@KrzysztofCwalina - what do you want me to do to resolve this?  I can sync up and change all my usages to long? .... Or I can revert your change?

eerhardt:
@KrzysztofCwalina - I added some casts to make the code start working again.

PR #1933, opened 2017-11-14
Made Position fields private

KrzysztofCwalina:
@terrajobst, @davidfowl


KrzysztofCwalina:

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
```diff
@@ -1,47 +1,64 @@
...
-        public int IntegerPosition;
-        public int Tag;
+        long _index;
+        object _obj;
 
-        public static readonly Position First = new Position();
-        public static readonly Position AfterLast = new Position() { IntegerPosition = int.MaxValue - 1 };
+        public static explicit operator int(Position position) => (int)position._index;
+        public static implicit operator long(Position position) => position._index;
+        public T As<T>() => _obj==null||IsInfinity?default:(T)_obj;
```
YohDeadfall: nit: spaces

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/ReadWriteBytes.cs
```diff
@@ -42,37 +42,36 @@ public ReadWriteBytes(Memory<byte> first, IMemorySequence<byte> rest, long lengt
 
         public bool TryGet(ref Position position, out Memory<byte> value, bool advance = true)
         {
-            if (position == Position.First)
+            if (position.IsDefault)
```
YohDeadfall: nit: why not `position == default` like everywhere?

src/System.IO.Pipelines.Extensions/ReadableBufferSequence.cs
```diff
@@ -16,41 +16,33 @@ public ReadableBufferSequence(ReadableBuffer buffer) : this()
...
-                        {
-                            position = Position.AfterLast;
-                        }
+                        position.Set(_buffer.Start.Segment.Next);
                     }
                 }
                 return true;
             }
-            else if (position == Position.AfterLast)
+            else if (position == Position.Infinity)
```
YohDeadfall: If you made `Position.Infinity` why did you removed `First`?

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
```diff
@@ -1,47 +1,64 @@
...
+        public static Position operator +(Position position, int value) => new Position(position._obj, position._index + value);
+        public static Position operator +(Position position, long value) => new Position(position._obj, position._index + value);
+
+        public static Position operator -(Position position, int value) => new Position(position._obj, position._index - value);
+        public static Position operator -(Position position, long value) => new Position(position._obj, position._index - value);
+
+        public static bool operator ==(Position left, Position right) => left._index == right._index && left._obj == right._obj;
+        public static bool operator !=(Position left, Position right) => left._index != right._index || left._obj != right._obj;
+
+        public void Set(int value) => _index = value;
```
YohDeadfall: `int` is implicitly convertible to `long`, so there are no reasons to have `Set(int)`.

samples/LowAllocationWebServer/LowAllocationWebServerLibrary/ReadWriteBytes.cs
KrzysztofCwalina: Will fix. Thanks,

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
KrzysztofCwalina: Will fix. Thanks.

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
KrzysztofCwalina: Will remove.

src/System.IO.Pipelines.Extensions/ReadableBufferSequence.cs
KrzysztofCwalina: First is default. Infinity is a special value.

YohDeadfall: `DateTime.MinValue` is default too, but it exists because it's a well known value. Also such constants allow you to have uniform code style. Just compare:
```csharp
if (position = default) { }
else if (position = Position.Ininity) { }
// and
if (position = Position.First) { }
else if (position = Position.Ininity) { }
```

KrzysztofCwalina: Yeah, there is a trade off here. I used to have Position.First for the reason you stated, but then the code was not consistent. In some places people were using First and in some default. Then on the other side people have to worry if First == default or if they need to do two checks. Removal of First makes it clear.

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
KrzysztofCwalina: Actually, I do need the int overload. Otherwise Set<T> is called on int argument.

KrzysztofCwalina:

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
YohDeadfall: Maybe `T` should be limited by the `class` constraint? The value is stored as `object` so that limitation will prevent implicit boxing inside `Set<T>`. As a bonus - performance optimization because `EqualityComparer<T>.Default.Equals(value, default)` will be replaced by `value == null`.

KrzysztofCwalina: Yeah, I like the idea. Thanks.

PR #1935, opened 2017-11-15
Finished Position Cleanup

KrzysztofCwalina:
Common usage examples:

1. Calling ISequence\<T\>
```c#
ISequence<T> collection = ...

Position position = default;
while (collection.TryGet(ref position, out int item)) {
    Console.WriteLine(item);
}
```
2. Array-based data structure implementation, e.g. [ResizableArray\<T\>](https://github.com/dotnet/corefxlab/blob/master/src/System.Collections.Sequences/System/Collections/Sequences/ResizableArray.cs#L107)
```c#
public bool TryGet(ref Position position, out T item, bool advance = true)
{
    int index = (int)position;
    if (index < _count) {
        item = _array[index];
        if (advance) { position += 1; }
        return true;
    }

    item = default;
    position = Position.End;
    return false;
}
```

3. Linked List implementation, e.g. [LinkedContainer\<T\>](https://github.com/dotnet/corefxlab/blob/master/tests/System.Collections.Sequences.Tests/SampleCollections/LinkedContainer.cs#L29)
```c#

public bool TryGet(ref Position position, out T item, bool advance = true)
{
    if(_count == 0)
    {
        item = default;
        return false;
    }

    if (position == default)
    {
        item = _head._item;
        if (advance) position.SetItem(_head._next);
        return _count > 0;
    }

    var node = position.GetItem<Node>();
    if (node == null) {
        item = default;
        position = Position.End;
        return false;
    }

    item = node._item;
    if (advance) { position.SetItem(node._next); }
    return true;
}
````
4. Buffer sequence with running index implementation (e.g. [ReadOnlyBytes](https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Experimental/System/Buffers/ReadOnlyBytes.cs#L44)) 
```c#
public bool TryGet(ref Position position, out ReadOnlyMemory<byte> value, bool advance = true)
{
    if (position == default)
    {
        value = _first;
        if (advance) position.Advance(Rest, _first.Length);
        return (!_first.IsEmpty || _all != null);
    }

    var (rest, runningIndex) = position.Get<IReadOnlyMemoryList<byte>>();
    if (rest == null)
    {
        value = default;
        return false;
    }

    value = rest.First;
    // We need to slice off the last segment based on length of this. 
    // ReadOnlyBytes is a potentially shorted view over a longer buffer list.
    if (value.Length + runningIndex > _totalLength)
    {
        value = value.Slice(0, (int)(_totalLength - runningIndex));
        if (value.Length == 0) return false;
    }

    if (advance) position.Advance(rest.Rest, value.Length);
    return true;
}
```

cc: @ahsonkhan, @davidfowl, @halter73, @terrajobst, @GrabYourPitchforks, @YohDeadfall

### DIFF
--- a/src/System.Collections.Sequences/System.Collections.Sequences.csproj
+++ b/src/System.Collections.Sequences/System.Collections.Sequences.csproj
@@ -9,5 +9,8 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
+    <PackageReference Include="System.ValueTuple">
+      <Version>4.5.0-preview2-25707-02</Version>
+    </PackageReference>
   </ItemGroup>
 </Project>
--- a/src/System.Collections.Sequences/System.Collections.Sequences.csproj
+++ b/src/System.Collections.Sequences/System.Collections.Sequences.csproj
@@ -9,8 +9,6 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
-    <PackageReference Include="System.ValueTuple">
-      <Version>4.5.0-preview2-25707-02</Version>
-    </PackageReference>
+    <PackageReference Include="System.ValueTuple" Version="$(CoreFxStableVersion)" />
   </ItemGroup>
 </Project>
--- a/src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
+++ b/src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
@@ -10,38 +10,58 @@ namespace System.Collections.Sequences
     public struct Position : IEquatable<Position>
     {
         long _index;
-        object _obj;
+        object _item;
 
         public static explicit operator int(Position position) => (int)position._index;
+
         public static implicit operator long(Position position) => position._index;
-        public T As<T>() => _obj == null || IsInfinity ? default : (T)_obj;
-        public bool IsInfinity => this == Infinity;
 
-        public static readonly Position Infinity = new Position(new object(), long.MaxValue);
+        public long GetIndex() => _index;
 
-        public static Position operator +(Position position, long value) => new Position(position._obj, position._index + value);
+        public T GetItem<T>() => _item == null || IsEnd ? default : (T)_item;
 
-        public static Position operator -(Position position, long value) => new Position(position._obj, position._index - value);
+        public (T item, long index) Get<T>() => (GetItem<T>(), this);
 
-        public static bool operator ==(Position left, Position right) => left._index == right._index && left._obj == right._obj;
-        public static bool operator !=(Position left, Position right) => left._index != right._index || left._obj != right._obj;
-
-        public void Set(long value) => _index = value;
-
-        public void Set(int value) => _index = value;
+        public void SetIndex(long index) => _index = index;
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void Set<T>(T item)
+        public void SetItem<T>(T item) where T:class
         {
-            if(EqualityComparer<T>.Default.Equals(item, default)) this = Infinity;
-            else _obj = item;
+            if (item == null) this = End;
+            else _item = item;
         }
 
-        private Position(object obj, long index)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Set<T>(T item, long index) where T : class
         {
-            _obj = obj;
-            _index = index;
+            if (item == null) this = End;
+            else
+            {
+                _item = item;
+                _index = index;
+            }
         }
+
+        public void Advance<T>(T item, long offset) where T:class
+        {
+            if (item == null) this = End;
+            else
+            {
+                _item = item;
+                _index += offset;
+            }
+        }
+
+        public static Position operator +(Position position, long offset) => new Position(position._item, position._index + offset);
+
+        public static Position operator -(Position position, long offset) => new Position(position._item, position._index - offset);
+
+        public static readonly Position End = new Position(new object(), long.MaxValue);
+
+        public bool IsEnd => this == End;
+
+        public static bool operator ==(Position left, Position right) => left._index == right._index && left._item == right._item;
+        public static bool operator !=(Position left, Position right) => left._index != right._index || left._item != right._item;
 
         [EditorBrowsable(EditorBrowsableState.Never)]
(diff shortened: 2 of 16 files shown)


src/System.Collections.Sequences/System.Collections.Sequences.csproj
```diff
@@ -9,5 +9,8 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
+    <PackageReference Include="System.ValueTuple">
+      <Version>4.5.0-preview2-25707-02</Version>
```
ahsonkhan: Can you use CoreFxStableVersion (i.e. 4.3.0)?

If not, consider adding a property SystemValueTypeVersion similar to SystemMemoryVersion in dependencies.props and update the version to latest.
https://github.com/dotnet/corefxlab/blob/master/tools/dependencies.props#L6

We will then have to update the update-dependencies program to update the version whenever a new package becomes available:
https://github.com/dotnet/corefxlab/blob/master/scripts/update-dependencies/Program.cs#L98

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
```diff
@@ -10,39 +10,59 @@ namespace System.Collections.Sequences
...
-        public static bool operator ==(Position left, Position right) => left._index == right._index && left._obj == right._obj;
-        public static bool operator !=(Position left, Position right) => left._index != right._index || left._obj != right._obj;
+        public (T item, long index) Get<T>() => (GetItem<T>(), this);
 
-        public void Set(long value) => _index = value;
+        public void SetIndex(long index) => _index = index;
 
-        public void Set(int value) => _index = value;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void SetItem<T>(T item) where T:class
```
ahsonkhan: nit: spacing

Also, why do we need to constraint T to a class here/elsewhere?

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
```diff
@@ -10,39 +10,59 @@ namespace System.Collections.Sequences
...
+            if (item == null) this = End;
+            else
+            {
+                _item = item;
+                _index = index;
+            }
         }
 
-        private Position(object obj, long index)
+        public void Advance<T>(T item, long offset) where T:class
```
ahsonkhan: nit: same here, spacing

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
```diff
@@ -10,39 +10,59 @@ namespace System.Collections.Sequences
...
-            _index = index;
+            if (item == null) this = End;
+            else
+            {
+                _item = item;
+                _index += offset;
+            }
         }
 
+        public static Position operator +(Position position, long offset) => new Position(position._item, position._index + offset);
```
ahsonkhan: Should this be checked to catch OverflowException?

tests/System.Buffers.Experimental.Tests/ReadOnlyBytesTests.cs
```diff
@@ -372,8 +380,16 @@ public void ReadOnlyBytesEnumeration()
         public void EmptyReadOnlyBytesEnumeration()
         {
             var bytes = ReadOnlyBytes.Empty;
-            Position position = default;
-            Assert.False(bytes.TryGet(ref position, out ReadOnlyMemory<byte> segment));
+            {
```
ahsonkhan: What is the reason for the nested scopes? Is it to re-use `segment` as the variable name?

tests/System.Collections.Sequences.Tests/BasicUnitTests.cs
```diff
@@ -38,7 +46,7 @@ private static ArrayList<int> CreateArrayList(int[] array)
         [InlineData(new int[] { 1, 2, 3 })]
         public void LinkedContainer(int[] array)
         {
-            LinkedContainer<int> collection = CreateLinkedContainer(array);
+             LinkedContainer<int> collection = CreateLinkedContainer(array);
```
ahsonkhan: nit: remove extra space?

src/System.Collections.Sequences/System.Collections.Sequences.csproj
KrzysztofCwalina: will do

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
KrzysztofCwalina: I will fix the spacing. 
The constraint is here so that people don't accidentally box and so the null check inside the method can be efficient

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
KrzysztofCwalina: will fix

src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
KrzysztofCwalina: I am torn. I am not sure it adds enough value. If long index overflows, there will most certainly other check that will kick in. Also, if it overflows, it's just an int that then subsequently is to be used in data structures (e.g. arrays) that do checks. But I will think about it more.

tests/System.Buffers.Experimental.Tests/ReadOnlyBytesTests.cs
KrzysztofCwalina: The compiler was complaining about "segment" being redefined even if I don't define it again.

tests/System.Collections.Sequences.Tests/BasicUnitTests.cs
KrzysztofCwalina: will do

tests/System.Buffers.Experimental.Tests/ReadOnlyBytesTests.cs
ahsonkhan: That is strange.

Wouldn't something as follows work?
```C#
var bytes = ReadOnlyBytes.Empty;
Position position = default;
Assert.False(bytes.TryGet(ref position, out ReadOnlyMemory<byte> memory));
foreach (var segment in bytes)
{
    Assert.False(true); // As an aside, why is this needed? Can't we check some property on ReadOnlyBytes instead?
}
```

PR #1936, opened 2017-11-16
Prepared Position for usage in Pipelines

KrzysztofCwalina:
Started to think how Position would be used as ReadCursor and discovered a few missing APIs.
 
@davidfowl, @ahsonkhan, @halter73

### DIFF
--- a/src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
+++ b/src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
@@ -22,10 +22,8 @@ namespace System.Collections.Sequences
 
         public static implicit operator long(Position position) => position._index;
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public long GetIndexLong() => _index;
 
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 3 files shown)


src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
```diff
@@ -12,15 +11,31 @@ public struct Position : IEquatable<Position>
...
+            position.Set(item, index);
+            return position;
+        }
+
         public static explicit operator int(Position position) => (int)position._index;
 
         public static implicit operator long(Position position) => position._index;
 
-        public long GetIndex() => _index;
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ahsonkhan: Is it necessary to mark all these APIs with AggresiveInlining attribute? I would think the JIT would automatically inline them.

As per Andy, anything under 16 bytes of IL gets inlined.

cc @AndyAyersMS, @vancem

KrzysztofCwalina: Yeah, I will actually remove all the ones that are one liners.

PR #1944, opened 2017-11-17
Removed Rob.Cursors and Use Position Instead

KrzysztofCwalina:
- I removed ReadOnlyBytes.Cursor and now I am using Position
- Position now represents node + index within the node (always), which is exactly like ReadCursor
- Rationalized ReadOnlyBytes and ReadWriteBytes. These are just read-only and read-write views over MemoryList

The APIs now look like follows:
```c#
namespace System.Buffers {
    
    // logically it's a tuple of (node:T,index:int)
    public struct Position : IEquatable<Position> {
        public static readonly Position End;
        public int Index { get; set; }
        public bool IsEnd { get; }
        public static Position Create(int index);
        public static Position Create<T>(int index, T item) where T : class;
        public bool Equals(Position position);
        public override bool Equals(object obj);
        public ValueTuple<T, int> Get<T>();
        public override int GetHashCode();
        public T GetItem<T>();
        public static bool operator ==(Position left, Position right);
        public static explicit operator int (Position position);
        public static bool operator !=(Position left, Position right);
        public void Set<T>(T item, int index) where T : class;
        public void SetItem<T>(T item) where T : class;
        public override string ToString();
    }

    public interface ISequence<T> {
        bool TryGet(ref Position position, out T item, bool advance=true);
    }

    public interface IMemoryList<T> : ISequence<Memory<T>>, ISequence<ReadOnlyMemory<T>> {
        Memory<T> Memory { get; }
        IMemoryList<T> Rest { get; }
        long VirtualIndex { get; }
        int CopyTo(Span<T> buffer);
    }

    public class MemoryList : ISequence<Memory<byte>>, ISequence<ReadOnlyMemory<byte>> {
        public MemoryList(Memory<byte> bytes);
        public Memory<byte> Memory { get; }
        public IMemoryList<byte> Rest { get; }
        public long VirtualIndex { get; }
        public MemoryList Append(Memory<byte> bytes);
        public int CopyTo(Span<byte> buffer);
        public static ValueTuple<MemoryList, long> Create(params byte[][] buffers);
        public bool TryGet(ref Position position, out Memory<byte> item, bool advance=true);
        public bool TryGet(ref Position position, out ReadOnlyMemory<byte> item, bool advance=true);
    }
    
    public struct ReadOnlyBytes : ISequence<ReadOnlyMemory<byte>> {
        public ReadOnlyBytes(IMemoryList<byte> segments, long length);
        public ReadOnlyBytes(ReadOnlyMemory<byte> memory);
        public static ReadOnlyBytes Empty { get; }
        public ReadOnlyMemory<byte> First { get; }
        public bool IsEmpty { get; }
        public long Length { get; }
        public int CopyTo(Span<byte> buffer);
        public SequenceEnumerator<ReadOnlyMemory<byte>, ReadOnlyBytes> GetEnumerator();
        public long IndexOf(byte value);
        public long IndexOf(ReadOnlySpan<byte> bytes);
        public Position PositionAt(int index);
        public Position PositionOf(byte value);
        public ReadOnlyBytes Slice(Position from);
        public ReadOnlyBytes Slice(Position from, Position to);
        public ReadOnlyBytes Slice(long index);
        public ReadOnlyBytes Slice(long index, long length);
        public ReadOnlyBytes Slice(Range range);
        public bool TryGet(ref Position position, out ReadOnlyMemory<byte> value, bool advance=true);
    }

    public struct ReadWriteBytes : ISequence<Memory<byte>> {
        public ReadWriteBytes(IMemoryList<byte> segments, long length);
        public ReadWriteBytes(Memory<byte> memory);
        public static ReadWriteBytes Empty { get; }
        public Memory<byte> First { get; }
        public bool IsEmpty { get; }
        public long Length { get; }
        public int CopyTo(Span<byte> buffer);
        public SequenceEnumerator<Memory<byte>, ReadWriteBytes> GetEnumerator();
        public long IndexOf(byte value);
        public Position PositionAt(int index);
        public Position PositionOf(byte value);
        public ReadWriteBytes Slice(Position from);
        public ReadWriteBytes Slice(Position from, Position to);
        public ReadWriteBytes Slice(long index);
        public ReadWriteBytes Slice(long index, long length);
        public ReadWriteBytes Slice(Range range);
        public bool TryGet(ref Position position, out Memory<byte> value, bool advance=true);
    }
}
```

cc: @ahsonkhan, @pakrym, @davidfowl, @halter73, @GrabYourPitchforks, @joshfree, @ianhays, @terrajobst


src/System.Buffers.Experimental/System/Buffers/MemoryList.cs
```diff
@@ -0,0 +1,115 @@
...
+        public int CopyTo(Span<byte> buffer)
+        {
+            int copied = 0;
+            Position position = default;
+            var free = buffer;
+            while (TryGet(ref position, out ReadOnlyMemory<byte> segment, true))
+            {
+                if (segment.Length > free.Length)
+                {
+                    segment.Span.Slice(0, free.Length).CopyTo(free);
```
ahsonkhan: If we had a CopyTo method that returned how much it copied (rather than throwing if not enough space), this could be written more cleanly.

```C#
int copied = 0;
Position position = default;
var free = buffer;
while (TryGet(ref position, out ReadOnlyMemory<byte> segment, true))
{
  copied += segment.Span.CopyTo(free);
  free = buffer.Slice(copied);
  if (free.Length == 0) break;
}
```

@KrzysztofCwalina, does this code motivate the change to our span CopyTo API?

src/System.Buffers.Experimental/System/Buffers/MemoryList.cs
```diff
@@ -0,0 +1,115 @@
...
+
+            var sequence = position.GetItem<MemoryList>();
+            item = sequence._data;
+            if (advance) { position.SetItem(sequence._next); }
+            return true;
+        }
+
+        public static (MemoryList list, long length) Create(params byte[][] buffers)
+        {
+            if(buffers.Length == 0 || (buffers.Length == 1 && buffers[0].Length == 0))
```
ahsonkhan: nit: spacing

ahsonkhan: Can you explain this check? `(buffers.Length == 1 && buffers[0].Length == 0)`

What happens if buffers.Length > 1 but all buffers[i].Length == 0 for all i?

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public ReadWriteBytes Slice(long index, long length)
+        {
+            if (length == 0) return Empty;
+
+            if (_all == null)
+            {
+                if (index > _first.Length) throw new ArgumentOutOfRangeException(nameof(index));
```
ahsonkhan: Use throw helper?

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
```diff
@@ -0,0 +1,322 @@
...
+                if (length > _first.Length - index) throw new ArgumentOutOfRangeException(nameof(length));
```
ahsonkhan: What about index/length being negative?

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
```diff
@@ -0,0 +1,322 @@
...
+
+                return new ReadWriteBytes(_first.Slice((int)index, (int)length), (int)(_totalLengthOrOffset + index));
+            }
+
+            var first = First;
+            if (first.Length >= length + index)
```
ahsonkhan: Could `length + index` overflow?

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
```diff
@@ -0,0 +1,322 @@
...
+                return Slice(indexFrom, indexTo - indexFrom + 1);
+            }
+
+            var headIndex = _all.VirtualIndex + _all.Memory.Length - _first.Length;
+            var newHeadIndex = fromSegment.VirtualIndex + fromIndex;
+            var newEndIndex = toSegment.VirtualIndex + toIndex;
+            var slicedOffFront = newHeadIndex - headIndex;
+            var length = newEndIndex - newHeadIndex;
+            // TODO: this could be optimized to avoid the Slice
+            var a = new ReadWriteBytes(fromSegment, length + fromIndex);
```
ahsonkhan: nit: use more meaningful variable name

src/System.Buffers.Experimental/System/Buffers/MemoryList.cs
KrzysztofCwalina: Yes

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
```diff
@@ -0,0 +1,322 @@
...
+                    return Empty;
+                }
+                else
+                {
+                    throw new ArgumentOutOfRangeException("index or length");
+                }
+            }
+
+            // TODO (pri 2): this could be optimized
+            var rest = new ReadWriteBytes(Rest, length + index - _first.Length);
```
ahsonkhan: Do we need to validate the index/length here?

src/System.Buffers.Experimental/System/Buffers/MemoryList.cs
KrzysztofCwalina: this seems like the right spacing, i.e. it's the default VS editor formatting

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
KrzysztofCwalina: I have lots of optimizations to do once we settle on the design.

src/System.Buffers.Experimental/System/Buffers/MemoryList.cs
ahsonkhan: `if(buffers.Length == 0` -> `if (buffers.Length == 0`

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
KrzysztofCwalina: Memory ctor will do the check that.

... this is the never ending dilemma: how much double checking we want to do.

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
KrzysztofCwalina: Yes, and then Memory.Slice will fail and throw the right exception (coincidentally).

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
KrzysztofCwalina: will rename to "slice". Same in ROB

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
ahsonkhan: Fair point. Why do these checks then? Slice will throw if out of bounds.

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
KrzysztofCwalina: I added validation to Slice(int,int)

src/System.Buffers.Experimental/System/Buffers/ReadWriteBytes.cs
KrzysztofCwalina: I added the checks. It helps later in code. At some point we can optimize them out, but I will do it later if needed.

davidfowl:
>         public static ValueTuple<MemoryList, long> Create(params byte[][] buffers);

What's the deal with this API?

KrzysztofCwalina:
> What's the deal with this API?
It's not really an API. It's more of a helper for testing. I will think how to expose it in a cleaner way.

BTW, instead of using this helper you can do:
```c#
MemoryList list = new MemoryList(buffer1).Append(buffer2).Append(buffer3)....;
```

davidfowl:
I see. 

>        long VirtualIndex { get; }
What about this one?

PR #1950, opened 2017-11-20
Made IndexOf to be an extension method over ISequence

KrzysztofCwalina:
Last week, I wanted to add ```IndexOf``` extension method over ```ISequence<Memory<byte>>```. I was not able to do it based on the existing interface, and so I implemented the extension over ```IMemoryList```.

This PR changes the ```ISequence<T>``` interface to implementation of IndexOf extension (and probably all other things we need). The change adds a property called ```First``` to ```ISequence<T>```.  Without the change, the position representing the first segment was ```default(Position)```, and such default value did not have the item field set.

@ahsonkhan, @davidfowl, @halter73


src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
```diff
@@ -8,26 +8,37 @@ namespace System.Buffers
...
+        // If I make it an extensions method, the compiler complains Span<T> cannot
+        // be used as a type parameter.
+        public static long IndexOf<TSequence>(TSequence sequence, byte value) where TSequence : ISequence<ReadOnlyMemory<byte>>
         {
             Position position = default;
             int totalIndex = 0;
             while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> memory))
             {
-                var index = memory.Span.IndexOf(value);
+                var index = SpanExtensions.IndexOf(memory.Span, value);
```
ahsonkhan: SpanExtensions -> MemoryExtensions

src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
```diff
@@ -8,26 +8,37 @@ namespace System.Buffers
 {
     public static class SequenceExtensions
     {
-        public static long IndexOf(this ISequence<ReadOnlyMemory<byte>> sequence, byte value)
+        // TODO: this cannot be an extension method (as I would like iot to be).
```
davidfowl: typo: iot -> it

PR #1954, opened 2017-11-25
Fix cursor computation issue with empty segments

davidfowl:
- See https://github.com/aspnet/KestrelHttpServer/issues/2151 for more context on the original issue


src/System.IO.Pipelines/System/IO/Pipelines/ReadableBufferReader.cs
```diff
@@ -117,7 +112,11 @@ public void Skip(int length)
                     break;
                 }
 
-                length -= (_currentSpan.Length - _index);
+                var remaining = (_currentSpan.Length - _index);
+
+                _index += remaining;
```
pakrym: `_index += currentSpan.Length - _index;`

```
remaining -= _currentSpan.Length - _index;
_index = _currentSpan.Length;
```

PR #1956, opened 2017-11-27
Readonly pipe structs and remove lenght

pakrym:
Does not affect the throughput.

### DIFF
--- a/src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
+++ b/src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
@@ -39,8 +39,6 @@ namespace System.IO.Pipelines
                     break;
                 }
             }
-
-            _buffer.ClearCursors();
         }
 
     }
--- a/src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
+++ b/src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
@@ -39,6 +39,8 @@ namespace System.IO.Pipelines
                     break;
                 }
             }
+
+            _buffer = default;
         }
 
     }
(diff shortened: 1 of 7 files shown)


YohDeadfall:
Would be `ReadCursor` replaced by `Position` from sequences?

pakrym:
@YohDeadfall it's still being discussed.

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
```diff
@@ -39,8 +39,6 @@ public void Dispose()
                     break;
                 }
             }
-
-            _buffer.ClearCursors();
```
davidfowl: Unrelated?

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
```diff
@@ -9,24 +9,23 @@ namespace System.IO.Pipelines
...
         /// Length of the <see cref="ReadableBuffer"/> in bytes.
         /// </summary>
-        public long Length => BufferLength;
+        public long Length => BufferStart.GetLength(BufferEnd);
 
         /// <summary>
         /// Determines if the <see cref="ReadableBuffer"/> is empty.
         /// </summary>
-        public bool IsEmpty => BufferLength == 0;
+        public bool IsEmpty => Length == 0;
```
benaadams: Does removing actual length cause issue for `IsEmpty` as its now a calculation? 

Can it just check the first segment rather than full chain?

pakrym: We need to figure out some issues with cursors and switch to cursor comparison for `IsEmpty`

PR #1961, opened 2017-11-29
Cleaned up MemoryPool

KrzysztofCwalina:
Renamed BufferPool to MemoryPool\<T\>
Made the pool generic
Cleaned up file locations, and some naming

@davidfowl, what do you think about this as the abstraction?

It would be good to make pipelines MemoryPool internal (and just have a factory on Pipe). I think it could be done if we add callbacks to MemoryPool<T> for when memory is allocated. The callbacks would give you access to OwnedMemory<T>, but I think it's all you need, i.e. Length and native pointer.

cc: @ahsonkhan, @pakrym,


src/System.Buffers.Primitives/System/Buffers/OwnedArray.cs
```diff
@@ -19,7 +20,7 @@ public OwnedArray(int length)
 
         public OwnedArray(T[] array)
         {
-            if (array == null) BufferPrimitivesThrowHelper.ThrowArgumentNullException(nameof(array)); 
+            if (array == null) new ArgumentNullException(nameof(array));
```
ahsonkhan: nit: use throw helper here like in other places

src/System.IO.Pipelines/System/IO/Pipelines/PipeOptions.cs
```diff
@@ -31,6 +31,6 @@ public class PipeOptions
 
         public IScheduler ReaderScheduler { get; }
 
-        public BufferPool BufferPool { get; }
+        public MemoryPool BufferPool { get; }
```
pakrym: Rename property too

src/System.IO.Pipelines/System/IO/Pipelines/PipeOptions.cs
```diff
@@ -10,7 +10,7 @@ namespace System.IO.Pipelines
     public class PipeOptions
     {
         public PipeOptions(
-            BufferPool bufferPool,
+            MemoryPool bufferPool,
```
pakrym: Rename argument

src/System.IO.Pipelines/System/Buffers/MemoryPool.cs
```diff
@@ -78,6 +78,12 @@ public override OwnedMemory<byte> Rent(int size)
             return block;
         }
 
+        public override void Return(OwnedMemory<byte> buffer)
+        {
+            var block = buffer as MemoryPoolBlock;
+            if (block != null) Return(block);
```
pakrym: Throw?

pakrym: I don't think we need this public, only thing that returns is `OnZeroReferences` of `MemoryPoolBlock`

src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
```diff
@@ -3,23 +3,25 @@
...
-    public abstract class BufferPool : IDisposable
+    public abstract class MemoryPool<T> : IDisposable
     {
-        public static BufferPool Default => Internal.ManagedBufferPool.Shared;
+        public static MemoryPool<T> Default => Internal.ArrayMemoryPool<T>.Shared;
 
-        public abstract OwnedMemory<byte> Rent(int minimumBufferSize);
+        public abstract OwnedMemory<T> Rent(int minimumBufferSize);
+
+        public abstract void Return(OwnedMemory<T> buffer);
```
pakrym: Do we need `Return` at all if we have `Release` in `OwnedMemory<T>`?

davidfowl: Yea, this was a point of contention before.

src/System.Buffers.Primitives/System/Buffers/OwnedArray.cs
KrzysztofCwalina: will do

src/System.IO.Pipelines/System/IO/Pipelines/PipeOptions.cs
KrzysztofCwalina: will do

src/System.IO.Pipelines/System/IO/Pipelines/PipeOptions.cs
KrzysztofCwalina: will do

src/System.IO.Pipelines/System/Buffers/MemoryPool.cs
KrzysztofCwalina: Yeah, I was thinking about it. I will change to throw.

src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
KrzysztofCwalina: I added the method to allow pooling of the OwnedMemory instances.

KrzysztofCwalina:
Yeah, let's discuss. I added it to allow pooling OwnedMemory instances and for parity with other pools (e.g. ArrayPool). OwnedMemory instances are free to store the pool in them and call Return(this) when they are disposed. I.e. I don't see the negative of adding this method. Do you?

pakrym:
It's a mixed message from abstraction, do you as consumer have to `Return` or `Release` is enough?

benaadams:
> Yeah, let's discuss. I added it to allow pooling OwnedMemory instances and for parity with other pools (e.g. ArrayPool)

`ArrayPool<T>` the management is external and manual (as the array doesn't sidecar the pool); so you **can't** use a dispose pattern.

Should either use `Dispose` allowing a nice scoping with `using`, manual `Dispose` when longer lived; or manual `Return` but not both. 

Dispose is probably a better pattern to go for

KrzysztofCwalina:
OK, the arguments make sense. I will remove Return.

KrzysztofCwalina:

PR #1963, opened 2017-11-30
Pipelines Pool Abstraction

KrzysztofCwalina:
This is an experiment in making pipelines pooling into a true abstraction, i.e. pipelines use MemoryPool<byte> instead of MemoryPool concrete class. The only usage of the concrete class is to create the default instances.

@davidfowl, @pakrym, @ahsonkhan, any thoughts?

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
@@ -17,11 +17,10 @@ namespace System.Buffers.Internal
 
         public override int MaxBufferSize => 1024 * 1024 * 1024;
 
-        public override OwnedMemory<T> Rent(int minimumBufferSize = DefaultSize)
+        public override OwnedMemory<T> Rent(int minimumBufferSize = AnySize)
         {
-            if (minimumBufferSize == 0) minimumBufferSize = 32;
-            else if (minimumBufferSize == AnySize) minimumBufferSize = DefaultSize;
-            else if (minimumBufferSize > MaxBufferSize) throw new ArgumentOutOfRangeException(nameof(minimumBufferSize));
+            if (minimumBufferSize == AnySize) minimumBufferSize = DefaultSize;
+            else if (minimumBufferSize > MaxBufferSize || minimumBufferSize < 1) throw new ArgumentOutOfRangeException(nameof(minimumBufferSize));
             return new ArrayPoolMemory(minimumBufferSize);
         }
 
--- a/src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
@@ -14,8 +14,11 @@ namespace System.Buffers.Internal
 
         public static ArrayMemoryPool<T> Shared => s_shared;
 
+        public override int MaxBufferSize => 1024 * 1024 * 1024;
+
         public override OwnedMemory<T> Rent(int minimumBufferSize)
         {
+            if (minimumBufferSize > MaxBufferSize) throw new ArgumentOutOfRangeException(nameof(minimumBufferSize));
             return new ArrayPoolMemory(minimumBufferSize);
         }
 
--- a/src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
@@ -1,7 +1,6 @@
 // Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Runtime;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Threading;
@@ -10,15 +9,19 @@ namespace System.Buffers.Internal
 {
     internal sealed class ArrayMemoryPool<T> : MemoryPool<T>
     {
+        const int DefaultSize = 4096;
+
         readonly static ArrayMemoryPool<T> s_shared = new ArrayMemoryPool<T>();
 
         public static ArrayMemoryPool<T> Shared => s_shared;
 
         public override int MaxBufferSize => 1024 * 1024 * 1024;
 
-        public override OwnedMemory<T> Rent(int minimumBufferSize)
+        public override OwnedMemory<T> Rent(int minimumBufferSize = DefaultSize)
         {
-            if (minimumBufferSize > MaxBufferSize) throw new ArgumentOutOfRangeException(nameof(minimumBufferSize));
+            if (minimumBufferSize == 0) minimumBufferSize = 32;
+            else if (minimumBufferSize == AnySize) minimumBufferSize = DefaultSize;
+            else if (minimumBufferSize > MaxBufferSize) throw new ArgumentOutOfRangeException(nameof(minimumBufferSize));
             return new ArrayPoolMemory(minimumBufferSize);
         }
 
--- a/src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
@@ -9,6 +9,8 @@ namespace System.Buffers
 
         public abstract OwnedMemory<T> Rent(int minimumBufferSize);
 
+        public abstract int MaxBufferSize { get;  }
+
         public void Dispose()
         {
             Dispose(true);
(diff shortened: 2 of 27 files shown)


KrzysztofCwalina:

src/System.IO.Pipelines/System/Buffers/MemoryPoolSlab.cs
```diff
@@ -9,7 +9,7 @@ namespace System.Buffers
     /// Slab tracking object used by the byte buffer memory pool. A slab is a large allocation which is divided into smaller blocks. The
     /// individual blocks are then treated as independant array segments.
     /// </summary>
-    public class MemoryPoolSlab : IDisposable
+    internal class MemoryPoolSlab : OwnedMemory<byte>
```
davidfowl: Why is this an OwnedMemory?

KrzysztofCwalina: The allocation/deallocation callbacks need OwnedMemory. Since you call the callbacks when slabs are allocated/deallocated, I either needed to inherit slab from OwnedMemory or create a temp (throwaway) OwnedMemory when calling the callbacks. 

I also considered the callbacks to take something else than OwnedMemory (e.g. tuple of pointer/len), but I think the pool creates an illusion of OwnedMemory instances being the unit of allocation, and so OwnedMemory as a parameter of the callbacks seems to be consistent with the abstraction.

pakrym: Are callbacks only used in RIO sample, @davidfowl?

davidfowl: I'd rather remove the callbacks altogether and keep the slabs as is. Copy this pool as is into the RIO code and make this one internal.

davidfowl: @pakrym yes it is.

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
```diff
@@ -10,12 +9,19 @@ namespace System.Buffers.Internal
...
+        const int DefaultSize = 4096;
+
         readonly static ArrayMemoryPool<T> s_shared = new ArrayMemoryPool<T>();
 
         public static ArrayMemoryPool<T> Shared => s_shared;
 
-        public override OwnedMemory<T> Rent(int minimumBufferSize)
+        public override int MaxBufferSize => 1024 * 1024 * 1024;
+
+        public override OwnedMemory<T> Rent(int minimumBufferSize = DefaultSize)
```
pakrym: I don't think we should use default value here. Especially while having `AnySize` option.

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
```diff
@@ -10,12 +9,19 @@ namespace System.Buffers.Internal
...
         {
+            if (minimumBufferSize == 0) minimumBufferSize = 32;
+            else if (minimumBufferSize == AnySize) minimumBufferSize = DefaultSize;
```
pakrym: Why do we support both `AnySize` and `0`?

src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
```diff
@@ -5,9 +5,13 @@ namespace System.Buffers
...
     public abstract class MemoryPool<T> : IDisposable
     {
+        public const int AnySize = int.MinValue;
+
         public static MemoryPool<T> Default => Internal.ArrayMemoryPool<T>.Shared;
 
-        public abstract OwnedMemory<T> Rent(int minimumBufferSize);
+        public abstract OwnedMemory<T> Rent(int minimumBufferSize = AnySize);
+
+        public abstract int MaxBufferSize { get;  }
```
pakrym: Why is this required?

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
KrzysztofCwalina: Default and Any are different. Default is a const. Any is a sentinel. Any can return different sizes each time it's called. Default will return the same size.

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
KrzysztofCwalina: What would you prefer to happen when 0 is passed?

src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
KrzysztofCwalina: So callers can query it to know that Rent won't throw (like your slab pool does)

pakrym: How would they react to the knowledge?

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
pakrym: We need to decide what const are we using for `AnySize` and throw in all other non-valid cases.

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
pakrym: But it get's compiled into consumer code. So if implementation changes it would require a recompile to react.

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
KrzysztofCwalina: OK, I will change it to throw for zero

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
KrzysztofCwalina: I don't get it: who needs to react to what? If the calling code does not pass anything to Rent, it apparently does not care what size it gets back. 

Having said that, I will change it to Any to be consistent across all pools.

src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
pakrym: Usually renting code lives deep inside the library and won't be able to do anything except to throw, but pool already does throwing.

src/System.Buffers.Primitives/System/Buffers/Pooling/ArrayMemoryPool.cs
pakrym: Imagine we have pipes `MemoryPool` with block size `4096` and `minimumBufferSize` being the same. Then we ship a new version with smaller block size. While user code looks like it doesn't care about the renting size it would start failing until rebuilt against a newer version of the library.

src/System.Buffers.Primitives/System/Buffers/Pooling/MemoryPool.cs
KrzysztofCwalina: I don't agree with "and won't be able to do anything" part. 

Also, see https://github.com/dotnet/corefxlab/blob/010238eb489ccfb04a47774cc606653afb82464a/src/System.IO.Pipelines.Networking.Sockets/SocketConnection.cs#L356

pakrym: Okay, I'm fine with keeping it, maybe consumer would be able to adjust to block size.

pakrym: What if there is no limit? `int.MaxValue`? Maybe it's better to make it nullable to make it more explicit?

KrzysztofCwalina: int.Max value is the limit, as OwnedMemory<T> cannot be larger.

KrzysztofCwalina:
@pakrym, good to merge?

src/System.IO.Pipelines/System/Buffers/MemoryPoolSlab.cs
```diff
@@ -20,7 +20,7 @@ public class MemoryPoolSlab : IDisposable
         private byte[] _data;
 
         private bool _isActive;
-        internal Action<MemoryPoolSlab> _deallocationCallback;
+        internal Action<OwnedMemory<byte>> _deallocationCallback;
```
davidfowl: These can be removed now.

PR #1964, opened 2017-11-30
Added BytesReader.TryRead(out int)

KrzysztofCwalina:
I also added perf tests comparing BytesReader to raw Utf8Formatter and to ReadableBufferReader (raw APIs). All three are in the same ballpark, which is good! ... especially considering that BytesReader is doing an additional check that it important for parsing straddling buffers.

 Benchmarks.dll                                  | Metric               | Unit  | Average 
:----------------------------------------------- |:-------------------- |:-----:|----------:
 BytesReaderBench.ParseInt32BytesReader          | Duration             | msec  |      1.056 
 BytesReaderBench.ParseInt32BytesReader          | Instructions Retired | count | 1.495E+007 
 BytesReaderBench.ParseInt32BytesReader          | Cache Misses         | count |   1478.656 
 BytesReaderBench.ParseInt32ReadableBufferReader | Duration             | msec  |      1.235 
 BytesReaderBench.ParseInt32ReadableBufferReader | Instructions Retired | count | 1.637E+007 
 BytesReaderBench.ParseInt32ReadableBufferReader | Cache Misses         | count |   2240.512 
 BytesReaderBench.ParseInt32Utf8Parser           | Duration             | msec  |      0.881 
 BytesReaderBench.ParseInt32Utf8Parser           | Instructions Retired | count | 1.249E+007 
 BytesReaderBench.ParseInt32Utf8Parser           | Cache Misses         | count |   1196.032 

@ahsonkhan, @davidfowl, @jkotas, @vancem, @pakrym, @joshfree

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/Text/BytesReader.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/Text/BytesReader.cs
@@ -1,21 +1,13 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-using System.Buffers.Text;
+using System.Buffers.Binary;
 using System.Collections.Sequences;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace System.Buffers.Text
 {
-    public static class BytesReader
-    {
-        public static BytesReader<T> Create<T>(T sequence) where T : ISequence<ReadOnlyMemory<byte>>
-        {
-            return new BytesReader<T>(sequence);
-        }
-    }
-
     /// <summary>
     /// Used to read text from byte buffers
     /// </summary>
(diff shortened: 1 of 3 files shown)


src/System.Buffers.Experimental/System/Buffers/Text/BytesReader.cs
```diff
@@ -109,81 +93,129 @@ public PositionRange ReadRange(ReadOnlySpan<byte> delimiter)
...
         }
 
-        public bool TryParseUInt64(out ulong value)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public bool TryParse(out int value)
+        {
+            var unread = Unread;
+            if (Utf8Parser.TryParse(unread, out value, out int consumed))
+            {
+                if (unread.Length > consumed) {
```
ahsonkhan: Should this be `if (unread.Length >= consumed)`?

KrzysztofCwalina: No, the check is here to ensure that the parser does not "quit" early when there are more digits in the subsequent segment, i.e. [1, 2, 3] [4, x] should parse as 1234, not 123.

PR #1967, opened 2017-12-05
Support for arrays in ReadCursor

pakrym:


src/System.IO.Pipelines/System/IO/Pipelines/MemoryEnumerator.cs
```diff
@@ -8,46 +8,109 @@ namespace System.IO.Pipelines
...
+                return true;
+            }
+
+            PipelinesThrowHelper.ThrowNotSupportedException();
+            return default;
+        }
+
+        private void ThrowEndNotSeen()
+        {
+            throw new InvalidOperationException("Segments ended by end was never seen");
```
davidfowl: These error messages need some love

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -257,7 +257,8 @@ internal void Append(ReadableBuffer buffer)
 
             EnsureAlloc();
 
-            var clonedBegin = BufferSegment.Clone(buffer.Start, buffer.End, out BufferSegment clonedEnd);
+            // TODO: this has to support array buffers
+            var clonedBegin = BufferSegment.Clone(buffer.Start.GetSegment(), buffer.Start.Index, buffer.End.GetSegment(), buffer.End.Index, out BufferSegment clonedEnd);
```
davidfowl: What happens if the underlying item is an array? Does it throw? Should we have a skipped test for this?

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
```diff
@@ -39,9 +39,6 @@ public void Dispose()
                     break;
                 }
             }
-
-            _buffer = default;
```
davidfowl: Why undo this?

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -285,12 +335,27 @@ public override int GetHashCode()
...
+            {
+                return null;
+            }
+            if (Segment is BufferSegment bufferSegment)
+            {
+                return bufferSegment;
+            }
+
+            ThrowWrongType();
+            return default;
```
davidfowl: Null not good enough? 😬

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
```diff
@@ -2,13 +2,16 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
 using System.Text;
 
 namespace System.IO.Pipelines
 {
     /// <summary>
     /// Represents a buffer that can read a sequential series of bytes.
     /// </summary>
+    [DebuggerDisplay("")]
```
davidfowl: ?

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
```diff
@@ -60,9 +63,15 @@ internal ReadableBuffer(BufferSegment startSegment, int startIndex, BufferSegmen
             BufferEnd = new ReadCursor(endSegment, endIndex);
         }
 
+        internal ReadableBuffer(byte[] startSegment, int startIndex, int lenght)
```
davidfowl: Fix he length argument (everywhere)

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
```diff
@@ -60,9 +63,15 @@ internal ReadableBuffer(BufferSegment startSegment, int startIndex, BufferSegmen
...
+        {
+            BufferStart = new ReadCursor(startSegment, startIndex);
+            BufferEnd = new ReadCursor(startSegment, startIndex + lenght);
+        }
+
         private ReadableBuffer Clone(in ReadableBuffer buffer)
         {
-            var segmentHead = BufferSegment.Clone(buffer.BufferStart, buffer.BufferEnd, out var segmentTail);
+            var segmentHead = BufferSegment.Clone(buffer.BufferStart.GetSegment(), buffer.BufferStart.Index, buffer.BufferEnd.GetSegment(), buffer.BufferEnd.Index, out var segmentTail);
```
davidfowl: Throw if this is an array for now.

tests/System.IO.Pipelines.Tests/SeekTests.cs
```diff
@@ -330,45 +246,45 @@ public static IEnumerable<object[]> SeekIteratorLimitData
...
 
-                data.Add(new object[] { "hello, world", 'h', 'd', 'h' });
-                data.Add(new object[] { "hello, world", ' ', 'd', ' ' });
-                data.Add(new object[] { "hello, world", 'd', 'd', 'd' });
-                data.Add(new object[] { "hello, world", '!', 'd', -1 });
-                data.Add(new object[] { "hello, world", 'h', 'w', 'h' });
-                data.Add(new object[] { "hello, world", 'o', 'w', 'o' });
-                data.Add(new object[] { "hello, world", 'r', 'w', -1 });
-                data.Add(new object[] { "hello, world", 'd', 'w', -1 });
+                //data.Add(new object[] { "hello, world", 'h', 'd', 'h' });
```
davidfowl: What’s up with this?

tests/System.IO.Pipelines.Tests/TestBufferFactory.cs
```diff
@@ -0,0 +1,76 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.IO.Pipelines.Testing;
+using System.Linq;
+using System.Text;
+
+namespace System.IO.Pipelines.Tests
+{
+    internal abstract class TestBufferFactory
```
davidfowl: Add one that mimics the pipelines allocation pattern.

davidfowl: Adding an empty block at the end of a block with data

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
pakrym: It doesn't make sense. It's a struct there might be hundreds of copies cleaning one wouldn't magically return memory.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
pakrym: I use `default` when I want to convey "we don't care what's returned". I like the pattern and how it works for both reference and value types.

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
pakrym: I added implementation that clones.

tests/System.IO.Pipelines.Tests/SeekTests.cs
pakrym: Debugging leftovers will fix.

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
pakrym: Debugging leftovers will fix.

tests/System.IO.Pipelines.Tests/TestBufferFactory.cs
pakrym: `BytePerSegmentTestBufferFactory` has an empty block in the beginning, end, and between blocks with data.

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
pakrym: It throws. Pipe doesn't support non linked list buffers and there are not tests for this.

src/System.IO.Pipelines/System/IO/Pipelines/MemoryEnumerator.cs
pakrym: Copied from another file. We'll give them love in separate pass.

src/System.IO.Pipelines.Testing/BufferUtilities.cs
```diff
@@ -11,7 +11,7 @@ public class BufferUtilities
     {
         public static ReadableBuffer CreateBuffer(params byte[][] inputs)
         {
-            if (inputs == null || inputs.Length == 0)
```
halter73: Nit: indexoutofrange

src/System.IO.Pipelines/System/IO/Pipelines/MemoryEnumerator.cs
```diff
@@ -8,46 +8,114 @@ namespace System.IO.Pipelines
...
+
+                Current = bufferSegment.Memory.Slice(start, end - start);
+                _cursorSegment = bufferSegment;
+                _cursorStart = start;
+                return true;
+            }
+
+            if (segment is byte[] array)
+            {
+                Current = ((Memory<byte>)array).Slice(_startIndex, _endIndex - _startIndex);
```
halter73: Probably slightly more efficient to do `Current = new Memory<byte>(array, _startIndex, _endIndex - _startIndex)` so you only create one Memory object instead of two.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -178,29 +225,40 @@ internal bool TryGetBuffer(ReadCursor end, out Memory<byte> data)
...
-
-                data = Memory<byte>.Empty;
-                return false;
             }
-            else
+
+            if (segment is byte[] array)
             {
-                return TryGetBufferMultiBlock(end, out data);
+                data = ((Memory<byte>) array).Slice(Index, end.Index - Index);
```
halter73: Use ctor instead?

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
```diff
@@ -6,11 +6,11 @@ namespace System.IO.Pipelines
...
-    public struct PreservedBuffer : IDisposable
+    public readonly struct PreservedBuffer : IDisposable
     {
-        private ReadableBuffer _buffer;
+        private readonly ReadableBuffer _buffer;
 
-        internal PreservedBuffer(ref ReadableBuffer buffer)
+        internal PreservedBuffer(in ReadableBuffer buffer)
         {
             _buffer = buffer;
```
halter73: Should we assert that the ReadableBuffer is references a BufferSegment and not an Array?

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
```diff
@@ -25,23 +25,22 @@ internal PreservedBuffer(ref ReadableBuffer buffer)
...
 
-                if (returnSegment == returnEnd)
+                while (true)
                 {
-                    break;
+                    var returnSegment = returnStart;
+                    returnStart = returnStart?.Next;
+                    returnSegment?.ResetMemory();
+
+                    if (returnSegment == returnEnd)
```
halter73: do/while

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
```diff
@@ -25,23 +25,22 @@ internal PreservedBuffer(ref ReadableBuffer buffer)
...
+                    {
+                        break;
+                    }
                 }
             }
-
-            _buffer = default;
```
halter73: Why remove this?

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -17,12 +19,24 @@ internal ReadCursor(BufferSegment segment)
             Index = segment?.Start ?? 0;
         }
 
+        internal ReadCursor(object segment, int index)
```
halter73: Who calls this?

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -17,12 +19,24 @@ internal ReadCursor(BufferSegment segment)
...
 
+        internal ReadCursor(byte[] array, int index)
+        {
+            Segment = array;
+            Index = index;
+        }
+
         internal bool IsDefault => Segment == null;
 
         internal bool IsEnd
```
halter73: No longer used?

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -36,25 +50,38 @@ internal bool IsEnd
...
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private bool IsEndMultiSegment()
+        private void ThrowWrongType()
+        {
+            throw new InvalidOperationException();
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static bool IsEndMultiSegment(BufferSegment segment)
```
halter73: Also not used?

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -122,15 +156,18 @@ private ReadCursor SeekMultiSegment(long bytes, ReadCursor end, bool checkEndRea
...
                 // if end is not trusted
                 if (!foundResult)
                 {
-                    if (segmentPart.Length >= bytes)
+                    if (memory.Length >= bytes)
                     {
-                        result = new ReadCursor(segmentPart.Segment, segmentPart.Start + (int)bytes);
+                        result = enumerator.CreateCursor((int)bytes);
                         foundResult = true;
                         if (!checkEndReachable)
```
halter73: I don't even like checkEndReachable as an option. For those willing to sacrifice perf for a runtime correctness check, I would just add a `bool ReadCursor.CheckEndReachable(ReadCursor end)`.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -155,9 +192,19 @@ private ReadCursor SeekMultiSegment(long bytes, ReadCursor end, bool checkEndRea
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         internal void BoundsCheck(ReadCursor newCursor)
```
halter73: I don't think it makes sense to have a BoundCheck that only verifies the newCursor comes after.

Every time we bounds check, we should be able to come up with both a lower and upper bound. If we are bounds checking at all, it means we don't really trust the input. Having an API that allows you to specify one boundary encourages doing only a partial bounds check.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
halter73: As it stands, I think this never throws if `ReadCursor.IsDefault` since `Segment != end.Segment` and `enumerator.MoveNext()` would return false. I don't think we should even allow null Segments. It doesn't seem well supported to begin with.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -211,15 +269,15 @@ private bool TryGetBufferMultiBlock(ReadCursor end, out Memory<byte> data)
 
             while (true)
             {
-                var wasLastSegment = segment.Next == null || end.Segment == segment;
+                var wasLastSegment = segment.Next == null || end == segment;
```
halter73: if `segment.Next == null`, doesn't that mean end didn't follow start? Why not throw?

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -235,24 +293,17 @@ private bool TryGetBufferMultiBlock(ReadCursor end, out Memory<byte> data)
...
         {
-            if (IsEnd)
-            {
-                return "<end>";
-            }
-
             var sb = new StringBuilder();
-            Span<byte> span = Segment.Buffer.Span.Slice(Index, Segment.End - Index);
-            SpanLiteralExtensions.AppendAsLiteral(span, sb);
             return sb.ToString();
```
halter73: This could be optimized to `return "";` 😆

pakrym: Removed the method entirely.

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
pakrym: It's clearing the field inside a specific copy of a struct which doesn't guarantee anything

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
pakrym: For now, there is `Preserve` implementation for array-based readable buffers, it clones an array and while it doesn't specifically require freeing it can support the same pattern.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -17,12 +19,24 @@ internal ReadCursor(BufferSegment segment)
             Index = segment?.Start ?? 0;
         }
 
+        internal ReadCursor(object segment, int index)
+        {
+            Segment = segment;
+            Index = index;
+        }
+
         internal ReadCursor(BufferSegment segment, int index)
```
KrzysztofCwalina: Should this ctor be removed?

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -1,14 +1,16 @@
...
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Text;
 
 namespace System.IO.Pipelines
 {
+    [DebuggerDisplay("{Segment}[{Index}]")]
     public readonly struct ReadCursor : IEquatable<ReadCursor>
```
KrzysztofCwalina: Can we rename it to Position?

pakrym: We are getting there. Want to keep PR size reasonable.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -36,25 +50,38 @@ internal bool IsEnd
                 {
                     return true;
                 }
-                else if (Index < segment.End)
-                {
-                    return false;
-                }
-                else if (segment.Next == null)
+
+                if (segment is BufferSegment bufferSegment)
```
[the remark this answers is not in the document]
pakrym: Comming in next PR.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
pakrym: Done.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
pakrym: https://github.com/dotnet/corefxlab/issues/1969

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
pakrym: Removed.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
pakrym: Removed.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
pakrym: This logic needs to be merged with `MemoryEnumerator.MoveNext`

src/System.IO.Pipelines/System/IO/Pipelines/PreservedBuffer.cs
halter73: Of course not. I figured it was about referencing PreservedBuffer.Buffer after disposal. Better to nullref than allow it.

I wouldn't have mentioned it if it wasn't there before, but since you went out of your way to remove it.

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
KrzysztofCwalina: ok

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
```diff
@@ -225,13 +277,26 @@ public static ReadableBuffer Create(byte[] data, int offset, int length)
...
+
+            return new ReadableBuffer(data, 0, data.Length);
         }
 
         /// <summary>
-        /// Create a <see cref="ReadableBuffer"/> over an OwnedMemory.
+        /// Create a <see cref="ReadableBuffer"/> over an <see cref="OwnedMemory{Byte}"/>.
         /// </summary>
-        public static ReadableBuffer Create(OwnedMemory<byte> data, int offset, int length)
+        public static PreservedBuffer Create(OwnedMemory<byte> data, int offset, int length)
```
KrzysztofCwalina: Ultimately this type should not depend on PreservedBuffer

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
```diff
@@ -201,18 +221,50 @@ public BufferEnumerator GetEnumerator()
             return new BufferEnumerator(BufferStart, BufferEnd);
         }
 
+        public ReadCursor Move(ReadCursor cursor, long count)
```
KrzysztofCwalina: This method is strange. It's not clear if it offsets or moves to a new location. Also, it's strange that it moves cursor but it's an instance method on RB.

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
```diff
@@ -201,18 +221,50 @@ public BufferEnumerator GetEnumerator()
...
+            }
+            return cursor.Seek(count, BufferEnd, false);
+        }
+
         /// <summary>
-        /// Create a <see cref="ReadableBuffer"/> over an array.
+        /// Create a <see cref="ReadableBuffer"/> over an <see cref="IEnumerable{OwnedMemory{Byte}}"/>.
         /// </summary>
-        public static ReadableBuffer Create(byte[] data)
+        public static PreservedBuffer Create(IEnumerable<OwnedMemory<byte>> data)
```
KrzysztofCwalina: Can we move the test only APIs to some extensions library?

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
pakrym: It needs `End` cursor to check bounds, that's why it's on RB.

KrzysztofCwalina:
All my comments are just questions/suggestions. Feel free to merge without waiting on me. We can/should discuss further changes separately.

src/System.IO.Pipelines/System/IO/Pipelines/ReadableBuffer.cs
pakrym: It was an exercise in creating RB for any linked list without before we decided to unify on `IMemoryList`. Will remove.

pakrym:
Perf doesn't seem to regress on available benchmarks.
master:
```
                                          Method |     Mean |     Error |    StdDev |        Op/s | Allocated |
------------------------------------------------ |---------:|----------:|----------:|------------:|----------:|
                         ParseLiveAspNetTwoTasks | 370.2 ns | 25.304 ns | 37.874 ns | 2,700,929.5 |       1 B |
                           ParseLiveAspNetInline | 231.7 ns |  3.098 ns |  4.638 ns | 4,316,083.1 |       0 B |
                                 LongWriteInline | 892.1 ns | 11.574 ns | 17.324 ns | 1,120,967.2 |       0 B |
                          WritePlaintextResponse | 464.3 ns |  8.102 ns | 12.127 ns | 2,153,579.3 |       0 B |
                      WriteSpanPlaintextResponse | 569.6 ns |  7.436 ns | 11.130 ns | 1,755,508.7 |       0 B |
 WriteableBufferWriterWriteFastPlaintextResponse | 313.8 ns |  9.013 ns | 13.490 ns | 3,186,533.9 |       0 B |

             Method |     Mean |     Error |    StdDev |      Op/s | Allocated |
------------------ |---------:|----------:|----------:|----------:|----------:|
  MemoryEnumerator | 1.873 us | 0.0248 us | 0.0371 us | 534,041.0 |       0 B |
 SegmentEnumerator | 1.004 us | 0.0185 us | 0.0277 us | 995,809.6 |       0 B |
```

branch:

```
                                       Method |     Mean |     Error |    StdDev |        Op/s | Allocated |
------------------------------------------------ |---------:|----------:|----------:|------------:|----------:|
                         ParseLiveAspNetTwoTasks | 362.3 ns | 19.623 ns | 29.371 ns | 2,760,188.2 |       1 B |
                           ParseLiveAspNetInline | 233.0 ns |  2.435 ns |  3.644 ns | 4,292,164.0 |       0 B |
                                 LongWriteInline | 893.7 ns | 14.510 ns | 21.717 ns | 1,118,887.2 |       0 B |
                          WritePlaintextResponse | 460.7 ns |  9.494 ns | 14.210 ns | 2,170,670.4 |       0 B |
                      WriteSpanPlaintextResponse | 608.1 ns | 41.370 ns | 61.921 ns | 1,644,418.4 |       0 B |
 WriteableBufferWriterWriteFastPlaintextResponse | 308.0 ns |  6.278 ns |  9.396 ns | 3,246,316.2 |       0 B |

           Method |     Mean |     Error |    StdDev |      Op/s | Allocated |
----------------- |---------:|----------:|----------:|----------:|----------:|
 MemoryEnumerator | 2.010 us | 0.0398 us | 0.0596 us | 497,516.5 |       0 B |
```

pakrym:
:up::date:

src/System.IO.Pipelines/System/IO/Pipelines/ReadCursor.cs
```diff
@@ -178,29 +159,40 @@ internal bool TryGetBuffer(ReadCursor end, out Memory<byte> data)
...
         }
 
-        private bool TryGetBufferMultiBlock(ReadCursor end, out Memory<byte> data)
+        private static bool TryGetBufferMultiBlock(BufferSegment start, int startIndex, BufferSegment end, int endIndex, out Memory<byte> data)
         {
-            var segment = Segment;
-            var index = Index;
+            var segment = start;
 
             // Determine if we might attempt to copy data from segment.Next before
```
halter73: Nit: Remove this comment and any copies of it if there are any others.

davidfowl:
@pakrym have we taken this change into kestrel as yet? It would be good to change things more incrementally so we can more easily find regressions

pakrym:
@davidfowl we did as of yesterday.

pakrym:
@davidfowl and we see a massive perf hit of taking new corefx and corefxlab packages. I'll look into it today.

PR #1977, opened 2017-12-13
Added BufferWriter<TOutput>

KrzysztofCwalina:
- Added BufferWriter\<TOutput\> that can write to multi-segmented buffers
- Renamed single segment SpanWriter to BufferWriter
- Reorganized files (split SpanWriter APIs into several files based on types being written)

BufferWriter\<TOutput\> is similar to WritableBufferWriter but:
1. Added the ability to write IWritable, string, int (for now; just to illustrate)
2. Does not "flush" on every write. This is done to be able to write to cached Span<byte> without the need to get a new span after every Advance, which is what IOutput contract currently requires. I will experiment with adding optional "AutoFlush" bool property to make it easier to use the writer. Alternatively, we can change the IOutput contract to require that Advance does not create a new buffer, but this has some other negative implications.

@ahsonkhan, @pakrym,


src/System.Azure.Experimental/System/Azure/CosmosDbAuthorizationHeader.cs
```diff
@@ -20,29 +20,32 @@ public struct CosmosDbAuthorizationHeader : IBufferFormattable {
         public string Version;
         public DateTime Time;
 
+        static TransformationFormat s_toBase64 = new TransformationFormat(Base64Experimental.BytesToUtf8Encoder);
```
ahsonkhan: nit: add explicit `private` access modifier

src/System.Azure.Experimental/System/Azure/StorageAccessSignature.cs
```diff
@@ -6,29 +6,31 @@
...
 using System.Binary.Base64Experimental;
 using System.Buffers.Text;
+using System.Buffers;
 
 namespace System.Azure.Authentication
 {
     public static class StorageAccessSignature
     {
         static Utf8String s_emptyHeaders = (Utf8String)"\n\n\n\n\n\n\n\n\n\n\nx-ms-date:"; // this wont be needed once we have UTF8 literals
+        static TransformationFormat s_toBase64 = new TransformationFormat(Base64Experimental.BytesToUtf8Encoder);
```
ahsonkhan: nit: add explicit `private` access modifier, here and elsewhere

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_writable.cs
```diff
@@ -0,0 +1,122 @@
...
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+using System.Text.Utf8;
+
+namespace System.Buffers.Text
+{
+    public ref partial struct BufferWriter
+    {
+        #region Byte
+        public bool TryWriteBytes(byte[] bytes)
```
ahsonkhan: Do we need this API? Why can't we use implicit cast from array to ReadOnlySpan?

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_writable.cs
```diff
@@ -0,0 +1,122 @@
...
+
+        public void WriteBytes(ReadOnlyMemory<byte> bytes)
+            => WriteBytes(bytes.Span);
+
+        #endregion
+
+        #region IWritable
+        public bool TryWriteBytes<T>(T value, StandardFormat format) where T : IWritable
+        {
+            int written;
```
ahsonkhan: nit: inline declaration of out variable?

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
```diff
@@ -0,0 +1,200 @@
...
+using System.Text.Utf8;
+
+namespace System.Buffers.Text
+{
+    public ref partial struct BufferWriter
+    {
+        #region String
+        public bool TryWrite(string text)
+        {
+            if (!TryWrite(text, out var written)) return false;
```
ahsonkhan: nit: specify the type instead of using `var`

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
```diff
@@ -0,0 +1,200 @@
...
+            while (!TryWriteLine(text, format)) Resize();
+        }
+        #endregion
+
+        #region UTF8String
+        public bool TryWrite(Utf8Span text)
+        {
+            int written = text.Bytes.Length;
+            if (!text.Bytes.TryCopyTo(Free)) return false;
+            _written += written;
```
ahsonkhan: We don't need `written` local here. Why not just do:
```C#
_written += text.Bytes.Length;
```

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
```diff
@@ -0,0 +1,200 @@
...
+        public bool TryWriteLine(Utf8Span text)
+        {
+            if (!TryWriteLine(text, out var written)) return false;
+            _written += written;
+            return true;
+        }
+
+        public void WriteLine(Utf8Span text)
+        {
+            while (!TryWriteLine(text)) Resize();
```
ahsonkhan: Could this ever lead to a non-terminating loop?

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
```diff
@@ -0,0 +1,200 @@
...
+                }
+                free[0] = (byte)character;
+                _written++;
+            }
+            else
+            {
+                Span<byte> utf8Span = stackalloc byte[4];
+                Span<char> utf16Span = stackalloc char[1];
+                utf16Span[0] = character;
+                if(Encodings.Utf8.FromUtf16(utf16Span.AsBytes(), utf8Span, out int consumed, out int written) == OperationStatus.Done)
```
ahsonkhan: The call to AsBytes is likely too expensive here.

Can we use bitwise operations and get the two bytes that represent the character ourselves and change utf16Span to the following?
```C#
Span<byte> utf16ByteSpan = stackalloc byte[2];
```

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
```diff
@@ -0,0 +1,200 @@
...
+                        Resize();
+                    }
+                    _written += written;
+                }
+            }
+        }
+
+        private bool TryWrite(string text, out int written)
+        {
+            var status = Encodings.Utf16.ToUtf8(text.AsReadOnlySpan().AsBytes(), Free, out var consumed, out written);
```
ahsonkhan: nit: we don't use the `consumed` out parameter. Replace with `out _`

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
```diff
@@ -0,0 +1,200 @@
...
+
+        private bool TryWriteLine(Utf8Span text, out int written)
+        {
+            if (!text.Bytes.TryCopyTo(Free))
+            {
+                written = 0;
+                return false;
+            }
+            written = text.Bytes.Length;
+            if (!NewLine.TryCopyTo(Free.Slice(written))) return false;
```
ahsonkhan: Do we need to check that written is less than Free.Length? Otherwise the Slice could throw.

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_datetime.cs
```diff
@@ -0,0 +1,34 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Buffers.Text
+{
+    public ref partial struct BufferWriter
+    {
+        public bool TryWrite(DateTime date, StandardFormat format)
+        {
+            if (!Utf8Formatter.TryFormat(date, Free, out var bytesWritten, format)) return false;
```
ahsonkhan: nit: use explicit type instead of `var`

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
ahsonkhan: Also, do we want written to be set to text.Bytes.Length if this TryCopyTo fails or should it be 0?

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs
```diff
@@ -0,0 +1,67 @@
...
+            }
+        }
+
+        public void Clear()
+            => _written = 0;
+        
+        private void Resize(int desiredBufferSize = 0)
+        {
+            if (Enlarge == null) throw new BufferTooSmallException();
+            if (desiredBufferSize <= 0) desiredBufferSize = _buffer.Length * 2 + 1;
```
ahsonkhan: Do we add 1 here to make sure that if Length == 0, that we still increase the desiredBufferSize? If so, can you please add a comment.

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs
```diff
@@ -0,0 +1,67 @@
...
+            if (desiredBufferSize < _buffer.Length + 1) throw new ArgumentOutOfRangeException(nameof(desiredBufferSize));
```
ahsonkhan: Should this be an else if instead?
i.e.
```C#
if (desiredBufferSize <= 0) desiredBufferSize = _buffer.Length * 2 + 1;
else if (desiredBufferSize < _buffer.Length + 1) throw new ArgumentOutOfRangeException(nameof(desiredBufferSize));
```

This is because if` (desiredBufferSize <= 0)` is true, then `_buffer.Length * 2 + 1 < _buffer.Length + 1` will never be true.

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs
```diff
@@ -0,0 +1,67 @@
...
+        Span<byte> _buffer;
+        int _written;
+
+        public ReadOnlySpan<byte> NewLine;
+        public Func<int, Memory<byte>> Enlarge;
+
+        static byte[] s_defaultNewline = new byte[] { (byte)'\n' };
+
+        public static BufferWriter Create(Span<byte> buffer) => new BufferWriter(buffer);
+        public static BufferWriter<TOutput> Create<TOutput>(TOutput output)
```
ahsonkhan: nit: add new line between the two Create methods.

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_writable.cs
KrzysztofCwalina: For now I want to have this API as it came from WritableBufferWriter (with all its perf optimizations), but it's a good point about the overloads. I will measure perf and possibly remove/rename.

src/System.Azure.Experimental/System/Azure/CosmosDbAuthorizationHeader.cs
KrzysztofCwalina: will do

src/System.Azure.Experimental/System/Azure/StorageAccessSignature.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_writable.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
KrzysztofCwalina: will fix

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
KrzysztofCwalina: I dont think so, the Resize method throws when the buffer does not actually resize

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
KrzysztofCwalina: Yeah, I forgot to add a comment that this is a naive implementation. I think we need char encoding APIs on Utf8/Utf16 classes. But I dont think char fits in 2 bytes. I think it fits in 3, but I wanted to err on the safe side (especially for a implementation that will be optimized later)

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_datetime.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter_strings.cs
KrzysztofCwalina: yeah, good point.

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs
KrzysztofCwalina: yes, and I will add the comment.

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs
KrzysztofCwalina: good point.

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs
KrzysztofCwalina: will do

KrzysztofCwalina:

src/System.Buffers.Experimental/System/Buffers/Text/BufferWriter.cs
```diff
@@ -0,0 +1,69 @@
...
+        }
+
+        public void Clear()
+            => _written = 0;
+        
+        private void Resize(int desiredBufferSize = 0)
+        {
+            if (Enlarge == null) throw new BufferTooSmallException();
+            // double the size; add one to ensure that an empty buffer still resizes
+            if (desiredBufferSize <= 0) desiredBufferSize = _buffer.Length * 2 + 1;
```
ahsonkhan: This could cause integer overflow resulting in desiredBufferSize to be negative.

KrzysztofCwalina: Do you worry about the exception being strange or about not being able to resize to Int32.MaxValue? i.e. if the value wraps, the next line will throw ArgumentOutOfRange. The issue is that it will throw it "too early".

ahsonkhan: if _buffer.Length == int.MaxValue, _buffer.Length * 2 + 1 == _buffer.Length + 1 == -1, so we end up calling Enlarge(-1).Span, not throw ArgumentOutOfRangeException.

PR #1979, opened 2017-12-14
Changed System.Range based on language discussions

KrzysztofCwalina:
This is just the first stab.

cc: @terrajobst, @eerhardt, @jcouv, @joshfree, @ahsonkhan, @jaredpar, @khyperia, @MadsTorgersen

### DIFF
--- a/src/System.Buffers.Experimental/System/Range.cs
+++ b/src/System.Buffers.Experimental/System/Range.cs
@@ -11,20 +11,34 @@ namespace System
     {
         public const int UnboundedFirst = Int32.MinValue;
         public const int UnboundedLast = Int32.MaxValue;
+        public const uint UnboundedLength = UInt32.MaxValue;
 
         public readonly int First;
-        public readonly uint Length;
-        public int Last => (int)(Length + First - 1);
+        public readonly int Last; // Last is exclusive
+
+        public uint Length
+        {
+            get {
+                if (First == UnboundedFirst || Last == UnboundedLast) return UnboundedLength;
+                return (uint)((long)Last - (long)First);
+            }
+        }
 
         public Range(int first, uint length)
         {
-            if (length < 0)
-            {
-                throw new ArgumentOutOfRangeException("Length must be non-negative.", nameof(length));
-            }
-
+            if (first == UnboundedFirst)
+                throw new ArgumentOutOfRangeException(nameof(first));
+            
             First = first;
-            Length = length;
+            Last = (int)(first + length);
+
+            if (Last < First) throw new ArgumentOutOfRangeException(nameof(length));
+        }
+
+        private Range(int first, int last)
+        {
+            First = first;
+            Last = last;
         }
 
         public void Deconstruct(out int first, out int last)
@@ -32,13 +46,8 @@ namespace System
             first = First;
             last = Last;
         }
-        public static Range Construct(int first, int last = UnboundedLast)
-        {
-            if (last == UnboundedLast) return new Range(first, (uint)((long)int.MaxValue - first));
-            long length = (long)last - first + 1;
-            if (length > uint.MaxValue || length < 0) throw new ArgumentOutOfRangeException(nameof(last));
-            return new Range(first, (uint)length);
-        }
+        public static Range Construct(int first, int last)
+            => new Range(first, last);
 
         public struct Enumerator : IEnumerator<int>
         {
@@ -54,7 +63,7 @@ namespace System
             public bool MoveNext()
             {
                 _first++;
-                return _first <= _last;
+                return _first < _last;
             }
 
             public int Current => _first;
--- a/src/System.Buffers.Experimental/System/Range.cs
+++ b/src/System.Buffers.Experimental/System/Range.cs
@@ -1,26 +1,78 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
-namespace System {
-    public struct Range
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System
+{
+
+    public struct Range : IEnumerable<int>
     {
-        public Range(long index, long length)
+        public const int UnboundedFirst = Int32.MinValue;
+        public const int UnboundedLast = Int32.MaxValue;
+
+        public readonly int First;
+        public readonly uint Length;
+        public int Last => (int)(Length + First - 1);
+
+        public Range(int first, uint length)
         {
             if (length < 0)
             {
                 throw new ArgumentOutOfRangeException("Length must be non-negative.", nameof(length));
             }
 
-            Index = index;
+            First = first;
             Length = length;
         }
-        public readonly long Index;
-        public readonly long Length;
 
-        public void Deconstruct(out long index, out long length)
+        public void Deconstruct(out int first, out int last)
         {
-            index = Index;
-            length = Length;
+            first = First;
+            last = Last;
         }
+        public static Range Construct(int first, int last = UnboundedLast)
+        {
+            if (last == UnboundedLast) return new Range(first, (uint)((long)int.MaxValue - first));
+            long length = (long)last - first + 1;
+            if (length > uint.MaxValue || length < 0) throw new ArgumentOutOfRangeException(nameof(last));
+            return new Range(first, (uint)length);
+        }
+
+        public struct Enumerator : IEnumerator<int>
+        {
+            int _first;
+            int _last;
+
+            internal Enumerator(int first, int last)
+            {
+                _first = first - 1;
+                _last = last;
+            }
+
+            public bool MoveNext()
+            {
+                _first++;
+                return _first <= _last;
+            }
+
+            public int Current => _first;
+
+            object IEnumerator.Current => _first;
+
+            void IDisposable.Dispose() { }
+
+            void IEnumerator.Reset() => throw new NotSupportedException();
+        }
+
+        public Enumerator GetEnumerator()
+            => new Enumerator(First, Last);
+
+        IEnumerator<int> IEnumerable<int>.GetEnumerator()
+            => GetEnumerator();
+
+        IEnumerator IEnumerable.GetEnumerator()
+            => GetEnumerator();
     }
 }
--- a/src/System.Buffers.Experimental/System/Range.cs
+++ b/src/System.Buffers.Experimental/System/Range.cs
@@ -6,12 +6,11 @@ using System.Collections.Generic;
 
 namespace System
 {
-
+    // TODO: consider allowing Last > First. Ennumeration will count down.
     public struct Range : IEnumerable<int>
     {
         public const int UnboundedFirst = Int32.MinValue;
         public const int UnboundedLast = Int32.MaxValue;
-        public const uint UnboundedLength = UInt32.MaxValue;
 
         public readonly int First;
         public readonly int Last; // Last is exclusive
@@ -19,8 +18,8 @@ namespace System
         public uint Length
         {
             get {
-                if (First == UnboundedFirst || Last == UnboundedLast) return UnboundedLength;
-                return (uint)((long)Last - (long)First);
+                if (IsBound) return (uint)((long)Last - (long)First);
+                throw new InvalidOperationException("cannot get length of unbound range");
             }
         }
 
@@ -41,6 +40,8 @@ namespace System
             Last = last;
         }
 
+        public bool IsBound => First != UnboundedFirst && Last != UnboundedLast;
+
         public void Deconstruct(out int first, out int last)
         {
             first = First;
@@ -51,32 +52,36 @@ namespace System
 
         public struct Enumerator : IEnumerator<int>
         {
-            int _first;
+            int _current;
             int _last;
 
             internal Enumerator(int first, int last)
             {
-                _first = first - 1;
+                _current = first - 1;
                 _last = last;
             }
 
             public bool MoveNext()
             {
-                _first++;
-                return _first < _last;
+                _current++;
+                return _current < _last;
             }
 
-            public int Current => _first;
+            public int Current => _current;
 
-            object IEnumerator.Current => _first;
+            object IEnumerator.Current => _current;
 
             void IDisposable.Dispose() { }
 
             void IEnumerator.Reset() => throw new NotSupportedException();
         }
 
+        // TODO: write benchmark for this
         public Enumerator GetEnumerator()
-            => new Enumerator(First, Last);
+        {
+            if(IsBound) return new Enumerator(First, Last);
+            throw new InvalidOperationException("cannot enumerate unbound range");
+        }
 
         IEnumerator<int> IEnumerable<int>.GetEnumerator()
             => GetEnumerator();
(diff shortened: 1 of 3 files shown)


src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,82 @@
...
 
-        public void Deconstruct(out long index, out long length)
+        public void Deconstruct(out int first, out long last)
+        {
+            first = First;
+            last = Last;
+        }
+        public static Range Construct(int first, long last = UnboundedLast)
+        {
+            if (last == UnboundedLast) return new Range(first, int.MaxValue);
```
ahsonkhan: Should this be `last >= UnboundedLast`?

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,82 @@
...
+            if (length > int.MaxValue) throw new ArgumentOutOfRangeException(nameof(last));
+            return new Range(first, (int)length);
+        }
+
+        public struct Enumerator : IEnumerator<long>
+        {
+            long _first;
+            long _last;
+
+            internal Enumerator(int first, long last)
```
ahsonkhan: Why is `first` an Int32, but the struct field `_first` an Int64?

tests/System.Buffers.Experimental.Tests/RangeTests.cs
```diff
@@ -0,0 +1,44 @@
...
+using Xunit;
+
+namespace System.Buffers.Tests
+{
+    public class RangeTests
+    {
+        [Fact]
+        public void Basics() {
+
+            for (int first = 0; first < 10; first++)
```
ahsonkhan: If first is allowed to be negative, can we start the for loop from -10 instead of 0?

tests/System.Buffers.Experimental.Tests/RangeTests.cs
```diff
@@ -0,0 +1,44 @@
...
+                for (int length = 0; length < 10; length++)
+                {
+                    var range = new Range(first, length);
+
+                    Assert.Equal(first, range.First);
+                    Assert.Equal(length, range.Length);
+                    Assert.Equal(length + first - 1, range.Last);
+
+                    long sum = 0;
+                    foreach (var value in range)
```
ahsonkhan: nit: use explicit type instead of var

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,82 @@
...
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System
+{
+
+    public struct Range : IEnumerable<int>
     {
-        public Range(long index, long length)
+        public const int UnboundedFirst = Int32.MinValue;
```
eerhardt: @jaredpar - when we spoke yesterday, this was `static`.  Will `const` work as well?  Or does the compiler need it to be `static`?

eerhardt:
>                         newDimensions[newDimensionCount] = (int)range.Length;

This cast can go away as well.  And I think the `checked` section can be moved up to not include this chunk of code anymore.

---
Refers to: src/System.Numerics.Tensors/System/Numerics/Tensor.cs:1478 in 998d6da. [](commit_id = 998d6da1aa6d88822481fbd3270962ae40980317, deletion_comment = False)

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1458,19 +1458,19 @@ public virtual Tensor<T> Slice(ReadOnlySpan<Range> ranges)
                 Debug.Assert(range.Length > 0);
 
                 int currentDimension = Dimensions[i];
-                if ((uint)range.Index >= (uint)currentDimension)
+                if ((uint)range.First >= (uint)currentDimension)
                 {
-                    throw new ArgumentOutOfRangeException($"Range.Index at position '{i}' must be non-negative and less than the Dimension '{currentDimension}' at that position. The Range.Index value is '{range.Index}'.", nameof(ranges));
+                    throw new ArgumentOutOfRangeException($"Range.Index at position '{i}' must be non-negative and less than the Dimension '{currentDimension}' at that position. The Range.Index value is '{range.First}'.", nameof(ranges));
```
eerhardt: There are also usages of `Range.Index` in the string that should be replaced here and below.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,82 @@
...
+        public const long UnboundedLast = Int32.MaxValue;
+
+        public readonly int First;
+        public readonly int Length;
+        public long Last => ((long)Length) + First - 1;
```
eerhardt: Does `Last` imply that it is included in the Range?

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: How can last be >= long.MaxValue?

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,82 @@
...
-        public void Deconstruct(out long index, out long length)
+        public void Deconstruct(out int first, out long last)
+        {
+            first = First;
+            last = Last;
+        }
+        public static Range Construct(int first, long last = UnboundedLast)
+        {
+            if (last == UnboundedLast) return new Range(first, int.MaxValue);
+            long length = ((long)last) - first + 1L;
```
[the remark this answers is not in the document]
KrzysztofCwalina: last cannot be > than UnboundedLast. Having said that, last can be long.Maxvalue -1 and first can be int.MinValue, and then it will overflow. I will fix this.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: BTW, I thought about this whole design more. I want to remove any longs from the surface area. I updated the PR

tests/System.Buffers.Experimental.Tests/RangeTests.cs
KrzysztofCwalina: will fix.

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
```diff
@@ -1458,19 +1458,19 @@ public virtual Tensor<T> Slice(ReadOnlySpan<Range> ranges)
...
                 {
-                    if (range.Index + range.Length > currentDimension)
+                    if (range.First + range.Length > currentDimension)
                     {
-                        throw new ArgumentOutOfRangeException($"The Range at position '{i}' targets values outside of the Dimension '{currentDimension}' at that position. The Range.Index value is '{range.Index}'. The Range.Length value is '{range.Length}'.", nameof(ranges));
+                        throw new ArgumentOutOfRangeException($"The Range at position '{i}' targets values outside of the Dimension '{currentDimension}' at that position. The Range.Index value is '{range.First}'. The Range.Length value is '{range.Length}'.", nameof(ranges));
                     }
 
-                    offset += (int)(range.Index * strides[i]);
+                    offset += (int)(range.First * strides[i]);
```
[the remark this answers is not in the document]
KrzysztofCwalina: will fix

src/System.Numerics.Tensors/System/Numerics/Tensor.cs
KrzysztofCwalina: will fix

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,78 @@
...
 
-namespace System {
-    public struct Range
+using System.Collections;
+using System.Collections.Generic;
+
+namespace System
+{
+
+    public struct Range : IEnumerable<int>
```
KrzysztofCwalina: Should we call it Range32?

YohDeadfall: Sounds like Win32. It's better to use `IntRange` or just `Range`.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -13,8 +13,9 @@ public struct Range : IEnumerable<int>
         public const int UnboundedLast = Int32.MaxValue;
 
         public readonly int First;
-        public readonly uint Length;
-        public int Last => (int)(Length + First - 1);
+        public readonly int Last;
+
+        public uint Length => (uint)((long)Last - (long)First);
```
eerhardt: This is going to force consumers of Length to cast to int right?  Say if they need to pass Range.Length into String.IndexOf or similar methods that take a Length/count parameter.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,89 @@
...
-        public readonly long Length;
+        public static Range Construct(int first, int last)
+            => new Range(first, last);
 
-        public void Deconstruct(out long index, out long length)
+        public struct Enumerator : IEnumerator<int>
         {
-            index = Index;
-            length = Length;
+            int _first;
```
jcouv: Consider naming `_current` instead.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,89 @@
...
+            public int Current => _first;
+
+            object IEnumerator.Current => _first;
+
+            void IDisposable.Dispose() { }
+
+            void IEnumerator.Reset() => throw new NotSupportedException();
         }
+
+        public Enumerator GetEnumerator()
```
jcouv: Should we allow enumerating unbounded Ranges?

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,89 @@
...
+        public struct Enumerator : IEnumerator<int>
         {
-            index = Index;
-            length = Length;
+            int _first;
+            int _last;
+
+            internal Enumerator(int first, int last)
+            {
+                _first = first - 1;
```
jcouv: If we allow enumerating unbounded ranges, we'll have a problem when first is `Int32.Min` here.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,89 @@
...
+
+    public struct Range : IEnumerable<int>
     {
-        public Range(long index, long length)
+        public const int UnboundedFirst = Int32.MinValue;
+        public const int UnboundedLast = Int32.MaxValue;
+        public const uint UnboundedLength = UInt32.MaxValue;
+
+        public readonly int First;
+        public readonly int Last; // Last is exclusive
```
jcouv: I don't have a strong opinion here, but we should follow-up on naming: First+Last vs. Start+End.

tests/System.Buffers.Experimental.Tests/RangeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+
+using Xunit;
+
+namespace System.Buffers.Tests
+{
+    public class RangeTests
+    {
+        [Fact]
+        public void Basics()
+        {
```
jcouv: Nit: un-necessary empty line

tests/System.Buffers.Experimental.Tests/RangeTests.cs
```diff
@@ -0,0 +1,81 @@
...
+                for (uint length = 0; length < 20; length++)
+                {
+                    var range = new Range(first, length);
+
+                    Assert.Equal(first, range.First);
+                    Assert.Equal(length, range.Length);
+                    Assert.Equal(length + first, range.Last);
+
+                    long sum = 0;
+                    foreach (int value in range)
```
jcouv: Would you mind doing a micro-benchmark just to make sure that `foreach (var i in Range.Construct(i, j)) { ... }` is fast as we expect?

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,89 @@
...
+
+            internal Enumerator(int first, int last)
+            {
+                _first = first - 1;
+                _last = last;
+            }
+
+            public bool MoveNext()
+            {
+                _first++;
```
jcouv: It seems that we're assuming that first < last, and therefore we should increment to go from `first` to `last`.
What if I create a `Range.Construct(10, 0)`?

I think such ranges should be allowed, but the enumerator should count down.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Range.cs
YohDeadfall: Last can't be exclusive, but End means that everywhere.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: Yeah, this whole thing with unbound ends will be very confusing. Length does not make sense either for unbound ranges. I was thinking about making Length throw for unbound ranges and have a property Range.IsBound. If we do it, then enumeration can throw.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: I changed it. Will submit update soon.

tests/System.Buffers.Experimental.Tests/RangeTests.cs
KrzysztofCwalina: will fix

tests/System.Buffers.Experimental.Tests/RangeTests.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: I added TODO to consider.

src/System.Buffers.Experimental/System/Range.cs
YohDeadfall: For example, the `System.Linq.Enumerable` class has the `First<T>` and `Last<T>` methods. The second returns the last element in the sequence or throws. So having the `Last` field which holds an excluded value would be confusing.

```csharp
var range = new Range(0, 10);
var last1 = range.Last; // using the Last field, 10
var last2 = range.Last(); // using the Enumerable.Last<T> method, 9
```
But they should return the same result. Isn't?

KrzysztofCwalina: Good point. I will change it.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -1,26 +1,92 @@
...
+
+        public bool IsBound => First != UnboundedFirst && Last != UnboundedLast;
+
+        public void Deconstruct(out int first, out int last)
+        {
+            first = First;
+            last = Last;
+        }
+        public static Range Construct(int first, int last)
+            => new Range(first, last);
```
eerhardt: Should this check for `last < first` and throw?

PR #1981, opened 2017-12-15
Added IsValid, Bind, and BindToValid. Renamed Last to End

KrzysztofCwalina:
I added some helpers and overrode ToString to make it easier to work with ranges.
Renamed Last to End as "Last" is confusing given the value is exclusive.

@ahsonkhan, @joshfree, @eerhardt, @jcouv

### DIFF
--- a/src/System.Buffers.Experimental/System/Range.cs
+++ b/src/System.Buffers.Experimental/System/Range.cs
@@ -48,6 +48,7 @@ namespace System
             first = First;
             end = End;
         }
+
         public static Range Construct(int first, int end)
             => new Range(first, end);
 
@@ -97,7 +98,7 @@ namespace System
         }
 
         /// <summary>
-        /// Returns true if this Range is a valid range for a zero based index of sspecified length.
+        /// Returns true if this Range is a valid range for a zero based index of specified length.
         /// </summary>
         /// <param name="length">zero based length.</param>
         /// <returns></returns>
@@ -115,10 +116,14 @@ namespace System
                 if (First > length) return false;
                 return End <= length;
             }
-
-            throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Converts an unbound Range (IsBound == false) to a bound one (IsBound == true).
+        /// </summary>
+        /// <param name="length">zero based length of an indexable 'list' the range is being bound to.</param>
+        /// <returns>Bound Range (IsBound == true).</returns>
+        /// <remarks>The method throws ArgumentOutOfRangeException Range.IsValid(lenght) returns false.</remarks>
         public Range Bind(int length)
         {
             if (!IsValid(length)) throw new ArgumentOutOfRangeException(nameof(length));
--- a/src/System.Buffers.Experimental/System/Range.cs
+++ b/src/System.Buffers.Experimental/System/Range.cs
@@ -3,22 +3,23 @@
 
 using System.Collections;
 using System.Collections.Generic;
+using System.Text;
 
 namespace System
 {
     // TODO: consider allowing Last > First. Ennumeration will count down.
-    public struct Range : IEnumerable<int>
+    public readonly struct Range : IEnumerable<int>
     {
         public const int UnboundedFirst = Int32.MinValue;
-        public const int UnboundedLast = Int32.MaxValue;
+        public const int UnboundedEnd = Int32.MaxValue;
 
         public readonly int First;
-        public readonly int Last; // Last is exclusive
+        public readonly int End; // End is exclusive
 
         public uint Length
         {
             get {
-                if (IsBound) return (uint)((long)Last - (long)First);
+                if (IsBound) return (uint)((long)End - (long)First);
                 throw new InvalidOperationException("cannot get length of unbound range");
             }
         }
@@ -29,42 +30,42 @@ namespace System
                 throw new ArgumentOutOfRangeException(nameof(first));
             
             First = first;
-            Last = (int)(first + length);
+            End = (int)(first + length);
 
-            if (Last < First) throw new ArgumentOutOfRangeException(nameof(length));
+            if (End < First) throw new ArgumentOutOfRangeException(nameof(length));
         }
 
-        private Range(int first, int last)
+        private Range(int first, int end)
         {
             First = first;
-            Last = last;
+            End = end;
         }
 
-        public bool IsBound => First != UnboundedFirst && Last != UnboundedLast;
+        public bool IsBound => First != UnboundedFirst && End != UnboundedEnd;
 
-        public void Deconstruct(out int first, out int last)
+        public void Deconstruct(out int first, out int end)
         {
             first = First;
-            last = Last;
+            end = End;
         }
-        public static Range Construct(int first, int last)
-            => new Range(first, last);
+        public static Range Construct(int first, int end)
+            => new Range(first, end);
 
         public struct Enumerator : IEnumerator<int>
         {
             int _current;
-            int _last;
+            int _end;
 
-            internal Enumerator(int first, int last)
+            internal Enumerator(int first, int end)
             {
                 _current = first - 1;
-                _last = last;
+                _end = end;
             }
 
             public bool MoveNext()
             {
                 _current++;
-                return _current < _last;
+                return _current < _end;
             }
 
             public int Current => _current;
@@ -79,7 +80,7 @@ namespace System
         // TODO: write benchmark for this
         public Enumerator GetEnumerator()
         {
-            if(IsBound) return new Enumerator(First, Last);
+            if(IsBound) return new Enumerator(First, End);
             throw new InvalidOperationException("cannot enumerate unbound range");
         }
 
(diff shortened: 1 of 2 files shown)


src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -29,42 +30,42 @@ public Range(int first, uint length)
...
-        private Range(int first, int last)
+        private Range(int first, int end)
         {
             First = first;
-            Last = last;
+            End = end;
         }
 
-        public bool IsBound => First != UnboundedFirst && Last != UnboundedLast;
+        public bool IsBound => First != UnboundedFirst && End != UnboundedEnd;
```
ahsonkhan: Why can't First == UnboundedFirst? I thought only End is exclusive.

KrzysztofCwalina: The property returns true if both of the ends are bound. Otherwise false. The property can be used to see if some operations on the range would work, e.g. GetEnumerator works only for bound ranges. "Unbound" means "I don't know/care what the value is". It does not make sense to enumerate such ranges.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -29,42 +30,42 @@ public Range(int first, uint length)
...
-        public void Deconstruct(out int first, out int last)
+        public void Deconstruct(out int first, out int end)
         {
             first = First;
-            last = Last;
+            end = End;
         }
-        public static Range Construct(int first, int last)
-            => new Range(first, last);
+        public static Range Construct(int first, int end)
```
ahsonkhan: nit: add new line

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -3,22 +3,23 @@
...
 {
     // TODO: consider allowing Last > First. Ennumeration will count down.
-    public struct Range : IEnumerable<int>
+    public readonly struct Range : IEnumerable<int>
     {
         public const int UnboundedFirst = Int32.MinValue;
-        public const int UnboundedLast = Int32.MaxValue;
+        public const int UnboundedEnd = Int32.MaxValue;
 
         public readonly int First;
```
YohDeadfall: `First` should be renamed to `Start` to follow naming convention as @jcouv mentioned earlier in the previous PR: `Start` and `End`.

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -88,5 +89,77 @@ IEnumerator<int> IEnumerable<int>.GetEnumerator()
...
+        {
+            if (First == UnboundedFirst)
+            {
+                if (End == UnboundedEnd) return true;
+                return End <= length;
+            }
+            else // First is bounded
+            {
+                if (First < 0) return false;
+                if (End == UnboundedEnd) return First <= length;
```
ahsonkhan: Should First be < length here?

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -88,5 +89,77 @@ IEnumerator<int> IEnumerable<int>.GetEnumerator()
...
             => GetEnumerator();
+
+        public bool Contains(int value)
+        {
+            if (!IsBound) throw new InvalidOperationException("Unbound ranges cannot contain");
+            return value >= First && value < End;
+        }
+
+        /// <summary>
+        /// Returns true if this Range is a valid range for a zero based index of sspecified length.
```
ahsonkhan: nit: spelling - sspecified

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -88,5 +89,77 @@ IEnumerator<int> IEnumerable<int>.GetEnumerator()
...
+            }
+            else // First is bounded
+            {
+                if (First < 0) return false;
+                if (End == UnboundedEnd) return First <= length;
+                if (First > length) return false;
+                return End <= length;
+            }
+
+            throw new NotImplementedException();
```
ahsonkhan: Will we ever reach here?

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -88,5 +89,77 @@ IEnumerator<int> IEnumerable<int>.GetEnumerator()
...
+        }
+
+        public Range Bind(int length)
```
ahsonkhan: Can you please add a comment here explaining what Bind is meant to do?

tests/System.Buffers.Experimental.Tests/RangeTests.cs
```diff
@@ -106,7 +141,7 @@ public void Errors()
             });
             Assert.Throws<ArgumentOutOfRangeException>(() =>
             {
-                // MaxValue is used as a sentinel for Last, so Last cannot endup being it.
+                // MaxValue is used as a sentinel for End, so End cannot endup being it.
```
ahsonkhan: nit: spelling endup

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: I made it so that range can start after the last element. This is to allow empty ranges.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: will fix. thanks

src/System.Buffers.Experimental/System/Range.cs
ahsonkhan: I see. So, you can't enumerate from [int.MinValue..int.MinValue + 5] similar to int.MaxValue?

int.MaxValue being the sentinel for End makes sense given End is exclusive (for example [0..int.MaxValue] works fine. Are we treating the lower bound the same way?

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -88,5 +89,77 @@ IEnumerator<int> IEnumerable<int>.GetEnumerator()
 
         IEnumerator IEnumerable.GetEnumerator()
             => GetEnumerator();
+
+        public bool Contains(int value)
+        {
+            if (!IsBound) throw new InvalidOperationException("Unbound ranges cannot contain");
```
YohDeadfall: If it's unbound, then it contains any value. If it has unbound start, check the end only. Same for ranges with unbound start.

```csharp
if (IsBound) return Start <= value && value < End;
if (Start == UnboundStart) return value < End;
if (End == UnboundEnd) return Start <= value;
return false;
```

tests/System.Buffers.Experimental.Tests/RangeTests.cs
KrzysztofCwalina: will fix

src/System.Buffers.Experimental/System/Range.cs
YohDeadfall: Do not forget to check direction.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: will do

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: nah, I will remove.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: Yeah, I am not thrilled with Start. It sounds like an action is about to start. That's why I originally did not use End. But I will think about the naming a bit more.

src/System.Buffers.Experimental/System/Range.cs
KrzysztofCwalina: Theoretically yes, but what are the scenarios for calling Contains on unbound ranges?

src/System.Buffers.Experimental/System/Range.cs
shaggygi: I agree with @YohDeadfall  as `Start` seems to fit with `End`.  However, I also see @KrzysztofCwalina point.  I like `First` and `Last` better, though.

YohDeadfall: `Begin` would be fine, but unfortunately it shouldn't be used for historical reasons since everywhere in .NET only `Start`/`End` is used for such situations in parameter names, fields, properties. Also `Begin` and `End` are used for pre-`Task` async operations in method names.

src/System.Buffers.Experimental/System/Range.cs
YohDeadfall: Exceptions are okay for `GetEnumerator` since we can't iterate over infinity sequences, but `Contains` can work and therefore should. Otherwise, it wouldn't be useful, and developers would be managed to write they own implementation. For example:

```csharp
class Example
{
    private Range _range;

    public Example(Range range) => _range = range;

    public void DoSomething(int value)
    {
        if (_range.Contains(value))
        {
        }
        else
        {
        }
    }
}
```

What if `range` provided to the constructor is unbound? Should I use `(_range.Start == Range.UnboundStart || _range.Start <= value) && (_range.End == Range.UnboundEnd || value < _range.End)` to know if the `value` is in `_range` to be sure that the code will not throw?

src/System.Buffers.Experimental/System/Range.cs
```diff
@@ -29,42 +30,43 @@ public Range(int first, uint length)
...
-        private Range(int first, int last)
+        private Range(int first, int end)
         {
             First = first;
-            Last = last;
+            End = end;
         }
 
-        public bool IsBound => First != UnboundedFirst && Last != UnboundedLast;
+        public bool IsBound => First != UnboundedFirst && End != UnboundedEnd;
```
YohDeadfall: `IsBound` should take into account that the range could be reversed:

* `Range(UnboundedLast, UnboundedFirst)` is unbound;
* `Range(5, UnboundedFirst)` is unbound;
* `Range(UnboundedLast, 5)` is unbound too.

src/System.Buffers.Experimental/System/Range.cs
YohDeadfall: Also the `in` pattern matching will use `Contains`.

```csharp
switch (variable)
{
    case in ..0: // should not throw
        break;
    // ...
    case in 5..: // should not throw
        break;
}
```

PR #1982, opened 2017-12-15
Changed IMemoryList based on Pipelines API Review

KrzysztofCwalina:
- Renamed Rest to Next
- Removed interfaces it extended
- Removed CopyTo
- Updated pipelines to use this (shared) interface

@pakrym, @ahsonkhan, @joshfree

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
@@ -9,6 +9,118 @@ using System.Numerics;
 
 namespace System.Buffers
 {
+
+    public static class MemoryListExtensions
+    {
+        // span creation helpers:
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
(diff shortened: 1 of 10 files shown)


davidfowl:
Getting closer

src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -9,6 +9,118 @@
 
 namespace System.Buffers
 {
+
+    public static class MemoryListExtensions
```
davidfowl: Are we going to use the ReadableBuffer implementation?

KrzysztofCwalina: For? 

BTW, this project is not even referenced by the pipelines. I needed to add all these extension methods below to be able to compile existing code (in this project) after removing CopyTo, and ISequence implementations from IMemoryList. i.e. there is no plan to ship these members.

davidfowl: Ah I see

PR #1988, opened 2017-12-19
Removed Position.End

KrzysztofCwalina:
cc: @pakrym, @davidfowl, @joshfree, @ahsonkhan 

When sequences move to end default(Position) is returned from TryGet
Before calling ISequence<T>.TryGet for the first time, the calling code has to call ISequence<T>.First, as default position no longer mean the first item (necessarily)
APIs like buffer.PositionOf(byte value) return Position? (nullable). They return null if value is not found

All these changes were requested by @pakrym to allow him to use Position in pipelines instead of ReadCursor.


KrzysztofCwalina:

PR #1990, opened 2017-12-20
Simplified Position API

KrzysztofCwalina:
cc: @pakrym, @joshfree, @davidfowl

### DIFF
--- a/src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
+++ b/src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
@@ -16,20 +16,19 @@ namespace System.Collections.Sequences
 
         public static implicit operator Position(int index) => new Position(index, null);
 
-        public static explicit operator int(Position position) => position.Index;
-
-        public int Index => _index;
+        public static explicit operator int(Position position) => position._index;
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public T GetItem<T>() => _item == null ? default : (T)_item;
+        public (T item, int index) Get<T>() {
+            var item = _item == null ? default : (T) _item;
+            return (item, _index);
+        }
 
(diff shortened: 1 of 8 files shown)


src/System.Collections.Sequences/System/Collections/Sequences/Position.cs
```diff
@@ -16,20 +16,19 @@ namespace System.Collections.Sequences
 
         public static implicit operator Position(int index) => new Position(index, null);
```
pakrym: Do we need all these operators, `Get<T>`, and static `Create` method? 
Can we just expose item and index and add a single public constructor that takes them?

pakrym: Also in all of the usages of `Get<T>` you never check that item is not null. Do we need to add some kind of validation that position was of an expected type?

KrzysztofCwalina: That's how it was originally, but you guys wanted to make it harder for people to get access to the object. Get/Create allow you to store an internal type in object to make it impossible to retrieve by third party code.

pakrym: You can always do `Get<object>` and get around it. And `ReadOnlyBuffer` stores publicly available types there. I don't think we can protect it from retrieving the item.

KrzysztofCwalina: Yes, but I wanted to change the check to only return the object if typeof(T) == _item.GetType().
I was planing to do it in a subsequent PR.

KrzysztofCwalina:
Let me merge this one. I will play with the typeof(T) == _item.GetType() check, and if it does not work, we can go back to Item and Index properties.

PR #1993, opened 2017-12-22
Prepared System.Buffer.Primitives for API Review

KrzysztofCwalina:
- Unified Positions
- Moved ReadCursorOperations around. These APIs are not general purpose Position APIs. They only work with ReadOnlyBuffer.
- Renamed ReadOnlyBufferReader to BufferReader
- Simplified namespaces

@pakrym, @davidfowl, @joshfree, @terrajobst


src/System.Buffers.Experimental/System/Buffers/BufferExtensions.cs
```diff
@@ -56,7 +56,7 @@ public static int CopyTo(this IMemoryList<byte> list, Span<byte> destination)
             {
                 var current = list.Memory.Span;
                 var index = current.IndexOf(value);
-                if (index != -1) return Collections.Sequences.Position.Create(list, index);
+                if (index != -1) return new Collections.Sequences.Position(list, index);
```
benaadams: `System.Collections.Sequences` is imported, is there a name clash? 

i.e. can this be just `Position`?

KrzysztofCwalina: Ah, forgot about it. I will remove.

tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
```diff
@@ -180,5 +181,129 @@ public void Create_WorksWithOwnedMemory()
             b => b.Slice(0, 70).Slice(0, b.End),
             b => b.Slice(70, b.Start)
         };
+
+        static bool TryGetBuffer(Position begin, Position end, out ReadOnlyMemory<byte> data, out Position next)
```
pakrym: Why is this here?

tests/System.IO.Pipelines.Tests/TestBufferFactory.cs
```diff
@@ -0,0 +1,93 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers;
+using System.Collections.Generic;
+using System.IO.Pipelines.Testing;
+using System.Linq;
+using System.Text;
+
+namespace System.IO.Pipelines.Tests
```
pakrym: This is useful not on only for pipelines, why is it moved back?

src/System.Buffers.Primitives/System/Collections/Position.cs
```diff
@@ -0,0 +1,81 @@
...
+        {
+            switch (Segment)
+            {
+                case null:
+                    return default;
+                case T segment:
+                    return segment;
+            }
+
+            //ThrowHelper.ThrowInvalidOperationException(ExceptionResource.UnexpectedSegmentType);
```
pakrym: Why remove throwing?

src/System.IO.Pipelines/System/IO/Pipelines/PositionOperations.cs
```diff
@@ -8,7 +8,7 @@
 
 namespace System.IO.Pipelines
 {
-    public static class ReadCursorOperations
+    public static class PositionOperations
```
pakrym: Why is it moved back to pipelines?

tests/System.IO.Pipelines.Tests/TestBufferFactory.cs
KrzysztofCwalina: It's used just by pipelines tests

src/System.Buffers.Primitives/System/Collections/Position.cs
KrzysztofCwalina: I have no idea. Good catch.

tests/System.IO.Pipelines.Tests/TestBufferFactory.cs
pakrym: No, see https://github.com/dotnet/corefxlab/blob/master/tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs#L21

pakrym: Not sure why it builds but it's used by ROB and RBR tests that are in primitives.

src/System.IO.Pipelines/System/IO/Pipelines/PositionOperations.cs
KrzysztofCwalina: Well, it's pipelines specific. It does not work with an arbitrary position. It only works with positions of IMemoryList, and possibly other things that ReadOnlyBuffer supports.

pakrym: It works with everything that `ReadOnlyBuffer` supports that's why it's in buffers. There is no need to have another copy in Pipelines.

tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
KrzysztofCwalina: I can move the tests that call Seek to Pipelines.Tests and then eliminate it. What do you think?

src/System.IO.Pipelines/System/IO/Pipelines/PositionOperations.cs
KrzysztofCwalina: Position is more general than Buffers. Look in System.Collections.Sequences. There is a Hashtable and ArrayList implementation that uses it.

pakrym: Then move these methods to be on ROB

tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
pakrym: These tests were testing internal implementations of `ROB` why would we want to move them to `Pipelines`? Or even worse duplicate code being tested inside the test itself.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
```diff
@@ -0,0 +1,203 @@
...
+using System.Buffers;
+using System.Collections.Sequences;
+using System.Runtime.CompilerServices;
+
+namespace System.Buffers
+{
+    public readonly partial struct ReadOnlyBuffer 
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        static bool TryGetBuffer(Position begin, Position end, out ReadOnlyMemory<byte> data, out Position next)
```
pakrym: internal

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
```diff
@@ -0,0 +1,203 @@
...
+                    return true;
+            }
+
+            ThrowHelper.ThrowNotSupportedException();
+            next = default;
+            return false;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        static Position Seek(Position begin, Position end, long bytes, bool checkEndReachable = true)
```
pakrym: internal

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
```diff
@@ -0,0 +1,203 @@
...
+            else
+            {
+                cursor = SeekMultiSegment(begin, end, bytes, checkEndReachable);
+            }
+
+            return cursor;
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        static Position SeekMultiSegment(Position begin, Position end, long bytes, bool checkEndReachable)
```
pakrym: private for everything else

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
KrzysztofCwalina: Why internal? It can be private

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
KrzysztofCwalina: will do

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
KrzysztofCwalina: same: why not private?

tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
KrzysztofCwalina: What do you propose I do? Is there a way to test this without calling ROB's private APIs?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
pakrym: So the tests from tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs  do not require copied code or being moved.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
pakrym: The same thing, there is a test in tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs  using it.

tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
pakrym: Make them internal. It has been done in other places to verify implementation details.

tests/System.IO.Pipelines.Tests/TestBufferFactory.cs
KrzysztofCwalina: There are two copies. I will remove one of them (i.e. the one in pipelines)

src/System.Buffers.Primitives/System/Collections/Position.cs
```diff
@@ -0,0 +1,82 @@
...
+            _index = 0;
+        }
+
+        public object Segment => _segment;
+        public int Index => _index;
+
+        public static explicit operator int(Position position) => position._index;
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public (T segment, int index) Get<T>()
```
pakrym: Do we expect consumers to check if the segment is non-null/default?

src/System.Buffers.Primitives/System/Collections/Position.cs
```diff
@@ -0,0 +1,82 @@
...
+            }
+
+            ThrowHelper.ThrowInvalidOperationException(ExceptionResource.UnexpectedSegmentType);
+            return default;
+        }
+
+        public static bool operator ==(Position left, Position right) => left._index == right._index && left._segment == right._segment;
+        public static bool operator !=(Position left, Position right) => left._index != right._index || left._segment != right._segment;
+
+        public static Position operator +(Position value, int index)
```
pakrym: I thought we were going to limit the ability to modify positions?

src/System.IO.Pipelines.Extensions/ReadableBufferSequence.cs
```diff
@@ -16,16 +15,16 @@ public ReadableBufferSequence(ReadOnlyBuffer buffer) : this()
             _buffer = buffer;
         }
 
-        public Collections.Sequences.Position First => Collections.Sequences.Position.Create(_buffer.Start.Segment, _buffer.Start.Index);
+        public Position Start => new Position(_buffer.Start.Segment, _buffer.Start.Index);
```
pakrym: Why have start at all if we treat `default` as a `Start`?

src/System.IO.Pipelines/System/IO/Pipelines/PositionOperations.cs
KrzysztofCwalina: Hmmm, I have done it initially and it did not work. But now it does. I will fix.

KrzysztofCwalina: I removed it and added InternalsVisible from Pipelines.Tests to ReadOnlyBuffer. We need to design proper seek APIs though.

pakrym: Wait, why do you need `InternalsVisisbleTo` to `Pipelines.Tests`? Just don't move `SeekTests.cs` from primitives.

src/System.Buffers.Primitives/System/Collections/Position.cs
KrzysztofCwalina: We now have public ctors, Index, and Segment. The + operator is much more hidden than those. 
Also, without the operator, code that uses position as index is very wacky. It does not care about the Segment, yet it has to deal with it (in the ctor, at least).

We can think about it more, but I would treat this as a separate PR.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
```diff
@@ -10,8 +10,65 @@ namespace System.Buffers
 {
     public readonly partial struct ReadOnlyBuffer 
     {
+        public static int Seek(Position begin, Position end, out Position result, byte byte0)
```
pakrym: Change them to be extension/instance methods on `ReadOnlyBuffer`

src/System.Buffers.Primitives/System/Collections/Position.cs
KrzysztofCwalina: Depends on what they use the position for.

src/System.IO.Pipelines.Extensions/ReadableBufferSequence.cs
KrzysztofCwalina: I dont think we treat default as start. I remove this early this week.

src/System.Buffers.Primitives/System/Collections/Position.cs
pakrym: I think in most cases they either use it or not. If they don't use it, direct access to Index is enough. If they need both then I prefer throwing to returning default.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
KrzysztofCwalina: Let's do it separately. I think I have seen some code that calls the APIs with positions other that the buffer's start/end

pakrym: I think `Seek` is the last one, but we can do it in a separate pass, just don't like moving things back and forth.

src/System.Buffers.Primitives/System/Collections/Position.cs
KrzysztofCwalina: Sure, we can play with this. But I still want to do an experiment with changing the check to typeof(T) == typeof(Segment). I will try to do it after the new year. If if does not work, we can remove this method.

PR #1999, opened 2017-12-25
Fast TaskRunScheduler

benaadams:
`Task.Factory.StartNew` is slow compared to `QueueUserWorkItem` or the newer lower contention `QueueUserWorkItem(preferLocal: true)` https://github.com/dotnet/corefx/issues/12442

Since the returned Task is just ignored; might as well use the faster versions (which `Task.Factory.StartNew` essentially ends up calling - more or less...)

Resolves: https://github.com/dotnet/corefxlab/issues/2002
Resolves: #2000 (due to rebase on  #2001)

/cc @davidfowl @pakrym

### DIFF
--- a/src/System.IO.Pipelines/System/Threading/Scheduler.cs
+++ b/src/System.IO.Pipelines/System/Threading/Scheduler.cs
@@ -5,9 +5,13 @@ namespace System.Threading
 {
     public abstract class Scheduler
     {
+        private protected static readonly Action<object> ScheduleAction = o => ((Action)o)();
+
         public static Scheduler TaskRun { get; } = new TaskRunScheduler();
         public static Scheduler Inline { get; } = new InlineScheduler();
 
         public abstract void Schedule(Action<object> action, object state);
+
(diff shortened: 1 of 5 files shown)


benaadams:
Can be improved post https://github.com/dotnet/corefxlab/pull/2001

benaadams:
Rebased on https://github.com/dotnet/corefxlab/pull/2001

src/System.IO.Pipelines/System/Threading/Scheduler.cs
```diff
@@ -5,9 +5,13 @@ namespace System.Threading
 {
     public abstract class Scheduler
     {
+        private protected static readonly Action<object> ScheduleAction = o => ((Action)o)();
+
         public static Scheduler TaskRun { get; } = new TaskRunScheduler();
```
davidfowl: Now I think we should rename this to ThreadPool.

benaadams:
Rebased and superseded by https://github.com/dotnet/corefxlab/pull/2026

Issue #2000, opened 2017-12-25
Add Scheduler.Schedule(Action action)


benaadams:
Otherwise you end up with the wrap for 
```csharp
private static readonly Action<object> _scheduleContinuation = o => ((Action)o)();

private static void TrySchedule(Scheduler scheduler, Action action)
{
    if (action != null)
    {
        scheduler.Schedule(_scheduleContinuation, action);
    }
}
```
Which means a double wrap and allocation for a Faster TaskScheduler https://github.com/dotnet/corefxlab/pull/1999 as it also needs to do the same wrap to enqueue.

i.e. the wrapping should be handled by the Scheduler not the pipe

davidfowl:
This looks like a fine change now that we have a base class.

PR #2018, opened 2018-01-03
Add single argument Schedule overload

pakrym:
https://github.com/dotnet/corefxlab/issues/2000


benaadams:
Pipe.cs should ditch 
```csharp
private static readonly Action<object> _scheduleContinuation = o => ((Action)o)();
```

And call
```csharp
scheduler.Schedule(action);
```
Rather than
```csharp
scheduler.Schedule(_scheduleContinuation, action);
```

PR #2022, opened 2018-01-04
Another set of API review changes

pakrym:
1. Make `ReadOnlyBuffer(Position, Position)` private.
2. Rename `IMemoryList<T>` to `IBufferList`.
3. Remove `CreatePosition` from `Enumerable`.
4. Remove multi-byte seek methods
5. Rename `Seek` to `PositionOf` and change signature.
6. Rename `Move` to `Seek`
7. Add `ReadOnlyBuffer(IEnumerable<Memory<byte>>)` constructor

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
@@ -1,6 +1,7 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Collections.Generic;
 using System.Collections.Sequences;
 using System.Diagnostics;
 using System.IO.Pipelines;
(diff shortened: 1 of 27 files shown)


src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -252,7 +252,64 @@ public bool TryGet(ref Position cursor, out ReadOnlyMemory<byte> data, bool adva
 
             return result;
         }
-        
+
+        public int Seek(out Position result, byte byte0)
```
KrzysztofCwalina: Why does it return an int? Isn't the int always equal to byte0 or -1? If yes, the API should be:
```c#
 public bool TrySeek(out Position position, byte value);
// or
public Position? Seek(byte value);
```

Also, since we call integer index returning method IndesOf, why not call these PositionOf?

pakrym: It works for single byte overload, what about multi-byte ones? We returned int to be able to quickly know what byte was found, without slicing or creating the reader.

KrzysztofCwalina: The only overload that you are actually using in the pipelines/parser the one that takes a single value.

The other overloads should be called PositionOfAny (to be consistent with IIndexOfAny). These are not used by the http parser, and so maybe we don't have to return the found value, just like we don't return it from IndexOfAny. Alternativelly, we should redesign IndexOfAny (in MemoryExtensions)

pakrym: I like changing it to be more like `IndexOfAny`, but not returning byte that was found would make things harder because you need to slice just to get it.

pakrym: I'll change it to `PositionOf[Any]`

KrzysztofCwalina: Why do we need the [Any] versions? Nothing uses them except for tests.

pakrym: If we go with, "cut everything we don't use" - no. But then we can cut all of them and implement it in kestrel.

KrzysztofCwalina: Sorry, I was not clear. I was asking whether we need the super optimized versions that return the byte found.
 
What about the following:
```c#
public Position? PositionOf(byte value);
public Position? PositionOf(ReadOnlySpan<byte> value);
public Position? PositionOfAny(ReadOnlySpan<byte> values);
```

pakrym: Often the first things you do after you found the position is to check what character is it. We did it for URL parsing, looked for `'?', ' '` and depending what was found executed different logic.

KrzysztofCwalina:
Looks good.

Issue #2019, opened 2018-01-04
BufferReader API : consider moving to 64-bit (System.Buffers.Primitives)


mgravell:
In particular, `ConsumedBytes` - but possibly also `Skip(long length)`.

Ultimately, the reader can be spanning an arbitrary number of spans; the classic implementation is over a `ReadOnlyBuffer`, and that has already moved to `public long Length => ...` / `Slice(long start, long length)`. So since the *thing it wraps* clearly considers the >2GiB world, so should this.

Obviously things like `Index` are unaffected, since that relates to the current span which is `int`-indexed

Drawaes:
I hit this issue when experimenting with making a ReadOnlyBuffer over a large memory mapped file. The 2GiB limit was something I then had to code around

pakrym:
Yep, it's a bug.

mgravell:
@pakrym  any thoughts on the `Skip` aspect? (64-bit)

pakrym:
I think if you are skipping 2GB of memory Reader is not the right thing to use. ReadOnlyBuffer can do it much more efficiently,

mgravell:
Maybe, but general purpose consumer code isn't necessarily well placed to
switch readily between those two. Conversely, if ROB is better at this,
perhaps large skips could be achieved by a check inside Skip, without the
caller needing to worry about it?

if(len > LARGE)
{
    var rob = new ROB(Cursor, End).Slice(len);
    this = new BufferReader(rob, bytesConsumed + Len);
}

?


On 5 Jan 2018 4:10 p.m., "Pavel Krymets" <notifications@github.com> wrote:

I think if you are skipping 2GB of memory Reader is not the right thing to
use. ReadOnlyBuffer can do it much more efficiently,

—
You are receiving this because you authored the thread.
Reply to this email directly, view it on GitHub
<https://github.com/dotnet/corefxlab/issues/2019#issuecomment-355593467>,
or mute the thread

PR #2023, opened 2018-01-04
Public sequence, long ConsumedBytes

pakrym:
Public `Sequence` allows some nice extension methods on `BufferReader` for example `SliceToEnd()`

https://github.com/dotnet/corefxlab/issues/2019

### DIFF
--- a/src/System.Text.Http.Parser/HttpParser.cs
+++ b/src/System.Text.Http.Parser/HttpParser.cs
@@ -356,7 +356,7 @@ namespace System.Text.Http.Parser
             finally
             {
                 consumed = reader.Position;
-                consumedBytes = reader.ConsumedBytes;
+                consumedBytes = (int)reader.ConsumedBytes;
 
                 if (done)
                 {
--- a/tests/Benchmarks/BytesReaderBench.cs
+++ b/tests/Benchmarks/BytesReaderBench.cs
@@ -84,7 +84,7 @@ public class BytesReaderBench
 
             using (iteration.StartMeasurement())
             {
-                while(Utf8Parser.TryParse(reader.Span.Slice(reader.ConsumedBytes), out int value, out int consumed)){
+                while(Utf8Parser.TryParse(reader.Span.Slice((int)reader.ConsumedBytes), out int value, out int consumed)){
                     reader.Skip(consumed + 1);
                 }
             }
--- a/tests/System.Buffers.Primitives.Tests/ReadableBufferReaderFacts.cs
+++ b/tests/System.Buffers.Primitives.Tests/ReadableBufferReaderFacts.cs
@@ -84,6 +84,34 @@ namespace System.IO.Pipelines.Tests
             reader.Take();
             Assert.True(reader.End);
         }
+        
+        [Fact]
+        public void SliceAftereWorks()
+        {
+            var reader = BufferReader.Create(Factory.CreateWithContent(new byte[] { 1, 2, 3, 4, 5 }));
+            var b1 = reader.Take();
+            var b2 = reader.Take();
+            var b3 = reader.Take();
+
+            Assert.Equal(1, b1);
+            Assert.Equal(2, b2);
+            Assert.Equal(3, b3);
+            Assert.Equal(new byte[] { 4, 5 }, reader.SliceAfter().ToArray());
+        }
+
+        [Fact]
+        public void SliceBeforeWorks()
+        {
+            var reader = BufferReader.Create(Factory.CreateWithContent(new byte[] { 1, 2, 3, 4, 5 }));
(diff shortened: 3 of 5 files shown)


pakrym:
Closing in favour of larger https://github.com/dotnet/corefxlab/pull/2024

pakrym:
That PR got closed, reopening.

KrzysztofCwalina:
I don't think we should be exposing sequence (or any other APIs) without validating that they are needed and sufficient to implement the extensions methods.

pakrym:
Reader doesn't have anything else useful that we don't expose. Having sequence is enough for what I think of: I want to implement a method that returns `ROB` from the current position of `Reader` to the end.

What's the reason not to expose it?

KrzysztofCwalina:
The main reason to not expose it is that we need to validate. You are adding APIs that we don't exercise. Moreover, I suspect raw sequence exposed will be very error prone as extensions will have to remember how much they already read.

Secondary reason is that I just implemented many extension methods and did not have to use the sequence.

pakrym:
The extension would be for `BufferReader<ReadOnlyBuffer>` not `BufferReader<T>` and use Position to Slice to the end.

I've looked at some example of how BuffferReader is used and there seems to be a popular theme of wanting to slice underlying buffer using current position after some reading.

pakrym:
> Secondary reason is that I just implemented many extension methods and did not have to use the sequence.
So how would you implement `Slice to the end of underlying buffer` without having access to buffer itself?

KrzysztofCwalina:
We need to add the following primitives (as instance methods) to the reader:
PositionOf(byte)
PositionOf(ReadOnlySpan<byte>)
AdvanceTo(Position)

pakrym:
How is it related?

KrzysztofCwalina:
> How is it related?

Sorry, I was answering: "So how would you implement Slice to the end of underlying buffer without having access to buffer itself?"

pakrym:
I still don't see how to do it even having those methods.

My implementation would look like this:

```
public static ReadOnlyBuffer SliceToEnd(ref this BufferReader<ReadOnlyBuffer> reader)
{
    return reader.Sequence.Slice(reader.BytesConsumed);
}
```

KrzysztofCwalina:
Your implementation would not compile. Sequence does not have a method Slice. Which kind of proves my point that we need to add the property together with actual extensions that we hope to implement.

pakrym:
It is on `BufferReader<ReadOnlyBuffer>`.

KrzysztofCwalina:
> It is on BufferReader<ReadOnlyBuffer>

This is an indication that the BufferReader<T> abstraction is not complete.

pakrym:
No, it's an indication that ISequence abstraction is not complete. And it's impossible to build everything on top of it, you at lease need Slice.

KrzysztofCwalina:
Fair, either of the abstractions is not complete. Which again proves my point that we need to add the extension methods that we hope to implement in the same PR as the APIs we are adding to implement such extensions.

pakrym:
Added extension methods and tests.

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -356,7 +356,7 @@ public HttpParser(bool showErrorDetails)
             finally
             {
                 consumed = reader.Position;
-                consumedBytes = reader.ConsumedBytes;
+                consumedBytes = (int)reader.ConsumedBytes;
```
mgravell: consider `checked((int)reader.ConsumedBytes);` ?

tests/Benchmarks/BytesReaderBench.cs
```diff
@@ -84,7 +84,7 @@ static void ParseInt32ReadableBufferReader()
 
             using (iteration.StartMeasurement())
             {
-                while(Utf8Parser.TryParse(reader.Span.Slice(reader.ConsumedBytes), out int value, out int consumed)){
+                while(Utf8Parser.TryParse(reader.Span.Slice((int)reader.ConsumedBytes), out int value, out int consumed)){
```
mgravell: consider checked cast

pakrym:
Closing, pending design decisions,

tests/System.Buffers.Primitives.Tests/ReadableBufferReaderFacts.cs
```diff
@@ -84,6 +84,34 @@ public void CursorIsCorrectAtEnd()
             reader.Take();
             Assert.True(reader.End);
         }
+        
+        [Fact]
+        public void SliceAftereWorks()
+        {
```
bruno-garcia: typo: SliceAfter**e**Works -> SliceAfterWorks

PR #2024, opened 2018-01-04
Unified Readers

KrzysztofCwalina:
Removed System.Buffer.Experimental buffer reader
Added parsing, search and binary reading methods that used to be on the experimental reader
The changes to the buffer reader look bigger than they are. I think git got confused about the file move/rename. I don't think I changed anything in the existing reader (from pipelines) except for some filed and property renames.

@ahsonkhan, @pranavkm, @joshfree


pranavkm:
Think you meant @pakrym

KrzysztofCwalina:
> Think you meant @pakrym

Ah, yes. Thanks.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        private TSequence _sequence;
+        private Position _currentSegmentPosition;
+        private Position _nextSegmentPosition;
+
+        private ReadOnlySpan<byte> _currentSegment;
+        private int _currentSegmentConsumedBytes;
+        
+        // TODO: this should be long, but it will impact the HTTP parser. We should meassure perf impact in isolation
+        private int _totalConsumedBytes;
```
pakrym: It has to be long to be correct, so it doesn't really matter if it does impact parser.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+    {
+        public static BufferReader<TSequence> Create<TSequence>(TSequence buffer) where TSequence : ISequence<ReadOnlyMemory<byte>>
+        {
+            return new BufferReader<TSequence>(buffer);
+        }
+    }
+
+    public ref partial struct BufferReader<TSequence> where TSequence : ISequence<ReadOnlyMemory<byte>>
+    {
+        private TSequence _sequence;
```
pakrym: Expose this as public, please. Having ability to write extension methods for `BufferReader<ROB>` and use `ROB` is very nice.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+            MoveNext();
+        }
+
+        public bool End => _hasCompleted;
+
+        public int CurrentSegmentIndex => _currentSegmentConsumedBytes;
+
+        public Position CurrentPosition => _currentSegmentPosition + _currentSegmentConsumedBytes;
+
+        public ReadOnlySpan<byte> CurrentSegment => _currentSegment;
```
pakrym: `CurrentSpan`?

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+
+        public ReadOnlySpan<byte> UnreadSegment => _currentSegment.Slice(_currentSegmentConsumedBytes);
```
pakrym: When is this useful?

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+            _sequence = buffer;
+            _currentSegmentPosition = _sequence.Start;
+            _nextSegmentPosition = _currentSegmentPosition;
+            _currentSegment = default;
+            MoveNext();
+        }
+
+        public bool End => _hasCompleted;
+
+        public int CurrentSegmentIndex => _currentSegmentConsumedBytes;
```
pakrym: `CurrentIndex`?

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
```diff
@@ -0,0 +1,154 @@
...
+        public bool TryReadUntill(out ReadOnlyBuffer bytes, byte delimiter)
+        {
+            var range = new PositionRange(CurrentPosition, AdvanceToDelimiter(delimiter).GetValueOrDefault());
+            if (range.Start == default || range.End == default)
+            {
+                bytes = default;
+                return false;
+            }
+
+            bytes = new ReadOnlyBuffer(range.Start, range.End);
```
pakrym: This is not correct. You cannot just create `ROB` from random positions.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
```diff
@@ -0,0 +1,154 @@
...
+                bytes = default;
+                return false;
+            }
+            if (range.End != default)
+            {
+                SkipTo(range.End);
+                Skip(delimiter.Length);
+            }
+
+            bytes = new ReadOnlyBuffer(range.Start, range.End);
```
pakrym: Same here, `T` has to be ReadOnlyBuffer to be able to do it.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
pakrym: Or just `Span`

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
pakrym: Or just `Index`

davidfowl: Index

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_binary.cs
```diff
@@ -0,0 +1,53 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Buffers.Binary;
+using System.Collections.Sequences;
+
+namespace System.Buffers
+{
+    public ref partial struct BufferReader<TSequence> where TSequence : ISequence<ReadOnlyMemory<byte>>
+    {
+        public bool TryRead(out int value, bool littleEndian = false)
```
davidfowl: Can we not add these APIs back? I don't think they are up to the quality bar and they shouldn't be part of the API review

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
```diff
@@ -0,0 +1,154 @@
...
+                currentSegmentConsumedBytes = 0;
+                unread = memory.Span;
+                index = unread.IndexOf(value);
+                if (index != -1) return currentSegmentPosition + index;
+            }
+        }
+        #endregion
+    }
+
+    struct PositionRange
```
pakrym: Seems suspiciously similar to ROB

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+                _currentSegmentConsumedBytes = 0;
+                if (_currentSegment.Length > 0)
+                {
+                    return;
+                }
+            }
+            _hasCompleted = true;
+        }
+
+        public void Skip(int byteCount)
```
davidfowl: Lets rename this to Advance

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
```diff
@@ -0,0 +1,149 @@
...
+                MoveNext();
+            }
+
+            if (byteCount > 0)
+            {
+                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.length);
+            }
+        }
+
+        public void SkipTo(Position position)
```
davidfowl: When is this useful?

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
```diff
@@ -0,0 +1,154 @@
...
+                }
+                previousPosition = _nextSegmentPosition;
+            }
+
+            _nextSegmentPosition = nextPosition;
+            _currentSegmentPosition = currentPosition;
+            return null;
+        }
+
+        private Position? PositionOf(ReadOnlySpan<byte> value)
```
davidfowl: Isn't this basically IndexOf?

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
KrzysztofCwalina: will do

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
KrzysztofCwalina: Let's do it separately. I worry that consumers will need more than the sequence to make it work.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
KrzysztofCwalina: I think logically the type is a sequence of segments (that's what we call the concept in Position). Span is a type that we happen to use in this API to represent a segment. I would rather keep property names that describe the concept, not the representation.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
KrzysztofCwalina: I think it will be super useful for the extension APIs. It's already used in some of the reading APIs I added.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
KrzysztofCwalina: The reason I renamed it is that we have two indexes: this one and the "global" index.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
KrzysztofCwalina: Why not? Can we fix it in ROB?

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
KrzysztofCwalina: ?

KrzysztofCwalina:
> They should remain extension methods and we should have a model for writing more instead of baking them all in. We need more real world samples.

OK, that's fair. I will try to make them into extensions. I just wanted to remove the experimental reader, but you are right that these APIs need more bake time.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
pakrym: Position is transparent, the implementor of ISequence can put random things there. `ROB ` cannot support them all.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
pakrym: Let's just do it, it's super easy, and I was doing in other PR that I had to close.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader.cs
pakrym: We have memory in `IMemoryList` not `Segment`, `Segment` is just another term wit didn't use before.

src/System.Buffers.Primitives/System/Buffers/ReaderWriter/BufferReader_search.cs
pakrym: Use ROB, why do we need another version of the basically same type?

PR #2025, opened 2018-01-05
Removed Old BufferReader

KrzysztofCwalina:
I changed all members of the old reader into extension methods over the pipelines reader, and deleted the old reader.

@davidfowl, @pakrym


src/System.Buffers.Experimental/System/Buffers/BufferReader_search.cs
```diff
@@ -0,0 +1,40 @@
...
+        public static bool TryReadUntill<TSequence>(ref this BufferReader<TSequence> reader, out ReadOnlyBuffer bytes, byte delimiter)
+            where TSequence : ISequence<ReadOnlyMemory<byte>>
+        {
+            var position = reader.PositionOf(delimiter);
+            if (position == null)
+            {
+                bytes = default;
+                return false;
+            }
+            bytes = new ReadOnlyBuffer(reader.Position, position.Value);
```
pakrym: Still not correct.

KrzysztofCwalina: as we discussed, we need to make it correct, or remove the ctor.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -42,7 +44,9 @@ public BufferReader(TSequence buffer)
 
         public Position Position => _currentPosition + _index;
 
-        public ReadOnlySpan<byte> Span => _currentSpan;
+        public ReadOnlySpan<byte> CurrentSegment => _currentSpan;
```
pakrym: Same comment about `Span`. We don't call it Segment in `IMemoryList` why call it segment here.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -94,36 +98,143 @@ private void MoveNext()
...
+                _currentSpan = memory.Span;
+            }
+            else
+            {
+                _currentSpan = default;
+            }
+            _index = 0;
+        }
+      
+        public Position? PositionOf(byte value)
```
pakrym: There is same logic in `ROB.Seek`

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: Cause IMemoryList is a list of '''Memory<T>``` instances. ReadOnlyBuffer is a sequence of segments, not sequence of spans.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -94,36 +98,143 @@ private void MoveNext()
...
+                }
+                currentSegmentConsumedBytes = 0;
+                unread = memory.Span;
+
+                index = unread.IndexOf(value);
+                if (index != -1) return currentSegmentPosition + index;
+            }
+        }
+
+        public static int CopyTo(BufferReader<TSequence> bytes, Span<byte> buffer)
```
pakrym: There is same logic on `ROB`. Lets not duplicate it.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: Yeah, but I don't have access to ROB

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
pakrym: It's a list of things you can get Memory from, same as here, list of things you can get Span from.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: Yeah, but I don't have access to ROB. I agree that it would be good to somehow refactor to reuse the logic. Any ideas for how to refactor?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
pakrym: That doesn't mean we have to duplicate everything on reader now.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
pakrym: We decided some time ago that we won't try to implement everything on top of `ISequence` and now we are trying to do it.

We should expose `Sequence` and so allow access to all these implementations on `ROB`. We can expand ISequence in the future to accommodate new features but we shouldn't go into rabbit hole right now. Especially considering these implementations are not used in production code and may not be thoroughly tested.

pakrym: We can expose `SliceBefore()` and `SliceAfter()` extension methods in the way I talked in other thread so the consumer can get `ROB` of either and call existing methods on it.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
davidfowl: I think we should call it Span

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
davidfowl: Agreed. We should keep this API minimal.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -42,7 +44,9 @@ public BufferReader(TSequence buffer)
 
         public Position Position => _currentPosition + _index;
 
-        public ReadOnlySpan<byte> Span => _currentSpan;
+        public ReadOnlySpan<byte> CurrentSegment => _currentSpan;
+
+        public ReadOnlySpan<byte> UnreadSegment => _currentSpan.Slice(_index);
```
davidfowl: This is a useful property 😄 . I think we should expose this as the RemainingSpan and we should also have RemainingBytes

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -94,36 +98,143 @@ private void MoveNext()
...
+            {
+                var segment = memory.Span;
+                index = segment.IndexOf(value);
+                if (index != -1) return currentSegmentPosition + index;
+                currentSegmentPosition = nextSegmentPosition;
+            }
+            return default;
+        }
+
+        public Position? PositionOf(ReadOnlySpan<byte> value)
```
davidfowl: Can we remove this for now? I'd like to keep this API minimal and fully tested. Expose whatever needs to be exposed to implement this in the experimental APIs

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
pakrym: Remaining sounds like you are talking about entire buffer, especially when RemainingBytes is absolute.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: I don't think it can be removed, and I don't think we can sweep the issue under the rug. The remaining abstraction is not powerful enough to implement reading of variable length data (based on searches). We would have to expose many fields of the reader (in addition to the sequence) to implement for example ReadUntill, at which point using the reader would be very error prone as the callers could mess up its state too easily.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: I am not sure how it helps here. I don't have BufferReader<ROB>; I have BufferReader<TSequence>

pakrym: As I said, in a previous meeting we decided not to try building everything on top of `ISequence`, especially while we have only one real implementation. We can live with supporting these things only on ROB and see what other things would implement `ISequence<ReadOnlyMemory<byte>>` if there would be any other at all.

pakrym: Maybe moving the reader to `ISequence` was a mistake too.

pakrym:
I think moving reader to use `ISequence` and treating it as an abstraction was a mistake, we should concentrate our time on finishing things that around `ROB`, not trying to add more stuff that supports `ISequence` especially considering we have exactly one implementation of it.

We may also want to consider not shipping interface at all until we find a real need and at least another implementer for it because at the current state it clearly does not allow all the scenarios we are trying to use it for (lack of `Slice` is one of the bigger problems.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: We have several problems in various abstractions: ISequence, Position, ROB. But I would not throw baby with the bathwater. I implemented many extensions over the APIs we have now (the no merge PR) and it seems to be almost working. I think we need to solve a couple of small problems to enable both reading over sequences and slicing.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: I want to eliminate the old reader. To do this I need a method roughly like the following:
```c#
        public static bool TryReadUntill(ref this BufferReader<ReadOnlyBuffer> reader, out ReadOnlyBuffer bytes, byte delimiter)
        {
            var position = reader.PositionOf(delimiter);
            if (position == null)
            {
                bytes = default;
                return false;
            }
            bytes = new ReadOnlyBuffer(reader.Position, position.Value);
            reader.SkipTo(position.Value);
            reader.Skip(1);
            return true;
        }
```

Even if I change the this parameter to reader over ROB, I am not sure how to implement it. Can you suggest a way to implement it without adding PositionOf to the reader?

pakrym: `while (!reader.IsEnd || reader.Read() != delimiter) {}` until you find `delimiter` and use

KrzysztofCwalina: The method is a Try, but Read has side effects.

pakrym: Copy reader to local and copy back if needed.

KrzysztofCwalina: I think it's ok as a workaround for missing APIs, but it's not a good long term solution. The current implementation is super fast: does not copy large structs, uses vectorized IndexOf to find matches, in common cases where the match is found in the first segment, it simply adjusts the _index (once). Your implementation (workaround) is not even in the same ballpark.

pakrym: I know, I'm not claiming that it is efficient. But we have fast one in ROB and we need to decide how to reuse it and avoid code duplication.

KrzysztofCwalina:
OK, I removed all but Skip(Position) and CopyTo. Are we good with these two?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: I think in general calling these two concepts "Span" has the same issue: it's not clear what the span contains. Segment is very clear: the type is a list of segments and it's one of these.

pakrym:
They both exist in `ROB` in forms of `Slice` and `CopyTo`, why duplicate?

pakrym:
There are also no tests for them.

KrzysztofCwalina:
> They both exist in ROB in forms of Slice and CopyTo, why duplicate?

Cause I believe we cannot develop hundreds of BCL APIs operating over ROB; it's too policy heavy. I think we need a ROB abstraction. There are many ways to design a multi-segment buffer and ROB is just one of them. In the BCL we don't like to take dependencies on one particular design. Whenever we do, we endup having to redesign and reimplement these APIs operating over such policy heavy types.

> There are also no tests for them.

Thus "NO MERGE". I said in the PR message that I will add tests once we agree on the APIs.

KrzysztofCwalina:
OK, I added tests to CopyTo, the only method we agreed to keep on the reader.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -17,6 +17,8 @@ namespace System.Buffers
         internal readonly Position BufferStart;
         internal readonly Position BufferEnd;
 
+        public static readonly ReadOnlyBuffer Empty = new ReadOnlyBuffer(new byte[0]);
```
pakrym: `Array.Empty<>`?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -6,21 +6,27 @@
...
 {
-    public class BufferReader
+    public static class BufferReader
     {
         public static BufferReader<TSequence> Create<TSequence>(TSequence buffer) where TSequence : ISequence<ReadOnlyMemory<byte>>
         {
             return new BufferReader<TSequence>(buffer);
         }
+
+        public static int CopyTo<TSequence>(BufferReader<TSequence> reader, Span<byte> destination)
```
pakrym: Why do you need it here and not on the instance?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -94,36 +102,70 @@ private void MoveNext()
...
             }
 
-            if (length > 0)
+            if (byteCount > 0)
             {
                 ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.length);
             }
         }
+
+        internal static int CopyTo(BufferReader<TSequence> bytes, Span<byte> destination)
```
pakrym: It doesn't follow usual `CopyTo` conventions about destination sizes. Maybe we should name it differently? `Read(span)`?

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
KrzysztofCwalina: Not available in NS 1.1

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: It's a limitation of the C# compiler. The Span can be stack allocated but the instance CopyTo method could store the passed in span into the _currentSegment field.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: It does not advance the index, so it cannot be called Read, but I will rename it to Peek

pakrym: Maybe it should advance? Wouldn't it be the most common use case?

pakrym: Or `TryRead(Span)`.

KrzysztofCwalina:

KrzysztofCwalina:

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -6,21 +6,27 @@
...
     public ref struct BufferReader<TSequence> where TSequence : ISequence<ReadOnlyMemory<byte>>
     {
         private ReadOnlySpan<byte> _currentSpan;
         private int _index;
+
         private TSequence _sequence;
         private Position _currentPosition;
         private Position _nextPosition;
+               
         private int _consumedBytes;
```
pakrym: Long?

Issue #2002, opened 2017-12-25
Task.Factory.StartNew is slower than QUWI for fire & forget tasks


benaadams:
`TaskRunScheduler` uses `Task.Factory.StartNew`; Kestrel's `SocketsTransport` uses `TaskRunScheduler` - this will be effecting its performance.

/cc @davidfowl @pakrym

PR #2026, opened 2018-01-06
Scheduler improvements

benaadams:
Scheduler Devirtualization
* Seal internal types `TaskRunScheduler` and `InlineScheduler`
* Maintain shared `TaskRunScheduler` and `InlineScheduler` in derived typed members; rather than base type
* Expose via base type property that will devirtualize on inlining when the Jit sees the derived types (as type is maintained above, and derived type is sealed) https://github.com/dotnet/coreclr/pull/15766

Resolves: https://github.com/dotnet/corefxlab/issues/2027

Fast Fire-and-Forget Tasks for `TaskRunScheduler`
* Queue directly to `ThreadPool`; when available, rather than have the extra overhead of a `Task` which is ignored.
* On netcoreapp2.1 use `QueueUserWorkItem(preferLocal: true)` which is lower contention and more closely matches `Task` behaviour dotnet/corefx#12442
* If the `Action` fails create a failed `Task` so it can be captured as an unobserved task exception as per `Task.Run`

Resolves: https://github.com/dotnet/corefxlab/issues/2002

Rename TaskRun -> ThreadPool https://github.com/dotnet/corefxlab/pull/2026#pullrequestreview-87068486

/cc @davidfowl @pakrym

### DIFF
--- a/src/System.IO.Pipelines/System.IO.Pipelines.csproj
+++ b/src/System.IO.Pipelines/System.IO.Pipelines.csproj
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>An abstraction for doing efficient asynchronous IO</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFrameworks>netstandard1.1;netstandard2.0;netcoreapp2.1</TargetFrameworks>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <PackageIconUrl>http://go.microsoft.com/fwlink/?linkid=833199</PackageIconUrl>
     <!--<DefineConstants Condition="'$(Configuration)' == 'Debug'">$(DefineConstants);BLOCK_LEASE_TRACKING</DefineConstants>-->
--- a/src/System.IO.Pipelines/System/Threading/Scheduler.cs
+++ b/src/System.IO.Pipelines/System/Threading/Scheduler.cs
@@ -5,8 +5,11 @@ namespace System.Threading
 {
     public abstract class Scheduler
     {
-        public static Scheduler TaskRun { get; } = new TaskRunScheduler();
-        public static Scheduler Inline { get; } = new InlineScheduler();
+        private static TaskRunScheduler _taskRunScheduler = new TaskRunScheduler();
+        private static InlineScheduler _inlineScheduler = new InlineScheduler();
+
+        public static Scheduler TaskRun => _taskRunScheduler;
+        public static Scheduler Inline => _inlineScheduler;
         
         public abstract void Schedule(Action action);
         public abstract void Schedule(Action<object> action, object state);
--- a/src/System.IO.Pipelines/System/Threading/Scheduler.cs
+++ b/src/System.IO.Pipelines/System/Threading/Scheduler.cs
@@ -5,10 +5,10 @@ namespace System.Threading
 {
     public abstract class Scheduler
     {
-        private static TaskRunScheduler _taskRunScheduler = new TaskRunScheduler();
+        private static ThreadPoolScheduler _threadPoolScheduler = new ThreadPoolScheduler();
         private static InlineScheduler _inlineScheduler = new InlineScheduler();
 
-        public static Scheduler TaskRun => _taskRunScheduler;
+        public static Scheduler ThreadPool => _threadPoolScheduler;
         public static Scheduler Inline => _inlineScheduler;
         
         public abstract void Schedule(Action action);
--- a/src/System.IO.Pipelines/System/Threading/ThreadPoolScheduler.cs
+++ b/src/System.IO.Pipelines/System/Threading/ThreadPoolScheduler.cs
@@ -12,9 +12,9 @@ namespace System.Threading
         {
 #if NETCOREAPP2_1
             // Queue to low contention local ThreadPool queue; rather than global queue as per Task
-            Threading.ThreadPool.QueueUserWorkItem(_actionAsTask, action, preferLocal: true);
+            Threading.ThreadPool.QueueUserWorkItem(_actionWaitCallback, action, preferLocal: true);
 #elif NETSTANDARD2_0
-            Threading.ThreadPool.QueueUserWorkItem(_actionAsTask, action);
+            Threading.ThreadPool.QueueUserWorkItem(_actionWaitCallback, action);
 #else
             Task.Factory.StartNew(action);
 #endif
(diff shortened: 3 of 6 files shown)


benaadams:
myget seems unhappy
```
21:23:12   Retrying 'FindPackagesByIdAsyncCore' for source 'https://dotnet.myget.org/F/dotnet-corefxlab/FindPackagesById()?id='Microsoft.NETCore.Compilers''.
21:23:12   Response status code does not indicate success: 503 (Service Unavailable).
```

src/System.IO.Pipelines/System/Threading/Scheduler.cs
```diff
@@ -5,8 +5,11 @@ namespace System.Threading
 {
     public abstract class Scheduler
     {
-        public static Scheduler TaskRun { get; } = new TaskRunScheduler();
-        public static Scheduler Inline { get; } = new InlineScheduler();
+        private static TaskRunScheduler _taskRunScheduler = new TaskRunScheduler();
+        private static InlineScheduler _inlineScheduler = new InlineScheduler();
+
+        public static Scheduler TaskRun => _taskRunScheduler;
```
davidfowl: ThreadPool

benaadams: Done

benaadams:
Still
```
Failure to restore a package: Failed to restore packages.
```
:sad:

benaadams:
myget issue https://github.com/dotnet/coreclr/issues/15769

src/System.IO.Pipelines/System/Threading/ThreadPoolScheduler.cs
```diff
@@ -0,0 +1,83 @@
...
+#if NETCOREAPP2_1 || NETSTANDARD2_0
+        // Catches only the exception into a failed Task, so the fire-and-forget action 
+        // can be queued directly to the ThreadPool without the extra overhead of as Task
+        private readonly static WaitCallback _actionAsTask = state =>
+        {
+            try
+            {
+                ((Action)state)();
+            }
+            catch (Exception ex)
```
benaadams: Do you want to catch the Exception? Currently assuming so as side-effect is quite bad - however is it always wrapped (for logging etc)

KrzysztofCwalina:
What are we doing with this PR? We have an API review of these APIs on Friday (I think) and it would be good to merge or close this one before.

@benaadams, @pakrym, @terrajobst, @davidfowl

benaadams:
@dotnet test CI

benaadams:
> What are we doing with this PR?

Just shooting yourself in the foot; perf-wise, not to use `QueueUserWorkItem` over `Task.Run` if you aren't interested in the `Task` (especially the `preferLocal: true` variant).

src/System.IO.Pipelines/System/Threading/Scheduler.cs
```diff
@@ -5,8 +5,11 @@ namespace System.Threading
 {
     public abstract class Scheduler
     {
-        public static Scheduler TaskRun { get; } = new TaskRunScheduler();
```
davidfowl: Why this change?

benaadams: Because you said so here https://github.com/dotnet/corefxlab/pull/2026#pullrequestreview-87068486

davidfowl: Should have been more specific, why the syntax change.

benaadams: Storing it in a `Scheduler` field means the exact type gets lost for:
```csharp
Scheduler.ThreadPool.Schedule((() => {})
```
So it remains a virtual call. 

Storing it in an exact type field `ThreadPoolScheduler` means when `Scheduler.ThreadPool` inlines the Jit knows exactly what it is and can devitalize it to a direct call.

Won't directly help pipelines as the Transports store the configured values in a `Scheduler` field; but will help anyone calling it directly.

src/System.IO.Pipelines/System/Threading/ThreadPoolScheduler.cs
```diff
@@ -0,0 +1,83 @@
...
+    {
+        public override void Schedule(Action action)
+        {
+#if NETCOREAPP2_1
+            // Queue to low contention local ThreadPool queue; rather than global queue as per Task
+            Threading.ThreadPool.QueueUserWorkItem(_actionAsTask, action, preferLocal: true);
+#elif NETSTANDARD2_0
+            Threading.ThreadPool.QueueUserWorkItem(_actionAsTask, action);
+#else
+            Task.Factory.StartNew(action);
```
davidfowl: Should this just be QueueUserWorkItem?

benaadams: Doesn't exist on `netstandard1.1` came in at `netstandard2.0` 😢

src/System.IO.Pipelines/System.IO.Pipelines.csproj
```diff
@@ -2,7 +2,7 @@
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
     <Description>An abstraction for doing efficient asynchronous IO</Description>
-    <TargetFramework>netstandard1.1</TargetFramework>
+    <TargetFrameworks>netstandard1.1;netstandard2.0;netcoreapp2.1</TargetFrameworks>
```
davidfowl: This is probably the most unfortunate part of this change but meh..

src/System.IO.Pipelines/System/Threading/ThreadPoolScheduler.cs
```diff
@@ -0,0 +1,83 @@
...
+            {
+                try
+                {
+                    _action(_state);
+                }
+                catch (Exception ex)
+                {
+                    // Create faulted Task for the TaskScheulder to handle exception
+                    // rather than letting it escape onto the ThreadPool and crashing the process
+                    Task.FromException(ex);
```
davidfowl: This is pretty atypical. I think it should be left to crash.

benaadams: K, will clean up the code a lot to remove it

benaadams: Removed

juliusfriedman:
Somehow I know the Scheduler is faux paus; it should be IScheduler or more technically "Inate" or     'IInate' with a 'Resolution' field for Time units...

It should be able to accept ISchedulables e
g.  Itenarius or otherwise such as Tasks; I feel as if a distinction should be made for CoRoutines as well such that I can schedule a Thread or Task with a specific core topology and have the Scheduler ensure such and if not able to ensure such I need a way to aggregate the failure and handle it if necessary optionally stopping propagation and allowing it to be restarted at the whim of the application (the event fiber).

I would appreciate hardware support in the form of kernels which I would be able to combine with Tensors, Manifolds and ISlicables... (useful when block reading and or writing of a scratch pad medium or otherwise... not to mention Pipes and eventually; hopefully - "Buses" which are [time] erg coalesced Pipes which operate on a Manifold and it's Tensor metrics... think raw memory and hardware first than USB or HDMI 2nd... would love to see UsbClient and other stuff shift to completely managed code with as thinly placed drivers which are generic as possible. Ideally just the kernels interface should be enough and consumers of the kernels can produce other meaningful interactions through Communiquae or more technically 'Signals' and their abstractions of the erg they represent.

Scheduler, Fiber, Aggregate Exception. ThreadPriority structure, IHardware, Intrinsic, Abstraction, Interface, etc and more (Array / Slice competition - would be even better with an internal base class under array which could be used to subvert the need for Slice in general) are all here...

Apache 2 license:

https://stackoverflow.com/questions/38379278/how-can-i-improve-this-wrapper-for-thread/38444258#38444258

http://net7mma.codeplex.com/SourceControl/latest#Concepts/Classes/Intrinsics.cs

http://net7mma.codeplex.com/SourceControl/latest#Concepts/Classes/Arrays.cs

.[>..//<]

juliusfriedman:
Please also see: http://net7mma.codeplex.com/SourceControl/latest#Common/Classes/TaggedException.cs ; I believe that it made it's way into the framework as Pattern Matching... I can only assume such co-compiler of ye generation 'next' still forth cometh...

src/System.IO.Pipelines/System/Threading/Scheduler.cs
juliusfriedman: Ben, what do you know about guns or shooting them at that? Perhaps shooting yourself in the foot is an exercise the application is required to perform albeit under controlled circumstance; in short we are already using Task what's left to shoot for? I would hope for something over thread like 🥇 

http://net7mma.codeplex.com/SourceControl/latest#Concepts/Classes/Threading/Threading.cs

And

http://net7mma.codeplex.com/SourceControl/latest#Concepts/Classes/Hardware.cs

Now that we finally have unsafe and call indirect... (thanks me)

http://net7mma.codeplex.com/SourceControl/latest#Concepts/Classes/CommonIntermediateLanguage.cs

http://net7mma.codeplex.com/SourceControl/latest#Concepts/Classes/Unsafe.cs

*sigh*...

..."Of the greatest King of Kings am I. If any other want to know how great I am or where I lie, let them outdo me in thy works." [In the poem Diodorus becomes "a traveller from an antique land...]

And whence we met the traveller(s) from such a car away and antique land, 
They spake and said—“Two vast and trunkless legs of stone 
Stand in the desert, one of brick and one of stone. . . . Near them, on the sand, 
Half sunk a shattered visage lies, whose frown, 
And wrinkled lip, and sneer of cold command, 
Tell that its sculptor well those passions read 
Which yet survive, stamped on these lifeless things, 
The hand that mocked them, and the heart that fed; 
And on the pedestal, these words appear: 

The flood lifted up the fish on to the cliff-bank;the whale became sad, where he swam on the shingle.Whale's bone.Magi.
--

Here fight Titus and the Jews.
Here the inhabitants flee from Jerusalem.

Doom.
Hostages.

Romulus and Remus, two brothers: a she-wolf fed them in Rome city, far from their native land.

Here the horse stands above the mound of woe,
It suffers tribulation; just as to her Erta appointed anxiety,
A grave of grief, in sorrow and anguish of heart.

Wood. Biter. Rush.


My name is not important, King of Kings I am, important. 
Look on my Works, ye Mighty, and despair! 
Nothing beside remains. Round the decay 
Of that colossal Wreck, boundless and bare 
Thy lone and level sands stretch far away... 

After the water recedes take sand and lament with it as I have, set thine creation in the sun and let is bask; In three days time whence my clay is hardednded  bring it forth so I may give it great scrutiny and critique.

src/System.IO.Pipelines/System.IO.Pipelines.csproj
terrajobst: Meh indeed. It's in implementation details with zero bearings on the user's experience.

davidfowl: > It's in implementation details with zero bearings on the user's experience.

In this case yes.

PR #2032, opened 2018-01-08
Add MinimumSegmentSize

pakrym:
So it can work with pools that return exact buffer size that was requested.

### DIFF
--- a/tests/System.IO.Pipelines.Tests/PipePoolTests.cs
+++ b/tests/System.IO.Pipelines.Tests/PipePoolTests.cs
@@ -93,11 +93,33 @@ namespace System.IO.Pipelines.Tests
             await pipe.Writer.WriteAsync(new byte[writeSize]);
         }
 
+        [Fact]
+        public async Task RentsMinimumSegmentSize()
+        {
+            var pool = new DisposeTrackingBufferPool();
+            var writeSize = 512;
+
+            var pipe = new Pipe(new PipeOptions(pool, minimumSegmentSize: 2020));
+
+            var buffer = pipe.Writer.Alloc(writeSize);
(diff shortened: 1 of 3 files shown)


tests/System.IO.Pipelines.Tests/PipePoolTests.cs
```diff
@@ -93,11 +93,33 @@ public class PipePoolTests
...
+            var allocatedSize = buffer.Buffer.Length;
+            buffer.Advance(buffer.Buffer.Length);
+            buffer.Ensure(1);
+            var ensuredSize = buffer.Buffer.Length;
+            await buffer.FlushAsync();
+
+            pipe.Reader.Complete();
+            pipe.Writer.Complete();
+
+            Assert.Equal(2020, ensuredSize);
```
KrzysztofCwalina: Should these ensure that the size is at least 2020, not that it is exactly 2020?

pakrym: It uses mock implementation of pool that returns exactly what asked.

pakrym: So for 512 request it has to return minimum sized segment.

KrzysztofCwalina: Your call on the PR (i.e. my comment is non blocking), but you don't want to test the mocked pool. You just want to make sure small buffers are not returned. Also, the test could be reused with other pool (that don't return exact size), if you changed the check to >=

pakrym: I'm testing pipe implementation and verifying if it requests the right thing from the pool. It's not specific to pool implementation.

PR #2035, opened 2018-01-08
Switch OutputWriter.Write to use ReadOnlySpan<byte>

pakrym:


src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
```diff
@@ -35,79 +35,42 @@ public void Advance(int count)
             _output.Advance(count);
         }
 
-        public void Write(byte[] source)
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Write(Span<byte> source)
```
exyi: Why not use `ReadOnlySpan<byte>`? I think that `Write` is not supposed to write into the `source`

src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
```diff
@@ -35,79 +35,42 @@ public void Advance(int count)
...
-
-                Unsafe.CopyBlockUnaligned(ref pDest, ref pSource, (uint)length);
-
-                Advance(length);
-            }
-            else
-            {
-                WriteMultiBuffer(source, offset, length);
-            }
+            _output.Enlarge(count);
```
KrzysztofCwalina: Not related to this PR, but we need to make sure we standardize Enlarge/Ensure (naming) and what the parameter means.

PR #2044, opened 2018-01-10
Remove ReadOnlyBytes

KrzysztofCwalina:
```<edit>```
I just moved ISlicable to Experimental package and changed the tests to use an adapter. A long term solution would mean ReadOnlyBytes implementing ISlicable, but for now I want to merge this PR to remove ReadOnlyBytes
```</edit>```

This PR shows what it would take to remove ReadOnlyBytes and still support TryReadUntil on BufferReader. TryReadUntill reads till some delimiter and then returns the bytes read. The difference between TryReadUntill and all the other "read" methods we already support is that its result is not fixed-size, i.e. the result is ReadOnlyBytes itself. Other read methods return ints, floats, datetimes, etc.

To summarize: 
1. We would need to add ```ISlicable<T>``` interface. If BufferReader wraps a type implementing ISlicable, the reader would support TryReadUntil; otherwise TryReadUntil would not be available (the extension constraint would not be fulfilled).
2. We would need to expose Sequence property on BufferReader.

It works, but I wish it all was simpler. 

cc: @pakrym , @davidfowl, @joshfree, @terrajobst, @halter73, @ahsonkhan 

BTW, I assume/assert we must support something like TryReadUntil on the reader, i.e. our abstractions are not shippable if we cannot make it work.

Other alternatives we explored (and my commentary on them):

1. Make new ReadOnlyBytes(Position start, Position end) work. It's not clear this is possible without severely constraining what can be done with Position, but maybe it's worth trying.
2. Change BufferReder to be specific to ReadOnlyBuffer, as opposed to being over ISequence. I worry that this will result in less APIs being developed that operate in buffer sequences. With ISequence, I imagine we will develop many transportation and parsing APIs (e.g. compress sequence, parse JSON, etc. ) that operate on ISequence. I worry that ReadOnlyBuffer is too policy heavy to embark on major effort to develop all these APIs that would work just on it.
3. Have TryReadUntil to always incur a data copy. I think this goes too much against our philosophy of non-alloc.

Any other ideas?

### CODE REVIEW

**Comment by ahsonkhan** [comment]
cc @pakrym (not Pranav)

**Comment by KrzysztofCwalina** [comment]
I just moved ISlicable to Experimental package and changed the tests to use an adapter. A long term solution would mean ReadOnlyBytes implementing ISlicable, but for now I want to merge this PR to remove ReadOnlyBytes completly.

@pakrym, are you ok with this merge?

File: src/System.Buffers.Experimental/System/Buffers/Sequences/ReadOnlyBytes.cs [hunk 1]
```diff
@@ -1,383 +1,384 @@
...
-
-        public override string ToString()
-        {
-            var buffer = ToSpan().ToArray();
-            return Encoding.UTF8.GetString(buffer, 0, buffer.Length);
-
-        }
-    }
-}
+﻿//// Copyright (c) Microsoft. All rights reserved.
```
davidfowl: Why is this commented out?

juliusfriedman:
@KrzysztofCwalina, I support the removal of ReadOnlyBytes, please do keep the ISliceable interface, its very valuable especially considering the API underlying Memory is so intricately designed; As so far as ideas, how about a BitsReservoir or more generally applicable in the context of Tensors and Strides a Manifold semantic so I can read interpret and repeat the underlying (erg-in-binary) as necesaatied by the application. Keep up the work for good.

src/System.Buffers.Experimental/System/Buffers/Sequences/ReadOnlyBytes.cs
juliusfriedman: @davidfowl, Sarah Faraphallower said it wasn't required...

juliusfriedman: Won't immediately say I disagree either, my 3 Sat and CRT/JMT logic tables work without it....

KrzysztofCwalina: @davidfowl, something went very wrong with the merge. I really don't understand how the tests passed. Anyway, I meant to delete this whole file, and I will shortly.

KrzysztofCwalina:

KrzysztofCwalina:

PR #2045, opened 2018-01-10
Removed Position.Get

KrzysztofCwalina:
@pakrym, @davidfowl, @terrajobst, @joshfree, @ahsonkhan


joshfree:
lgtm

juliusfriedman:
Uh oh... was at camp crystal lake and Jason was saying that we needed position . Missionary and position . DoggyStyle... Don't wanna forget the signed unsigned goodness and the issues with nint/nuint (nfloat) etc...

Perhaps [sic] (one) should [sic] to solve both issues at the same time [link needed]

PR #2048, opened 2018-01-11
Made IBufferList and ReadOnlyBuffer generic

davidfowl:
- Updated the extension methods to use ReadOnlyBuffer<byte>


src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -238,29 +236,33 @@ public void CopyTo(Span<byte> destination)
...
+                    var sb = new StringBuilder();
+                    foreach (var buffer in bytes)
+                    {
+                        SpanLiteralExtensions.AppendAsLiteral(buffer.Span, sb);
+                    }
+                    return sb.ToString();
+                }
             }
-            return sb.ToString();
+            return base.ToString();
```
pakrym:
```
foreach (buffer in this)
 foreach (item in buffer) 
sb.Append("{0}, ", item);
```

PR #2049, opened 2018-01-11
Implement IOutput on IPipeWriter

pakrym:
This PR is based on other IOutput PR.

### DIFF
--- a/samples/System.IO.Pipelines.Samples/HttpServer/HttpConnection.cs
+++ b/samples/System.IO.Pipelines.Samples/HttpServer/HttpConnection.cs
@@ -142,7 +142,7 @@ namespace System.IO.Pipelines.Samples.Http
 
         private async Task EndResponse()
         {
-            var buffer = _output.Alloc();
+            var buffer = _output;
 
             if (!HasStarted)
             {
@@ -172,7 +172,7 @@ namespace System.IO.Pipelines.Samples.Http
 
         public Task WriteAsync(Span<byte> data)
         {
-            var buffer = _output.Alloc();
+            var buffer = _output;
 
             if (!HasStarted)
             {
@@ -194,12 +194,12 @@ namespace System.IO.Pipelines.Samples.Http
             return FlushAsync(buffer);
         }
 
-        public async Task FlushAsync(WritableBuffer buffer)
+        public async Task FlushAsync(IPipeWriter buffer)
         {
             await buffer.FlushAsync();
         }
 
-        private void WriteBeginResponseHeaders(WritableBuffer buffer)
+        private void WriteBeginResponseHeaders(IPipeWriter buffer)
         {
             if (HasStarted)
             {
@@ -217,7 +217,7 @@ namespace System.IO.Pipelines.Samples.Http
             ResponseHeaders.CopyTo(_autoChunk, buffer);
         }
 
-        private void WriteEndResponse(WritableBuffer buffer)
+        private void WriteEndResponse(IPipeWriter buffer)
         {
             buffer.Write(_chunkedEndBytes);
         }
--- a/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
+++ b/src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
@@ -21,7 +21,6 @@ namespace System.IO.Pipelines.Networking.Libuv
         private volatile bool _stopping;
 
         private Task _sendingTask;
-        private WritableBuffer? _inputBuffer;
         private MemoryHandle _inputBufferPin;
 
         public UvTcpConnection(UvThread thread, UvTcpHandle handle)
@@ -149,8 +148,6 @@ namespace System.IO.Pipelines.Networking.Libuv
                 // A zero status does not indicate an error or connection end. It indicates
                 // there is no data to be read right now.
                 // See the note at http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb.
-                _inputBuffer?.Commit();
-                _inputBuffer = null;
                 UnpinInputBuffer();
 
                 return;
@@ -172,7 +169,7 @@ namespace System.IO.Pipelines.Networking.Libuv
                 handle.Libuv.Check(status, out Exception uvError);
                 error = new IOException(uvError.Message, uvError);
 
-                _inputBuffer?.Commit();
+                _input.Writer.Commit();
                 UnpinInputBuffer();
 
                 // REVIEW: Should we treat ECONNRESET as an error?
@@ -181,18 +178,15 @@ namespace System.IO.Pipelines.Networking.Libuv
             }
             else
             {
-                var inputBuffer = _inputBuffer.Value;
 
-                _inputBuffer = null;
-
-                inputBuffer.Advance(readCount);
-                inputBuffer.Commit();
+                _input.Writer.Advance(readCount);
+                _input.Writer.Commit();
                 UnpinInputBuffer();
 
                 // Flush if there was data
                 if (readCount > 0)
                 {
-                    var awaitable = inputBuffer.FlushAsync();
+                    var awaitable = _input.Writer.FlushAsync();
 
                     if (!awaitable.IsCompleted)
                     {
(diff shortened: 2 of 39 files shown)


samples/System.IO.Pipelines.Samples/HttpServer/HttpConnection.cs
```diff
@@ -194,12 +194,12 @@ public Task WriteAsync(Span<byte> data)
             return FlushAsync(buffer);
         }
 
-        public async Task FlushAsync(WritableBuffer buffer)
+        public async Task FlushAsync(IPipeWriter buffer)
         {
-            await buffer.FlushAsync();
+            await buffer.FlushAsync(default);
```
davidfowl: What is this argument?

src/System.IO.Pipelines.Networking.Libuv/UvTcpConnection.cs
```diff
@@ -149,8 +148,7 @@ private static void ReadCallback(UvStreamHandle handle, int status, object state
                 // A zero status does not indicate an error or connection end. It indicates
                 // there is no data to be read right now.
                 // See the note at http://docs.libuv.org/en/v1.x/stream.html#c.uv_read_cb.
-                _inputBuffer?.Commit();
-                _inputBuffer = null;
+                _input.Writer.Commit();
```
davidfowl: This can be removed right?

pakrym: Yeah

src/System.IO.Pipelines.Text.Primitives/PipelineTextOutput.cs
```diff
@@ -23,34 +22,23 @@ public PipelineTextOutput(IPipeWriter writer, SymbolTable symbolTable)
...
         {
-            _writableBuffer.Advance(bytes);
+            _writer.Advance(bytes);
         }
 
         public Memory<byte> GetMemory(int desiredFreeBytesHint = 0)
         {
-            _writableBuffer.Ensure(desiredFreeBytesHint == 0 ? 2048 : desiredFreeBytesHint);
-            return _writableBuffer.Buffer;
+            return _writer.GetMemory();
```
davidfowl: Pass the argument desiredFreeBytesHint argument to GetMemory?

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -363,7 +344,7 @@ internal ValueAwaiter<FlushResult> FlushAsync(CancellationToken cancellationToke
         /// <param name="exception">Optional Exception indicating a failure that's causing the pipeline to complete.</param>
         void IPipeWriter.Complete(Exception exception)
         {
-            if (_writingState.IsActive)
+            if (_writingState.IsStarted && _currentWriteLength > 0)
```
davidfowl: We should just remove the IsStarted/IsActive flag and use the current write length for everything

pakrym: We still need it for some things - Advance before calling GetMemory, and to prevent returning of the last segment if it was returned from GetMemory and might be written to.

pakrym:
Rebased and updated.

src/System.IO.Pipelines/System/IO/Pipelines/IPipeWriter.cs
```diff
@@ -1,20 +1,16 @@
...
+using System.Buffers;
+using System.Threading;
+
 namespace System.IO.Pipelines
 {
     /// <summary>
     /// Defines a class that provides a pipeline to which data can be written.
     /// </summary>
-    public interface IPipeWriter
+    public interface IPipeWriter: IOutput
```
KrzysztofCwalina: I wonder if we should not rename the type to IAsyncOutput

pakrym: Which type? IPipeWriter?

pakrym:
@dotnet-bot test this please

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -132,64 +132,51 @@ WritableBuffer IPipeWriter.Alloc(int minimumSize)
...
 
                 segment.SetNext(nextSegment);
 
                 _writingHead = nextSegment;
             }
+
+            return Buffer;
         }
 
+        Span<byte> IOutput.GetSpan(int minimumSize) => ((IOutput)this).GetMemory(minimumSize).Span;
```
davidfowl: zomg

src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
davidfowl: Hmm, I wonder if other IOutput implementations also have this problem

src/System.IO.Pipelines/System/IO/Pipelines/PipelinesThrowHelper.cs
```diff
@@ -125,7 +125,7 @@ private static string GetResourceString(ExceptionResource argument, string locat
                     resourceString = "Already writing.";
                     break;
                 case ExceptionResource.NotWritingNoAlloc:
-                    resourceString = "No writing operation. Make sure Alloc() was called.";
+                    resourceString = "No writing operation. Make sure GetMemory() was called.";
```
davidfowl: File an issue for these error messages.

src/System.IO.Pipelines/System/IO/Pipelines/WritableBuffer.cs
```diff
@@ -6,91 +6,17 @@
 
 namespace System.IO.Pipelines
```
davidfowl: 1. Change the name of this file.
2. Can we move this method to System.Buffers.Primitives?

pakrym:
I have second thoughts about this one, the change would cause every small write even when using the writer to do an interface dispatch for Advance. It'll kill perf.

KrzysztofCwalina:
Just like on the reading side, we need a BufferWriter to make it efficient. It will fish out the span, advance a filed index, and then use the interface call to commit a big chunk when the span fills in.

pakrym:
BufferWriter still calls advance every time https://github.com/dotnet/corefxlab/blob/master/src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs#L35.

So if you do

```
for (int i...)
{
   writer.WriteInt(i);
}
```

It would be very inefficient.

KrzysztofCwalina:
Yeah, for this to be devirtualized, the WritableBuffer would have to implement IOutput/IPipeWriter. Alternativelly, the BufferWriter should not call IOutput.Advance but rather advance an internal in and then periodically flush.

pakrym:
If WriteableBuffer implements IOutput it would have to take `Pipe` to be non-virtual and we lose having IPipeWriter to be real abstraction decoupled from Pipe.

Second options would make using BufferWriter harder. You can't even make it disposable because it has to be passed around by-ref. 

We need to figure it out before merging this change.

KrzysztofCwalina:

KrzysztofCwalina:
I say we make IPipeWriter extend IOutput and get rid of WritableBuffer. :-)
Then we rename IPipeWriter to IAsyncOutput and we have a very simple and easy to understand API. As it stands now IPipeWriter give people WritableBuffer that has to be wrapped in a BufferWriter. Seems like we are wrapping, unwrapping, and wrapping again.

pakrym:
> rename IPipeWriter to IAsyncOutput 
It has to much IPipe specific things on it, CancelPending*, Complete, OnCompleted

You didn't day anything about how do we keep performance the same going to interfaces? Forcing people to flush their BufferWriters before flushing pipe?

KrzysztofCwalina:
The writer would wrap IAsynOutput. It would make one call to get the Span buffer then write 4KB incrementing a filed in the writer, then flush. Two interface calls per 4KB written should be fast.

pakrym:
How do you know when to flush Writer if you do
```
for (int i...)
{
   writer.WriteInt(i);
}
```
?

KrzysztofCwalina:
At the end. You can flush sooner to avoid buffers being held.

pakrym:

pakrym:
@KrzysztofCwalina somethign is really broken on OSX

PR #2052, opened 2018-01-16
Address API review comments

pakrym:
https://github.com/dotnet/apireviews/tree/master/2018/System.IO.Pipelines

1. Add `Slice(int, int)` overloads
2. Rename `Seek` to `GetPosition`
4. Rename `VirtualIndex` to `RunningLength`
5. Remove `ROB.ctor(OwnedMemory)` add `ROB.ctor(Memory)`

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
@@ -61,7 +61,7 @@ namespace System.Buffers
             BufferEnd = end;
         }
 
-        public ReadOnlyBuffer(IBufferList<T> startSegment, int offset, IBufferList<T> endSegment, int endIndex)
+        public ReadOnlyBuffer(IMemoryListNode<T> startSegment, int offset, IMemoryListNode<T> endSegment, int endIndex)
         {
             Debug.Assert(startSegment != null);
             Debug.Assert(endSegment != null);
@@ -271,7 +271,7 @@ namespace System.Buffers
             return new Enumerator(this);
         }
 
-        public Position Seek(Position origin, long offset)
+        public Position GetPosition(Position origin, long offset)
         {
             if (offset < 0)
             {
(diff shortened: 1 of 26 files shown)


src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -146,50 +133,71 @@ public ReadOnlyBuffer(IEnumerable<Memory<T>> buffers)
...
+        /// <param name="offset">The index at which to begin this slice.</param>
+        /// <param name="length">The length of the slice</param>
+        public ReadOnlyBuffer<T> Slice(int offset, int length) => Slice((long)offset, length);
+
+        /// <summary>
+        /// Forms a slice out of the given <see cref="ReadOnlyBuffer"/>, beginning at 'start', ending at 'end' (inclusive).
+        /// </summary>
+        /// <param name="offset">The index at which to begin this slice.</param>
+        /// <param name="end">The end (inclusive) of the slice</param>
+        public ReadOnlyBuffer<T> Slice(int offset, Position end) => Slice((long)offset, end);
```
davidfowl: I think this defeats the purpose of the int overload.

pakrym: I was adding it to get surface to the place we want.

pakrym: Fixed.

pakrym:
Any objection to merging this?

/cc @KrzysztofCwalina @davidfowl @joshfree @muratg

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -146,46 +133,71 @@ public ReadOnlyBuffer(IEnumerable<Memory<T>> buffers)
         /// <summary>
         /// Forms a slice out of the given <see cref="ReadOnlyBuffer"/>, beginning at 'start', and is at most length bytes
```
sebastienros: Rename 'start' parameter

sebastienros: Not the only occurence btw.

KrzysztofCwalina:
I like all the changes except IBufferList to IMemoryListNode. I remember a discussion about this, but I was not aware that we made a decision to go through with the rename. I think IMemoryListNode is a decent name from the perspective of the implementer, but not the consumer, and we usually optimize for consumers. The consumer does not care how the datastructure is implemented. In fact, IBufferList can be implemented without any nodes, i.e. as an array.

pakrym:
They still look like nodes of the linked list with `data` and `next` no matter the implementation. What name do you like?

KrzysztofCwalina:
I like IBufferList<T> or ISegmentList<T> with the Memory property being renamed to Segment.

pakrym:
OK, I'll rename it back, for now, so everything else can be merged.

KrzysztofCwalina:
looks good

PR #2059, opened 2018-01-19
Fixed a bug in MoveNext

KrzysztofCwalina:
MoveNext was not correctly updating _currentPosition if the loop moved past an empty buffer. The bug is not observable (as all other APIs would skip the empty buffer), but @pranavkm thought it would be good to fix it to ensure we have the right semantics if we ever add APIs that can observe the difference, e.g. expose CurrentPosition property

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
@@ -92,13 +92,14 @@ namespace System.Buffers
             var previous = _nextPosition;
             while (_sequence.TryGet(ref _nextPosition, out var memory, true))
             {
-                _currentPosition = previous;
                 _currentSpan = memory.Span;
-                _index = 0;
                 if (_currentSpan.Length > 0)
                 {
+                    _index = 0;
+                    _currentPosition = previous;
                     return;
                 }
+                previous = _nextPosition;
             }
             _end = true;
         }


src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -92,13 +92,14 @@ private void MoveNext()
             var previous = _nextPosition;
             while (_sequence.TryGet(ref _nextPosition, out var memory, true))
             {
-                _currentPosition = previous;
                 _currentSpan = memory.Span;
-                _index = 0;
                 if (_currentSpan.Length > 0)
                 {
+                    _index = 0;
```
pakrym: Hmmmm, what if the last segment of the sequence is empty and we don't clear index? I think this code should be outside if.

pakrym:
```
[xUnit.net 00:00:24.4949470] System.IO.Pipelines.Tests.ReadableBufferReaderFacts+OwnedMemory.CopyToLargerBufferWorks [FAIL]

``` 

The issue didn't seem to fix itself.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
KrzysztofCwalina: Then we don't also change the current span and so the index points to the end of the span, and in turn Unread is empty. The problem with the current (before this PR code) is that MoveNext might reset the index to zero, but not reset _currentIndex, which will make the last segment available to read again (unless the reader checks the _isEnd bit).

KrzysztofCwalina:

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
pakrym: After this change, if the last segment is empty we wouldn't move to it, and GetPosition would have to traverse again. Not a big deal but if we are optimizing re-walking we should do it all way.

PR #2066, opened 2018-01-24
Use Position.Index sign bit for buffer type

pakrym:


src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
```diff
@@ -389,5 +397,35 @@ public void Reset()
...
+                begin.Index | (Start.Index & ~IndexBitMask),
+                end.Segment,
+                end.Index | (End.Index & ~IndexBitMask)
+            );
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private  BufferType GetBufferType()
+        {
+            return (BufferType)((((uint)Start.Index >> 30) & 2) | (uint)End.Index >> 31);
```
pakrym: This is really ugly. I'm sure there is a better way.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer_helpers.cs
```diff
@@ -81,57 +91,109 @@ internal static bool TryGetBuffer(Position begin, Position end, out ReadOnlyMemo
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        internal static Position Seek(Position begin, Position end, int bytes, bool checkEndReachable = true)
+        internal Position Seek(Position start, Position end, int bytes, bool checkEndReachable = true)
```
benaadams: These `AggressiveInlining`s are going to be ugly...

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
benaadams: Use the low bits instead?
```csharp
return (BufferType)((Start.Index & 1) << 1) | End.Index & 1);
```
Then to get value its 
```csharp
private static int GetIndex(int index)
{
    return index >> 1;
}
```
Add flagging; shift 1 + or
```csharp
BufferStart = new Position(first, (0 << 1) | MemoryListStartFlag);
BufferEnd = new Position(segment, (segment.Memory.Length << 1) | MemoryListEndFlag);
```

pakrym: Interesting idea. I guess we don't really care which bit to use.

`return index >> 1;` wouldn't shift cause problems if we set sign bit? It would start filling with ones instead of zeros.

benaadams: Store it as a `uint`?

tests/System.Text.Http.Parser.Tests/HttpParserTests.cs
```diff
@@ -373,8 +373,8 @@ public void ParseRequestLineSplitBufferWithoutNewLineDoesNotUpdateConsumed()
             Assert.Equal(1, pairs.Length);
             Assert.Equal(headerName, pairs[0].Key);
             Assert.Equal(expectedHeaderValue, pairs[0].Value);
-            Assert.Equal(buffer.End, consumed);
```
pakrym: This is an interesting side effect of the change: Positions returned from GetPosition do not have type information in them (we need 2 bits for it) so cannot be directly compared to other cursors (Start/End)

davidfowl:
Need performance numberz

pakrym:
Before:
```

               Method |     Mean |    Error |    StdDev |        Op/s | Scaled | ScaledSD | Allocated |
--------------------- |---------:|---------:|----------:|------------:|-------:|---------:|----------:|
 PlaintextTechEmpower | 455.3 ns | 5.297 ns |  7.929 ns | 2,196,536.6 |   1.00 |     0.00 |       0 B |
           LiveAspNet | 714.4 ns | 3.951 ns |  5.913 ns | 1,399,774.5 |   1.57 |     0.03 |       0 B |
              Unicode | 964.1 ns | 9.115 ns | 13.642 ns | 1,037,251.1 |   2.12 |     0.05 |       0 B |
```
After:
```
               Method |     Mean |     Error |     StdDev |        Op/s | Scaled | ScaledSD | Allocated |
--------------------- |---------:|----------:|-----------:|------------:|-------:|---------:|----------:|
 PlaintextTechEmpower | 396.0 ns |  2.545 ns |   3.810 ns | 2,525,498.8 |   1.00 |     0.00 |       0 B |
           LiveAspNet | 675.0 ns | 88.328 ns | 132.205 ns | 1,481,550.8 |   1.70 |     0.33 |       0 B |
              Unicode | 904.8 ns | 18.496 ns |  27.684 ns | 1,105,185.6 |   2.29 |     0.07 |       0 B |
```

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBuffer.cs
benaadams: `>> 1` is less ugly than `& 0x7FFFFFFF` e.g. ` & IndexBitMask`
And
`& 1` is less ugly than `& 0x80000000` e.g. ` & ~IndexBitMask`

KrzysztofCwalina:
decent win!

davidfowl:
@dotnet-bot build Innerloop OSX10.12

davidfowl:
@dotnet-bot build OSX10.12

KrzysztofCwalina:

PR #2085, opened 2018-01-29
Fix issue where GetPosition might cross outside of ROB

pakrym:
Also simplify advance logic a bit


tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
```diff
@@ -140,7 +140,7 @@ public void SegmentStartIsConsideredInBoundsCheck()
         }
 
         [Fact]
-        public void MovePrefersNextSegment()
```
halter73: Nit: might as well rename other test methods containing "Move" in the name.

pakrym: Fixed

PR #2088, opened 2018-01-30
API Cleanup

pakrym:
1. Move WriteAsync to PipeWriter
2. Move PosisionOf to extension class
3. Remove Get<> from Position


src/System.IO.Pipelines/System/IO/Pipelines/PipeWriter.cs
```diff
@@ -33,5 +36,24 @@ public abstract class PipeWriter: IOutput
         public abstract void Advance(int bytes);
         public abstract Memory<byte> GetMemory(int minimumLength = 0);
         public abstract Span<byte> GetSpan(int minimumLength = 0);
+
+        public virtual Task WriteAsync(ReadOnlyMemory<byte> source)
```
pakrym: Not entirely sure about this, maybe it needs to return `ValueTask<FlushResult>`

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -323,7 +323,7 @@ public HttpParser(bool showErrorDetails)
                                 var current = reader.Position;
                                 var subBuffer = buffer.Slice(reader.Position);
                                 // Split buffers
-                                var lineEnd = subBuffer.PositionOf(ByteLF);
+                                var lineEnd = ReadOnlyBufferExtensions.PositionOf(subBuffer, ByteLF);
```
davidfowl: ?

src/System.IO.Pipelines/System/IO/Pipelines/PipeWriter.cs
davidfowl: Yea, lets do ValueAwaiter<FlushResult>

src/System.Text.Http.Parser/HttpParser.cs
pakrym: It compares two implementations of `PositionOf` on ISequence and ROB, now when both of them are extension methods invocation is ambiguous

src/System.IO.Pipelines/System/IO/Pipelines/PipeWriter.cs
```diff
@@ -11,6 +12,8 @@ namespace System.IO.Pipelines
     /// </summary>
     public abstract class PipeWriter: IOutput
     {
+        private static readonly Task _completedTask = Task.FromResult(0);
```
davidfowl: Remove this.

src/System.IO.Pipelines/System/IO/Pipelines/PipeWriter.cs
pakrym: Fixed

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferExtensions.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+namespace System.Buffers
+{
+    public static class ReadOnlyBufferExtensions
```
KrzysztofCwalina: If we want to have this as an extension member, I would prefer to have it on MemoryExtensions class (an existing extensions method holder). I don't think we should be adding several extension method holders.

pakrym: We can move it to that type during corefx move.

KrzysztofCwalina: ok

src/System.Text.Http.Parser/HttpParser.cs
davidfowl: Isn't this a problem?

pakrym: No, the other extension method doesn't ship it's in buffers.Experimental

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferExtensions.cs
pakrym: They are in different namespaces. MemoryExtensions is in System.

PR #2090, opened 2018-01-31
Fix date formatting

jkotas:
This test was consistently failing on my Mac when I tried to repro the memory corruption. My Mac ended up with non-standard timezone offset somehow. This code was not handling non-standard timezone offsets correctly. I have just copied the logic from matching productized corefx utf8 formatter.

### DIFF
--- a/src/System.Text.Primitives/System/Text/Formatters/Utf16Formatter_time.cs
+++ b/src/System.Text.Primitives/System/Text/Formatters/Utf16Formatter_time.cs
@@ -199,17 +199,16 @@ namespace System.Buffers.Text
 
             if (kind == DateTimeKind.Local)
             {
-                int hours = offset.Hours;
                 char sign = Plus;
 
-                if (offset.Hours < 0)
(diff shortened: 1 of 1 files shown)


src/System.Text.Primitives/System/Text/Formatters/Utf16Formatter_time.cs
```diff
@@ -199,17 +199,16 @@ private static bool TryFormatO(DateTime value, TimeSpan offset, Span<byte> buffe
 
             if (kind == DateTimeKind.Local)
             {
-                int hours = offset.Hours;
                 char sign = Plus;
 
-                if (offset.Hours < 0)
+                if (offset < default(TimeSpan) /* a "const" version of TimeSpan.Zero */)
```
joshfree: is there an actual significant performance win for using default(TimeSpan) in place of TimeSpan.Zero?  If so, is it worth creating a codegen bug on this to see what improvements could be made here?

Issue #2087, opened 2018-01-30
FlushResult needs a constructor


davidfowl:
Today it doesn't have any and can't be constructed with IsCompleted or IsCancelled.

PR #2095, opened 2018-02-02
Make FlushResult constructable

davidfowl:
- Added a constructor to allow setting IsCompleted and IsCancelled

#2087


src/System.IO.Pipelines/System/IO/Pipelines/FlushResult.cs
```diff
@@ -7,6 +7,21 @@ public struct FlushResult
     {
         internal ResultFlags ResultFlags;
```
pakrym: Do we still set this directly? Make it private if not.

davidfowl: We do.

davidfowl:
@pakrym Yea, test, I'll add one.

PR #2097, opened 2018-02-04
Pass BufferReader around as readonly byref

jkotas:
BufferReader is a big struct. Passing big structs by value produces bloated inefficient code. Now that this bloated inefficient served its purpose by finding stress bug in the runtime, we can use readonly byref that is much more appropriate in this case.

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
@@ -13,9 +13,9 @@ namespace System.Buffers
             return new BufferReader<TSequence>(buffer);
         }
 
-        public static int Peek<TSequence>(BufferReader<TSequence> reader, Span<byte> destination)
+        public static int Peek<TSequence>(in BufferReader<TSequence> reader, Span<byte> destination)
             where TSequence : ISequence<ReadOnlyMemory<byte>>
-            => BufferReader<TSequence>.Peek(reader, destination);
+            => BufferReader<TSequence>.Peek(in reader, destination);
     }
 
     public ref struct BufferReader<TSequence> where TSequence : ISequence<ReadOnlyMemory<byte>>
@@ -135,7 +135,7 @@ namespace System.Buffers
             }
         }
 
-        internal static int Peek(BufferReader<TSequence> bytes, Span<byte> destination)
+        internal static int Peek(in BufferReader<TSequence> bytes, Span<byte> destination)
         {
             var first = bytes.UnreadSegment;
             if (first.Length > destination.Length)
--- a/src/System.Buffers.Experimental/System/Buffers/BufferReader_binary.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BufferReader_binary.cs
@@ -27,7 +27,7 @@ namespace System.Buffers
             }
 
             Span<byte> tempSpan = stackalloc byte[4];
-            var copied = BufferReader.Peek(reader, tempSpan);
+            var copied = BufferReader.Peek(in reader, tempSpan);
             if (copied < 4)
             {
                 value = default;
--- a/src/System.Buffers.Experimental/System/Buffers/BufferReader_text.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/BufferReader_text.cs
@@ -22,7 +22,7 @@ namespace System.Buffers
             }
 
             Span<byte> tempSpan = stackalloc byte[5];
-            var copied = BufferReader.Peek(reader, tempSpan);
+            var copied = BufferReader.Peek(in reader, tempSpan);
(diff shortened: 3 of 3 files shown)


jkotas:
CopyToLargerBufferWorks test will stop hitting https://github.com/dotnet/coreclr/issues/16044 as side-effect of this change.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
```diff
@@ -135,7 +135,7 @@ public void Advance(int byteCount)
             }
         }
 
-        internal static int Peek(BufferReader<TSequence> bytes, Span<byte> destination)
+        internal static int Peek(in BufferReader<TSequence> bytes, Span<byte> destination)
         {
             var first = bytes.UnreadSegment;
```
benaadams: Will cause copy when passed `in`? (as `BufferReader<TSequence>` is not readonly)

Should the property be expanded here to work on the fields

jkotas: Hmmm, you are right. This just moves the inefficient copies of the whole BufferReader to different places. I do not know how to avoid them with this API shape. @KrzysztofCwalina Thoughts?

jkotas: > Think the only change needed is

There are still going to be copies where Peek is called. For example, `var copied = BufferReader.Peek(reader, tempSpan);` will have a copy. You can tell that it has a copy because of if you explicitly pass it as `in reader`, it won't compile anymore.

>  C# compiler/analyser to issue a warning that its going to do a copy of a readonly variable 

Agree. It is a pit of failure that it is very easy to fall into, with `in` in particular.

KrzysztofCwalina: Can the parameter by by ref (as opposed to in)?

BTW, This method is a hack in general. I originally had it as an instance method but after changing the struct to be a by-ref struct, the compiler started to complain that I cannot pass stack allocated spans to it. @VSadov suggested this fix (hack?) and when I talked to @jaredpar he suggested that at some point we can add readonly methods to the language.

jkotas: When I made the parameter byref, the compiler started complain about passing spans to it.

tannergooding: IIRC, `in` will only cause a copy if you don't explicitly specify.

That is `Peek(bytes, destination)` may cause a copy, but `Peek(in bytes, destination)` will elide the copy as you explicitly opted in.

@jcouv or @VSadov may remember better 😄

jkotas: > `Peek(in bytes, destination)`

The problem is this does not compile. It fails to compile with:

```
System\Buffers\BufferReader_text.cs(73,26): error CS8350: This combination of arguments to 'BufferReader.Peek<TSequence>(in BufferReader<TSequence>, Span<byte>)' is disallowed because it may expose variables referenced by parameter 'destination' outside of their declaration scope
```

VSadov:
The problem is that you have a large mutable ref-like struct that is _returnable_ and you want to call a method (instance or otherwise) that takes that struct by reference and also takes another unreturnable ref-like struct.

The only way to guarantee that it is safe is to convey that no mutations could actually happen to the large instance. Otherwise the call could as well directly or indirectly stash the unreturnable value into the large instance.

We do not have a way to say "this instance method alone treats `this` " as `readonly`, but it is expressible for extension methods. If we had, it would have same issues with copying though. Invoking other not readonly methods on `this` would have to copy the receiver.

src/System.Buffers.Primitives/System/Buffers/ReadOnlyBufferReader.cs
VSadov: Re; warning on implicit copying. FWIW this was a behavior since v1. Invoking a regular (and so formally mutating) method on a readonly variable does not warn/error, but instead makes a copy of the receiver. 

I am not happy about it but I see how it was considered better than making things like "myStruct.ToString" a warning/error.

VSadov: Yes, 

Replacing
```C#
var first = bytes.UnreadSegment;
```
with
```C#
var first = bytes._currentSpan.Slice(bytes._index);
```
would eliminate a copy here.

Or you can make `UnreadSegment` an `in` static method...

ahsonkhan:
@dotnet-bot test this please

ahsonkhan:
@jkotas, what are your plans for this PR?

jkotas:
No plans.

Issue #2098, opened 2018-02-04
MemoryPool shouldn't discard space per block


benaadams:
As MemoryPool's blocks are page sized and page aligned and the cpu doesn't prefetch across a page boundary the 64 byte buffer is not required.

Having the full 4096 bytes available is helpful when a pipe is used as output/input to a file which also reads and writes in 4096 chunks. https://github.com/aspnet/KestrelHttpServer/pull/2049

ahsonkhan:
MemoryPool lives in corefx now. Can you please move the issue there? Thanks!
https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Buffers/MemoryPool.cs#L10

benaadams:
That's `MemoryPool<T>` rather than `MemoryPool : MemoryPool<byte>`

davidfowl:
This got moved and done in ASP.NET

PR #2099, opened 2018-02-04
Use full page for MemoryPool blocks

benaadams:
CPUs doesn't prefetch across page boundaries so the cache line sized buffer isn't required.

Helps when reading from files which like page sized reads. 

Also uses the page size on .NET Core and .NET Standard rather than the const 4096

Resolves https://github.com/dotnet/corefxlab/issues/2098

/cc @davidfowl @pakrym @halter73

### DIFF
--- a/src/System.IO.Pipelines/System/Buffers/MemoryPool.cs
+++ b/src/System.IO.Pipelines/System/Buffers/MemoryPool.cs
@@ -12,17 +12,17 @@ namespace System.Buffers
     /// </summary>
     public class MemoryPool : MemoryPool<byte>
     {
+#if NETSTANDARD1_1
         /// <summary>
         /// The gap between blocks' starting address. 4096 is chosen because most operating systems are 4k pages in size and alignment.
         /// </summary>
-        private const int _blockStride = 4096;
-
+        private const int _pageSize = 4096;
+#else
         /// <summary>
-        /// The last 64 bytes of a block are unused to prevent CPU from pre-fetching the next 64 byte into it's memory cache.
-        /// See https://github.com/aspnet/KestrelHttpServer/issues/117 and https://www.youtube.com/watch?v=L7zSU9HI-6I
+        /// The gap between blocks' starting address.
         /// </summary>
-        private const int _blockUnused = 64;
-
+        private static readonly int _pageSize = Environment.SystemPageSize;
+#endif
         /// <summary>
         /// Allocating 32 contiguous blocks per slab makes the slab size 128k. This is larger than the 85k size which will place the memory
         /// in the large object heap. This means the GC will not try to relocate this array, so the fact it remains pinned does not negatively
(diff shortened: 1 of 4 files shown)


benaadams:
Haunting me; not sure why this change should effect it
```
The active test run was aborted. Reason: FailFast: 
17:35:41 *** Block being garbage collected instead of returned to pool ***
17:35:41    at System.Diagnostics.Debug.Assert(Boolean condition, String message, String detailMessage)
17:35:41    at System.Buffers.MemoryPoolBlock.Finalize() in D:\j\workspace\windows_nt_de---4b678204\src\System.IO.Pipelines\System\Buffers\MemoryPoolBlock.cs:line 65
17:35:41    at System.Environment.FailFast(System.String, System.Exception)
17:35:41    at System.Buffers.MemoryPoolBlock.Finalize()
```

pakrym:
I don't think it's your change.

We are going to remove MemoryPool from corefxlab and move it back to Kestrel, I'll include this edit in kestrel PR.

src/System.IO.Pipelines/System/Buffers/MemoryPool.cs
```diff
@@ -12,39 +12,34 @@ namespace System.Buffers
...
         /// larger values can be leased but they will be disposed after use rather than returned to the pool.
         /// </summary>
-        public override int MaxBufferSize { get; } = _blockLength;
+        public override int MaxBufferSize { get; } = _pageSize;
 
         /// <summary>
         /// 4096 * 32 gives you a slabLength of 128k contiguous bytes allocated per slab
         /// </summary>
-        private const int _slabLength = _blockStride * _blockCount;
+        private static readonly int _slabLength = _pageSize * _blockCount;
```
halter73: What are the possible values for _pageSize? I assume it will never be less than 4096, but I want to be sure of this so we don't end up with a _slabLength so short that it doesn't end up in the LOH. We should probably update some comments that assume a 4096 block stride.

benaadams: Page size is always a power of 2 as its integral for determining what page an address is in (OS/CPU)

Don't know if could be less than 4096; perhaps
```csharp
private static readonly int _pageSize = Math.Max(4096, Environment.SystemPageSize);
``` 
Would still be valid for pages sizes less than 4096. However, might want to also adjust number of blocks; incase someone has set page size to 2MB

benaadams: Huge pages may become more of a thing on linux due to Meltdown; so should cap the multiplier so it doesn't allocate a vast amount per slab

pakrym:
I incorporated this change into https://github.com/aspnet/Common/pull/316

davidfowl:
@pakrym that's brave...

pakrym:
If things stop working it's easy to undo.

benaadams:
@pakrym is wise

PR #2102, opened 2018-02-06
Return write head if completed after GetMemory

pakrym:


src/System.IO.Pipelines/System/IO/Pipelines/Pipe.cs
```diff
@@ -615,7 +615,9 @@ private void CompletePipe()
 
                 _disposed = true;
                 // Return all segments
-                var segment = _readHead;
+                // if _readHead is null we need to try return _commitHead
+                // because there might be a block allocated for writing
+                var segment = _readHead ?? _commitHead;
```
davidfowl: Shouldn't this be `_writingHead`?

pakrym: No, `_writingHead` points to the last segment of current write. `_commitHead` on other hand points to the first one, in case if we attached more then one.

PR #2111, opened 2018-02-10
Update package version, react to ReadOnlyBuffer changes, and fix tests

ahsonkhan:
Supersedes https://github.com/dotnet/corefxlab/pull/2110

Updating corefxlab to react to changes from https://github.com/dotnet/corefx/pull/26694 and https://github.com/dotnet/corefx/pull/26807

cc @dotnet/corefxlab-contrib, @pakrym, @eerhardt, @davidfowl


tests/System.Numerics.Tensors.Tests/TensorSliceTests.cs
```diff
@@ -322,7 +322,6 @@ public void SliceIndexOutOfRange(TensorConstructor constructor)
         {
             var slice = Get3DSlice(constructor);
 
-            Assert.Throws<ArgumentNullException>(() => slice[(int[])null]);
```
ahsonkhan: @eerhardt, please note this change. Is this fine?

tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
```diff
@@ -106,14 +96,6 @@ public void ReadableBufferGetPosition_DoesNotAlowNegative()
             Assert.Throws<ArgumentOutOfRangeException>(() => buffer.GetPosition(buffer.Start, -1));
         }
 
-        [Fact]
-        public void ReadCursorSeekChecksEndIfNotTrustingEnd()
```
ahsonkhan: These tests are testing APIs that no longer exist (in the public surface area).

tests/System.Buffers.Primitives.Tests/BasicUnitTests.cs
```diff
@@ -116,34 +116,6 @@ public void ByteReadOnlySpanCtorWithRangeValidCases(byte[] bytes, int start, int
             ReadOnlySpan<byte> span = new ReadOnlySpan<byte>(bytes, start, length);
         }
 
-        [Fact]
-        public void ByteSpanCtorWithRangeThrowsArgumentExceptionOnNull()
```
ahsonkhan: Already covered with tests like https://github.com/dotnet/corefx/blob/master/src/System.Memory/tests/Span/CtorArray.cs#L72

tests/System.Buffers.Experimental.Tests/StringSliceTests.cs
```diff
@@ -59,15 +59,6 @@ public static void StringSliceIntIntPastEnd()
             Assert.Equal(0, span.Length);
         }
 
-        [Fact]
-        public static void StringSliceNullChecked()
```
ahsonkhan: Already covered with https://github.com/dotnet/corefx/blob/master/src/System.Memory/tests/ReadOnlySpan/AsReadOnlySpan.cs#L119

src/System.IO.Pipelines.Text.Primitives/PipelineTextBufferWriter.cs
```diff
@@ -11,7 +11,6 @@ namespace System.IO.Pipelines.Text.Primitives
     public class PipelineTextBufferWriter : ITextBufferWriter
     {
         private readonly PipeWriter _writer;
-        private bool _needAlloc = true;
```
ahsonkhan: This wasn't being used anywhere. @pakrym, is this change ok?

tests/System.Buffers.Primitives.Tests/ReadableBufferFacts.cs
```diff
@@ -76,12 +66,12 @@ public void ToArrayIsCorrect(int length)
             Assert.Equal(data, buffer.ToArray());
         }
 
-        [Theory]
+        [Theory(Skip = "Need to fix the exception being thrown for various scenarios. Some ReadableBufferFacts throw InvalidOperationException")]
         [MemberData(nameof(OutOfRangeSliceCases))]
-        public void ReadableBufferDoesNotAllowSlicingOutOfRange(Action<ReadOnlyBuffer<byte>> fail)
+        public void ReadableBufferDoesNotAllowSlicingOutOfRange(Action<ReadOnlySequence<byte>> fail)
```
ahsonkhan: @pakrym, @davidfowl - we need to fix this test based on the recent changes.

I have disabled this test for now to get us unblocked, but we have some ReadableBufferFacts that throw InvalidOperationException and some that throw ArgumentOutOfRangeException (for various scenarios).

For example, we throw InvalidOperationException here: https://github.com/dotnet/corefx/blob/master/src/System.Memory/src/System/Buffers/ReadOnlySequence_helpers.cs#L47

This should probably be tested in corefx (see https://github.com/dotnet/corefx/issues/26768)

src/System.Text.Http.Parser/HttpParser.cs
```diff
@@ -446,7 +446,7 @@ public HttpParser(bool showErrorDetails)
                                 return false;
                             }
 
-                            var headerSpan = buffer.Slice(index, end - index + 1).ToSpan();
+                            var headerSpan = Sequence.ToSpan(buffer.Slice(index, end - index + 1));
```
davidfowl: I don't think this is right. There was another ToSpan

ahsonkhan: Ah yes, https://github.com/dotnet/corefxlab/blob/71ea1a077c6852a2869909117828c66822051b74/src/System.Text.Http.Parser/PipelineExtensions.cs#L12.

I will update this to call the right ToSpan method. Do we need to update it anywhere else?

Other places use: https://github.com/dotnet/corefxlab/blob/71ea1a077c6852a2869909117828c66822051b74/src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs#L12

tests/System.Numerics.Tensors.Tests/TensorSliceTests.cs
eerhardt: We pass this arg directly into Span's constructor.  What does Span do now when you give it a null `int[]`?

I think we should keep this test, and add an explicit check for null in the indexer.  No matter what, the test is valuable to ensure the behavior of passing null into this method is as expected.

---
In reply to: [167384904](https://github.com/dotnet/corefxlab/pull/2111#discussion_r167384904) [](ancestors = 167384904)

ahsonkhan: Passing null to Span returns `default(Span<T>)`

eerhardt: The test isn't testing passing a null to Span.  The test is testing passing a null to the indexer of Tensor<T>.  It just so happens that the implementation of Tensor<T> is directly passing it into a Span.  But the test doesn't care what it does under the covers.

ahsonkhan: Would it hurt performance too much if we added a null check to the indexer?

I will add back the test case and update the indexer to check for null.

PR #2113, opened 2018-02-10
Reduce OutputWriter's virtual calls and Memory slicing

benaadams:
Currently every Write and Advance call makes a virtual call to Pipe and also Slices the underlining Memory; mostly defeating the purpose of using the Span.

This leads to the explosion of calls to `BufferSegment.set_End` via virtual `Advance` from writing output headers (once per name, once per value):

image

Change to only Advance when `Commit()` is called or the Span runs out of space. 

`Commit()` must be called after a set of Writing to "flush" (Advance) any buffered data to pipe

/cc @davidfowl

### DIFF
--- a/src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
@@ -71,7 +71,7 @@ namespace System.Buffers
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private void EnsureMore(int count = 1)
+        private void EnsureMore(int count = 0)
         {
             if (_buffered > 0)
             {
--- a/src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
+++ b/src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
@@ -18,9 +18,11 @@ namespace System.Buffers
     {
         private T _output;
         private Span<byte> _span;
+        private int _buffered;
 
         public OutputWriter(T output)
         {
+            _buffered = 0;
             _output = output;
             _span = output.GetSpan();
         }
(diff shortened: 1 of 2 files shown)


benaadams:
This wasn't the error I was expecting...
```
09:48:05 dotnet.exe not installed, downloading and installing.
09:48:05 dotnet-install: Downloading link: https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300-preview2-008171/dotnet-sdk-2.1.300-preview2-008171-osx-x64.tar.gz
09:49:08 curl: (56) SSLRead() return error -9806
09:49:08 dotnet-install: Cannot download: https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300-preview2-008171/dotnet-sdk-2.1.300-preview2-008171-osx-x64.tar.gz
09:49:08 dotnet-install: Downloading alternate link: https://dotnetcli.azureedge.net/dotnet/Sdk/2.1.300-preview2-008171/dotnet-dev-osx-x64.2.1.300-preview2-008171.tar.gz
09:49:08 curl: (22) The requested URL returned error: 404 Not Found
09:49:08 Failed to install latest dotnet.exe, exit code 1, aborting build.
```

benaadams:
e.g.
```diff
public void WriteResponseHeaders(int statusCode, string reasonPhrase, HttpResponseHeaders responseHeaders)
{
    lock (_contextLock)
    {
        if (_completed)
        {
            return;
        }

        var buffer = _pipeWriter;
        var writer = OutputWriter.Create(buffer);

        writer.Write(_bytesHttpVersion11);
        var statusBytes = ReasonPhrases.ToStatusBytes(statusCode, reasonPhrase);
        writer.Write(statusBytes);
        responseHeaders.CopyTo(ref writer);
        writer.Write(_bytesEndHeaders);
+
+       writer.Advance();
        buffer.Commit();
    }
}
```

src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
```diff
@@ -48,9 +61,23 @@ public void Write(ReadOnlySpan<byte> source)
...
         public void Ensure(int count = 1)
         {
+            if (_span.Length < count)
+            {
+                EnsureMore(count);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void EnsureMore(int count = 1)
```
pakrym: count can be 0

benaadams:
Yeah, wasn't really sure about name. Isn't truly a commit or a flush (at the pipe level)

pakrym:
It would be fine if we drop Commit from pipe.

src/System.Buffers.Primitives/System/Buffers/OutputWriter.cs
benaadams: Does `_output.GetMemory(0)` always return `span.Length >= 1`?

pakrym: Yes, since we don't do `Append` anymore there is no reason to return empty when someone asks for memory.

PR #2118, opened 2018-02-13
AzCopy-like sample using non-allocating APIs and pipelines

KrzysztofCwalina:
@joshfree, @davidfowl, @JeffreyRichter, @ahsonkhan, @shiftylogic, @terrajobst, @davidsh, @stephentoub, @rmkerr 

@Davidsh, @rmkerr, you might want to take a look how this sample avoids lots of allocations during requests and responses, e.g. only headers the code cares about are parsed. Of course, the programming model is not as convinient as HttpClient, but if you want to optimize as the cost of usability, such API might be quite nice. See ```StorageResponse``` struct for details. Similarly, headers writing (in request) is done with close to zero allocations, see ```GetFileRequest``` struct.

### DIFF
--- a/samples/AzCopyCore/AzCopyCore.sln
+++ b/samples/AzCopyCore/AzCopyCore.sln
@@ -1,0 +1,28 @@
+﻿
+Microsoft Visual Studio Solution File, Format Version 12.00
+# Visual Studio 15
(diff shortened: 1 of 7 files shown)


davidfowl:
👏

PR #2122, opened 2018-02-14
Updated AzCopyCore sample to new corfxlab APIs

KrzysztofCwalina:


samples/AzCopyCore/AzCopyCore/AzCopyCore.csproj
```diff
@@ -17,13 +17,13 @@
...
 
   <ItemGroup>
-    <PackageReference Include="System.Azure.Experimental" Version="0.1.0-preview2-180209-2" />
-    <PackageReference Include="System.Buffers.Experimental" Version="0.1.0-preview2-180209-2" />
-    <PackageReference Include="System.IO.Pipelines" Version="0.1.0-preview2-180209-2" />
-    <PackageReference Include="System.Memory" Version="4.5.0-preview2-26209-05" />
-    <PackageReference Include="System.Text.Http" Version="0.1.0-preview2-180209-2" />
-    <PackageReference Include="System.Text.Http.Parser" Version="0.1.0-preview2-180209-2" />
-    <PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.0-preview2-26209-05" />
+    <PackageReference Include="System.Azure.Experimental" Version="0.1.0-preview2-180213-4" />
```
ahsonkhan: Consider adding it to the solution and reference the libraries directly (we have some samples which are written this way already). This would remove the need to keep updating these versions.

KrzysztofCwalina: I kind like the samples being standalone apps as opposed to having to build them with the whole repo.

ahsonkhan: I am fine with that. Should we add this standalone sample to the build script though so it gets built as part of the CI run?

KrzysztofCwalina: Yeah, I will do it tomorrow.

PR #2137, opened 2018-02-21
Remove System.IO.Pipelines

pakrym:


src/System.IO.Pipelines.Text.Primitives/PipelineTextBufferWriter.cs
```diff
@@ -32,6 +32,8 @@ public Memory<byte> GetMemory(int desiredFreeBytesHint = 0)
 
         public Span<byte> GetSpan(int minimumLength) => GetMemory(minimumLength).Span;
 
+        public int MaxBufferSize { get; } = Int32.MaxValue;
```
davidfowl: _writer.MaxBufferSize

src/System.Text.Formatting/System/Text/Formatting/Formatters/SequenceFormatter.cs
```diff
@@ -108,5 +109,7 @@ void IBufferWriter.Advance(int bytes)
 
             _totalWritten += bytes;
         }
+
+        public int MaxBufferSize { get; } = Int32.MaxValue;
```
KrzysztofCwalina: What's the contract of MaxBufferSize? i.e. I don't think consumers of this implementation can call GetSpan with int.MaxValue and expect it to work. In general, are we happy with the MaxBufferSize property being added to IBufferWriter? I have second thoughts about it.

@ahsonkhan, @GrabYourPitchforks, @joshfree

tools/dependencies.props
```diff
@@ -3,10 +3,11 @@
...
     <RoslynVersion>2.6.0-beta3-62316-02</RoslynVersion>
-    <SystemMemoryVersion>4.5.0-preview2-26221-01</SystemMemoryVersion>
-    <SystemCompilerServicesUnsafeVersion>4.5.0-preview2-26221-01</SystemCompilerServicesUnsafeVersion>
-    <SystemNumericsVectorsVersion>4.5.0-preview2-26221-01</SystemNumericsVectorsVersion>
-    <SystemBuffersVersion>4.5.0-preview2-26221-01</SystemBuffersVersion>
+    <SystemMemoryVersion>4.5.0-preview2-26221-05</SystemMemoryVersion>
+    <SystemCompilerServicesUnsafeVersion>4.5.0-preview2-26221-05</SystemCompilerServicesUnsafeVersion>
+    <SystemNumericsVectorsVersion>4.5.0-preview2-26221-05</SystemNumericsVectorsVersion>
+    <SystemBuffersVersion>4.5.0-preview2-26221-05</SystemBuffersVersion>
+    <SystemIOPipelinesVersion>4.5.0-preview2-26221-05</SystemIOPipelinesVersion>
```
ahsonkhan: Can you update the program that auto-updates the versions to include this? See https://github.com/dotnet/corefxlab/blob/master/scripts/update-dependencies/Program.cs#L98

src/System.Text.Formatting/System/Text/Formatting/Formatters/SequenceFormatter.cs
pakrym: MaxBufferSize is to be used only in `Math.Min(required, MaxBufferSize)` expression.

pakrym: On the other hand, it's nice to have sensible default sometimes.

KrzysztofCwalina: if required > MaxBufferSize, the required buffer will not be created -> bug
If required < MaxBufferSize, you don't need MaxBufferSize.

pakrym: What's the semantics of MemoryPool.MaxBufferSize? This is pretty much the same.

KrzysztofCwalina: I don't understand this comment. 
What I am arguing is that I think the property is useless: there are lots of examples where we return int.MaxValue, yet the caller won't be able to get int.MaxValue-sized buffers. And so, the property just creates an illusion of robust code, while in reality there are failures lurking all over the place.

pakrym: I'm saying that MemoryPool<byte> abstraction has exactly same property with exactly same semantics and if we are fixing things both should be fixed.

KrzysztofCwalina: I am not sure the same applies to pools. Pools do know what's the max size buffer they can provide. The problem with IBufferWriter is that not all implementations use pooled memory, and some implementations dont know their max size.

PR #2161, opened 2018-03-15
Update the compiler version and the language version to 7.3

ahsonkhan:
cc @KrzysztofCwalina, @dotnet/corefxlab-contrib 

Do not merge yet. Waiting for VS support for language version 7.3 (the VSIX doesn't seem to work).
image


GrabYourPitchforks:
You sure this is a VSIX issue? Seems like an error in the package manifest or a mismatched version somewhere. (Though weirdly enough it does build correctly from the command line.)

ahsonkhan:
> You sure this is a VSIX issue? Seems like an error in the package manifest or a mismatched version somewhere. 

No, I am not sure if it is a VSIX issue. I will investigate tomorrow.

> Though weirdly enough it does build correctly from the command line.

Yep, which is why the CI is also passing.

ahsonkhan:
csc.exe doesn't list 7.3 yet.

```text
csc -langversion:?
Supported language versions:
default
1
2
3
4
5
6
7.0 (default)
7.1
7.2 (latest)
latest
```

ahsonkhan:
@agocke, do you have any suggestions on how to unblock us here?

jasonmalinowski:
@ahsonkhan Which csc.exe was that from? Is there still an older version of some other csc installed from a previous extension that's messing this up?

ahsonkhan:
> Which csc.exe was that from?

C:\Program Files (x86)\Microsoft Visual Studio\Preview\Enterprise\MSBuild\15.0\Bin\Roslyn\csc.exe

Where can I find the csc from the extension?

agocke:
> @agocke, do you have any suggestions on how to unblock us here?

It sounds like the compiler works. What are you blocked on?

ahsonkhan:
> It sounds like the compiler works. What are you blocked on?

Building the solution using the command line works. However, I can't get it to build in Visual Studio.
I get several build errors like:
```text
Error	MSB4062	The "Microsoft.CodeAnalysis.BuildTasks.Csc" task could not be loaded from the assembly C:\Users\ahkha\.nuget\packages\microsoft.netcore.compilers\2.8.0-beta2-62719-08\build\..\tools\Microsoft.Build.Tasks.CodeAnalysis.dll. Could not load file or assembly 'System.Runtime, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.	System.Azure.Experimental.Tests	C:\Users\ahkha\.nuget\packages\microsoft.netcore.compilers\2.8.0-beta2-62719-08\tools\Microsoft.CSharp.Core.targets	52	
```

agocke:
Right, the problem is that you can't use the Microsoft.NETCore.Compilers package in VS. The build task inside targets netstandard2.0/netcoreapp2.0, but the MSBuild used in VS only targets desktop .NET.

What you *can* do is use a combination of the Microsoft.NETCore.Compilers package and the Microsoft.Net.Compilers package. You can decide which one to use based on the `MSBuildRuntimeType` property, which will be set to Core or Full depending on which MSBuild you're running on in any given build.

ahsonkhan:
> Right, the problem is that you can't use the Microsoft.NETCore.Compilers package in VS. The build task inside targets netstandard2.0/netcoreapp2.0, but the MSBuild used in VS only targets desktop .NET.

Why would updating the compiler package version change this behavior?

> You can decide which one to use based on the `MSBuildRuntimeType` property, which will be set to Core or Full depending on which MSBuild you're running on in any given build.

How (and where) is that property set?

agocke:
>Why would updating the compiler package version change this behavior? 

Because I changed our dependency to netstandard2.0 and you updated to the package that contains the change.

> Also, what happens if I have a .NET Core application/library? Would we still end up with the MSBuild that targets desktop?

In VS? Yes.

> How (and where) is that property set?

It is an MSBuild built-in property that is always set.

GrabYourPitchforks:
Does this work now with the latest VS previews?

ahsonkhan:
> Does this work now with the latest VS previews?

I haven't tried it in a while. I will re-visit this after 2.1 RC and update this thread then.

ahsonkhan:
Superseded by https://github.com/dotnet/corefxlab/pull/2298. Closing.

PR #2176, opened 2018-03-22
BufferWriter.WriteLine String/Utf8String

KrzysztofCwalina:
Added WriteLine to BufferWriter
Refactored and renamed some files.


src/System.Buffers.ReaderWriter/System/Buffers/Writer/BufferWriterT.cs
```diff
@@ -19,6 +19,8 @@ public BufferWriter(T output)
             _span = output.GetSpan();
         }
 
+        public ReadOnlySpan<byte> NewLine;
```
YohDeadfall: Public field?

src/System.Buffers.ReaderWriter/System/Buffers/Writer/BufferWriterT_ints.cs
```diff
@@ -0,0 +1,34 @@
...
+
+namespace System.Buffers.Writer
+{
+    public ref partial struct BufferWriter<T> where T : IBufferWriter<byte>
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Write(int value, StandardFormat format = default)
+        {
+            int written;
+            while (!Utf8Formatter.TryFormat(value, Buffer, out written, format))
```
YohDeadfall: Just for myself. Why don't you use `out` variables? Is it because of the coding style?

src/System.Buffers.ReaderWriter/System/Buffers/Writer/BufferWriterT.cs
KrzysztofCwalina: yeah, I copied/pasted it from the non-generic writer. It should be a property. I will change.

src/System.Buffers.ReaderWriter/System/Buffers/Writer/BufferWriterT_ints.cs
KrzysztofCwalina: You mean why don't I do "out int written"? If yes, then because of scoping. I need written outside of the while loop.

YohDeadfall: Ah, thanks. Didn't know the expression variables scoping is different for loops.

PR #2180, opened 2018-03-22
Added the ability to upload files larger than 4MB

KrzysztofCwalina:
cc: @ahsonkhan


samples/AzCopyCore/AzCopyCore/Program.cs
```diff
@@ -197,14 +197,27 @@ static void TransferStorageFileToDirectory(ReadOnlySpan<char> storageFile, ReadO
...
+                    var putRequest = new PutRangeRequest(storagePath, bytes, index, length);
+                    response = await client.SendRequest(putRequest).ConfigureAwait(false);
+                    if (response.StatusCode != 201)
+                    {
+                        Log.TraceEvent(TraceEventType.Error, 0, "Response Status Code {0}", response.StatusCode);
+                        return false;
+                    }
+                    index += length;
+                    bytesLeft -= length;
+                    if (bytesLeft == 0) break;
```
ahsonkhan: nit: why not use this as the while condition, especially since we do this check at the end of the loop anyway?

samples/AzCopyCore/AzCopyCore/StorageRequests.cs
```diff
@@ -102,23 +102,30 @@ public static ReadOnlyMemory<byte> AsString(Http.Method verb)
...
 
         // TODO (pri 3): I dont like how the client property is a public API
         public StorageClient Client { get; set; }
 
-        public PutRangeRequest(string filePath, Stream fileContent)
+        public PutRangeRequest(string filePath, Stream fileContent, long offset, int length)
         {
-            FilePath = filePath;
-            FileContent = fileContent;
+            if (offset < 0) throw new ArgumentOutOfRangeException(nameof(offset));
```
ahsonkhan: Do we need to add checks for offset greater than length?

samples/AzCopyCore/AzCopyCore/StorageRequests.cs
```diff
@@ -131,14 +138,21 @@ class Writer : StorageRequestWriter<PutRangeRequest>
...
 
             protected override void WriteXmsHeaders(ref BufferWriter writer, ref PutRangeRequest arguments)
             {
-                long size = arguments.FileContent.Length;
+                long size = arguments._fileContent.Length;
                 writer.WriteHeader("x-ms-date", Time, 'R');
                 // TODO (pri 3): this allocation should be eliminated
-                writer.WriteHeader("x-ms-range", $"bytes=0-{size - 1}");
+
+                var start = arguments._offset;
```
ahsonkhan: nit: use of var

samples/AzCopyCore/AzCopyCore/Helpers/PipelinesExtensions.cs
```diff
@@ -44,16 +44,21 @@ static class PipelinesExtensions
...
         /// </summary>
-        public static async Task WriteAsync(this PipeWriter writer, Stream stream)
+        public static async Task WriteAsync(this PipeWriter writer, Stream stream, long bytesToWrite)
         {
             if (!stream.CanRead) throw new ArgumentException("Stream.CanRead returned false", nameof(stream));
-            while (true)
+            while (bytesToWrite > 0)
             {
                 Memory<byte> buffer = writer.GetMemory();
+                if(buffer.Length > bytesToWrite)
```
ahsonkhan: nit: formatting, add space

samples/AzCopyCore/AzCopyCore/StorageRequests.cs
```diff
@@ -131,14 +138,21 @@ class Writer : StorageRequestWriter<PutRangeRequest>
             public override Http.Method Verb => Http.Method.Put;
 
             protected override async Task WriteBody(PipeWriter writer, PutRangeRequest arguments)
-                => await writer.WriteAsync(arguments.FileContent);
+            {
+                var stream = arguments._fileContent;
```
ahsonkhan: nit: use of var

samples/AzCopyCore/AzCopyCore/Program.cs
KrzysztofCwalina: will do

samples/AzCopyCore/AzCopyCore/StorageRequests.cs
KrzysztofCwalina: offset can be greater, equal, or less than length.

samples/AzCopyCore/AzCopyCore/StorageRequests.cs
KrzysztofCwalina: will fix

samples/AzCopyCore/AzCopyCore/Program.cs
```diff
@@ -27,8 +27,23 @@ static void PrintUsage()
 
     static void Main(string[] args)
     {
+        var u = new string[]
```
ahsonkhan: Please revert this change and roll your keys.

samples/AzCopyCore/AzCopyCore/StorageRequests.cs
```diff
@@ -102,23 +102,30 @@ public static ReadOnlyMemory<byte> AsString(Http.Method verb)
...
 
         // TODO (pri 3): I dont like how the client property is a public API
         public StorageClient Client { get; set; }
 
-        public PutRangeRequest(string filePath, Stream fileContent)
+        public PutRangeRequest(string filePath, Stream fileContent, long offset, int length)
         {
-            FilePath = filePath;
-            FileContent = fileContent;
+            if (offset < 0 || offset > fileContent.Length - length) throw new ArgumentOutOfRangeException(nameof(offset));
```
ahsonkhan: Hmm. What about the case when `length > fileContent.Length`?

samples/AzCopyCore/AzCopyCore/Program.cs
```diff
@@ -212,7 +212,7 @@ static void TransferStorageFileToDirectory(ReadOnlySpan<char> storageFile, ReadO
                     }
                     index += length;
                     bytesLeft -= length;
-                    if (bytesLeft == 0) break;
+                    Debug.Assert(bytesLeft >= 0);
```
ahsonkhan: I don't think this Debug.Assert is necessary, given the loop condition is already stronger. Maybe a better option would be to add a Debug.Assert after the loop exits, if we know that bytesLeft should never be negative:
`Debug.Assert(bytesLeft == 0);`

KrzysztofCwalina: The assert ensures bytesLeft is no negative

samples/AzCopyCore/AzCopyCore/StorageRequests.cs
KrzysztofCwalina: AORE will be thrown. fileContent.Length - length will be negative and will be < offset (which has to be positive at the time of this check)

Issue #1912, opened 2017-11-07
Need unit tests for parsing methods on Date and Time types


mattjohnsonpint:
Need to add unit tests for all `Parse`, `ParseExact`, `TryParse`, and `TryParseExact` public methods on the `Time` and `Date` types.

gmilani:
Hello, I want to start contributing. Can I start with this task?

mj1856:
Yes, please do!

Note that ultimately these types will move forward to corefx via the API proposal process, which won't necessarily include the implementation or tests - so your code may possibly not make it through to the shipping product.  However, the final implementer may indeed borrow or copy from the implementation here as a starting point, so it is still useful to be as complete as possible.

AndreKraak:
Hi,

I am looking into contributing to open source projects, it will be my first time doing so.
I found this project/issue and would like to try and help. Is this issue still available to work on?

I have started with a test [DateParseTests.zip](https://github.com/dotnet/corefxlab/files/1821716/DateParseTests.zip) and before continuing I would like to ask some questions.
If there is a more appropriate forum to asks these questions please let know and I will post them there.

1. Are there any coding guidelines I should be aware off and following?
For example I see that the code explicitly uses data types in the variable declaration instead of 'var'.
2. Based on [DateFormattingTests](https://github.com/dotnet/corefxlab/blob/master/tests/System.Time.Tests/DateFormattingTests.cs) it seems that testing is only done for InvariantCulture, is this correct?
3. When testing should I also test with the possible values off [TwoDigitYearMax](https://docs.microsoft.com/en-us/dotnet/api/system.globalization.calendar.twodigityearmax?view=netframework-4.7.1)?
Depending on the value "29 December" is parsed differently.

Cheers,
André

AndreKraak:
> A string with a time but no date component.The method assumes the current date unless you call the Parse(String, IFormatProvider, DateTimeStyles) overload and include DateTimeStyles.NoCurrentDateDefaultin the styles parameter, in which case the method assumes January 1, 0001.

AndreKraak:
What is the expected result when parsing a string without a date?

Right now parsing "10:49:12" with Date.Parse result in a Date for today. This is as documented in [DateTime.Parse](https://msdn.microsoft.com/en-us/library/system.datetime.parse(v=vs.110).aspx).
[...]

PR #2182, opened 2018-03-24
[WIP] Added tests for Date.Parse(string s) (#1912)

AndreKraak:
* CannotParseData fails for 23 theories.
  The string being parsed also contains time information or only time information.
  As per https://github.com/dotnet/corefxlab/issues/1912#issuecomment-374647391 they are seen as invalid string at the moment.
> [mj1856](https://github.com/mj1856): If parsing a Date without enough information to construct a date, then yes I'd want a FormatException to be thrown. Same for TimeOfDay without a time data. I think also that we'd want to throw if the string contained extraneous data, such as if given 3/20/2018 8:43 AM, that should be parsable only as a DateTime. If parsed as a Date or a TimeOfDay, I'd expect a FormatException as well.

### DIFF
--- a/tests/System.Time.Tests/DateParseTests.cs
+++ b/tests/System.Time.Tests/DateParseTests.cs
@@ -1,0 +1,140 @@
+﻿using System.Collections.Generic;
+using System.Globalization;
+using Xunit;
+
+namespace System._Time.Tests
+{
+    public class DateParseTests
+    {
+        public DateParseTests()
+        {
+            CultureInfo.CurrentCulture = CultureInfo.InvariantCulture;
+            CultureInfo.CurrentUICulture = CultureInfo.InvariantCulture;
+        }
+
+        [Fact]
+        public static void CannotParseInvalidArguments()
+        {
+            Assert.Throws<ArgumentNullException>(() => Date.Parse(null));
+
+            Assert.Throws<FormatException>(() => Date.Parse(""));
+        }
+
+        public static IEnumerable<object[]> Parse_CurrentYearWithoutTime_TestData()
+        {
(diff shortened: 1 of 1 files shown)


AndreKraak:
Pesonally I think I would like the `Date.Parse(string s)` method to return a date when the string contains time information. The method intention is to retreive a date and the fact that time information is present is not important for me.

The way these string would need to be parsed would be by using `DateTime.Parse(string s)` and then use the `GetDate()` extension to get the date I need.

The only exception would be the string that only contains time information, parsing this should not return a date.

AndreKraak:
I have been thinking on it and I would like the `Date.Parse(string s)` method to return a date when the string also contains time information. By using the method I am indicating my interest in the date information and that I am not concerned about the time information.

Currently the way to get the date would be to first parse the string using `DateTime.Parse(string s)` and then use the `GetDate()` extension.
If I have a list off string I intend to parse for data, I would either need to determine which parse method to use for each item or parse them all with `DateTime.Parse(string s)` and get the date from its result.
Needing to use the `DateTime.Parse(string s)` to make sure that all strings can be parsed would make the having a `Date.Parse(string s)` not as usefull in my opinion.

The exception is when the string only contains time information, in which case I would like to get a `FormatException`.

ahsonkhan:
Any updates to this PR?

danmosemsft:
I'm going to close this PR. Feel free to reopen if you plan to continue work on it.

YohDeadfall:
@mj1856 I would like to finish `Date` and `Time` structs, but to implement it correctly I need to do some tweaks in `DateTimeParse`. What is the best approach to do that? Should I copy `DateTimeParse` here and add changes to it?

Issue #1534, opened 2017-05-09
System.IO.FileSystem.Watcher.Polling uses Windows specific API FindFirstFileExW


ahsonkhan:
https://github.com/dotnet/corefxlab/blob/master/src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs#L95

Hence, the tests only run for Windows for now and disabled for the rest, here:
https://github.com/dotnet/corefxlab/pull/1532

cc @KrzysztofCwalina

ahsonkhan:
@MisinformedDNA, can this issue be closed now?

MisinformedDNA:
Yes please.

PR #2208, opened 2018-04-13
Remove Windows specific code

MisinformedDNA:
#1534


src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
```diff
@@ -84,43 +88,20 @@ private FileChangeList ComputeChangesAndUpdateState()
...
-                            if (!IsWatched(fileData.cFileName)) continue;
-                            UpdateState(directory, ref changes, ref fileData, fileData.cFileName);
-                        }
-                        while (DllImports.FindNextFileW(handle, pFileData));
-                    }
-                    finally
-                    {
-                        DllImports.FindClose(handle);
-                    }
+                var files = Directory.EnumerateFiles(_directory, "*", _includeSubdirectories ? SearchOption.AllDirectories : SearchOption.TopDirectoryOnly);
```
JeremyKuhne: Making this change would significantly impact the performance of the watcher as it will now allocate full path strings for every file when there were no allocations before.

MisinformedDNA: Dumb question: where do I find `FileSystemEnumerable<T>`?

YohDeadfall: In the `System.IO.FileSystem` package.

JeremyKuhne: It is in the System.IO.Enumeration namespace: https://apisof.net/catalog/System.IO.Enumeration.FileSystemEnumerable%3CTResult%3E

MisinformedDNA: I had assumed that was the package it would be in, but the [last NuGet package](https://www.nuget.org/packages/System.IO.FileSystem/4.3.0) was in 2016 and the [last MyGet package](https://dotnet.myget.org/feed/dotnet-core/package/nuget/System.IO.FileSystem/4.4.0-beta-24913-02) was in the beginning of 2017. So where do I find this updated package?

JeremyKuhne: @KrzysztofCwalina, how do we consume the latest bits in a project in CoreFXLab?

MisinformedDNA: If I create a brand new Core console app (and change the target to 2.1), I still can't find the updated code. I would expect it to be in NuGet marked "Preview", but like I said that code hasn't been updated in a while. So while my issue is certainly in CoreFXLab, the issue extends to "normal" development.

ahsonkhan: > how do we consume the latest bits in a project in CoreFXLab?

These are [netstandard libraries](https://github.com/dotnet/corefxlab/blob/master/src/System.IO.FileSystem.Watcher.Polling/System.IO.FileSystem.Watcher.Polling.csproj#L5), so what exactly do you want to consume in corefxlab?

> If I create a brand new Core console app (and change the target to 2.1), I still can't find the updated code.

Aren't System.IO.FileSystem/Enumeration part of [Microsoft.NetCoreApp](https://www.nuget.org/packages/Microsoft.NETCore.App/2.1.0-preview2-26406-04) package that comes as part of 2.1? Why do you need an explicit package reference? I think if your app targets netcoreapp2.1, you should get the API.

If you want to consume packages/libraries built from corefxlab, include the following MyGet feed: https://dotnet.myget.org/feed/dotnet-corefxlab/package/nuget/System.IO.FileSystem.Watcher.Polling

Let me know if you are still blocked. If you could share the csproj of your console application, I could take a look and help further.

MisinformedDNA: This is my CSPROJ:

```
<Project Sdk="Microsoft.NET.Sdk">

  <PropertyGroup>
    <OutputType>Exe</OutputType>
    <TargetFramework>netcoreapp2.1</TargetFramework>
  </PropertyGroup>

</Project>
```

I've also attached a ZIP of the (very tiny) project. [ConsoleApp1.zip](https://github.com/dotnet/corefxlab/files/1921680/ConsoleApp1.zip)

And notice that the Enumeration namespace is missing in my Object Explorer:
image

ahsonkhan:
```diff
- using System.IO.Enumerable; // namespace not found
+ using System.IO.Enumeration; // this is the right namespace
```

```diff
- FileSystemEnumerable test;  // type not found
+ FileSystemEnumerable<int> test;  // type is generic
```

MisinformedDNA: Thanks for the corrections, but the type/namespace still doesn't load.

MisinformedDNA: Next question: The `PollingWatcher` is currently targetting .NET Standard 1.3. Even if I target .NET Standard, it still won't include your new APIs. Can I make the PollingWatcher project target .NET Core? I don't know how concerned we are with compatibility with the .NET Framework.

ahsonkhan: > Sigh. I accidentally installed Preview 1. Is there a way we can mark old releases [here](https://www.microsoft.com/net/download/dotnet-core/sdk-2.1.300-preview1)?

cc @livarcocc 

> Can I make the PollingWatcher project target .NET Core? I don't know how concerned we are with compatibility with the .NET Framework.

I don't own this project so I can't make the call. If we need to take a dependency on APIs only available on a higher version of net standard, or .net core 2.1 only (and we are ok with making this library less portable), then updating the target version would be fine. I don't think any other projects in corefxlab depend on this one. However, it is usually recommended for portable libraries to target the lowest possible version (and maybe have if/defs, if absolutely necessary).

cc @JeremyKuhne

JeremyKuhne: Yes, go ahead and target 2.1. We don't want to rewrite the interop logic and until we find that we really have to have this down level we can address that later. It should be designed around the high-performance APIs I've mentioned. :)

@KrzysztofCwalina, if you have any concerns we can discuss, but I believe this is in line with what you were expecting.

KrzysztofCwalina:
> @KrzysztofCwalina, if you have any concerns we can discuss, but I believe this is in line with what you were expecting.

@MisinformedDNA, @JeremyKuhne, the main objective for this component was to have zero allocations when nothing changes in the watched directories. Is this still true after the change? If yes, the change is great. Otherwise, I would like to see how much is allocated when the pooling timer runs.

MisinformedDNA:
I updated the code to use the new APIs. I don't see any allocations in the Diagnostics windows.

The one thing of note in this commit is that the extension comparison, is now case-sensitive (which can be wrong for Windows), but it was previously case-insensitive (which is wrong for Linux).

I will file an API Proposal, that resolves this while making it more flexible and more inline with the new .NET APIs.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
```diff
@@ -129,93 +116,43 @@ private FileChangeList ComputeChangesAndUpdateState()
...
         {
+            if (entry.IsDirectory) return false;
             if (_extensionsToWatch == null) return true;
-            var length = GetLength(filename);
-            foreach(var extension in _extensionsToWatch)
+            foreach (var extension in _extensionsToWatch)
             {
-                if(EndsWith(filename, length, extension))
-                {
+                if (Path.GetExtension(entry.FileName).SequenceEqual(extension))
```
JeremyKuhne: There is an `EndsWith()` extension for `ReadOnlySpan<char>` that also takes a `StringComparison`. It is in the `System` namespace and comes from the `System.Memory` library (which is part of 2.1).

MisinformedDNA: As I mentioned above, it will either be wrong for Windows or wrong for Linux (or we do if defines). My [proposal](https://github.com/dotnet/corefxlab/issues/2222) is that we remove this code and let the user define this part through `ShouldRecursePredicate`.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
```diff
@@ -84,43 +89,25 @@ private FileChangeList ComputeChangesAndUpdateState()
...
-                        }
-                        while (DllImports.FindNextFileW(handle, pFileData));
-                    }
-                    finally
-                    {
-                        DllImports.FindClose(handle);
-                    }
+                var files = new FileSystemEnumerable<FileInfo>(
+                    _directory,
+                    (ref FileSystemEntry entry) => (FileInfo)entry.ToFileSystemInfo(),
```
JeremyKuhne: This will allocate more than you need to. In the delegate you can call `UpdateState` with the `FileSystemEntry`. As that will capture `changes` the delegate won't get cached, however (resulting in an extra allocation). You can eliminate that by deriving directly from `FileSystemEnumerator<TResult>`. `FileSystemEnumerable` is the 80% case helper (it wraps `FileSystemEnumerator` and we use it internally for existing IO APIs). `FileSystemEnumerator` is the advanced, full control option. :)

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
JeremyKuhne: Specifying StringComparison (or `bool ignoreCase`) as an option is the best place to start on this change if you want to expose right away. You can, by default, pick the one you think is applicable based on the platform. MacOS & Windows should be insensitive by default.

I'll review your proposal as soon as I get a chance.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
MisinformedDNA: Thanks for helping me make my code awesome! Is this what you had in mind:

```csharp
var files = new FileSystemEnumerable<FileInfo>(
    _directory,
    (ref FileSystemEntry entry) => (FileInfo)entry.ToFileSystemInfo(),
    new EnumerationOptions { IgnoreInaccessible = true, RecurseSubdirectories = _includeSubdirectories })
{
    ShouldIncludePredicate = (ref FileSystemEntry entry) => IsWatched(ref entry)
};
var fileEnumerator = files.GetEnumerator();
while( fileEnumerator.MoveNext())
{
    UpdateState(_directory, ref changes, fileEnumerator.Current);
}
```

I don't feel like I gained anything. But you also said I should derive from `FileSystemEnumerator`, but I'm not really sure where to go with that. `FileSystemEnumaerable` seems to be tied to `FileSystemEnumerator`, so did you want me to derive from `FileSystemEnumerable`

I'm also not seeing the extra allocations that I need to get rid of, so all pointers are welcome. Thanks for your help!

src/System.IO.FileSystem.Watcher.Polling/System/IO/PathToFileStateHashtable.cs
```diff
@@ -121,41 +101,21 @@ private int NextCandidateBucket(int bucket)
...
-        [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private unsafe bool Equal(FullPath fullPath, string directory, string file)
-        {
-            if (!String.Equals(fullPath.Directory, directory, StringComparison.Ordinal))
-            {
-                return false;
-            }
-
-            if (!String.Equals(fullPath.File, file, StringComparison.Ordinal))
+            if (!MemoryExtensions.Equals(fullPath.File, file, StringComparison.Ordinal))
```
JeremyKuhne: Nit- use the extension syntax. (`fullPath.File.Equals(...`)

JeremyKuhne: Other way around, actually. :) `file.Equals(fullPath.File...`  (so you don't have to explicitly cast)

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
```diff
@@ -33,43 +35,47 @@ public class PollingWatcher : IDisposable
...
 
         /// <summary>
         /// Extensions to watch
         /// </summary>
-        /// <param name="extension">for examople "txt", "doc", etc., i.e don't use wildcards</param>
+        /// <param name="extension">for example "txt", "doc", etc., i.e don't use wildcards</param>
         /// <remarks>
         /// By default all extensions are watched. Once this method is called, only extensions in added are watched.
         /// </remarks>
         public void AddExtension(string extension)
```
JeremyKuhne: We should eventually remove this and make it `Filters`. We didn't have a non-allocating matcher when this was first written.

MisinformedDNA: This is coming in my next PR which aims to bring the `PollingFileSystemWatcher` API in closer alignment with `FileSystemWatcher`.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
```diff
@@ -129,93 +108,44 @@ private FileChangeList ComputeChangesAndUpdateState()
...
         {
+            if (entry.IsDirectory) return false;
             if (_extensionsToWatch == null) return true;
-            var length = GetLength(filename);
-            foreach(var extension in _extensionsToWatch)
+            foreach (var extension in _extensionsToWatch)
             {
-                if(EndsWith(filename, length, extension))
-                {
+                var ignoreCase = !RuntimeInformation.IsOSPlatform(OSPlatform.Linux);
```
JeremyKuhne: It might be better to ignore case for Mac & Windows rather than not Linux. Other non-Linux based Unix variants should also be case sensitive. Also a nit- use `bool` and `string`, not var.  The existing code wasn't following our standards regarding var.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingWatcher.cs
```diff
@@ -129,93 +108,44 @@ private FileChangeList ComputeChangesAndUpdateState()
...
-            }
-
             var previousState = _state.Values[index];
-            if (file.LastWrite != previousState.LastWrite || file.FileSize != previousState.FileSize) {
+            if (file.LastWriteTimeUtc != previousState.LastWrite || file.Length != previousState.FileSize)
+            {
                 changes.AddChanged(directory, previousState.Path);
-                _state.Values[index].LastWrite = file.LastWrite;
-                _state.Values[index].FileSize = file.FileSize;
+                _state.Values[index].LastWrite = file.LastWriteTimeUtc;
```
JeremyKuhne: We should eventually change these to match the names that `FileSystemEntry` is using. (Which, in turn, aligns with prior System.IO terminology)

MisinformedDNA:
Fixed the extension and var issues. Will fix the others in subsequent PR (as mentioned earlier).

JeremyKuhne:
@MisinformedDNA can you click on the Details link for the Pre-Merge signing tool and sign off?

ahsonkhan:
> can you click on the Details link for the Pre-Merge signing tool and sign off?

I don't think external contributors can do it. It is also not necessary.

JeremyKuhne:
> I don't think external contributors can do it. It is also not necessary.

Hmm. I'm clearly misinformed about what this is. :)

Issue #2224, opened 2018-04-19
Fix ability to run tests


MisinformedDNA:
When I try to run tests in VS2017P3, I get this error

> Microsoft.VisualStudio.TestPlatform.ObjectModel.TestPlatformException: Testhost process exited with error: It was not possible to find any compatible framework version
> The specified framework 'Microsoft.NETCore.App', version '2.1.0-preview3-26410-06' was not found.
>   - Check application dependencies and target a framework version installed at:
>       C:\Program Files\dotnet\
>   - Installing .NET Core prerequisites might help resolve this problem:
>       http://go.microsoft.com/fwlink/?LinkID=798306&clcid=0x409
>   - The .NET Core framework and SDK can be installed from:
>       https://aka.ms/dotnet-download
>   - The following versions are installed:
>       1.0.1 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       1.0.4 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       1.0.5 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       1.1.1 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       1.1.2 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       2.0.0 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       2.0.3 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       2.0.5 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       2.0.6 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       2.1.0-preview1-26216-03 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
>       2.1.0-preview2-26406-04 at [C:\Program Files\dotnet\shared\Microsoft.NETCore.App]
> 
>    at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyOperationManager.SetupChannel(IEnumerable`1 sources, CancellationToken cancellationToken)
>    at Microsoft.VisualStudio.TestPlatform.CrossPlatEngine.Client.ProxyExecutionManager.StartTestRun(TestRunCriteria testRunCriteria, ITestRunEventsHandler eventHandler)
 
[...]

PR #2226, opened 2018-04-20
Always install latest dotnet version

MisinformedDNA:
#2224

### DIFF
--- a/scripts/build.ps1
+++ b/scripts/build.ps1
@@ -14,23 +14,6 @@ Write-Host "SkipTests=$SkipTests."
 
 if (!(Test-Path "dotnetcli\dotnet.exe")) {
     Write-Host "dotnet.exe not installed, downloading and installing."
-    if ($Version -eq "<default>") {
-        $Version = (Get-Content "$PSScriptRoot\..\DotnetCLIVersion.txt" -Raw).Trim()
-    }
-    Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Channel master -Version $Version -InstallDir $PSScriptRoot\..\dotnetcli"
-    if ($lastexitcode -ne $null -and $lastexitcode -ne 0) {
-        Write-Error "Failed to install latest dotnet.exe, exit code [$lastexitcode], aborting build."
-        exit -1
-    }
-
-    # Temporary workaround until CLI, Core-Setup, CoreFx are all in sync with the shared runtime.
-    $SharedVersion = (Get-Content "$PSScriptRoot\..\SharedRuntimeVersion.txt" -Raw).Trim()
-    Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Channel master -Version $SharedVersion -InstallDir $PSScriptRoot\..\dotnetcli -SharedRuntime"
-    if ($lastexitcode -ne $null -and $lastexitcode -ne 0) {
-        Write-Error "Failed to install latest 2.1.0 shared runtime (version $SharedVersion), exit code [$lastexitcode], aborting build."
-        exit -1
-    }
-
     Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Version 1.0.0 -InstallDir $PSScriptRoot\..\dotnetcli"
     if ($lastexitcode -ne $null -and $lastexitcode -ne 0) {
         Write-Error "Failed to install framework version 1.0.0, exit code [$lastexitcode], aborting build."
@@ -44,10 +27,36 @@ if (!(Test-Path "dotnetcli\dotnet.exe")) {
     }
 }
 
+$dotnetExePath=".\dotnetcli\dotnet.exe"
+
+$cliVersion = Invoke-Expression "$dotnetExePath --version"
+if ($Version -eq "<default>") {
+    $Version = (Get-Content ".\DotnetCLIVersion.txt" -Raw).Trim()
+}
+if ($cliVersion -ne $Version) {
+    Write-Host "Newest version of dotnet cli not installed, downloading and installing."
+    Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Channel master -Version $cliVersion -InstallDir $PSScriptRoot\..\dotnetcli"
+    if ($lastexitcode -ne $null -and $lastexitcode -ne 0) {
+        Write-Error "Failed to install latest dotnet.exe, exit code [$lastexitcode], aborting build."
+        exit -1
+    }
+}
+
+# Temporary workaround until CLI, Core-Setup, CoreFx are all in sync with the shared runtime.
+$installedRuntimeVersions = Invoke-Expression "$dotnetExePath --list-runtimes" | %{ $_.split(" ")[1] } | Select-Object
+$SharedVersion = (Get-Content ".\SharedRuntimeVersion.txt" -Raw).Trim()
+if (!($installedRuntimeVersions.Contains($SharedVersion))) {
+    Write-Host "Newest version of dotnet runtime not installed, downloading and installing."
+    Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Channel master -Version $SharedVersion -InstallDir $PSScriptRoot\..\dotnetcli -SharedRuntime"
+    if ($lastexitcode -ne $null -and $lastexitcode -ne 0) {
+        Write-Error "Failed to install latest 2.1.0 shared runtime (version $SharedVersion), exit code [$lastexitcode], aborting build."
+        exit -1
+    }
+}
+
 $env:DOTNET_SKIP_FIRST_TIME_EXPERIENCE = 1
 $env:DOTNET_MULTILEVEL_LOOKUP = 0
 
-$dotnetExePath="$PSScriptRoot\..\dotnetcli\dotnet.exe"
 
 $file = "corefxlab.sln"
 


MisinformedDNA:
@ahsonkhan

scripts/build.ps1
```diff
@@ -14,40 +14,49 @@ Write-Host "SkipTests=$SkipTests."
...
     if ($lastexitcode -ne $null -and $lastexitcode -ne 0) {
-        Write-Error "Failed to install framework version 1.0.0, exit code [$lastexitcode], aborting build."
+        Write-Error "Failed to install latest dotnet.exe, exit code [$lastexitcode], aborting build."
         exit -1
     }
+}
 
-    Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Version 2.0.0 -InstallDir $PSScriptRoot\..\dotnetcli"
+# Temporary workaround until CLI, Core-Setup, CoreFx are all in sync with the shared runtime.
+$installedRuntimeVersions = Invoke-Expression "$dotnetExePath --list-runtimes" | %{ $_.split(" ")[1] } | Select-Object
```
ahsonkhan: What happens if $_.split(" ") results in an empty array? Should we guard against that?

scripts/build.ps1
```diff
@@ -14,40 +14,49 @@ Write-Host "SkipTests=$SkipTests."
...
+    Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Version 2.0.0 -InstallDir $PSScriptRoot\..\dotnetcli"
     if ($lastexitcode -ne $null -and $lastexitcode -ne 0) {
-        Write-Error "Failed to install latest 2.1.0 shared runtime (version $SharedVersion), exit code [$lastexitcode], aborting build."
+        Write-Error "Failed to install framework version 2.0.0, exit code [$lastexitcode], aborting build."
         exit -1
     }
+}
 
-    Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Version 1.0.0 -InstallDir $PSScriptRoot\..\dotnetcli"
+$dotnetExePath=".\dotnetcli\dotnet.exe"
```
ahsonkhan: Why was it necessary to change this from `$dotnetExePath="$PSScriptRoot\..\dotnetcli\dotnet.exe"`?

scripts/build.ps1
```diff
@@ -14,40 +14,49 @@ Write-Host "SkipTests=$SkipTests."
 
 if (!(Test-Path "dotnetcli\dotnet.exe")) {
     Write-Host "dotnet.exe not installed, downloading and installing."
-    if ($Version -eq "<default>") {
```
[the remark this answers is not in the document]
MisinformedDNA: `dotnet --list-runtimes` does not exist for 2.0. Do we really need 1.0 and 2.0 installed at all?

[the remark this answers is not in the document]
ahsonkhan: We probably don't need 1.0 but I am not 100% certain we can remove 2.0. Try to remove them and see if the build fails or if there is any issues with the tests. As far as I can tell, it should be fine to remove them both.

scripts/build.ps1
MisinformedDNA: Powershell won't throw an error. We should be fine. Try `"".split(" ")[1]` in PowerShell.

scripts/build.ps1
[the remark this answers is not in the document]
ahsonkhan: We largely keep them for testing purposes, so I would prefer to keep them for now. It would still be good to know that we **can** remove them if we wanted to without unexpected failures.

scripts/build.ps1
ahsonkhan: Yes, but `$installedRuntimeVersions.Contains($SharedVersion)` would throw if it was set to null.

MisinformedDNA: `$installedRuntimeVersions` can only be null if `Invoke-Expression "$dotnetExePath --list-runtimes"` didn't return any runtimes. Since we are installing the latest first, I wouldn't think that would be possible.

ahsonkhan: You are right that in normal circumstances, we will have at least one runtime installed. But if we have only one runtime installed, wouldn't splitting on space and indexing at [1] still return null since we only have one element? Are there always at least 2 elements (some header text maybe)?

**As an aside:** What happens if `--list-runtimes` doesn't exist (for instance if 2.0 cli was installed)? What type of error will the user get?

MisinformedDNA: As long as there is one line with one space (even if there are no other characters surrounding it), there is no issue.

`--list-runtimes` returns an error message in 2.0 with a space, so there's no issue there either.

The only way this fails is if `--list-runtimes` returns absolutely nothing.

MisinformedDNA: If you want the check, I'll add the check. I just disagree that it's needed. :)

MisinformedDNA:
My first time bash scripting!

scripts/build.ps1
ahsonkhan: I think we need to look through the entire list and find all the versions of Microsoft.NETCore.App.

Just looking at the the first one listed could get the wrong result.

`dotnet --list-runtimes`
```text
Microsoft.AspNetCore.All 2.1.0-preview2-final [D:\GitHub\Fork\corefxlab\dotnetcli\shared\Microsoft.AspNetCore.All]
Microsoft.AspNetCore.App 2.1.0-preview2-final [D:\GitHub\Fork\corefxlab\dotnetcli\shared\Microsoft.AspNetCore.App]
Microsoft.NETCore.App 1.0.4 [D:\GitHub\Fork\corefxlab\dotnetcli\shared\Microsoft.NETCore.App]
Microsoft.NETCore.App 1.1.1 [D:\GitHub\Fork\corefxlab\dotnetcli\shared\Microsoft.NETCore.App]
Microsoft.NETCore.App 2.0.0 [D:\GitHub\Fork\corefxlab\dotnetcli\shared\Microsoft.NETCore.App]
Microsoft.NETCore.App 2.1.0-preview2-26406-04 [D:\GitHub\Fork\corefxlab\dotnetcli\shared\Microsoft.NETCore.App]
Microsoft.NETCore.App 2.1.0-preview3-26416-01 [D:\GitHub\Fork\corefxlab\dotnetcli\shared\Microsoft.NETCore.App
```

Right now it is checking `2.1.0-preview2-final == 2.1.0-preview3-26416-01` and returning false, causing us to re-install the existing version.

ahsonkhan: > If you want the check, I'll add the check. I just disagree that it's needed. :)

You are right, it is not necessary, as long as you have at least one version installed.

build.sh
```diff
@@ -64,13 +67,42 @@ if [ ! -d "dotnetcli" ]; then
...
   fi
+else
+  echo "dotnet.exe is installed, checking for latest."
+  cliVersion=$(./$dotnetExePath --version)
+  if [ $cliVersion != $Version ]; then
+    echo "Newest version of dotnet cli not installed, downloading and installing."
+    ./scripts/install-dotnet.sh -Channel master -Version "$Version" -InstallDir "dotnetcli"
+    ret=$?
+	if [ $ret -ne 0 ]; then
+      echo "Failed to install latest dotnet.exe, exit code $ret, aborting build.."
```
ahsonkhan: nit: extra period at the end of the sentence

build.sh
```diff
@@ -64,13 +67,42 @@ if [ ! -d "dotnetcli" ]; then
...
+      exit -1
+    fi
+  fi
+
+  installedRuntimeVersions=($(./$dotnetExePath --list-runtimes | cut -d ' ' -f2))
+  found=0
+  for i in "${installedRuntimeVersions[@]}"
+  do
+      if [ "$i" == "$yourValue" ] ; then
```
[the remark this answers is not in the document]
MisinformedDNA: (headbang)

ahsonkhan:
cc @dotnet/corefxlab-contrib FYI

ahsonkhan:
@MisinformedDNA, there seems to be a bug here.

> dotnet --version

2.1.300-preview2-008533

DotnetCLIVersion.txt has 2.2.100-preview1-008633

There is a mismatch, so I expected a new version to be installed (and it started to). But then it failed with, version 8533 already installed.

> D:\GitHub\Fork\corefxlab>build.cmd
Configuration=Debug.
Restore=true.
Version=<default>.
BuildVersion=preview2-180420-1.
SkipTests=false.
dotnet.exe is installed, checking for latest.
Newest version of dotnet cli not installed, downloading and installing.
**dotnet-install: .NET SDK version 2.1.300-preview2-008533 is already installed.**
dotnet-install: Adding to current process PATH: "D:\GitHub\Fork\corefxlab\dotnetcli\". Note: This change will not be visible if PowerShell was run as a child process.
Newest version of dotnet runtime not installed, downloading and installing.
D:
dotnet-install: Downloading https://dotnetcli.azureedge.net/dotnet/Runtime/2.2.0-preview1-26418-01/dotnet-runtime-2.2.0-preview1-26418-01-win-x64.zip
dotnet-install: Extracting zip from https://dotnetcli.azureedge.net/dotnet/Runtime/2.2.0-preview1-26418-01/dotnet-runtime-2.2.0-preview1-26418-01-win-x64.zip
dotnet-install: Adding to current process PATH: "D:\GitHub\Fork\corefxlab\dotnetcli\". Note: This change will not be visible if PowerShell was run as a child process.
dotnet-install: Installation finished

scripts/build.ps1
```diff
@@ -42,12 +44,34 @@ if (!(Test-Path "dotnetcli\dotnet.exe")) {
         Write-Error "Failed to install framework version 2.0.0, exit code [$lastexitcode], aborting build."
         exit -1
     }
+} else {
+    Write-Host "dotnet.exe is installed, checking for latest."
+
+    $cliVersion = Invoke-Expression "$dotnetExePath --version"
+    if ($cliVersion -ne $Version) {
+        Write-Host "Newest version of dotnet cli not installed, downloading and installing."
+        Invoke-Expression -Command "$PSScriptRoot\install-dotnet.ps1 -Channel master -Version $cliVersion -InstallDir $PSScriptRoot\..\dotnetcli"
```
ahsonkhan: `-Channel master -Version $cliVersion`

This needs to be `-Channel master -Version $Version`

MisinformedDNA: Doh! Thanks.

Issue #2215, opened 2018-04-16
Extend the SpanUsage code analyzer to check for Memory<T>.Slice(...).Span


ahsonkhan:
`Memory<T>.Slice(...).Span` => `Memory<T>.Span.Slice(...)`

https://github.com/dotnet/corefxlab/pull/2206

PR #2236, opened 2018-05-01
Add a code analyzer to fix certain Memory<T> usage

ahsonkhan:
Fixes https://github.com/dotnet/corefxlab/issues/2215

Swap the order of `Memory<T>.Slice(start, length).Span` to be `Memory<T>.Span.Slice(start, length)`

**Screenshots:**
1) image

2) image

3) image

**Majority of the code here is template code. Focus on the following files:**
- [samples/SpanUsage/MemoryUsage/MemoryUsage/MemoryUsageAnalyzer.cs](https://github.com/dotnet/corefxlab/pull/2236/files#diff-9996611a5850352a18ddfa18026d6a07)
- [samples/SpanUsage/MemoryUsage/MemoryUsage/MemoryUsageCodeFixProvider.cs](https://github.com/dotnet/corefxlab/pull/2236/files#diff-79add863d28a8badebf61fb73c9705ac)
- [samples/SpanUsage/MemoryUsage/MemoryUsage.Test/MemoryUsageUnitTests.cs](https://github.com/dotnet/corefxlab/pull/2236/files#diff-2d20c37fe6b4ac869e9da53e6752b0ef)

corefxlab fix - https://github.com/dotnet/corefxlab/pull/2237

cc @dotnet/corefxlab-contrib, @khellang, @VSadov, @KrzysztofCwalina, @terrajobst


KrzysztofCwalina:
Would it be possible to write a analyzer that detects when there are multiple unnecessary calls to Memory.Span in a single method body?

ahsonkhan:
> Would it be possible to write a analyzer that detects when there are multiple unnecessary calls to Memory.Span in a single method body?

I believe so, but it would be significantly more complex and require semantic tracking and analysis. Feel free to file an issue for it. I am assuming it would only highlight the issue, since offering a fix in that case seems futile.

PR #2238, opened 2018-05-01
BenchmarkDotNet init

adamsitnik:


tests/Benchmarks/Benchmarks.csproj
```diff
@@ -20,10 +21,10 @@
     <NoWarn>$(NoWarn);CS0618</NoWarn>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.10.14.525" />
+    <PackageReference Include="BenchmarkDotNet.Diagnostics.Windows" Version="0.10.14.525" />
```
KrzysztofCwalina: Can the benchmarks be run on Linux/Mac?

tests/Benchmarks/Buffers/Writer.cs
```diff
@@ -11,9 +11,8 @@
 using System.Runtime.CompilerServices;
 using System.Text;
 using System.Text.Utf8;
-using System.Threading;
 
-public class BufferWriterBench
+public class Writer
```
KrzysztofCwalina: Why this change? Isn't "Writer" a bit too generic of a name?

tests/Benchmarks/Benchmarks.csproj
```diff
@@ -20,10 +21,10 @@
     <NoWarn>$(NoWarn);CS0618</NoWarn>
   </PropertyGroup>
   <ItemGroup>
+    <PackageReference Include="BenchmarkDotNet" Version="0.10.14.525" />
```
ahsonkhan: nit: Add BenchmarkDotNetVersion property in dependencies.props and reference it here: https://github.com/dotnet/corefxlab/blob/master/tools/dependencies.props

tests/Benchmarks/Program.cs
```diff
@@ -0,0 +1,10 @@
+﻿using BenchmarkDotNet.Running;
```
ahsonkhan: Requires license header.

tests/Benchmarks/Encodings/AsciiDecoding.cs
```diff
@@ -0,0 +1,32 @@
...
+        private byte[] bytes;
+
+        [Params("/plaintext", "text/plain,text/html;q=0.9,application/xhtml+xml;q=0.9,application/xml;q=0.8,*/*;q=0.7")]
+        public string Text;
+
+        [GlobalSetup]
+        public void Setup() => bytes = Encoding.ASCII.GetBytes(Text);
+
+        [Benchmark]
+        public string AsciiToStringPrimitives() => System.Buffers.Text.Encodings.Ascii.ToUtf16String((ReadOnlySpan<byte>)bytes);
```
KrzysztofCwalina: Is the cast required?

tests/Benchmarks/Program.cs
```diff
@@ -0,0 +1,10 @@
...
+
+namespace Benchmarks
+{
+    public class Program
+    {
+        public static void Main(string[] args)
+            => BenchmarkSwitcher.FromAssembly(typeof(Program).Assembly).Run(args);
```
KrzysztofCwalina: Is there a way to run the tests on both CLR and Core without changing the sources? I think it would be interesting to be able to run some CLR-Core comparisons for APIs that use fast and slow span.

tests/Benchmarks/Encodings/AsciiDecoding.cs
```diff
@@ -0,0 +1,32 @@
...
+
+        [Benchmark]
```
KrzysztofCwalina: Should this be the baseline, i.e. existing CLR implementation being the baseline.

tests/Benchmarks/Encodings/AsciiDecoding.cs
```diff
@@ -0,0 +1,32 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Text;
+using System.Text.Utf8;
+using BenchmarkDotNet.Attributes;
+
+namespace Benchmarks.Encodings
+{
+    public class AsciiDecoding
```
KrzysztofCwalina: If I want to measure allocations, can I do it without changing the sources? i.e. through the "switcher"

tests/Benchmarks/Buffers/Writer.cs
ahsonkhan: Let's set a consistent structure for the namespaces. I like the Benchmarks.Encodings for the other tests. For this one, can we put it in the Benchmarks.Buffers namespace?

tests/Benchmarks/Program.cs
KrzysztofCwalina: And more generally, it would be useful to add some comment to this Main explaining what command line args can be passed (maybe simple inline, and then link to a BDN docs).

tests/Benchmarks/Buffers/Writer.cs
ahsonkhan: I believe the objective is to reduce duplication within the naming. The folder structure already contains "Buffers", "Benchmarks". @adamsitnik, please clarify if I am mistaken.

tests/Benchmarks/Buffers/Writer.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-using Microsoft.Xunit.Performance;
```
KrzysztofCwalina: It would be good if we have a naming convention for the sub-folders containing tests. In thins case the sub-folder is called  "Buffers", but we have two projects System.Buffers.Primitives and System.Buffers.Experimental". "Buffers" makes it not clear which API project these tests are for. Maybe the folder should be called "SystemBuffersPrimitives" or "BuffersPrimitives" or something like that.

tests/Benchmarks/Program.cs
ahsonkhan: We should also add a README.md summarizing how we can run the tests and common args that are useful here (similar to what we had before https://github.com/dotnet/corefxlab/blob/master/scripts/PerfHarness/README.md). That doc can evolve as we do the cleanup.

tests/Benchmarks/Benchmarks.csproj
adamsitnik: yes, `BenchmarkDotNet.Diagnostics.Windows` is optional and will be used only for HardwareCounters and InliningDiagnoser. It targets .NET Standard 2.0 so it will build on any OS

tests/Benchmarks/Buffers/Writer.cs
adamsitnik: @ahsonkhan you are right. The project is called `Benchmarks`, the subfolder is called `Buffers` so I don't want to duplicate the names in full name

`Benchmarks.Buffers.Writer` vs
`Benchmarks.BufferWriterBench`

I think it's cleaner solution but it's just my personal preference so I can change it if you want @KrzysztofCwalina

tests/Benchmarks/Program.cs
adamsitnik: > Is there a way to run the tests on both CLR and Core without changing the sources? 

not yet. I can add sample config to the code and add comment for how to use it

Unfortunately I don't have time to add the docs now. I will serarch for some `.md` file with docs that we could reuse.

tests/Benchmarks/Buffers/Writer.cs
KrzysztofCwalina: I am fine with the naming you have. Just wanted to make sure the change is what we as a team like.

tests/Benchmarks/Encodings/AsciiDecoding.cs
KrzysztofCwalina: And you want to avoid double cast? i.e. byte[] => Span => ReadOnlySpan?

KrzysztofCwalina:

PR #2240, opened 2018-05-02
Ported BufferReaderBench to BenchmarkDotNet

KrzysztofCwalina:
``` ini

BenchmarkDotNet=v0.10.14.525-nightly, OS=Windows 10.0.16299.371 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7-6700K CPU 4.00GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
Frequency=3914067 Hz, Resolution=255.4887 ns, Timer=TSC
.NET Core SDK=2.2.0-preview1-007460
  [Host]     : .NET Core 2.2.0-preview1-26424-02 (CoreCLR 4.6.26423.05, CoreFX 4.6.26423.04), 64bit RyuJIT
  DefaultJob : .NET Core 2.2.0-preview1-26424-02 (CoreCLR 4.6.26423.05, CoreFX 4.6.26423.04), 64bit RyuJIT


```
|                    Method |       Mean |      Error |     StdDev | Scaled | ScaledSD |
|-------------------------- |-----------:|-----------:|-----------:|-------:|---------:|
|    ParseInt32BufferReader | 2,405.1 us | 14.3076 us | 12.6833 us |   3.48 |     0.02 |
|      ParseInt32Utf8Parser |   691.1 us |  0.4016 us |  0.3353 us |   1.00 |     0.00 |
| ParseInt32BufferReaderRaw |   903.2 us |  5.2700 us |  4.9296 us |   1.31 |     0.01 |

It's interesting how the BufferReader parsing ints in a loop (ParseInt32BufferReader) is slower than manually parsing the buffer inside the reader (ParseInt32BufferReaderRaw). I suspect it's because of the additional call to Advance (to advance the delimiter), but we should investigate and fix the perf problem. Without the fix, I am not sure how viable is the reader.

cc: @joshfree, @pakrym


tests/Benchmarks/Buffers/Reader.cs
```diff
@@ -0,0 +1,65 @@
...
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using System;
+using System.Buffers;
+using System.Buffers.Reader;
+using System.Buffers.Text;
+using System.Text;
+
+public class Reader
```
KrzysztofCwalina: @adamsitnik, I think I don't like the convention of calling the test types with a single word, e.g. "Reader". The reason is that the type names are used as the "menu" items in the BenchmarkSwitcher's menu. I suspect we will have several readers and it will be impossible to tell them apart in the menu.

PR #2243, opened 2018-05-02
Ported HttpParserBench to BenchmarkDotNet

KrzysztofCwalina:
``` ini

BenchmarkDotNet=v0.10.14.525-nightly, OS=Windows 10.0.16299.371 (1709/FallCreatorsUpdate/Redstone3)
Intel Core i7-6700K CPU 4.00GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
Frequency=3914067 Hz, Resolution=255.4887 ns, Timer=TSC
.NET Core SDK=2.2.0-preview1-007460
  [Host]     : .NET Core 2.2.0-preview1-26424-02 (CoreCLR 4.6.26423.05, CoreFX 4.6.26423.04), 64bit RyuJIT
  DefaultJob : .NET Core 2.2.0-preview1-26424-02 (CoreCLR 4.6.26423.05, CoreFX 4.6.26423.04), 64bit RyuJIT


```
|      Method |     Mean |     Error |    StdDev |
|------------ |---------:|----------:|----------:|
| RequestLine | 136.0 ns | 0.1786 ns | 0.1583 ns |
|     Headers | 251.0 ns | 1.2776 ns | 1.1951 ns |
| FullRequest | 327.8 ns | 0.2701 ns | 0.2526 ns |


tests/Benchmarks/Http/HttpParser.cs
```diff
@@ -0,0 +1,111 @@
...
+    {
+        //Method = method;
+        //Version = version;
+        //Path = PrimitiveEncoder.DecodeAscii(path);
+        //Query = PrimitiveEncoder.DecodeAscii(query);
+        //Target = PrimitiveEncoder.DecodeAscii(target);
+    }
+}
+
+struct RequestStruct : IHttpHeadersHandler, IHttpRequestLineHandler
```
adamsitnik: @krwq  this type is not used (and was not before the port), should we keep it or remove it?

tests/Benchmarks/Http/HttpParser.cs
```diff
@@ -0,0 +1,111 @@
...
+
+    private static readonly ReadOnlySequence<byte> s_plaintextTechEmpowerRequestRos = new ReadOnlySequence<byte>(s_plaintextTechEmpowerRequestArray);
+    private static readonly ReadOnlySequence<byte> s_plaintextTechEmpowerHeadersRos = new ReadOnlySequence<byte>(s_plaintextTechEmpowerHeadersArray);
+
+    private static readonly System.Text.Http.Parser.HttpParser s_parser = new System.Text.Http.Parser.HttpParser();
+
+    [Benchmark]
+    public void RequestLine()
+    {
+        var request = new Request();
```
adamsitnik: @KrzysztofCwalina do parsing methods have any side effects on `Request`? if not, maybe we should make it a field too?

KrzysztofCwalina: The code with side effects is commented out. After removing the comments the request cannot be a filed. Also, the cost of creating the request object is not avoidable, so I think it's best include it in the cost.

tests/Benchmarks/Http/HttpParser.cs
KrzysztofCwalina: Yeah, I was thinking about it when I was porting the tests. I actually would like to add the struct option to these tests. But for now, we are just porting, and I will do it as a separate workitem.

PR #2246, opened 2018-05-02
Add advanced options when deriving from PollingFileSystemWatcher

MisinformedDNA:


src/System.IO.FileSystem.Watcher.Polling/System/IO/FileSystemChangesEnumerator.cs
```diff
@@ -32,7 +32,12 @@ protected override bool ShouldIncludeEntry(ref FileSystemEntry entry)
             if (_currentDirectory == null)
                 _currentDirectory = entry.Directory.ToString();
 
-            return _watcher.IsWatched(ref entry);
+            return _watcher.ShouldIncludeEntry(ref entry);
+        }
+
+        protected override bool ShouldRecurseIntoEntry(ref FileSystemEntry entry)
+        {
+            return _watcher.ShouldRecurseIntoEntry(ref entry);
```
YohDeadfall: nit: I think it would be better to use an expression bodied method here. It's an one liner after all.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
```diff
@@ -89,6 +89,11 @@ internal bool IsWatched(ref FileSystemEntry entry)
             return false;
         }
 
+        protected internal virtual bool ShouldRecurseIntoEntry(ref FileSystemEntry entry)
+        {
+            return true;
```
YohDeadfall: And here too.

MisinformedDNA:
Can we merge this? Any other comments?

JeremyKuhne:
> Can we merge this? Any other comments?

You should add tests for the new extension point. Adding sample projects is something we don't do. Other than that it looks fine.

KrzysztofCwalina:
BTW, I have no opinions on the sample. I like the cleanup/renames in the APIs.

MisinformedDNA:
I'm working on adding unit tests for already merged code: https://github.com/dotnet/corefxlab/pull/2273

I'll add some unit tests here next.

JeremyKuhne:
> I'll add some unit tests here next.

I'm fine with a follow-up as this is the labs repo. I'm even ok with merging this if you remove the sample project as a part of that.

The problem with sample projects is that there is no effective way to maintain them. How do we validate they work? What value do they provide if we can't validate? (Outside of "they compile", which can be done in an existing test project.) They are a tax that can't be absorbed on the scale that we operate. If we embraced them we would have thousands of projects to compile and otherwise maintain. If we had a well constrained process for managing them that didn't add overhead to the core development cycle maybe.

I _do_ think it is good to have scenario tests that cover samples and help ensure that we're explicitly covering what we have written up in blogs/docs/etc. Have thousands of additional tests in _existing_ projects is fine. :)

MisinformedDNA:
Sample removed.

As for testing, I'm not sure how to test this part. There are two main challenges: 1) The methods are `protected` and 2) `FileSystemEntry` is of type `ref struct`.

The first one I got around with Moq, but I can't find a way to get `FileSystemEntry` to get passed in as an argument.

```csharp
    [Fact]
    public static void ShouldIncludeEntryTest()
    {
        Mock<PollingFileSystemWatcher> mock = new Mock<PollingFileSystemWatcher>();
        mock.Protected()
            .Setup<bool>("ShouldIncludeEntry", It.IsAny<FileSystemEntry>())  // The type 'FileSystemEntry' may not be used as a type argument
            .Returns(false)
            .Verifiable();

        var watcher = mock.Object;
        watcher.Start();

        mock.Verify();
    }
```

`It.IsAny<DateTime>` works fine, so I assume that this error has something to do with `ref struct`, but I'm not sure how to work around it.

JeremyKuhne:
> As for testing, I'm not sure how to test this part. 

You have to manually create your own mock as we don't use a mocking framework. Just derive from the watcher and track whatever you need to. An example: https://github.com/dotnet/corefx/blob/master/src/System.IO.FileSystem/tests/Enumeration/ErrorHandlingTests.netcoreapp.cs

MisinformedDNA:
@JeremyKuhne Merge?

Issue #420, opened 2015-11-06
Intermittent test failure in System.IO.FileSystem.Watcher.Polling.Test


krwq:
```
19:03:57   Discovering: System.Text.Formatting.Tests
19:03:57   Discovered:  System.Collections.Tests
19:03:57   Discovered:  System.IO.FileSystem.Watcher.Polling.Tests
19:03:57   Discovered:  System.Text.Formatting.Tests
19:03:57   Starting:    System.Collections.Tests
19:03:58   Starting:    System.IO.FileSystem.Watcher.Polling.Tests
19:03:58   Starting:    System.Text.Formatting.Tests
19:03:58   InvariantFormatStructClr : Elapsed 722ms
19:03:59   
19:03:59   Unhandled Exception: System.ObjectDisposedException: Cannot access a disposed object.
19:03:59      at System.Threading.TimerQueueTimer.Change(UInt32 dueTime, UInt32 period)
19:03:59      at System.IO.FileSystem.PollingWatcher.TimerHandler(Object context)
19:03:59      at System.Threading.ExecutionContext.Run(ExecutionContext executionContext, ContextCallback callback, Object state)
19:03:59      at System.Threading.TimerQueueTimer.CallCallback()
19:03:59      at System.Threading.TimerQueueTimer.Fire()
19:03:59      at System.Threading.TimerQueue.FireNextTimers()
19:03:59   InvariantFormatIntHexClr : Elapsed 668ms
19:04:00 d:\j\workspace\dotnet_corefxlab\Debug_prtest\packages\Microsoft.DotNet.BuildTools.1.0.25-prerelease-00107\lib\tests.targets(129,5): warning MSB3073: The command "CoreRun.exe xunit.console.netcore.exe System.IO.FileSystem.Watcher.Polling.Tests.dll  -xml testResults.xml -notrait category=nonwindowstests  -notrait category=OuterLoop -notrait category=failing " exited with code -532462766. [d:\j\workspace\dotnet_corefxlab\Debug_prtest\src\System.IO.FileSystem.Watcher.Polling\tests\System.IO.FileSystem.Watcher.Polling.Tests.csproj]
[...]

PR #2248, opened 2018-05-02
Migrate Utf8String perf benchmarks to benchmarkdotnet

GrabYourPitchforks:
Ended up using `IParam` due to the complex setup that must be performed before each test case. To reviewers: does this look like a correct conversion?


GrabYourPitchforks:
Need to figure out how not to have `[ParamsSource]` across both fields, since it's causing benchmarkdotnet to try different combinations of field values, even if the method under test doesn't use that particular field. It's causing the benchmark to take 20 minutes to run.

tests/Benchmarks/Utf8String/CodePointEnumeration.cs
```diff
@@ -0,0 +1,62 @@
...
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using BenchmarkDotNet.Attributes;
+using BenchmarkDotNet.Code;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using Utf8StringRealType = System.Text.Utf8.Utf8String;
+
+namespace Benchmarks.Utf8String
```
ahsonkhan: `Benchmarks.SystemTextUtf8String` (with the folder name changed to `Benchmarks/SystemTextUtf8String/`), similar to https://github.com/dotnet/corefxlab/pull/2249/files#diff-9803e8894ed7238070c254047cf81cf7R13?

tests/Benchmarks/Utf8String/ConstructFromString.cs
```diff
@@ -0,0 +1,69 @@
...
+            public object Value { get; }
+
+            public string ToSourceCode()
+            {
+                string valueAsString = (string)Value;
+
+                StringBuilder sb = new StringBuilder("\"");
+                foreach (char ch in valueAsString)
+                {
+                    sb.AppendFormat(CultureInfo.InvariantCulture, "\\u{0:X4}", (uint)ch);
```
ahsonkhan: Can you explain, in comments, why we format the string this way for `ToSourceCode()`?

tests/Benchmarks/Utf8String/CodePointEnumeration.cs
```diff
@@ -0,0 +1,62 @@
...
+{
+    public partial class Utf8String
```
ahsonkhan: Rename this class to CodePointEnumeration and then we can remove `Utf8StringRealType`

tests/Benchmarks/Utf8String/ConstructFromString.cs
```diff
@@ -0,0 +1,69 @@
...
+using BenchmarkDotNet.Code;
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.Text;
+using Utf8StringRealType = System.Text.Utf8.Utf8String;
+
+namespace Benchmarks.Utf8String
+{
+    public partial class Utf8String
```
ahsonkhan: Similarly, rename this to ConstructFromString?

ahsonkhan:
https://ci.dot.net/job/dotnet_corefxlab/job/master/job/ubuntu16.04_debug_prtest/1826/consoleFull#-8536224876a086b3e-df04-41d2-bc4d-43e8f9406d07

FYI @JeremyKuhne, @MisinformedDNA, @krwq 

We should track and fix this intermittent failure (or maybe even disable the test for now).

From https://github.com/dotnet/corefxlab/issues/420
> The three current tests rely on Thread.Sleep and, thus, race conditions.

tests/Benchmarks/Utf8String/CodePointEnumeration.cs
ahsonkhan: I am fine with that.

KrzysztofCwalina:

GrabYourPitchforks:
Waiting for checks to go green. Thanks @ahsonkhan and @adamsitnik for all the discussion! :)

adamsitnik:
I am going to merge it now because the benchmarks are not run as part of CI so there is no need  to wait for OSX build if it can be compiled on other OSes.

PR #2253, opened 2018-05-03
Fix PollingFileSystemWatcher unit tests

MisinformedDNA:
Fixes #420

Issue #2239, opened 2018-05-01
Write tests for PollingFileSystemWatcher


MisinformedDNA:
There are currently only three tests.

PR #2258, opened 2018-05-03
Add safer dispose method

MisinformedDNA:


MisinformedDNA:
@JeremyKuhne per #2253. Is this what you were looking for? My tests look promising, locally at least.

JeremyKuhne:
> Is this what you were looking for?

Part of it. I commented in more detail on the linked issue. https://github.com/dotnet/corefxlab/pull/2253#issuecomment-386451112

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
```diff
@@ -25,6 +25,7 @@ public class PollingFileSystemWatcher : IDisposable
         Timer _timer;
         PathToFileStateHashtable _state; // stores state of the directory
         byte _version; // this is used to keep track of removals. // TODO: describe the algorithm
+        private bool disposed = false;
```
JeremyKuhne: `_disposed` Can you add `private` to the others while you're here?

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
```diff
@@ -131,12 +133,33 @@ internal void UpdateState(string directory, ref FileChangeList changes, ref File
...
         public void Dispose(WaitHandle notifyObject)
         {
-            _timer.Dispose(notifyObject);
+            Dispose(true, notifyObject);
+            GC.SuppressFinalize(this);
+        }
+
+        protected virtual void Dispose(bool disposing, WaitHandle notifyObject = null)
+        {
+            if (disposed) return;
```
JeremyKuhne: I was actually thinking of checking in the callback and not calling `.Change()` again if disposed.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
```diff
@@ -131,12 +133,33 @@ internal void UpdateState(string directory, ref FileChangeList changes, ref File
...
+
+            if (disposing)
+            {
+                if (notifyObject != null)
+                {
+                    _timer.Dispose(notifyObject);
```
JeremyKuhne: Don't do this here. Presumably we always want the timer disposed, even if the derived class doesn't call base.

MisinformedDNA: `_timer` is always disposed. It's just switching which `Dispose` method to use. Am I missing something?

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
MisinformedDNA: Good idea!

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
JeremyKuhne: If a derived class overrides your new virtual Dispose() and never calls .Base() you won't get called. I think it is a good idea to have a virtual Dispose(), we just shouldn't do anything in it in the base class.

MisinformedDNA: Uh, so you don't want me to dispose `_timer`?

JeremyKuhne: > Uh, so you don't want me to dispose _timer?

No, I don't want you to dispose it in a virtual method. Put the dispose calls back into the non-virtual methods.

MisinformedDNA: I was following the pattern here: https://docs.microsoft.com/en-us/dotnet/standard/garbage-collection/implementing-dispose#implementing-the-dispose-pattern-for-a-base-class

MisinformedDNA: Shouldn't it be up to the developer to properly inherit?

JeremyKuhne: > Shouldn't it be up to the developer to properly inherit?

If you can make it more stable, why wouldn't you? Particularly because this one has timed events. Some examples:

1. What if the derived class takes forever and another event fires? Or worse, in rare cases?
2. What if the derived class throws and another event fires because we didn't "unhook"?

Disposals in my experience can go horribly wrong. If we were to expose the timer (e.g. via protected), then I'd think differently in this case. (I'd rather we didn't, btw.)

MisinformedDNA: Is this what you are thinking?

```csharp
        public void Dispose()
        {
            _timer.Dispose();
            Dispose(true);
            GC.SuppressFinalize(this);
        }

        public void Dispose(WaitHandle notifyObject)
        {
            _timer.Dispose(notifyObject);
            Dispose(true);
            GC.SuppressFinalize(this);
        }

        protected virtual void Dispose(bool disposing)
        {
            disposed = true;
        }
```

JeremyKuhne: > Is this what you are thinking?

Closer. If we care about _disposed, we should set it before we call _timer.Dispose() so that we can skip calling _timer.Change(). You might want to consider checking for it in Start() as well- potentially worth throwing ObjectDisposed if that happens.

One possibility is to not add another member and just set _timer to null and check for null where we care.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
```diff
@@ -49,8 +51,10 @@ public PollingFileSystemWatcher(string path, string filter = "*.*", EnumerationO
 
         public void Start()
         {
+            if (_disposed) throw new ObjectDisposedException(nameof(_timer));
```
JeremyKuhne: Should be `nameof(PollingFileSystemWatcher)`.

MisinformedDNA:
@JeremyKuhne Can we just skip the tests, for now, to fix https://github.com/dotnet/corefxlab/pull/2258 and ultimately #420. There's #2239 to ensure we write the tests. @ahsonkhan wants the unit tests either fixed or disabled.

ahsonkhan:
If this change reliably fixes #420, I think we can merge it and add more tests in subsequent PRs.

@dotnet-bot test this please

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
ahsonkhan: @MisinformedDNA, after the CI goes green again, can you submit a commit for this and we can merge this in (barring any other feedback from @JeremyKuhne).

JeremyKuhne:
> If this change reliably fixes #420, I think we can merge it and add more tests in subsequent PRs (running in CI a couple of times to make sure).

I'm ok with postponing this time, *if* the test failures are causing grief. Building up a backlog of tests is much harder to deal with then writing them as you go. We're going to miss coverage by not being able to match test changes to code changes.

ahsonkhan:

ahsonkhan:
>  I already started working on the unit tests, but suddenly Live Testing stopped working.

I am not sure why.

I see similar errors for other tests too:
```text
[14:30:45.446 Info] Started executing tests from D:\GitHub\Fork\corefxlab\.vs\corefxlab\lut\0\t\System.Collections.Sequences.Tests\Debug\netcoreapp2.1\System.Collections.Sequences.Tests.dll.
[14:30:46.450 Error] [TestRunner 1] [xUnit.net 00:00:00.5309356] System.Collections.Sequences.Tests: Exception filtering tests: No tests matched the filter because it contains one or more properties that are not valid (TestCategory). Specify filter expression containing valid properties (DisplayName, FullyQualifiedName) and try again.
[14:30:46.506 Info] [TestRunner 1] No test is available in D:\GitHub\Fork\corefxlab\.vs\corefxlab\lut\0\t\System.Collections.Sequences.Tests\Debug\netcoreapp2.1\System.Collections.Sequences.Tests.dll. Make sure that installed test discoverers & executors, platform & framework version settings are appropriate and try again.
```

Maybe ask someone who works on xunit tests or file an issue: https://github.com/xunit/xunit

PR #2274, opened 2018-05-09
Clean up Utf8StringPerf benchmark

GrabYourPitchforks:
This re-introduces the descriptive strings to the benchmark tests. Also cleaned up the enumeration test to make sure the JIT can't optimize away the enumerator.

### DIFF
--- a/tests/Benchmarks/System.Text.Utf8/Utf8StringPerf.cs
+++ b/tests/Benchmarks/System.Text.Utf8/Utf8StringPerf.cs
@@ -8,32 +8,49 @@ namespace System.Text.Utf8.Benchmarks
 {
     public class Utf8StringPerf
     {
-        [Params(5, 50000)]
-        public int Length;
+        [Params(TestCaseType.ShortAscii, TestCaseType.ShortMultiByte, TestCaseType.LongAscii, TestCaseType.LongMultiByte)]
+        public TestCaseType TestCase;
 
-        [Params(126, 0xD7FF)]
-        public int MaxCodePoint;
-
-        private string randomString;
-        private Utf8String utf8String;
+        private string _string;
+        private Utf8String _utf8String;
 
         [GlobalSetup]
         public void Setup()
         {
+            int length;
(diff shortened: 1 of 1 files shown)


tests/Benchmarks/System.Text.Utf8/Utf8StringPerf.cs
```diff
@@ -8,32 +8,49 @@ namespace System.Text.Utf8.Benchmarks
...
-        public Utf8String ConstructFromString() => new Utf8String(randomString);
+        public Utf8String ConstructFromString() => new Utf8String(_string);
 
         [Benchmark]
-        public void EnumerateCodePoints()
+        public uint EnumerateCodePoints()
         {
-            foreach (uint codePoint in utf8String)
+            uint retVal = default;
+            foreach (uint codePoint in _utf8String)
```
ahsonkhan: > Also cleaned up the enumeration test to make sure the JIT can't optimize away the enumerator.

@adamsitnik, in what scenarios do we have to worry about JIT optimizing away the benchmark?

adamsitnik: @ahsonkhan it depends on the JIT, but as far as I know, JIT does not remove empty loops as of today.

@GrabYourPitchforks change prevents from suffering from any future changes. 

the only thing you need to keep in mind is if you are ever going to be comparing Utf8String vs String, the benchmarks need to do exactly the same thing (loop over it, add elements, return value) to make sure you are comparing apples to apples

GrabYourPitchforks:
I've seen JIT remove empty loops for strings and arrays, but those types are special cased by the JIT.

adamsitnik:
> I've seen JIT remove empty loops for strings and arrays, but those types are special cased by the JIT.

Nice! In case you ever see `0 ns` as the result when running benchmarks with BenchmarkDotNet it means that the benchmark got optimized by JIT to an empty method/method returning const.

PR #2279, opened 2018-05-10
Exploring json deserializer built on top of the utf-8 json reader

ahsonkhan:
Skeleton of a prototype deserializer built on top of the json reader as a starting off point to help with the performance investigation (and help with validating the reader APIs).

- Only supports basic primitive types (Utf8String, int, bool, etc.). TODO: arrays and objects.
- Currently uses reflection to set the properties rather than Ref.Emit.
- Some benchmark code clean up.
- It is too early for real perf tests, but for json payload containing integers and bools only, it is generally on par with json.net, as is.

cc @GrabYourPitchforks, @KrzysztofCwalina


src/System.Text.JsonLab.Dynamic/System.Text.JsonLab.Dynamic.csproj
```diff
@@ -8,6 +8,7 @@
   </PropertyGroup>
   <ItemGroup>
     <PackageReference Include="System.Dynamic.Runtime" Version="$(CoreFxStableVersion)" />
+    <PackageReference Include="System.Reflection.Emit.LightWeight" Version="$(CoreFxStableVersion)" />
```
KrzysztofCwalina: You said you don't yet do ref.emit.

src/System.Text.JsonLab.Dynamic/System/Text/Json/Deserializer.cs
```diff
@@ -0,0 +1,40 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Buffers.Text;
+using System.Collections.Generic;
+using System.Dynamic;
+using System.Reflection;
+using System.Reflection.Emit;
+
+namespace System.Text.JsonLab
```
KrzysztofCwalina: It think it's fine that you changed the package name, but I would not change the namespace name, i.e. I would live it as System.Text.Json.

src/System.Text.JsonLab.Dynamic/System/Text/Json/Deserializer.cs
```diff
@@ -0,0 +1,40 @@
...
+{
+    public partial class JsonDynamicObject : DynamicObject, IBufferFormattable
```
KrzysztofCwalina: Why is the deserializer a DynamicObject? I would think that it's going to deserialize to a strongly typed object, i.e. no need for dynamic.

src/System.Text.JsonLab.Dynamic/System/Text/Json/Deserializer.cs
```diff
@@ -0,0 +1,40 @@
...
+            {
+                DynamicMethod method = new DynamicMethod("CreateInstanceDynamicMethod", typeof(T), null, restrictedSkipVisibility: true);
+
+                // GetConstructors() is only available on netstandard 2.0
+                IEnumerable<ConstructorInfo> ctors = typeof(T).GetTypeInfo().DeclaredConstructors;
+                ConstructorInfo constructor = default;
+                foreach (ConstructorInfo ci in ctors)
+                {
+                    constructor = ci;
+                    break;
```
KrzysztofCwalina: So you just pick the first ctor regardless of parameters?

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -20,6 +21,78 @@ private JsonDynamicObject(Dictionary<JsonProperty, JsonValue> properties)
             _properties = properties;
         }
 
+        private static Dictionary<Type, Dictionary<int, PropertyInfoLinkedList>> TypeCache = new Dictionary<Type, Dictionary<int, PropertyInfoLinkedList>>();
```
KrzysztofCwalina: WOuld be good to add a comment describing what the mapping is. In particular it's not clear what the int is.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -20,6 +21,78 @@ private JsonDynamicObject(Dictionary<JsonProperty, JsonValue> properties)
...
+
+        public static T Deserialize<T>(ReadOnlySpan<byte> utf8)
```
KrzysztofCwalina: Shouldn't the API tell the called how many bytes were consumed?

KrzysztofCwalina: Also, how will this work if the payload is in multiple spans/segments?

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -20,6 +21,78 @@ private JsonDynamicObject(Dictionary<JsonProperty, JsonValue> properties)
...
+        {
+            if (!TypeCache.TryGetValue(typeof(T), out Dictionary<int, PropertyInfoLinkedList> dictionary))
+            {
+                dictionary = GetTypeMap(typeof(T));
+                TypeCache.Add(typeof(T), dictionary);
+            }
```
KrzysztofCwalina: @Maoni0, this is a cache where we might want to experiment with the memory APIs.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -20,6 +21,78 @@ private JsonDynamicObject(Dictionary<JsonProperty, JsonValue> properties)
...
+
+            T instance = Deserializer<T>.Deserialize(utf8);
```
KrzysztofCwalina: BTW, the name of the method above is very confusing. It's not actually deserializing the whole thing; it's just creating the top level instance. correct?

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -20,6 +21,78 @@ private JsonDynamicObject(Dictionary<JsonProperty, JsonValue> properties)
...
+                                pi = GetPropertyInfo(dictionary, key, reader.Value);
+                                pi.SetValue(instance, false);
+                                break;
+                            case JsonValueType.Null:
+                                pi = GetPropertyInfo(dictionary, key, reader.Value);
+                                pi.SetValue(instance, null);
+                                break;
+                            case JsonValueType.Number:
+                                pi = GetPropertyInfo(dictionary, key, reader.Value);
+                                if (!Utf8Parser.TryParse(reader.Value, out int result, out _))
```
KrzysztofCwalina: I think you need to first try long and then double (as opposed to an int)

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -20,6 +21,78 @@ private JsonDynamicObject(Dictionary<JsonProperty, JsonValue> properties)
...
+                                {
+                                    throw new InvalidCastException();
+                                }
+                                pi.SetValue(instance, result);
```
KrzysztofCwalina: BTW, will this deserialized work with structs? i.e. can I deserialize into a struct?

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -218,6 +291,67 @@ private static unsafe bool TryEncodeControlChar(SymbolTable symbolTable, byte va
...
+                hash = hash * 31 + element;
+            }
+            return hash;
+        }
+
+        private static Dictionary<int, PropertyInfoLinkedList> GetTypeMap(Type type)
+        {
+            var dictionary = new Dictionary<int, PropertyInfoLinkedList>();
+            IEnumerable<PropertyInfo> properties = type.GetRuntimeProperties();
+            foreach (PropertyInfo pi in properties)
```
KrzysztofCwalina: Would it make sense for this map to be lazy? i.e. you add new propoerties on demand.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -218,6 +291,67 @@ private static unsafe bool TryEncodeControlChar(SymbolTable symbolTable, byte va
...
+            foreach (byte element in span)
+            {
+                hash = hash * 31 + element;
+            }
+            return hash;
+        }
+
+        private static Dictionary<int, PropertyInfoLinkedList> GetTypeMap(Type type)
+        {
+            var dictionary = new Dictionary<int, PropertyInfoLinkedList>();
```
KrzysztofCwalina: Long term you might want to play with custom comparers that would work better for matching property names. Also, you might consider writing a custom hashtable optmized for this scenario.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -218,6 +291,67 @@ private static unsafe bool TryEncodeControlChar(SymbolTable symbolTable, byte va
...
+            IEnumerable<PropertyInfo> properties = type.GetRuntimeProperties();
+            foreach (PropertyInfo pi in properties)
+            {
+                byte[] encoded = Encoding.UTF8.GetBytes(pi.Name);
+                int hashcode = GetHashCode(encoded);
+
+                if (dictionary.ContainsKey(hashcode))
+                {
+                    dictionary.TryGetValue(hashcode, out PropertyInfoLinkedList list);
```
KrzysztofCwalina: why do you map from hashcodes? and then store lists? This is what dictionary is doing again (in it's implementation).

src/System.Text.JsonLab.Dynamic/System/Text/Json/PropertyInfoLinkedList.cs
```diff
@@ -0,0 +1,37 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Reflection;
+
+namespace System.Text.JsonLab
+{
+    public class PropertyInfoNode
```
KrzysztofCwalina: Why is it public?

src/System.Text.JsonLab.Dynamic/System.Text.JsonLab.Dynamic.csproj
ahsonkhan: I am starting to play with ref.emit (the instance of T returned is constructed using ref emit instead of Activator.CreateInstance). What I meant was that the property setters are still using reflection.

src/System.Text.JsonLab.Dynamic/System/Text/Json/Deserializer.cs
ahsonkhan: I had to rename the namespace, here (https://github.com/dotnet/corefxlab/pull/2265) to avoid conflicts with benchmark tests.

> Renamed System.Text.Json to System.Text.JsonLab for now to avoid naming conflict with the System.Text.Json NuGet package by [YSharp](https://github.com/ysharplanguage/FastJsonParser)

Once we port those tests out of the repo (and the investigation is complete), we can change the namespace back.

src/System.Text.JsonLab.Dynamic/System/Text/Json/Deserializer.cs
ahsonkhan: It was already a DynamicObject. I just left it as is for now.

src/System.Text.JsonLab.Dynamic/System/Text/Json/Deserializer.cs
ahsonkhan: I will add a comment, TODO.

I understand it doesn't cover most scenarios (for instance if the type to deserialize has custom, non-default constructors). Wanted to start with supporting deserializing of simple types first.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: Atm, the entire json payload is expected to be within the input span, so all of the data would be consumed, in which case, bytesConsumed isn't necessary.

ahsonkhan: Good point. Since our reader only works on single segments, I have the deserializer as such. Once the reader has support for multiple spans, I will change the deserializer.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: Yes, for now. I will change the method name. The intention is to have the entire deserialize method (the reader loop) by emitted in IL.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: Yes, I agree. Atm, this is hand-wavy and only support the basic scenario of just ints (for testing purposes). I will add a TODO. 

For now, I am assuming we only have ints.

Naiively, we could start off with something like this:
```C#
pi = GetPropertyInfo(dictionary, key, reader.Value);
Type propertyType = pi.GetType();
object value = default;
if (propertyType == typeof(double))
{
    if (!Utf8Parser.TryParse(reader.Value, out double result, out _))
    {
        throw new InvalidCastException();
    }
    value = result;
}
else if (propertyType == typeof(long))
{
    if (!Utf8Parser.TryParse(reader.Value, out long result, out _))
    {
        throw new InvalidCastException();
    }
    value = result;
}
// ... other numeric types
pi.SetValue(instance, value);
```

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: This is a one time operation (per type), so the cost is amortized over multiple calls to Deserialize. Regardless, we could make it lazy if we generally expect the json payload to partially fill the object.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: This (hashcode, PropertyInfoList) mapping is needed since our json reader returns a `ReadOnlySpan<byte>` as the value which cannot be stored in a collection without allocating. 

So, from the input type T, we have a list of property names on the type (which are string). For fast access of the property info that matches the property name, I calculate the hashcode of the `ReadOnlySpan<byte>` returned by the reader and use that for the look up of the property info. Since it is possible for their to be a hash conflict (where two properties end up with the same hashcode), our dictionary needs to be (int, PropertyInfoList) rather than (int, PropertyInfo). In most cases, the list would only contain a single PropertyInfo anyway.

src/System.Text.JsonLab.Dynamic/System/Text/Json/PropertyInfoLinkedList.cs
ahsonkhan: Will fix.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: Do you mean struct instead of a class? No, it doesn't work atm since the CreateDelegate is looking for a default constructor (which doesn't work for structs).

Also, atm, nested structs/etc. don't work either.

> case JsonValueType.Object:
    throw new NotImplementedException("object support not implemented yet.");

ahsonkhan: Also, pi.SetValue doesn't work on structs either. The ref.emit based setters should fix that.

ahsonkhan:
I am going to merge this in, for now (however, the feature is still very much a work-in-progress).

Issue #2222, opened 2018-04-18
API Proposal: PollingWatcher


MisinformedDNA:
```csharp
namespace System.IO.Enumeration  // was System.IO.FileSystem
{
    public class FileSystemMonitor : IDisposable  // was PollingWatcher
    {
        public FileSystemMonitor(string directory, EnumerationOptions options = null);
        public int PollingInterval { get; set; }
        // REMOVED: public PollingWatcher(string directory, bool includeSubdirectories = false, int pollingIntervalInMilliseconds = 1000);

        public FindPredicate ShouldRecursePredicate { get; set; }
        public FindPredicate ShouldIncludePredicate { get; set; }
        // REMOVED: public void AddExtension(string extension);

        public TraceSource Tracing { get; private set; }

        public void Start();
        public void Start(List<FileState> fileSystemState);
        public ReadOnlyCollection<FileState> GetSystemState();

        public Action Changed;
        public Action<FileChange[]> ChangedDetailed;
        public delegate bool FindPredicate(ref FileSystemEntry entry);

        public void Dispose();
    }

    public struct FileChange
    {
        public ReadOnlySpan<char> Path { get; }  // Full path
        // REMOVED: public string Name { get; }
        // REMOVED/REDUNDANT: public ReadOnlySpan<char> Directory { get; }

        public ChangeType ChangeType { get; }
    }

    public class FileState
    {
        public ReadOnlySpan<char> Path { get; }  // Full Path
        public DateTimeOffset LastWriteTimeUtc { get; }
        public long Length { get; }
        public override string ToString() { get; }

        public FileState(string path, DateTimeOffset lastWriteTimeUtc, long length);
    }
}
```

**Notes:**
 - Not a huge fan of the namespace `Enumeration` in general, but it seems like it belongs with `FileSystemEnumerable`.
 - Renamed to `FileSystemMonitor`. `PollingWatcher` is both not specific enough and is too similar to `FileSystemWatcher`, from which this is a pretty big departure.
 - Changed constructor to accept same parameters as `FileSystemEnumerable` (minus the transform).
 - Added similar predicate members from `FileSystemEnumerable` to pass through.
 - Added a `Start` overload to specify starting file set.
 - Add a method to retrieve file set.
    - The purpose of retrieving and setting the file set is so the user can decide to save state in case the system goes down, or only needs to be run on occassion. This helps prevent changes from being missed.
 - `FileState` is mutable as an internal version number changes; so it's a class

[...]

PR #2280, opened 2018-05-10
Make PollingFileSystemWatcher serializable

MisinformedDNA:
Part of #2222


MisinformedDNA:
@ahsonkhan @JeremyKuhne Can we get this reviewed?

src/System.IO.FileSystem.Watcher.Polling/System/IO/FileState.cs
```diff
@@ -0,0 +1,29 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Diagnostics;
+
+namespace System.IO
+{
+    [Serializable]
+    struct FileState
```
JeremyKuhne: Accessibility modifier

src/System.IO.FileSystem.Watcher.Polling/System/IO/PathToFileStateHashtable.cs
```diff
@@ -4,15 +4,15 @@
...
 
-namespace System.IO.FileSystem
+namespace System.IO
 {
     // this is a quick an dirty hashtable optimized for the PollingFileSystemWatcher
     // It allows mutating struct values (FileState) contained in the hashtable
     // It has optimized Equals and GetHasCode
     // It implements removals by marking values as "removed" (Path==null) and then garbage collecting them when table is resized
+    [Serializable]
     class PathToFileStateHashtable
```
JeremyKuhne: Please fix the accessibility modifiers when you're changing these files

src/System.IO.FileSystem.Watcher.Polling/System/IO/PathToFileStateHashtable.cs
```diff
@@ -138,7 +132,7 @@ private int ComputeBucket(ReadOnlySpan<char> file)
         private void Resize()
         {
             // this is because sometimes we just need to garbade collect instead of increase size
-            var newSize = Math.Max(_count * 2, 4);
+            var newSize = Math.Max(Count * 2, 4);
```
JeremyKuhne: Also try to remove var when you're making changes. Var is ok for tests, but generally not for code.

JeremyKuhne:
@KrzysztofCwalina, why did you want to make the watcher serializable?

src/System.IO.FileSystem.Watcher.Polling/System/IO/FileState.cs
MisinformedDNA: Is there a reason we don't have a solution-wide analyzer to remind us about style guidelines?

MisinformedDNA:
@JeremyKuhne @KrzysztofCwalina 

Part of the purpose of `PollingFileSystemWatcher` is that file events are too important to miss. If that is the case, then we need to account for file changes during downtime. I initially proposed import/export functions, but I was guided to aim for the more consistent `ISerializable`. Whatever direction we go, I believe we need a story for "file changes during downtime".

src/System.IO.FileSystem.Watcher.Polling/System/IO/FileState.cs
JeremyKuhne: > Is there a reason we don't have a solution-wide analyzer to remind us about style guidelines?

You do get the feedback in VS, but it isn't super in your face. I think that was intentional.

image

In any case I'll bring it up with the team.

JeremyKuhne:
> Part of the purpose of PollingFileSystemWatcher is that file events are too important to miss. If that is the case, then we need to account for file changes during downtime.

@MisinformedDNA Ah, I forgot that you were trying to do that. Sorry.

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
```diff
@@ -47,17 +51,44 @@ public PollingFileSystemWatcher(string path, string filter = "*", EnumerationOpt
...
             if (_disposed) throw new ObjectDisposedException(nameof(PollingFileSystemWatcher));
+            if (_started) return;
 
+            _started = true;
             ComputeChangesAndUpdateState(); // captures the initial state
-            _timer.Change(PollingIntervalInMilliseconds, Timeout.Infinite);
+            _timer.Change(PollingInterval, Timeout.Infinite);
+        }
+
+        private void AttemptRestart()
```
JeremyKuhne: This should just be inline as there appears to be only one caller.

PR #2283, opened 2018-05-11
 Add JsonReaderTests and update benchmarks with different json string inputs.

ahsonkhan:
**See the raw results here:** https://gist.github.com/ahsonkhan/4f8dbb517e701e84c2f6d1385afff2e1

image

**Conclusions:**
- JsonLab Reader is 2.5-3.7x faster than Json.NET Reader, in general, across the board.
- JsonLab Reader is better at reading UTF-8 than UTF-16 when compared to Json.NET Reader (Json.NET is forced to use StreamReader which is slower than StringReader)
- When returning a copy of the data, JsonLab Reader allocates 1.1x of the input json size for UTF-8 and 2.2x for UTF-16. Json.NET Reader ends up allocating 6-7x of the input json size for both UTF-8 and UTF-16 (32 bytes more in the case of UTF-16 due to the allocation of a new StringReader).

image

image

image

image

**EmptyLoop just goes through the entire json using each reader, without accessing any value or token:**
```C#
public void ReaderNewtonsoftReaderEmptyLoop()
{
    TextReader reader;
    if (IsUTF8Encoded)
    {
        _stream.Seek(0, SeekOrigin.Begin);
        reader = _reader;
    }
    else
    {
        reader = new StringReader(_jsonString);
    }
    var json = new Newtonsoft.Json.JsonTextReader(reader);
    while (json.Read()) ;
}

public void ReaderSystemTextJsonLabEmptyLoop()
{
    JsonReader json = IsUTF8Encoded ?
        new JsonReader(_dataUtf8, SymbolTable.InvariantUtf8) :
        new JsonReader(_dataUtf16, SymbolTable.InvariantUtf16);

    while (json.Read()) ;
}
```

**ReturnBytes, ReturnString get each value within the json from the reader and concat it together (either as a string output, or byte array).  The results of these loops are essentially identical (which is what our unit tests compare and assert).**
```C#
private string NewtonsoftReturnStringHelper(TextReader reader)
{
    StringBuilder sb = new StringBuilder();
    var json = new Newtonsoft.Json.JsonTextReader(reader);
    while (json.Read())
    {
        if (json.Value != null)
        {
            sb.Append(json.Value).Append(", ");
        }
    }

    return sb.ToString();
}

private byte[] JsonLabReturnBytesHelper(byte[] data, SymbolTable symbolTable, int utf16Multiplier = 1)
{
    byte[] outputArray = new byte[data.Length];

    Span<byte> destination = outputArray;
    var json = new JsonReader(data, symbolTable);
    while (json.Read())
    {
        JsonTokenType tokenType = json.TokenType;
        ReadOnlySpan<byte> valueSpan = json.Value;
        switch (tokenType)
        {
            case JsonTokenType.PropertyName:
                valueSpan.CopyTo(destination);
                destination[valueSpan.Length] = (byte)',';
                destination[valueSpan.Length + 1 * utf16Multiplier] = (byte)' ';
                destination = destination.Slice(valueSpan.Length + 2 * utf16Multiplier);
                break;
            case JsonTokenType.Value:
                var valueType = json.ValueType;

                switch (valueType)
                {
                    case JsonValueType.True:
                        destination[0] = (byte)'T';
                        destination[1 * utf16Multiplier] = (byte)'r';
                        destination[2 * utf16Multiplier] = (byte)'u';
                        destination[3 * utf16Multiplier] = (byte)'e';
                        destination = destination.Slice(4 * utf16Multiplier);
                        break;
                    case JsonValueType.False:
                        destination[0] = (byte)'F';
                        destination[1 * utf16Multiplier] = (byte)'a';
                        destination[2 * utf16Multiplier] = (byte)'l';
                        destination[3 * utf16Multiplier] = (byte)'s';
                        destination[4 * utf16Multiplier] = (byte)'e';
                        destination = destination.Slice(5 * utf16Multiplier);
                        break;
                }

                valueSpan.CopyTo(destination);
                destination[valueSpan.Length] = (byte)',';
                destination[valueSpan.Length + 1 * utf16Multiplier] = (byte)' ';
                destination = destination.Slice(valueSpan.Length + 2 * utf16Multiplier);
                break;
            default:
                break;
        }
    }
    return outputArray;
}
```

**Edit:** Updated the results and gist.

<details>
<summary>Out-dated results </summary>

!~[image](https://user-images.githubusercontent.com/6527137/39912453-9f797fd4-54b3-11e8-87a8-a07be0ebaec7.png)~

**Conclusions:**
- JsonLab is 3-4x faster than Json.NET, in general, across the board.
- JsonLab is better at reading UTF-8 than UTF-16 when compared to Json.NET (Json.NET is forced to use StreamReader which is slower than StringReader)
- When returning a copy of the data, Json.Lab allocates 1x of the input json size for UTF-8 and 2x for UTF-16. Json.NET ends up allocating 10x of the input json size for UTF-8 and 7x for UTF-16.

!~[image](https://user-images.githubusercontent.com/6527137/39912463-a6cca856-54b3-11e8-8ce7-8cc5709b7214.png)~

!~[image](https://user-images.githubusercontent.com/6527137/39912465-aa18e268-54b3-11e8-9cab-9a402f9b56ab.png)~

!~[image](https://user-images.githubusercontent.com/6527137/39912468-acff3022-54b3-11e8-8ac2-2dfe7ab41b62.png)~

!~[image](https://user-images.githubusercontent.com/6527137/39912474-b03d2870-54b3-11e8-9a93-cef82a001245.png)~
</details>

cc @adamsitnik, @KrzysztofCwalina

### DIFF
--- a/tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
+++ b/tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
@@ -10,10 +10,11 @@ using System.IO;
 
 namespace System.Text.JsonLab.Benchmarks
 {
+    // Keep the JsonStrings resource names in sync with TestCaseType enum values.
     public enum TestCaseType
     {
         HelloWorld,
-        Basic,
+        BasicJson,
         BasicLargeNum,
         SpecialNumForm,
         ProjectLockJson,
@@ -51,55 +52,7 @@ namespace System.Text.JsonLab.Benchmarks
         [GlobalSetup]
         public void Setup()
         {
-            switch (TestCase)
-            {
-                case TestCaseType.HelloWorld:
-                    _jsonString = JsonStrings.HelloWorld;
-                    break;
-                case TestCaseType.Basic:
-                    _jsonString = JsonStrings.BasicJson;
-                    break;
-                case TestCaseType.BasicLargeNum:
-                    _jsonString = JsonStrings.BasicJsonWithLargeNum;
-                    break;
-                case TestCaseType.SpecialNumForm:
-                    _jsonString = JsonStrings.JsonWithSpecialNumFormat;
-                    break;
-                case TestCaseType.ProjectLockJson:
-                    _jsonString = JsonStrings.ProjectLockJson;
-                    break;
-                case TestCaseType.FullSchema1:
-                    _jsonString = JsonStrings.FullJsonSchema1;
-                    break;
-                case TestCaseType.FullSchema2:
-                    _jsonString = JsonStrings.FullJsonSchema2;
-                    break;
-                case TestCaseType.DeepTree:
-                    _jsonString = JsonStrings.DeepTree;
-                    break;
-                case TestCaseType.BroadTree:
-                    _jsonString = JsonStrings.BroadTree;
-                    break;
-                case TestCaseType.LotsOfNumbers:
-                    _jsonString = JsonStrings.LotsOfNumbers;
-                    break;
-                case TestCaseType.LotsOfStrings:
-                    _jsonString = JsonStrings.LotsOfStrings;
-                    break;
-                case TestCaseType.Json400B:
-                    _jsonString = JsonStrings.Json400B;
-                    break;
-                case TestCaseType.Json4KB:
-                    _jsonString = JsonStrings.Json4KB;
-                    break;
-                case TestCaseType.Json40KB:
-                    _jsonString = JsonStrings.Json40KB;
-                    break;
-                case TestCaseType.Json400KB:
-                    _jsonString = JsonStrings.Json400KB;
-                    break;
-            }
-
+            _jsonString = JsonStrings.ResourceManager.GetString(TestCase.ToString());
             if (IsUTF8Encoded)
             {
                 _dataUtf8 = Encoding.UTF8.GetBytes(_jsonString);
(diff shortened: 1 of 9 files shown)


tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
```diff
@@ -2,70 +2,277 @@
...
 
-        public static IEnumerable<string> ValuesForJsonString() => new[] { JsonStrings.HeavyNestedJson, JsonStrings.HelloWorld };
+        public static IEnumerable<TestCaseType> TestCaseValues() => (IEnumerable<TestCaseType>)Enum.GetValues(typeof(TestCaseType));
 
         [GlobalSetup]
         public void Setup()
         {
-            _symbolTable = GetTargetEncoder(Target);
-            _data = EncodeTestData(Target, JsonString);
+            switch (TestCase)
```
adamsitnik: this switch grew really big. I would either make it a dictionary to save some space or name enums and string resource keys in the same way to have sth like `_jsonString = JsonStrings.GetString(TestCase.ToString());`

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
```diff
@@ -2,70 +2,277 @@
...
+            }
         }
 
         [Benchmark(Baseline = true)]
-        public void ReaderNewtonsoftStringReader()
+        public void ReaderNewtonsoftReaderEmptyLoop()
         {
-            var json = new Newtonsoft.Json.JsonTextReader(new StringReader(JsonString));
-            while (json.Read()) ;
+            if (IsUTF8Encoded)
```
adamsitnik: I think that it would be nice to simplify the code by reducing the repeated code, an example:

```cs
_stream.Seek(0, SeekOrigin.Begin);

var reader = IsUTF8Encoded ? _reader : new StringReader(_jsonString);
var json = new Newtonsoft.Json.JsonTextReader(reader);
while (json.Read()) ;
```

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
```diff
@@ -2,70 +2,277 @@
...
+        public string ReaderNewtonsoftReaderReturnString()
         {
-            _stream.Seek(0, SeekOrigin.Begin);
-            using (var json = new Newtonsoft.Json.JsonTextReader(_reader))
-                while (json.Read()) ;
+            StringBuilder sb = new StringBuilder();
+            if (IsUTF8Encoded)
+            {
+                _stream.Seek(0, SeekOrigin.Begin);
+                using (var json = new Newtonsoft.Json.JsonTextReader(_reader))
```
adamsitnik: why one of the readers is wrapped with `using` block, and the other is not?

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
```diff
@@ -2,70 +2,277 @@
...
+                    while (json.Read())
+                    {
+                        if (json.Value != null)
+                        {
+                            sb.Append(json.Value + ", ");
```
adamsitnik: nit: I would remove this string concatenation here `sb.Append(json.Value).Append(", ");`

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
```diff
@@ -2,70 +2,277 @@
...
+                            destination[json.Value.Length] = (byte)',';
+                            destination[json.Value.Length + 2] = (byte)' ';
+                            destination = destination.Slice(json.Value.Length + 4);
+                            break;
+                        default:
+                            break;
+                    }
+                }
+                return outputArray;
+            }
```
adamsitnik: I would rather create a single method and make it parameterizable. When the code is being duplicated and we want to measure the difference we can get into a situation where the copies are not identical and the difference might be due some code mismatch. Not different perf

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
ahsonkhan: I don't want to incur the cost of resetting the stream in all cases, but I will reduce the duplicated code.

```C#
TextReader reader;
if (IsUTF8Encoded)
{
    _stream.Seek(0, SeekOrigin.Begin);
    reader = _reader;
}
else
{
    reader = new StringReader(_jsonString);
}
var json = new Newtonsoft.Json.JsonTextReader(reader);
while (json.Read()) ;
```

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
ahsonkhan: Will this have any impact of the performance?

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
ahsonkhan: How do you suggest we parameterize this? ArgumentSource? If so, how can we do the GlobalSetup correctly?

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
ahsonkhan: Fixed to be consistent.

PR #2284, opened 2018-05-11
Fix test case failing in PR #2281

MisinformedDNA:
It was failing to create a directory in the specified location.


tests/System.IO.FileSystem.Watcher.Polling.Tests/Utility/Utility.cs
```diff
@@ -42,4 +42,11 @@ public static void EnsureDelete(string path)
             Directory.Delete(path, true);
         }
     }
+
+    public static string GetRandomDirectory()
+    {
+        var path = Path.Combine(Path.GetTempPath(), Path.GetRandomFileName());
```
ahsonkhan: nit: don't use var

PR #2254, opened 2018-05-03
Ported JsonWriterPerfTests to BenchmarkDotNet

ahsonkhan:
``` ini

BenchmarkDotNet=v0.10.14.525-nightly, OS=Windows 10.0.16299.371 (1709/FallCreatorsUpdate/Redstone3)
Intel Xeon CPU E5-1620 v2 3.70GHz, 1 CPU, 8 logical and 4 physical cores
Frequency=3604597 Hz, Resolution=277.4235 ns, Timer=TSC
.NET Core SDK=2.1.300-rc1-008662
  [Host]     : .NET Core 2.2.0-preview1-26424-02 (CoreCLR 4.6.26423.05, CoreFX 4.6.26423.04), 64bit RyuJIT
  DefaultJob : .NET Core 2.2.0-preview1-26424-02 (CoreCLR 4.6.26423.05, CoreFX 4.6.26423.04), 64bit RyuJIT


```
|                         Method |         Target | Formatted |        Mean |        Error |        StdDev |  Gen 0 | Allocated |
|------------------------------- |--------------- |---------- |------------:|-------------:|--------------:|-------:|----------:|
|      **WriterSystemTextJsonBasic** | **InvariantUtf16** |     **False** | **72,467.4 ns** | **1,360.429 ns** | **1,397.0607 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic | InvariantUtf16 |     False | 18,095.1 ns |   403.423 ns |   414.2858 ns | 0.0610 |     416 B |
| WriterSystemTextJsonHelloWorld | InvariantUtf16 |     False |    548.0 ns |     3.680 ns |     3.0733 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld | InvariantUtf16 |     False |    332.7 ns |     5.978 ns |     4.3226 ns | 0.0486 |     256 B |
|      **WriterSystemTextJsonBasic** | **InvariantUtf16** |      **True** | **97,282.2 ns** | **1,771.764 ns** | **1,570.6226 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic | InvariantUtf16 |      True | 30,431.1 ns |   465.610 ns |   388.8057 ns | 0.0916 |     584 B |
| WriterSystemTextJsonHelloWorld | InvariantUtf16 |      True |    749.0 ns |    12.867 ns |    12.0354 ns |      - |       0 B |
[...]

PR #2291, opened 2018-05-16
 Optimized the JsonWriter APIs and expanded the unit test.

ahsonkhan:
Compared to [previous benchmark results](https://github.com/dotnet/corefxlab/pull/2254#issuecomment-386201486), we are 2-4x faster, which makes the writer faster than Newtonsoft.

**Optimizations:**
- Made ArrayFormatter GetSpan by avoiding unnecessary construction of ArrayFormatter and casts.
- Collapsed several method chains to avoid redundant IsUtf8/IsUtf16/etc. checks, and reduced calls to EnsureBuffer/Advance by pre-calculating exactly how many bytes are required.
- Added separate private methods for each encoding (utf-8, utf-16, slow), which resulted in some code duplication.
- Ported over the optimized Utf8Formatter APIs to WriteDigitsUInt64D, to make WriteValue(long) faster, and skip the StandardFormatter switch statement all together. TODO: Port this fix for all numeric overloads (mainly ulong).
- Optimized EnsureBuffer by removing unnecessary checks.
- Added ThrowHelpers to help with inlining.

**Other fixes:**
- Some code formatting clean-up and fix of `var` usage.

``` ini

Intel Xeon CPU E5-1620 v2 3.70GHz, 1 CPU, 8 logical and 4 physical cores
Frequency=3604598 Hz, Resolution=277.4234 ns, Timer=TSC
.NET Core SDK=2.1.300-rtm-008838
  [Host]     : .NET Core 2.1.0-rtm-26514-02 (CoreCLR 4.6.26514.02, CoreFX 4.6.26514.02), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0-rtm-26514-02 (CoreCLR 4.6.26514.02, CoreFX 4.6.26514.02), 64bit RyuJIT


```
|                         Method |         Target | Formatted |        Mean |       Error |      StdDev |      Median |  Gen 0 | Allocated |
|------------------------------- |--------------- |---------- |------------:|------------:|------------:|------------:|-------:|----------:|
|      **WriterSystemTextJsonBasic** | **InvariantUtf16** |     **False** | **14,794.9 ns** | **136.8339 ns** | **114.2625 ns** | **14,760.6 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic | InvariantUtf16 |     False | 17,417.6 ns | 258.7934 ns | 242.0755 ns | 17,504.0 ns | 0.0610 |     416 B |
| WriterSystemTextJsonHelloWorld | InvariantUtf16 |     False |    149.4 ns |   0.2963 ns |   0.2474 ns |    149.4 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld | InvariantUtf16 |     False |    282.2 ns |   4.9845 ns |   4.6625 ns |    279.6 ns | 0.0486 |     256 B |
|      **WriterSystemTextJsonBasic** | **InvariantUtf16** |      **True** | **15,751.3 ns** | **298.0006 ns** | **292.6764 ns** | **15,703.0 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic | InvariantUtf16 |      True | 31,092.0 ns |  72.0731 ns |  42.8896 ns | 31,089.7 ns | 0.0916 |     584 B |
| WriterSystemTextJsonHelloWorld | InvariantUtf16 |      True |    151.3 ns |   2.9176 ns |   2.8654 ns |    153.2 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld | InvariantUtf16 |      True |    384.4 ns |   1.9225 ns |   1.3901 ns |    384.1 ns | 0.0806 |     424 B |
|      **WriterSystemTextJsonBasic** |  **InvariantUtf8** |     **False** | **15,637.4 ns** | **320.2385 ns** | **785.5505 ns** | **15,267.8 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |  InvariantUtf8 |     False | 17,951.5 ns | 335.2059 ns | 313.5518 ns | 17,902.8 ns | 0.0610 |     416 B |
| WriterSystemTextJsonHelloWorld |  InvariantUtf8 |     False |    207.7 ns |   0.6127 ns |   0.4053 ns |    207.6 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |  InvariantUtf8 |     False |    285.1 ns |   4.2451 ns |   3.5449 ns |    286.4 ns | 0.0486 |     256 B |
|      **WriterSystemTextJsonBasic** |  **InvariantUtf8** |      **True** | **16,288.8 ns** | **224.7983 ns** | **210.2764 ns** | **16,222.2 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |  InvariantUtf8 |      True | 30,375.8 ns | 472.5879 ns | 442.0589 ns | 30,474.8 ns | 0.0916 |     584 B |
| WriterSystemTextJsonHelloWorld |  InvariantUtf8 |      True |    214.5 ns |   4.5213 ns |   4.6430 ns |    212.8 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |  InvariantUtf8 |      True |    386.6 ns |   0.8402 ns |   0.7016 ns |    386.5 ns | 0.0806 |     424 B |

### DIFF
--- a/src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
@@ -1,11 +1,12 @@
 ﻿// Copyright (c) Microsoft. All rights reserved.
 // Licensed under the MIT license. See LICENSE file in the project root for full license information.
 
+using System.Buffers;
 using System.Buffers.Text;
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 using System.Text.Formatting;
-using System.Text.Utf8;
 
 namespace System.Text.JsonLab
 {
@@ -55,25 +56,65 @@ namespace System.Text.JsonLab
         {
             if (UseFastUtf8)
             {
-                WriteItemSeperatorUtf8();
-                WriteSpacingUtf8(false);
-                WriteControlUtf8(JsonConstants.OpenBrace);
+                WriteStartUtf8(CalculateStartBytesNeeded(sizeof(byte)), JsonConstants.OpenBrace);
             }
             else if (UseFastUtf16)
             {
-                WriteItemSeperatorUtf16();
-                WriteSpacingUtf16(false);
-                WriteControlUtf16(JsonConstants.OpenBrace);
+                WriteStartUtf16(CalculateStartBytesNeeded(sizeof(char)), JsonConstants.OpenBrace);
             }
             else
             {
-                WriteItemSeperatorSlow();
-                WriteSpacingSlow(false);
-                WriteControlSlow(JsonConstants.OpenBrace);
+                WriteStartSlow(JsonConstants.OpenBrace);
             }
 
             _firstItem = true;
             _indent++;
+        }
+
+        private void WriteStartUtf8(int bytesNeeded, byte token)
+        {
+            Span<byte> byteBuffer = EnsureBuffer(bytesNeeded);
+
+            int idx = 0;
+            if (!_firstItem)
+                byteBuffer[idx++] = JsonConstants.ListSeperator;
+
+            if (_prettyPrint)
+            {
+                int indent = _indent;
+
+                while (indent-- >= 0)
+                {
+                    byteBuffer[idx++] = JsonConstants.Space;
+                    byteBuffer[idx++] = JsonConstants.Space;
+                }
+            }
+
+            byteBuffer[idx] = token;
+            _bufferWriter.Advance(bytesNeeded);
+        }
+
+        private void WriteStartUtf16(int bytesNeeded, byte token)
+        {
+            Span<char> charBuffer = MemoryMarshal.Cast<byte, char>(EnsureBuffer(bytesNeeded));
+
+            int idx = 0;
+            if (!_firstItem)
+                charBuffer[idx++] = (char)JsonConstants.ListSeperator;
+
+            if (_prettyPrint)
+            {
+                int indent = _indent;
+
+                while (indent-- >= 0)
+                {
+                    charBuffer[idx++] = (char)JsonConstants.Space;
+                    charBuffer[idx++] = (char)JsonConstants.Space;
+                }
+            }
+
+            charBuffer[idx] = (char)token;
+            _bufferWriter.Advance(bytesNeeded);
         }
 
         /// <summary>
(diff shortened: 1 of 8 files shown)


src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
```diff
@@ -145,8 +531,23 @@ public void WriteAttribute(string name, string value)
         /// <param name="value">The signed integer value to be written to JSON data.</param>
         public void WriteAttribute(string name, long value)
         {
-            WriteStartAttribute(name);
-            WriteNumber(value);
+            if (UseFastUtf8)
+            {
+                ReadOnlySpan<byte> nameSpan = MemoryMarshal.AsBytes(name.AsSpan());
+                int bytesNeeded = CalculateStartAttributeBytesNeeded(nameSpan, sizeof(byte));
```
KrzysztofCwalina: Is it faster to calculate and then write? Why not be optimistic and use TryWrite pattern?

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
```diff
@@ -235,21 +820,230 @@ public void WriteValue(string value)
...
+                value /= 10;
+                buffer[i] = (byte)(temp - (value * 10));
+            }
+
+            Debug.Assert(value < 10);
+            buffer[0] = (byte)('0' + value);
+        }
+
+        private void WriteValueUtf8(long value, int bytesNeeded)
+        {
```
KrzysztofCwalina: What can we do to Utf8Formatter to be able to use it here directly?

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
```diff
@@ -568,22 +1511,237 @@ private void WriteSpacingUtf16(bool newline)
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private Span<byte> EnsureBuffer(int needed = 256)
```
KrzysztofCwalina: Yeah, as I said above, I am not thrilled about all the calls to Ensure, Calculate, and other checks. I think we should write everything with Try methods, as in most cases the buffer will be large enough.

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
ahsonkhan: We have to take a similar approach as we did for the default format, where we have a short-circuit for it: https://github.com/dotnet/corefx/pull/25424, but I am not sure how feasible it will be for a general purpose formatter.

Currently, we have the overhead of a switch statement for the 'D' Standard Format and then we have unnecessary checks based on the precision (which is default for our JsonWriter, i.e. NoPrecision).

It might be sufficient to keep the switch statement, but add an `if (precision.IsDefault)` short-circuit in all the implementations. That alone might get us most of the way there.

```C#
        TryFormat(long value, Span<byte> destination, out int bytesWritten, StandardFormat format = default)
            => TryFormatInt64(value, 0xffffffffffffffff, destination, out bytesWritten, format);


        private static bool TryFormatInt64(long value, ulong mask, Span<byte> destination, out int bytesWritten, StandardFormat format)
        {
            if (format.IsDefault)
            {
                  ...
            }
            switch (format.Symbol)
            {
                  ...
                  case 'D':
                    return TryFormatInt64D(value, format.Precision, destination, out bytesWritten);
            }
        ...

        private static bool TryFormatInt64D(long value, byte precision, Span<byte> destination, out int bytesWritten)
        {
            ...
            return TryFormatUInt64D((ulong)value, precision, destination, insertNegationSign, out bytesWritten);
        }

        private static bool TryFormatUInt64D(ulong value, byte precision, Span<byte> destination, bool insertNegationSign, out int bytesWritten)
        {
            int leadingZeroCount = ((precision == StandardFormat.NoPrecision) ? 0 : (int)precision) - digitCount;
            if (leadingZeroCount < 0)
            {
                leadingZeroCount = 0;
            }

            int requiredBufferLength = digitCount + leadingZeroCount;
```

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
ahsonkhan: How would we figure out when to return false when using the TryWrite pattern? The only way I can think of is to call GetSpan() and compare it's length to how much we need (repeatedly). We have to compute bytes needed (or some upper bound) and make at least one call to GetSpan to do that anyway, and those calls are expensive (interface dispatch). Therefore, to make sure we make only one call to GetSpan and get at least what we need, we have to pre-compute the bytes needed (which is a relatively cheap operation). The alternative is several calls to GetSpan (which is what we originally had), or over-asking unnecessarily.

>  In most cases the buffer is large enough

That behavior is already built into the GetSpan implementation in ArrayFormatter, where we just return back the FreeSpan (without doing the expensive resize operation) when we have enough space.

The same issue applies with Advance, which is another benefit of pre-computing. We reduce the calls to Advance to just one, since we know exactly how much we need to move, ahead of time.

KrzysztofCwalina: It depends what you will be doing with this buffer. ILet's say you will be copying static data into it: use TryCopy. Or formar and int, use TryFormat, etc. i.e. turtles all the way down (ok, maybe all the way to the indexer setting a value in the span).

ahsonkhan: Interesting, I tried the optimistic approach, and for methods where pre-computing is expensive, we do get a decent win! I will file an issue for this to apply "optimistic writing" to specific methods where pre-computing is non-trivial. We save ~15% for instance, for writing a string attribute (`WriteAttribute(string, string)`), where calculating the bytes needed requires a call to the encoder.

But, for methods like `WriteValue(long)`, where pre-computing the bytes needed is trivial (and we have to call DigitCount anyway), it is worse to use the TryWrite pattern (~20% slower).

I am not 100% certain that these savings from the TryWrite pattern can successfully offset the cost to re-do the work when we have to expand the buffer. They probably can, if we apply it selectively.

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
KrzysztofCwalina: We should consider exposing TryFormatInt64D (and others). If we need it, our customers will too.
cc: @GrabYourPitchforks

PR #2298, opened 2018-05-17
Update Roslyn compiler to latest, language version to latest

GrabYourPitchforks:
Also includes a workaround where latest netcore compiler doesn't work in VS, use the one built in to VS instead if compiling from the IDE. I've confirmed that command-line build continues to work correctly, and we can now build and run tests from within VS2017 using the latest language features.

This PR replaces https://github.com/dotnet/corefxlab/pull/2161.

@agocke, we also tried the below props, but it resulted in errors at build time (only within the IDE).

```xml
  <ItemGroup Condition="'$(MSBuildRuntimeType)' == 'Core'">
    <PackageReference Include="Microsoft.NETCore.Compilers" Version="$(RoslynVersion)" PrivateAssets="All" />
  </ItemGroup>
  <ItemGroup Condition="'$(MSBuildRuntimeType)' == 'Full'">
    <PackageReference Include="Microsoft.Net.Compilers" Version="$(RoslynVersion)" PrivateAssets="All" />
  </ItemGroup>
```

```txt
Error	MSB4062	The "Microsoft.CodeAnalysis.BuildTasks.Csc" task could not be loaded from the assembly C:\Users\levib\.nuget\packages\microsoft.net.compilers\2.8.2-beta6-62916-08\build\..\tools\Microsoft.Build.Tasks.CodeAnalysis.dll. Could not load file or assembly 'System.Runtime, Version=4.2.0.0, Culture=neutral, PublicKeyToken=b03f5f7f11d50a3a' or one of its dependencies. The system cannot find the file specified. Confirm that the <UsingTask> declaration is correct, that the assembly and all its dependencies are available, and that the task contains a public class that implements Microsoft.Build.Framework.ITask.	System.Text.JsonLab.Tests	C:\Users\levib\.nuget\packages\microsoft.net.compilers\2.8.2-beta6-62916-08\tools\Microsoft.CSharp.Core.targets	52	
```

The particular project I'm compiling targets netcoreapp2.1, so I would've assumed that the _System.Runtime_ dependency would've been normalized. Any ideas on how we can use the preferred switching on the _MSBuildRuntimeType_ property?

### DIFF
--- a/tools/dependencies.props
+++ b/tools/dependencies.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <CoreFxStableVersion>4.3.0</CoreFxStableVersion>
-    <RoslynVersion>2.8.2-beta6-62916-08</RoslynVersion>
+    <RoslynVersion>2.8.0-beta4</RoslynVersion>
     <SystemMemoryVersion>4.6.0-preview1-26517-01</SystemMemoryVersion>
     <SystemCompilerServicesUnsafeVersion>4.6.0-preview1-26517-01</SystemCompilerServicesUnsafeVersion>
     <SystemNumericsVectorsVersion>4.6.0-preview1-26517-01</SystemNumericsVectorsVersion>
--- a/tools/dependencies.props
+++ b/tools/dependencies.props
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <CoreFxStableVersion>4.3.0</CoreFxStableVersion>
-    <RoslynVersion>2.6.0-beta3-62316-02</RoslynVersion>
+    <RoslynVersion>2.8.2-beta6-62916-08</RoslynVersion>
     <SystemMemoryVersion>4.6.0-preview1-26517-01</SystemMemoryVersion>
     <SystemCompilerServicesUnsafeVersion>4.6.0-preview1-26517-01</SystemCompilerServicesUnsafeVersion>
     <SystemNumericsVectorsVersion>4.6.0-preview1-26517-01</SystemNumericsVectorsVersion>
--- a/tools/common.props
+++ b/tools/common.props
@@ -13,9 +13,9 @@
     <SignAssembly>true</SignAssembly>
     <PublicSign Condition=" '$(OS)' != 'Windows_NT' ">true</PublicSign>
     <AssemblyOriginatorKeyFile>$(MSBuildThisFileDirectory)Key.snk</AssemblyOriginatorKeyFile>
-    <LangVersion>7.2</LangVersion>
+    <LangVersion>latest</LangVersion>
   </PropertyGroup>
-  <ItemGroup>
+  <ItemGroup Condition="'$(BuildingInsideVisualStudio)' != 'true'">
     <PackageReference Include="Microsoft.NETCore.Compilers" Version="$(RoslynVersion)" PrivateAssets="All" />
   </ItemGroup>
 </Project>


tools/dependencies.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <CoreFxStableVersion>4.3.0</CoreFxStableVersion>
-    <RoslynVersion>2.6.0-beta3-62316-02</RoslynVersion>
+    <RoslynVersion>2.8.2-beta6-62916-08</RoslynVersion>
```
ahsonkhan: Should we use beta4 (dev15.8, instead of dev15.8p1), or do you need bleeding edge features as they come out?

Buildtools for corefx is using beta4. https://github.com/dotnet/buildtools/blob/4aa764b049d9df84e57f89d0450951be99934593/src/Microsoft.DotNet.Build.Tasks/PackageFiles/Roslyn.Common.props#L4

If we want to stick with the latest, we should consider auto-updating this version by hooking into maestro (like the other dependency updates auto PRs): https://github.com/dotnet/versions/blob/master/build-info/dotnet/roslyn/dev15.8/Latest.txt#L1

GrabYourPitchforks: 2.8.0-beta4 should be reasonable. We can update as needed.

MisinformedDNA:
I'll take a look.

tools/dependencies.props
```diff
@@ -1,7 +1,7 @@
 <Project>
   <PropertyGroup>
     <CoreFxStableVersion>4.3.0</CoreFxStableVersion>
-    <RoslynVersion>2.6.0-beta3-62316-02</RoslynVersion>
+    <RoslynVersion>2.8.0-beta4</RoslynVersion>
```
ahsonkhan: I can't find 2.8.0-beta4 version of Microsoft.NETCore.Compilers within the build tools myget feed:
https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.NETCore.Compilers

There is a 2.8.0-beta4 version of Microsoft.Net.Compilers only:
https://dotnet.myget.org/feed/dotnet-buildtools/package/nuget/Microsoft.NET.Compilers

It looks like we are pulling this package from nuget.org: 
https://www.nuget.org/packages/Microsoft.NETCore.Compilers/2.8.0-beta4

MisinformedDNA:
I can't repo the test failure. Is there a way to see what specific test failed?

GrabYourPitchforks:
I don't see any particular test having failed - looks like maybe the test runner itself timed out? Is there any known flakiness with this test?

MisinformedDNA:
Yes, we always had intermittent failures with these tests. They seems to be happening less often than before, but still not 100%.

And while it doesn't list the test that failed, we do see this error message

> 13:20:38 The active test run was aborted. Reason: Unhandled Exception: System.ObjectDisposedException: Cannot access a disposed object.

It also says it errored out on line 208. But line [206](https://github.com/dotnet/corefxlab/blob/master/src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs#L206) checks that it isn't disposed.

MisinformedDNA:
Hmmm... I got it to repro, but still not getting a specific test case. I'm going to try and narrow it down.

GrabYourPitchforks:
@dotnet-bot test Windows_NT Release Build and Test

For reference, the failing test run is at https://ci.dot.net/job/dotnet_corefxlab/job/master/job/windows_nt_release_prtest/1906/.

GrabYourPitchforks:

PR #2304, opened 2018-05-24
Fix System.reflecion.metadata.cil to use latest S.R.Metadata package

safern:
I'm trying to bring this back to life on my spare time to be able to have a cross-plat kind of ildasm based on this.

cc: @ahsonkhan @nguerrera

### DIFF
--- a/src/System.Reflection.Metadata.Cil/Decoder/CilTypeProvider.cs
+++ b/src/System.Reflection.Metadata.Cil/Decoder/CilTypeProvider.cs
@@ -4,7 +4,6 @@
 
 using System.Collections.Immutable;
 using System.Diagnostics;
-using System.Reflection.Metadata.Decoding;
 using System.Text;
 
 namespace System.Reflection.Metadata.Cil.Decoder
@@ -12,7 +11,7 @@ namespace System.Reflection.Metadata.Cil.Decoder
     /* 
         TO DO: Change Signature decoder, to decode type with byte indicating whether should have the class or valutype prefix or not.
     */
-    public struct CilTypeProvider : ISignatureTypeProvider<CilType>
+    public struct CilTypeProvider : ISignatureTypeProvider<CilType, object>
     {
         private readonly MetadataReader _reader;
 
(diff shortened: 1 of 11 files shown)


src/System.Reflection.Metadata.Cil/SignatureDecoder.cs
```diff
@@ -44,16 +44,16 @@ internal static ImmutableArray<CilType> DecodeMethodSpecificationSignature(BlobH
             return NewDecoder(provider).DecodeMethodSpecificationSignature(ref blobReader);
         }
 
-        internal static CilType DecodeType(EntityHandle handle, CilTypeProvider provider, bool? isValueType)
+        internal static CilType DecodeType(EntityHandle handle, CilTypeProvider provider, byte? isValueType)
```
nguerrera: This parameter name looks wrong now. Should be rawTypeCode, I believe.

safern: Yes, that is correct. What does that rawTypeCode means? 0 if it is class and 1 if it is valuetype?

src/System.Reflection.Metadata.Cil/Decoder/CilTypeProvider.cs
```diff
@@ -197,27 +196,27 @@ public CilType GetSZArrayType(CilType elementType)
...
             CilType type = new CilType(GetFullName(Reader.GetTypeDefinition(handle)), _isValueType, _isClass);
             return type;
         }
 
-        public CilType GetTypeFromReference(MetadataReader reader, TypeReferenceHandle handle, SignatureTypeHandleCode code)
+        public CilType GetTypeFromReference(MetadataReader reader, TypeReferenceHandle handle, byte rawTypeKind)
         {
-            bool _isValueType = (code == SignatureTypeHandleCode.Class);
-            bool _isClass = (code == SignatureTypeHandleCode.ValueType);
+            bool _isValueType = rawTypeKind == 1;
```
joshfree: can these be a byte backed enum instead of just hardcoded values?

PR #2308, opened 2018-05-29
Initial commit of GPIO Planning documents

ianhays:


src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace System.Devices.Gpio
```
KrzysztofCwalina: Are we going to vahe more devices? If not, I would put it in System.IO.Ports.Gpio (maybe?)

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+    /// of pins. At the very least, we should support these.
+    /// - Open a object that is a representation of a pin with the given pin number
+    /// - Support closing a pin to release the resources owned by that pin object
+    /// - Represent the mode of the pin that details how the pin handles reads and writes (e.g. Input, Output)
+    /// - Allow a resistor to be added to a pin such that it can be set as pullup or pulldown (or no pull)
+    /// - Support setting a PWM value on a pin
+    /// </summary>
+    #region GPIO - Basic Scenarios
+
+    public partial class GPIOPin : IDisposable
```
KrzysztofCwalina: This does not follow our casing guidelines. It should be GpioPin. Also, I would remove "partial" from API spec. It's a source code concept.

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+    {
+        int PinNumber { get { throw new NotImplementedException(); } }
+        bool Read() { throw new NotImplementedException(); }
+        void Write(bool value) { throw new NotImplementedException(); }
+        void Dispose() { throw new NotImplementedException(); }
```
KrzysztofCwalina: Why is it disposable?

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+        GPIOPin(GPIOController controller, int pinNumber, GPIOPinMode pinMode = GPIOPinMode.Input) { }
+
+        GPIOPinMode PinMode { get { throw new NotImplementedException(); } set { throw new NotImplementedException(); } }
```
KrzysztofCwalina: The property should be called "Mode". It's already on the Pin type.

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+        bool IsPinModeSupported(GPIOPinMode pinMode) { throw new NotImplementedException(); }
+        int PWMValue { set { throw new NotImplementedException(); } }
```
KrzysztofCwalina: I searched for PVM and nothing interesting came up in Bing and Google search. This means we should not use the acronym. :-)

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+    }
+
+    public partial class GPIOController : IDisposable
+    {
+        GPIOPin this[int pinNumber] { get { throw new NotImplementedException(); } }
```
KrzysztofCwalina: What's the difference between getting a pin by opening it or by using the indexer?

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+        GPIOPin OpenPin(int pinNumber) { throw new NotImplementedException(); }
+        void ClosePin(int pinNumber) { throw new NotImplementedException(); }
```
KrzysztofCwalina: Why is it here and not on Pin?

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+        void ClosePin(GPIOPin pin) { throw new NotImplementedException(); }
```
KrzysztofCwalina: Why here and not on Pin?

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+        int PinCount() { throw new NotImplementedException(); }
+        IEnumerable<GPIOPin> ConnectedPins { get { throw new NotImplementedException(); } }
+        void Dispose() { throw new NotImplementedException(); }
+    }
```
KrzysztofCwalina: Why do we need this controller at all. Why not:
`c#
var pin = GpioPin.Create(5);
pin.Set();
pin.Clear();
pin.Close();
```

or something like that.

File: src/System.Devices.Gpio/GPIORef.cs [hunk 10]
```diff
@@ -0,0 +1,188 @@
...
+
+    public enum GPIOPinMode
+    {
+        Pull_Down,
```
KrzysztofCwalina: You should not have underscores in public APIs (per guidelines).

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+    /// GPIO - Advanced Scenarios
+    /// - Analog Reads and Writes - Most GPIO works with digital pins, but sometimes analog pins are used. The difference in the 
+    ///     Analog pins is that they have a range of potential values instead of just being on/off like the digital pins.
+    /// - Listeners - Polling, interrupts, etc. There should be some way to listen for a change and respond accordingly
+    /// - Edge Detection - Used with listeners/eventing as a way of definining the circumstances under which an event/callback will be raised
+    /// - Allow setting a Debounce duration to ignore quickly occuring events during some timespan.
+    /// </summary>
+    #region GPIO - Advanced Scenarios
+
+    public partial class GPIOPin
```
KrzysztofCwalina: Should there be a separate type for analog pin?

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+
+    // BCM vs BOARD
+    public partial class GPIOController
+    {
+        GPIOController(GPIOScheme numbering = GPIOScheme.BOARD) { throw new NotImplementedException(); }
+    }
+
+    public enum GPIOScheme
+    {
+        BOARD,
```
KrzysztofCwalina: Casing

src/System.Devices.Gpio/GPIORef.cs
```diff
@@ -0,0 +1,188 @@
...
+        BCM
+    }
+
+    public partial class GPIOPin
+    {
+        // Waiters
+        public bool ReadWait(TimeSpan timeout) { throw new NotImplementedException(); }
```
KrzysztofCwalina: Could these be async APIs?

KrzysztofCwalina: Can we support cancellations?

src/System.Devices.Gpio/Planning.md
```diff
@@ -0,0 +1,403 @@
+# Goals
```
KrzysztofCwalina: Why is this file here and not in https://github.com/dotnet/corefxlab/tree/master/docs/specs

KrzysztofCwalina:
Cool! Glad we started working on this! Good first stab at the space.

src/System.Devices.Gpio/GPIORef.cs
ahsonkhan: [Pulse-width_modulation](https://www.bing.com/search?q=pwm&qs=n&form=QBRE&sp=-1&pq=pwm&sc=8-3&sk=&cvid=6AB8F0887B0E4283B00ADFE3A6D8B683)

PR #2328, opened 2018-06-02
Add new overloads to Clone() methods on subtypes of Tensor<T>

i3arnon:
Addresses #1976 

Since a class can't declare 2 methods differing only in the return type I added a `protected abstract CloneTensor` to support the polymorphism.

So there are:
- `new ConcreteTensor.Clone` that returns `ConcreteTensor`
- `Tensor.Clone` that calls `ConcreteTensor.CloneTensor` that in turn calls `ConcreteTensor.Clone`

Personally I would name `CloneTensor` `CloneInternal`. Not sure what you prefer here.

### DIFF
--- a/src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
@@ -391,7 +391,7 @@ namespace System.Numerics
             return new CompressedSparseTensor<T>(values.ToArray(), compressedCounts.ToArray(), indices.ToArray(), nonZeroCount, dimensions, IsReversedStride);
         }
 
-        protected override Tensor<T> CloneTensor() => Clone();
+        protected override Tensor<T> CloneInternal() => Clone();
 
         /// <summary>
         /// Creates a new Tensor of a different type with the specified dimensions and the same layout as this tensor with elements initialized to their default value.
--- a/src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
+++ b/src/System.Numerics.Tensors/System/Numerics/CompressedSparseTensor.cs
@@ -386,10 +386,12 @@ namespace System.Numerics
         /// Creates a shallow copy of this tensor, with new backing storage.
         /// </summary>
         /// <returns>A shallow copy of this tensor.</returns>
-        public override Tensor<T> Clone()
+        public new CompressedSparseTensor<T> Clone()
         {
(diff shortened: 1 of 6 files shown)


ericstj:
It makes me nervous that we rely on folks only implementing CloneTensor by calling Clone in it and requiring that folks implement new Clone in a particular way.  That's a few cases where people need to read the manual to get it right.  It also makes me nervous that we rely on 'new'.  I like your idea though, very clever.

What if we did it as an extension method instead using the *good* name for the extension method and the then use a more descriptive name for the abstract?
```C#
    public static class TensorExtensions
    {
        public static TTensor Clone<TTensor, T>(this TTensor tensor) where TTensor : Tensor<T>
        {
            TTensor copy = tensor.DeepClone() as TTensor;

            return copy;
        }
    }
```

Another option, if we didn't want to use extension methods is to put it on the static Tensor class, where we have other cases where we want to benefit from generic type inference.  /cc @eerhardt @KrzysztofCwalina

i3arnon:
> It makes me nervous that we rely on folks only implementing CloneTensor by calling Clone in it and requiring that folks implement new Clone in a particular way

Before this PR you already relied on them implementing the `Clone` override correctly. Now they need to implement `CloneTensor` instead. If they just do that everything still works, they just don't have the extra benefit of the `new Clone` returning the concrete type.

It's a small loss IMO (assuming few implementers, and a high threshold) compared with exposing 2 public methods doing the same thing (i.e. `Clone`, `DeepClone`).

eerhardt:
The thing I like about the current proposal is that there is only 1 public method name I see in intellisense - `Clone`.  I don't see 2 methods on `Tensor<T>` when I am using it from outside of the class.  The extension method approach would require 2 different methods, right?  The extension `Clone` and the `DeepClone` method that is called by the extension.

FWIW - I'd pick a different name than `CloneTensor` or `DeepClone` for the abstract method.  `Deep` is wrong IMO because it doesn't actually clone the individual elements (I can create a `Tensor<object>`).  `CloneTensor` is just confusing from the normal `Clone` method.  `CloneInternal` does sound better to me, or maybe `CloneProtected`.  `CloneImplementation`?

ericstj:
Yeah, I can see how this is better for the users at a small cost to the derivers.  Given there will be a relatively small set of derivers I'm ok with forcing them to "read the manual" on what to do for Clone.  We should also add a test that calls the base clone and ensures that it returns the derived type.

danmosemsft:
@tannergooding you'll need to figure out what to do with this PR at some point.

danmosemsft:
@tannergooding should this be closed or moved to corefx

tannergooding:
@i3arnon, could you please move this PR to target the relocated code in CoreFX?

Issue #2315, opened 2018-05-30
Optimize JsonWriterUtf8 and JsonWriterUtf16 WriteNull() methods


ahsonkhan:
Follow the example from #2291 to reduce unnecessary calls to EnsureBuffer and Advance.

https://github.com/dotnet/corefxlab/blob/cae77712b8274a686fee4a4b62db770e093a701b/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs#L568
https://github.com/dotnet/corefxlab/blob/cae77712b8274a686fee4a4b62db770e093a701b/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs#L546

PR #2329, opened 2018-06-02
Optimize JsonWriterUtf8 and JsonWriterUtf16 WriteNull() methods (#2315)

cshung:
https://github.com/dotnet/corefxlab/issues/2315

### DIFF
--- a/src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs
@@ -37,6 +37,8 @@ namespace System.Text.JsonLab
         public static readonly byte[] FalseValue = { (byte)'f', (byte)'a', (byte)'l', (byte)'s', (byte)'e' };
         public static readonly byte[] NullValue = { (byte)'n', (byte)'u', (byte)'l', (byte)'l' };
 
+        public static readonly byte[] NullValueUtf16 = { (byte)'n', 0, (byte)'u', 0, (byte)'l', 0, (byte)'l', 0 };
+
         #endregion Common values
     }
 }
--- a/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
@@ -543,12 +543,37 @@ namespace System.Text.JsonLab
         /// </summary>
         public void WriteNull()
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf16();
-            _firstItem = false;
-            WriteSpacingUtf16();
+            int charsNeeded = (_firstItem ? 0 : 1) + (_prettyPrint ? 2 + (_indent + 1) * 2 : 0);
+            int bytesNeeded = charsNeeded * 2 + JsonConstants.NullValueUtf16.Length;
+            Span<byte> byteBuffer = EnsureBuffer(bytesNeeded);
+            Span<char> charBuffer = MemoryMarshal.Cast<byte, char>(byteBuffer);
+            int idx = 0;
+            if (_firstItem)
+            {
+                _firstItem = false;
+            }
+            else 
+            {
+                charBuffer[idx++] = (char)JsonConstants.ListSeperator;
+            }
 
-            WriteJsonValue(JsonConstants.NullValue);
+            if (_prettyPrint) 
+            {
+                int indent = _indent;
+                charBuffer[idx++] = (char)JsonConstants.CarriageReturn;
+                charBuffer[idx++] = (char)JsonConstants.LineFeed;
+
+                while (indent-- >= 0)
+                {
+                    charBuffer[idx++] = (char)JsonConstants.Space;
+                    charBuffer[idx++] = (char)JsonConstants.Space;
+                }
+            }
+
+            Debug.Assert(byteBuffer.Slice(idx * 2).Length >= JsonConstants.NullValueUtf16.Length);
+            JsonConstants.NullValueUtf16.CopyTo(byteBuffer.Slice(idx * 2));
+
+            _bufferWriter.Advance(bytesNeeded);
         }
 
         private void WriteAttributeUtf16(ReadOnlySpan<char> nameSpanChar, int bytesNeeded)
--- a/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
@@ -565,12 +565,35 @@ namespace System.Text.JsonLab
         /// </summary>
         public void WriteNull()
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf8();
-            _firstItem = false;
-            WriteSpacingUtf8();
+            int bytesNeeded = (_firstItem ? 0 : 1) + (_prettyPrint ? 2 + (_indent + 1) * 2 : 0) + JsonConstants.NullValue.Length;
+            Span<byte> byteBuffer = EnsureBuffer(bytesNeeded);
+            int idx = 0;
+            if (_firstItem)
+            {
+                _firstItem = false;
+            }
+            else 
+            {
+                byteBuffer[idx++] = JsonConstants.ListSeperator;
+            }
 
-            WriteJsonValue(JsonConstants.NullValue);
+            if (_prettyPrint) 
+            {
+                int indent = _indent;
+                byteBuffer[idx++] = JsonConstants.CarriageReturn;
+                byteBuffer[idx++] = JsonConstants.LineFeed;
+
+                while (indent-- >= 0)
+                {
+                    byteBuffer[idx++] = JsonConstants.Space;
+                    byteBuffer[idx++] = JsonConstants.Space;
+                }
+            }
+
+            Debug.Assert(byteBuffer.Slice(idx).Length >= JsonConstants.NullValue.Length);
+            JsonConstants.NullValue.CopyTo(byteBuffer.Slice(idx));
+
+            _bufferWriter.Advance(bytesNeeded);
         }
 
         private void WriteAttributeUtf8(ReadOnlySpan<byte> nameSpanByte, int bytesNeeded)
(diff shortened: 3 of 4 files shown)


ahsonkhan:
Thanks for the contribution. Can you also add a call to WriteNull within the test as well?
https://github.com/dotnet/corefxlab/blob/master/tests/System.Text.JsonLab.Tests/JsonWriterTests.cs#L188
Update the GetExpectedString and the test by adding a null phoneNumber value (within the array).

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -565,12 +565,9 @@ public void WriteValue(Guid value)
...
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf8();
-            _firstItem = false;
-            WriteSpacingUtf8();
-
-            WriteJsonValue(JsonConstants.NullValue);
+            ReadOnlySpan<byte> valueSpan = JsonConstants.NullValue;
+            int bytesNeeded = CalculateValueBytesNeeded(valueSpan, sizeof(byte), 2);
+            WriteValueUtf8(valueSpan, bytesNeeded);
```
ahsonkhan: I don't think calling the existing WriteValueUtf8 is correct here. This method writes the value surrounded by quotations. We only want the characters `n u l l` to be written. You should add your own method, like WriteNullUtf8 for now (or re-write WriteJsonValue).

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -565,12 +565,9 @@ public void WriteValue(Guid value)
...
         public void WriteNull()
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf8();
-            _firstItem = false;
-            WriteSpacingUtf8();
-
-            WriteJsonValue(JsonConstants.NullValue);
+            ReadOnlySpan<byte> valueSpan = JsonConstants.NullValue;
+            int bytesNeeded = CalculateValueBytesNeeded(valueSpan, sizeof(byte), 2);
```
ahsonkhan: Similarly, here, CalculateValueBytesNeeded is adding the bytes needed to add the quotes, which shouldn't be needed here.

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -565,10 +565,32 @@ public void WriteValue(Guid value)
...
+                while (indent-- >= 0)
+                {
+                    byteBuffer[idx++] = JsonConstants.Space;
+                    byteBuffer[idx++] = JsonConstants.Space;
+                }
+            }
+
+            _bufferWriter.Advance(bytesNeeded);
 
             WriteJsonValue(JsonConstants.NullValue);
```
ahsonkhan: I think we can remove the WriteJsonValue call as well (for JsonWriterUtf8).
The WriteJsonValue just does a TryCopyTo call and we already know the length of JsonConstants.NullValue:
https://github.com/dotnet/corefxlab/blob/3b9952e99cbd50c9018e7976b31e08820b821a57/src/System.Text.Primitives/System/Text/SymbolTable/SymbolTable_utf8.cs#L51-L63

So, to avoid unnecessary calls to Advance, let's change this to something like:
```C#
JsonConstants.NullValue.CopyTo(byteBuffer.Slice(idx));
```

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -565,10 +565,32 @@ public void WriteValue(Guid value)
...
+                byteBuffer[idx++] = JsonConstants.LineFeed;
+
+                while (indent-- >= 0)
+                {
+                    byteBuffer[idx++] = JsonConstants.Space;
+                    byteBuffer[idx++] = JsonConstants.Space;
+                }
+            }
+
+            _bufferWriter.Advance(bytesNeeded);
```
ahsonkhan: With the change to remove the call to WriteJsonValue, move the call to Advance after it.

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -565,10 +565,32 @@ public void WriteValue(Guid value)
         /// </summary>
         public void WriteNull()
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf8();
-            _firstItem = false;
-            WriteSpacingUtf8();
+            int bytesNeeded = (_firstItem ? 0 : 1) + (_prettyPrint ? 2 + (_indent + 1) * 2 : 0);
```
ahsonkhan: Add JsonConstants.NullValue.Length to bytesNeeded here.

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
```diff
@@ -543,11 +543,34 @@ public void WriteValue(Guid value)
...
+                while (indent-- >= 0)
+                {
+                    charBuffer[idx++] = (char)JsonConstants.Space;
+                    charBuffer[idx++] = (char)JsonConstants.Space;
+                }
+            }
+
+            _bufferWriter.Advance(bytesNeeded);
+            Debug.Assert(idx == charsNeeded);
             WriteJsonValue(JsonConstants.NullValue);
```
ahsonkhan: I think we should do something similar here as Utf8 (and use copyto).

Given the value we are writing is a constant, it doesn't make sense to keep calling TryEncode on it. We should create a UTF-16 encoded NullValue as a byte array and write that to the output buffer
https://github.com/dotnet/corefxlab/blob/3b9952e99cbd50c9018e7976b31e08820b821a57/src/System.Text.Primitives/System/Text/SymbolTable/SymbolTable_utf16.cs#L54-L64

```C#
public static readonly byte[] NullValueUtf16 = { (byte)'n', 0, (byte)'u', 0, (byte)'l', 0, (byte)'l', 0 };
```

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
```diff
@@ -543,12 +543,37 @@ public void WriteValue(Guid value)
...
         public void WriteNull()
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf16();
-            _firstItem = false;
-            WriteSpacingUtf16();
+            int charsNeeded = (_firstItem ? 0 : 1) + (_prettyPrint ? 2 + (_indent + 1) * 2 : 0);
+            int bytesNeeded = charsNeeded * 2 + JsonConstants.NullValueUtf16.Length;
+            Span<byte> byteBuffer = EnsureBuffer(bytesNeeded);
+            Span<char> charBuffer = MemoryMarshal.Cast<byte, char>(byteBuffer );
```
ahsonkhan: nit: remove whitespace `MemoryMarshal.Cast<byte, char>(byteBuffer )`

ahsonkhan:
cc @KrzysztofCwalina

src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs
```diff
@@ -37,6 +37,8 @@ static class JsonConstants
         public static readonly byte[] FalseValue = { (byte)'f', (byte)'a', (byte)'l', (byte)'s', (byte)'e' };
         public static readonly byte[] NullValue = { (byte)'n', (byte)'u', (byte)'l', (byte)'l' };
 
+        public static readonly byte[] NullValueUtf16 = { (byte)'n', 0, (byte)'u', 0, (byte)'l', 0, (byte)'l', 0 };
```
GrabYourPitchforks: This should be NullValueUtf16LE.

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
```diff
@@ -543,12 +543,37 @@ public void WriteValue(Guid value)
...
         public void WriteNull()
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf16();
-            _firstItem = false;
-            WriteSpacingUtf16();
+            int charsNeeded = (_firstItem ? 0 : 1) + (_prettyPrint ? 2 + (_indent + 1) * 2 : 0);
+            int bytesNeeded = charsNeeded * 2 + JsonConstants.NullValueUtf16.Length;
+            Span<byte> byteBuffer = EnsureBuffer(bytesNeeded);
+            Span<char> charBuffer = MemoryMarshal.Cast<byte, char>(byteBuffer);
```
GrabYourPitchforks: Is EnsureBuffer _guaranteed_ to return a span that is 16-bit aligned? This cast has undefined behavior if the input span isn't properly aligned for the destination type.

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
```diff
@@ -543,12 +543,37 @@ public void WriteValue(Guid value)
         /// </summary>
         public void WriteNull()
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf16();
-            _firstItem = false;
-            WriteSpacingUtf16();
+            int charsNeeded = (_firstItem ? 0 : 1) + (_prettyPrint ? 2 + (_indent + 1) * 2 : 0);
```
GrabYourPitchforks: The calculation of `charsNeeded` and `bytesNeeded` should be in a checked block without some other mechanism guaranteeing that they can't integer overflow.

GrabYourPitchforks:
I opened https://github.com/dotnet/corefxlab/issues/2337 to track addressing the PR feedback I left. These feedback items are all high-priority.

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
ahsonkhan: Do you think that an _indent value of 2^30 is realistic for writing some json payload? Only then would a calculation like this overflow.

cshung: It appear to me that none of the calculate buffer size logic are checked - is that really an issue?

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
cshung: EnsureBuffer does not ensure the span is 16-bit aligned, but in this particular use case, it should, because we are manipulating a character stream here. Is there a way I can write an assert to ensure it is aligned?

Issue #1450, opened 2017-04-07
Eliminate ReadHex and use Parsers


ahsonkhan:
Used here:
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs#L252

KrzysztofCwalina:
I think this is done/outdated.

ahsonkhan:
We are still using ReadHex instead of our new Parsers here:
https://github.com/dotnet/corefxlab/blob/master/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs#L302


```C#
        private static int UnescapePercentEncoding(ref int scan, Span<byte> buffer)
        {
            if (buffer[scan++] != '%')
            {
                return -1;
            }

            var probe = scan;

            var value1 = ReadHex(ref probe, buffer);
            if (value1 == -1)
            {
                return -1;
            }
```

cc @GrabYourPitchforks

i3arnon:
@karelz, I would like to take this issue.

PR #2330, opened 2018-06-02
Use Utf8Parser to unescape percent encoding

i3arnon:
Addresses #1450

### DIFF
--- a/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
+++ b/src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
@@ -297,79 +297,30 @@ namespace System.Text.Encodings.Web.Utf8
                 return -1;
             }
 
-            var probe = scan;
-
-            var value1 = ReadHex(ref probe, buffer);
-            if (value1 == -1)
+            buffer = buffer.Slice(scan);
+            if (buffer.Length < 2)
             {
                 return -1;
             }
 
-            var value2 = ReadHex(ref probe, buffer);
-            if (value2 == -1)
+            if (!Utf8Parser.TryParse(buffer.Slice(0, 2), out byte value, out var bytesConsumed, 'X'))
+            {
+                return -1;
+            }
+            
+            if (bytesConsumed != 2)
             {
                 return -1;
             }
 
-            if (SkipUnescape(value1, value2))
-            {
-                return -1;
-            }
-
-            scan = probe;
-            return (value1 << 4) + value2;
-        }
-
-
-        /// <summary>
-        /// Read the next char and convert it into hexadecimal value.
-        /// 
-        /// The <paramref name="scan"/> index will be moved to the next
-        /// byte no matter no matter whether the operation successes.
-        /// </summary>
-        /// <param name="scan">The index of the byte in the buffer to read</param>
-        /// <param name="buffer">The byte span from which the hex to be read</param>
-        /// <returns>The hexadecimal value if successes, otherwise -1.</returns>
-        private static int ReadHex(ref int scan, Span<byte> buffer)
-        {
-            if (scan == buffer.Length)
-            {
-                return -1;
-            }
-
-            var value = buffer[scan++];
-            var isHead = ((value >= '0') && (value <= '9')) ||
-                         ((value >= 'A') && (value <= 'F')) ||
-                         ((value >= 'a') && (value <= 'f'));
-
-            if (!isHead)
-            {
-                return -1;
-            }
-
-            if (value <= '9')
-            {
-                return value - '0';
-            }
-            else if (value <= 'F')
-            {
-                return (value - 'A') + 10;
-            }
-            else // a - f
-            {
-                return (value - 'a') + 10;
-            }
-        }
-
-        private static bool SkipUnescape(int value1, int value2)
-        {
             // skip %2F - '/'
-            if (value1 == 2 && value2 == 15)
+            if (value == 0x2F)
             {
-                return true;
+                return -1;
             }
 
-            return false;
+            scan += 2;
+            return value;
         }
 
         static UrlEncoder()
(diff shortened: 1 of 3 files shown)


src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
```diff
@@ -297,79 +297,30 @@ private static int UnescapePercentEncoding(ref int scan, Span<byte> buffer)
...
-            var value2 = ReadHex(ref probe, buffer);
-            if (value2 == -1)
+            if (!Utf8Parser.TryParse(buffer.Slice(0, 2), out byte value, out var bytesConsumed, 'X'))
             {
                 return -1;
             }
-
-            if (SkipUnescape(value1, value2))
+            
+            if (bytesConsumed != 2)
```
ahsonkhan: When will `bytesConsumed != 2` here if TryParse returned true?

ahsonkhan:
@i3arnon, would you be willing to write a performance test for this method and share the results of before/after? See the unit tests for more info on how to call these APIs: https://github.com/dotnet/corefxlab/blob/8e67e5fd0ed7f1863ee71814ca699f9970478e59/tests/System.Text.Encodings.Web.Utf8.Tests/UnescapeUrlPathInPlaceTests.cs

I want to make sure that as part of this, we are validating that our parser APIs have good performance.

src/System.Text.Encodings.Web.Utf8/UrlEncoder.cs
i3arnon: It could be a single digit hex (i.e. "%FZ")

ahsonkhan: Ah yes, good point.

GrabYourPitchforks: Is there a unit test for the "%FZ" condition? If not, we should add one. Overall LGTM.

i3arnon: @GrabYourPitchforks there wasn't. I added one.

i3arnon:
@ahsonkhan I've added benchmarks. If they're good enough then there's a slight degradation.

Before:

``` ini

Intel Core i7-7600U CPU 2.80GHz (Kaby Lake), 1 CPU, 4 logical and 2 physical cores
Frequency=2835936 Hz, Resolution=352.6173 ns, Timer=TSC
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT


```
|        Method |     Mean |     Error |    StdDev |
|-------------- |---------:|----------:|----------:|
| DecodeInPlace | 35.67 us | 0.5011 us | 0.4687 us |

After:

``` ini

Intel Core i7-7600U CPU 2.80GHz (Kaby Lake), 1 CPU, 4 logical and 2 physical cores
Frequency=2835936 Hz, Resolution=352.6173 ns, Timer=TSC
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT


```
|        Method |     Mean |     Error |    StdDev |
|-------------- |---------:|----------:|----------:|
| DecodeInPlace | 36.71 us | 0.6127 us | 0.5732 us |

tests/Benchmarks/System.Text.Encodings.Web.Utf8/UrlEncoderPerf.cs
```diff
@@ -0,0 +1,39 @@
...
+        private byte[] _encodedBytes;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            var random = new Random(42);
+            var builder = new StringBuilder();
+            for (int i = 0; i < 10000; i++)
+            {
+                if (random.NextDouble() < 0.15)
```
ahsonkhan: Why did you decide on 0.15 here? Out of curiosity, can you build the string to be all hex characters (locally) and share the results? I want to try and exaggerate the use of ReadHex (before) vs Utf8Parser.TryParse (after) to see if the difference grows.

Otherwise, this change looks good to me. Thanks!

cc @adamsitnik

tests/Benchmarks/System.Text.Encodings.Web.Utf8/UrlEncoderPerf.cs
```diff
@@ -0,0 +1,39 @@
...
+    public class UrlEncoderPerf
+    {
+        private byte[] _encodedBytes;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            var random = new Random(42);
+            var builder = new StringBuilder();
+            for (int i = 0; i < 10000; i++)
```
ahsonkhan: nit: change 10000 to 10_000 to make it easier to read.

tests/Benchmarks/System.Text.Encodings.Web.Utf8/UrlEncoderPerf.cs
i3arnon: I wanted something close to real life data so I picked some ratio that looks reasonable. I can change it if you have some better guesstimate.

The results for all hex chars:

Before:

|        Method |     Mean |    Error |   StdDev |   Median |
|-------------- |---------:|---------:|---------:|---------:|
| DecodeInPlace | 262.7 us | 12.84 us | 37.66 us | 245.6 us |

After:

|        Method |     Mean |    Error |   StdDev |
|-------------- |---------:|---------:|---------:|
| DecodeInPlace | 234.7 us | 4.680 us | 6.085 us |

:)
I'm not sure how consistent these tests are though.

adamsitnik: It's ok to use Random for input data, if the seed and length are constant. (so we always have the same input and compare apples to apples between the runs)

Issue #2321, opened 2018-05-31
Cleanup Json PerfSmokeTests and port to BDN


ahsonkhan:
- Remove flaky aspects of the tests (that uses GC APis to measure allocations).
- Port the performance measurement parts to use BDN.

https://github.com/dotnet/corefxlab/blob/89ea88c64f2dabe362c0a3df21d07313542c1d80/tests/System.Text.JsonLab.Tests/PerfSmokeTests.cs#L18

Issue #561, opened 2016-01-14
Ability to regression test GC allocations


KrzysztofCwalina:

ahsonkhan:
I am assuming this is part of performance testing? Does using BenchmarkDotNet to track allocations resolve this issue?

cc @adamsitnik

KrzysztofCwalina:
GC allocations are very predictable. I worry that BDN's harness is just way too heavyweight for it. I was imagining something very lightweight that can be used within a simple unit test, i,.e. capture allocations before the test, run test, ensure that only expected number of bytes are allocated after the test runs. 

I think this can be done with the new thread allocation metrics exposed in core 2.1?

PR #2331, opened 2018-06-02
Use GetAllocatedBytesForCurrentThread to get stable allocation measurements

ahsonkhan:
Fixes https://github.com/dotnet/corefxlab/issues/2321

Removing PerfSmokeTests.cs since we have such benchmarks already. https://github.com/dotnet/corefxlab/blob/3b9952e99cbd50c9018e7976b31e08820b821a57/tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs#L36


ahsonkhan:
Relates to https://github.com/dotnet/corefxlab/issues/561#issuecomment-393610173.
> I think this can be done with the new thread allocation metrics exposed in core 2.1?

That is what I ended up using here.

cc @KrzysztofCwalina, @adamsitnik

adamsitnik:
@ahsonkhan LGTM!

PR #2332, opened 2018-06-03
[WIP] Optimize JsonWriterUtf8 WriteValue overloads

cshung:
Does this pattern sounds like a way to go?

### DIFF
--- a/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
@@ -507,15 +507,23 @@ namespace System.Text.JsonLab
         /// <param name="value">The boolean value to be written to JSON data.</param>
         public void WriteValue(bool value)
         {
-            //TODO: Optimize, just like WriteValue(long value)
-            WriteItemSeperatorUtf8();
-            _firstItem = false;
-            WriteSpacingUtf8();
+            ReadOnlySpan<byte> literal = value ? JsonConstants.TrueValue : JsonConstants.FalseValue;
 
-            if (value)
-                WriteJsonValue(JsonConstants.TrueValue);
-            else
-                WriteJsonValue(JsonConstants.FalseValue);
+            int seperatorSize = ItemSeperatorSize();
+            int spacingSize = SpacingSize();
+            int literalSize = LiteralSize(literal);
+
+            int bytesNeeded = seperatorSize + spacingSize + literalSize;
+
+            Span<byte> byteBuffer = EnsureBuffer(bytesNeeded);
+
+            int idx = 0;
+            idx += this.WriteItemSeperator(byteBuffer.Slice(idx, seperatorSize));
+            idx += this.WriteSpacing(byteBuffer.Slice(idx, spacingSize));
+            idx += this.WriteLiteral(byteBuffer.Slice(idx, literalSize), literal);
+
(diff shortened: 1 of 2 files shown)


ahsonkhan:
> Does this pattern sounds like a way to go?

I am not sure. I think we should compare the performance of refactoring the code this way. 

Update this performance benchmark to write bool values instead of long values and see how performance is affected:
https://github.com/dotnet/corefxlab/blob/da50fe338613fc061d5652b2cda651ebb770cd9c/tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs#L125

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -507,15 +507,23 @@ public void WriteValue(ulong value)
...
-            WriteSpacingUtf8();
+            ReadOnlySpan<byte> literal = value ? JsonConstants.TrueValue : JsonConstants.FalseValue;
 
-            if (value)
-                WriteJsonValue(JsonConstants.TrueValue);
-            else
-                WriteJsonValue(JsonConstants.FalseValue);
+            int seperatorSize = ItemSeperatorSize();
+            int spacingSize = SpacingSize();
+            int literalSize = LiteralSize(literal);
```
ahsonkhan: This seems unnecessary. We should just use literal.Length where its needed.

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -507,15 +507,23 @@ public void WriteValue(ulong value)
...
+
+            int bytesNeeded = seperatorSize + spacingSize + literalSize;
+
+            Span<byte> byteBuffer = EnsureBuffer(bytesNeeded);
+
+            int idx = 0;
+            idx += this.WriteItemSeperator(byteBuffer.Slice(idx, seperatorSize));
```
ahsonkhan: Can we do all these in a single method to avoid the multiple span slices and additions to idx?

src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf8.cs
```diff
@@ -915,5 +909,66 @@ private int AddNewLineAndIndentation(Span<byte> buffer)
...
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int SpacingSize()
+        {
+            return (_prettyPrint ? 2 + (_indent + 1) * 2 : 0);
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private int WriteSpacing(Span<byte> byteBuffer)
```
ahsonkhan: Wouldn't the return values of these methods always equal what the "{X}Size()" methods?

WriteSpacing would result whatever SpacingSize does. So, why do we need to return an integer here?

cshung:
Thanks for the review and the pointer to measuring performance. I am experimenting with performance now.

ahsonkhan:
> WriterSystemTextJsonBasic |          True |     False | 17,008.23 ns |
WriterNewtonsoftBasic |          True |     False | 14,822.00 ns |

What changes did you make to get those results? It seems to be an out-lier. All other UTF-8 benchmarks show SystemTextJson is faster. We should try to optimize the implementation further for this case as well.

PR #2334, opened 2018-06-03
Migrate UTF8 decoding benchmarks to Benchmark.DotNet

### LINKED ISSUE #2319: Port Utf8Utility Performance test to BenchmarkDotNet
We have some Utf8Utility performance tests that need to be ported from xunit.performance to BenchmarkDotNet.
https://github.com/dotnet/corefxlab/blob/89ea88c64f2dabe362c0a3df21d07313542c1d80/tests/System.Text.Primitives.Tests/Encoding/Performance/Utf8UtilityPerfTests.cs#L11

Follow the examples from previous efforts to port the performance tests:
https://github.com/dotnet/corefxlab/pull/2240
https://github.com/dotnet/corefxlab/pull/2241
https://github.com/dotnet/corefxlab/pull/2242
https://github.com/dotnet/corefxlab/pull/2243
https://github.com/dotnet/corefxlab/pull/2249
https://github.com/dotnet/corefxlab/pull/2254
https://github.com/dotnet/corefxlab/pull/2260

cc @adamsitnik, @GrabYourPitchforks

kant2002:
Closes #2319


tests/Benchmarks/System.Text.Primitives/GetUtf16CharCount.cs
```diff
@@ -0,0 +1,64 @@
...
+using BenchmarkDotNet.Attributes;
+using System.Buffers.Text;
+using System.Collections.Generic;
+using System.IO;
+
+namespace System.Text.Primitives.Benchmarks
+{
+    public class GetUtf16CharCount
+    {
+        private const int InnerCount = 10000;
```
adamsitnik: dead code: `InnerCount` is not used anymore

tests/Benchmarks/System.Text.Primitives/GetUtf16CharCount.cs
```diff
@@ -0,0 +1,64 @@
...
+        private static readonly byte[] _utf8BOM = new byte[] { 0xEF, 0xBB, 0xBF };
+
+        [ParamsSource(nameof(FileNames))]
+        public string FileName;
+
+        static byte[] utf8Text;
+
+        [GlobalSetup]
+        public void Setup()
+        {
```
adamsitnik: nit: you can use expressions for such simple methods

`public void Setup() => utf8Text = ReadTestResource(FileName).ToArray();`

kant2002:
@adamsitnik I fix PR comments

ahsonkhan:
Can you please share the results of the benchmark?

GrabYourPitchforks:
FWIW we should try to avoid code like the following going forward.

```cs
public void Setup() => utf8Text = ReadTestResource(FileName).ToArray();
```

The lambda operator (`=>`) usually implies that the member is returning a value or that the member is a pass-through to another method that performs the actual computation. We generally _don't_ have side effect-causing code immediately on the right of the operator, such as the field setter code which is there now. Such code harms readability over the more imperative form.

```cs
public void Setup() {
    utf8Text = ReadTestResource(FileName).ToArray();
}
```

kant2002:
@ahsonkhan @adamsitnik @GrabYourPitchforks  Should I add separate PR with additional comments? What are the consensus regarding method body expressions ?

adamsitnik:
> The lambda operator (=>) usually implies that the member is returning a value

adamsitnik:
> The lambda operator (=>) usually implies that the member is returning a value

I think that it's personal preference, to me the less the code the better.

ahsonkhan:
> Should I add separate PR with additional comments?

Up to you. I don't have a preference either way, but I did notice that if we tend to change the performance test or update it to be multi-line, then it is more work if we keep it as a method body expression. If it is a regular method, it is trivial to expand. Once written, benchmarks shouldn't change too much (especially if they are used to track performance regression). In this case, however, there might be changes given the API surface and design is not set in stone yet. So, I would lean towards using a regular method.

Please share the performance results, if you have them :)

Issue #2312, opened 2018-05-30
Add a performance benchmark for Range GetEnumerator


ahsonkhan:
We use BenchmarkDotNet for our performance benchmarks. See https://github.com/dotnet/corefxlab/tree/master/tests/Benchmarks for examples.

Add a performance test that measures the cost of iterating through a range of values.
https://github.com/dotnet/corefxlab/blob/d68a9a4cfae4c62a0cb05c204cf8d98d6876833a/src/System.Buffers.Experimental/System/Range.cs#L81

cc @KrzysztofCwalina

i3arnon:
I'll take it.

kant2002:
This closed by #2336

karelz:
Thanks for reminder! Closing.

PR #2336, opened 2018-06-03
Add a benchmark for Range enumeration

i3arnon:
Addresses #2312 

Not sure there's much more to do here.

The results:

``` ini

Intel Core i7-7600U CPU 2.80GHz (Kaby Lake), 1 CPU, 4 logical and 2 physical cores
Frequency=2835936 Hz, Resolution=352.6173 ns, Timer=TSC
.NET Core SDK=2.1.300
  [Host]     : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0 (CoreCLR 4.6.26515.07, CoreFX 4.6.26515.06), 64bit RyuJIT


```
|      Method | Length |         Mean |       Error |      StdDev |
|------------ |------- |-------------:|------------:|------------:|
| **Enumeration** |     **10** |     **21.23 ns** |   **0.0559 ns** |   **0.0496 ns** |
| **Enumeration** |    **100** |    **199.84 ns** |   **0.8654 ns** |   **0.7226 ns** |
| **Enumeration** |   **1000** |  **1,849.24 ns** |   **2.0702 ns** |   **1.8352 ns** |
| **Enumeration** |  **10000** | **18,728.23 ns** | **258.6625 ns** | **215.9949 ns** |

### DIFF
--- a/tests/Benchmarks/System/RangePerf.cs
+++ b/tests/Benchmarks/System/RangePerf.cs
@@ -1,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+
+namespace System.Benchmarks
+{
+    public class RangePerf
+    {
+        [Params(10, 100, 1000, 10_000)]
+        public uint Length;
+
+        private Range _range;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            _range = new Range(0, Length);
+        }
+
+        [Benchmark]
+        public void Enumeration()
+        {
+            foreach (var value in _range)
+            {
+                
+            }
+        }
+    }
+}
--- a/src/System.Buffers.Experimental/System/Range.cs
+++ b/src/System.Buffers.Experimental/System/Range.cs
@@ -78,7 +78,6 @@ namespace System
             void IEnumerator.Reset() => throw new NotSupportedException();
         }
 
-        // TODO: write benchmark for this
         public Enumerator GetEnumerator()
         {
             if(IsBound) return new Enumerator(First, End);
--- a/tests/Benchmarks/System/RangeEnumeration.cs
+++ b/tests/Benchmarks/System/RangeEnumeration.cs
@@ -4,9 +4,9 @@
 
 using BenchmarkDotNet.Attributes;
 
-namespace System.Benchmarks
+namespace System.Buffers.Experimental.Benchmarks
 {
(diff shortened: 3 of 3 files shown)


tests/Benchmarks/System/RangePerf.cs
```diff
@@ -0,0 +1,31 @@
...
+        [GlobalSetup]
+        public void Setup()
+        {
+            _range = new Range(0, Length);
+        }
+
+        [Benchmark]
+        public void Enumeration()
+        {
+            foreach (var value in _range)
```
ahsonkhan: nit: don't use var here

ahsonkhan:
> Not sure there's much more to do here.

@KrzysztofCwalina, do you have some baseline that we should be comparing this benchmark to?

tests/Benchmarks/System/RangePerf.cs
```diff
@@ -0,0 +1,31 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+
+namespace System.Benchmarks
```
ahsonkhan: Please follow the namespace and directory structure of other benchmarks.
For instance see this:
https://github.com/dotnet/corefxlab/blob/67620c5a10103a72c5125cc85ac123e56943ed60/tests/Benchmarks/System.Text.Primitives/AsciiDecoding.cs#L8

This should be within the System.Buffers.Experiemental.Benchmarks namespace and be placed within the tests/Benchmarks/System.Buffers.Experiemental folder.

Also consider renaming the file to exclude words like Perf/Benchmark/etc as they are redundant. RangeEnumeration works.

i3arnon: > Please follow the namespace

I thought I did, but went with the type's namespace (i.e. `System`) instead.

> Also consider renaming the file to exclude words like Perf/Benchmark/etc as they are redundant. RangeEnumeration works.

I though that's the convention after seeing `Utf8StringPerf`, `JsonReaderPerf`, `JsonWriterPerf`. etc.

I fixed both issues.

ahsonkhan: > I though that's the convention after seeing `Utf8StringPerf`, `JsonReaderPerf`, `JsonWriterPerf`. etc.

Great point. It looks like we don't have a consistent rule on this. Thanks for fixing it.

joshfree:
Thank you for your contribution @i3arnon

PR #2338, opened 2018-06-04
Address PR feedback from #2329

cshung:
@GrabYourPitchforks
@ahsonkhan

### DIFF
--- a/src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs
@@ -37,7 +37,8 @@ namespace System.Text.JsonLab
         public static readonly byte[] FalseValue = { (byte)'f', (byte)'a', (byte)'l', (byte)'s', (byte)'e' };
         public static readonly byte[] NullValue = { (byte)'n', (byte)'u', (byte)'l', (byte)'l' };
 
-        public static readonly byte[] NullValueUtf16 = { (byte)'n', 0, (byte)'u', 0, (byte)'l', 0, (byte)'l', 0 };
+        public static readonly byte[] NullValueUtf16LE = { (byte)'n', 0, (byte)'u', 0, (byte)'l', 0, (byte)'l', 0 };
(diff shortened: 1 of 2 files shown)


src/System.Text.JsonLab/System/Text/Json/JsonWriterUtf16.cs
```diff
@@ -570,10 +570,17 @@ public void WriteNull()
...
-            _bufferWriter.Advance(bytesNeeded);
+            if (BitConverter.IsLittleEndian)
+            {
+                Debug.Assert(byteBuffer.Slice(idx * 2).Length >= JsonConstants.NullValueUtf16LE.Length);
+                JsonConstants.NullValueUtf16LE.CopyTo(byteBuffer.Slice(idx * 2));
+                _bufferWriter.Advance(bytesNeeded);
+            }
+            else
+            {
+                _bufferWriter.Advance(bytesNeeded);
```
ahsonkhan: Add a NullValueUtf16BE and use that instead?

Alternatively, try to isolate and measure the difference in performance between calling the UTF-16 to UTF-8 encode method for a small input (like 4-byte null array) and compare that to directly copying a pre-computed byte array. If there is little performance impact, we could keep the encoding step, which appropriately deals with the endianness.

ahsonkhan:
@dotnet-bot test this please

@GrabYourPitchforks, any feedback, especially on the other points you mentioned previously?
- [ensure buffer](https://github.com/dotnet/corefxlab/pull/2329#discussion_r192606825)
- [integer overflow][(https://github.com/dotnet/corefxlab/pull/2329#discussion_r192606868)

cshung:
@ahsonkhan 
I have merged the conflicts. Shall we check this in so that we fix the issue on big-endian machine? We can keep the issue open to track the alignment and overflow if necessary.

ahsonkhan:
> I have merged the conflicts. Shall we check this in so that we fix the issue on big-endian machine? We can keep the issue open to track the alignment and overflow if necessary.

Yes, sounds good to me.

PR #2339, opened 2018-06-04
Added System.Memory.Polyfill Project

KrzysztofCwalina:
This project is a skeleton for an adapter package that allows writing portable code using ```Span<T>```-enabled APIs that are found in .NET Core 2.1 but not in other versions/redists of .NET.

For example, .NET Core 2.1 Stream has a Read method that takes a ```Span<byte>``` parameter. This member does not exist in other .NET versions, which makes is difficult to write portable code (such portable code has to "manually" cross-target). 

This library hide cross-targeting inside adapter APIs.


src/System.Memory.Polyfill/System/Parse.cs
```diff
@@ -0,0 +1,18 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System
+{
+    public static partial class Int32Polyfill
+    {
+        public static bool TryParse(ReadOnlySpan<char> buffer, out int value)
```
benaadams: Would,'t work for implicit cast from `Span<char>`?

stephentoub: Why not?  It's not an extension method.

src/System.Memory.Polyfill/System.Memory.Polyfill.csproj
```diff
@@ -0,0 +1,18 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\tools\common.props" />
+  <PropertyGroup>
+    <Description>Adapters for Span APIs</Description>
+    <TargetFrameworks>netstandard1.3;netcoreapp2.1</TargetFrameworks>
+    <PackageTags>Span Memory Adapter</PackageTags>
+    <!-- disable automatic compile item inclusion so that we can set the correct metadata on TextTemplate-generated sources -->
```
ericstj: You don't actually have any of these, so you can remove this and the compile entry below.

src/System.Memory.Polyfill/System/Parse.cs
```diff
@@ -0,0 +1,18 @@
...
+        {
+#if NETCOREAPP2_1
+            return int.TryParse(buffer, out value);
```
ericstj: You may want to avoid using the versioned define as it makes it more difficult if you decide to retarget.  Instead use `#if NETCOREAPP`

tests/System.Memory.Polyfill.Tests/System.Memory.Polyfill.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="..\..\tools\common.props" />
+  <PropertyGroup>
+    <TargetFrameworks>net46;netcoreapp2.1</TargetFrameworks>
+    <AssemblyOriginatorKeyFile>../../tools/test_key.snk</AssemblyOriginatorKeyFile>
```
ericstj: All the properties from here on look unnecessary for a test project.

src/System.Memory.Polyfill/System/Parse.cs
KrzysztofCwalina: But this api exists only in netcoreapp 2.1

tests/System.Memory.Polyfill.Tests/System.Memory.Polyfill.Tests.csproj
KrzysztofCwalina: yeah, I just copied them from some other test, but I will remove/simplify. Thanks.

src/System.Memory.Polyfill/System.Memory.Polyfill.csproj
KrzysztofCwalina: will do.

src/System.Memory.Polyfill/System/Parse.cs
ericstj: You aren't defining targeting here, you're just ifdefing the code.  Imagine that you wanted to cross-compile for netcoreapp2.2 as well in the future, you'd need to touch this ifdef for no reason.  I think you're more likely to cross-compile for a future version of netcoreapp than you are an older version that doesn't have this API.
Alternatively you can define your own feature flag that represents what you want and make sure that's defined as appropriate.

src/System.Memory.Polyfill/System.Memory.Polyfill.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="System.Memory" Version="$(SystemMemoryVersion)" />
+    <PackageReference Include="System.Buffers" Version="$(SystemMemoryVersion)" />
```
ahsonkhan: Use SystemBuffersVersion instead of SystemMemoryVersion.

src/System.Memory.Polyfill/System/Stream.cs
```diff
@@ -0,0 +1,32 @@
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers;
+using System.IO;
+
+namespace System
+{
+    public static partial class MemoryPolyfill
```
ahsonkhan: StreamPolyFill instead?

tests/System.Memory.Polyfill.Tests/Int32.cs
```diff
@@ -0,0 +1,20 @@
...
+
+namespace System.Polyfill.Tests
+{
+    public class Int32Tests
+    {
+        [Fact]
+        public void Int32TryParse()
+        {
+            char[] buffer = int.MaxValue.ToString().ToCharArray();
+            ReadOnlySpan<char> span = buffer.AsSpan();
```
ahsonkhan: `ReadOnlySpan<char> span = int.MaxValue.ToString().AsSpan();`?

src/System.Memory.Polyfill/System/Parse.cs
KrzysztofCwalina: I am not sure I understand. Can you clarify?

KrzysztofCwalina:

src/System.Memory.Polyfill/System/Parse.cs
KrzysztofCwalina: Yeah, I agree it would work today, but it won't if we ever add new targets. Correct?

src/System.Memory.Polyfill/System/Stream.cs
KrzysztofCwalina: I think we want to minimize the number of extensions types.

tests/System.Memory.Polyfill.Tests/System.Memory.Polyfill.Tests.csproj
```diff
@@ -1,13 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
     <TargetFrameworks>net46;netcoreapp2.1</TargetFrameworks>
-    <AssemblyOriginatorKeyFile>../../tools/test_key.snk</AssemblyOriginatorKeyFile>
```
ahsonkhan: Why did you remove these properties?

tests/System.Memory.Polyfill.Tests/System.Memory.Polyfill.Tests.csproj
```diff
@@ -1,13 +1,10 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
+    <TargetFrameworks>netcoreapp2.1</TargetFrameworks>
+  </PropertyGroup>
+  <PropertyGroup Condition="'$(TargetsWindows)' == 'true'">
```
ahsonkhan: Is listing netcoreapp2.1 TFM here redundant or required (for Windows only)?

tests/System.Memory.Polyfill.Tests/Int32.cs
```diff
@@ -11,8 +11,7 @@ public class Int32Tests
         [Fact]
         public void Int32TryParse()
         {
-            char[] buffer = int.MaxValue.ToString().ToCharArray();
-            ReadOnlySpan<char> span = buffer.AsSpan();
+            ReadOnlySpan<char> span = int.MaxValue.ToString().ToCharArray().AsSpan();
```
ahsonkhan: Why is the `.ToCharArray()` necessary?

KrzysztofCwalina: fixed in the new PR

tests/System.Memory.Polyfill.Tests/System.Memory.Polyfill.Tests.csproj
KrzysztofCwalina: @ericstj suggested that they were not necessary for tests.

tests/System.Memory.Polyfill.Tests/System.Memory.Polyfill.Tests.csproj
KrzysztofCwalina: I changed it in the new PR. This actually did not work in the CI. I changed it to use netcoreapp2.0 to force the fallback path.

PR #2348, opened 2018-06-06
Merge back JsonWriterUtf8 and JsonWriterUtf16

ahsonkhan:
Reverting the split that was done in https://github.com/dotnet/corefxlab/pull/2292.

There is only a 1-8% regression due to the `if (_isUtf8)` checks.

It isn't enough to justify splitting up the types.

``` ini

BenchmarkDotNet=v0.10.14.534-nightly, OS=Windows 10.0.16299.431 (1709/FallCreatorsUpdate/Redstone3)
Intel Xeon CPU E5-1620 v2 3.70GHz, 1 CPU, 8 logical and 4 physical cores
Frequency=3604597 Hz, Resolution=277.4235 ns, Timer=TSC
.NET Core SDK=2.2.100-preview1-008958
  [Host]     : .NET Core 2.1.0-rtm-26515-03 (CoreCLR 4.6.26515.03, CoreFX 4.6.26515.03), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.0-rtm-26515-03 (CoreCLR 4.6.26515.03, CoreFX 4.6.26515.03), 64bit RyuJIT


```
|                         Method | IsUTF8Encoded | Formatted |         Mean |       Error |      StdDev |  Gen 0 | Allocated |
|------------------------------- |-------------- |---------- |-------------:|------------:|------------:|-------:|----------:|
|      **WriterSystemTextJsonBasic** |         **False** |     **False** | **13,529.82 ns** | **159.0390 ns** | **148.7652 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |     False | 25,062.88 ns |  31.7473 ns |  26.5105 ns | 0.0610 |     472 B |
| WriterSystemTextJsonHelloWorld |         False |     False |     85.79 ns |   0.2155 ns |   0.1911 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |     False |    267.93 ns |   0.9935 ns |   0.7757 ns | 0.0591 |     312 B |
|      **WriterSystemTextJsonBasic** |         **False** |      **True** | **15,564.28 ns** |  **48.1039 ns** |  **40.1689 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |      True | 38,421.90 ns | 734.3675 ns | 686.9277 ns | 0.0610 |     640 B |
[...]

Issue #2177, opened 2018-03-22
API Proposal: BufferWriter<T>


KrzysztofCwalina:
The prototype of this type is in https://github.com/dotnet/corefxlab/tree/master/src/System.Buffers.ReaderWriter/System/Buffers/Writer

```c#
namespace System.Buffers.Writer {
    public ref struct BufferWriter<T> where T : IBufferWriter<byte> {

        public BufferWriter(T output);
        public ReadOnlySpan<byte> NewLine; 

        public Span<byte> Buffer { get; }
        public void Advance(int count);
        public void Ensure(int count=1);
        public void Flush();

        // primitive APIs will be available for:
        // byte, sbyte, ushort, short, uint, int, ulong, long, float, double, char, Utf8Char
        // DateTime, DateTimeOffset, TimeSpan, Guid, Uri, BigInteger, Decimal
        public void Write(int value);
        public void WriteLine(int value);
        public void Write(int value, StandardFormat format=default);
        public void WriteLine(int value, StandardFormat format=default);
        public void Write(int value, TransformationFormat format);
        public void WriteLine(int value, TransformationFormat format);

        // string APIs
        public void Write(string value);
        public void WriteLine(string value);
        public void Write(string value, TransformationFormat format);
        public void WriteLine(string value, TransformationFormat format);

        public void Write(ReadOnlySpan<char> value);
        public void WriteLine(ReadOnlySpan<char>  value);
        public void Write(ReadOnlySpan<char> value, TransformationFormat format);
        public void WriteLine(ReadOnlySpan<char> value, TransformationFormat format);

        public void Write(Utf8String value);
        public void WriteLine(Utf8String value);
        public void Write(Utf8String value, TransformationFormat format);
        public void WriteLine(Utf8String value, TransformationFormat format);

        public void Write(ReadOnlySpan<Utf8Char> value);
        public void WriteLine(ReadOnlySpan<Utf8Char>  value);
        public void Write(ReadOnlySpan<Utf8Char> value, TransformationFormat format);
        public void WriteLine(ReadOnlySpan<Utf8Char> value, TransformationFormat format);

        // writables
        public void Write<TWritable>(TWritable value) where TWritable : IWritable;
        public void WriteLine<TWritable>(TWritable value) where TWritable : IWritable;

        public void Write<TWritable>(TWritable value, StandardFormat format) where TWritable : IWritable;
[...]

PR #2358, opened 2018-06-12
 Add support for BufferWriter<T> to the JsonWriter where T is IBufferWriter<byte>

ahsonkhan:
Since JsonWriter contains a BufferWriter field now, it has to be a ref struct.

- Maybe `_bufferWriter.Ensure` should return back the span. I tend to call Ensure and then access the buffer right after (_bufferWriter.Buffer).
- Also, calling Flush after each call to Advance is a nuisance.

There is a ~20% regression in performance for the HelloWorld case, across the board (higher constant overhead due to the BufferWriter constructor, roughly 19 ns). However, we get a 15-20% performance improvement for writing a larger json payload. I am not sure why exactly that is. Any ideas?

``` ini

BenchmarkDotNet=v0.10.14.534-nightly, OS=Windows 10.0.14393.0 (1607/AnniversaryUpdate/Redstone1)
Frequency=3328125 Hz, Resolution=300.4695 ns, Timer=TSC
.NET Core SDK=2.2.100-preview1-009007
  [Host]     : .NET Core 2.2.0-preview1-26529-01 (CoreCLR 4.6.26525.04, CoreFX 4.6.26525.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.2.0-preview1-26529-01 (CoreCLR 4.6.26525.04, CoreFX 4.6.26525.01), 64bit RyuJIT


```

**BEFORE:**

|                         Method | IsUTF8Encoded | Formatted |         Mean |       Error |      StdDev |       Median |  Gen 0 | Allocated |
|------------------------------- |-------------- |---------- |-------------:|------------:|------------:|-------------:|-------:|----------:|
|      **WriterSystemTextJsonBasic** |         **False** |     **False** | **11,296.64 ns** | **216.3541 ns** | **212.4887 ns** | **11,262.65 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |     False | 21,245.27 ns | 418.2912 ns | 663.4542 ns | 21,252.39 ns | 0.0916 |     472 B |
| WriterSystemTextJsonHelloWorld |         False |     False |     67.73 ns |   1.3718 ns |   1.2832 ns |     67.62 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |     False |    216.21 ns |   2.9374 ns |   2.4529 ns |    217.23 ns | 0.0739 |     312 B |
|      **WriterSystemTextJsonBasic** |         **False** |      **True** | **13,256.87 ns** | **278.9258 ns** | **481.1333 ns** | **13,191.37 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |      True | 29,857.70 ns | 489.6843 ns | 458.0510 ns | 29,798.08 ns | 0.1221 |     632 B |
| WriterSystemTextJsonHelloWorld |         False |      True |     71.91 ns |   1.4367 ns |   2.0605 ns |     71.70 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |      True |    313.56 ns |   6.2629 ns |  17.5618 ns |    308.08 ns | 0.1121 |     472 B |
|      **WriterSystemTextJsonBasic** |          **True** |     **False** | **11,486.53 ns** | **224.4030 ns** | **275.5871 ns** | **11,413.29 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |     False | 18,858.90 ns | 393.5725 ns | 483.3424 ns | 18,687.17 ns | 0.0916 |     416 B |
| WriterSystemTextJsonHelloWorld |          True |     False |    116.67 ns |   0.7400 ns |   0.6560 ns |    116.68 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |     False |    245.37 ns |   2.3336 ns |   1.9487 ns |    245.17 ns | 0.0606 |     256 B |
|      **WriterSystemTextJsonBasic** |          **True** |      **True** | **12,622.99 ns** |  **17.7916 ns** |  **14.8568 ns** | **12,621.65 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |      True | 27,754.24 ns |  63.0735 ns |  52.6692 ns | 27,741.40 ns | 0.1221 |     576 B |
| WriterSystemTextJsonHelloWorld |          True |      True |    112.05 ns |   0.7270 ns |   0.6444 ns |    111.83 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |      True |    343.16 ns |   6.8418 ns |   7.0260 ns |    344.03 ns | 0.0987 |     416 B |

**AFTER:**

|                         Method | IsUTF8Encoded | Formatted |         Mean |      Error |     StdDev |  Gen 0 | Allocated |
|------------------------------- |-------------- |---------- |-------------:|-----------:|-----------:|-------:|----------:|
|      **WriterSystemTextJsonBasic** |         **False** |     **False** |  **9,538.89 ns** | **186.580 ns** | **242.607 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |     False | 19,977.28 ns | 226.011 ns | 200.353 ns | 0.0916 |     472 B |
| WriterSystemTextJsonHelloWorld |         False |     False |     82.10 ns |   1.630 ns |   1.673 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |     False |    242.20 ns |   2.962 ns |   2.312 ns | 0.0739 |     312 B |
|      **WriterSystemTextJsonBasic** |         **False** |      **True** | **11,179.64 ns** | **185.413 ns** | **164.363 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |      True | 28,603.38 ns | 566.520 ns | 847.940 ns | 0.1221 |     632 B |
| WriterSystemTextJsonHelloWorld |         False |      True |     87.42 ns |   1.712 ns |   2.285 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |      True |    316.48 ns |   5.985 ns |   5.598 ns | 0.1121 |     472 B |
|      **WriterSystemTextJsonBasic** |          **True** |     **False** |  **9,215.93 ns** | **176.737 ns** | **165.320 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |     False | 20,158.62 ns | 402.585 ns | 638.543 ns | 0.0916 |     416 B |
| WriterSystemTextJsonHelloWorld |          True |     False |    142.76 ns |   2.516 ns |   2.354 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |     False |    280.29 ns |   4.935 ns |   6.417 ns | 0.0606 |     256 B |
|      **WriterSystemTextJsonBasic** |          **True** |      **True** | **11,498.84 ns** | **229.579 ns** | **298.518 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |      True | 28,570.19 ns | 551.134 ns | 541.287 ns | 0.1221 |     576 B |
| WriterSystemTextJsonHelloWorld |          True |      True |    142.99 ns |   1.430 ns |   1.268 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |      True |    351.14 ns |   7.372 ns |  13.294 ns | 0.0987 |     416 B |


tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
```diff
@@ -100,7 +102,7 @@ private TextWriter GetWriter()
             return writer;
         }
 
-        private static void WriterSystemTextJsonBasicUtf8(bool formatted, ArrayFormatter output, int[] data)
+        private static void WriterSystemTextJsonBasicUtf8(bool formatted, BufferWriter<IBufferWriter<byte>> output, int[] data)
```
benaadams: Again add a convenience helper
```csharp
public ref partial struct BufferWriter
{
    public static BufferWriter<T> Create<T>(TBufferWriter bufferWriter) where TBufferWriter : IBufferWriter<byte>
    {
        return new BufferWriter<TBufferWriter>(bufferWriter);
    }
}
```
Though maybe they should be called `BufferedWriter` and `BufferedWriter<TBufferWriter>`?

```csharp
private static void WriterSystemTextJsonBasicUtf8<TBufferWriter>(bool formatted, TBufferWriter output, int[] data)
    where TBufferWriter : IBufferWriter<byte>
{
    WriterSystemTextJsonBasicUtf8(formatted, BufferWriter.Create(output), data);
}

private static void WriterSystemTextJsonBasicUtf8(bool formatted, BufferWriter<TBufferWriter> output, int[] data)
    where TBufferWriter IBufferWriter<byte>
{
    var  json = JsonWriter.Create(output, isUtf8: true, formatted);
    // ...
```

JeremyKuhne:
> I am not sure why exactly that is. Any ideas?

Did you try measuring with just changing JsonWriter to a ref struct? No specific reason I think that would make a difference, but I am curious.

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
```diff
@@ -129,7 +131,7 @@ private static void WriterSystemTextJsonBasicUtf8(bool formatted, ArrayFormatter
             json.WriteObjectEnd();
         }
 
-        private static void WriterSystemTextJsonBasicUtf16(bool formatted, ArrayFormatter output, int[] data)
+        private static void WriterSystemTextJsonBasicUtf16(bool formatted, BufferWriter<IBufferWriter<byte>> output, int[] data)
```
benaadams: As above

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
```diff
@@ -199,7 +201,7 @@ private static void WriterNewtonsoftBasic(bool formatted, TextWriter writer, int
             }
         }
 
-        private static void WriterSystemTextJsonHelloWorldUtf8(bool formatted, ArrayFormatter output)
+        private static void WriterSystemTextJsonHelloWorldUtf8(bool formatted, BufferWriter<IBufferWriter<byte>> output)
```
benaadams: As above

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
```diff
@@ -208,7 +210,7 @@ private static void WriterSystemTextJsonHelloWorldUtf8(bool formatted, ArrayForm
             json.WriteObjectEnd();
         }
 
-        private static void WriterSystemTextJsonHelloWorldUtf16(bool formatted, ArrayFormatter output)
+        private static void WriterSystemTextJsonHelloWorldUtf16(bool formatted, BufferWriter<IBufferWriter<byte>> output)
```
benaadams: As above

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -17,15 +19,17 @@ public class JsonWriterTests
         public void WriteJsonUtf8()
         {
             var formatter = new ArrayFormatter(1024, SymbolTable.InvariantUtf8);
-            var json = new JsonWriter(formatter, true, prettyPrint: false);
+            var bufferWriter = new BufferWriter<IBufferWriter<byte>>(formatter);
```
benaadams:
```csharp
var bufferWriter = BufferWriter.Create(formatter);
var json = JsonWriter.Create(formatter, isUtf8: true, prettyPrint: false);
```

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -17,15 +19,17 @@ public class JsonWriterTests
...
+            var json = new JsonWriter(bufferWriter, true, prettyPrint: false);
             Write(ref json);
 
             var formatted = formatter.Formatted;
             var str = Encoding.UTF8.GetString(formatted.Array, formatted.Offset, formatted.Count);
             Assert.Equal(expected, str.Replace(" ", ""));
 
             formatter.Clear();
-            json = new JsonWriter(formatter, true, prettyPrint: true);
+            bufferWriter = new BufferWriter<IBufferWriter<byte>>(formatter);
```
benaadams: As above

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -37,15 +41,17 @@ public void WriteJsonUtf8()
         public void WriteJsonUtf16()
         {
             var formatter = new ArrayFormatter(1024, SymbolTable.InvariantUtf16);
-            var json = new JsonWriter(formatter, false, prettyPrint: false);
+            var bufferWriter = new BufferWriter<IBufferWriter<byte>>(formatter);
```
benaadams: As above

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -37,15 +41,17 @@ public void WriteJsonUtf8()
...
+            var json = new JsonWriter(bufferWriter, false, prettyPrint: false);
             Write(ref json);
 
             var formatted = formatter.Formatted;
             var str = Encoding.Unicode.GetString(formatted.Array, formatted.Offset, formatted.Count);
             Assert.Equal(expected, str.Replace(" ", ""));
 
             formatter.Clear();
-            json = new JsonWriter(formatter, false, prettyPrint: true);
+            bufferWriter = new BufferWriter<IBufferWriter<byte>>(formatter);
```
benaadams: As above

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -86,7 +92,8 @@ public void WriteHelloWorldJsonUtf16(bool prettyPrint)
             string expectedStr = GetHelloWorldExpectedString(prettyPrint, isUtf8: false);
 
             var output = new ArrayFormatter(1024, SymbolTable.InvariantUtf16);
-            var jsonUtf16 = new JsonWriter(output, false, prettyPrint);
+            var bufferWriter = new BufferWriter<IBufferWriter<byte>>(output);
```
benaadams: As above

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -106,7 +113,8 @@ public void WriteHelloWorldJsonUtf8(bool prettyPrint)
             string expectedStr = GetHelloWorldExpectedString(prettyPrint, isUtf8: true);
 
             var output = new ArrayFormatter(1024, SymbolTable.InvariantUtf8);
-            var jsonUtf8 = new JsonWriter(output, true, prettyPrint);
+            var bufferWriter = new BufferWriter<IBufferWriter<byte>>(output);
```
benaadams: As above

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -128,7 +136,8 @@ public void WriteBasicJsonUtf16(bool prettyPrint)
             string expectedStr = GetExpectedString(prettyPrint, isUtf8: false, data);
 
             var output = new ArrayFormatter(1024, SymbolTable.InvariantUtf16);
-            var jsonUtf16 = new JsonWriter(output, false, prettyPrint);
+            var bufferWriter = new BufferWriter<IBufferWriter<byte>>(output);
```
benaadams: As above

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
```diff
@@ -170,7 +179,8 @@ public void WriteBasicJsonUtf8(bool prettyPrint)
             string expectedStr = GetExpectedString(prettyPrint, isUtf8: true, data);
 
             var output = new ArrayFormatter(1024, SymbolTable.InvariantUtf8);
-            var jsonUtf8 = new JsonWriter(output, true, prettyPrint);
+            var bufferWriter = new BufferWriter<IBufferWriter<byte>>(output);
```
benaadams: As above

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
```diff
@@ -57,9 +59,9 @@ public void WriterSystemTextJsonBasic()
         {
             _arrayFormatter.Clear();
             if (IsUTF8Encoded)
-                WriterSystemTextJsonBasicUtf8(Formatted, _arrayFormatter, _data);
+                WriterSystemTextJsonBasicUtf8(Formatted, new BufferWriter<IBufferWriter<byte>>(_arrayFormatter), _data);
```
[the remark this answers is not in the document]
benaadams: i.e. revert

benaadams:
> There is a ~20% regression in performance for the HelloWorld case, across the board (higher constant overhead due to the BufferWriter constructor, roughly 19 ns). However, we get a 15-20% performance improvement for writing a larger json payload.

At a guess the regression is from interface dispatch, and the gain is from buffering in span (reduced virtual dispatch + full advance call costs). Don't think much will be from the constructor.

benaadams:
Suggested changes hopefully will keep the gains while removing the regressions.

> There is a ~20% regression in performance for the HelloWorld case, across the board (higher constant overhead due to the BufferWriter constructor, roughly 19 ns). However, we get a 15-20% performance improvement for writing a larger json payload.

At a guess the regression is from both boxing the formatters and interface dispatch, and the gain is from buffering in span (reduced costs from less virtual dispatch + full advance calls). Don't think much will be from the BufferWriter constructor.

The suggested changes should remove both the boxing to `IBufferWriter` and the additional interface dispatch costs against those boxes

samples/LowAllocationWebServer/Shared/SampleServer.cs
```diff
@@ -62,7 +63,7 @@ static void WriteResponseForGetJson(HttpRequest request, ReadOnlySequence<byte>
             response.AppendEoh();
 
             // write response JSON
-            var jsonWriter = new JsonWriter(response, true, prettyPrint: false);
+            var jsonWriter = new JsonWriter(new BufferWriter<IBufferWriter<byte>>(response), true, prettyPrint: false);
```
benaadams:
```csharp
var jsonWriter = JsonWriter.Create(BufferWriter.Create(response), true, prettyPrint: false);
```

samples/LowAllocationWebServer/Shared/SampleServer.cs
```diff
@@ -94,7 +95,7 @@ static void WriteResponseForPostJson(HttpRequest request, ReadOnlySequence<byte>
...
```
benaadams:
```csharp
var jsonWriter = JsonWriter.Create(BufferWriter.Create(response), true, prettyPrint: false);
```

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
```diff
@@ -3,30 +3,31 @@
...
 using System.Buffers.Text;
+using System.Buffers.Writer;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 namespace System.Text.JsonLab
 {
-    public struct JsonWriter
+    public ref struct JsonWriter
```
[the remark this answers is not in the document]
ahsonkhan: Can you help me understand why that would be beneficial?

Also, if we make JsonWriter generic (with the constraint), how can we add support for stream?

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
ahsonkhan: This is good feedback specific to BufferWriter.
https://github.com/dotnet/corefxlab/issues/2177

cc @KrzysztofCwalina, @JeremyKuhne

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
ahsonkhan: Did you mean to pass-in bufferWriter to JsonWriter.Create? If not, then why would we need to create bufferWriter at all?

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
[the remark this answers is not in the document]
benaadams: > Can you help me understand why that would be beneficial?

Two main reasons
1. If you pass `BufferWriter<IBufferWriter<byte>>` a struct formatter; it will box to the `IBufferWriter<byte>` type, allocation and perf issue. Using the generic will avoid the allocation and boxing.
2. The methods called internally to `BufferWriter<T>` (e.g. `GetSpan` and `Advance`) will go via generic interface dispatch in a shared generic; which can't inline and is also the slowest calling convention for the clr. Changing it to the generic means when you pass it a struct type it will use a non-shared generic and direct, inlinable calls.

> Also, if we make JsonWriter generic (with the constraint), how can we add support for stream?

I assume you mean something not currently present?

Make a struct wrapper for Stream and provide a `Create` overload:
```csharp
public struct StreamWriter : IBufferWriter<byte>
{
    private Stream _stream;

    public StreamWriter(Stream stream)
    {
        _stream = stream;
    }

    public void Advance(int count)
    {
        // ...
    }

    public Memory<byte> GetMemory(int sizeHint = 0)
    {
        // ...
    }

    public Span<byte> GetSpan(int sizeHint = 0)
    {
        // ...
    }
}

public static class JsonWriter
{
    public static JsonWriter<TBufferWriter> Create<TBufferWriter>(
        TBufferWriter bufferWriter,
        bool isUtf8,
        bool prettyPrint = false)
    where TBufferWriter : IBufferWriter<byte>
    {
        return new JsonWriter<TBufferWriter>(bufferWriter, isUtf8, prettyPrint);
    }

    // Stream overload
    public static JsonWriter<StreamWriter> Create<TBufferWriter>(
        Stream stream,
        bool isUtf8,
        bool prettyPrint = false)
    where TBufferWriter : IBufferWriter<byte>
    {
        return new JsonWriter<StreamWriter>(new StreamWriter(stream), isUtf8, prettyPrint);
    }
}
```

tests/System.Text.JsonLab.Tests/JsonWriterTests.cs
benaadams: Yes, edited :)

tests/Benchmarks/System.IO.Pipelines/E2E.cs
```diff
@@ -57,7 +61,7 @@ public void TechEmpowerJsonNoIO(int numberOfRequests, int concurrentConnections)
                 formatter.Append("\r\n\r\n");
 
                 // write body
-                var writer = new JsonWriter(formatter, true);
+                var writer = new JsonWriter(new BufferWriter<IBufferWriter<byte>>(formatter), true, true);
```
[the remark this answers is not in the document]
benaadams: Fixed Create method, had it wrong before; it just needs to pass through the generic type

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
benaadams: Added comment for issue

benaadams: Change to overload is still relevant for this PR though

benaadams:
> Also, calling Flush after each call to Advance is a nuisance.

The methods shouldn't be calling `Flush`, only `Advance`. 

`Flush` should be done by the caller after a group of operations.

benaadams:
> Also, calling Flush after each call to Advance is a nuisance.

The methods shouldn't be calling `Flush`, only `Advance`. 

`Flush` should be done by the caller after a group of operations; and before finishing with the `BufferWriter`.

i.e. `Advance` calls "Flush" when it needs to, and an explicit `Flush` is to clear out anything remaining in the Span that hasn't been `Advanced` in the underlying Wrtier (e.g. is still buffered)

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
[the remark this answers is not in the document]
MarcoRossignoli: > The methods called internally to BufferWriter<T> (e.g. GetSpan and Advance) will go via generic interface dispatch in a shared generic; which can't inline and is also the slowest calling convention for the clr. Changing it to the generic means when you pass it a struct type it will use a non-shared generic and direct, inlinable calls.

@benaadams is related to CLR behaviour explained here https://blogs.msdn.microsoft.com/joelpob/2004/11/17/clr-generics-and-code-sharing/ ?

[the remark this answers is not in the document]
benaadams: The `T` parameter is explicitly being specified as an interface: `IBufferWriter<byte>`; so if you create that `BufferWriter<IBufferWriter<byte>>` and give it a struct as the T, then its boxed to the interface and stored in the `BufferWriter<IBufferWriter<byte>>` as the boxed `IBufferWriter<byte>` type

KrzysztofCwalina:
> Since JsonWriter contains a BufferWriter field now, it has to be a ref struct.

The root reason is that now JsonWriter stores/caches the Span. It's done through BufferWriter, but even if you did not use BufferWriter, you would want to cache the span from IBufferWriter, wouldn't you?

On a related note: this PR moves from not caching the span to caching the span and it is only 20% perf improvement? 

Also, writing hello world is what JSON Techempower does and so it would be good to drill into why there is a regression here.

Lastly, I suspect the perf loss is because you now have the overhead of BufferWriter without the benefits of caching the span in cases where the JsonWriter is used to write (access the span) a small number of times. As the number of times you access the span increases, the cost of the buffer writer will be amortized but the cost of individual writes will be smaller.

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
```diff
@@ -77,6 +78,7 @@ private void WriteStartUtf8(int bytesNeeded, byte token)
 
             byteBuffer[idx] = token;
             _bufferWriter.Advance(bytesNeeded);
+            _bufferWriter.Flush();
```
KrzysztofCwalina: I don't think we can flush on every write. Flush is very expensive when called on Pipelines IBufferWriter. 

cc: @pakrym, @davidfowl

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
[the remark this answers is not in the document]
ahsonkhan: > I assume you mean something not currently present?

Yes.

> Make a struct wrapper for Stream and provide a `Create` overload:

That is a great idea! I was trying to see if there was a way to add stream support directly to JsonWriter without exposing additional public types (like the StreamWriter wrapper). However, the boxing and performance benefits you mentioned are definitely compelling.

ahsonkhan:
> The methods shouldn't be calling `Flush`, only `Advance`. 

> `Flush` should be done by the caller after a group of operations; and before finishing with the `BufferWriter`.

> [EDIT] The current PR calls BufferWriter.Flush in every json write method, this is a perf killer.

> I don't think we can flush on every write. Flush is very expensive when called on Pipelines IBufferWriter. In general, calling flush on every write defeats the purpose of BufferWriter (as Flush will cause Buffer Writer to fish out a new span from Memory)

I agree and it makes sense to me. I will make this change and update the results.

cc @GrabYourPitchforks, since we just talked about this offline as well.

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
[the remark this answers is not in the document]
ahsonkhan: > If you pass `BufferWriter<IBufferWriter<byte>>` a struct formatter; it will box to the `IBufferWriter<byte>` type, allocation and perf issue. Using the generic will avoid the allocation and boxing.

I wasn't seeing any allocations because the formatter used within the benchmarks is a class (ArrayFormatter). Good point.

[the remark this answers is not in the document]
benaadams: Ah, missed that. Can always measure the effects of the call costs by creating a second struct that just wraps and using that in a test instead of `ArrayFormatter` e.g.
```csharp
public struct ArrayFormatterWrapper : ITextBufferWriter, IDisposable
{
    private ArrayFormatter _arrayFormatter;

    public ArrayFormatter(int capacity, SymbolTable symbolTable, ArrayPool<byte> pool = null)
    {
        _arrayFormatter = new ArrayFormatterInternal(capacity, symbolTable, pool);
    }

    public int CommitedByteCount => _arrayFormatter.CommitedByteCount;

    public void Clear()
    {
        _arrayFormatter.Clear();
    }

    public ArraySegment<byte> Free => _arrayFormatter.Free;
    public ArraySegment<byte> Formatted => _arrayFormatter.Formatted;

    public SymbolTable SymbolTable => _arrayFormatter.SymbolTable;

    public Memory<byte> GetMemory(int minimumLength = 0)
    {
        _arrayFormatter.GetMemory(minimumLength);
    }

    public Span<byte> GetSpan(int minimumLength = 0)
    {
        _arrayFormatter.GetSpan(minimumLength);
    }

    public void Advance(int bytes)
    {
        _arrayFormatter.Advance(bytes);
    }
    public int MaxBufferSize { get; } = _arrayFormatter.MaxBufferSize;

    public void Dispose()
    {
        _arrayFormatter.Dispose();
    }
}
```

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
ahsonkhan: > Again add a convenience helper


We already have such a convenience method:
https://github.com/dotnet/corefxlab/blob/a02d39de49df0257b9b3a92bc7504ede33847f1e/src/System.Buffers.ReaderWriter/System/Buffers/Writer/BufferWriter.cs#L24-L26

src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs
davidfowl: Agreed

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
benaadams: Concrete type it, but not its parameter
```csharp
public static class JsonWriter
{
    public static JsonWriter<TBufferWriter> Create<TBufferWriter>(
        TBufferWriter bufferWriter,
        bool isUtf8,
        bool prettyPrint = false)
    where TBufferWriter : IBufferWriter<byte>
    {
        return new JsonWriter<TBufferWriter>(BufferWriter.Create(bufferWriter), isUtf8, prettyPrint);
    }

    // Stream overload
    public static JsonWriter<StreamWriter> Create(
        Stream stream,
        bool isUtf8,
        bool prettyPrint = false)
    {
        return new JsonWriter<StreamWriter>(BufferWriter.Create(new StreamWriter(stream)), isUtf8, prettyPrint);
    }
}

public ref struct JsonWriter<TBufferWriter>
    where TBufferWriter : IBufferWriter<byte>
{
    private BufferWriter<TBufferWriter> _bufferWriter;

    public JsonWriter(BufferWriter<TBufferWriter> bufferWriter, bool isUtf8, bool prettyPrint = false)
    {
        _bufferWriter = bufferWriter;
        // ...
    }
}
```

ahsonkhan: > Concrete type it, but not its parameter?

I don't follow. We still have the issue with JsonWriter.Create since the first argument a generic type:
```C#
var formatter = new ArrayFormatter(1024, SymbolTable.InvariantUtf8);
var bufferWriter = BufferWriter.Create(formatter);
var json = JsonWriter.Create(bufferWriter, isUtf8: true, prettyPrint: false); // Build error
```

Error	CS0306	The type `BufferWriter<ArrayFormatter>` may not be used as a type argument	System.Text.JsonLab.Tests	E:\GitHub\Fork\corefxlab\tests\System.Text.JsonLab.Tests\JsonWriterTests.cs	23	Active

benaadams: Ah, gotcha. Just need to add concrete Create overload
```csharp
public static class JsonWriter
{
    public static JsonWriter<TBufferWriter> Create<TBufferWriter>(
        TBufferWriter bufferWriter,
        bool isUtf8,
        bool prettyPrint = false)
    where TBufferWriter : IBufferWriter<byte>
    {
        return new JsonWriter<TBufferWriter>(BufferWriter.Create(bufferWriter), isUtf8, prettyPrint);
    }

    // extra overload
    public static JsonWriter<TBufferWriter> Create<TBufferWriter>(
        BufferWriter<TBufferWriter> bufferWriter,
        bool isUtf8,
        bool prettyPrint = false)
    where TBufferWriter : IBufferWriter<byte>
    {
        return new JsonWriter<TBufferWriter>(bufferWriter, isUtf8, prettyPrint);
    }
}
```

ahsonkhan:
> The root reason is that now JsonWriter stores/caches the Span. It's done through BufferWriter, but even if you did not use BufferWriter, you would want to cache the span from IBufferWriter, wouldn't you?

> On a related note: this PR moves from not caching the span to caching the span and it is only 20% perf improvement?

We actually end up improving 30%-35%. Here are the updated results with the calls to flush moved to the caller rather than within the JsonWriter APIs (reducing from multiple calls to a single call).

|                         Method | IsUTF8Encoded | Formatted |         Mean |       Error |      StdDev |       Median |  Gen 0 | Allocated |
|------------------------------- |-------------- |---------- |-------------:|------------:|------------:|-------------:|-------:|----------:|
|      **WriterSystemTextJsonBasic** |         **False** |     **False** |  **7,540.44 ns** | **150.3238 ns** | **140.6130 ns** |  **7,587.65 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |     False | 20,027.49 ns | 171.1728 ns | 142.9370 ns | 20,002.59 ns | 0.0916 |     472 B |
| WriterSystemTextJsonHelloWorld |         False |     False |     78.16 ns |   1.5752 ns |   3.1819 ns |     77.67 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |     False |    222.25 ns |   4.2226 ns |   4.1472 ns |    221.30 ns | 0.0741 |     312 B |
|      **WriterSystemTextJsonBasic** |         **False** |      **True** |  **9,551.74 ns** | **155.0085 ns** | **129.4391 ns** |  **9,557.50 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |      True | 29,313.71 ns | 348.3948 ns | 308.8429 ns | 29,253.00 ns | 0.1221 |     632 B |
| WriterSystemTextJsonHelloWorld |         False |      True |     78.19 ns |   1.5312 ns |   1.8228 ns |     78.64 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |      True |    318.94 ns |   1.6915 ns |   1.3206 ns |    318.67 ns | 0.1121 |     472 B |
|      **WriterSystemTextJsonBasic** |          **True** |     **False** |  **7,792.17 ns** | **115.3570 ns** | **102.2609 ns** |  **7,765.99 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |     False | 19,222.77 ns | 444.5917 ns | 371.2542 ns | 19,043.86 ns | 0.0916 |     416 B |
| WriterSystemTextJsonHelloWorld |          True |     False |    125.39 ns |   0.6210 ns |   0.5186 ns |    125.50 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |     False |    261.20 ns |   5.2267 ns |   6.0191 ns |    261.31 ns | 0.0606 |     256 B |
|      **WriterSystemTextJsonBasic** |          **True** |      **True** | **10,282.52 ns** | **202.2167 ns** | **403.8487 ns** | **10,190.40 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |      True | 29,028.33 ns | 574.0739 ns | 589.5316 ns | 29,175.77 ns | 0.1221 |     576 B |
| WriterSystemTextJsonHelloWorld |          True |      True |    143.44 ns |   3.0589 ns |   8.8256 ns |    140.09 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |      True |    359.92 ns |   7.1477 ns |  11.9422 ns |    358.37 ns | 0.0987 |     416 B |

ahsonkhan:
@benaadams, I followed your suggestion and made JsonWriter generic, to avoid boxing. This regressed performance significantly (2+ times slower than before).

Changed JsonWriter from `public ref struct JsonWriter` to `public ref struct JsonWriter<TBufferWriter> where TBufferWriter : IBufferWriter<byte>`

Changed the ctor from `public JsonWriter(BufferWriter<IBufferWriter<byte>> bufferWriter,...)` to `public JsonWriter(BufferWriter<TBufferWriter> bufferWriter,...)`

Reverted the performance test changes back to what they were before and used JsonWriter.Create helper method.

|                         Method | IsUTF8Encoded | Formatted |        Mean |      Error |       StdDev |      Median |  Gen 0 | Allocated |
|------------------------------- |-------------- |---------- |------------:|-----------:|-------------:|------------:|-------:|----------:|
|      **WriterSystemTextJsonBasic** |         **False** |     **False** | **15,334.4 ns** | **219.009 ns** |   **194.145 ns** | **15,376.6 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |     False | 19,837.1 ns |  78.258 ns |    69.373 ns | 19,837.9 ns | 0.0916 |     472 B |
| WriterSystemTextJsonHelloWorld |         False |     False |    109.7 ns |   2.024 ns |     1.794 ns |    108.9 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |     False |    220.3 ns |   4.455 ns |     6.097 ns |    217.7 ns | 0.0741 |     312 B |
|      **WriterSystemTextJsonBasic** |         **False** |      **True** | **20,123.9 ns** | **105.785 ns** |    **76.489 ns** | **20,123.3 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |         False |      True | 30,560.0 ns | 608.719 ns | 1,666.358 ns | 29,923.4 ns | 0.1221 |     632 B |
| WriterSystemTextJsonHelloWorld |         False |      True |    132.3 ns |   1.571 ns |     1.312 ns |    132.5 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |         False |      True |    342.7 ns |   6.887 ns |    18.263 ns |    337.9 ns | 0.1121 |     472 B |
|      **WriterSystemTextJsonBasic** |          **True** |     **False** | **16,718.1 ns** | **366.409 ns** |   **376.275 ns** | **16,745.5 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |     False | 20,227.8 ns | 401.237 ns |   961.337 ns | 19,912.5 ns | 0.0916 |     416 B |
| WriterSystemTextJsonHelloWorld |          True |     False |    155.2 ns |   3.028 ns |     3.605 ns |    153.4 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |     False |    271.0 ns |   5.424 ns |     8.444 ns |    269.8 ns | 0.0606 |     256 B |
|      **WriterSystemTextJsonBasic** |          **True** |      **True** | **20,833.2 ns** | **412.624 ns** |   **344.560 ns** | **20,753.4 ns** |      **-** |       **0 B** |
|          WriterNewtonsoftBasic |          True |      True | 30,173.9 ns | 611.857 ns | 1,149.215 ns | 29,645.0 ns | 0.1221 |     576 B |
| WriterSystemTextJsonHelloWorld |          True |      True |    175.4 ns |   1.478 ns |     1.234 ns |    175.7 ns |      - |       0 B |
|     WriterNewtonsoftHelloWorld |          True |      True |    364.1 ns |   5.279 ns |     4.408 ns |    364.0 ns | 0.0987 |     416 B |

benaadams:
Interesting, and using a struct wrapper over ArrayWriterFormatter? https://github.com/dotnet/corefxlab/pull/2358#discussion_r194947621

benaadams:
Interesting, and using a struct wrapper over ArrayWriterFormatter, https://github.com/dotnet/corefxlab/pull/2358#discussion_r194947621 to move it completely away from a shared generic?

I wonder if some methods that previously inlined are no longer inlining due to the additional il complexity that the Jit sees. 

Moving from an explicit interface to a generic constraint shouldn't cause a 2x drop in perf for a object type, or there's serious runtime issues?

ahsonkhan:
> I wonder if some methods that previously inlined are no longer inlining due to the additional il complexity that the Jit sees.

> Moving from an explicit interface to a generic constraint shouldn't cause a 2x drop in perf for a object type, or there's serious runtime issues?

cc @AndyAyersMS, maybe you can provide some insight here, as to why we are seeing the regression. Any ideas?

**Version 1 - baseline:**
https://github.com/ahsonkhan/corefxlab/blob/5370b7b26ab29aacaafbb129d1ec3527650b4f7c/src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs#L13-L38

**Version 2 - 2x slower:**
https://github.com/dotnet/corefxlab/blob/1d2604910630b3b64f025bb7101165aaf8fefee2/src/System.Text.JsonLab/System/Text/Json/JsonWriter.cs#L13-L38

Changes to the performance test between v1 and v2:
https://github.com/dotnet/corefxlab/pull/2358/commits/1d2604910630b3b64f025bb7101165aaf8fefee2#diff-ca666b8a8d4a2a773e161e08cdd3cd67L208

AndyAyersMS:
My hunch is you may be running into limits inlining methods from one shared generic class into methods from some other shared generic class. I can take a more detailed look later.

benaadams:
Less than great, but you could go for a double implementation. One for structs and an interface based one for classes via interface: [sharplab.io](https://sharplab.io/#v2:EYLgZgpghgLgrgJwgZwLRIMaOQSwG4SoAOsMECAdsgDQwg4A2APgAIAMABCwIwDcAsACghLAMwccFMgjBQMEDgEkAQnDCQEAdQQ5pAHgAqAPg4BvAL4jxSMB2QwEcDDA6r15bbvKGTAdwAW5AoGHCBKbhqe+sAAnmRGQqZCHClc4hEeOtIAFCHAahq+WeQAlBZClsKCYhw2dg5OLhlaxQiJyamdNTwAbK4FmV4IhgDycDBE4yYAwkiwEKPjkzBGuWMT4xwA9kvjJR2dhxwBQRwG68uh4QMtQ3qx8QdHqQC8JhQQvv3ut/rnuytsjsNjASgJBJUrLUILZ7I5nBwAFLILYUKLeAzNdEIPyBJBnLGtK5wxrUa4/bH3OIQBKCJKCQ5EHR4ebfSKtQyEoYmAD6+QprXBhyeaSRKLRrWyXL+0vIJn57KGZOAWy2DAkyAAqjAwAAOZWq9VMiAwGAxAAKOikHBeHFkDGQEH2DM69OeKT5N2xNo4CsG0iFnUqkOq1hh9XhLmRqOx7RdKSZ+FZsuGKi9HIeNN5ft+5EDKSeNWjEqG2TTArumfl6aVvsNGu1eoNao4xtNFqtLlt9sdztdItSnor0h9Hy+Kb05cV0WpRlWOexYKelU6hfEvTFMY5mJr0hmczInN3cuyA5SO+H5F9x4Q1DPdZbOC1Ov195VLbbZstki7dqgDqdY48WCFNiQaZwySnf1vCrJ43WeFgAHYODHTcSxlG9VhTAA6WZoDIbIF1aEoySfRt9VbJB22/KQl3jDgQy6dduD6YtKQvac5Q4PD5iPS8cVPejOgnDjoJxa9+LvITUnfdUyJfKT3VkyiTS/TsfR7QCTnxUTcwQMDI0gidYPo+CjiQlDPjQ9iU3nG9SOfJsVOozs6KDJ41y4FjrKJHiCPvKC9KpeIJM42833reSmwij8qLUn8NP/Xs4PvCzULYyUiNrKKKM/Dsfzc1JgyEEQACYOAAQQQBAoBib0wkCykq3KQQSQRABlcCYHq8kwuCmkWqhFhyp4RC4yYry+hYAAWLzdWyPtUjMw4WX0qAvltVCqpqurJUKo4MqGDgACtfG4H1DukXCDwgbJ1rJBoIDJTT9pWqB9OQDbLK+TrI2xBb80OS7vF+xpsRMU7yttYGEGu/Dbs+h7HCev8ANe4MgA==)
```csharp
public ref struct JsonWriter<TBufferWriter> where TBufferWriter : struct, IBufferWriter<byte>
{
    private BufferWriter<TBufferWriter> _bufferWriter;

    public JsonWriter(BufferWriter<TBufferWriter> bufferWriter, bool isUtf8, bool prettyPrint = false)
    {
        _bufferWriter = bufferWriter;
    }
}

public ref struct JsonWriter
{
    private BufferWriter<IBufferWriter<byte>> _bufferWriter;

    public JsonWriter(IBufferWriter<byte> bufferWriter, bool isUtf8, bool prettyPrint = false)
    {
        _bufferWriter = new BufferWriter<IBufferWriter<byte>>(bufferWriter);
    }

    // Helper Create methods
    public static JsonWriter<TBufferWriter> Create<TBufferWriter>(
        TBufferWriter bufferWriter,
        bool isUtf8,
        bool prettyPrint = false) where TBufferWriter : struct, IBufferWriter<byte>
    {
        return new JsonWriter<TBufferWriter>(BufferWriter.Create(bufferWriter), isUtf8, prettyPrint);
    }

    public static JsonWriter<TBufferWriter> Create<TBufferWriter>(
        BufferWriter<TBufferWriter> bufferWriter,
        bool isUtf8,
        bool prettyPrint = false) where TBufferWriter : struct, IBufferWriter<byte>
    {
        return new JsonWriter<TBufferWriter>(bufferWriter, isUtf8, prettyPrint);
    }

    public static JsonWriter Create(
        IBufferWriter<byte> bufferWriter,
        bool isUtf8,
        bool prettyPrint = false)
    {
        return new JsonWriter(bufferWriter, isUtf8, prettyPrint);
    }
}

class ArrayWriter : IBufferWriter<byte> { }
struct StructWriter : IBufferWriter<byte> { }

public class Program
{
    public static void Main()
    {
        var aw = new ArrayWriter();
        JsonWriter jw1 = JsonWriter.Create(aw, true, false);
        var sw = new StructWriter();
        JsonWriter<StructWriter> jw2 = JsonWriter.Create(sw, true, false);
    }
}
```
Then it should completely bypass shared generics

ahsonkhan:
> Interesting, and using a struct wrapper over ArrayWriterFormatter, https://github.com/dotnet/corefxlab/pull/2358#discussion_r194947621 to move it completely away from a shared generic?

FYI, wrapping the ArrayFormatter class in a struct does remove the regression.

ahsonkhan:
> Also, writing hello world is what JSON Techempower does and so it would be good to drill into why there is a regression here.

> Lastly, I suspect the perf loss is because you now have the overhead of BufferWriter without the benefits of caching the span in cases where the JsonWriter is used to write (access the span) a small number of times. As the number of times you access the span increases, the cost of the buffer writer will be amortized but the cost of individual writes will be smaller.

I agree. The ctor of BufferWriter alone has noticeable overhead, and as you said, the cost gets amortized when we have multiple writes. For instance, we break even if we write HelloWorld twice rather than just once.

ahsonkhan:
> Less than great, but you could go for a double implementation. One for structs and an interface based one for classes via interface

> Remove the enforced interface dispatch and use generic writer type.

For now, I have reverted this change and kept the enforced interface dispatch. I will revisit this change in another PR once I can make sense of the performance trade-offs. The 2x regression needs to be addressed. https://github.com/dotnet/corefxlab/commit/1d2604910630b3b64f025bb7101165aaf8fefee2

ahsonkhan:

AndyAyersMS:
I looked at the "slow" generic version 1d26049 vs on the first test `WriterSystemTextJsonBasic` and here's what I saw:
* benchmark is basically timing the speed of `WriteValue(long)`
* in both versions inlining is comparable and neither `WriteValue(long)` or `WriteValueUtf16(long, ..)` are inlined. So my hunch above was wrong.
* in the slow version the time spent in `WriteValueUtf16` is considerably higher because of the relatively high cost of static field access in a generic type (in this case `s_newLineUtf16.Length`).

So the likely fix is to rework how the statics are implemented -- for instance move them into a non-generic helper class.

I also think perf might be improved in general by not fusing together utf8 and utf16 handling into one class -- both because many methods must check at runtime which mode the writer is in and also because it makes the method code size larger and will tend to reduce the amount of inlining that will happen.

ahsonkhan:
> So the likely fix is to rework how the statics are implemented -- for instance move them into a non-generic helper class.

That ends up making a significant impact. Thanks for the analysis, Andy!
There is still a gap in performance though.

**Original (fastest)**

|                    Method | IsUTF8Encoded | Formatted |     Mean |     Error |    StdDev | Allocated |
|-------------------------- |-------------- |---------- |---------:|----------:|----------:|----------:|
| **WriterSystemTextJsonBasic** |         **False** |     **False** | **7.565 us** | **0.1456 us** | **0.1290 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |         **False** |      **True** | **9.460 us** | **0.1880 us** | **0.1759 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |          **True** |     **False** | **7.637 us** | **0.0827 us** | **0.0645 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |          **True** |      **True** | **9.626 us** | **0.1793 us** | **0.1677 us** |       **0 B** |

**Static field access in a generic type (2x slower)**

|                    Method | IsUTF8Encoded | Formatted |     Mean |     Error |    StdDev | Allocated |
|-------------------------- |-------------- |---------- |---------:|----------:|----------:|----------:|
| **WriterSystemTextJsonBasic** |         **False** |     **False** | **16.06 us** | **0.3475 us** | **0.4984 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |         **False** |      **True** | **20.72 us** | **0.2223 us** | **0.2080 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |          **True** |     **False** | **15.86 us** | **0.1656 us** | **0.1383 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |          **True** |      **True** | **20.46 us** | **0.3870 us** | **0.4141 us** |       **0 B** |

**Using a static helper class (25% slower than original, but significant improvement from before)**

|                    Method | IsUTF8Encoded | Formatted |     Mean |     Error |    StdDev | Allocated |
|-------------------------- |-------------- |---------- |---------:|----------:|----------:|----------:|
| **WriterSystemTextJsonBasic** |         **False** |     **False** | **10.08 us** | **0.1562 us** | **0.1461 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |         **False** |      **True** | **11.78 us** | **0.2171 us** | **0.2031 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |          **True** |     **False** | **10.36 us** | **0.1123 us** | **0.1051 us** |       **0 B** |
| **WriterSystemTextJsonBasic** |          **True** |      **True** | **12.33 us** | **0.2444 us** | **0.6772 us** |       **0 B** |


> I also think perf might be improved in general by not fusing together utf8 and utf16 handling into one class -- both because many methods must check at runtime which mode the writer is in and also because it makes the method code size larger and will tend to reduce the amount of inlining that will happen.

We had done that previously, but reverted it since it had too little impact on performance to justify splitting the types. We could revisit this again.
https://github.com/dotnet/corefxlab/pull/2348

If, in general, the WriteX methods are not getting inlined anyway, then splitting the types wouldn't improve perf significantly, correct?

benaadams:
Not being generic would be an issue currently for `BufferWriterFormatter`, `StreamFormatter` as they are structs; also limit user defined formatters (as it would make structs an issue); so ideally this could be resolved via the generic path.

ahsonkhan:
> Not being generic would be an issue currently for `BufferWriterFormatter`, `StreamFormatter` as they are structs; also limit user defined formatters (as it would make structs an issue); so ideally this could be resolved via the generic path.

For the struct case, we would avoid boxing and allocation via the generic path, but we still have to account for the ~25% overhead for reference types like [PipeWriter](https://github.com/dotnet/corefx/blob/4ad640888343cd5e6d73a00e87fe4953559daef8/src/System.IO.Pipelines/src/System/IO/Pipelines/PipeWriter.cs#L14) (which is an abstract class). Would the guideline then be to wrap any classes that implement `IBufferWriter<byte>` in a struct before passing it (or maybe, the JsonWriter would do wrap it in a struct as part of its ctor)?

benaadams:
Do you have a branch with "Using a static helper class"?

ahsonkhan:
> Do you have a branch with "Using a static helper class"?

https://github.com/ahsonkhan/corefxlab/tree/UseStaticHelper
- Move statics to a non-generic helper class (fd5b7a3c16787e92e5f5fe00db98340a9a401f68)
- Add a generic JsonWriter to try to optimize the struct case (355934358494d155759bf21a469c6cf817f93bb1)

benaadams:
Looking...

benaadams:
Might have something...

benaadams:
How's this? https://github.com/dotnet/corefxlab/pull/2366; not sure it the types aren't a bit strange...

AndyAyersMS:
As far as separate classes vs internal utf8/utf16 switching, there is possibly some small code locality benefit as the way the methods are set up now half of each method is going to be coldish code.

I also think of splitting types as future proofing as much as anything. Someday (hopefully soon) we'll be able to give the jit a more adaptive inline policy but the apparent size of a method will always be a consideration in the inlining decision. By fusing two different paths into one method you make it less likely that the method will ever get inlined.

ahsonkhan:
Superseded by https://github.com/dotnet/corefxlab/pull/2374. Closing.

PR #2366, opened 2018-06-18
Add support for BufferWriter<T> to the JsonWriter

benaadams:
Follow up to https://github.com/dotnet/corefxlab/pull/2358

/cc @ahsonkhan


benaadams:
Pre
```
                         Method | IsUTF8Encoded | Formatted |         Mean | Allocated |
------------------------------- |-------------- |---------- |-------------:|----------:|
      WriterSystemTextJsonBasic |         False |     False | 13,405.64 ns |       0 B |
 WriterSystemTextJsonHelloWorld |         False |     False |     76.70 ns |       0 B |
      WriterSystemTextJsonBasic |         False |      True | 15,203.72 ns |       0 B |
 WriterSystemTextJsonHelloWorld |         False |      True |     82.64 ns |       0 B |
      WriterSystemTextJsonBasic |          True |     False | 13,232.31 ns |       0 B |
 WriterSystemTextJsonHelloWorld |          True |     False |    126.33 ns |       0 B |
      WriterSystemTextJsonBasic |          True |      True | 15,041.09 ns |       0 B |
 WriterSystemTextJsonHelloWorld |          True |      True |    134.63 ns |       0 B |
```
Post
```
                         Method | IsUTF8Encoded | Formatted |        Mean | Allocated |
------------------------------- |-------------- |---------- |------------:|----------:|
      WriterSystemTextJsonBasic |         False |     False |  8,271.6 ns |       0 B |
 WriterSystemTextJsonHelloWorld |         False |     False |    107.9 ns |       0 B |
      WriterSystemTextJsonBasic |         False |      True | 10,033.6 ns |       0 B |
 WriterSystemTextJsonHelloWorld |         False |      True |    126.1 ns |       0 B |
      WriterSystemTextJsonBasic |          True |     False |  8,262.6 ns |       0 B |
 WriterSystemTextJsonHelloWorld |          True |     False |    163.8 ns |       0 B |
      WriterSystemTextJsonBasic |          True |      True | 10,083.9 ns |       0 B |
 WriterSystemTextJsonHelloWorld |          True |      True |    172.2 ns |       0 B |
```

ahsonkhan:
What is the performance impact of keeping a single type rather than splitting it into encoding specific writers (JsonWriterUtf8/JsonWriterUtf16)? From my previous testing, the performance difference wasn't worth splitting it up (https://github.com/dotnet/corefxlab/pull/2348).

> Post

For the hello world case, the regression is quite bad. What's causing it? Is the Create method the bulk of the cost?

cc @KrzysztofCwalina

benaadams:
Inlining the ctors does help
```
                         Method | IsUTF8Encoded | Formatted |        Mean | Allocated |
------------------------------- |-------------- |---------- |------------:|----------:|
      WriterSystemTextJsonBasic |         False |     False |  7,658.4 ns |       0 B |
 WriterSystemTextJsonHelloWorld |         False |     False |    109.7 ns |       0 B |
      WriterSystemTextJsonBasic |         False |      True |  9,478.8 ns |       0 B |
 WriterSystemTextJsonHelloWorld |         False |      True |    108.5 ns |       0 B |
      WriterSystemTextJsonBasic |          True |     False |  8,533.1 ns |       0 B |
 WriterSystemTextJsonHelloWorld |          True |     False |    154.7 ns |       0 B |
      WriterSystemTextJsonBasic |          True |      True |  9,874.7 ns |       0 B |
 WriterSystemTextJsonHelloWorld |          True |      True |    164.4 ns |       0 B |
```
Still looking into the "Hello world"

benaadams:
What I find a bit weird is why uft8 is slower than utf16 for Hello World

ahsonkhan:
> What I find a bit weird is why uft8 is slower than utf16 for Hello World

It's the encoding step since we are dealing with strings (utf16) rather than utf8 string literals.
https://github.com/dotnet/corefxlab/pull/2366/files#diff-3db8612306b944da32a84c0b13a8e6c2R219

benaadams:
> For the hello world case, the regression is quite bad. What's causing it? Is the Create method the bulk of the cost?

Yeah think its the `vmovdqu` shuffling from the wrappers, will see if I can isolate a test case for the Jit
 
/cc @AndyAyersMS 
```
System.Text.JsonLab.Benchmarks.JsonWriterPerf.WriterSystemTextJsonHelloWorldUtf8(Boolean, System.Text.Formatting.ArrayFormatter) |  
-- | --
  | JsonWriterUtf8 json = JsonWriter.CreateUtf8(output, formatted);             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | mov     rcx,rdx
  | lea     rdx,[rsp+88h]
  | vxorps  xmm0,xmm0,xmm0
  | vmovdqu xmmword ptr [rdx],xmm0
  | vmovdqu xmmword ptr [rdx+10h],xmm0
  | vmovdqu xmmword ptr [rdx+20h],xmm0
  | xor     edx,edx
  | mov     dword ptr [rsp+88h],edx
  | lea     rdx,[rsp+90h]
  | mov     qword ptr [rdx],rcx
  | lea     rdx,[rsp+98h]
  | mov     r11,7FFC140F0340h
  | xor     r8d,r8d
  | mov     rax,7FFC140F0340h
  | cmp     dword ptr [rcx],ecx
  | call    qword ptr [rax]
  | mov     rcx,7FFC1435F2C8h
  | xor     edx,edx
  | call    coreclr!GC_VersionInfo+0x25a30
  | mov     rcx,21F58C52C90h
  | mov     rcx,qword ptr [rcx]
  | test    rcx,rcx
  | jne     M01_L00
  | xor     eax,eax
  | xor     edx,edx
  | jmp     M01_L01
M01_L00 |  
  | lea     rax,[rcx+10h]
  | mov     edx,dword ptr [rcx+8]
M01_L01 |  
  | lea     rcx,[rsp+0A8h]
  | mov     qword ptr [rcx],rax
  | mov     dword ptr [rcx+8],edx
  | vmovdqu xmm0,xmmword ptr [rsp+88h]
  | vmovdqu xmmword ptr [rsp+0B8h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+98h]
  | vmovdqu xmmword ptr [rsp+0C8h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+0A8h]
  | vmovdqu xmmword ptr [rsp+0D8h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+0B8h]
  | vmovdqu xmmword ptr [rsp+58h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+0C8h]
  | vmovdqu xmmword ptr [rsp+68h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+0D8h]
  | vmovdqu xmmword ptr [rsp+78h],xmm0
  | xor     ecx,ecx
  | lea     rax,[rsp+20h]
  | vxorps  xmm0,xmm0,xmm0
  | vmovdqu xmmword ptr [rax],xmm0
  | vmovdqu xmmword ptr [rax+10h],xmm0
  | vmovdqu xmmword ptr [rax+20h],xmm0
  | mov     qword ptr [rax+30h],rcx
  | vmovdqu xmm0,xmmword ptr [rsp+58h]
  | vmovdqu xmmword ptr [rsp+28h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+68h]
  | vmovdqu xmmword ptr [rsp+38h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+78h]
  | vmovdqu xmmword ptr [rsp+48h],xmm0
  | movzx   ecx,sil
  | mov     byte ptr [rsp+24h],cl
  | mov     dword ptr [rsp+20h],0FFFFFFFFh
  | mov     byte ptr [rsp+25h],1
  | vmovdqu xmm0,xmmword ptr [rsp+20h]
  | vmovdqu xmmword ptr [rsp+0E8h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+30h]
  | vmovdqu xmmword ptr [rsp+0F8h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+40h]
  | vmovdqu xmmword ptr [rsp+108h],xmm0
  | mov     rcx,qword ptr [rsp+50h]
  | mov     qword ptr [rsp+118h],rcx
  | vmovdqu xmm0,xmmword ptr [rsp+0E8h]
  | vmovdqu xmmword ptr [rsp+120h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+0F8h]
  | vmovdqu xmmword ptr [rsp+130h],xmm0
  | vmovdqu xmm0,xmmword ptr [rsp+108h]
  | vmovdqu xmmword ptr [rsp+140h],xmm0
  | mov     rcx,qword ptr [rsp+118h]
  | mov     qword ptr [rsp+150h],rcx
  | json.WriteObjectStart();             ^^^^^^^^^^^^^^^^^^^^^^^^
  | lea     rcx,[rsp+120h]
  | call    System.Text.JsonLab.JsonWriterUtf8`1[[System.Text.JsonLab.IBufferWriter, System.Text.JsonLab]].WriteObjectStart()
  | json.WriteAttribute("message", "Hello, World!");             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  | mov     rdx,21F58C59578h
  | mov     rdx,qword ptr [rdx]
  | mov     r8,21F58C59580h
  | mov     r8,qword ptr [r8]
  | lea     rcx,[rsp+120h]
  | call    System.Text.JsonLab.JsonWriterUtf8`1[[System.Text.JsonLab.IBufferWriter, System.Text.JsonLab]].WriteAttribute(System.String, System.String)
  | json.WriteObjectEnd();             ^^^^^^^^^^^^^^^^^^^^^^
  | lea     rcx,[rsp+120h]
  | call    System.Text.JsonLab.JsonWriterUtf8`1[[System.Text.JsonLab.IBufferWriter, System.Text.JsonLab]].WriteObjectEnd()
  | json.Flush();             ^^^^^^^^^^^^^
  | mov     edx,dword ptr [rsp+128h]
  | test    edx,edx
  | jle     M01_L02
  | xor     ecx,ecx
  | mov     dword ptr [rsp+128h],ecx
  | mov     rcx,qword ptr [rsp+130h]
  | mov     r11,7FFC140F0348h
  | mov     rax,7FFC140F0348h
  | cmp     dword ptr [rcx],ecx
  | call    qword ptr [rax]
  | }         ^
M01_L02 |  
  | nop
```

benaadams:
> What is the performance impact of keeping a single type rather than splitting it into encoding specific writers (JsonWriterUtf8/JsonWriterUtf16)? 

Is also asm code gen to consider; every `TBufferWriter` type will generate asm for every instance method for both Utf16 and Utf8; even if only Utf8 is ever used

benaadams:
Raised issue "Inlined Struct wrapper .ctors are non-zero cost" https://github.com/dotnet/coreclr/issues/18542

ahsonkhan:
@benaadams, if you plan to merge these changes in (rather than using it to investigate design and performance trade-offs), could you merge back the Utf-8/Utf-16 types? I think we should evaluate that separately, especially since we still need to decide if we even need a UTF-16 specific type (or if the Reader/Writer could be UTF-8 only).

benaadams:
> could you merge back the Utf-8/Utf-16 types?

Sure, but having them in the same type is really ugly for the asm; pretty print is similar.

For the fastest version you'd probably want to copy the src then delete the encoding you don't want and delete all the pretty print branches

ahsonkhan:
@benaadams, I checked out your branch and if I change ArrayFormatter to a struct, the tests (as they are written) fail (due to struct copy). See my comment here: https://github.com/dotnet/corefxlab/pull/2369#issuecomment-399232955
Is it feasible to have a struct-based implementation of `IBufferWriter<byte>`? If not, what is the advantage of making the JsonWriter generic to try to avoid boxing when such a scenario doesn't exist (or isn't feasible).

benaadams:
Let me get back to you...

benaadams:
> Is it feasible to have a struct-based implementation of IBufferWriter<byte>? 

`StreamFormatter` and `BufferWriterFormatter<TOutput>` are both stucts and `IBufferWriter<byte>` so yes.

Shame this isn't valid
```csharp
public ref struct JsonWriterUtf8<TBufferWriter> where TBufferWriter : IBufferWriter<byte>
{
    private ref BufferWriter<TBufferWriter> _bufferWriter;

    public JsonWriterUtf8(ref BufferWriter<TBufferWriter> bufferWriter, bool prettyPrint = false)
    {
        ref _bufferWriter = ref bufferWriter;
```

benaadams:
Mostly they are struct converters for class types. Additional use is a struct wrapper to force a non-shared instantiation and avoid interface calls (whereas `IBufferWriter<byte>` direct forces interface calls though shared interface dispatch stubs)

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -7,6 +7,8 @@
 
 namespace System.Text.JsonLab
 {
+
+    [StructLayout(LayoutKind.Auto)]
```
ahsonkhan: What is the benefit of adding this attribute?

benaadams: Rearranges and packs the fields in the most efficient layout; rather than the order you define them (except child structs, which are on pointer boundaries).

I'd always assumed it was the default layout as it is for classes; and you had to specify  `[StructLayout(LayoutKind.Sequential)]` if you wanted that but seem to not always be the case https://blogs.msdn.microsoft.com/seteplia/2017/09/21/managed-object-internals-part-4-fields-layout/

Not sure it would make much difference; due to the object reference (forces Auto); but was trying to find where the drop was (think its now the issue I opened)

src/System.Text.JsonLab/System/Text/Json/JsonThrowHelper.cs
```diff
@@ -19,6 +19,17 @@ private static ArgumentException GetArgumentException(string message)
             return new ArgumentException(message);
         }
 
+        public static void ThrowArgumentExceptionInvalidUtf8String()
```
ahsonkhan: Why is this beneficial? Isn't `ThrowArgumentException(string message)` good enough?

benaadams: The il and asm are larger as it has to load the string reference before calling the Throw helper

Via String
```asm
G_M53912_IG05:
       48B9683016EFC1010000 mov      rcx, 0x1C1EF163068
       488B09               mov      rcx, gword ptr [rcx]
       E81FFCFFFF           call     JsonThrowHelper:ThrowArgumentException(ref)
       CC                   int3   
```
Via Enum is smaller as its just passing a numeric const
```asm
G_M12337_IG05:
       33C9                 xor      ecx, ecx
       E8F2FBFFFF           call     JsonThrowHelper:ThrowArgumentException(int)
       CC                   int3     
```
No Params is just a call
```asm
G_M36770_IG05:
       E8D4FBFFFF           call     JsonThrowHelper:ThrowArgumentExceptionInvalidUtf8String()
       CC                   int3   
```

benaadams: e.g. if you are inlining it all over the place, or duplicating it due to generics, better to be smaller

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
ahsonkhan: I haven't observed any difference in benchmarks when marking the struct layout as auto versus leaving it as default.

benaadams: Yeah, from the article it seems if there is a GC ref then it will switch to auto; and I think a Span or ArrayFormatter will count as that, so it shouldn't make any difference. Was just hoping...

benaadams:
Will close this

PR #2370, opened 2018-06-21
CalculateStartAttributeBytesNeeded calc issue

benaadams:
/cc @ahsonkhan


ahsonkhan:
Yep, noticed that myself when working on the performance investigation. Thanks for the fix!

Can you also add a test for it please?

PR #2373, opened 2018-06-27
Gpio API progress

edgardozoppi:
Hello,

This is the first of many PRs with the progress I made so far regarding Gpio API design. It is still work in progress but I just want to share it so you can have the opportunity to take a look and start making comments and suggestions.

@wfurt, @pjanotti, @stephentoub, @tommcdon, @dotnet/corefxlab-contrib 

Thanks!


src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -0,0 +1,165 @@
+﻿using System.Diagnostics;
```
joshfree: All files need this at the top


// Copyright (c) Microsoft. All rights reserved.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -0,0 +1,165 @@
...
+
+        private static void BlinkingLED(GpioDriver driver)
+        {
+            using (driver)
+            {
+                var led = new GpioPin(driver, GpioScheme.BCM, 26, GpioPinMode.Output);
+
+                for (var i = 0; i < 5; ++i)
+                {
+                    led.Write(GpioPinValue.High);
```
joshfree: Just a thought, but you could probably also add a Toggle API to flip a pin's state.

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
```diff
@@ -0,0 +1,65 @@
...
+        High = 1
+    }
+
+    public enum GpioScheme
+    {
+        Board,
+        BCM
+    }
+
+    public class GpioPin
```
joshfree: we should consider if it's possible to drop the redundant Gpio prefix on all types or whether it's necessary to keep it due to conflicts with other well-known types or keywords.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -0,0 +1,165 @@
...
+        }
+
+        private static void ButtonLED(GpioDriver driver)
+        {
+            using (driver)
+            {
+                var button = new GpioPin(driver, GpioScheme.BCM, 18, GpioPinMode.Input);
+                var led = new GpioPin(driver, GpioScheme.BCM, 26, GpioPinMode.Output);
+
+                var watch = Stopwatch.StartNew();
```
ahsonkhan: nit: don't use var where you can't statically infer the type by code inspection alone

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -0,0 +1,165 @@
...
+
+                while (watch.Elapsed.TotalSeconds < 15)
+                {
+                    var value = button.Read();
```
ahsonkhan: nit: don't use var here

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
ahsonkhan: nit: consider renaming the folder to just Samples rather than System.Devices.Gpio.Samples.

So, it would be `src/System.Devices.Gpio/Samples/Program.cs`

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
```diff
@@ -0,0 +1,65 @@
...
+            set
+            {
+                Driver.SetPinMode(Number, value);
+            }
+        }
+
+        public GpioPin(GpioDriver driver, GpioScheme numberKind, int number, GpioPinMode mode)
+        {
+            Driver = driver;
+            Number = driver.ConvertPinNumber(number, numberKind, GpioScheme.BCM);
```
joshfree: Does this mean that "Number" is always the BCM pin scheme value?  If so that seems like a bug given the Mode is settable.

src/System.Devices.Gpio/System.Devices.Gpio/System.Devices.Gpio.csproj
```diff
@@ -0,0 +1,23 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\..\tools\common.props" />
+  <PropertyGroup>
+    <Description></Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <PackageTags>Command line parsing support corefxlab</PackageTags>
```
ahsonkhan: Update this to match the package.

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
```diff
@@ -0,0 +1,159 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace System.Devices.Gpio
+{
+    public class UnixDriver : GpioDriver
+    {
+        private const string GpioPath = "/sys/class/gpio";
+
+        private ISet<int> ExportedPins;
```
ahsonkhan: private fields should start with `_` and follow camel casing:
https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
```diff
@@ -0,0 +1,65 @@
...
+            {
+                return Driver.GetPinMode(Number);
+            }
+            set
+            {
+                Driver.SetPinMode(Number, value);
+            }
+        }
+
+        public GpioPin(GpioDriver driver, GpioScheme numberKind, int number, GpioPinMode mode)
```
joshfree: Can we come up with a better name than "numberKind"?  How about "mode". 

What's called GpioPinMode right now feels like System.IO.FileAccess enum (Read, ReadWrite, Write).  Perhaps this should be a PinAccess enum.

src/System.Devices.Gpio/System.Devices.Gpio/GpioDriver.cs
```diff
@@ -0,0 +1,24 @@
+﻿using System.IO;
```
joshfree: Missing header

// Copyright (c) Microsoft. All rights reserved.

src/System.Devices.Gpio/publish.cmd
```diff
@@ -0,0 +1,23 @@
+@echo off
```
ahsonkhan: Why is the publish command within a batch script (.cmd) but the unpublish in a bash script (.sh)?

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
```diff
@@ -0,0 +1,159 @@
+﻿using System.Collections.Generic;
```
joshfree: Missing header

// Copyright (c) Microsoft. All rights reserved.

src/System.Devices.Gpio/publish_upload.txt
```diff
@@ -0,0 +1,3 @@
+cd ./Desktop
+put publish.zip
```
ahsonkhan: Did you intend to commit the publish.zip to the repo?

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
```diff
@@ -0,0 +1,488 @@
...
+
+        [StructLayout(LayoutKind.Explicit)]
+        private struct RegisterView
+        {
+            ///<summary>GPIO Function Select, 6x32 bits, R/W</summary>
+            [FieldOffset(0x00)]
+            public fixed UInt32 GPFSEL[6];
+
+            /////<summary>GPIO Function Select 0, 32 bits, R/W</summary>
+            //[FieldOffset(0x00)]
```
ahsonkhan: Why are all these commented out?

src/System.Devices.Gpio/publish_upload.txt
joshfree: please don't check in binaries / zips to git 😄

src/System.Devices.Gpio/System.Devices.Gpio/System.Devices.Gpio.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
+  </PropertyGroup>
+  <ItemGroup>
+    <Compile Include="System\**\*.*" />
+  </ItemGroup>
+  <ItemGroup>
+    <Compile Include="UnixDriver.cs" />
```
ahsonkhan: You shouldn't need to explicitly include all the files to compile. File globing should automatically grab everything within the System.Devices.Gpio folder.

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
```diff
@@ -300,10 +301,77 @@ public override void Dispose()
 
         public override void SetPinMode(int pin, GpioPinMode mode)
         {
-            if (pin < 0 || pin > 53) throw new ArgumentOutOfRangeException(nameof(pin));
-            if (mode < 0 || mode > GpioPinMode.Output) throw new ArgumentOutOfRangeException(nameof(mode));
+            if (pin < 0 || pin >= PinCount) throw new ArgumentOutOfRangeException(nameof(pin));
+            if (mode < 0 || mode > GpioPinMode.InputPullUp) throw new ArgumentOutOfRangeException(nameof(mode));
```
joshfree: this seems fragile.  Why don't you just add a default case to your switch below that throws for an unrecognized mode?

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
```diff
@@ -6,8 +6,10 @@ namespace System.Devices.Gpio
 {
     public enum GpioPinMode
     {
-        Input = 0,
-        Output = 1
+        Input,
```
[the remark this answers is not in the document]
joshfree: one more thing, can this be a flag instead such that writing "LightSensor.PinMode = GPIOPinMode.PullUp | GPIOPinMode.Input;" is possible?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -0,0 +1,165 @@
+﻿using System.Diagnostics;
+using System.Threading;
+
+namespace System.Devices.Gpio.Samples
+{
+    internal class Program
+    {
+        private static void Main(string[] args)
```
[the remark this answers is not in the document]
edgardozoppi: That project is just for me to test the API. It is not intended to be a console sample for users yet. Make sense to have samples, but since this code will change quite a lot in the future to follow API changes, I think is not worth to invest too much time on it for the moment.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -0,0 +1,165 @@
...
+                var option = -1;
+
+                if (args.Length > 0)
+                {
+                    option = Convert.ToInt32(args[0]);
+                }
+
+                switch (option)
+                {
+                    case 0: Unix_BlinkingLED(); break;
```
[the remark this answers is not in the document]
edgardozoppi: The Unix cases are using the GpioPin with UnixDriver (general driver using file interface).
The UnixDriver cases are using directly the UnixDriver (no GpioPin usage).
The RaspberryPi cases are using the GpioPin with RaspberryPiDriver (specific driver for that device).
The RaspberryPiDriver cases are using directly the RaspberryPiDriver (no GpioPin usage).

I would like to have the driver internal so users have to use GpioPin class, but also allow them to extend the API with their own drivers implementation some how. I am not sure how this will be possible yet.

src/System.Devices.Gpio/System.Devices.Gpio/Documentation/GPIORef.cs
```diff
@@ -104,8 +104,8 @@ public enum GPIOScheme
     public partial class GPIOPin
     {
         // Waiters
-        public bool ReadWait(TimeSpan timeout) { throw new NotImplementedException(); }
-        public int AnalogReadWait(TimeSpan timeout) { throw new NotImplementedException(); }
+        public void ReadWait(bool value, TimeSpan timeout) { throw new NotImplementedException(); }
```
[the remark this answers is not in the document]
edgardozoppi: That file is the original version created by Ian (I believe) to show how the API can look like. I just kept those files under Documentation folder for easy access but they are not compiled and are not part of the API. I will remove them later.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
edgardozoppi: Yes, that's a good idea. I think I will add that kind of useful and handy functionalities as extension methods, since they can be easily implemented using the already existing API.

edgardozoppi: Actually, I don't know if we can read the value of a pin that is configured as output. I will try it on Raspberry Pi, but even if it does maybe other devices work different.

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
edgardozoppi: Yes, it is always BCM number but the GpioPin class also has the GetNumber(GpioScheme kind) method to get the number for a different numbering scheme. Also, the user can specify the pin number using any numbering scheme in the constructor of this class.

An alternative could be having two peroperties: BcmNumber and BoardNumber. But I like more the other approach. Maybe we I should rename the Number property to be BcmNumber instead.

src/System.Devices.Gpio/publish.cmd
edgardozoppi: This is because the publish command is intended to run on the developers machine while the unpublish script has to run on the device itself.

src/System.Devices.Gpio/publish_upload.txt
edgardozoppi: No, I don't have the intention to commit the zip file to the repo.

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
edgardozoppi: Those are not needed because we are using the fixed array version of those registers. I just left them there as documentation but I can actually remove them.

src/System.Devices.Gpio/System.Devices.Gpio/System.Devices.Gpio.csproj
edgardozoppi: Thanks, I didn't do it myself. I think VS did it for me. It should be fixed in the next commit.

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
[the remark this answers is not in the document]
edgardozoppi: Yes, Input configures the pin to be just input with pull up/down disabled.

I thought about making those constants flags but since pull up/down options only make sense when they are combined with Input I decided that is more clear to do it without flags. That's also why I prefixed with Input both pull up/down constants.

Also, later on I will add PWM option (as well as SPI, I2C and other more advanced pin modes) and I thought it will be easier to understand if I prefix that constant name with Output, like OutputPWM. Other wise the user will need to know which flags can be combined with each other to generate valid modes. For example, GPIOPinMode.PullUp | GPIOPinMode.Output doesn't make any sense (I believe), nor GPIOPinMode.PWM | GPIOPinMode.Input.

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
edgardozoppi: Yes, I like that much more.

.gitignore
```diff
@@ -266,3 +266,6 @@ binaries/
 
 # Do not exclude any external files
 !external/**/*
+
+# Do not include these files
+/src/System.Devices.Gpio/publish.zip
```
ahsonkhan: I would change this to any zip file, i.e.
*.zip

src/System.Devices.Gpio/publish.cmd
pjanotti: The publish can come from a Linux box too, so a .sh is also welcome.

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
```diff
@@ -0,0 +1,80 @@
...
+    {
+        Low,
+        High,
+        SyncFallingEdge,
+        SyncRisingEdge,
+        AsyncFallingEdge,
+        AsyncRisingEdge
+    }
+
+    public enum GpioNumberingScheme
```
pjanotti: `PinNumberingScheme` perhaps?

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
```diff
@@ -0,0 +1,80 @@
...
+
+    public class GpioPin
+    {
+        protected GpioDriver _driver;
+
+        public int Number { get; }
+
+        public GpioPinMode Mode
+        {
+            get
```
pjanotti: nit: use expression-bodied property accessors for this property, e.g.:
```c#
get => _driver.GetPinMode(Number);
set => _driver.SetPinMode(Number, value);
```

src/System.Devices.Gpio/System.Devices.Gpio/GpioDriver.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        public abstract void Output(int pin, GpioPinValue value);
+
+        public abstract GpioPinValue Input(int pin);
+
+        public abstract void ClearDetectedEvent(int pin);
+
+        public abstract bool EventWasDetected(int pin);
+
+        public abstract void SetEventDetection(int pin, GpioEventKind kind, bool enabled);
```
pjanotti: What `enabled` means in this context?

src/System.Devices.Gpio/System.Devices.Gpio/GpioExtensions.cs
```diff
@@ -0,0 +1,35 @@
...
+        }
+
+        public static void Toogle(this GpioPin pin)
+        {
+            GpioPinValue value = pin.Read();
+
+            switch (value)
+            {
+                case GpioPinValue.Low: value = GpioPinValue.High; break;
+                case GpioPinValue.High: value = GpioPinValue.Low; break;
```
pjanotti: nit: not the focus at this moment but the enum can be cast from an integer and the API likely should throw if the value is not expected

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
```diff
@@ -0,0 +1,182 @@
...
+        private ISet<int> _exportedPins;
+
+        public UnixDriver()
+        {
+            _exportedPins = new HashSet<int>();
+        }
+
+        public override void Dispose()
+        {
+            while (_exportedPins.Count > 0)
```
pjanotti: At this stage this is a nit, more like something to look ahead: I am not sure that a HashSet and ISet are good abstractions here, perhaps, BitArray works better in this context - or depending on the number of pins accessing flags. Using a BitArray can let you use foreach instead of relying that UnexportPin remove it from the collection.

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
```diff
@@ -0,0 +1,182 @@
...
+                UnexportPin(pin);
+            }
+        }
+
+        public override GpioPinMode GetPinMode(int pin)
+        {
+            ExportPin(pin);
+
+            var directionPath = $"{GpioPath}/gpio{pin}/direction";
+            var stringMode = File.ReadAllText(directionPath);
```
pjanotti: nit: the coding style that we use just uses `var` when the right side is something like `new T()` in this case and below we are explicit about the type (I know, I like `var` better, but consistency...)

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
```diff
@@ -0,0 +1,182 @@
...
+            ExportPin(pin);
+
+            var directionPath = $"{GpioPath}/gpio{pin}/direction";
+            var stringMode = ModeToStringMode(mode);
+            File.WriteAllText(directionPath, stringMode);
+        }
+
+        public override GpioPinValue Input(int pin)
+        {
+            var valuePath = $"{GpioPath}/gpio{pin}/value";
```
pjanotti: Why when getting/setting mode you ensure the export but not for Input/Output?

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
```diff
@@ -0,0 +1,182 @@
...
+            return number;
+        }
+
+        #region Private Methods
+
+        private void ExportPin(int pin)
+        {
+            var pinPath = $"{GpioPath}/gpio{pin}";
+
+            if (!Directory.Exists(pinPath))
```
pjanotti: I wonder if we should throw if it doesn't exist but it is present in _exportedPins, it seems to indicate some overlapping usage. OTOH we could simply assume if it is in _exportedPins it is good to go (I know something else can change it...). I am not sure about the best path here, we need to think about it.

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
```diff
@@ -0,0 +1,182 @@
...
+
+        private GpioPinMode StringModeToPinMode(string value)
+        {
+            GpioPinMode result;
+
+            switch (value)
+            {
+                case "in": result = GpioPinMode.Input; break;
+                case "out": result = GpioPinMode.Output; break;
+                default: throw new Exception("Invalid GPIO Pin mode");
```
pjanotti: nit: for later, we need some kind of Gpio exception.

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
pjanotti: Removing those make easy to read the sources.

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
```diff
@@ -0,0 +1,828 @@
...
+        public RaspberryPiDriver()
+        {
+            var fileDescriptor = open(GpioMemoryFilePath, FileOpenFlags.O_RDWR | FileOpenFlags.O_SYNC);
+
+            if (fileDescriptor < 0)
+            {
+                throw new IOException($"open error number: {Marshal.GetLastWin32Error()}");
+            }
+
+            //Console.WriteLine($"file descriptor = {fileDescriptor}");
```
pjanotti: The info from this type of debug helpers should be available via strace. Any problem to run it on the device? An alternative is to use the trace mechanisms from CoreFx or debug output if they are valuable only for debug runs.

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
```diff
@@ -0,0 +1,828 @@
...
+            switch (mode)
+            {
+                case GpioPinMode.Input:
+                case GpioPinMode.InputPullDown:
+                case GpioPinMode.InputPullUp:
+                    SetInputPullMode(pin, PinModeToGPPUD(mode));
+                    break;
+            }                
+
+            SetPinMode(pin, PinModeToGPFSEL(mode));
```
pjanotti: My ignorance here: do you really want to SetPinMode after SetInputPullMode?

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
```diff
@@ -0,0 +1,828 @@
...
+            //Console.WriteLine($"{nameof(RegisterView.GPPUD)} original register value = {register:X8}");
+
+            register &= ~0b11U;
+            register |= mode;
+
+            //Console.WriteLine($"{nameof(RegisterView.GPPUD)} new register value = {register:X8}");
+
+            *gppudPointer = register;
+
+            Thread.SpinWait(150);
```
pjanotti: A comment for this SpinWait and its value is a good idea here.

pjanotti:
A general thing that would make easier reading the code is to add more comments.

src/System.Devices.Gpio/System.Devices.Gpio/GpioDriver.cs
edgardozoppi: Means if the user wants to detect (enabled == true) or not (enabled == false) the event kind passed in kind argument for that particular pin.

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
edgardozoppi: The problem is that for this driver implementation (UnixDriver) that is meant to be generic enough to be used in all Unix based OS among different devices, I don't know if there is a way to get how many GPIO pins the device has, so I cannot use the BitArray if I don't know how many bits I will need.

Maybe the pin count should be a parameter the user will have to provide in the UnixDriver constructor.

src/System.Devices.Gpio/System.Devices.Gpio/UnixDriver.cs
edgardozoppi: Ok, but for me is dificult to understand why C# designers added a feature (in this case var) if it is considered bad coding style haha. Do you know if there is some kind of Roslyn analyzer or similar that checks and warn (or even better, fix) my code to meet the coding style? That will be nice!

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
edgardozoppi: I cannot debug in the device.

src/System.Devices.Gpio/System.Devices.Gpio/RaspberryPiDriver.cs
edgardozoppi: Yes, because SetInputPullMode will just set the pull up/down mode and we still need to set the input/output mode using SetPinMode.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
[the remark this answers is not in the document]
joshfree: perhaps rename Unix to "UnixGeneric" or something then.  It's not obvious from the names how they're different.

src/System.Devices.Gpio/System.Devices.Gpio/Documentation/GPIORef.cs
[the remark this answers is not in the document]
joshfree: >> just kept those files under Documentation folder

OK, then should you revert your changes to this file then?

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
joshfree: or Mode or PinMode

src/System.Devices.Gpio/System.Devices.Gpio/GpioDriver.cs
joshfree: This isn't how C# users work with events.  Please see the observer design pattern https://docs.microsoft.com/en-us/dotnet/standard/events/ and examples of it on SerialPort https://docs.microsoft.com/en-us/dotnet/api/system.io.ports.serialport.datareceived?view=netframework-4.7.2

edgardozoppi: I think there is a confusion going on regarding names. The specification document of the BCM2835 chipset talk about events but they are actually not what you might be thinking. These are not similar to C# events, they are just bits that are set when a particular event occur. I tried to keep these names in the API design so people will know what this is about. C# events will be also available later on using interrupts.

src/System.Devices.Gpio/System.Devices.Gpio/Documentation/GPIORef.cs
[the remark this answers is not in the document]
edgardozoppi: OK

src/System.Devices.Gpio/publish.cmd
edgardozoppi: Those scripts and the Samples project are just for me, so I can easily try and test the API. They are not part of the API and are not intended to be used by users.

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
edgardozoppi: PinMode is already used for something else. Also I don't think the numbering scheme is a pin mode, it is just a way to identify a particular pin on the device. PinNumberingScheme makes more sense to me.

src/System.Devices.Gpio/System.Devices.Gpio/GPIOPin.cs
edgardozoppi: Gpio prefix removed.

PR #2378, opened 2018-06-29
Optimize JsonWriter APIs by using the TryWrite approach

ahsonkhan:
``` ini

BenchmarkDotNet=v0.10.14.534-nightly, OS=Windows 10.0.14393.2312 (1607/AnniversaryUpdate/Redstone1)
Intel Core i7-6700 CPU 3.40GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
Frequency=3328126 Hz, Resolution=300.4694 ns, Timer=TSC
.NET Core SDK=2.1.301
  [Host]     : .NET Core 2.1.1 (CoreCLR 4.6.26606.02, CoreFX 4.6.26606.05), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.1 (CoreCLR 4.6.26606.02, CoreFX 4.6.26606.05), 64bit RyuJIT


```
**BEFORE (master):**

|                         Method | Formatted | size |         Mean |       Error |      StdDev | Allocated |
|------------------------------- |---------- |----- |-------------:|------------:|------------:|----------:|
|  **WriterSystemTextJsonArrayOnly** |     **False** |    **1** |     **72.55 ns** |   **0.8945 ns** |   **0.7469 ns** |       **0 B** |
|  **WriterSystemTextJsonArrayOnly** |     **False** |   **10** |    **160.18 ns** |   **0.6331 ns** |   **0.5286 ns** |       **0 B** |
|  **WriterSystemTextJsonArrayOnly** |     **False** | **1000** | **12,708.05 ns** |  **76.4989 ns** |  **71.5571 ns** |       **0 B** |
|      **WriterSystemTextJsonBasic** |     **False** |    **?** |    **769.99 ns** |  **15.4300 ns** |  **27.0244 ns** |       **0 B** |
| WriterSystemTextJsonHelloWorld |     False |    ? |     93.48 ns |   1.9026 ns |   3.3819 ns |       0 B |
|  **WriterSystemTextJsonArrayOnly** |      **True** |    **1** |     **86.39 ns** |   **1.0231 ns** |   **0.9570 ns** |       **0 B** |
|  **WriterSystemTextJsonArrayOnly** |      **True** |   **10** |    **209.60 ns** |   **2.9768 ns** |   **2.7845 ns** |       **0 B** |
|  **WriterSystemTextJsonArrayOnly** |      **True** | **1000** | **16,923.45 ns** | **303.0954 ns** | **283.5156 ns** |       **0 B** |
[...]

PR #2379, opened 2018-06-30
Measure overhead of checking if escaping strings is required while writing to JsonWriter

ahsonkhan:
cc @GrabYourPitchforks, @benaadams, @Tornhoof 

The check whether to escape or not already has 30-80% overhead (and we still have some other characters that need to be escaped). There is definitely room for performance improvement here depending on how we want to structure our escaping. **Any suggestions or feedback on the best approach?**
- I am doing the check after encoding and copying to the output buffer rather than before and assuming having to escape is not a performance critical path.
  - Is that a valid assumption? Should we instead check while writing rather than checking ahead of time?
  - Should I process and escape the chars first (utf-16) and then encode/copy?
  - I see from the RFC that we need to escape `/` as well. Why?
  - Should we avoid using IndexOfAny? Use if or case statement instead?

For comparison, see results from master here - https://github.com/dotnet/corefxlab/pull/2378

|                         Method | Formatted | size |         Mean |       Error |      StdDev | Allocated |
|------------------------------- |---------- |----- |-------------:|------------:|------------:|----------:|
|  **WriterSystemTextJsonArrayOnly** |     **False** |    **1** |     **83.73 ns** |   **0.7384 ns** |   **0.6546 ns** |       **0 B** |
|  **WriterSystemTextJsonArrayOnly** |     **False** |   **10** |    **165.73 ns** |   **3.3113 ns** |   **3.4005 ns** |       **0 B** |
|  **WriterSystemTextJsonArrayOnly** |     **False** | **1000** | **11,351.73 ns** | **222.1803 ns** | **264.4898 ns** |       **0 B** |
|      **WriterSystemTextJsonBasic** |     **False** |    **?** |    **926.72 ns** |  **12.7365 ns** |  **11.9138 ns** |       **0 B** |
| WriterSystemTextJsonHelloWorld |     False |    ? |     92.22 ns |   1.8747 ns |   2.0837 ns |       0 B |


benaadams:
Looking at the Span code for `IndexOfAny` for 7 items will do 7 `IndexOf`s

From https://tools.ietf.org/html/rfc8259 errata
```
      string = quotation-mark *char quotation-mark

      char = unescaped /
          escape (
              %x22 /          ; "    quotation mark  U+0022
              %x5C /          ; \    reverse solidus U+005C
              %x2F /          ; /    solidus         U+002F
              %x62 /          ; b    backspace       U+0008
              %x66 /          ; f    form feed       U+000C
              %x6E /          ; n    line feed       U+000A
              %x72 /          ; r    carriage return U+000D
              %x74 /          ; t    tab             U+0009
              %x75 4HEXDIG )  ; uXXXX                U+XXXX

      escape = %x5C              ; \

      quotation-mark = %x22      ; "

      unescaped = %x20-21 / %x23-2E / %x30-5B / %x5D-10FFFF
```

So its `ch < 32 || ch == 34 || ch ==  47 || ch == 92`? Don't know if that's a great help though...

Could do an `IndexOfAny(span, 34, 47, 92)` which is vectorized and a custom vectorized `< 32` (or merge them both into one method?)

benaadams:
e.g.

```csharp
var vMatches = Vector.BitwiseOr(
                    Vector.BitwiseOr(
                        Vector.BitwiseOr(
                            Vector.Equals(vData, values0),
                            Vector.Equals(vData, values1)),
                        Vector.Equals(vData, values2)),
                    Vector.LessThan(vData, vSpaces));
```

Tornhoof:
Had to find my notes on this problems, I spent a few hours on this problem (and still do try a few things whenever I have time):
> So its ch < 32 || ch == 34 || ch == 47 || ch == 92

Yes that's correct, specifically the part < 32 is important, because quite a few serializers only implement those well-known escape symbols and not the full range, Utf8Json is one of the them.

> I see from the RFC that we need to escape / as well. Why?

Embedding JSON in html/xml, see https://stackoverflow.com/questions/1580647/json-why-are-forward-slashes-escaped for links.

> [...] having to escape is not a performance critical path.

I looked through a bit of production data on LOB apps I have access to, and found out the following:
- HATEOAS für REST APIs requires URLs to be encoded
- User comments are multiline
- Data for trees has slashes (stored in a db as json)
Bottom line of that is that for the tree app escaping in Json is actually a large bottleneck (and that app needs to reevaluate it's storage system). I imagine for e.g. Stack Overflow the comments might be more important.

> Should we avoid using IndexOfAny? Use if or case statement instead?

My evaluations produced fairly similar output in both performance for if/switch, I did not try IndexOfAny.
It is possible to do the scan with SSE fairly efficiently, RapidJson is using SSE to do it, I experimented with it a bit, but not much (I'm waiting for  ``PCMPISTR`` to be implemented for Whitespace checks)
https://github.com/Tencent/rapidjson/blob/7e68aa0a21b7800ec98133cb106e49bd6536e25c/include/rapidjson/writer.h#L597

In SpanJson I settled with the following (for now):
1. Scan through input string
2. If escaping is necessary
2. a Escape necessary: Utf8 encode and copy everything up to this char to the output, escape char and continue
2. b Escape not necessary: continue
3. Write encode and copy remaining input

https://github.com/Tornhoof/SpanJson/blob/663300d8f362f7a6571cd5de59338d1adccb9474/SpanJson/JsonWriter.Utf8.cs#L270 ( I noticed that the comments are not fully up to date).

The overhead is the now the loop and a check against the escapee list for each character, if there are no escapes to do. Multiple utf8 encodings and (possibly buffer resizes) if there are. 
In my benchmarks the UTF8 encoding process still is the main time sink, but your 30% overhead sounds about right.

benaadams:
Specifically I was referring to https://github.com/dotnet/corefx/blob/bcbf63568b81239ceb1048b5ad75e6c9f7eb58dc/src/Common/src/CoreLib/System/SpanHelpers.Byte.cs#L582-L586
```csharp
var vMatches = Vector.BitwiseOr(
                Vector.BitwiseOr(
            	    Vector.Equals(vData, values0),
            	    Vector.Equals(vData, values1)),
                Vector.Equals(vData, values2));
```
Which could have the addition of `Vector.LessThan` (in addition to the non-vector path)
```csharp
var vMatches = Vector.BitwiseOr(
                    Vector.BitwiseOr(
                        Vector.BitwiseOr(
                            Vector.Equals(vData, values0),
                            Vector.Equals(vData, values1)),
                        Vector.Equals(vData, values2)),
                    Vector.LessThan(vData, vSpaces));
```
To reduce it from 7 scans to 1 scan; to determine if it needed escaping

src/System.Text.JsonLab/System/Text/Json/JsonWriterHelper.cs
```diff
@@ -10,6 +10,11 @@ internal static class JsonWriterHelper
     {
         public static readonly byte[] NewLineUtf8 = Encoding.UTF8.GetBytes(Environment.NewLine);
 
+        // TODO: Add any other characters that need escaping, control characters 0x0 to 0x1F
```
joshfree: you should chat with @GrabYourPitchforks about escaping

tests/Benchmarks/System.Text.JsonLab/JsonWriterPerf.cs
```diff
@@ -76,32 +76,32 @@ public void WriterSystemTextJsonHelloWorld()
...
         public void WriterUtf8JsonHelloWorld()
         {
             WriterUtf8JsonHelloWorldHelper(_output);
         }
 
         [Benchmark]
         [Arguments(1)]
-        [Arguments(2)]
-        [Arguments(5)]
+        //[Arguments(2)]
```
joshfree: small nit: why not just delete these commented out cases

ahsonkhan:
5-10% improvement

image

PR #2385, opened 2018-07-07
Renamed EventKind to PinEvent

shaggygi:
EventKind seems reasonable, but there could be other future event types for things like SPI, I2C, UART, etc. that is mentioned in planning.  This PR renames pin event related code to more specifics.

cc: @edgardozoppi, @ianhays


joshfree:
Thanks. This seems reasonable

PR #2393, opened 2018-07-11
Tidy up BufferReader a bit.

JeremyKuhne:
- Reorder the properties slightly.
- Dial back var usage
- Change private MoveNext() to GetNextSegment()
- Make buffer argument for Peek in

I'm not sure why the constructor was public and the factory method doesn't do anything. For now I'm presuming we want the factory method and I've made the constructor private.

I'm looking at making the various BufferReaderExtensions methods into extension methods.

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -10,7 +10,7 @@ namespace System.Buffers.Reader
         private SequencePosition _currentSequencePosition;
         private SequencePosition _nextSequencePosition;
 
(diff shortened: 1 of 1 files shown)


davidfowl:
This is the latest [BufferReader](https://github.com/aspnet/Common/blob/d7a3ef5d126c2d767a26879db9ebc3d805b89014/shared/Microsoft.Extensions.Buffers.Sources/BufferReader.cs) impl that ASP.NET Core is currently using (just an FYI).

JeremyKuhne:
@pakrym / @davidfowl curious as to why there are no checks for `IsSingleSegment`. Was that ever investigated as an optimization? Is it uncommon for ASP.NET to create these with single segments?

JeremyKuhne:
I've updated the PR to get the changes the ASP.NET has that this doesn't.

PR #2394, opened 2018-07-13
Benchmarks cleanup + Readme

adamsitnik:
@JeremyKuhne you have asked me yesterday how to run the benchmarks and then I realized that this repo lacks of README.md file for the benchmarks

1. I added the readme
2. I removed old xunit.performance stuff
3. I ported last remaining benchmarks to BenchmarkDotNet
4. I updated BDN to latest version to take advantage of latest features (including glob filtering from command line)

/cc @KrzysztofCwalina @ahsonkhan


tests/Benchmarks/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+2. Restore this project
+   
+   (`..\..\dotnetcli\dotnet.exe restore`)
+
+3a. Run the benchmarks in Release, choose one of the benchmarks when prompted
+   
+   (`..\..\dotnetcli\dotnet.exe run -c Release`)
+   
+3b. To run specific tests only, pass in the filter to the harness:
```
ahsonkhan: Nice!

tests/Benchmarks/README.md
```diff
@@ -0,0 +1,32 @@
...
+
+**Pre-requisite:** The dotnet cli is available at the root in the dotnetcli directory. On a clean repo, build.cmd at the root installs the latest available dotnet.exe.
+
+**Pre-requisite:** To use dotnet cli from the root directory remember to set `DOTNET_MULTILEVEL_LOOKUP` environment variable to `0`!
+
+    $env:DOTNET_MULTILEVEL_LOOKUP=0
+
+1. Navigate to the benchmarks directory (corefxlab\tests\Benchmarks\)
+
+2. Restore this project
```
ahsonkhan: Is this step necessary?

tests/Benchmarks/System.IO.Pipelines/E2E.cs
```diff
@@ -8,11 +8,9 @@
 using System.Text.Formatting;
 using System.Text.JsonLab;
 using BenchmarkDotNet.Attributes;
-using BenchmarkDotNet.Engines;
 
 namespace System.IO.Pipelines.Benchmarks
 {
-    [SimpleJob(RunStrategy.Monitoring)] // the setting for benchmarks with not-steady state
```
ahsonkhan: What changed that allows us to remove this attribute?

tests/Benchmarks/System.Text.Primitives/PrimitiveParserBoolPerfTests.cs
```diff
@@ -0,0 +1,29 @@
...
+using System.Buffers.Text;
+using System.Collections.Generic;
+using BenchmarkDotNet.Attributes;
+using Benchmarks.System.Text.Primitives;
+
+namespace System.Text.Primitives.Benchmarks
+{
+    public partial class PrimitiveParserPerfTests
+    {
+        [Benchmark]
```
ahsonkhan: Can we mark this benchmark as baseline?

tests/Benchmarks/System.Text.Primitives/PrimitiveParserBoolPerfTests.cs
```diff
@@ -0,0 +1,29 @@
...
+        [Arguments("True")]
+        [Arguments("False")]
+        public void BaselineStringToBool(string text) => bool.TryParse(text, out _);
+
+        [Benchmark]
+        [ArgumentsSource(nameof(Utf8ByteArrays))]
+        public void PrimitiveParserByteSpanToBool(Utf8ByteArrayArgument text) => Utf8Parser.TryParse(text.CreateSpan(), out bool value, out int bytesConsumed);
```
ahsonkhan: Anyway to get rid of the overhead for creating the span? Does it have much affect on the end results of these benchmarks, given parsing primitives like bool/ints is quire fast?

tests/Benchmarks/System.Text.Primitives/PrimitiveParserInt32PerfTests.cs
```diff
@@ -0,0 +1,89 @@
...
+            "2147",
+            "-2147",
+            "-214748",
+            "-2147483",
+            "214748",
+            "-2147483648",
+            "2147483647",
+            "21",
+            "2147483",
+            "-214"
```
ahsonkhan: I think it would be worthwhile to add test cases like these, no?
"000000000000000000001235abcdfg"
"2abcdefghijklmnop"

ahsonkhan:
> 1. I added the readme
> 2. I removed old xunit.performance stuff
> 3. I ported last remaining benchmarks to BenchmarkDotNet
> 4. I updated BDN to latest version to take advantage of latest features (including glob filtering from command line)

Awesome! Thanks for taking the time to do this @adamsitnik.

FYI, @L-Dogg

ahsonkhan:
> I realized that this repo lacks of README.md file for the benchmarks

We should update this section at the root of the repo to point to the README you added:
https://github.com/dotnet/corefxlab#measuring-performance

tests/Benchmarks/README.md
```diff
@@ -0,0 +1,32 @@
...
+3a. Run the benchmarks in Release, choose one of the benchmarks when prompted
+   
+   (`..\..\dotnetcli\dotnet.exe run -c Release`)
+   
+3b. To run specific tests only, pass in the filter to the harness:
+   
+   (`..\..\dotnetcli\dotnet.exe run -c Release -- --filter namespace*`)
+   (`..\..\dotnetcli\dotnet.exe run -c Release -- --filter *typeName*`)
+   (`..\..\dotnetcli\dotnet.exe run -c Release -- --filter *.methodName`)
+   (`..\..\dotnetcli\dotnet.exe run -c Release -- --filter namespace.typeName.methodName`)
```
ahsonkhan: nit: The steps are not getting formatted correctly when we view the README within the browser.

- The numbered list is not aligned correctly.
- The various filter commands all end up on the same line.

image

tests/Benchmarks/README.md
adamsitnik: No, I forgot that dotnet run does restore the project as well

tests/Benchmarks/System.IO.Pipelines/E2E.cs
adamsitnik: I made BDN smarter ;) Now when the benchmark is very time consuming it will run it once per iteration, not 16 times as it used to.

tests/Benchmarks/System.Text.Primitives/PrimitiveParserBoolPerfTests.cs
adamsitnik: there can be only 1 baseline benchmark per type so if you split `PrimitiveParserPerfTests` into multiple types then yes

tests/Benchmarks/System.Text.Primitives/PrimitiveParserBoolPerfTests.cs
adamsitnik: @ahsonkhan this is a very good question.

I believe that this particular parsing operation should include the cost of the Span creation. Why? Because even if we have some huge buffer of memory and we want to parse just a part of it using the Span API we need to create a Span. So we can't use it without creating a Span instance.

If we would like to amortize the cost of Span creation we could use the `OperationsPerInvoke` feature:

```cs
[Benchmark(OperationsPerInvoke = 10)]
[ArgumentsSource(nameof(Utf8ByteArrays))]
public void PrimitiveParserByteSpanToBool(Utf8ByteArrayArgument text)
{
     bool result = default;
     var span = text.CreateSpan();

     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 1
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 2
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 3
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 4
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 5
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 6
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 7
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 8
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 9 
     result = Utf8Parser.TryParse(span, out bool value, out int bytesConsumed); // 10

    return result;
}
```

PR #2398, opened 2018-07-18
Adding LCD sample and SPI support

edgardozoppi:
This PR includes support for SPI connections on devices running Unix based OS and a GPIO sample to display text on a LCD screen.


src/System.Devices.Gpio/System.Devices.Gpio/SpiConnectionSettings.cs
```diff
@@ -0,0 +1,33 @@
...
+        public int ChipSelectLine { get; set; }
+        public SpiMode Mode { get; set; }
+        public int DataBitLength { get; set; }
+        public int ClockFrequency { get; set; }
+
+        public SpiConnectionSettings(int chipSelectLine)
+        {
+            ChipSelectLine = chipSelectLine;
+            DataBitLength = 8;
+            ClockFrequency = 5 * 100 * 1000;
```
joshfree: don't use magic numbers.  Use a private const long with a meaningful name and comment this code.

Here's an example https://github.com/dotnet/corefx/blob/master/src/Common/src/CoreLib/System/TimeSpan.cs#L33-L45

src/System.Devices.Gpio/System.Devices.Gpio/UnixSpiDevice.cs
```diff
@@ -0,0 +1,200 @@
...
+        {
+            if (_deviceFileDescriptor >= 0) return;
+
+            SpiConnectionSettings settings = ConnectionSettings;
+            string devicePath = $"/dev/spidev0.{settings.ChipSelectLine}";
+            _deviceFileDescriptor = open(devicePath, FileOpenFlags.O_RDWR);
+
+            if (_deviceFileDescriptor < 0)
+            {
+                throw new GpioException($"open error number: {Marshal.GetLastWin32Error()}");
```
joshfree: This should be an IOException given it's about opening a file on disk

src/System.Devices.Gpio/System.Devices.Gpio.Samples/LiquidCrystal.cs
```diff
@@ -0,0 +1,460 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Devices.Gpio.Samples
+{
+    // Ported from Microsoft.IoT.Lightning Nuget package
```
joshfree: Please update the PR to include a THIRD-PARTY-NOTICES.TXT for the corefxlab root directory and add the License notice for Microsoft.IoT.Ligning to it.  See https://github.com/dotnet/corefx/blob/master/THIRD-PARTY-NOTICES.TXT as an example.

Please do not commit this PR, until I get a chance to review the PR again with this change included.

src/System.Devices.Gpio/System.Devices.Gpio/UnixSpiDevice.cs
```diff
@@ -138,6 +138,18 @@ private unsafe void Initialize()
             }
         }
 
+        public override void Read(byte[] buffer)
+        {
+            byte[] writeBuffer = new byte[buffer.Length];
```
joshfree: Allocations are not free.  We shouldn't be allocated unnecessarily or doing a R+W when only a "half duplex" operation was requested.

src/System.Devices.Gpio/System.Devices.Gpio/UnixSpiDevice.cs
edgardozoppi: Yes, but the user doesn't know about that because it is an implementation detail. I will change it anyway.

joshfree:
Thanks @edgardozoppi for making the changes requested.

PR #2400, opened 2018-07-24
Adding Bme280 class and samples using SPI and LCD

edgardozoppi:
BME280 sensor allows to read pressure, humidity and temperature data. It supports both SPI and I2C serial protocols. The later is not yet supported by the current implementation.

I have also added an integrated sample that displays the data read from the sensor to an LCD using both SPI and GPIO pins.

### DIFF
--- a/src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
+++ b/src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
@@ -1,0 +1,409 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Devices.Gpio.Samples
+{
+    public class Bme280 : IDisposable
+    {
+        private const byte BME280_ADDRESS = 0x77; // Default I2C address
+
+        // Name of Registers used in the BME280
+
+        private const byte BME280_DIG_T1_REG = 0x88;
+        private const byte BME280_DIG_T2_REG = 0x8A;
+        private const byte BME280_DIG_T3_REG = 0x8C;
+        private const byte BME280_DIG_P1_REG = 0x8E;
+        private const byte BME280_DIG_P2_REG = 0x90;
+        private const byte BME280_DIG_P3_REG = 0x92;
+        private const byte BME280_DIG_P4_REG = 0x94;
+        private const byte BME280_DIG_P5_REG = 0x96;
+        private const byte BME280_DIG_P6_REG = 0x98;
+        private const byte BME280_DIG_P7_REG = 0x9A;
+        private const byte BME280_DIG_P8_REG = 0x9C;
+        private const byte BME280_DIG_P9_REG = 0x9E;
+
+        private const byte BME280_DIG_H1_REG = 0xA1;
+        private const byte BME280_DIG_H2_REG = 0xE1;
+        private const byte BME280_DIG_H3_REG = 0xE3;
+        private const byte BME280_DIG_H4_REG = 0xE4;
+        private const byte BME280_DIG_H5_REG = 0xE5;
+        private const byte BME280_DIG_H6_REG = 0xE7;
+
+        private const byte BME280_REGISTER_CHIPID = 0xD0;
+        private const byte BME280_REGISTER_VERSION = 0xD1;
+        private const byte BME280_REGISTER_SOFTRESET = 0xE0;
+        private const byte BME280_REGISTER_CAL26 = 0xE1;
+        private const byte BME280_REGISTER_CONTROLHUMID = 0xF2;
+        private const byte BME280_REGISTER_CONTROL = 0xF4;
+        private const byte BME280_REGISTER_CONFIG = 0xF5;
+        private const byte BME280_REGISTER_PRESSUREDATA = 0xF7;
+        private const byte BME280_REGISTER_TEMPDATA = 0xFA;
+        private const byte BME280_REGISTER_HUMIDDATA = 0xFD;
+
+        // Structure to hold the calibration data that is
+        // programmed into the sensor in the factory
+        // during manufacture
+        private struct Bme280_Calibration_Data
+        {
+            public ushort dig_T1;
+            public short dig_T2;
+            public short dig_T3;
+            public ushort dig_P1;
+            public short dig_P2;
+            public short dig_P3;
+            public short dig_P4;
+            public short dig_P5;
+            public short dig_P6;
+            public short dig_P7;
+            public short dig_P8;
+            public short dig_P9;
+            public byte dig_H1;
+            public short dig_H2;
+            public byte dig_H3;
+            public short dig_H4;
+            public short dig_H5;
+            public sbyte dig_H6;
+        };
+
+        private Bme280_Calibration_Data _cal_data;
+
+        private float _temperature;
+        private float _humidity;
+        private float _pressure;
+        private int _tFine;
+
+        private SpiConnectionSettings _spiSettings;
+        private SpiDevice _spiDevice;
+
+        private Pin _csPin;
+        private Pin _mosiPin;
+        private Pin _misoPin;
+        private Pin _sckPin;
+
+        /// <summary>
+        /// Use this for hardware SPI.
+        /// </summary>
+        public Bme280(Pin cs, SpiConnectionSettings spiSettings)
+        {
+            _csPin = cs;
+            _spiSettings = spiSettings;
+        }
+
+        /// <summary>
+        /// Use this for software SPI.
+        /// </summary>
+        public Bme280(Pin cs, Pin mosi, Pin miso, Pin sck)
+        {
+            _csPin = cs;
+            _mosiPin = mosi;
+            _misoPin = miso;
+            _sckPin = sck;
+        }
+
+        public void Dispose()
+        {
+            if (_spiDevice != null)
+            {
+                _spiDevice.Dispose();
+                _spiDevice = null;
+            }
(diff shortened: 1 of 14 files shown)


src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -98,21 +97,33 @@ private static void Main(string[] args)
...
                         Unix_ButtonWait();
                         break;
-
                     case 21:
                         RaspberryPi_ButtonWait();
                         break;
 
                     case 22:
-                        Unix_LiquidCrystal();
+                        Unix_LCD();
```
joshfree: Can you make this "Lcd" to be consistent with how you've cases Spi and Bme? 

https://msdn.microsoft.com/en-us/library/ms229043(v=vs.100).aspx

src/System.Devices.Gpio/System.Devices.Gpio.Samples/LiquidCrystal.cs
```diff
@@ -426,14 +426,14 @@ private void Write8bits(byte value)
         }
 
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private void DigitalWrite(Pin pin, int value)
+        private static void DigitalWrite(Pin pin, int value)
         {
             PinValue state = HasFlag(value, 0x01) ? PinValue.High : PinValue.Low;
```
joshfree: what is the magic number 0x01 in this context? can this be an enum or const?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
...
+
+        private Pin _csPin;
+        private Pin _mosiPin;
+        private Pin _misoPin;
+        private Pin _sckPin;
+
+        /// <summary>
+        /// Use this for hardware SPI.
+        /// </summary>
+        public Bme280(Pin cs, SpiConnectionSettings spiSettings)
```
joshfree: cs should be "chipSelectLine"

joshfree: Use the variable names from Windows.Devices.Spi as a starting point.

https://docs.microsoft.com/en-us/uwp/api/windows.devices.spi.spidevice

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
...
+        {
+            _csPin = cs;
+            _spiSettings = spiSettings;
+        }
+
+        /// <summary>
+        /// Use this for software SPI.
+        /// </summary>
+        public Bme280(Pin cs, Pin mosi, Pin miso, Pin sck)
```
joshfree: cs should be "chipSelectLine" 
sck should be "serialClock"
miso could be "masterInSlaveOut"
mosi could be "masterOutSlaveIn"

Can you think of better names for miso/mosi that are more approachable to people who don't know these concepts?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
...
+            Dispose();
+
+            _csPin.Mode = PinMode.Output;
+            DigitalWrite(_csPin, PinValue.High);
+
+            if (_spiSettings != null)
+            {
+                // Hardware SPI
+                _spiSettings.Mode = SpiMode.Mode0;
+                _spiSettings.ClockFrequency = 500 * 1000;
```
joshfree: "500 * 1000" could be written as 500_000;  Is the clock speed always 500KHz?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
+// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+
+using System.Runtime.CompilerServices;
+
+namespace System.Devices.Gpio.Samples
```
joshfree: For a sample, this is missing A LOT of comments ☹️ 

Could you draw inspiration from https://github.com/Microsoft/Windows-universal-samples/blob/master/Samples/IoT-I2C/cs/Scenario1_ReadData.xaml.cs#L42-L63 ?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
...
+            if (chipId != 0x60)
+            {
+                Console.WriteLine($"chipId = '{chipId:X2}'");
+                return false;
+            }
+
+            ReadSensorCoefficients();
+
+            // Set Humidity oversampling to 1
+            // Set before CONTROL (DS 5.4.3)
```
joshfree: Why do we want to set either of these values this way?  can you add more information?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
...
+            Write8(BME280_REGISTER_CONTROLHUMID, 0x01);
+            Write8(BME280_REGISTER_CONTROL, 0x3F);
+            return true;
+        }
+
+        private void ReadTemperature()
+        {
+            int var1, var2;
```
joshfree: can you think of a better name than "var1" and "var2"?  could this be "rawTemperatureReading" instead?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
...
+        }
+
+        private void ReadPressure()
+        {
+            long var1, var2, p;
+
+            uint adc_P = Read24(BME280_REGISTER_PRESSUREDATA);
+            adc_P >>= 4;
+
+            var1 = ((long)_tFine) - 128000;
```
joshfree: as a sample this should have better variable names and comments

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -0,0 +1,419 @@
...
+                    var writeBuffer = new byte[] { x };
+
+                    _spiDevice.TransferFullDuplex(writeBuffer, readBuffer);
+                }
+
+                result = readBuffer[0];
+            }
+            else
+            {
+                // Software SPI
```
joshfree: Have you tested the software SPI code path?  I wouldn't expect this to work reliably on a non real time OS

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Program.cs
edgardozoppi: Yes, sorry I didn't realized about this.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/LiquidCrystal.cs
edgardozoppi: It means just to set the pin to high. I will make a local constant for it.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
edgardozoppi: Ok, but I think people who will use this API or sample must have to understand what are those names because they need to connect the wires between the device and sensor and they use those names.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
edgardozoppi: I searched for it and it seems to support I2C up to 3.4MHz and SPI up to 10MHz. So maybe I should give the user the possibility to select the clock speed they want to use.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
edgardozoppi: They don't explain to much about it, they just say: "Changes to this register only became effective after a write operation to CONTROL register.". I will add this comment to the source code.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
edgardozoppi: Idem as above comment.

Those are the names used by the official algorithm in the sensor specification. Not sure what they mean.

https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BME280_DS001-12.pdf

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
edgardozoppi: Those are the names used by the official algorithm in the sensor specification. Not sure what they mean.

https://ae-bst.resource.bosch.com/media/_tech/media/datasheets/BST-BME280_DS001-12.pdf

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
edgardozoppi: Yes, and is not working at all. Don't know why, probably because of what you mentioned. Maybe I should remove that posibility from the code.

PR #2405, opened 2018-07-28
I2C support and samples using Bme280 sensor and LCD

edgardozoppi:
This PR adds support for I2C serial protocol.
It also extends the Bme280 sensor API to allow I2C in addition to SPI.

### DIFF
--- a/src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
+++ b/src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
@@ -67,6 +67,12 @@ namespace System.Devices.Gpio.Samples
             public sbyte dig_H6;
         };
 
+        private enum ConnectionProtocol
+        {
+            Spi,
+            I2c
+        }
+
         private Bme280_Calibration_Data _calibrationData;
 
         private float _temperature;
@@ -77,12 +83,24 @@ namespace System.Devices.Gpio.Samples
         private SpiConnectionSettings _spiSettings;
         private SpiDevice _spiDevice;
 
+        private I2cConnectionSettings _i2cSettings;
+        private I2cDevice _i2cDevice;
+
+        private readonly ConnectionProtocol _protocol;
         private Pin _csPin;
 
         public Bme280(Pin chipSelectLine, SpiConnectionSettings spiSettings)
         {
             _csPin = chipSelectLine ?? throw new ArgumentNullException(nameof(chipSelectLine));
             _spiSettings = spiSettings ?? throw new ArgumentNullException(nameof(spiSettings));
+            _protocol = ConnectionProtocol.Spi;
+        }
+
+        public Bme280(Pin chipSelectLine, I2cConnectionSettings i2cSettings)
+        {
+            _csPin = chipSelectLine ?? throw new ArgumentNullException(nameof(chipSelectLine));
+            _i2cSettings = i2cSettings ?? throw new ArgumentNullException(nameof(i2cSettings));
+            _protocol = ConnectionProtocol.I2c;
         }
 
         public void Dispose()
@@ -91,6 +109,12 @@ namespace System.Devices.Gpio.Samples
             {
                 _spiDevice.Dispose();
                 _spiDevice = null;
+            }
+
+            if (_i2cDevice != null)
+            {
+                _i2cDevice.Dispose();
+                _i2cDevice = null;
             }
         }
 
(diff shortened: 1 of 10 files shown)


src/System.Devices.Gpio/System.Devices.Gpio/UnixI2CDevice.cs
```diff
@@ -0,0 +1,349 @@
...
+        }
+
+        private unsafe void Initialize()
+        {
+            if (_deviceFileDescriptor >= 0)
+            {
+                return;
+            }
+
+            string devicePath = $"/dev/i2c-{_settings.BusId}";
```
joshfree: Can the "/dev/i2c-" be a public property setting?  In case a device has a different path

src/System.Devices.Gpio/System.Devices.Gpio/UnixI2CDevice.cs
```diff
@@ -0,0 +1,349 @@
...
+            }
+
+            if (_functionalities.HasFlag(I2cFunctionalityFlags.I2C_FUNC_I2C))
+            {
+                //Console.WriteLine("I2c functionality supported");
+
+                I2cMessageFlags flags = TransferKindToI2cMessageFlags(kind);
+                I2cTransfer(buffer, flags);
+            }
+            //else if (_functionalities.HasFlag(I2cFunctionalityFlags.I2C_FUNC_SMBUS_BLOCK_DATA))
```
joshfree: should this commented out code be removed?

src/System.Devices.Gpio/System.Devices.Gpio/UnixI2CDevice.cs
```diff
@@ -0,0 +1,349 @@
...
+        //    switch (kind)
+        //    {
+        //        case TrasnferKind.Read: return SmbusMessageFlags.I2C_SMBUS_WRITE;
+        //        case TrasnferKind.Write: return SmbusMessageFlags.I2C_SMBUS_READ;
+
+        //        default: throw new NotSupportedException();
+        //    }
+        //}
+
+        //private unsafe void SmbusTransfer(byte[] buffer, SmbusMessageFlags flags)
```
joshfree: should this commented out code be removed?

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
```diff
@@ -253,41 +288,62 @@ private void ReadSensorCoefficients()
...
 
-            byte result = readBuffer[0];
+            byte result = buffer[0];
             return result;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void Write(byte x)
+        {
+            var buffer = new byte[] { x };
```
joshfree: We shouldn't have to allocate on every Write call.  Is it possible to refactor the internals so you can avoid this new byte[]{x}?

src/System.Devices.Gpio/System.Devices.Gpio/UnixI2CDevice.cs
```diff
@@ -0,0 +1,349 @@
...
+            public SmbusTransactionType size;
+            public byte command;
+            public byte* data;
+        }
+
+        /// <summary>As specified in SMBus standard</summary>
+        private const int I2C_SMBUS_BLOCK_MAX = 32;
+
+        [StructLayout(LayoutKind.Explicit)]
+        private unsafe struct i2c_smbus_data
```
joshfree: which native header file did these struct definitions come from?  Could you add a comment pointing to the source so that future maintainers can find it more easily?

src/System.Devices.Gpio/System.Devices.Gpio/I2cConnectionSettings.cs
```diff
@@ -0,0 +1,47 @@
...
+        /// Gets or sets the I2c connection device address.
+        /// </summary>
+        public int DeviceAddress { get; set; }
+
+        /// <summary>
+        /// Initializes new instance of <see cref="I2cConnectionSettings"/>.
+        /// </summary>
+        /// <param name="busId">The I2c bus id on which the connection will be made.</param>
+        /// <param name="deviceAddress">The I2c connection device address.</param>
+        public I2cConnectionSettings(int busId, int deviceAddress)
```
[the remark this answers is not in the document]
edgardozoppi: No it can't as far as I know. I will change it to be of type uint instead.

src/System.Devices.Gpio/System.Devices.Gpio.Samples/Bme280.cs
edgardozoppi: Yes, I will change it to reuse always the same buffer.

PR #2411, opened 2018-08-01
 Remove use of unsafe code from JsonReader

ahsonkhan:
Roughly 20% regression.

**Before:**

|                               Method |   TestCase |          Mean |         Error |        StdDev | Allocated |
|------------------------------------- |----------- |--------------:|--------------:|--------------:|----------:|
| **ReaderSystemTextJsonLabSpanEmptyLoop** |  **BasicJson** |     **453.89 ns** |      **9.070 ns** |     **12.715 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** | **HelloWorld** |      **73.24 ns** |      **2.057 ns** |      **1.924 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |   **Json400B** |     **683.08 ns** |     **13.565 ns** |     **15.077 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |  **Json400KB** | **640,655.58 ns** | **12,779.191 ns** | **17,492.301 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |    **Json4KB** |   **6,198.38 ns** |     **38.388 ns** |     **32.056 ns** |       **0 B** |

**After:**

|                               Method |   TestCase |          Mean |          Error |         StdDev |        Median | Allocated |
|------------------------------------- |----------- |--------------:|---------------:|---------------:|--------------:|----------:|
| **ReaderSystemTextJsonLabSpanEmptyLoop** |  **BasicJson** |     **618.19 ns** |     **12.1167 ns** |     **19.2184 ns** |     **611.96 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** | **HelloWorld** |      **72.88 ns** |      **0.6524 ns** |      **0.6103 ns** |      **72.72 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |   **Json400B** |     **823.59 ns** |      **4.0467 ns** |      **3.5873 ns** |     **823.42 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |  **Json400KB** | **919,037.82 ns** | **18,193.0694 ns** | **42,165.2048 ns** | **904,594.38 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |    **Json4KB** |   **6,799.30 ns** |    **107.2116 ns** |     **89.5265 ns** |   **6,816.31 ns** |       **0 B** |

### DIFF
--- a/src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
@@ -167,7 +167,7 @@ namespace System.Text.JsonLab
                 return false;
             }
 
-            char ch = (char)_buffer[0];
+            byte ch = _buffer[0];
 
             switch (TokenType)
             {
@@ -203,14 +203,13 @@ namespace System.Text.JsonLab
                     break;
 
                 case JsonTokenType.PropertyName:
-                    if (ConsumeValueUtf8(ch) == 0) return false;
+                    ConsumeValueUtf8(ch);
                     break;
 
                 case JsonTokenType.EndArray:
                 case JsonTokenType.EndObject:
                 case JsonTokenType.Value:
-                    if (ConsumeNextUtf8(ch) == 0) return false;
-                    break;
+                    return ConsumeNextUtf8(ch);
             }
 
             return true;
@@ -302,19 +301,18 @@ namespace System.Text.JsonLab
         /// This method consumes the next token regardless of whether we are inside an object or an array.
         /// For an object, it reads the next property name token. For an array, it just reads the next value.
         /// </summary>
-        private int ConsumeNextUtf8(char marker)
+        private bool ConsumeNextUtf8(byte marker)
         {
-            _buffer = _buffer.Slice(1);
             switch (marker)
             {
-                case (char)JsonConstants.ListSeperator:
+                case JsonConstants.ListSeperator:
                     {
-                        SkipWhiteSpaceUtf8();
+                        SkipWhiteSpaceUtf8(1);
                         if (InObject)
                         {
                             if (_buffer.Length < 1)
                             {
-                                return 0;
+                                return false;
                             }
                             if (_buffer[0] != JsonConstants.Quote) JsonThrowHelper.ThrowJsonReaderException();
                             _buffer = _buffer.Slice(1);
@@ -324,33 +322,33 @@ namespace System.Text.JsonLab
                         {
                             if (_buffer.Length < 1)
                             {
-                                return 0;
+                                return false;
                             }
 
-                            char ch = (char)_buffer[0];
-                            return ConsumeValueUtf8(ch);
+                            ConsumeValueUtf8(_buffer[0]);
                         }
                         else
                         {
                             JsonThrowHelper.ThrowJsonReaderException();
-                            return default;
                         }
                     }
                     break;
 
-                case (char)JsonConstants.CloseBrace:
+                case JsonConstants.CloseBrace:
+                    _buffer = _buffer.Slice(1);
                     EndObject();
                     break;
 
-                case (char)JsonConstants.CloseBracket:
+                case JsonConstants.CloseBracket:
+                    _buffer = _buffer.Slice(1);
                     EndArray();
                     break;
 
                 default:
                     JsonThrowHelper.ThrowJsonReaderException();
-                    return default;
+                    break;
             }
-            return 1;
+            return true;
         }
 
         private int ConsumeValueUtf8MultiSegment(char marker)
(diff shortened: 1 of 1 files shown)


src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -161,24 +160,19 @@ private bool ReadMultiSegment()
...
-            ref byte bytes = ref MemoryMarshal.GetReference(_buffer);
-            int length = _buffer.Length;
-            int skip = SkipWhiteSpaceUtf8(ref bytes, length);
+            SkipWhiteSpaceUtf8();
 
-            ref byte next = ref Unsafe.Add(ref bytes, skip);
-            length -= skip;
-
-            if (length < 1)
+            if (_buffer.Length < 1)
```
benaadams: `(uint)_buffer.Length < 1` help skip bounds check?

ahsonkhan: I don't see any difference in disassembly. Which bounds check would get skipped?

image

benaadams: The access `byte ch = _buffer[0];`
```
cmp ebx, 0
jbe 00007FFB7...
```
Same with `Unsafe.Add(ref src, 0)` that have become `_buffer[0]` etc

Try greater than?
```csharp
if (1u > (uint)_buffer.Length)
{
    return false;
}
byte ch = buffer[0];
```
Otherwise perhaps it needs to be a local rather than field access?
```csharp
var buffer = _buffer;
if (1u > (uint)_buffer.Length)
{
    return false;
}
byte ch = buffer[0];
```

But the Jit doesn't remove the bounds check if its not `uint` and doesn't work that out even if its a constant https://github.com/dotnet/coreclr/issues/18934

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -534,20 +533,19 @@ private int ConsumeNaNUtf8MultiSegment()
...
+        private void ConsumeNaNUtf8()
         {
             Value = ReadOnlySpan<byte>.Empty;
             ValueType = JsonValueType.NaN;
 
-            if (length < 3
-                || Unsafe.Add(ref src, 0) != 'N'
-                || Unsafe.Add(ref src, 1) != 'a'
-                || Unsafe.Add(ref src, 2) != 'N')
+            if (_buffer.Length < 3
```
benaadams: e.g.
```csharp
if (3u > (uint)_buffer.Length
    || _buffer[0] != 'N'
    || _buffer[1] != 'a'
    || _buffer[2] != 'N')
{
    JsonThrowHelper.ThrowJsonReaderException();
}
_buffer = _buffer.Slice(3);
```

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
ahsonkhan: **Casting to uint or changing to greater than has no affect.**

Copying the span into a local gets rid of the extra bounds checks. However, if there is only one access, then the difference is between a struct copy vs. a bounds check (a wash).

https://sharplab.io/#v2:C4LghgzgtgPgrhAlgOwOYAIDKBPCwCmUA3ALABQADnAEYA2iAxugE74Bm6ezcDw6AwugDe5dGPQVmiAG5gC6AEr4wAEwDyyWtkwUwyADzVsBAHzoA+tThs2+ZqTLj0o8VTqN01APZfa6AHIAjAAUAJTCLk5iiBzBlta2zAB0ADL4aMAAFuj66ADM6DAwFlY2dgDaAAwAuugAvHXogYXF8WXM5YG1DegATC0lCRW93Y15oZFRIo5Rs6zAcMzI6GxgtBD4DrNiAL6TTvOLy8Dcm5N7M2Ju9EzevgG9YRGXUTHowcFwKMChbYmp6VQWRy+TgAz+FRq9UazSKg3anVGfXBpUS5RG0PyExeTmm2yihyWKzWGy22wu+MJx1OZN2k2uHjufn8eSeeO2SlUGi0Oj0hmM+DMqLs9XhiVpr1iwuSaQy2VyBTh0qqSNhxWVXUx/SVQw6GJ64324nZ+LEVOJ6zOOPEFO25pOcCtTltVxoN08PmZABY2UborE8mCzJ9vr9pQC5SjdSrMWqxRVNT1ta1lfqxtj8SbTebVpaJc6/Sx8AsiQ6nTbyDsgA===

If we have multiple span accesses, then it helps.
image

benaadams: > Copying the span into a local gets rid of the extra bounds checks. However, if there is only one access, then the difference is between a struct copy vs. a bounds check (a wash).

Its a shame the copy to local is needed to elide the bounds check. It makes sense with an array as the field reference may be changed to a different one by another thread (I'm assuming that's why its conservative); whereas the span is stack local so another thread won't be changing it so the Jit doesn't have to be as conservative? Might be completely wrong about this though... /cc @AndyAyersMS @mikedn

AndyAyersMS: Yes, when a field comes from a byref-like type then the jit should be able to assume the field is not cross-thread visible and not be quite so conservative.

mikedn: AFAIR I looked at a similar issue recently. Being stack only does not mean that other threads can't touch it, they may receive the address of the local by some means (e.g. via a static pointer typed variable). However, this seems impossible to do in a verifiable manner so the JIT could probably be more liberal with ref struct field accesses.

PR #2418, opened 2018-08-07
Add stream-based ctor to JsonReader

ahsonkhan:
This is mainly a convenience ctor. We can't have async APIs on JsonReader since its a ref struct so we are reading all the data from the stream and storing it within the span before processing it.

cc @davidfowl, @JeremyKuhne, @GrabYourPitchforks

### DIFF
--- a/src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
@@ -3,7 +3,8 @@
 
 using System.Buffers;
 using System.Buffers.Reader;
-using System.Runtime.CompilerServices;
+using System.IO;
+using System.Threading.Tasks;
 
 namespace System.Text.JsonLab
 {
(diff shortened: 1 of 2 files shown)


src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -83,6 +84,19 @@ public Utf8JsonReader(in ReadOnlySequence<byte> data)
             ValueType = JsonValueType.Unknown;
         }
 
+        public Utf8JsonReader(Stream stream)
+        {
+            _reader = default;
+            _isSingleSegment = true;
+            _buffer = ReadFromStreamAsync(stream).GetAwaiter().GetResult();
```
davidfowl: Why call any async methods?

Tornhoof:
Maybe add the special case handling for memorystream?
Like
```csharp
if (stream is MemoryStream ms && ms.TryGetBuffer(out var buffer))
```
to directly get the buffer from it?
You need to use the right memorystream ctor though.

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
Tornhoof: I would put stream handling one or more levels higher, and handle the copy via Arraypool. Memorystream is not really light on array copy. If you still want to do stream copy, then check for canseek first, so you may be able to get the length of the stream to create a buffer with the correct size.

davidfowl: I agree, it's strange to have the Stream API here.

ahsonkhan:
> I agree, it's strange to have the Stream API here.

Closing.

PR #2419, opened 2018-08-07
Create optimized TryRead<T>.

JeremyKuhne:
Move helpers to instance methods, split out slow paths and inline fast paths where needed.

Add some perf tests.

Overhead from implicit struct copies, method calls was significant. The best pattern I can come up with is to keep the fast (single span) reads in tight methods with callouts for aggregation. Changed the pattern of peek to return a direct span rather than a copy when it isn't necessary.

**Before:**

```
    Method |     Mean |    Error |   StdDev |
---------- |---------:|---------:|---------:|
 ReadInt32 | 328.9 us | 2.700 us | 2.525 us |
 PeekSpan  | 2.589 ms | 0.0125 ms | 0.0117 ms |
```

**After**

```
---------- |---------:|----------:|----------:|
 ReadInt32 | 102.1 us | 0.8000 us | 0.7483 us |
  PeekSpan | 240.5 us | 2.4859 us | 2.2037 us |
```

Roughly 3x / 10x improvement for binary reads. Parsing reads from 20% (split input) - 37% improvement in the benchmarks I added.


JeremyKuhne:
I also removed endianness in the binary read. Doesn't seem worth taking a hit for. If we really want it we can add slower overloads.

davidfowl:
If you're going to pick an endianess to support, you'd want big endian since most networking protocols are big endian.

davidfowl:
cc @mgravell @benaadams

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -2,74 +2,114 @@
...
+            }
+
+            return TryParseSlow(out value);
+        }
+
+        private unsafe bool TryParseSlow(out int value)
         {
-            var unread = reader.UnreadSegment;
-            if (Utf8Parser.TryParse(unread, out value, out int consumed))
+            const int MaxLength = 15;
```
ahsonkhan: What happens if the buffer contains leading zeros? Would MaxLength = 15 work in that case?

Also, where is 15 coming from?

tests/Benchmarks/Benchmarks.csproj
```diff
@@ -16,6 +16,11 @@
   <ItemGroup>
     <EmbeddedResource Include="System.Text.Primitives\SampleTexts\*.txt" />
   </ItemGroup>
+  <ItemGroup>
+    <Compile Include="..\System.Buffers.Primitives.Tests\BufferSegment.cs" Link="BufferSegment.cs" />
```
ahsonkhan: Include a project reference to the entire System.Buffers.Primitives.Tests instead, similar to other project references?

https://github.com/dotnet/corefxlab/pull/2419/files#diff-2d9cd050a80433427e0c8791e9b28ed5R51

tests/Benchmarks/System.Buffers/Reader.cs
```diff
@@ -45,7 +45,7 @@ public void ParseInt32BufferReader()
         {
             var reader = BufferReader.Create(s_ros);
 
-            while (BufferReaderExtensions.TryParse(ref reader, out int value))
+            while (reader.TryParse(out int value))
```
ahsonkhan: Why are we advancing by 1 each time? Shouldn't it be 4 since we have "1234 1234 1234 ..."?

tests/Benchmarks/System.Buffers/Reader_Binary.cs
```diff
@@ -0,0 +1,48 @@
...
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using System.Buffers.Reader;
+
+namespace System.Buffers.Benchmarks
+{
+    public class Reader_Binary
+    {
+        static byte[] s_array;
```
ahsonkhan: nit: add access modifiers explicitly

tests/Benchmarks/System.Buffers/Reader_Binary.cs
```diff
@@ -0,0 +1,48 @@
...
+            while (reader.TryRead(out int value))
+            {
+            }
+        }
+
+        [Benchmark]
+        public void PeekSpan()
+        {
+            const int Count = 1000;
+            const int Iterations = 100000;
```
ahsonkhan: Why are we manually setting up the iterations here? Is the objective here to measure reader.Peek and amortize the overhead of the setup of Span/BufferReader (which are ref structs and can't be saved as fields in global setup)?

cc @adamsitnik

tests/Benchmarks/System.Buffers/Reader_Binary.cs
```diff
@@ -0,0 +1,48 @@
...
+    public class Reader_Binary
+    {
+        static byte[] s_array;
+        static ReadOnlySequence<byte> s_ros;
+        static byte[] s_buffer;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            s_array = new byte[100000];
```
ahsonkhan: nit: use underscores as digit separators

tests/System.Buffers.Primitives.Tests/BufferUtilities.cs
```diff
@@ -8,28 +8,64 @@ namespace System.Buffers.Testing
 {
     public class BufferUtilities
     {
+        public static ReadOnlySequence<byte> CreateSplitBuffer(byte[] buffer, int minSize, int maxSize)
+        {
+            if (buffer == null || buffer.Length == 0 || minSize <= 0 || maxSize <= 0)
```
ahsonkhan: Also if `minSize > maxSize`?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -2,74 +2,114 @@
...
+    public ref partial struct BufferReader
     {
-        public static bool TryParse(ref BufferReader reader, out bool value) 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe bool TryParse(out bool value)
         {
-            var unread = reader.UnreadSegment;
-            if (Utf8Parser.TryParse(unread, out value, out int consumed))
+            var unread = UnreadSegment;
+            if (Utf8Parser.TryParse(unread, out value, out int consumed) && consumed < unread.Length)
```
ahsonkhan: Why do we need to compare `consumed` with `unread.Length` here? Isn't parsing bool known to be of constant length (i.e. either 4 or 5)? Shouldn't we return true if TryParse returns true, regardless?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -2,74 +2,114 @@
...
-    public static partial class BufferReaderExtensions 
+    public ref partial struct BufferReader
     {
-        public static bool TryParse(ref BufferReader reader, out bool value) 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public unsafe bool TryParse(out bool value)
         {
-            var unread = reader.UnreadSegment;
-            if (Utf8Parser.TryParse(unread, out value, out int consumed))
+            var unread = UnreadSegment;
```
ahsonkhan: nit: avoid using var here.

benaadams:
> Overhead from implicit struct copies, method calls was significant.

Can you add more detail/examples to "Inlined Struct wrapper .ctors are non-zero cost" https://github.com/dotnet/coreclr/issues/18542

tests/Benchmarks/System.Buffers/Reader_Binary.cs
adamsitnik: @ahsonkhan yes, you are right. However this benchmark was more "ported" from xunit than written in "BDN style"

benaadams:
> If you're going to pick an endianess to support, you'd want big endian since most networking protocols are big endian.

When using can always add a non-branching byteswap after (using `BitConverter.IsLittleEndian`  taken to a `readonly static` to change it to a Jit const); though probably also want @GrabYourPitchforks' "[WIP] Add support for BSWAP intrinsic" https://github.com/dotnet/coreclr/pull/18398

Custom protocols; IPC and file formats aren't necessarily big endian (e.g. why convert to big endian then back to little endian when 90% of the machines are little endian these days)

benaadams:
Ah, `BitConverter.IsLittleEndian` is an intrinsic now in CoreClr https://github.com/dotnet/coreclr/pull/9789

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: You have to be at the start- only the user can know what is relevant. 15 is the maximum size of the incoming string in the given format.

tests/Benchmarks/Benchmarks.csproj
JeremyKuhne: Tried that first, it didn't work out. We link in those files in multiple test projects. I should probably move the helpers to a shared project.

tests/Benchmarks/System.Buffers/Reader_Binary.cs
JeremyKuhne: Yes, I couldn't hold the reader as you mention.

tests/Benchmarks/System.Buffers/Reader_Binary.cs
JeremyKuhne: Yeah, sorry, copying an existing test. I'll fix that one too.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: I'll clean up the existing vars.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: For bool I can see dropping out. The others don't know if there is more valid data beyond a segment. I'll change this one and add a comment.

JeremyKuhne:
> Can you add more detail/examples to "Inlined Struct wrapper .ctors are non-zero cost" 

@benaadams I didn't look at the resulting IL, I just measured. Here are the numbers I saw (before breaking the slow path out of the main method).

```
Use static method in class that takes instance

    Method |     Mean |    Error |   StdDev |
---------- |---------:|---------:|---------:|
 ReadInt32 | 318.0 us | 5.955 us | 5.279 us |

Use instance method

    Method |     Mean |    Error |   StdDev |
---------- |---------:|---------:|---------:|
 ReadInt32 | 272.9 us | 5.203 us | 5.110 us |
```

tests/Benchmarks/System.Buffers/Reader_Binary.cs
ahsonkhan: @adamsitnik, do you plan to add support for ref struct fields in BDN?

JeremyKuhne:
I've added some endian sample methods `public unsafe bool TryReadInt32LittleEndian(out int value)`. If you're aligned with the proper platform it ends up being just about as fast:

```
----------------------- |---------:|----------:|----------:|
              ReadInt32 | 101.7 us | 0.5264 us | 0.4924 us |
    ReadInt32_BigEndian | 137.2 us | 1.0393 us | 0.9722 us |
 ReadInt32_LittleEndian | 103.5 us | 0.6634 us | 0.6206 us |
```


@mgravell, @davidfowl, @Drawaes, @benaadams Thoughts?

JeremyKuhne:
Still taking feedback of course- merging in the current state.

tests/Benchmarks/Benchmarks.csproj
ahsonkhan: > Tried that first, it didn't work out.

What ended up breaking?

Also, I tried to run the Reader_Binary tests locally, and i can't. I am getting the following errors for some reason. Have you come across this before @JeremyKuhne?
```text
// Validating benchmarks:
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=99, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf16toUtf8 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=999, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf16toUtf8 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=9999, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf16toUtf8 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=99, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf32toUtf16 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=999, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf32toUtf16 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=9999, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf32toUtf16 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=99, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf8toUtf16 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=999, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf8toUtf16 has 7
Only 1 benchmark method in a group can have "Baseline = true" applied to it, group DefaultJob-[Length=9999, CodePointInfo=Syste(...)Point [43]] in class EncodeFromUtf8toUtf16 has 7
```

Any ideas? cc @adamsitnik

tests/Benchmarks/System.Buffers/Reader_Binary.cs
adamsitnik: @ahsonkhan no, at least as of today. It would imply too many changes for our engine (today we derive a class from type with benchmarks, with structs it's impossible) +the stack only has too many limitations, example: it can't be used for `[MemberData]` because it requires to cast to object

```cs
 public IEnumerator<object[]> GetEnumerator()
    {
        yield return new object[] { 1, 2, 3 };
        yield return new object[] { -4, -6, -10 };
        yield return new object[] { -2, 2, 0 };
        yield return new object[] { int.MinValue, -1, int.MaxValue };
    }
```

we also can't use `Func<Span<T>>` because it can't be generic argument ;(

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
```diff
@@ -2,52 +2,167 @@
...
+            Advance(sizeof(T));
+            return true;
+        }
+
+        private unsafe bool TryReadSlow<T>(out T value) where T : unmanaged
+        {
+            Debug.Assert(UnreadSegment.Length < sizeof(T));
+
+            // Not enough data in the current segment, try to peek for the data we need.
+            byte* buffer = stackalloc byte[sizeof(T)];
```
davidfowl: Why isn't this directly using stackalloc Span?

JeremyKuhne: Because it won't let you pass it to Peek (for fear of being captured as the reader is a ref struct). Readonly methods should address some of this. I'm discussing with @jaredpar etc. about the possibilities of adding a feature that allows methods to take spans and specify that they won't stash the incoming span argument.

PR #2421, opened 2018-08-08
Moving Gpio samples project

edgardozoppi:
To avoid generating a MyGet package for the Sample console project.
Also includes circuit diagrams for all the samples we have so far.


src/System.Devices.Gpio/.editorconfig
```diff
@@ -1,160 +0,0 @@
...
-[*]
-insert_final_newline = true
-indent_style = space
-indent_size = 4
-
-[project.json]
-indent_size = 2
-
-# C# files
-[*.cs]
```
ahsonkhan: Consider adding such an editorconfig to the root of the repo (rather than removing it). It will help enforce some of the coding guidelines throughout the repo (like only use var when it's obvious what the variable type is).

edgardozoppi: There is an already existing .editorconfig file in the corefxlab root directory, next to the solution file. That's why I removed the copy under "src/System.Devices.Gpio" folder. I'm not sure where exactly that file has to be placed. Is enough to leave it in the same directory of the solution file o it has to be also next to each project file?

ahsonkhan: > Is enough to leave it in the same directory of the solution file

I believe so.
Just replace the contents of https://github.com/dotnet/corefxlab/blob/master/.editorconfig with what you had in your editorconfig that is being removed.

PR #2422, opened 2018-08-08
Optimize TryReadUntil

JeremyKuhne:
Move the methods to instance and optimize. I did not optimize the TryReadUntil that takes a Span delimiter yet.

For a single segment single delimiter I can get about 25% gain. The cost of comparing multiple delimiters eats a greater portion of the time pretty rapidly. I think were pretty close to optimal here (i.e. most of the time is in IndexOf/Any).

**Before (single segment)**
```
                            Method |       Mean |     Error |    StdDev |
---------------------------------- |-----------:|----------:|----------:|
      TryReadUntil_Span_OneSegment |   179.8 us |  3.220 us |  2.854 us |
 TryReadUntilAny_2_Span_OneSegment |   633.3 us |  6.364 us |  5.953 us |
 TryReadUntilAny_5_Span_OneSegment | 3,116.2 us | 16.157 us | 15.113 us |
```

**After (single segment)**
```
                            Method |       Mean |     Error |    StdDev |
---------------------------------- |-----------:|----------:|----------:|
      TryReadUntil_Span_OneSegment |   142.5 us |  2.651 us |  2.479 us |
 TryReadUntilAny_2_Span_OneSegment |   616.5 us |  6.070 us |  5.678 us |
 TryReadUntilAny_5_Span_OneSegment | 3,104.0 us | 51.189 us | 47.882 us |
```

For multiple segments I can get better improvements.

**Before (multiple segments)**
```
--------------------------- |---------:|----------:|----------:|
      TryReadUntil_Sequence | 2.559 ms | 0.0390 ms | 0.0365 ms |
 TryReadUntilAny_2_Sequence | 5.848 ms | 0.0439 ms | 0.0390 ms |
 TryReadUntilAny_5_Sequence | 6.537 ms | 0.0312 ms | 0.0292 ms |
          TryReadUntil_Span | 3.278 ms | 0.0149 ms | 0.0133 ms |
     TryReadUntilAny_2_Span | 6.348 ms | 0.0305 ms | 0.0285 ms |
     TryReadUntilAny_5_Span | 8.161 ms | 0.0414 ms | 0.0367 ms |
```

**After (multiple segments)**
```
--------------------------- |---------:|----------:|----------:|
      TryReadUntil_Sequence | 1.278 ms | 0.0051 ms | 0.0045 ms |
 TryReadUntilAny_2_Sequence | 2.185 ms | 0.0115 ms | 0.0089 ms |
 TryReadUntilAny_5_Sequence | 5.009 ms | 0.0486 ms | 0.0455 ms |
          TryReadUntil_Span | 1.817 ms | 0.0139 ms | 0.0130 ms |
     TryReadUntilAny_2_Span | 2.906 ms | 0.0104 ms | 0.0092 ms |
     TryReadUntilAny_5_Span | 5.548 ms | 0.0402 ms | 0.0376 ms |
```

cc: @benaadams, @mgravell, @Drawaes

### DIFF
--- a/tests/Benchmarks/System.Buffers/Reader_Search.cs
+++ b/tests/Benchmarks/System.Buffers/Reader_Search.cs
@@ -1,0 +1,115 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using System.Buffers.Reader;
+using System.Buffers.Testing;
+using System.Buffers.Text;
+
+namespace System.Buffers.Benchmarks
+{
+    public class Reader_Search
+    {
+        private static byte[] s_array;
+        private static ReadOnlySequence<byte> s_ros;
+        private static ReadOnlySequence<byte> s_rosSplit;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            s_array = new byte[1_000_000];
+            Random r = new Random(42);
+            r.NextBytes(s_array);
+            s_ros = new ReadOnlySequence<byte>(s_array);
+            s_rosSplit = BufferUtilities.CreateSplitBuffer(s_array, 10, 100);
+        }
+
+        [Benchmark]
+        public void TryReadUntil_Sequence()
+        {
+            BufferReader reader = BufferReader.Create(s_rosSplit);
+            while (reader.TryReadUntil(out ReadOnlySequence<byte> bytes, 42))
+            {
+            }
+        }
+
+        [Benchmark]
+        public void TryReadUntilAny_2_Sequence()
+        {
+            BufferReader reader = BufferReader.Create(s_rosSplit);
+            byte[] delimiters = { 0, 255 };
+            while (reader.TryReadUntilAny(out ReadOnlySequence<byte> bytes, delimiters))
+            {
+            }
+        }
+
+        [Benchmark]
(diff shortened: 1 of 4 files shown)


tests/Benchmarks/System.Buffers/Reader_Search.cs
```diff
@@ -0,0 +1,115 @@
...
+        private static byte[] s_array;
+        private static ReadOnlySequence<byte> s_ros;
+        private static ReadOnlySequence<byte> s_rosSplit;
+
+        [GlobalSetup]
+        public void Setup()
+        {
+            s_array = new byte[1_000_000];
+            Random r = new Random(42);
+            r.NextBytes(s_array);
```
ahsonkhan: To get deterministic performance numbers, does it make sense to filter out the delimiters we are looking for so that we measure the worst case performance of TryReadUntil?

Currently, it is hard to reason about how TryReadUntil_Sequence compares to TryReadUntilAny_2_Sequence/etc., since its possible that 42 is in segment 1, while 0, 255 are in segment N.

JeremyKuhne: > does it make sense to filter out the delimiters

As a separate test, sure.

> it is hard to reason about how TryReadUntil_Sequence 

I would hope that the large set of input data makes this easy to reason about- i.e. it distributes across a pretty wide set of cases. I'm splitting to something like 20,000 segments here.

JeremyKuhne:
I think I inadvertently broke the JSON reader by making the methods consistent. They weren't all moving past the delimiter. Investigating...

PR #2423, opened 2018-08-09
Constructor work & perf tweaks

JeremyKuhne:
Making the constructor public cuts the construction time by about a third.

Tweak UnreadSpan to inline and give back the current span if at the start.

Skip out on getting the next memory from the sequence if we're single segment.

Cuts 5-10% off of some of the perf tests.

Also change terms to Span from Segment. It was mentally taxing to juggle Span/Segment/Sequence/SequencePosition.

cc: @benaadams, @mgravell, @Drawaes, @davidfowl

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -10,40 +10,34 @@ namespace System.Buffers.Reader
         private SequencePosition _currentPosition;
         private SequencePosition _nextPosition;
 
+        /// <summary>
+        /// Create a <see cref="BufferReader" over the given <see cref="ReadOnlySequence{byte}"/>./>
+        /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private BufferReader(in ReadOnlySequence<byte> buffer)
+        public BufferReader(in ReadOnlySequence<byte> buffer)
         {
-            CurrentSegmentIndex = 0;
+            CurrentSpanIndex = 0;
             ConsumedBytes = 0;
             Sequence = buffer;
             _currentPosition = Sequence.Start;
             _nextPosition = _currentPosition;
 
-            if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<byte> memory, true))
+            if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<byte> memory, advance: true))
             {
                 End = false;
-                CurrentSegment = memory.Span;
-                if (CurrentSegment.Length == 0)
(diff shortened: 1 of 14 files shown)


tests/Benchmarks/System.Buffers/Reader_Construction.cs
```diff
@@ -0,0 +1,41 @@
...
+// See the LICENSE file in the project root for more information.
+
+using BenchmarkDotNet.Attributes;
+using System.Buffers.Reader;
+using System.Buffers.Testing;
+using System.Buffers.Text;
+
+namespace System.Buffers.Benchmarks
+{
+    public class Reader_Construction
```
ahsonkhan: Can you please post the benchmark results?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -10,42 +10,36 @@ namespace System.Buffers.Reader
...
         {
-            CurrentSegmentIndex = 0;
+            CurrentSpanIndex = 0;
             ConsumedBytes = 0;
             Sequence = buffer;
             _currentPosition = Sequence.Start;
             _nextPosition = _currentPosition;
 
-            if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<byte> memory, true))
+            if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<byte> memory, advance: true))
```
ahsonkhan: Special case IsSingleSegment to avoid the unnecessary call to ReadOnlySequence\<byte\>.TryGet?

```C#
if (buffer.IsSingleSegment)
{
    CurrentSpan = buffer.First.Span;
    End = CurrentSpan.Length == 0;
}
else if (buffer.TryGet(...))
{
...
}
else
{
...
}
```

JeremyKuhne:
Constructor results:

**Before**

```
----------------------- |---------:|----------:|----------:|
 ConstructSingleSegment | 37.66 ns | 0.6909 ns | 0.6125 ns |
  ConstructMultiSegment | 34.22 ns | 0.4174 ns | 0.3700 ns |
```

**After**

```
----------------------- |---------:|----------:|----------:|
 ConstructSingleSegment | 28.38 ns | 0.1411 ns | 0.1320 ns |
  ConstructMultiSegment | 26.55 ns | 0.2087 ns | 0.1952 ns |
```

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: Tanked multisegment- I'll look a bit more

PR #2424, opened 2018-08-09
Update JsonReader multi-segment code paths to follow changes made to the single-segment case

ahsonkhan:
- ~Added IsSingleSegment check in BufferReader ctor.~
- Added more JsonReader tests.


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -19,7 +19,12 @@ private BufferReader(in ReadOnlySequence<byte> buffer)
             _currentPosition = Sequence.Start;
             _nextPosition = _currentPosition;
 
-            if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<byte> memory, true))
+            if (buffer.IsSingleSegment)
```
JeremyKuhne: That hurt my measurements for multisegment. Hiccup perhaps?

tests/Benchmarks/System.Text.JsonLab/JsonReaderPerf.cs
```diff
@@ -174,7 +174,7 @@ public byte[] ReaderSystemTextJsonLabReturnBytes()
             return outputArray;
         }
 
-        [Benchmark]
+        //[Benchmark]
```
JeremyKuhne: Comment

PR #2425, opened 2018-08-09
 Remove use of BufferReader within JsonReader to measure overhead

ahsonkhan:
Attempting to use ReadOnlySequence\<byte\> directly to measure performance overhead of using BufferReader.

There is ~15% overhead of using BufferReader APIs rather than custom implementations and ReadOnlySequence\<byte\> directly.

``` ini

BenchmarkDotNet=v0.10.14.683-nightly, OS=Windows 10.0.17713
.NET Core SDK=3.0.100-alpha1-20180720-2
  [Host]     : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT
  DefaultJob : .NET Core 2.1.2 (CoreCLR 4.6.26628.05, CoreFX 4.6.26629.01), 64bit RyuJIT


```
**Before:**

|                                            Method |   TestCase |           Mean |        Error |       StdDev |         Median |  Gen 0 | Allocated |
|-------------------------------------------------- |----------- |---------------:|-------------:|-------------:|---------------:|-------:|----------:|
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |  **BasicJson** |     **1,149.8 ns** |    **22.911 ns** |    **47.315 ns** |     **1,121.5 ns** | **0.0076** |      **40 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** | **HelloWorld** |       **353.3 ns** |     **2.199 ns** |     **1.950 ns** |       **353.4 ns** | **0.0091** |      **40 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |   **Json400B** |     **1,500.5 ns** |    **10.395 ns** |     **9.215 ns** |     **1,501.6 ns** | **0.0114** |      **56 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |  **Json400KB** | **1,195,981.8 ns** | **7,753.480 ns** | **7,252.609 ns** | **1,197,271.0 ns** |      **-** |       **0 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |   **Json40KB** |   **118,374.7 ns** |   **449.618 ns** |   **420.573 ns** |   **118,243.9 ns** |      **-** |       **0 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |    **Json4KB** |    **10,846.9 ns** |    **95.483 ns** |    **79.733 ns** |    **10,830.6 ns** | **0.0916** |     **416 B** |

**After:**

|                                            Method |   TestCase |           Mean |         Error |        StdDev |         Median |  Gen 0 | Allocated |
|-------------------------------------------------- |----------- |---------------:|--------------:|--------------:|---------------:|-------:|----------:|
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |  **BasicJson** |       **984.7 ns** |     **19.435 ns** |     **38.364 ns** |       **969.8 ns** | **0.0076** |      **40 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** | **HelloWorld** |       **318.8 ns** |      **1.730 ns** |      **1.534 ns** |       **318.2 ns** | **0.0091** |      **40 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |   **Json400B** |     **1,284.3 ns** |     **11.059 ns** |     **10.344 ns** |     **1,281.8 ns** | **0.0114** |      **56 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |  **Json400KB** | **1,048,252.0 ns** | **12,956.988 ns** | **11,486.031 ns** | **1,047,165.8 ns** |      **-** |       **0 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |   **Json40KB** |   **102,040.0 ns** |  **1,581.665 ns** |  **1,320.762 ns** |   **101,601.2 ns** |      **-** |       **0 B** |
| **ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop** |    **Json4KB** |     **9,351.8 ns** |    **194.793 ns** |    **346.245 ns** |     **9,269.9 ns** | **0.0916** |     **416 B** |

cc @JeremyKuhne, @AndyAyersMS

### DIFF
--- a/src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
@@ -663,7 +663,6 @@ namespace System.Text.JsonLab
             }
         }
 
-
         public bool TryReadUntilAny(out ReadOnlySpan<byte> span, ReadOnlySpan<byte> delimiters)
         {
             ReadOnlySpan<byte> remaining = _buffer;
@@ -683,7 +682,7 @@ namespace System.Text.JsonLab
             ReadOnlySequence<byte> copy = _sequence;
             if (skip > 0)
                 Advance(skip);
-            
+
             while (_buffer.Length > 0)
             {
                 ReadOnlySpan<byte> remaining = _buffer;
@@ -893,7 +892,7 @@ namespace System.Text.JsonLab
             ReadOnlySequence<byte> copy = _sequence;
             if (skip > 0)
                 Advance(skip);
-            
+
             while (_buffer.Length > 0)
             {
                 ReadOnlySpan<byte> remaining = _buffer;
@@ -920,13 +919,13 @@ namespace System.Text.JsonLab
                         }
                     }
 
-                    Done:
+                Done:
                     ReadOnlySequence<byte> sequence = copy.Slice(copy.First.Span.Length - skip, index + skip);
                     span = sequence.IsSingleSegment ? sequence.First.Span : sequence.ToArray();
                     Advance(1);
                     return true;
                 }
-                KeepLooking:
+            KeepLooking:
                 Advance(remaining.Length);
             }
             span = default;


Tornhoof:
A few thoughts from my side, since I tried to tackle buffered reading/writing vs. reading/writing everything from only one buffer in SpanJson multiple times and below is the method, with which I came the closest to my goals (and still missed it)...
Let's assume a fairly well sized buffer, i.e. 4k for everything, nothing too small, everything above 1k is probably fine.
One of the possible goals is to get around large arrays, i.e LOH arrays, which is both important from a performance perspective and to keep memory in general in check. I will ignore the _it's more performant to start parsing/writing the json before I have all the data_ faction for now, as this will only apply for huge sizes and the parsing/writing speed of Json isn't the problem on modern systems.

**I imagine that most of these following is nothing new for you, but maybe it helps anyway**.

***What datatypes are affected?***
Simple answer: ``String``, a ``string`` has a dynamic length and we need the value from start to end.
This affects both json attribute names and the actual payloads.
I don't really know the maximum length of an c# identifier (theoretically it's apparently not limited), but Stack Overflow posts basically point to like 1000 chars and then Visual Studio will break.
For dictionaries I imagine that only a few rare cases if at all will serialize/deserialize arbitrary dictionaries with keys of such a size.
_Everything else_ will fit in less than ~50 bytes (decimal/double values).

Anyway, most of the time, a 4k buffer will be plenty enough for one string.

***What about arrays?***
Handle them higher up in the call chain and not in the Reader/Writer.

***What to do?***
In my experiments, always using ``ReadOnlySequence`` hits the performance quite hard, resizing the buffer only in the ``string`` case and simply continue resizing it until everything fits was way better. @davidfowl has some thoughts about actually resizing buffers in ``ReadOnlySequence`` over in corefx, if I remember correctly.
For all other cases, consider the buffer as a _window_ and slide it along.

***When to do it?***
****Reading****
``SkipWhitespace``, this method needs to be called before any parsing anyway, can move the window around (e.g. slide it further if the remaining length is < 50 bytes). Rare cases of purely empty buffers might apply here. Can be highly optimized with evil SSE/AVX code.
``ReadString``: Find the end, if it fails, slide the window to the beginning like above, if that fails again, resize buffer until it fits. Can be highly optimized with evil SSE/AVX code.
****Writing****
As soon as you need to grow your buffer, just flush the data to the underlying stream and set the offset to 0.

***Performance***
I enabled the code above only if the expected size of the json (SpanJson tracks the serialized/deserialized size of a type where possible) is above ~65k bytes (32k chars).
****Reading****
I was able to get within 3-5% of the _everything in buffer_ case with the above method (I'm still not happy with that, because it's technically a  performance regression in SpanJson).
****Writing****
It beats the classical grow array method even for medium jsons, well below the LOH size, at around 40k bytes.

davidfowl:
> In my experiments, always using ReadOnlySequence hits the performance quite hard, resizing the buffer only in the string case and simply continue resizing it until everything fits was way better. @davidfowl has some thoughts about actually resizing buffers in ReadOnlySequence over in corefx, if I remember correctly.
For all other cases, consider the buffer as a window and slide it along.

That's definitely the case if you don't use the `ReadOnlySequence<byte>` as a carrier type. It should only ever be slow when you end up crossing the boundary and depending on the segment size (usually 2K or 4K), that should be rare. The BufferReader's intent was to be that super light stack only abstraction that basically does the "expose a Span at a time" style parsing but it possibly isn't as light as it should be (as this proves).

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -92,35 +100,97 @@ public bool Read()
...
+            Advance(1);
+            return val;
+        }
+
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private void Advance(int byteCount)
+        {
+            if (byteCount < _buffer.Length)
+            {
+                _buffer = _buffer.Slice(byteCount);
```
davidfowl: I'm betting it would be even better if you stored the index instead of slicing every time.

ahsonkhan: I can try that and see if it has significant impact. Slicing the _buffer makes a lot of other operations faster since accesses can start at 0. I wanted to get an order-of-magnitude perf impact of using BufferReader, to see if using it is viable. I think 15% overhead is low enough to continue using it (and benefit from the convenience/less code) and try to improve that instead.

PR #2427, opened 2018-08-13
Point CLI/SDK auto-update at Core-SDK

dagood:
This is the corefxlab side of https://github.com/dotnet/core-eng/issues/3974: uptaking new master SDK versions after the CLI refactor. `core-sdk` is a "fake package" that will be put on dotnet/versions once https://github.com/dotnet/core-sdk/pull/50 is merged into core-sdk.

(This commit also fixes an issue where dependency names appear multiple times in the PR title and commit message due to multiple enumeration.)

### DIFF
--- a/scripts/update-dependencies/Program.cs
+++ b/scripts/update-dependencies/Program.cs
@@ -25,10 +25,11 @@ namespace Microsoft.Dotnet.Scripts
             if (!ParseArgs(args, out bool updateOnly))
                 return;
 
-            IEnumerable<IDependencyUpdater> updaters = GetUpdaters();
-            IEnumerable<DependencyBuildInfo> buildInfoDependencies = GetBuildInfoDependencies();
+            IEnumerable<IDependencyUpdater> updaters = GetUpdaters().ToArray();
+            IEnumerable<DependencyBuildInfo> buildInfoDependencies = GetBuildInfoDependencies().ToArray();
 
             DependencyUpdateResults updateResults = DependencyUpdateUtils.Update(updaters, buildInfoDependencies);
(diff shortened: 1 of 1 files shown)


scripts/update-dependencies/Program.cs
```diff
@@ -102,7 +103,7 @@ private static IEnumerable<IDependencyUpdater> GetUpdaters()
             yield return CreateRegexPropertyUpdater(config.DependencyFilePath, "SystemNumericsVectorsVersion", "System.Numerics.Vectors");
             yield return CreateRegexPropertyUpdater(config.DependencyFilePath, "SystemBuffersVersion", "System.Buffers");
             yield return CreateRegexPropertyUpdater(config.DependencyFilePath, "SystemIOPipelinesVersion", "System.IO.Pipelines");
-            yield return CreateFileUpdater(config.CLIVersionFilePath, "Microsoft.DotNet.Cli.Utils");
+            yield return CreateFileUpdater(config.CLIVersionFilePath, "core-sdk");
```
ahsonkhan: Should we rename this (and the file name) to something more appropriate now (something related to CoreSDK?
https://github.com/dotnet/corefxlab/blob/511aa547cf87eda83e2fc86f5e86132d73c44b18/scripts/update-dependencies/Config.cs#L37

dagood: Probably. However, after moving to Arcade, this file won't exist anymore, so it might not be worth the effort. (It's also still called DotnetCLIVersion.txt in CoreFX master, for what it's worth.)

ahsonkhan: Sounds good to me.

PR #2432, opened 2018-08-16
Make BufferReader generic

JeremyKuhne:
- Make BufferReader generic to allow usage with byte/char/etc.
- Read/Peek now return bool, and out T values (default (T) when at end)
- Move binary reader methods to extension methods
- Move parsing methods to extension methods (will add char variants)
- Rename "*Until" methods to "*To"
- Add "SkipTo" and "SkipPast" methods
- Improve perf of TryReadTo for sequence of delimiters (12x faster)
- Refactor basic functionality tests to run against byte & char


JeremyKuhne:
This change did not impact performance for existing APIs meaningfully.

cc: @danmosemsft, @davidfowl, @benaadams, @mgravell, @Drawaes

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -68,39 +68,42 @@ public BufferReader(in ReadOnlySequence<byte> buffer)
...
-        /// <returns>The next byte or -1 if at the end of the buffer.</returns>
+        /// <returns>The next T or default if at the end of the buffer.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public int Read()
+        public T Read()
         {
             if (End)
             {
-                return -1;
+                return default;
```
JanEggers: i think the method should be try read and return false instead of retuning default / -1

JeremyKuhne: I'm changing this to `bool Peek/Read(out T)`

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -68,46 +68,61 @@ public BufferReader(in ReadOnlySequence<byte> buffer)
...
         /// <summary>
-        /// Peeks at the next byte value without advancing the reader.
+        /// Peeks at the next value without advancing the reader.
         /// </summary>
-        /// <returns>The next byte or -1 if at the end of the buffer.</returns>
+        /// <param name="value">The next value or default if at the end.</param>
+        /// <returns>False if at the end of the reader.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public int Peek() => End ? -1 : CurrentSpan[CurrentSpanIndex];
+        public bool Peek(out T value)
```
ahsonkhan: TryPeek/TryRead?

JeremyKuhne: Hmm- probably should have been that way from the start. :)

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -167,38 +170,153 @@ public bool TryReadUntilAnyInternal(out ReadOnlySequence<byte> sequence, ReadOnl
...
                 return true;
             }
 
-            int matched = 0;
-            BufferReader copy = this;
-            SequencePosition start = Position;
-            SequencePosition end = Position;
+            BufferReader<T> copy = this;
+
+            T* b = stackalloc T[delimiter.Length];
```
ahsonkhan: We shouldn't stackalloc without bounding the user input to some known constant.

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -10,10 +10,14 @@ namespace System.Text.JsonLab
     {
         // We are using a ulong to represent our nested state, so we can only go 64 levels deep.
         private const int MaxDepth = sizeof(ulong) * 8;
+        private static byte[] s_nullBytes = new byte[4] { (byte)'n', (byte)'u', (byte)'l', (byte)'l' };
```
ahsonkhan: These aren't needed. We already have them defined within JsonConstants.
https://github.com/dotnet/corefxlab/blob/defd7e3ce7104b2fdd44d8bf5c542e2696ddb569/src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs#L39-L45

JeremyKuhne: Ah, missed that.

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -91,25 +95,18 @@ public bool Read()
             return _isSingleSegment ? ReadSingleSegment(ref _buffer) : ReadMultiSegment();
         }
 
-        private void SkipWhiteSpace()
+        private unsafe void SkipWhiteSpace()
```
ahsonkhan: Why is this unsafe?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: I'm not so sure on this one. As performance critical as this stuff is- who would want to match multiple KB? Do we support it and fall to the heap? Do we just fail past a certain point? I feel awkward penalizing everyone for an outlier, but maybe it doesn't really add up..

tests/Benchmarks/System.Buffers/Reader_Search.cs
```diff
@@ -22,92 +26,174 @@ public void Setup()
...
+            while (reader.TrySkipTo(42))
+            {
+            }
+        }
+
+        [Benchmark]
+        public unsafe void TrySkipToAny_AndSkipPastAny()
+        {
+            BufferReader<byte> reader = new BufferReader<byte>(s_rosSplit);
+            byte* b = stackalloc byte[] { 0x00, 0xFF };
```
ahsonkhan: Is this type of declaration required by the compiler or can we use safely defined spans here?

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
JeremyKuhne: It was, sorry. Fixing.

tests/System.Buffers.Primitives.Tests/BufferSegment.cs
```diff
@@ -3,13 +3,14 @@
 // See the LICENSE file in the project root for more information.
 
 using System.Diagnostics;
+using System.Runtime.InteropServices;
 using System.Text;
 
 namespace System.Buffers.Testing
 {
-    internal class BufferSegment : ReadOnlySequenceSegment<byte>
+    internal class BufferSegment<T> : ReadOnlySequenceSegment<T> where T : struct
```
ahsonkhan: What's the distinction between T : struct and T : unmanaged? I am assuming unmanaged is stricter and means T can't be a referencing containing struct. Should BufferReader\<T\> be constrained to T : struct? Can it be?

tests/Benchmarks/System.Buffers/Reader_Search.cs
JeremyKuhne: Required for now- I didn't want to make test specific statics.

tests/System.Buffers.Primitives.Tests/BufferSegment.cs
```diff
@@ -99,12 +100,17 @@ public override string ToString()
...
-            SpanLiteralExtensions.AppendAsLiteral(Memory.Span, builder);
-            return builder.ToString();
+            if (typeof(T) == typeof(byte))
+            {
+                var builder = new StringBuilder();
+                SpanLiteralExtensions.AppendAsLiteral(MemoryMarshal.AsBytes(Memory.Span), builder);
+                return builder.ToString();
+            }
+
+            return Memory.Span.ToString();
```
ahsonkhan: Just checking, this will output the right thing (i.e. the actual string contents) if T = char, correct?

tests/System.Buffers.Primitives.Tests/BufferSegment.cs
JeremyKuhne: Unmanaged is blittable. You can take sizeof(), pointers, etc. I figured starting there was better- might give flexibility.

tests/System.Buffers.Primitives.Tests/BufferSegment.cs
JeremyKuhne: Should- this was a convenience for the test code.

tests/System.Buffers.ReaderWriter.Tests/ReadableBufferFacts.cs
```diff
@@ -1,196 +0,0 @@
...
-
-        [Fact]
-        public void ReadableBufferGetPosition_DoesNotAlowNegative()
-        {
-            var buffer = Factory.CreateOfSize(20);
-            Assert.Throws<ArgumentOutOfRangeException>(() => buffer.GetPosition(-1));
-        }
-
-        [Fact]
-        public void SegmentStartIsConsideredInBoundsCheck()
```
ahsonkhan: Are we losing some test coverage here (including other tests in this file)?

JeremyKuhne: I assume we've moved these tests to CoreFX as they were all tests for ReadOnlySegment

JeremyKuhne: Er, ReadOnlySequence

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
ahsonkhan: I think it is a security/reliability concern. Whether we support it or not via allocating on the heap (either choice is feasible), we shouldn't be allocating an amount on the stack that is unbounded (based on user input). At least that is what was recommended to me.

cc @GrabYourPitchforks

JeremyKuhne: StackOverflow is worse than OutOfMemory I guess.

``` C#
Span<T> peekBuffer;
if (delimiter.Length * sizeof(T) < 512)
{
    T* t = stackalloc T[delimiter.Length];
    peekBuffer = new Span<T>(t, delimiter.Length);
}
else
{
    peekBuffer = new Span<T>(new T[delimiter.Length]);
}
```

JeremyKuhne: Note that I can't do the above without the unmanaged constraint.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -68,46 +68,61 @@ public BufferReader(in ReadOnlySequence<byte> buffer)
...
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get => CurrentSpanIndex == 0 ? CurrentSpan : CurrentSpan.Slice(CurrentSpanIndex);
         }
 
         /// <summary>
-        /// The total number of bytes processed by the reader.
+        /// The total number of {T}s processed by the reader.
         /// </summary>
-        public int ConsumedBytes { get; private set; }
+        public int ConsumedValues { get; private set; }
```
davidfowl: Consumed? Remove Values/Bytes/Things

pakrym:
>Read/Peek now return T, and default when at end (instead of int / -1)

So idea is you always have to use `.End` before operations?

tests/System.Buffers.ReaderWriter.Tests/ReadableBufferFacts.cs
pakrym: Yes, ROS tests can be removed.

JeremyKuhne:
> So idea is you always have to use .End before operations?

I changed it to `bool TryRead(out T)`. Updated the comment.

tests/System.Buffers.Primitives.Tests/BufferSegment.cs
davidfowl: We don't need the constraint if we don't use it anywhere.

davidfowl:
@JeremyKuhne are you planning to rewrite the Kestrel parser to see how convenient these API changes are?

tests/System.Buffers.Primitives.Tests/BufferSegment.cs
JeremyKuhne: I just added the first usage to BufferReader. I determine whether or not to allocate on the stack based on size(T).

JeremyKuhne:
> are you planning to rewrite the Kestrel parser to see how convenient these API changes are?

@davidfowl I've been primarily looking at interaction with the JsonReader so far. Where is the Kestrel parser? I was going to look more at System.Text.Http.Parser in CoreFXlab.

davidfowl:
> @davidfowl I've been primarily looking at interaction with the JsonReader so far. Where is the Kestrel parser? I was going to look more at System.Text.Http.Parser in CoreFXlab.

It's very similar. https://github.com/aspnet/KestrelHttpServer/blob/release/2.2/src/Kestrel.Core/Internal/Http/HttpParser.cs

mgravell:
If anyone wants more comparisons, I also have a fully span-ized redis ("RESP") parser and protobuf parser - both make use of home-crafted versions of something akin to BufferReader. One is *largely* text-based, the other is *largely* binary based. All fun.

Issue #2406, opened 2018-07-30
Proposal: Provide specialized collections aka Power Collections


danmosemsft:
@ViktorHofer commented on [Thu Jul 19 2018](https://github.com/dotnet/corefx/issues/31191)

## Proposal

We are hitting limitations with our in-built collections all over the places, either performance based (i.e. allocation) or feature based). After talking with @jkotas offline we agreed that it would be ideal if we would provide commonly used specialized collections (like a Dictionary<string, string> or a PriorityQueue) as a source based nuget package. That would allow us to provide basic enhanced functionality over the general purpose inbox types but still the ability to tweak them on demands.

Considerations:
- What kind of collection types would benefit from specialization? 
- Where do we see gaps to our competition (again, performance or feature wise)?
- Where would such a package be hosted and how would it be deployed? (corefx vs community corefx repo)
- Is a source based nuget package feasible and are we easily able to use it inside corefx?

## Example

Currently working in Regex to avoid allocation I stumbled upon places where Unicode categories are statically stored in a Dictionary<string, string>. Trying to get the value out of it involves allocation because there is currently no way to pass a ReadOnlySpan<char> as a key in to get the value. Providing a specialized version would allow us to add an additional overload for ReadOnlySpan<char>.

## References

Dictionary/List ref api additions: https://github.com/dotnet/coreclr/pull/17405#issuecomment-378627247
PriorityQueue: https://github.com/dotnet/corefx/issues/574#issuecomment-308491122 & https://github.com/dotnet/corefxlab/pull/1850
OrderedDictionary<,>: https://github.com/dotnet/corefx/issues/26634

cc @jkotas @karelz @benaadams @danmosemsft  

---

@danmosemsft commented on [Thu Jul 19 2018](https://github.com/dotnet/corefx/issues/31191#issuecomment-406339761)

Perhaps we could have a Dictionary implementation specialized for keys of type `int` -- that would be used by k-nucleotide (given they accept use of `Long2IntOpenHashMap` [by Java](https://benchmarksgame-team.pages.debian.net/benchmarksgame/program/knucleotide-java-1.html) 😕 ).

Why do you propose a source based package?

@stephentoub 

---

@danmosemsft commented on [Thu Jul 19 2018](https://github.com/dotnet/corefx/issues/31191#issuecomment-406340267)

cc @AnthonyLloyd for that.

---

[...]

PR #2439, opened 2018-08-21
Add Microsoft.Experimental.Collections package and move MultivalueDictionary to it

safern:
This is the beginning of a new package to experiment new collections before adding them to .NET Core.

Related to: https://github.com/dotnet/corefxlab/issues/2406

I tried to split it per commit to make it easier to review since I did code cleanup and some modifications for the source to be able to build.

cc: @ahsonkhan @danmosemsft @joshfree @terrajobst


src/Microsoft.Experimental.Collections/Microsoft.Experimental.Collections.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\tools\common.props" />
+  <PropertyGroup>
+    <Description>This package provides collections that are being considered for future versions of .NET Core.</Description>
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageTags>collections, system, microsoft, fastdictionary, multivaluedictionary</PackageTags>
```
danmosemsft: As I discovered recently, nuget splits tags on spaces. You should have no commas

src/Microsoft.Experimental.Collections/Microsoft.Experimental.Collections.csproj
```diff
@@ -0,0 +1,24 @@
+<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\tools\common.props" />
+  <PropertyGroup>
+    <Description>This package provides collections that are being considered for future versions of .NET Core.</Description>
```
danmosemsft: What about something like `This package contains some useful collections that are not currently included in .NET Core itself.`

danmosemsft: Oh, I see you are using the existing description. It's fine let's stick with that.

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
```diff
@@ -0,0 +1,176 @@
...
+                dict.ContainsKey(RandomKey); dict.ContainsKey(RandomKey); dict.ContainsKey(RandomKey);
+                dict.ContainsKey(RandomKey); dict.ContainsKey(RandomKey); dict.ContainsKey(RandomKey);
+                dict.ContainsKey(RandomKey); dict.ContainsKey(RandomKey); dict.ContainsKey(RandomKey);
+            }
+        }
+
+        [Benchmark]
+        public void ContainsValue()
+        {
+            for (int i = 0; i <= 20000; i++)
```
ahsonkhan: Is it necessary to have such a for loop (along with loop unrolling)? Also, with different iteration counts, how do we reason about/compare the perf results?

cc @adamsitnik

src/Microsoft.Experimental.Collections/Microsoft.Experimental.Collections.csproj
danmosemsft: I don't think anyone will search for "fastdictionary"

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
```diff
@@ -0,0 +1,176 @@
...
+        [Benchmark]
+        public void Clear()
+        {
+            dict.Clear();
+        }
+
+        [Benchmark]
+        public void Ctor()
+        {
+            for (int i = 0; i <= 20000; i++)
```
ahsonkhan: Couldn't some of these tests simply be the Ctor call with nothing else?

```C#
[Benchmark]
public void Ctor()
{
    var dict = new MultiValueDictionary<int, string>();
}
```

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
```diff
@@ -0,0 +1,176 @@
...
+            {
+                dict.Add(i * 10 + 1, 0); dict.Add(i * 10 + 2, 0); dict.Add(i * 10 + 3, 0);
+                dict.Add(i * 10 + 4, 0); dict.Add(i * 10 + 5, 0); dict.Add(i * 10 + 6, 0);
+                dict.Add(i * 10 + 7, 0); dict.Add(i * 10 + 8, 0); dict.Add(i * 10 + 9, 0);
+            }
+        }
+
+        public List<int> values;
```
ahsonkhan: nit: Should this be a private field? If not, use PascalCasing.

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
```diff
@@ -0,0 +1,176 @@
...
+    public class MultiValueDictionaryPerformanceTests
+    {
+        [Params(1000, 10000, 10000)]
+        public int Size { get; set; }
+
+        public int RandomKey { get; set; }
+
+        public MultiValueDictionary<int, int> dict;
+        
+        [GlobalSetup(Targets = new[] { nameof(Add), nameof(Clear), nameof(GetItem), nameof(GetKeys), nameof(ContainsValue) })]
```
ahsonkhan: Interesting use of GlobalSetup. I didn't know you could specify benchmark specific setups like this. Good to know :)

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
```diff
@@ -0,0 +1,176 @@
...
+using BenchmarkDotNet.Attributes;
+
+namespace Microsoft.Collections.Extensions.Tests
+{
+    public class MultiValueDictionaryPerformanceTests
+    {
+        [Params(1000, 10000, 10000)]
+        public int Size { get; set; }
+
+        public int RandomKey { get; set; }
```
ahsonkhan: In other benchmarks, we tend to use private fields rather than public properties. Any specific reason why you opted for properties instead?

src/Microsoft.Experimental.Collections/Microsoft.Experimental.Collections.csproj
```diff
@@ -0,0 +1,24 @@
...
+    <TargetFramework>netstandard2.0</TargetFramework>
+    <PackageTags>collections, system, microsoft, fastdictionary, multivaluedictionary</PackageTags>
+    <VersionPrefix>1.0.4</VersionPrefix>
```
ahsonkhan: Why are we adding a custom version prefix?

We have one coming from common.props:
https://github.com/dotnet/corefxlab/blob/3e2932bdb580538be39cccb9cc1fa306bb41cc02/tools/common.props#L6

src/Microsoft.Experimental.Collections/Microsoft.Experimental.Collections.csproj
safern: Ok, will change to spaces.

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
safern: I was going to use `[Params]` in this property and that is why I used public. I didn't know if when using `[Params]` it can be public, don't know how benchmark initializes the property with the specified parameters in the attribute.

However, I will update to make the dictionary, list and randomkey private 😄

src/Microsoft.Experimental.Collections/Microsoft.Experimental.Collections.csproj
safern: This package already exists in NuGet

https://www.nuget.org/packages/Microsoft.Experimental.Collections/1.0.3-alpha

So, unfortunately we need to have a custom version for this package unaligned for the other packages in this repo.

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
safern: Will change to be private.

src/Microsoft.Experimental.Collections/Microsoft.Experimental.Collections.csproj
safern: I did change the description, the base is the same, just changed .NET Framework -> .NET Core.

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
adamsitnik: `[Params]` requires it to be either a public property with public setter or public non-readonly field

safern:
I updated the perf tests and the description to contain the results.

@adamsitnik I'm getting this warnings:

```
// * Warnings *
MinIterationTime
  MultiValueDictionaryPerformanceTests.Add: InvocationCount=1, UnrollFactor=1           -> MinIterationTime = 1.2000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.AddRange: InvocationCount=1, UnrollFactor=1      -> MinIterationTime = 12.5000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.ContainsValue: InvocationCount=1, UnrollFactor=1 -> MinIterationTime = 13.9000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.Add: InvocationCount=1, UnrollFactor=1           -> MinIterationTime = 1.8000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.AddRange: InvocationCount=1, UnrollFactor=1      -> MinIterationTime = 97.3000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.ContainsValue: InvocationCount=1, UnrollFactor=1 -> MinIterationTime = 129.8000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.Add: InvocationCount=1, UnrollFactor=1           -> MinIterationTime = 3.9000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.AddRange: InvocationCount=1, UnrollFactor=1      -> MinIterationTime = 958.6000 us which is very small. It's recommended to increase it.
  MultiValueDictionaryPerformanceTests.ContainsValue: InvocationCount=1, UnrollFactor=1 -> MinIterationTime = 1.6391 ms which is very small. It's recommended to increase it.
MultimodalDistribution
  MultiValueDictionaryPerformanceTests.Ctor: Default                                    -> It seems that the distribution is bimodal (mValue = 3.36)
  MultiValueDictionaryPerformanceTests.Ctor_Size: Default                               -> It seems that the distribution can have several modes (mValue = 2.87)
  MultiValueDictionaryPerformanceTests.GetItem: Default                                 -> It seems that the distribution is bimodal (mValue = 3.79)
  MultiValueDictionaryPerformanceTests.GetKeys: Default                                 -> It seems that the distribution is bimodal (mValue = 3.84)
  MultiValueDictionaryPerformanceTests.AddRange: InvocationCount=1, UnrollFactor=1      -> It seems that the distribution is bimodal (mValue = 3.9)
  MultiValueDictionaryPerformanceTests.ContainsValue: InvocationCount=1, UnrollFactor=1 -> It seems that the distribution is multimodal (mValue = 4.63)
  MultiValueDictionaryPerformanceTests.Add: InvocationCount=1, UnrollFactor=1           -> It seems that the distribution is bimodal (mValue = 3.48)
  MultiValueDictionaryPerformanceTests.ContainsValue: InvocationCount=1, UnrollFactor=1 -> It seems that the distribution can have several modes (mValue = 2.95)
  MultiValueDictionaryPerformanceTests.Clear: Default                                   -> It seems that the distribution is bimodal (mValue = 3.43)
  MultiValueDictionaryPerformanceTests.Ctor: Default                                    -> It seems that the distribution can have several modes (mValue = 3.14)
  MultiValueDictionaryPerformanceTests.GetKeys: Default                                 -> It seems that the distribution is bimodal (mValue = 4)
```

Could you please help me on how to solve them?

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
```diff
@@ -0,0 +1,143 @@
...
+        private int _value;
+
+        [GlobalSetup(Targets = new[] { nameof(Clear), nameof(GetKeys) })]
+        public void CreateMultiValueDictionary_SingleValues()
+        {
+            _dict = new MultiValueDictionary<int, int>();
+            Random rand = new Random(11231992);
+
+            while (_dict.Count < Size)
+                _dict.Add(rand.Next(), rand.Next());
```
adamsitnik: what if the dictionary contains given key? will `Add` throw an exception?

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
```diff
@@ -0,0 +1,143 @@
...
+        [Benchmark]
+        public void Clear()
+        {
+            _dict.Clear();
+        }
+
+        [Benchmark]
+        public void Ctor()
+        {
+            var _ = new MultiValueDictionary<int, string>();
```
adamsitnik: to prevent from dead code elimination by JIT you need to return the result from a method. This applies to this and all other benchmarks

safern:
> The benchmarks you have written are OK, you might take a look at what I wrote and copy the ideas if you want to.
>
> Please remember to return the result from benchmark method to avoid dead code elimination.

Thanks for the pointer. I will take a look and update accordingly.

tests/Benchmarks/Microsoft.Experimental.Collections/MultiValueDictionaryPerformanceTests.cs
safern: No, since this is a multivalue dictionary it allows to have multiple values for the same key, so if the key already exists it just adds the value to the key's collection of values. 

https://github.com/dotnet/corefxlab/blob/b17275841c60ece2e5e8e8b2f089588f4ca91a2e/src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/MultiValueDictionary.cs#L637

safern:
I updated the perf tests to actually return a value and also updated the PR description to include perf results for netcoreapp2.1 and net472.

danmosemsft:
@adamsitnik isn't there a way to get BDN to show both A and B results in the same table, ie. showing the ratio, so I don't have to scroll back and forth? Perhaps that's not possible here as it's different frameworks.

safern:
> @adamsitnik isn't there a way to get BDN to show both A and B results in the same table, ie. showing the ratio, so I don't have to scroll back and forth? Perhaps that's not possible here as it's different frameworks.

That would be handy.

danmosemsft:
Interesting that .NET Core is not faster/equal in every case. Eg., `TryGetValue | 16 | 1000`. Significant slower.

adamsitnik:
> isn't there a way to get BDN to show both A and B results in the same table, ie. showing the ratio, so I don't have to scroll back and forth? Perhaps that's not possible here as it's different frameworks.

@danmosemsft sure!

First of all we need to add .NET 4.7.2 to `Benchmarks.csproj`

`<TargetFrameworks>netcoreapp2.1;net472</TargetFrameworks>`

Then from console line arguments:

`--runtimes clr core`

an example: `dotnet run -c Release -f netcoreapp2.1 -- -f *Corefx* --runtimes clr core`

More info in the [docs](https://benchmarkdotnet.org/articles/guides/console-args.html#runtimes)

PR #2459, opened 2018-08-31
Update corefxlab license to match corefx

GrabYourPitchforks:
These were changed:

- The header at the top of each .cs source file.
- The headers generated by .tt files.
- Any copyright information (as XML comments) embedded in .csproj files.
- The `LICENSE` file.

These were __not__ changed:

- The `<copyright>` element in .csproj files, as they're baked into DLLs.
- The line at https://github.com/dotnet/corefxlab/blob/29d7aa090242c8e9db327c34812e5f370e67761c/src/System.Reflection.Metadata.Cil/Visitor/CilToStringVisitor.cs#L55, because it's just copyright information without any accompanying license.


src/System.Text.Primitives/System/Text/Parsers/Utf16Parser_signed.cs
```diff
@@ -3,8 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Licensed to the .NET Foundation under one or more agreements.
```
ahsonkhan: This text is duplicate. We should remove it.

src/System.Text.Primitives/System/Text/Parsers/Utf16Parser_signed_hex.cs
```diff
@@ -3,8 +3,9 @@
 // See the LICENSE file in the project root for more information.
 
 
-// Copyright (c) Microsoft. All rights reserved.
-// Licensed under the MIT license. See LICENSE file in the project root for full license information.
+// Licensed to the .NET Foundation under one or more agreements.
```
ahsonkhan: Same here, and other places under System.Text.Primitives.

safern:
Thanks @GrabYourPitchforks 😄

src/System.Text.Primitives/System/Text/Parsers/Utf16Parser_signed.cs
GrabYourPitchforks: Good catch - I'll update the regex to fix this.

scripts/install-dotnet.ps1
```diff
@@ -1,6 +1,7 @@
 ﻿#
-# Copyright (c) .NET Foundation and contributors. All rights reserved.
-# Licensed under the MIT license. See LICENSE file in the project root for full license information.
+# Licensed to the .NET Foundation under one or more agreements.
```
ahsonkhan: This file was borrowed from https://github.com/dotnet/cli/blob/master/scripts/obtain/dotnet-install.ps1, so I am not sure about the license here.

PR #2479, opened 2018-09-14
Update TryParse for int to match Utf8Parser

JeremyKuhne:
- Update to handle same inputs Utf8Parser.TryParse handles
- Allow specifying format char
- Update bool try parse


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -58,42 +64,88 @@ private unsafe static bool TryParseSlow(ref BufferReader<byte> reader, out bool
...
-                // Fast path had enough space but couldn't find valid data
-                value = default;
-                return false;
-            }
-
-            byte* buffer = stackalloc byte[MaxLength];
-            Span<byte> tempSpan = new Span<byte>(buffer, MaxLength);
-            if (Utf8Parser.TryParse(reader.PeekSlow(tempSpan), out value, out int consumed))
+                // stackalloc is separated as we can't represent that PeekSlow will not capture the span.
+                byte* stackBuffer = stackalloc byte[bufferSize];
```
KrzysztofCwalina: Why do you first create a pointer and then wrap it? you know you can stackalloc span buffers directly?

JeremyKuhne: That is what the comment is about. C# doesn't allow it as you could, in theory, capture the stack span in the reader. We have a request open to allow marking parameters as no capture so we can get around this.

JeremyKuhne: We could also force a copy of the reader by not taking it as ref, but that comes at a non-trivial cost.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -58,42 +64,88 @@ private unsafe static bool TryParseSlow(ref BufferReader<byte> reader, out bool
...
+            if (consumed < unread.Length)
             {
+                // The parser found a digit it wouldn't consume so we have all valid data.
                 reader.Advance(consumed);
                 return true;
             }
 
-            return TryParseSlow(ref reader, out value);
+            // If we ate all of our unread there may be more valid digits
+            return TryParseSlow(ref reader, out value, standardFormat);
```
KrzysztofCwalina: This means we will always hit the slow path at the end of the sequence/span, i.e consumed will be equal to unread.Length. Can we avoid that?

JeremyKuhne: Not easy to do as we would need to know if we're in the last span. We could add a check for single segment, but the expectation is that you won't hit this often and we want to keep the fast path as small as possible.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -58,42 +64,88 @@ private unsafe static bool TryParseSlow(ref BufferReader<byte> reader, out bool
...
+                Span<byte> buffer = new Span<byte>(stackBuffer, bufferSize);
+                ReadOnlySpan<byte> peekSpan = reader.PeekSlow(buffer);
+
+                if (!Utf8Parser.TryParse(peekSpan, out value, out int consumed, standardFormat))
+                {
+                    // Overflow
+                    return false;
+                }
+
+                if (consumed < peekSpan.Length || peekSpan.Length < buffer.Length)
```
KrzysztofCwalina: what if the int is at the end of the reader's buffer?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -58,42 +64,88 @@ private unsafe static bool TryParseSlow(ref BufferReader<byte> reader, out bool
...
+                {
+                    // The parser found a digit it wouldn't consume or we hit the end of the reader
+                    // (peekSpan was smaller than requested)so we have all valid data.
+                    reader.Advance(consumed);
+                    return true;
+                }
+
+                // Throw if doubling would overflow.
+                bufferSize = checked(bufferSize * 2);
+            } while (true);
```
KrzysztofCwalina: should we break out the loop after some threshold?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: That is what the `peekSpan.Length < buffer.Length` is checking. If we couldn't fill our requested peek, we're at the end.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
KrzysztofCwalina: Maybe the check could be inside Slow method but before we actually allocate the buffer and copy?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: We'll throw eventually with an overflow above. I'm not sure we can pick anything more rational than that. We could, in theory, return false- but we're starting to hit crazytown levels here (MBs of leading zeroes). I'm open to suggestions here.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
KrzysztofCwalina: Ah, I see.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: That we're single segment? We'd still need to parse again- and if we're truly single segment there will be no copy- peek will just slice.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
KrzysztofCwalina: I was wondering what happens if we request 15 bytes, get 15 bytes, and then parsed exactly 15 bytes? I gather we just loop and then the code path you are referring to will be hit. But this, together with my comment in the outer TryParse, means that if I parse exactly 15 bytes int, I will not only hit the slow code path, but also loop here. If it's not avoidable, then maybe it's ok, but it does seem wasteful at the first glance.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
KrzysztofCwalina: What does int.Parse do? Does it accept 1 billion leading zeros?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
KrzysztofCwalina: Ah, I see. I forgot that peek with simply return the last span (if it's single segment)

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: Well, up to almost int.MaxValue in theory. I considered doing our own scanning to get past these cases, but it is complicated and would fall over if the Utf8Parser changes behavior or adds another format type. There are a number of corner cases in that code.

I think it is reasonable to expect users to do their own skipping for crazy cases. We've got all the helpers for that.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: You'd have to have exactly 15 bytes left in the reader's remaining span **and** have them all be valid to hit this case. Normal cases will be < 15. Together I think that case should be really rare.

You can also come in with more than 15 bytes if you're matching the end of the span. Same sort of rareness. We could change behavior here to check for these cases to avoid extra work, but I would expect them to impact the perf of common case- which is 14 bytes crossing segments.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -58,42 +64,88 @@ private unsafe static bool TryParseSlow(ref BufferReader<byte> reader, out bool
...
             ReadOnlySpan<byte> unread = reader.UnreadSpan;
-            if (Utf8Parser.TryParse(unread, out value, out int consumed) && consumed < unread.Length)
+            if (!Utf8Parser.TryParse(unread, out value, out int consumed, standardFormat))
+            {
+                return false;
+            }
+
+            if (consumed < unread.Length)
             {
+                // The parser found a digit it wouldn't consume so we have all valid data.
```
ahsonkhan: How does this comment explain what's happening here? How do we know that the parser found a digit, specifically? Isn't it just that there is more data left in this segment?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -58,42 +64,88 @@ private unsafe static bool TryParseSlow(ref BufferReader<byte> reader, out bool
...
             {
-                // Fast path had enough space but couldn't find valid data
-                value = default;
-                return false;
-            }
-
-            byte* buffer = stackalloc byte[MaxLength];
-            Span<byte> tempSpan = new Span<byte>(buffer, MaxLength);
-            if (Utf8Parser.TryParse(reader.PeekSlow(tempSpan), out value, out int consumed))
+                // stackalloc is separated as we can't represent that PeekSlow will not capture the span.
```
ahsonkhan: represent -> guarantee?

JeremyKuhne: represent- we can technically guarantee, we just have no way to represent that to C# (yet)

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -58,42 +64,88 @@ private unsafe static bool TryParseSlow(ref BufferReader<byte> reader, out bool
...
-
-            ReadOnlySpan<byte> unread = reader.UnreadSpan;
+            // Note that ints can have leading zeros ("-00000000000000127") which makes predicting the max
+            // length difficult. Typically the int will be less than 15 characters ("-2,147,483,648") so we'll
+            // work in those increments.
+            int bufferSize = 15;
 
-            if (unread.Length > MaxLength)
+            // Try using the stack for scratch space.
+            do
```
ahsonkhan: Rather than continuously doubling (and potentially throwing if there are 2.1 billion leading zeroes, which should be valid), why not skip all leading zeroes ahead of time? In the common case, it will be a single check but the implementation would be clearer.

JeremyKuhne: We'd have to deal with '+', '-', '.' and other formats that might come in the future.  Doesn't seem wise to couple that closely and add that much complexity for something that arguably should be handled by the consumer of this API.

That said, I think might even be more aggressive here. Lets say someone throws an empty 1GB buffer at us- should we really cruise the entire thing and return 0? Is that really what a user would want?

JeremyKuhne:
Marked as no merge temporarily. Found an issue where we fall over reading segments that end with a single `+`, `-`, or `.`.

tests/System.Buffers.Experimental.Tests/BufferFactory.cs
```diff
@@ -53,6 +53,14 @@ public static ReadOnlySequence<byte> Create(params byte[][] buffers)
             return Create(list.ToArray());
         }
 
+        public static ReadOnlySequence<byte> CreateUtf8(params string[] buffers)
+        {
+            if (buffers.Length == 1) return new ReadOnlySequence<byte>(Encoding.UTF8.GetBytes(buffers[0]));
+            var list = new List<Memory<byte>>();
+            foreach (var b in buffers) list.Add(Encoding.UTF8.GetBytes(b));
```
ahsonkhan: nit: avoid using var here

also not: avoid single letter variable names

tests/System.Buffers.Primitives.Tests/BufferUtilities.cs
```diff
@@ -114,5 +115,20 @@ public static ReadOnlySequence<byte> CreateUtf8Buffer(params string[] inputs)
             }
             return CreateBuffer<T>(buffers);
         }
+
+        public static void FillIntegerUtf8Array(byte[] array, int minValue, int maxValue, int seed = 42)
+        {
+            Random r = new Random(42);
```
ahsonkhan: Use the seed that was passed in.

tests/System.Buffers.ReaderWriter.Tests/Reader_ParseFloatingPoint.cs
```diff
@@ -0,0 +1,149 @@
...
+using System.Buffers.Reader;
+using System.Globalization;
+using System.Text;
+using Xunit;
+
+namespace System.Buffers.Tests
+{
+    public class Reader_ParseFloatingPoint
+    {
+
```
ahsonkhan: nit: remove extra white space

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -3,22 +3,57 @@
...
+        private static ParseDelegate<uint> s_uintParse;
         private static ParseDelegate<long> s_longParse;
+        private static ParseDelegate<ulong> s_ulongParse;
+        private static ParseDelegate<float> s_floatParse;
+        private static ParseDelegate<double> s_doubleParse;
+        private static ParseDelegate<decimal> s_decimalParse;
+
+        private static ParseDelegate<byte> ByteParser
+        {
+            [MethodImpl(MethodImplOptions.NoInlining)]
```
ahsonkhan: Why is this necessary?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -56,9 +159,9 @@ private static ParseDelegate<long> LongParser
         /// <see cref="Utf8Parser.TryParse(ReadOnlySpan{byte}, out bool, out int, char)"/> for supported formats.
         /// </remarks>
         /// <exception cref="FormatException">Thrown if the given format isn't supported.</exception>
-        /// <returns>True if successfully parsed.</returns>
+        /// <returns>Number of consumed bytes if successful, otherwise 0.</returns>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public static unsafe bool TryParse(ref this BufferReader<byte> reader, out bool value, char standardFormat = '\0')
+        public static unsafe int TryParse(ref this BufferReader<byte> reader, out bool value, char standardFormat = '\0')
```
ahsonkhan: This method doesn't need to be unsafe.

ahsonkhan: Also, do we have precedence elsewhere that a TryX method doesn't return bool? If not, maybe we should rename it.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -68,91 +171,152 @@ public static unsafe bool TryParse(ref this BufferReader<byte> reader, out bool
             if (Utf8Parser.TryParse(unread, out value, out int consumed, standardFormat))
             {
                 reader.Advance(consumed);
-                return true;
+                return consumed;
+            }
+            else if (unread.Length >= MaxBoolLength)
```
ahsonkhan: I am not sure if this is correct. What if `unread = {tru1}`? TryParse returns false, `unread.Length is < MaxBoolLength`, and then we fall to TryParseSlow which isn't what we want to do here.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -3,138 +3,700 @@
...
+        /// </summary>
+        /// <param name="value">The parsed <see cref="byte"/> value or default if failed.</param>
+        /// <param name="standardFormat">Expected format.</param>
+        /// <remarks>
+        /// <see cref="Utf8Parser.TryParse(ReadOnlySpan{byte}, out byte, out int, char)"/> for supported formats.
+        /// </remarks>
+        /// <exception cref="FormatException">Thrown if the given format isn't supported.</exception>
+        /// <returns>Number of consumed bytes if successful, otherwise 0.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe int TryParse(
```
ahsonkhan: same here, doesn't need to be unsafe

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -3,138 +3,700 @@
...
+            int bufferSize,
+            ParseDelegate<T> parseDelegate) where T: unmanaged
+        {
+            // Note that in calling this with the expected max size of a normal instance of the given
+            // type may do more work than needed for the degenerate case of zero prefaced numbers at
+            // the end of the span (e.g. 00000000000000000000000000123). Doing the extra check slows
+            // down the perf of "normal" numbers so we just let the "bad" numbers eat up more cycles.
+
+            // stackalloc is separated as we can't represent in C# that Peek will not capture the span.
+            byte* stackBytes = stackalloc byte[MaxParseBytes];
```
ahsonkhan: Does it make sense to separate out the common path (where we stackalloc exactly bufferSize rather than MaxParseBytes)? If we fail on our first attempt, we can then fall back to the doubling loop below.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -3,138 +3,700 @@
...
+        /// <exception cref="FormatException">Thrown if the given format isn't supported.</exception>
+        /// <returns>Number of consumed bytes if successful, otherwise 0.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe int TryParse(
+            ref this BufferReader<byte> reader,
+            out int value,
+            char standardFormat = '\0')
         {
-            const int MaxLength = 15;
+            //return TryParseFast(ref reader, out value, standardFormat, 15, IntParser);
```
ahsonkhan: remove commented out code

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: It is what we want to do, if we want this code path to be as fast as possible. Every check shows up in the perf runs. All of these methods make some technically unnecessary trips down the slow path.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -68,91 +171,152 @@ public static unsafe bool TryParse(ref this BufferReader<byte> reader, out bool
...
+            }
 
-                if (!parseDelegate(peekSpan, out value, out int consumed, standardFormat))
-                {
-                    // Overflow
-                    return false;
-                }
+            // If we ate all of our unread there may be more valid digits.
+            // The typical max size for a short will be less than 4 bytes ("255").
+            return TryParseSlow(ref reader, out value, standardFormat, 4, ByteParser);
```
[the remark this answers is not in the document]
JeremyKuhne: Because you need a non-valid character after the valid data.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: Just being explicit- don't ever want to generate the delegate until we absolutely need it.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: No, because the logic would be really complicated to avoid stack pressure. Every `stackalloc` adds more to the stack, even in a loop.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: Whoops- remnant of discovering there is no way to do generics here without significant overhead. :(

tests/System.Buffers.Experimental.Tests/BufferFactory.cs
JeremyKuhne: Yeah, copying existing code around.

JeremyKuhne: Var is ok in this context btw

tests/System.Buffers.Primitives.Tests/BufferUtilities.cs
JeremyKuhne: Derp

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -3,138 +3,700 @@
...
+            [MethodImpl(MethodImplOptions.NoInlining)]
+            get
+            {
+                if (s_intParse == null)
+                    s_intParse = Utf8Parser.TryParse;
+                return s_intParse;
+            }
+        }
+
+        private static ParseDelegate<uint> UintParser
```
KrzysztofCwalina: nit:lower case "i"

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -3,138 +3,700 @@
...
+            [MethodImpl(MethodImplOptions.NoInlining)]
+            get
+            {
+                if (s_byteParse == null)
+                    s_byteParse = Utf8Parser.TryParse;
+                return s_byteParse;
+            }
+        }
+
+        private static ParseDelegate<sbyte> SbyteParser
```
KrzysztofCwalina: nit: I would call it "SByte" to match the type name

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -3,138 +3,700 @@
...
+        /// </summary>
+        /// <param name="value">The parsed <see cref="float"/> value or default if failed.</param>
+        /// <param name="standardFormat">Expected format.</param>
+        /// <remarks>
+        /// <see cref="Utf8Parser.TryParse(ReadOnlySpan{byte}, out float, out int, char)"/> for supported formats.
+        /// </remarks>
+        /// <exception cref="FormatException">Thrown if the given format isn't supported.</exception>
+        /// <returns>Number of consumed bytes if successful, otherwise 0.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public static unsafe int TryParse(
```
KrzysztofCwalina: Why did we return number of consumed bytes as opposed to having an out parameter? I assume it's for perf, but it's not consistent with the general pattern that we already shipped. Is it really worth departing from the pattern?

JeremyKuhne: Yeah, but it wasn't a huge difference. I'm going to change it.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
JeremyKuhne: I was matching the c# keywords. I can change these to Single etc.

PR #2486, opened 2018-09-21
Add line number and position to JsonReader error message.

ahsonkhan:
Results in code duplication since we have an internal type called `Utf8JsonReaderInstrumented` which re-processes the data when we find out that it is invalid so that we can track the line number and position and return that to the user with the JsonReaderException.

This is done purely for performance. Tracking line number/position resulted in significant performance regression, especially if the property name/string contains embedded new line characters (which we would have to track to get the line number correct and reset the position).

The other reason to re-process the data is to avoid allocations (of stacks) in the common case if we decide to add a path string to the error message in the future.


src/System.Text.JsonLab/System/Text/Json/Utf8JsonReaderInstrumented.cs
```diff
@@ -0,0 +1,667 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace System.Text.JsonLab
+{
+    //TODO: Add multi-segment support
+    internal ref struct Utf8JsonReaderInstrumented
```
KrzysztofCwalina: How are we going to be able to keep these two in-sync? Would it be possible to share some of the code?

ahsonkhan: > How are we going to be able to keep these two in-sync?

That is the maintainability concern. I am open to alternate suggestions. Maybe there is a way to do this with #ifdefs where we keep the source identical but have the exception/line number code within an ifdef that is false for Utf8JsonReader and true for Utf8JsonReaderInstrumented. Instead of propagating false whenever we hit an exception, I could throw the exception right away.

> Would it be possible to share some of the code?

Almost all the methods have certain distinctions which makes sharing difficult. I can try to refactor some of the code and see if that gets us anywhere.

KrzysztofCwalina: I am pretty sure refactoring will not be very useful, or will slow down the code. Ifdefs seem like a best bet, but I am not sure how easy it will be to include the same file twice with different defines in our build script.

ahsonkhan: I used file level #define to work around the issue. Now, JsonReader.cs and Utf8JsonReaderInstrumented.cs are identical (other than the `#define` at the beginning in one). That should help keep things in sync (at least for now).

Issue #2487, opened 2018-09-21
[Proposal] Create System.Devices.I2c API


shaggygi:
## Initial Rambles...
This proposal is to define a new API for I2C devices.  It is somewhat related to the new System.Devices.Gpio work.  However, the intent is to remove and separate into different APIs to focus on each specific type of I/O instead of one gigantic library (with a mix/match of unneeded dependencies.  There are many ways to accomplish this and hope others would chime in.  This proposal is opinionated and related to my initial thoughts here (#2388).  Also, this issue is a WIP and I will be making updates over the next few days/weeks during spare time.

As mentioned before, I believe we could take advantage of some of the work that came from the <a href="https://docs.microsoft.com/en-us/uwp/api/windows.devices" target="_blank">Windows.Devices.XXX APIs</a>.  It should be possible to have similar namespaces like `System.Devices.Gpio`, `System.Devices.I2c`, `System.Devices.Spi`, etc..  Later, we could have more advanced APIs following the same thinking like `System.Devices.Serial`, `System.Devices.Bluetooth`, `System.Devices.Usb`, etc..  I plan to create a PR soon with initial code, but is only a starting point.  Samples, tests, docs and such will be added overtime if this approach seems like a good path to research/implement.

This proposal is focused on `System.Devices.I2c` to verify capabilities and limitations.  However, I also have started some other work related to other APIs mentioned above (e.g. SPI, GPIO, etc.).

## System.Devices.I2c API

While GPIO is general (hence the name) and can reference I2C-type pins, I like having the option of separate APIs based on functionality and project requirements.  Sometimes you might only need to toggle a few pins and other times you need to communicate with just a bus (I2C/SPI) device such as temperature, real-time clocks or EEPROMs.  The `System.Devices.I2c` API will provide the basic classes and interfaces related to I2C.  This includes enums like bus speed and sharing modes.  It also will include advanced helpers to support off-loading [I2C Bus Providers](https://docs.microsoft.com/en-us/windows/iot-core/develop-your-app/busproviders) (e.g. Ardruino, etc.).

### System.Devices.II2cDevice

[...]

PR #2488, opened 2018-09-21
Added initial System.Devices.I2c project

shaggygi:
Added the initial System.Devices.I2c project to repo.  This includes the basic I2C classes/interfaces and related to (#2487).  More work to follow with a working I2cDevice implementation along with sample.

### DIFF
--- a/src/System.Devices.I2c/I2cDevice.cs
+++ b/src/System.Devices.I2c/I2cDevice.cs
@@ -31,7 +31,7 @@ namespace System.Devices.I2c
         /// A buffer that contains the data to write to the I²C device.
         /// This data should not include the bus address.
         /// </param>
-        public abstract void Write(Span<byte> buffer);
+        public abstract void Write(ReadOnlySpan<byte> buffer);
 
         /// <summary>
         /// Writes data to the inter-integrated circuit (I²C) bus on which the
@@ -46,7 +46,7 @@ namespace System.Devices.I2c
         /// Information about the success of the write operation and the actual
         /// number of bytes that the operation wrote into the buffer.
         /// </returns>
-        public abstract I2cTransferResult WritePartial(Span<byte> buffer);
+        public abstract I2cTransferResult WritePartial(ReadOnlySpan<byte> buffer);
 
         /// <summary>
         /// Reads data from the inter-integrated circuit (I²C) bus on which the
@@ -86,7 +86,7 @@ namespace System.Devices.I2c
         /// The buffer to read the data from the I²C bus.
         /// The length of the buffer determines how much data to request from the device.
         /// </param>
-        public abstract void WriteRead(Span<byte> writeBuffer, Span<byte> readBuffer);
+        public abstract void WriteRead(ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer);
 
         /// <summary>
         /// Performs an atomic operation to write data to and then read data from the
@@ -106,7 +106,7 @@ namespace System.Devices.I2c
         /// succeeded and the sum of the actual number of bytes that the operation
         /// wrote and the actual number of bytes that the operation read.
         /// </returns>
-        public abstract I2cTransferResult WriteReadPartial(Span<byte> writeBuffer, Span<byte> readBuffer);
+        public abstract I2cTransferResult WriteReadPartial(ReadOnlySpan<byte> writeBuffer, Span<byte> readBuffer);
 
         public abstract void Dispose();
         public abstract void Dispose(bool disposing);


src/System.Devices.I2c/II2cConnectionSettings.cs
```diff
@@ -0,0 +1,13 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Devices.I2c
+{
+    internal interface II2cConnectionSettings
```
KrzysztofCwalina: Why do we need this interface (and other interfaces in this assembly)? It seems like there is a single implementation for them and we don't have any APIs that actually consume these interfaces. In .NET, we add abstractions in cases when multiple types implement an abstractions and also some APIs consume (take as parameter, for example) the same abstraction.

Also, we prefer classes over interfaces for abstractions. We only use interfaces when we need types to implement more than one abstraction (i.e. multiple-inheritance).

src/System.Devices.I2c/II2cDevice.cs
```diff
@@ -0,0 +1,20 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Devices.I2c
+{
+    // TODO: Document
+    public interface II2cDevice : IDisposable
+    {
+        void Write(byte[] buffer);
```
KrzysztofCwalina: Should these take `Span<byte>`?

src/System.Devices.I2c/II2cDeviceStatics.cs
```diff
@@ -0,0 +1,11 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Devices.I2c
+{
+    // TODO: Need to review this.
+    internal interface II2cDeviceStatics
```
KrzysztofCwalina: Why an empty interface?

src/System.Devices.I2c/Provider/ProviderI2cBusSpeed.cs
```diff
@@ -0,0 +1,24 @@
...
+// See the LICENSE file in the project root for more information.
+
+namespace System.Devices.I2c.Provider
+{
+    /// <summary>
+    /// Describes the bus speeds that are available for connecting to an inter-integrated
+    /// circuit (I²C) device. The bus speed is the frequency at which to
+    /// clock the I²C bus when accessing the device.
+    /// </summary>
+    public enum ProviderI2cBusSpeed
```
KrzysztofCwalina: Why do we need two enums (and other types) for the client and the provider? They seem to share values.

src/System.Devices.I2c/II2cDeviceStatics.cs
shaggygi: I added and missed by mistake.  It was another based on Windows.Devices and appeared to include methods related to AQS to find items.  I planned to review the need, but will removed for now.

src/System.Devices.I2c/Provider/ProviderI2cBusSpeed.cs
shaggygi: Good point.  I agree and was thinking there could eventually be common enums.  In addition, I'm still scratching my head a little on functionality of Providers and planned that topic to be a future investigation.  Just thinking it would be good to go ahead for something to start with.  Until a review (and better strategy) how Providers work... I went ahead and removed the related files for now.

src/System.Devices.I2c/II2cDevice.cs
shaggygi: I believe so.  Honestly, I'm still trying to get a better understanding on Span.  I'm hoping working on this project helps me in this area.  I changed the byte[] to Span<byte> for now as I wasn't sure if there are some that could be ReadOnlySpan<byte>.

src/System.Devices.I2c/I2cDevice.cs
```diff
@@ -0,0 +1,114 @@
...
+        /// <summary>
+        /// Writes data to the inter-integrated circuit (I²C) bus on which the
+        /// device is connected, based on the bus address specified in the
+        /// <see cref="I2cConnectionSettings"/> object used to create the <see cref="I2cDevice"/> object.
+        /// </summary>
+        /// <param name="buffer">
+        /// A buffer that contains the data to write to the I²C device.
+        /// This data should not include the bus address.
+        /// </param>
+        public abstract void Write(Span<byte> buffer);
```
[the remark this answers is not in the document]
shaggygi: Updated write buffer as ReadOnlySpan.

shaggygi:
Closing out based on comments. Will try to clean up within same assembly later.

PR #2491, opened 2018-09-24
Optimize reading numbers by removing redundant search for end of number

ahsonkhan:
Instead of using IndexOfAny(JsonConstants.Delimiters) to find the end of a number and then validating the number, read through the number and validate as we go.

Resulted in significant improvement (up to 40% in the best case) depending on how many numbers existed within the payload.

``` ini

BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17754
.NET Core SDK=3.0.100-alpha1-009460
  Job-SSUQKN : .NET Core 2.1.3 (CoreCLR 4.6.26725.06, CoreFX 4.6.26725.05), 64bit RyuJIT

IterationCount=10  WarmupCount=5  

```

**BEFORE:**

|                               Method | IsDataCompact |       TestCase |          Mean |         Error |         StdDev | Allocated |
|------------------------------------- |-------------- |--------------- |--------------:|--------------:|---------------:|----------:|
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |      **BasicJson** |     **658.91 ns** |     **46.162 ns** |     **30.5334 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |  **BasicLargeNum** |     **599.76 ns** |     **11.982 ns** |      **7.1301 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |     **HelloWorld** |      **70.66 ns** |      **1.324 ns** |      **0.6926 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |      **Json400KB** | **889,774.57 ns** | **22,262.378 ns** | **14,725.1855 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |  **LotsOfNumbers** |   **3,888.03 ns** |     **74.180 ns** |     **44.1435 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** | **SpecialNumForm** |   **1,611.65 ns** |    **110.629 ns** |     **73.1742 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |      **BasicJson** |     **451.48 ns** |     **21.443 ns** |     **14.1832 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **BasicLargeNum** |     **734.43 ns** |     **24.650 ns** |     **16.3044 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |     **HelloWorld** |      **61.58 ns** |      **2.746 ns** |      **1.8165 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |      **Json400KB** | **665,780.11 ns** | **27,569.075 ns** | **16,405.9199 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **LotsOfNumbers** |   **3,291.76 ns** |    **150.555 ns** |     **99.5826 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** | **SpecialNumForm** |   **1,450.84 ns** |     **39.853 ns** |     **20.8439 ns** |       **0 B** |

**AFTER:**

|                               Method | IsDataCompact |       TestCase |          Mean |          Error |        StdDev | Allocated |
|------------------------------------- |-------------- |--------------- |--------------:|---------------:|--------------:|----------:|
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |      **BasicJson** |     **562.92 ns** |     **10.9657 ns** |     **7.2532 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |  **BasicLargeNum** |     **463.39 ns** |      **6.1524 ns** |     **3.2178 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |     **HelloWorld** |      **72.14 ns** |      **0.4956 ns** |     **0.2949 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |      **Json400KB** | **815,310.31 ns** | **13,052.7690 ns** | **8,633.5988 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** |  **LotsOfNumbers** |   **2,821.09 ns** |     **27.3199 ns** |    **16.2576 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |         **False** | **SpecialNumForm** |   **1,295.33 ns** |     **25.3528 ns** |    **15.0870 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |      **BasicJson** |     **336.49 ns** |      **7.1159 ns** |     **3.7218 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **BasicLargeNum** |     **467.04 ns** |     **12.1527 ns** |     **6.3561 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |     **HelloWorld** |      **63.68 ns** |      **7.8392 ns** |     **4.6650 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |      **Json400KB** | **590,439.59 ns** |  **7,408.7881 ns** | **4,900.4548 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **LotsOfNumbers** |   **1,934.43 ns** |     **23.8542 ns** |    **12.4763 ns** |       **0 B** |
| **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** | **SpecialNumForm** |   **1,205.45 ns** |     **88.8544 ns** |    **58.7717 ns** |       **0 B** |


src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -735,24 +735,20 @@ private void ConsumeSingleValue(ref BufferReader<byte> reader, byte marker)
             else if ((uint)(marker - '0') <= '9' - '0')
             {
                 ReadOnlySequence<byte> sequence = reader.Sequence.Slice(reader.Position);
-                Value = sequence.IsSingleSegment ? sequence.First.Span : sequence.ToArray();
+                Value = GetNumber(sequence.IsSingleSegment ? sequence.First.Span : sequence.ToArray());
```
KrzysztofCwalina: If it's not a single segment, we should first search the delimiter, and then allocate an array of the "right" size (i.e. fitting data up to the delimiter).

KrzysztofCwalina: Also, I wonder if it's worth, but maybe we could rent the array from a pool and let users return it to the same pool?

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
```diff
@@ -1471,21 +1365,34 @@ private void AdjustLineNumber(ReadOnlySpan<byte> span)
...
             {
                 if (i >= data.Length)
+#if UseInstrumented
+                    throw new JsonReaderException($"Invalid number. Last character read: '{(char)nextByte}'. Expected a digit.", _lineNumber, _position);
+#else
                     JsonThrowHelper.ThrowJsonReaderException(ref this);
+#endif
+
                 nextByte = data[i++];
+                if (delimiters.IndexOf(nextByte) != -1)
```
KrzysztofCwalina: This will do vectorized IndexOf, which we know is slower for very small buffers. How large is the delimiters buffer? Is it worth doing vectorized IndexOf here?

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
ahsonkhan: > If it's not a single segment, we should first search the delimiter, and then allocate an array of the "right" size (i.e. fitting data up to the delimiter).

Since this code path is for single json values only, the delimiter will be at the end of the sequence anyway (if the json is valid), so there is no point of searching. The right size that fits the data is the entire length of the sequence, so calling ToArray is the same as finding the right size.

For json with multiple values (within objects/arrays), we use BufferReader.TryReadToAny which allocates the minimal amount necessary if the data straddles between two segments (we search for the delimiter there).

I will see if I can avoid the separate code path for single json values all together to make the code easier to read/understand. However, for valid json we are not over-allocating. It is possible, in the worst-case, for us to allocate quite a bit if the json is invalid.

ahsonkhan: > Also, I wonder if it's worth, but maybe we could rent the array from a pool and let users return it to the same pool?

This requirement would need coordination with the BufferReader, since that is where the allocation is happening:
https://github.com/dotnet/corefxlab/blob/952ca00c3e5a6c930447cf53116f45b2aa007b67/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs#L122

cc @JeremyKuhne

JeremyKuhne: I don't know that we have a good pattern yet for passing out rented memory without allocating. @KrzysztofCwalina is there a reason we don't have a struct that implements IMemoryOwner? Is that a rational thing? Or should there be some other pattern for return-when-done?

src/System.Text.JsonLab/System/Text/Json/JsonReader.cs
ahsonkhan: The delimiters buffer is 7 elements: https://github.com/dotnet/corefxlab/blob/952ca00c3e5a6c930447cf53116f45b2aa007b67/src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs#L44

I am not sure if its worth it to special case it.

There was no noticeable difference between a naiive for loop and using IndexOf here (within error).
Reading an array of 10 million numbers:
Before: 281.8 ms
After: 282.3 ms

The break-even point tends to be for buffer of size 3 (if we find the search item at the end). Otherwise, for size <= 3, a for-loop is 25% faster.


Method | DataLength | Mean | Scaled
-- | -- | -- | --
  |   |   |  
IndexOf | 1 | 317.2 ns | 1
ForLoop | 1 | 231.1 ns | 0.73
  |   |   |  
IndexOf | 2 | 424.0 ns | 1
ForLoop | 2 | 320.9 ns | 0.76
  |   |   |  
IndexOf | 3 | 526.6 ns | 1
ForLoop | 3 | 383.5 ns | 0.73
  |   |   |  
IndexOf | 4 | 346.8 ns | 1
ForLoop | 4 | 501.5 ns | 1.45
  |   |   |  
IndexOf | 5 | 378.5 ns | 1
ForLoop | 5 | 539.5 ns | 1.43
  |   |   |  
IndexOf | 6 | 539.2 ns | 1
ForLoop | 6 | 609.2 ns | 1.13
  |   |   |  
IndexOf | 7 | 560.6 ns | 1
ForLoop | 7 | 655.2 ns | 1.17
  |   |   |  
IndexOf | 8 | 366.0 ns | 1
ForLoop | 8 | 727.7 ns | 1.99
  |   |   |  
IndexOf | 9 | 451.2 ns | 1
ForLoop | 9 | 770.9 ns | 1.71
  |   |   |  
IndexOf | 10 | 513.6 ns | 1
ForLoop | 10 | 821.5 ns | 1.6
  |   |   |  
IndexOf | 15 | 678.8 ns | 1
ForLoop | 15 | 1,080.6 ns | 1.59
  |   |   |  
IndexOf | 16 | 568.7 ns | 1
ForLoop | 16 | 1,153.8 ns | 2.03
  |   |   |  
IndexOf | 17 | 646.0 ns | 1
ForLoop | 17 | 1,217.2 ns | 1.88
  |   |   |  
IndexOf | 20 | 682.7 ns | 1
ForLoop | 20 | 1,383.9 ns | 2.03
  |   |   |  
IndexOf | 21 | 726.0 ns | 1
ForLoop | 21 | 1,414.0 ns | 1.95



Calling IndexOf for size 7 is up to 15% slower to  up to15% faster depending on whether we find the search item (and where).


Method | FoundAt | Mean | Scaled
-- | -- | -- | --
IndexOf | 1 | 233.8 ns | 1
ForLoop | 1 | 197.1 ns | 0.84
  |   |   |  
IndexOf | 2 | 272.6 ns | 1
ForLoop | 2 | 322.0 ns | 1.18
  |   |   |  
IndexOf | 3 | 285.4 ns | 1
ForLoop | 3 | 406.1 ns | 1.42
  |   |   |  
IndexOf | 4 | 302.1 ns | 1
ForLoop | 4 | 481.6 ns | 1.59
  |   |   |  
IndexOf | 5 | 482.2 ns | 1
ForLoop | 5 | 511.0 ns | 1.06
  |   |   |  
IndexOf | 6 | 574.7 ns | 1
ForLoop | 6 | 553.0 ns | 0.96
  |   |   |  
IndexOf | 7 | 604.9 ns | 1
ForLoop | 7 | 572.3 ns | 0.95

```C#
[Benchmark(Baseline = true)]
public int IndexOf()
{
    ReadOnlySpan<byte> delimiters = Delimiters.AsSpan(0, DataLength);
    byte lookup = delimiters[DataLength - 1];
    int sum = 0;
    for (int i = 0; i < 100; i++)
        sum += delimiters.IndexOf(lookup);
    return sum;
}
```

ahsonkhan: Note, if we expect not to find the lookup byte within the buffer, then the break-even point is 8 (not 3).

Search element not found:

Method | DataLength | Mean | Scaled
-- | -- | -- | --
  |   |   |  
IndexOf | 1 | 347.4 ns | 1
ForLoop | 1 | 202.4 ns | 0.58
  |   |   |  
IndexOf | 2 | 431.2 ns | 1
ForLoop | 2 | 234.9 ns | 0.55
  |   |   |  
IndexOf | 3 | 506.1 ns | 1
ForLoop | 3 | 354.3 ns | 0.7
  |   |   |  
IndexOf | 4 | 316.9 ns | 1
ForLoop | 4 | 437.9 ns | 1.38
  |   |   |  
IndexOf | 5 | 406.1 ns | 1
ForLoop | 5 | 373.2 ns | 0.92
  |   |   |  
IndexOf | 6 | 508.8 ns | 1
ForLoop | 6 | 396.9 ns | 0.78
  |   |   |  
IndexOf | 7 | 567.6 ns | 1
ForLoop | 7 | 420.2 ns | 0.74
  |   |   |  
IndexOf | 8 | 443.1 ns | 1
ForLoop | 8 | 446.4 ns | 1.01
  |   |   |  
IndexOf | 9 | 480.9 ns | 1
ForLoop | 9 | 488.9 ns | 1.02
  |   |   |  
IndexOf | 10 | 522.1 ns | 1
ForLoop | 10 | 532.5 ns | 1.02
  |   |   |  
IndexOf | 15 | 631.8 ns | 1
ForLoop | 15 | 687.5 ns | 1.09
  |   |   |  
IndexOf | 16 | 619.1 ns | 1
ForLoop | 16 | 731.9 ns | 1.18
  |   |   |  
IndexOf | 17 | 677.9 ns | 1
ForLoop | 17 | 779.4 ns | 1.15
  |   |   |  
IndexOf | 20 | 688.6 ns | 1
ForLoop | 20 | 837.6 ns | 1.22
  |   |   |  
IndexOf | 21 | 812.2 ns | 1
ForLoop | 21 | 1,083.0 ns | 1.33

ahsonkhan: > This will do vectorized IndexOf, which we know is slower for very small buffers. How large is the delimiters buffer? Is it worth doing vectorized IndexOf here?

Also, note that I removed that specific use of IndexOf (and a few others). Explicitly looking for a digit was the correct behavior, rather than looking for a delimiter. So, we are using IndexOf sparingly now (essentially once before returning in the common case of integers, twice for decimals, and 3x if we have the e-notation).

ahsonkhan:
@dotnet-bot test this please

ahsonkhan:
@dotnet-bot test this please

PR #2493, opened 2018-09-25
Split single segment and multi segment functionality and related code refactoring

ahsonkhan:
- Renamed JsonReader.cs to Utf8JsonReader.cs to match filename and typename
- Split the mutli-segment code paths from Utf8JsonReader.cs into Utf8JsonReader_MultiSegment.cs
- Renamed Index to CurrentIndex and StartLocation to TokenStartIndex
- ~Removed ConsumeSingleValue and using ConsumeValue for it (especially since it is the uncommon case).~ Results in a redundant `if` check if first character is not '{' or '[' (i.e. json is a single value). **Edit:** Moved the ConsumeNumber logic into ReadFirstToken since we need to avoid searching for the delimiter (we may not find it).
- Simplified the names of helper methods (removed unnecessary Utf8/MultiSegment suffix).
- Reordered methods/fields so single segment and multi segment files are easier to compare.


src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
```diff
@@ -0,0 +1,438 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers;
+using System.Buffers.Reader;
+
+namespace System.Text.JsonLab
+{
+#if UseInstrumented
+    internal ref partial struct Utf8JsonReaderInstrumented
+#else
```
KrzysztofCwalina: BTW, this is not specific to this PR (and it's a nit), but I would keep the type names the same and just ifdef the namespace name (and then fully qualify when you want to use the instrumented one. I think it will be easier to read the ctor code if there are no ifdefs just to rename the type.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
```diff
@@ -0,0 +1,438 @@
...
+                ValueType = JsonValueType.Array;
+            }
+            else if ((uint)(marker - '0') <= '9' - '0')
+            {
+                ConsumeNumber(ref reader);
+            }
+            else if (marker == '-')
+            {
+                //TODO: Is this a valid check or do we need to do an Advance to be sure?
+                if (reader.End) JsonThrowHelper.ThrowJsonReaderException(ref this);
```
KrzysztofCwalina: When reviewing recent JsonReader PRs, I have seem lots of code that simply throws when there is not enough data. But we will need to make JsonReader be able to continue after more data arrives. Is that something you thought about? how would we enable starting reasing a ROS that is not complete JSON, getting more data, and continuing parsing.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
```diff
@@ -0,0 +1,438 @@
...
+                reader.Advance(idx);
+                CurrentIndex += idx;
+                return;
+            }
+            ConsumeStringWithNestedQuotes(ref reader);
+        }
+
+        private void ConsumeStringWithNestedQuotes(ref BufferReader<byte> reader)
+        {
+            //TODO: Optimize looking for nested quotes
```
KrzysztofCwalina: Do we have an API on BufferReader that can read to end of a string? If not, should we add it? If yes, does it support escape sequences?

JeremyKuhne: We don't. `TryReadTo` can do scenarios that are in the first 128 chars (ASCII), which would be the most common need I'd think. You could compose off of `TryReadToAny` for escaping or we could provide a method for ASCII escape chars? Something like:

``` C#
/// <param name="escape">When encountered, skip the following {T}.</param>
public bool TryReadTo(out ReadOnlySpan<T> span, T delimiter, T escape, bool advancePastDelimiter = true)
```

JeremyKuhne: Other escape chars are really rare so perhaps we just support backslash.


``` C#
public bool TryReadToSkipEscaped(out ReadOnlySpan<T> span, T delimiter, bool advancePastDelimiter = true)
```

ahsonkhan: Such an API would definitely be useful for JSON reading.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
ahsonkhan: Yes, I plan to add that with a `hasMoreData` flag, which defaults to false that the user passes in. We still need to keep the throw logic for invalid/incomplete json data in the case when the user sets `hasMoreData` as false and we reach the end of data state.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
ahsonkhan: Fixed.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReaderInstrumented_MultiSegment.cs
```diff
@@ -0,0 +1,460 @@
...
+            else
+            {
+                JsonThrowHelper.ThrowJsonReaderException(ref this);
+            }
+        }
+
+        private void ConsumeNumber(ref BufferReader<byte> reader)
+        {
+            //TODO: Increment Index by number of bytes advanced
+            if (!reader.TryReadToAny(out ReadOnlySpan<byte> span, JsonConstants.Delimiters, advancePastDelimiter: false))
```
ahsonkhan: @JeremyKuhne, would it be possible for TryReadToAny to return the entire data as an output span if none of the delimiters are found? The fact that this returns an empty span when no delimiters are found that results in me having to special case the following:
https://github.com/dotnet/corefxlab/pull/2493/files#diff-15d23dd0f42f71cf1a5e5426e7a2c609R66

JeremyKuhne: It could, but I think that would potentially be confusing and, more importantly, allocate unnecessarily. Why doesn't ConsumeNumber contain this logic?

KrzysztofCwalina: Why would it allocate?

JeremyKuhne: > Why would it allocate?

Because he wants the rest of the ROS as a span- at least that is what the code does here.

KrzysztofCwalina: Ah, I just noticed that the span is an out parameter. We should talk how we want to read/parse variable length values. Such out parameter will allocate in a lot of cases and I think we need a way to do it in a non-allocating way. When I initially prototypes BufferReader, I think I would return Position and/or int from methods that read variable size values, e.g. strings.

PR #2494, opened 2018-09-25
Add Date and Guid parsing

JeremyKuhne:
Add initial tests, tweak some comments

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
@@ -13,9 +13,13 @@ namespace System.Buffers.Reader
         // (i.e. ReadOnlySequence) this behavior can be extremely expensive and/or lead to overflow and
         // out-of-memory situations. We'll cap the maximum number of bytes to consume.
         private const int MaxParseBytes = 128;
-        private const int MaxBoolLength = 5;
+
+        private const int MaxBoolLength = 5;       // False
+        private const int MaxGuidLength = 38;      // {9f21bcb9-f5c2-4b54-9b1c-9e0869bf9c16}
+        private const int MaxDateTimeLength = 34;  // 9999-12-31T15:59:59.9999999-08:00
 
         private delegate bool ParseDelegate<T>(ReadOnlySpan<byte> source, out T value, out int consumed, char standardFormat);
+        private static ParseDelegate<bool> s_boolParse;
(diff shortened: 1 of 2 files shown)


tests/System.Buffers.ReaderWriter.Tests/Reader_ParseStruct.cs
```diff
@@ -0,0 +1,114 @@
...
+namespace System.Buffers.Tests
+{
+    public class Reader_ParseStruct
+    {
+        private delegate bool ParseDelegate<T>(ref BufferReader<byte> reader, out T value, out int bytesConsumed, char standardFormat = '\0');
+
+        [Fact]
+        public void TryParseGuid_MultiSegment()
+        {
+            Guid expected = new Guid("9f21bcb9-f5c2-4b54-9b1c-9e0869bf9c16");
```
ahsonkhan: nit: cache the string into a local rather than repeating it multiple times.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_text.cs
```diff
@@ -27,9 +31,68 @@ public static partial class ReaderExtensions
         private static ParseDelegate<float> s_floatParse;
         private static ParseDelegate<double> s_doubleParse;
         private static ParseDelegate<decimal> s_decimalParse;
+        private static ParseDelegate<Guid> s_guidParse;
+        private static ParseDelegate<TimeSpan> s_timeSpanParse;
+        private static ParseDelegate<DateTime> s_dateTimeParse;
+        private static ParseDelegate<DateTimeOffset> s_dateTimeOffsetParse;
 
         // Note: We mark the getters as NoInlining as we want to make sure the delegate creation is lazy.
```
KrzysztofCwalina: Would it be better to do:
if(s_xxxParse != null) return s_xxxParse;
else return InitializeXxxParse();
... and then let the calls inline?

And more generally, do you think lazy creating all the delegates makes sense? We are paying non inlined call cost plus a branch, and it seems that at least some of the delegates will be quite commonly used.

JeremyKuhne: > Would it be better to do

I can try one and see if it measures.

> do you think lazy creating all the delegates makes sense?

As we have so many, I believe so. Everything is tuned so "normal" usage doesn't incur the overhead of creating all of the delegates. It should be reasonably common to hit little to no delegate creation.

PR #2495, opened 2018-09-26
Implement prototype System.Devices.GpioDriver as Windows10Driver.

JohnTasler:
Really, this is just a prototype using the intern's initial work, in order to completely understand the current design.


samples/System.Devices.Gpio.Samples/Program.cs
```diff
@@ -16,7 +16,8 @@ private enum DeviceKind
             Unknown,
             RaspberryPi = 1,
             Odroid = 2,
-            Hummingboard = 3
+            Hummingboard = 3,
+            RaspberryPiOnWindows = 4,
```
KrzysztofCwalina: nit: Should it be called WindowsOnRaspberryPi?

src/System.Devices.Gpio/GpioController.cs
```diff
@@ -21,6 +21,10 @@ public GpioController(PinNumberingScheme numbering = PinNumberingScheme.Gpio)
             {
                 driver = new UnixDriver();
             }
+            else if (RuntimeInformation.IsOSPlatform(OSPlatform.Windows))
+            {
+                driver = new Windows10Driver();
+            }
             else
             {
                 throw new NotSupportedException($"Platform '{RuntimeInformation.OSDescription}' not supported");
```
KrzysztofCwalina: Is there a way to query whether the device is supported? e.g. `if (GpioController.IsSupported) { ...}`

src/System.Devices.Gpio/Windows10Driver.cs
```diff
@@ -0,0 +1,343 @@
...
+using System.Linq;
+using System.Runtime.CompilerServices;
+using System.Threading;
+
+using GpioPinValueChangedHandler = global::Windows.Foundation.TypedEventHandler<global::Windows.Devices.Gpio.GpioPin, global::Windows.Devices.Gpio.GpioPinValueChangedEventArgs>;
+using WinGpio = global::Windows.Devices.Gpio;
+
+namespace System.Devices.Gpio
+{
+    public class Windows10Driver : GpioDriver
```
KrzysztofCwalina: Why is it public?

samples/System.Devices.Gpio.Samples/Program.cs
joshfree: this is basically a test app; I think the name is ok 😄

JohnTasler: The name is simply an artifact to the way the test program was written. It should take a driver name/number, then a test name, then parameters for the pin number(s).

The only reason I chose RaspberryPiOnWindows is so that I could use a different pin number, so that I can just use the on-board green LED to blink (RPi2 only - not RPi3).

src/System.Devices.Gpio/GpioController.cs
JohnTasler: I would also like to look into how to find out more information about the platform (Windows version, Linux distro, etc.)

src/System.Devices.Gpio/Windows10Driver.cs
JohnTasler: Since the Windows10Driver can be created from a different *Windows*.Devices.GpioController, the class needs to have a constructor that specifies _which_ controller to adapt to the .NET Core API. Windows supports multiple GPIO devices.

joperezr:
@dotnet-bot test this please

JohnTasler:
@dotnet-bot help

PR #2496, opened 2018-09-26
Add isFinalBlock in an attempt to make JsonReader re-entrant

ahsonkhan:
The user passing `isFinalBlock = false` moves JsonReader in the right direction to support incomplete data and make it re-entrant. However, that alone is not sufficient since we need to pass back state information as well. We need more than JsonTokenType to be passed back in.

For example, say we have an object with single key/value pair that is split up as follows: `{"age": 30}` => `{"age":` and `30}`

Reading through the first segment sets the `Depth` to `1`, `TokenType` to `PropertyName`, and `InObject` to `true`. All this information is necessary to read through the second segment to validate that the end of object occurs at the right depth/nesting.

**TODO:**
- What state information is actually necessary to make the reader re-entrant?
- Add `isFinalBlock = false` logic for multi-segment data.


src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -266,24 +270,61 @@ private void ReadFirstToken(byte first)
...
+            return true;
         }
 
         private bool ReadSingleSegment()
         {
             bool retVal = false;
 
             if (CurrentIndex >= (uint)_buffer.Length)
+            {
+                if (!_isSingleValue && _isFinalBlock)
```
KrzysztofCwalina: We invented isFinalBlock for parsers that cannot unambiguously tell between end of data and incomplete data. Is this the case here? Wouldn't it be possible to simply return either Incomplete or Done when the reader parses all data. This would avoid all the isFinalBlock checks that surely add overhead.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -485,27 +528,49 @@ private void ConsumeValue(byte marker)
...
+            if (!span.StartsWith(Value))
             {
+                if (_isFinalBlock)
+                {
+                    goto Throw;
+                }
+                else
+                {
+                    if (span.Length > 1 && span[1] != Value[1])
+                        goto Throw;
```
KrzysztofCwalina: Should the reader be designed just like our encoders, i.e. they don't throw when they encounter bad data, but rather return with some "bad data" status. This allows the caller to inspect the bytesConsumed value to see where the bad data starts, fixup, and possibly keep going. This is an honest question, i.e. I am not sure what we should do here. @jamesNk, do people every try to recover from bad JSON when parsing?

cc: @davidfowl, @GrabYourPitchforks

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -620,55 +739,69 @@ private void ConsumePropertyName()
...
         }
 
-        private void ConsumeString()
+        private bool ConsumeString()
         {
             //Create local copy to avoid bounds checks.
             ReadOnlySpan<byte> localCopy = _buffer;
 
-            int idx = localCopy.Slice(CurrentIndex).IndexOf(JsonConstants.Quote);
+            int idx = localCopy.Slice(CurrentIndex + 1).IndexOf(JsonConstants.Quote);
```
KrzysztofCwalina: will the slice always succeed? i.e. are you sure the current index is not at the end? Maybe at least worth asserting.

tests/System.Text.JsonLab.Tests/JsonReaderTests.cs
```diff
@@ -308,22 +308,141 @@ public static void TestJsonReaderUtf8SpecialString(string jsonString, string exp
...
+        [InlineData("10.5e-1", 6, 0)]    // "10.5e-"
+        //[InlineData("{\"ints\":[1, 2, 3, 4, 5]}", 21, 21)]    // "{\"ints\":[1, 2, 3, 4, 5"
+        //[InlineData("{\"strings\":[\"abc\", \"def\"], \"ints\":[1, 2, 3, 4, 5]}", 24, 24)]  // "{\"strings\":[\"abc\", \"def\""
+        //[InlineData("{\"age\":30, \"name\":\"test}:[]\", \"another string\" : \"tests\"}", 19, 18)]   // "{\"age\":30, \"name\":\"test}"
+        public static void SplitJsonAsync(string jsonString, int splitLocation, int consumed)
+        {
+            var buffer = new char[jsonString.Length];
+            var reader = new StringReader(jsonString);
+
+            reader.ReadAsync(buffer, 0, splitLocation);
```
KrzysztofCwalina: You don't happen to need to await here (as the call returns synchronously, but this is very fragile. 

Having said that, I would not do actual async calls in all unit test like this. I think it would be enough to simulate continuing with the reader by first parsing partial data and then parsing the rest with a new reader.

tests/System.Text.JsonLab.Tests/JsonReaderTests.cs
```diff
@@ -308,22 +308,141 @@ public static void TestJsonReaderUtf8SpecialString(string jsonString, string exp
...
+        [InlineData("{\"name\":\"Ahso", 8)]
+        [InlineData("-", 0)]
+        [InlineData("0.", 0)]
+        [InlineData("10.5e", 0)]
+        [InlineData("10.5e-", 0)]
+        [InlineData("{\"ints\":[1, 2, 3, 4, 5", 21)]
+        [InlineData("{\"strings\":[\"abc\", \"def\"", 24)]
+        [InlineData("{\"age\":30, \"name\":\"test}", 18)]
+        [InlineData("[[[[{\r\n\"temp1\":[[[[{\"temp2:[]}]]]]}]]]]", 20)]
+        public static void PartialJson(string jsonString, int consumed)
```
KrzysztofCwalina: Does this test anything beyond what the tests above does?

tests/System.Text.JsonLab.Tests/JsonReaderTests.cs
```diff
@@ -308,22 +308,141 @@ public static void TestJsonReaderUtf8SpecialString(string jsonString, string exp
...
+                while (json.Read()) ;
+                Assert.True(false, "Expected JsonReaderException was not thrown.");
+            }
+            catch (JsonReaderException ex)
+            {
+                Assert.Equal(expectedlineNumber, ex.LineNumber);
+                Assert.Equal(expectedPosition, ex.Position);
+            }
+        }
+
```
KrzysztofCwalina: The new tests parse byte arrays. We need tests that do the same for ReadOnlySeuences. Is that what the TODO comment refers to?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
davidfowl: No I think it should just throw.

JamesNK: There is a feature in Json.NET where you can attempt to catch an exception and retry. I found it caused more problems that it was worth.

I'd throw.

tests/System.Text.JsonLab.Tests/JsonReaderTests.cs
ahsonkhan: Yes.

tests/System.Text.JsonLab.Tests/JsonReaderTests.cs
ahsonkhan: Good point, not it doesn't.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
ahsonkhan: > Should the reader be designed just like our encoders, i.e. they don't throw when they encounter bad data, but rather return with some "bad data" status. This allows the caller to inspect the bytesConsumed value to see where the bad data starts, fixup, and possibly keep going. 

It would definitely provide more flexibility to the user, at the cost of additional complexity. Since we don't have a destination here, we have three states: Done, Invalid, NeedsMore. If throwing for invalid is the expected behaviour, having an OperationStatus instead of a bool for the two states seems unnecessary.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -266,24 +270,61 @@ private void ReadFirstToken(byte first)
...
-                ConsumeValue(first);
+                _isSingleValue = true;
+                if (ConsumeValue(first))
+                {
+                    if (CurrentIndex >= (uint)_buffer.Length)
+                    {
+                        TokenType = JsonTokenType.Value;
+                        return true;
+                    }
+#if UseInstrumented
```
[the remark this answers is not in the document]
ahsonkhan: I am not sure I follow. The entire point of the ifdef was to track more info for the actual exception. In most cases, the ifdefs surround codepaths where both readers throw an exception, but one of them is the trigger to re-process the json before throwing. Why would this particular ifdef unique?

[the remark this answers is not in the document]
KrzysztofCwalina: There are some differences in the readers that are not in error code paths (e.g. incrementing the position, line #). These need to be ifdefed as runtime checks would impact perf. But here, you are already in the error code path. The perf does not matter at that point (throwing exceptions is very slow), and so I don't think we should complicate the code and use ifdefs in error code paths.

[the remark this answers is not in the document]
ahsonkhan: I addressed this by adding an overload to JsonThrowHelper.ThrowJsonReaderException that accepts Instrumented.Utf8JsonReader, in an attempt to cleanly pass in the custom error message string. Let me know if you have feedback around that approach.

ahsonkhan:

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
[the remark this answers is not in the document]
ahsonkhan: I had to revert this change since it caused a significant performance regression (roughly 2x slower across the board). I believe it had to do with passing around a formatted string (which contained property/field values) so the methods were likely not getting optimized as heavily.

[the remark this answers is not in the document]
ahsonkhan: @AndyAyersMS, do you know what is causing the regression? Is there anything that the JIT can do here or are there tricks/workarounds that you know to reduce the perf hit?

This was the commit that caused a regression: https://github.com/dotnet/corefxlab/pull/2496/commits/c7c45b1893f6be72430c1d065edb4d490380778a

Essentially, moving from calling a basic throw helper: `JsonThrowHelper.ThrowJsonReaderException(ref this);` to something like `JsonThrowHelper.ThrowJsonReaderException(ref this, $"Depth of {Depth} within an object is larger than max depth of {MaxDepth}");`

The assumption was, since the throw helper would only be called in exceptional cases, it should have no negative affect on performance for the common case. That assumption ended up not being true.

src/System.Text.JsonLab/System/Text/Json/JsonReaderState.cs
```diff
@@ -0,0 +1,55 @@
...
+        {
+            return _containerMask == other._containerMask &&
+                   EqualityComparer<Stack<bool>>.Default.Equals(_stack, other._stack) &&
+                   _inObject == other._inObject &&
+                   _depth == other._depth &&
+                   _tokenType == other._tokenType &&
+                   _searchedNextLast == other._searchedNextLast;
+        }
+
+        public override int GetHashCode()
```
ahsonkhan: This implementation was generated automatically by VS. What's the common pattern for implementing GetHashCode?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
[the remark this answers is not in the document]
mikedn: Have you tried passing a constant string rather than a formatted one?

src/System.Text.JsonLab/System/Text/Json/JsonReaderState.cs
```diff
@@ -0,0 +1,55 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace System.Text.JsonLab
+{
+    public struct JsonReaderState : IEquatable<JsonReaderState>
```
ahsonkhan: I compare the JsonReaderState to the default value to infer if the caller passed in a state (retry attempt). This type doesn't necessarily need to be IEquatable. The equality operations might not be necessary (if I can find a way to remove `_isRetry`.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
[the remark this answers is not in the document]
AndyAyersMS: My guess is that the formatting creates enough register pressure that code paths outside this call are impacted. But you'd have to look at profiles or diffs to be sure. If it is really 2x slower it should readily show up in profiling.

If this is the issue, you need to defer formatting. Either pass the format string or pass an enum along with the values needed for message, and do all the formatting in the throw helper(s). This is more or less what the [CoreCLR throw helper](https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/ThrowHelper.cs) pattern does.

src/System.Text.JsonLab/System/Text/Json/JsonReaderState.cs
KrzysztofCwalina: Going forward, we have APIs to create hashcodes, but these are not yet in NetStandard. 
This type will probably not be used in hash tables, and so I would just simplify what you have. For example, I would remove _stack and the bools (_inObject and _searchedNextLast). The stack is too complex and bools probably don't add much entropy to the hash.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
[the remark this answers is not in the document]
ahsonkhan: > My guess is that the formatting creates enough register pressure that code paths outside this call are impacted.

That's what it looks like:
**Before vs after comparison:** https://www.diffchecker.com/qo1taY5Q

> ave you tried passing a constant string rather than a formatted one?

That doesn't have any impact on the disassembly (so likely won't cause the regression).

> If this is the issue, you need to defer formatting. Either pass the format string or pass an enum along with the values needed for message, and do all the formatting in the throw helper(s). This is more or less what the CoreCLR throw helper pattern does.

Sounds good, I will give that a shot. Thanks!

When deferring formatting, how do I pass in arbitrary arguments needed for formatting the string properly without a regression? If the throw helper accepts `params` there are extra `call` and `mov` instructions within the caller. Alternatively, I could add several overloads with specific args (results in N move instructions where N is number of args), but only if the jump statement is true.

PR #2498, opened 2018-09-26
Add a skip delimiter overload to TryReadTo

JeremyKuhne:
This allows you to handle escaping. Unescaping would be considered the callers responsibility. We could combine `TryReadToInternal` if we introduce additional condition checks.

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
@@ -48,37 +48,38 @@ namespace System.Buffers.Reader
         /// </summary>
         /// <param name="span">The read data, if any.</param>
         /// <param name="delimiter">The delimiter to look for.</param>
-        /// <param name="skipDelimiter">If found prior to <paramref name="delimiter"/> it will skip that occurence.</param>
+        /// <param name="delimiterEscape">If found prior to <paramref name="delimiter"/> it will skip that occurence.</param>
         /// <param name="advancePastDelimiter">True to move past the <paramref name="delimiter"/> if found.</param>
         /// <returns>True if the <paramref name="delimiter"/> was found.</returns>
-        public bool TryReadTo(out ReadOnlySpan<T> span, T delimiter, T skipDelimiter, bool advancePastDelimiter = true)
+        public bool TryReadTo(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape, bool advancePastDelimiter = true)
         {
             ReadOnlySpan<T> remaining = UnreadSpan;
             int index = remaining.IndexOf(delimiter);
 
             if (index != -1)
             {
-                if (index > 0 && remaining[index - 1].Equals(skipDelimiter))
+                if (index > 0 && remaining[index - 1].Equals(delimiterEscape))
                 {
-                    // This delimiter should be skipped, go down the slow path
-                    return TryReadToSlow(out span, delimiter, skipDelimiter, index, advancePastDelimiter);
+                    // This delimiter might be skipped, go down the slow path
+                    return TryReadToSlow(out span, delimiter, delimiterEscape, advancePastDelimiter);
                 }
                 span = index == 0 ? default : remaining.Slice(0, index);
                 Advance(index + (advancePastDelimiter ? 1 : 0));
                 return true;
             }
 
-            return TryReadToSlow(out span, delimiter, skipDelimiter, remaining.Length, advancePastDelimiter);
+            return TryReadToSlow(out span, delimiter, delimiterEscape, advancePastDelimiter);
         }
 
-        private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T skipDelimiter, int skip, bool advancePastDelimiter)
+        private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape, bool advancePastDelimiter)
         {
-            if (!TryReadToInternal(out ReadOnlySequence<T> sequence, delimiter, skipDelimiter, advancePastDelimiter, skip))
+            if (!TryReadTo(out ReadOnlySequence<T> sequence, delimiter, delimiterEscape, advancePastDelimiter))
             {
                 span = default;
                 return false;
             }
 
+            Debug.Assert(sequence.Length > 0);
             span = sequence.IsSingleSegment ? sequence.First.Span : sequence.ToArray();
             return true;
         }
@@ -95,7 +96,7 @@ namespace System.Buffers.Reader
             return TryReadToInternal(out sequence, delimiter, advancePastDelimiter);
         }
 
-        public bool TryReadToInternal(out ReadOnlySequence<T> sequence, T delimiter, bool advancePastDelimiter, int skip = 0)
+        private bool TryReadToInternal(out ReadOnlySequence<T> sequence, T delimiter, bool advancePastDelimiter, int skip = 0)
         {
             BufferReader<T> copy = this;
             if (skip > 0)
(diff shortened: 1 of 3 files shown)


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -42,6 +42,47 @@ private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, int skip, bool
...
+        /// <summary>
+        /// Try to read everything up to the given <paramref name="delimiter"/>, ignoring delimiters that are
+        /// preceeded by <paramref name="skipDelimiter"/>.
+        /// </summary>
+        /// <param name="span">The read data, if any.</param>
+        /// <param name="delimiter">The delimiter to look for.</param>
+        /// <param name="skipDelimiter">If found prior to <paramref name="delimiter"/> it will skip that occurence.</param>
+        /// <param name="advancePastDelimiter">True to move past the <paramref name="delimiter"/> if found.</param>
+        /// <returns>True if the <paramref name="delimiter"/> was found.</returns>
+        public bool TryReadTo(out ReadOnlySpan<T> span, T delimiter, T skipDelimiter, bool advancePastDelimiter = true)
```
KrzysztofCwalina: I think optionalDelimiterPrefix might be a better name

tests/System.Buffers.ReaderWriter.Tests/Reader_SkipDelimiter.cs
```diff
@@ -0,0 +1,84 @@
...
+            byte[] expected = Encoding.UTF8.GetBytes("This is our \\\"understanding\\\"");
+            ReadOnlySequence<byte> bytes = BufferFactory.CreateUtf8("This is our \\\"understanding\\\"\" you see.");
+            BufferReader<byte> reader = new BufferReader<byte>(bytes);
+            Assert.True(reader.TryReadTo(out ReadOnlySpan<byte> span, (byte)'"', (byte)'\\', advancePastDelimiter: true));
+            Assert.Equal(expected, span.ToArray());
+            Assert.True(reader.IsNext((byte)' '));
+            Assert.Equal(30, reader.Consumed);
+
+            reader = new BufferReader<byte>(bytes);
+            Assert.True(reader.TryReadTo(out span, (byte)'"', (byte)'\\', advancePastDelimiter: false));
```
KrzysztofCwalina: this API will probably be used for reading strings mainly. It might be better to take UnicodeScalar, especially if we have support from the language for crating scalar literals. Casing chars to bytes does not scale beyond ascii and the syntax is ugly. @GrabYourPitchforks

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -90,6 +131,76 @@ public bool TryReadToInternal(out ReadOnlySequence<T> sequence, T delimiter, boo
...
+        /// <param name="delimiter">The delimiter to look for.</param>
+        /// <param name="skipDelimiter">If found prior to <paramref name="delimiter"/> it will skip that occurence.</param>
+        /// <param name="advancePastDelimiter">True to move past the <paramref name="delimiter"/> if found.</param>
+        /// <returns>True if the <paramref name="delimiter"/> was found.</returns>
+        public bool TryReadTo(out ReadOnlySequence<T> sequence, T delimiter, T skipDelimiter, bool advancePastDelimiter = true)
+        {
+            return TryReadToInternal(out sequence, delimiter, skipDelimiter, advancePastDelimiter);
+        }
+
+        public bool TryReadToInternal(out ReadOnlySequence<T> sequence, T delimiter, T skipDelimiter, bool advancePastDelimiter, int skip = 0)
```
ahsonkhan: This (and a couple other) internal APIs shouldn't be marked as public.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -90,6 +131,76 @@ public bool TryReadToInternal(out ReadOnlySequence<T> sequence, T delimiter, boo
...
+
+            ReadOnlySpan<T> remaining = UnreadSpan;
+            bool shouldSkipDelimiter = false;
+
+            while (!End)
+            {
+                int index = remaining.IndexOf(delimiter);
+                if (index != -1)
+                {
+                    if ((shouldSkipDelimiter && index == 0) || (index > 0 && remaining[index - 1].Equals(skipDelimiter)))
```
ahsonkhan: If I am understanding this correctly, there could be an issue here:

Suppose this is your input: `abc\"de"`

The initial search finds index = 4. and we have the skipDelimiter (`\`) at index 3. So we call `TryReadToSlow`.

We then advance by 3 (`skip - 1`), check if we reached skipDelimiter, and advance by 1, if we have. Now, remaining contains: `"de"`
Searching for quotation within the loop:
`shouldSkipDelimiter = false`
`index = 0`

And now we return true saying we found the string, `abc\`, which is incorrect.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -90,6 +131,76 @@ public bool TryReadToInternal(out ReadOnlySequence<T> sequence, T delimiter, boo
...
+            BufferReader<T> copy = this;
+            if (skip > 0)
+            {
+                if (skip > 1)
+                {
+                    Advance(skip - 1);
+                }
+
+                TryRead(out T prior);
+                if (prior.Equals(skipDelimiter) && !End)
```
ahsonkhan: Some comments here explaining why we are re-doing this check would be useful. We only do this if we didn't find a quotation in our first search and pass `remaining.Length` as skip instead of the found `index`.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -90,6 +131,76 @@ public bool TryReadToInternal(out ReadOnlySequence<T> sequence, T delimiter, boo
...
+                    {
+                        // This delimiter should be skipped, move past it.
+                        shouldSkipDelimiter = false;
+                        Advance(index + 1);
+                        remaining = UnreadSpan;
+                        continue;
+                    }
+
+                    // Found the delimiter. Move to it, slice, then move past it.
+                    sequence = Sequence.Slice(copy.Position, Position);
```
ahsonkhan: What happens if the skipDelimiter character itself has been escaped?
`abc\\"def"` should return `abc\\`

That is why I had to keep track of number of `skipDelimiter` characters seen and mod that by 2:
https://github.com/dotnet/corefxlab/blob/ad31432950ab70b3ddb51ac982d36bb56da64f1f/src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs#L385

This is a decent test case that I used:
https://github.com/dotnet/corefxlab/blob/ad31432950ab70b3ddb51ac982d36bb56da64f1f/tests/System.Text.JsonLab.Tests/JsonReaderTests.cs#L295

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
ahsonkhan: For short variable names, delimiterEscape is fine.
If we are fine with a longer name: instead of prefixed with, how about `ignoreDelimiterIfEscapedWith`

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: Whoops- good catch- fixing.

tests/System.Buffers.ReaderWriter.Tests/Reader_SkipDelimiter.cs
JeremyKuhne: Performance tanks for non ASCII and most of the time (90% plus?) parsing involves ASCII delimiters. We can always create full Unicode specific ones that might even call to this in some cases.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: I'll make sure to test this case.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: I'll add the tests and update if needed.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: Stupid mistake on my part, I didn't actually copy the "move it forward" part.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: I've got the fix for multiple escape bytes- took awhile to get it down as there was a bug in our test implementation of `ReadOnlySequenceSegment<byte>` that I accidentally stumbled over.

JeremyKuhne:
@stephentoub this is the latest PR for BufferReader

PR #2504, opened 2018-09-29
Add instrument property to JsonReader and remove Instrumented type

ahsonkhan:
There is a ~5-15% regression (Instrument = false) due to overhead of new if checks and fields. but we no longer have the #ifdef code complexity.

If Instrument is set to true, there is an additional 20% regression.

**For example, for reading HelloWorld:**
Instrument = false
- 64 ns -> 74 ns (compact)
- 75 ns -> 82 ns (formatted)

Instrument = true
- 74 ns -> 92 ns (compact)
- 82 ns -> 111 ns (formatted)

I did not avoid incrementing `_position` in all places (when Instrument is false), since an if check is likely slower than just incrementing a field.


src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -40,6 +36,8 @@ public int MaxDepth
             }
         }
 
+        public bool Instrument { get; set; }
```
KrzysztofCwalina: Does it need to be public?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -197,9 +185,7 @@ private void StartObject()
             if (Depth > MaxDepth)
                 ThrowJsonReaderException(ref this, ExceptionResource.ObjectDepthTooLarge);
 
-#if UseInstrumented
             _position++;
```
KrzysztofCwalina: Why do you always increment the position now?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -383,16 +377,20 @@ private bool ReadSingleSegment()
...
             {
+                int prevCurrentIndex = CurrentIndex;
+                int prevPosition = _position;
+                JsonTokenType prevTokenType = TokenType;
                 if (ConsumeNextToken(first))
                 {
-                    _searchedNextLast = true;
                     return true;
                 }
+                CurrentIndex = prevCurrentIndex;
```
KrzysztofCwalina: Would it make sense to not store/manage position, but rather store the CurrentIndex of the last newline and then compute position by subtracting the stored value from the new CurrentIndex. This way you only need to increment CurrentIndex

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
ahsonkhan: I believe so. How else would the caller indicate they want line number/position info to be tracked?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
ahsonkhan: > I did not avoid incrementing `_position` in all places (when Instrument is false), since an if check is likely slower than just incrementing a field.

Alternatively, I would have `if (Instrumented) _position++;` everywhere. Essentially, if instrumented is false, the value `_position` is meaningless. Should I add if checks around all `_position`  increments?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
ahsonkhan: We would still need some sort of running index to implement that correctly. CurrentIndex is reset back to 0 when a new span segment is passed in so we can't subtract last newline index from CurrentIndex to get the position.

PR #2505, opened 2018-09-29
Another attempt on cleaning UnixI2cDevice

shaggygi:
Moved to own Unix directory and separated classes/structs/enums

I didn't use Spans for now as I wasn't completely sure on the related pointer code.  Plan to review/add later.


joperezr:
Thanks for putting this up! sorry that it's taking us some time to take a look at it, but we are now looking specifically into I2C and SPI protocols so we will get to this PR very soon.

PR #2506, opened 2018-10-02
Use new BufferReader.TryReadTo API to replace JsonReader string escape logic

ahsonkhan:
~16% regression, originally.

I tried to optimize the TryReadTo implementation to drop the regression down to ~10% regression for the common case. I think we can find more ways to optimize it.

image

cc @KrzysztofCwalina

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
@@ -56,24 +56,26 @@ namespace System.Buffers.Reader
             ReadOnlySpan<T> remaining = UnreadSpan;
             int index = remaining.IndexOf(delimiter);
 
-            if (index != -1)
+            if (index < 0)
             {
-                if (index > 0 && remaining[index - 1].Equals(delimiterEscape))
-                {
-                    // This delimiter might be skipped, go down the slow path
-                    return TryReadToSlow(out span, delimiter, delimiterEscape, advancePastDelimiter);
-                }
-                span = index == 0 ? default : remaining.Slice(0, index);
+                goto SlowPath;
+            }
+
+            if (index == 0 || !remaining[index - 1].Equals(delimiterEscape))
+            {
+                span = remaining.Slice(0, index);
                 Advance(index + (advancePastDelimiter ? 1 : 0));
                 return true;
             }
 
-            return TryReadToSlow(out span, delimiter, delimiterEscape, advancePastDelimiter);
+        SlowPath:
+            // This delimiter might be skipped, go down the slow path
+            return TryReadToSlow(out span, delimiter, delimiterEscape, index, advancePastDelimiter);
         }
 
-        private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape, bool advancePastDelimiter)
+        private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape, int index, bool advancePastDelimiter)
         {
-            if (!TryReadTo(out ReadOnlySequence<T> sequence, delimiter, delimiterEscape, advancePastDelimiter))
+            if (!TryReadToSlow(out ReadOnlySequence<T> sequence, delimiter, delimiterEscape, index, advancePastDelimiter))
             {
                 span = default;
                 return false;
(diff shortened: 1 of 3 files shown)


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -56,24 +56,26 @@ public bool TryReadTo(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape,
...
             }
 
-            return TryReadToSlow(out span, delimiter, delimiterEscape, advancePastDelimiter);
+        SlowPath:
+            // This delimiter might be skipped, go down the slow path
+            return TryReadToSlow(out span, delimiter, delimiterEscape, index, advancePastDelimiter);
         }
 
-        private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape, bool advancePastDelimiter)
+        private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape, int index, bool advancePastDelimiter)
```
ahsonkhan: Trying to avoid re-doing the first IndexOf call.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -84,6 +86,102 @@ private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEsc
...
+                    if (advancePastDelimiter)
+                    {
+                        Advance(1);
+                    }
+                    return true;
+                }
+                else
+                {
+                    // No delimiter, need to check the end of the span for odd number of escapes then advance
+                    if (remaining[remaining.Length - 1].Equals(delimiterEscape))
```
ahsonkhan: Separating out the common case where this will be false before looping.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -84,6 +86,102 @@ private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEsc
...
+                                break;
+                        }
+                        if (i < 0 && priorEscape)
+                        {
+                            // Started and ended with escape, increment once more
+                            escapeCount++;
+                        }
+                        escapeCount += index - 2 - i;
+
+                        if ((escapeCount & 1) != 0)
```
ahsonkhan: Using bitwise operation instead of modulo.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -84,6 +86,102 @@ private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEsc
...
+                        {
+                            if (!remaining[i].Equals(delimiterEscape))
+                                break;
+                        }
+                        if (i < 0 && priorEscape)
+                        {
+                            // Started and ended with escape, increment once more
+                            escapeCount++;
+                        }
+                        escapeCount += index - 2 - i;
```
ahsonkhan: Avoiding incrementing escapeCount for each iteration of the loop above. Instead doing it once at the end.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -84,6 +86,102 @@ private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEsc
...
         }
 
+        private bool TryReadToSlow(out ReadOnlySequence<T> sequence, T delimiter, T delimiterEscape, int index, bool advancePastDelimiter)
+        {
+            BufferReader<T> copy = this;
+
+            ReadOnlySpan<T> remaining = UnreadSpan;
+            bool priorEscape = false;
+
+            do
```
ahsonkhan: Changed to do/while instead of while and calling IndexOf at the end.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -84,6 +86,102 @@ private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEsc
...
+            {
+                if (index >= 0)
+                {
+                    if (index == 0 && priorEscape)
+                    {
+                        // We were in the escaped state, so skip this delimiter
+                        priorEscape = false;
+                        Advance(index + 1);
+                        goto Continue;
```
ahsonkhan: Do we need to update `remaining` before continuing to the next iteration of the loop?

The same question would apply to the existing TryReadTo method:
https://github.com/dotnet/corefxlab/blob/b89fc2d9b404131b36ef6eb41320d638d4e4458c/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs#L161

If yes, we are missing a test case.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -56,24 +56,26 @@ public bool TryReadTo(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape,
...
-                if (index > 0 && remaining[index - 1].Equals(delimiterEscape))
-                {
-                    // This delimiter might be skipped, go down the slow path
-                    return TryReadToSlow(out span, delimiter, delimiterEscape, advancePastDelimiter);
-                }
-                span = index == 0 ? default : remaining.Slice(0, index);
+                goto SlowPath;
+            }
+
+            if (index == 0 || !remaining[index - 1].Equals(delimiterEscape))
```
JeremyKuhne: Wouldn't it be clearer to combine the if's here?

``` C#
if (index == 0 || (index > 0 && !remaining[index - 1].Equals(delimiterEscape)))
```
```

Comment (ahsonkhan) [reply src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs]: Yes, that would work nicely and avoid the multiple return statements.

The goto was there to avoid multiple return points.

Comment (ahsonkhan) [reply src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs]: nit: I would invert the order of the checks though, since index > 0 would be the more common case

``` C#
if ((index > 0 && !remaining[index - 1].Equals(delimiterEscape)) || index == 0)
```

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: Yes, we do, this is where you should actually write the test to try and break this. ;)  At the end of TryReadTo_SkipDelimiter, add the following:

``` C#
            // Delimiter starts second segment.
            bytes = BufferFactory.CreateUtf8("^", "|a|b");
            reader = new BufferReader<byte>(bytes);
            Assert.True(reader.TryReadTo(out span, (byte)'|', (byte)'^', advancePastDelimiter: true));
            Assert.Equal(Encoding.UTF8.GetBytes("^|a"), span.ToArray());
            Assert.True(reader.IsNext((byte)'b'));
            Assert.Equal(4, reader.Consumed);
```

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: Sure, can you follow up?

ahsonkhan: Yep, will do.

ahsonkhan: Done: https://github.com/dotnet/corefxlab/pull/2514

Issue #1816, opened 2017-10-04
Tensor should be in System.Numerics.Tensors namespace


ericstj:
This was feedback from the API review.  Folks don't want us to put this in the root Numerics namespace.

KonH:
Hello! I made PR with proposed changes, please review it.

tannergooding:
This issue was moved to dotnet/corefx#34516

PR #2508, opened 2018-10-02
Move Tensor and related classes from System.Numerics to System.Numerics.Tensors

KonH:
Related to #1816


KrzysztofCwalina:
LGTM.

src/System.Numerics.Tensors/System/Numerics/ArrayTensorExtensions.cs
```diff
@@ -2,7 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
-namespace System.Numerics
+namespace System.Numerics.Tensors
```
eerhardt: Can you make a `Tensors` folder as well so the folders match the namespace?

ericstj:
In addition to the namespace we also follow the folder convention for namsepace, so these files should be moved into a Tensors subfolder folder.

KonH:
I added two more commits - now files moved and using directives in samples fixed.
@KrzysztofCwalina "LGTM" - do you mean that this PR is appropriate or may be I don't understand acroabbreviation and breaks some contribution rules? (unfortunately, I can't find detailed docs about it)

eerhardt:
> LGTM

Stands for "Looks good to me".  😄

PR #2514, opened 2018-10-03
Fix bug in TryReadTo, add tests, and optimize AdvanceToNextSpan for the common case

ahsonkhan:
Leftover feedback from https://github.com/dotnet/corefxlab/pull/2506#discussion_r222156375

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -187,6 +187,7 @@ namespace System.Buffers.Reader
                     break;
                 }
 
+                CurrentSpanIndex += remaining;
                 count -= remaining;
                 Debug.Assert(count >= 0);
 
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -2,6 +2,7 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
 namespace System.Buffers.Reader
@@ -133,18 +134,16 @@ namespace System.Buffers.Reader
         [MethodImpl(MethodImplOptions.NoInlining)]
         private void GetNextSpan()
         {
-            if (!Sequence.IsSingleSegment)
+            // If Sequence is a single segment, TryGet will return false
+            SequencePosition previousNextPosition = _nextPosition;
+            while (Sequence.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
             {
-                SequencePosition previousNextPosition = _nextPosition;
-                while (Sequence.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
+                if (memory.Length > 0)
                 {
-                    _currentPosition = previousNextPosition;
                     CurrentSpan = memory.Span;
+                    _currentPosition = previousNextPosition;
                     CurrentSpanIndex = 0;
-                    if (CurrentSpan.Length > 0)
-                    {
-                        return;
-                    }
+                    return;
                 }
             }
             _moreData = false;
@@ -177,21 +176,24 @@ namespace System.Buffers.Reader
         [MethodImpl(MethodImplOptions.NoInlining)]
         private void AdvanceToNextSpan(int count)
         {
-            while (!End && count > 0)
+            while (_moreData)
             {
-                if (CurrentSpanIndex < CurrentSpan.Length - count)
+                int remaining = CurrentSpan.Length - CurrentSpanIndex;
+
+                if (remaining > count)
                 {
                     CurrentSpanIndex += count;
                     count = 0;
                     break;
                 }
 
-                int remaining = (CurrentSpan.Length - CurrentSpanIndex);
-
-                CurrentSpanIndex += remaining;
                 count -= remaining;
+                Debug.Assert(count >= 0);
 
                 GetNextSpan();
+
+                if (count == 0)
+                    break;
             }
 
             if (count != 0)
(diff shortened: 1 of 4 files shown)


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -177,21 +176,24 @@ public void Advance(int count)
...
+                if (remaining > count)
                 {
                     CurrentSpanIndex += count;
                     count = 0;
                     break;
                 }
 
-                int remaining = (CurrentSpan.Length - CurrentSpanIndex);
-
-                CurrentSpanIndex += remaining;
```
ahsonkhan: GetNextSpan resets CurrentSpanIndex anyway.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -177,21 +176,24 @@ public void Advance(int count)
...
                 count -= remaining;
+                Debug.Assert(count >= 0);
 
                 GetNextSpan();
+
+                if (count == 0)
```
ahsonkhan: 3 comparisons (if checks) instead of 4 for the common case (i.e. we advance to the next segment only once).

**Before:**
while (!End && count > 0) <= true and true
while (!End && count > 0) <= true and true
break

**After:**
while (_moreData) <= true
if (count == 0) <= false
while (_moreData) <= true
break

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -133,18 +134,16 @@ public bool TryRead(out T value)
         [MethodImpl(MethodImplOptions.NoInlining)]
         private void GetNextSpan()
         {
-            if (!Sequence.IsSingleSegment)
+            // If Sequence is a single segment, TryGet will return false
```
ahsonkhan: It is unlikely for GetNextSpan to be called when Sequence contains a single segment only. No?

JeremyKuhne: It will always get called for single segment- I explicitly added this when measuring.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -133,18 +134,16 @@ public bool TryRead(out T value)
...
+            SequencePosition previousNextPosition = _nextPosition;
+            while (Sequence.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
             {
-                SequencePosition previousNextPosition = _nextPosition;
-                while (Sequence.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
+                if (memory.Length > 0)
                 {
-                    _currentPosition = previousNextPosition;
                     CurrentSpan = memory.Span;
+                    _currentPosition = previousNextPosition;
```
JeremyKuhne: This will break. We should have an empty span when we get to the end. Current position won't get updated either. Can you add tests to ensure that behavior does not change in the end state? Empty sequences at the end are particularly concerning.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: Which BufferReader API is calling GetNextSpan for single segment? This method has 3 references and all are for edge cases.

The only one that might be relevant is TryRead(), but I am not sure if we need to care about that case either:
https://github.com/dotnet/corefxlab/blob/ce1ee0c85afdefaff162e97db8c5a7caf6b4358b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs#L124

JeremyKuhne: `AdvanceToNextSpan` calls it from `Advance` _every_ time it runs out of space in the current segment.

ahsonkhan: > `AdvanceToNextSpan` calls it from `Advance` _every_ time it runs out of space in the current segment.

Yes, but in that case `Sequence.IsSingleSegment` would be false anyway, so why bother checking? The only time that would matter is if you had a single segment sequence, and you advanced past the end (and only in that case do we go through an extra iteration of the loop in `AdvanceToNextSpan` before throwing).

Where do we call `GetNextSpan` when `Sequence.IsSingleSegment` is true?

JeremyKuhne: > Yes, but in that case Sequence.IsSingleSegment would be false anyway

How? We'll run out of space in a single segment and try to get the next. There isn't any magic to the Advance- it _always_ tries to get more segments. We could move the `IsSingleSegment` check there, but that seems pointless.

ahsonkhan: Assume you have a singleSegment sequence = [a, b, c]

Advance(1) and Advance(2) will never call AdvanceToNextSpan since `if (CurrentSpanIndex < CurrentSpan.Length - count)` is true.

Advance(4) <= should throw, and since that is the exception case, I don't think we need to avoid the `Sequence.TryGet` call there.

Advance(3) (or more generally, Advance(singleSegment.Length) is the only case where the IsSingleSegment check helps avoid the unnecessary `Sequence.TryGet` call.

I will revert this change since I don't have strong data to support it, but it seems to me we are always doing the check for IsSingleSegment within GetNextSpan when it will rarely be true. In error-free Advancing of a single segment, the GetNextSpan will only be called when you want to advance past the end. So to avoid that, we are currently doing the IsSingleSegment check for all Advance calls that straddle segments in the multi-segment case.

ahsonkhan:

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: You're right- I didn't think about the fact that is an error case. The code here has likely changed a number of times since I first introduced that check.

PR #2516, opened 2018-10-03
Remove copying from IsNext.

### LINKED ISSUE #2515: Unconstrained stackalloc based on input that the caller passed
The following public method calls IsNextSlow with the span that the user passed in which could be of arbitrary length.
```C#
public bool IsNext(ReadOnlySpan<T> next, bool advancePast = false)
```

https://github.com/dotnet/corefxlab/blob/ce1ee0c85afdefaff162e97db8c5a7caf6b4358b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs#L723

JeremyKuhne:
Fixes #2515

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
@@ -717,19 +717,51 @@ namespace System.Buffers.Reader
         [MethodImpl(MethodImplOptions.NoInlining)]
         private unsafe bool IsNextSlow(ReadOnlySpan<T> next, bool advancePast)
         {
-            // Call PeekSlow directly since we know there isn't enough space.
-            Debug.Assert(UnreadSpan.Length < next.Length);
+            ReadOnlySpan<T> currentSpan = UnreadSpan;
 
-            T* t = stackalloc T[next.Length];
-            ReadOnlySpan<T> peek = PeekSlow(new Span<T>(t, next.Length));
+            // We should only come in here if we need more data than we have in our current span
+            Debug.Assert(currentSpan.Length < next.Length);
 
-            if (next.SequenceEqual(peek))
+            int length = next.Length;
+            SequencePosition nextPosition = _nextPosition;
+
+            while (true)
             {
-                if (advancePast)
+                if (!next.StartsWith(currentSpan))
                 {
-                    Advance(next.Length);
+                    break;
                 }
-                return true;
+
+                if (next.Length == currentSpan.Length)
+                {
+                    // Fully matched
+                    if (advancePast)
+                    {
+                        Advance(length);
+                    }
+                    return true;
+                }
+                next = next.Slice(currentSpan.Length);
+
+                // Need to check the next segment
+                while (true)
+                {
+                    if (!Sequence.TryGet(ref nextPosition, out ReadOnlyMemory<T> nextSegment, advance: true))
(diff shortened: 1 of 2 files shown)


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -717,19 +717,51 @@ public bool IsNext(ReadOnlySpan<T> next, bool advancePast = false)
...
+                        // Nothing left
+                        return false;
+                    }
+
+                    if (nextSegment.Length > 0)
+                    {
+                        currentSpan = nextSegment.Span;
+                        if (currentSpan.Length > next.Length)
+                        {
+                            currentSpan = currentSpan.Slice(0, next.Length);
```
ahsonkhan: This slice and then checking if next.Length == currentSpan.Length within the outerloop on success seems unnecessary.

Instead of slicing, why not have the outerloop break when next.StartsWith(currentSpan) returns true.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -717,19 +717,51 @@ public bool IsNext(ReadOnlySpan<T> next, bool advancePast = false)
...
-                    Advance(next.Length);
+                    break;
+                }
+
+                if (next.Length == currentSpan.Length)
+                {
+                    // Fully matched
+                    if (advancePast)
+                    {
+                        Advance(length);
```
ahsonkhan: This Advance will traverse the sequence again, no? Can we avoid that by advancing when we get the next segment?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: If we have `"ABCD"` left in `currentSpan`, it won't start `next` if it has `"AB"` left in it, but it should match.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: We can't advance the cursor until we know we've got a full match. We could try and slice off of `nextPosition`, but it would be really convoluted and I suspect it wouldn't save much. I want to be careful about making this code too complicated for small gains.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -717,19 +717,51 @@ public bool IsNext(ReadOnlySpan<T> next, bool advancePast = false)
...
-            T* t = stackalloc T[next.Length];
-            ReadOnlySpan<T> peek = PeekSlow(new Span<T>(t, next.Length));
+            // We should only come in here if we need more data than we have in our current span
+            Debug.Assert(currentSpan.Length < next.Length);
 
-            if (next.SequenceEqual(peek))
+            int length = next.Length;
+            SequencePosition nextPosition = _nextPosition;
+
+            while (true)
```
ahsonkhan: nit: Change to `while (next.StartsWith(currentSpan))`

and then we can get rid of the `if (!next.StartsWith(currentSpan)) break;` within the loop

PR #2512, opened 2018-10-03
Measure multi-segment reading versus single-segment reading for various segment sizes

ahsonkhan:
As you would expect, the larger the segment size, the better the performance. 4 KB is a reasonable size. There is a 5-10% regression if you use a smaller size like 1KB.

**The multi-segment case is 50% slower than the single segment case, even with 8 KB buffers (ideally this should be under 10% as seen here: https://github.com/dotnet/corefxlab/issues/2511).**

Issue to try to improve performance: https://github.com/dotnet/corefxlab/issues/2509

``` ini

BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17754
Intel Core i7-6700 CPU 3.40GHz (Skylake), 1 CPU, 8 logical and 4 physical cores
.NET Core SDK=3.0.100-alpha1-009460
  [Host]     : .NET Core 2.1.3 (CoreCLR 4.6.26725.06, CoreFX 4.6.26725.05), 64bit RyuJIT
  Job-KGUDKK : .NET Core 2.1.3 (CoreCLR 4.6.26725.06, CoreFX 4.6.26725.05), 64bit RyuJIT

IterationCount=5  WarmupCount=3  

```
|                Method |  TestCase | segmentSize |         Mean |       Error |      StdDev |   Gen 0 | Allocated |
|---------------------- |---------- |------------ |-------------:|------------:|------------:|--------:|----------:|
|  **MultiSegmentSequence** | **Json400KB** |        **1000** | **1,514.121 us** | **282.4030 us** |  **73.3531 us** | **11.7188** |   **52056 B** |
|  **MultiSegmentSequence** | **Json400KB** |        **2000** | **1,446.905 us** | **288.6973 us** |  **74.9880 us** |  **5.8594** |   **26440 B** |
|  **MultiSegmentSequence** | **Json400KB** |        **4000** | **1,338.191 us** | **122.2628 us** |  **31.7573 us** |  **1.9531** |   **14072 B** |
|  **MultiSegmentSequence** | **Json400KB** |        **8000** | **1,370.076 us** | **540.2406 us** | **140.3255 us** |       **-** |    **7208 B** |
| **SingleSegmentSequence** | **Json400KB** |           **?** |   **899.445 us** |  **25.3657 us** |   **6.5887 us** |       **-** |       **0 B** |
[...]

Issue #2513, opened 2018-10-03
Investigate ways to optimize BufferReader.TryReadTo APIs for the common case


ahsonkhan:
We should aim to optimize the case where characters are not escaped, and the delimiter being searched for is within the next segment.

See https://github.com/dotnet/corefxlab/pull/2506 for potential optimizations as a starting point (validate and apply to all overloads).

https://github.com/dotnet/corefxlab/blob/e137eb61d704720755440de59f06f5d3823b4716/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs#L242
```C#
public bool TryReadTo(out ReadOnlySpan<T> span, T delimiter, T delimiterEscape, bool advancePastDelimiter = true);
public bool TryReadTo(out ReadOnlySequence<T> sequence, T delimiter, T delimiterEscape, bool advancePastDelimiter = true);
```

PR #2518, opened 2018-10-04
Add rich error messaging and re-entrant support for multi-segment json

ahsonkhan:
BufferReader does not provide a reset/recovery mechanism which results in the new re-entrant JsonReader tests to fail.

For example, if I see the start of a property name (i.e. `"`), I advance past the quote and call ConsumeString, which ends up calling `if (reader.TryReadTo(out ReadOnlySpan<byte> value, JsonConstants.Quote, (byte)'\\', advancePastDelimiter: true))`

However, if that returns false because we don't have all the data yet, I need to undo the advance I did (to move the cursor back before the first `"`).

To work around this, I have to copy the BufferReader before advancing to rollback the change. This results in significant perf regression. **How can I do this in a sane way while still using the BufferReader APIs?**

**TODO:** Add more test cases (especially for invalid data with isFinalBlock true/false and check that _position and _lineNumber are set correctly).

cc @KrzysztofCwalina, @JeremyKuhne 

**Multi-segment is 2.3x - 2.8x slower than single-segment.** Previously, it was 1.5x - 1.7x slower.

``` ini

BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17754
.NET Core SDK=3.0.100-alpha1-009460
  Job-WKNCWB : .NET Core 2.1.3 (CoreCLR 4.6.26725.06, CoreFX 4.6.26725.05), 64bit RyuJIT

IterationCount=5  WarmupCount=3  

```
**NOW:**

|                                            Method | IsDataCompact |   TestCase |            Mean |         Error |         StdDev | Scaled | ScaledSD |  Gen 0 | Allocated |
|-------------------------------------------------- |-------------- |----------- |----------------:|--------------:|---------------:|-------:|---------:|-------:|----------:|
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **BasicJson** |       **460.45 ns** |     **32.342 ns** |      **8.4008 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |  BasicJson |     1,492.66 ns |     28.555 ns |      7.4172 ns |   3.24 |     0.05 | 0.0076 |      40 B |
|                                                   |               |            |                 |               |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** | **HelloWorld** |        **72.93 ns** |      **2.589 ns** |      **0.6724 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True | HelloWorld |       468.30 ns |      7.660 ns |      1.9896 ns |   6.42 |     0.06 | 0.0086 |      40 B |
|                                                   |               |            |                 |               |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |   **Json400B** |       **647.95 ns** |     **92.892 ns** |     **24.1285 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |   Json400B |     1,770.08 ns |     53.784 ns |     13.9703 ns |   2.73 |     0.09 |      - |       0 B |
|                                                   |               |            |                 |               |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **Json400KB** |   **674,107.19 ns** | **16,212.198 ns** |  **4,211.0581 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |  Json400KB | 1,529,843.71 ns | 19,541.360 ns |  5,075.7955 ns |   2.27 |     0.01 |      - |     440 B |
|                                                   |               |            |                 |               |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |   **Json40KB** |    **62,572.23 ns** |  **1,253.872 ns** |    **325.6885 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |   Json40KB |   163,639.15 ns | 55,185.637 ns | 14,334.2640 ns |   2.62 |     0.21 |      - |      40 B |
|                                                   |               |            |                 |               |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |    **Json4KB** |     **4,886.29 ns** |  **1,397.445 ns** |    **362.9812 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |    Json4KB |    13,672.74 ns |  1,170.834 ns |    304.1198 ns |   2.81 |     0.18 | 0.0916 |     416 B |

|                Method |  TestCase | segmentSize |         Mean |         Error |      StdDev |   Gen 0 | Allocated |
|---------------------- |---------- |------------ |-------------:|--------------:|------------:|--------:|----------:|
|  **MultiSegmentSequence** | **Json400KB** |        **1000** | **2,277.556 us** |   **614.6865 us** | **159.6625 us** | **11.7188** |   **52056 B** |
|  **MultiSegmentSequence** | **Json400KB** |        **2000** | **2,284.465 us** | **2,527.7209 us** | **656.5661 us** |  **5.8594** |   **26440 B** |
|  **MultiSegmentSequence** | **Json400KB** |        **4000** | **1,870.355 us** |    **23.9744 us** |   **6.2273 us** |  **1.9531** |   **14072 B** |
|  **MultiSegmentSequence** | **Json400KB** |        **8000** | **1,974.490 us** |   **411.5267 us** | **106.8925 us** |       **-** |    **7208 B** |
| **SingleSegmentSequence** | **Json400KB** |           **?** |   **975.793 us** |   **326.4452 us** |  **84.7929 us** |       **-** |       **0 B** |
|  **MultiSegmentSequence** |  **Json40KB** |        **1000** |   **208.756 us** |    **67.1059 us** |  **17.4305 us** |  **0.9766** |    **4408 B** |
|  **MultiSegmentSequence** |  **Json40KB** |        **2000** |   **188.887 us** |    **22.7712 us** |   **5.9147 us** |  **0.4883** |    **2176 B** |
|  **MultiSegmentSequence** |  **Json40KB** |        **4000** |   **185.975 us** |    **11.2977 us** |   **2.9345 us** |  **0.2441** |    **1312 B** |
|  **MultiSegmentSequence** |  **Json40KB** |        **8000** |   **193.254 us** |    **73.5458 us** |  **19.1033 us** |       **-** |     **784 B** |
| **SingleSegmentSequence** |  **Json40KB** |           **?** |    **94.631 us** |    **34.9899 us** |   **9.0885 us** |       **-** |       **0 B** |
|  **MultiSegmentSequence** |   **Json4KB** |        **1000** |    **17.845 us** |     **1.1112 us** |   **0.2886 us** |       **-** |     **112 B** |
|  **MultiSegmentSequence** |   **Json4KB** |        **2000** |    **17.567 us** |     **1.7424 us** |   **0.4526 us** |       **-** |      **72 B** |
|  **MultiSegmentSequence** |   **Json4KB** |        **4000** |    **18.052 us** |     **6.4056 us** |   **1.6638 us** |       **-** |       **0 B** |
|  **MultiSegmentSequence** |   **Json4KB** |        **8000** |     **7.394 us** |     **0.4464 us** |   **0.1159 us** |       **-** |       **0 B** |
| **SingleSegmentSequence** |   **Json4KB** |           **?** |     **7.466 us** |     **0.2275 us** |   **0.0591 us** |       **-** |       **0 B** |

**BEFORE:**

|                                            Method | IsDataCompact |   TestCase |            Mean |           Error |         StdDev | Scaled | ScaledSD |  Gen 0 | Allocated |
|-------------------------------------------------- |-------------- |----------- |----------------:|----------------:|---------------:|-------:|---------:|-------:|----------:|
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **BasicJson** |       **418.88 ns** |      **22.2616 ns** |      **5.7824 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |  BasicJson |       953.00 ns |      24.8139 ns |      6.4453 ns |   2.28 |     0.03 | 0.0086 |      40 B |
|                                                   |               |            |                 |                 |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** | **HelloWorld** |        **71.34 ns** |       **0.7378 ns** |      **0.1916 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True | HelloWorld |       404.76 ns |     121.1304 ns |     31.4632 ns |   5.67 |     0.39 | 0.0091 |      40 B |
|                                                   |               |            |                 |                 |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |   **Json400B** |       **639.78 ns** |      **50.9402 ns** |     **13.2315 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |   Json400B |     1,119.68 ns |      28.5724 ns |      7.4216 ns |   1.75 |     0.03 |      - |       0 B |
|                                                   |               |            |                 |                 |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |  **Json400KB** |   **666,694.11 ns** |  **31,682.0832 ns** |  **8,229.3032 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |  Json400KB | 1,022,328.20 ns | 348,547.6500 ns | 90,533.9553 ns |   1.53 |     0.12 |      - |     440 B |
|                                                   |               |            |                 |                 |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |   **Json40KB** |    **64,289.09 ns** |   **4,376.6212 ns** |  **1,136.8111 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |   Json40KB |    91,749.92 ns |   6,529.2427 ns |  1,695.9465 ns |   1.43 |     0.03 |      - |      40 B |
|                                                   |               |            |                 |                 |                |        |          |        |           |
|              **ReaderSystemTextJsonLabSpanEmptyLoop** |          **True** |    **Json4KB** |     **4,800.40 ns** |     **269.5291 ns** |     **70.0092 ns** |   **1.00** |     **0.00** |      **-** |       **0 B** |
| ReaderSystemTextJsonLabMultiSpanSequenceEmptyLoop |          True |    Json4KB |     7,880.91 ns |     113.0941 ns |     29.3758 ns |   1.64 |     0.02 | 0.0916 |     416 B |

https://github.com/dotnet/corefxlab/pull/2512

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -28,11 +28,16 @@ namespace System.Buffers.Reader
             if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
             {
                 _moreData = true;
-                CurrentSpan = memory.Span;
-                if (CurrentSpan.Length == 0)
+                
+                if (memory.Length == 0)
                 {
+                    CurrentSpan = default;
                     // No space in the first span, move to one with space
                     GetNextSpan();
+                }
+                else
+                {
+                    CurrentSpan = memory.Span;
                 }
             }
             else
@@ -129,6 +134,63 @@ namespace System.Buffers.Reader
         }
 
         /// <summary>
+        /// Move the reader back the specified number of positions.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Retreat(int count)
+        {
+            if (count < 0)
+            {
+                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.count);
+            }
+
+            Consumed -= count;
+
+            if (CurrentSpanIndex >= count)
+            {
+                CurrentSpanIndex -= count;
+            }
+            else
+            {
+                // Current segment doesn't have enough space, scan backward through segments
+                RetreatToPreviousSpan(count);
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void RetreatToPreviousSpan(int count)
+        {
+            CurrentSpanIndex = 0;
+            Consumed = 0;
+            _currentPosition = Sequence.Start;
+            _nextPosition = _currentPosition;
+
+            if (Sequence.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
+            {
+                _moreData = true;
+                
+                if (memory.Length == 0)
+                {
+                    CurrentSpan = default;
+                    // No space in the first span, move to one with space
+                    GetNextSpan();
+                }
+                else
+                {
+                    CurrentSpan = memory.Span;
+                }
+            }
+            else
+            {
+                // No space in any spans and at end of sequence
+                _moreData = false;
+                CurrentSpan = default;
+            }
+
+            Advance(Consumed);
+        }
+
+        /// <summary>
         /// Get the next segment with available space, if any.
         /// </summary>
         [MethodImpl(MethodImplOptions.NoInlining)]
@@ -160,18 +222,18 @@ namespace System.Buffers.Reader
         /// Move the reader ahead the specified number of positions.
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void Advance(int count)
+        public void Advance(long count)
         {
             if (count < 0)
             {
-                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.length);
+                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.count);
             }
 
             Consumed += count;
 
             if (CurrentSpanIndex < CurrentSpan.Length - count)
             {
-                CurrentSpanIndex += count;
+                CurrentSpanIndex += (int)count;
             }
             else
             {
@@ -181,7 +243,7 @@ namespace System.Buffers.Reader
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private void AdvanceToNextSpan(int count)
+        private void AdvanceToNextSpan(long count)
         {
             while (_moreData)
             {
@@ -189,7 +251,7 @@ namespace System.Buffers.Reader
 
                 if (remaining > count)
                 {
-                    CurrentSpanIndex += count;
+                    CurrentSpanIndex += (int)count;
                     count = 0;
                     break;
                 }
@@ -206,7 +268,7 @@ namespace System.Buffers.Reader
 
             if (count != 0)
             {
-                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.length);
+                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.count);
             }
         }
 
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -160,6 +160,12 @@ namespace System.Buffers.Reader
         [MethodImpl(MethodImplOptions.NoInlining)]
         private void RetreatToPreviousSpan(long consumed)
         {
+            ResetReader();
+            Advance(consumed);
+        }
+
+        private void ResetReader()
+        {
             CurrentSpanIndex = 0;
             Consumed = 0;
             _currentPosition = Sequence.Start;
@@ -186,8 +192,6 @@ namespace System.Buffers.Reader
                 _moreData = false;
                 CurrentSpan = default;
             }
-
-            Advance(consumed);
         }
 
         /// <summary>
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -28,7 +28,7 @@ namespace System.Buffers.Reader
             if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
             {
                 _moreData = true;
-                
+
                 if (memory.Length == 0)
                 {
                     CurrentSpan = default;
@@ -137,7 +137,7 @@ namespace System.Buffers.Reader
         /// Move the reader back the specified number of positions.
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public void Retreat(int count)
+        public void Rewind(long count)
         {
             if (count < 0)
             {
@@ -148,17 +148,17 @@ namespace System.Buffers.Reader
 
             if (CurrentSpanIndex >= count)
             {
-                CurrentSpanIndex -= count;
+                CurrentSpanIndex -= (int)count;
             }
             else
             {
                 // Current segment doesn't have enough space, scan backward through segments
-                RetreatToPreviousSpan(count);
+                RetreatToPreviousSpan(Consumed);
             }
         }
 
         [MethodImpl(MethodImplOptions.NoInlining)]
-        private void RetreatToPreviousSpan(int count)
+        private void RetreatToPreviousSpan(long consumed)
         {
             CurrentSpanIndex = 0;
             Consumed = 0;
@@ -168,7 +168,7 @@ namespace System.Buffers.Reader
             if (Sequence.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
             {
                 _moreData = true;
-                
+
                 if (memory.Length == 0)
                 {
                     CurrentSpan = default;
@@ -187,7 +187,7 @@ namespace System.Buffers.Reader
                 CurrentSpan = default;
             }
 
-            Advance(Consumed);
+            Advance(consumed);
         }
 
         /// <summary>
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
@@ -141,7 +141,7 @@ namespace System.Buffers.Reader
                 else
                 {
                     // No delimiter, need to check the end of the span for odd number of escapes then advance
-                    if (remaining[remaining.Length - 1].Equals(delimiterEscape))
+                    if (remaining.Length > 0 && remaining[remaining.Length - 1].Equals(delimiterEscape))
                     {
                         int escapeCount = 1;
                         int i = remaining.Length - 2;
--- a/src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
+++ b/src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
@@ -193,8 +193,8 @@ namespace System.Text.JsonLab
                     TokenStartIndex++;
                     int prevCurrentIndex = CurrentIndex;
                     int prevPosition = _position;
-
-                    //reader.Advance(1);    //TODO: How do we revert this?
+                    BufferReader<byte> copy = reader;
+                    reader.Advance(1);
                     CurrentIndex++;
                     _position++;
                     if (ConsumePropertyName(ref reader))
@@ -204,6 +204,7 @@ namespace System.Text.JsonLab
                     CurrentIndex = prevCurrentIndex;
                     TokenType = JsonTokenType.StartObject;
                     _position = prevPosition;
+                    reader = copy;
                     return false;
                 }
             }
@@ -230,6 +231,7 @@ namespace System.Text.JsonLab
                 int prevCurrentIndex = CurrentIndex;
                 int prevPosition = _position;
                 JsonTokenType prevTokenType = TokenType;
+                BufferReader<byte> copy = reader;
                 if (ConsumeNextToken(ref reader, first))
                 {
                     return true;
@@ -237,6 +239,7 @@ namespace System.Text.JsonLab
                 CurrentIndex = prevCurrentIndex;
                 TokenType = prevTokenType;
                 _position = prevPosition;
+                reader = copy;
                 return false;
             }
 
@@ -282,7 +285,7 @@ namespace System.Text.JsonLab
                     if (first != JsonConstants.Quote)
                         ThrowJsonReaderException(ref this);
 
-                    //reader.Advance(1);    //TODO: How do we revert this?
+                    reader.Advance(1);
                     CurrentIndex++;
                     _position++;
                     TokenStartIndex++;
@@ -314,11 +317,21 @@ namespace System.Text.JsonLab
         {
             if (marker == JsonConstants.Quote)
             {
-                //reader.Advance(1);    //TODO: How do we revert this?
+                int prevCurrentIndex = CurrentIndex;
+                int prevPosition = _position;
+                BufferReader<byte> copy = reader;
+                reader.Advance(1);
                 CurrentIndex++;
                 _position++;
                 TokenStartIndex++;
-                return ConsumeString(ref reader);
+                if (!ConsumeString(ref reader))
+                {
+                    CurrentIndex = prevCurrentIndex;
+                    _position = prevPosition;
+                    reader = copy;
+                    return false;
+                }
+                return true;
             }
             else if (marker == JsonConstants.OpenBrace)
             {
@@ -384,11 +397,7 @@ namespace System.Text.JsonLab
             Value = span;
             ValueType = JsonValueType.Number;
             TokenType = JsonTokenType.Value;
-            if (TryGetNumber(Value, out _))
-            {
-                return false;
-            }
-            return true;
+            return TryGetNumber(Value, out _);
         }
 
         private bool ConsumeNull(ref BufferReader<byte> reader)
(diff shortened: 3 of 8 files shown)


KrzysztofCwalina:
BufferReader will be used to write parsers, and parsers often need to backtrack. It seems like it would be super useful to support backtracking.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
```diff
@@ -145,17 +180,33 @@ private bool ReadMultiSegment(ref BufferReader<byte> reader)
...
+                    if (first != JsonConstants.Quote)
+                        ThrowJsonReaderException(ref this);
                     TokenStartIndex++;
-                    ConsumePropertyName(ref reader);
+                    int prevCurrentIndex = CurrentIndex;
+                    int prevPosition = _position;
+                    BufferReader<byte> copy = reader;
+                    reader.Advance(1);
+                    CurrentIndex++;
+                    _position++;
```
KrzysztofCwalina: Can we somehow simplify all these indexes? It seems like we basically have 3 different ints tracking pretty much the same position (with different origins). Can we save the origins, and then only increment one integer?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
```diff
@@ -164,20 +215,33 @@ private bool ReadMultiSegment(ref BufferReader<byte> reader)
                 {
                     reader.Advance(1);
                     CurrentIndex++;
+                    _position++;
```
KrzysztofCwalina: Same situation here.

KrzysztofCwalina:
The perf regression seems scary

JeremyKuhne:
> It seems like it would be super useful to support backtracking.

We do in that you can copy the BufferReader. I wouldn't think you can do much better than that. @ahsonkhan, can you try to add a method that lets you set position and measure? Setting a long seems to be the only rational way to do it- we could optimize to see if we're in the current segment, otherwise we'd can fully reset and Advance.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -141,7 +141,7 @@ private bool TryReadToSlow(out ReadOnlySequence<T> sequence, T delimiter, T deli
                 else
                 {
                     // No delimiter, need to check the end of the span for odd number of escapes then advance
-                    if (remaining[remaining.Length - 1].Equals(delimiterEscape))
```
JeremyKuhne: Can this be !End?

JeremyKuhne: Also, can you add a test for this?

ahsonkhan: Sure, will do, but I think this API still needs more work as part of  https://github.com/dotnet/corefxlab/issues/2513

For example, there is potentially unnecessary work happening if `remaining.Length == 1`.

ahsonkhan:
> can you try to add a method that lets you set position and measure? 

I don't follow. Can you clarify what type of method you are thinking of? One that sets the current SequencePosition within BufferReader?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader_MultiSegment.cs
ahsonkhan: I have been thinking about this issue but haven't been able to come up with a solution that works for all cases. **I will see if I can figure something out.** The issue is the origin is only useful for subsequent calls in the outerloop. Within the loop, I need the concept of where a token starts (specifically strings). Generally, it isn't too expensive to track these. _position and CurrentIndex are too different to reconcile since new lines can essentially appear anywhere (especially when dealing with partially processed tokens).

JeremyKuhne:
> I don't follow. Can you clarify what type of method you are thinking of? One that sets the current SequencePosition within BufferReader?

``` C#
long consumed = reader.Consumed;
// ...
reader.SetPosition(consumed);
```

``` C#
public void SetPosition(long position)
{
    // Check for > 0
    // Is position in current segment?
    // Yes? Calculate CurrentSpanPosition and set & set Consumed
    // No? Reset reader and Advance(position)
}
```

In theory the common case would be backtracking in the current span. Another option:

``` C#
public void Backtrack(long count)
{
    // Check for > 0
    // Is count in current segment?
    // Yes? Calculate CurrentSpanPosition and set & set Consumed
    // No? Reset reader and Advance(position)
}
```

@KrzysztofCwalina, thoughts?

ahsonkhan:
> Is it possible to backtrack across segment boundaries? i.e. don't we lose past segments?

Right. Even though we have the original ReadOnlySequence currently, we would have to move forward from the start of the sequence to the back-track point. That's because there is no way to back-track the underlying ReadOnlySequence (maybe something we should consider adding).

And if we don't keep the original ReadOnlySequence, then there is no way to backtrack across the segment boundaries.

JeremyKuhne:
> Is it possible to backtrack across segment boundaries? i.e. don't we lose past segments?

You don't, otherwise we'd be busted all over (you wouldn't be able to save a copy and continue). If you go to Pipe and grab another sequence you need to start all over in that case.

JeremyKuhne:
> there is no way to back-track the underlying ReadOnlySequence (maybe something we should consider adding).

You'd have to build a reverse linked list on the fly, which doesn't seem plausible or particularly useful. We could keep the prior SequencePosition if we thought that it would be frequently useful- but that would bloat us further.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -128,6 +133,63 @@ public bool TryRead(out T value)
             return true;
         }
 
+        /// <summary>
+        /// Move the reader back the specified number of positions.
+        /// </summary>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        public void Retreat(int count)
```
JeremyKuhne: Rewind is a better name I think- I'm putting that name in the proposal.

JeremyKuhne: Also, can you make this 'long'?

ahsonkhan:
**We went from 1.32x to 1.96x down to 1.58x. So were able to get back most of the regression :)**

Most of the 10% cost is to support the re-entrant reader which was expected.

``` ini

BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17754
.NET Core SDK=3.0.100-alpha1-009460
  DefaultJob : .NET Core 2.1.3 (CoreCLR 4.6.26725.06, CoreFX 4.6.26725.05), 64bit RyuJIT


```

**Master:**

|                Method |  TestCase | segmentSize |       Mean |    Error |   StdDev |  Gen 0 | Allocated |
|---------------------- |---------- |------------ |-----------:|---------:|---------:|-------:|----------:|
|  **MultiSegmentSequence** | **Json400KB** |        **4000** | **1,275.6 us** | **8.582 us** | **7.166 us** | **1.9531** |   **14072 B** |
| **SingleSegmentSequence** | **Json400KB** |           **?** |   **963.2 us** | **5.585 us** | **5.224 us** |      **-** |       **0 B** |

**Before:**

|                Method |  TestCase | segmentSize |       Mean |     Error |    StdDev | Allocated |
|---------------------- |---------- |------------ |-----------:|----------:|----------:|----------:|
|  **MultiSegmentSequence** | **Json400KB** |        **4000** | **1,900.5 us** | **17.172 us** | **16.063 us** |   **14072 B** |
| **SingleSegmentSequence** | **Json400KB** |           **?** |   **969.2 us** |  **9.211 us** |  **7.692 us** |       **0 B** |


**Now:**

|                Method |  TestCase | segmentSize |       Mean |     Error |   StdDev |  Gen 0 | Allocated |
|---------------------- |---------- |------------ |-----------:|----------:|---------:|-------:|----------:|
|  **MultiSegmentSequence** | **Json400KB** |        **4000** | **1,391.3 us** | **10.106 us** | **9.453 us** | **1.9531** |   **14072 B** |
| **SingleSegmentSequence** | **Json400KB** |           **?** |   **879.4 us** |  **3.936 us** | **3.287 us** |      **-** |       **0 B** |

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: Yep. Made the count in Advance long, should follow suit.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -28,12 +28,17 @@ public BufferReader(in ReadOnlySequence<T> buffer)
             if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
             {
                 _moreData = true;
-                CurrentSpan = memory.Span;
-                if (CurrentSpan.Length == 0)
+                
+                if (memory.Length == 0)
```
ahsonkhan: @JeremyKuhne, note this change. I wonder if we have other places that need such fixes.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -128,6 +133,63 @@ public bool TryRead(out T value)
...
+            Consumed -= count;
+
+            if (CurrentSpanIndex >= count)
+            {
+                CurrentSpanIndex -= count;
+            }
+            else
+            {
+                // Current segment doesn't have enough space, scan backward through segments
+                RetreatToPreviousSpan(count);
```
JeremyKuhne: Shouldn't this be Consumed?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -128,6 +133,63 @@ public bool TryRead(out T value)
...
+            }
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private void RetreatToPreviousSpan(int count)
```
JeremyKuhne: Can the reset of state be unified with the constructor logic?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -128,6 +133,63 @@ public bool TryRead(out T value)
...
+                }
+            }
+            else
+            {
+                // No space in any spans and at end of sequence
+                _moreData = false;
+                CurrentSpan = default;
+            }
+
+            Advance(Consumed);
```
JeremyKuhne: This would be `count` right?

ahsonkhan: No, why? I think this should be Consumed.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: We don't need the count argument at all actually. It isn't used anywhere.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: We don't need the `count` argument at all actually. It isn't used anywhere.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: Well, you set Consumed to 0 above.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: How? I can't extract out to a shared method since the ctor requires all fields to be set.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: Oh right, lol, oops.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: Yeah, I just don't want to ever fix a bug we find without making sure we add a test at this point.

ahsonkhan: > Yeah, I just don't want to ever fix a bug we find without making sure we add a test at this point.

Agreed :)

PR #2523, opened 2018-10-07
Enable build on ubuntu 18.04

danmosemsft:
Note I have to use `ubuntu` not `ubuntu18.04` as there is no installer at that URL for .NET Core 1.0. Maybe I should return `ubuntu16.04` instead?

(Why do you still install 1.0?)

### DIFF
--- a/scripts/install-dotnet.sh
+++ b/scripts/install-dotnet.sh
@@ -100,6 +100,10 @@ get_os_download_name_from_platform() {
             echo "ubuntu.16.10"
             return 0
             ;;
+        "ubuntu.18.04")
+            echo "ubuntu"
+            return 0
+            ;;
         "alpine.3.4.3")
             echo "alpine"
             return 0


joshfree:
>> (Why do you still install 1.0?)

Because no one has fixed it yet I'm guessing.  Why don't we drop 1.0 in place of 2.1 and then we can return the real ubuntu18.04 installer?

cc: @KrzysztofCwalina

scripts/install-dotnet.sh
```diff
@@ -100,6 +100,10 @@ get_os_download_name_from_platform() {
             echo "ubuntu.16.10"
             return 0
             ;;
+        "ubuntu.18.04")
+            echo "ubuntu"
```
joshfree: We should get rid of the .NET Core 1.0 installer dependency and change this to return ubuntu18.04.  As a super short term hack we could probably return 16.04 for now with a // bugbug comment, until the 1.0 -> 2.1 change is in

ahsonkhan:
This PR can probably be closed based on https://github.com/dotnet/corefxlab/pull/2526.

joshfree:
Closing now that .NET Core 1.0 has been removed as a dependency from dotnet/corefxlab.

Building on Ubuntu 18.04 should now work fine.

PR #2458, opened 2018-08-28
DictionarySlim

AnthonyLloyd:

PR #2525, opened 2018-10-08
Added an OrderedDictionary

### LINKED ISSUE #2456: Add a generic OrderedDictionary class
Moved from [corefx#26634](https://github.com/dotnet/corefx/issues/26634).

Often times I've come across places when needing a `Dictionary` where the insertion order of the elements is important to me. Unfortunately, .NET does not currently have a generic `OrderedDictionary` class. We've had a non-generic `OrderedDictionary` class since .NET Framework 2.0 which oddly enough was when generics were added but no generic equivalent. This has forced many to roll their own solution, typically by using a combination of a `List` and `Dictionary` field resulting in the worst of both worlds in terms of performance and resulting in larger memory usage, and even worse sometimes users instead rely on implementation details of `Dictionary` for ordering which is quite dangerous.

## Proposed API
```c#
namespace Microsoft.Collections.Extensions {
    public class OrderedDictionary<TKey, TValue> : IDictionary<TKey, TValue>, IReadOnlyDictionary<TKey, TValue>, IList<KeyValuePair<TKey, TValue>>, IReadOnlyList<KeyValuePair<TKey, TValue>> {
        public struct Enumerator : IEnumerator<KeyValuePair<TKey, TValue>> {
            public KeyValuePair<TKey, TValue> Current { get; }
            public void Dispose();
            public bool MoveNext();
        }
        public sealed class KeyCollection : IList<TKey>, IReadOnlyList<TKey> {
            public struct Enumerator : IEnumerator<TKey> {
                public TKey Current { get; }
                public void Dispose();
                public bool MoveNext();
            }
            public int Count { get; }
            public TKey this[int index] { get; }
            public Enumerator GetEnumerator();
        }
        public sealed class ValueCollection : IList<TValue>, IReadOnlyList<TValue> {
            public struct Enumerator : IEnumerator<TValue> {
                public TValue Current { get; }
                public void Dispose();
                public bool MoveNext();
            }
            public int Count { get; }
            public TValue this[int index] { get; }
            public Enumerator GetEnumerator();
        }
        public OrderedDictionary();
        public OrderedDictionary(int capacity);
        public OrderedDictionary(IEqualityComparer<TKey> comparer);
        public OrderedDictionary(int capacity, IEqualityComparer<TKey> comparer);
        public OrderedDictionary(IEnumerable<KeyValuePair<TKey, TValue>> collection);
        public OrderedDictionary(IEnumerable<KeyValuePair<TKey, TValue>> collection, IEqualityComparer<TKey> comparer);
        public IEqualityComparer<TKey> Comparer { get; }
        public int Count { get; }
        public KeyCollection Keys { get; }
        public ValueCollection Values { get; }
        public TValue this[int index] { get; set; }
        public TValue this[TKey key] { get; set; }
        public void Add(TKey key, TValue value);
        public void Clear();
        public bool ContainsKey(TKey key);
        public int EnsureCapacity(int capacity);
        public TValue GetAt(int index);
        public TValue GetAt(int index, out TKey key);
        public Enumerator GetEnumerator();
        public TValue GetOrAdd(TKey key, TValue value);
        public TValue GetOrAdd(TKey key, Func<TValue> valueFactory);
        public TValue GetValue(TKey key);
        public int IndexOf(TKey key);
        public void Insert(int index, TKey key, TValue value);
        public void Move(int fromIndex, int toIndex);
        public void MoveRange(int fromIndex, int toIndex, int count);
        public bool Remove(TKey key);
        public bool Remove(TKey key, out TValue value);
        public void RemoveAt(int index);
        public void SetAt(int index, TValue value);
        public void SetAt(int index, TKey key, TValue value);
        public void SetValue(TKey key, TValue value);
        public void TrimExcess();
        public void TrimExcess(int capacity);
        public bool TryAdd(TKey key, TValue value);
        public bool TryGetValue(TKey key, out TValue value);
    }
}
```

Perhaps one of the reasons there was no generic `OrderedDictionary` added initially was due to issues with having both a key and index indexer when the key is an `int`. A call to the indexer would be ambiguous. Roslyn prefers the non-generic parameter so in this case the index indexer will be called. If the user needs to access elements by key they can use the explicit `GetValue` and `SetValue` methods.

## API Details
* `Insert` allows `index` to be equal to `Count` to insert the element at the end.
* `SetValue` will insert if needed, similar to the `Dictionary` indexer.
* Performance will be the same as `Dictionary` for all operations except `Remove` which will necessarily be `O(n)`. `Insert` and `RemoveAt` which aren't members of `Dictionary` will also be `O(n)`.

## Performance Characteristics Comparison
Operation | Dictionary<K,V> | SortedDictionary<K,V> | SortedList<K,V> | OrderedDictionary<K,V>
-- | -- | -- | -- | --
this[key] | O(1) | O(log n) | O(log n) or O(n) | O(1)
this[index] | n/a | n/a | n/a | O(1)
Add(key,value) | O(1) | O(log n) | O(n) | O(1)
Remove(key) | O(1) | O(log n) | O(n) | O(n)
RemoveAt(index) | n/a | n/a | n/a | O(n)
ContainsKey(key) | O(1) | O(log n) | O(log n) | O(1)
IndexOf(key) | n/a | n/a | n/a | O(1)
Insert(index,key,value) | n/a | n/a | n/a | O(n)
Space efficiency | good | worst | medium | good

## Open Questions
* Should the non-generic interfaces `ICollection`, `IList`, `IDictionary`, and `IOrderedDictionary` be implemented?

## Updates
* Added constructor overloads for `IEnumerable<KeyValuePair<TKey, TValue>>`.
* Added `ContainsValue` method due to being needed for the `ValueCollection.Contains` method.
* Proposal no longer advocates for throwing an exception when using an indexer while the key is an int.
* Changed namespace for corefxlab.
* Removed `IDictionary<TKey, TValue>` constructor overloads as they can be handled in the `IEnumerable<KeyValuePair<TKey, TValue>>` overload.
* Removed `ContainsValue` method as it's being implemented in the `ValueCollection.Contains` method.
* Made `KeyCollection` and `ValueCollection`'s `CopyTo` method use explicit interface implementation.
* Added `EnsureCapacity(int capacity)` and `TrimExcess(int capacity)` methods.
* Added `Move` and `MoveRange` methods.
* Added the `SetAt(int, TKey, TValue)` method overload.
* Added ref returning methods.
* Removed ref returning methods as the ref return values become invalidated when the internal array resizes.
* Made `KeyCollection` and `ValueCollection` constructors `internal` as it doesn't seem valuable.
* Added getter-only indexers to `KeyCollection` and `ValueCollection`.

TylerBrinkley:
No documentation yet.

Closes #2456


tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
```diff
@@ -0,0 +1,276 @@
...
+            Assert.Equal(3, test.IndexOf("3"));
+            Assert.Equal(3, test[3]);
+            Assert.True(test.ContainsKey("4"));
+            Assert.Equal(4, test["4"]);
+            Assert.Equal(4, test.IndexOf("4"));
+            Assert.Equal(4, test[4]);
+
+            // Tests OrderedDictionary Enumeration
+            var temp = 0;
+            foreach (var pair in test)
```
ahsonkhan: nit: avoid using var here (and for `var temp`) since it isn't clear from reading the right hand side what the type is.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/HashHelpers.cs
```diff
@@ -0,0 +1,93 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+
```
ahsonkhan: nit: remove extra new line

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
```diff
@@ -0,0 +1,276 @@
...
+
+using System;
+using Xunit;
+
+namespace Microsoft.Collections.Extensions.Tests
+{
+    public class OrderedDictionaryTests
+    {
+        [Fact]
+        public void TestBasicOrderedDictionaryOperations()
```
ahsonkhan: nit: I would break these into separate API specific tests.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,768 @@
...
+
+        bool ICollection<KeyValuePair<TKey, TValue>>.Contains(KeyValuePair<TKey, TValue> item) => TryGetValue(item.Key, out TValue value) && EqualityComparer<TValue>.Default.Equals(value, item.Value);
+
+        void ICollection<KeyValuePair<TKey, TValue>>.CopyTo(KeyValuePair<TKey, TValue>[] array, int arrayIndex)
+        {
+            if (array == null)
+            {
+                throw new ArgumentNullException(nameof(array));
+            }
+            if (arrayIndex < 0 || arrayIndex > array.Length)
```
ahsonkhan: nit: re-write as a single if check
```C#
if ((uint)arrayIndex > (uint)array.Length)
```

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,768 @@
...
+                throw new ArgumentOutOfRangeException(nameof(arrayIndex));
+            }
+            int count = Count;
+            if (array.Length - arrayIndex < count)
+            {
+                throw new ArgumentException("Destination array is not long enough to copy all the items in the collection. Check array index and length.");
+            }
+
+            Entry[] entries = _entries;
+            for (int i = 0; i < count; ++i)
```
ahsonkhan: Is it guaranteed that `count` is within `entries.Length`?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,768 @@
...
+                {
+                    throw new InvalidOperationException("Collection was modified; enumeration operation may not execute.");
+                }
+
+                _index = 0;
+                Current = default;
+            }
+        }
+
+        public sealed class KeyCollection : IList<TKey>, IReadOnlyList<TKey>
```
ahsonkhan: nit: split this out into a separate file

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,768 @@
...
+                        throw new InvalidOperationException("Collection was modified; enumeration operation may not execute.");
+                    }
+
+                    _index = 0;
+                    Current = default;
+                }
+            }
+        }
+
+        public sealed class ValueCollection : IList<TValue>, IReadOnlyList<TValue>
```
ahsonkhan: same here: move to a separate file

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
```diff
@@ -0,0 +1,276 @@
...
+            foreach (var value in test.Values)
+            {
+                Assert.True(false);
+            }
+        }
+
+        [Fact]
+        public void TestLargeOrderedDictionaryOperations()
+        {
+            const int count = 1000000;
```
ahsonkhan: nit: use digit separators for readability (`1_000_000`)

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/HashHelpers.cs
TylerBrinkley: That's a direct copy/paste from coreclr.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
TylerBrinkley: Yes, the `Count` property is guaranteed to always be less than or equal to `_entries.Length`.

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
TylerBrinkley: I agree. Is there anywhere within the .NET Core repos with guidance on how to write good unit tests, especially in regards to generic collections.

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
TylerBrinkley: :) It's hard to break old habits.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
ahsonkhan: Great. Debug.Assert?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/HashHelpers.cs
ahsonkhan: Should these remain in sync with https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/shared/System/Collections/HashHelpers.cs?

If so, can you please add a comment stating where you got this from.

ahsonkhan: Also, regarding the nit, we should remove the newline from both places :)

Other files don't have the extra new lines, for example see: https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/Numerics/Hashing/HashHelpers.cs

TylerBrinkley:
I've split out the tests into individual unit tests and have added support for many others.

Frassle:
Could we get interfaces to match, IOrderedDictionary and IReadOnlyOrderedDictionary? I've had code that could of benefited from IReadOnlyOrderedDictionary where Keys and Values are IList not IEnumerable.

danmosemsft:
This can close #2456, but probably we should have some issue tracking getting it through API and possibly moving it into CoreFX later.

danmosemsft:
@TylerBrinkley apologies for the delay.

@stephentoub @ahsonkhan @safern any further comments? Note, this is not waiting on API review - that's the purpose of creating the pack in this repo. However any review feedback is welcome including to the API. Note original discussion was in https://github.com/dotnet/corefx/issues/26634.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            return newSize;
+        }
+
+        public TValue GetAt(int index) => GetAt(index, out _);
+
+        public TValue GetAt(int index, out TKey key)
+        {
+            if (index < 0 || index >= Count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index), "Index was out of range. Must be non-negative and less than the size of the collection.");
```
danmosemsft: Please use only strings from resx (although we're not localizing, all the other code does)

eg `SR.ArgumentOutOfRange_Index`

danmosemsft: Same elsewhere

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/IDictionaryDebugView.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    internal sealed class DictionaryKeyCollectionDebugView<TKey, TValue>
+    {
+        private readonly ICollection<TKey> _collection;
+
+        public DictionaryKeyCollectionDebugView(ICollection<TKey> collection)
+        {
+            if (collection == null)
```
safern: this can be simplified to `_collection = collection ?? throw new ...`

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/IDictionaryDebugView.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+    }
+
+    internal sealed class DictionaryValueCollectionDebugView<TKey, TValue>
+    {
+        private readonly ICollection<TValue> _collection;
+
+        public DictionaryValueCollectionDebugView(ICollection<TValue> collection)
+        {
+            if (collection == null)
```
safern: Ditto

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.KeyCollection.cs
```diff
@@ -0,0 +1,137 @@
...
+                set => throw new NotSupportedException();
+            }
+
+            TKey IReadOnlyList<TKey>.this[int index] => ((IList<TKey>)this)[index];
+
+            bool ICollection<TKey>.IsReadOnly => true;
+
+            public KeyCollection(OrderedDictionary<TKey, TValue> orderedDictionary)
+            {
+                if (orderedDictionary == null)
```
safern: Simplify with ??

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.KeyCollection.cs
```diff
@@ -0,0 +1,137 @@
...
+
+            bool ICollection<TKey>.Contains(TKey item) => _orderedDictionary.ContainsKey(item);
+
+            void ICollection<TKey>.CopyTo(TKey[] array, int arrayIndex)
+            {
+                if (array == null)
+                {
+                    throw new ArgumentNullException(nameof(array));
+                }
+                if (arrayIndex < 0 || arrayIndex > array.Length)
```
safern: Use (uint) casting as @ahsonkhan previously suggested

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.ValueCollection.cs
```diff
@@ -0,0 +1,145 @@
...
+                set => throw new NotSupportedException();
+            }
+
+            TValue IReadOnlyList<TValue>.this[int index] => _orderedDictionary[index];
+
+            bool ICollection<TValue>.IsReadOnly => true;
+
+            public ValueCollection(OrderedDictionary<TKey, TValue> orderedDictionary)
+            {
+                if (orderedDictionary == null)
```
safern: Simplify with ??

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.ValueCollection.cs
```diff
@@ -0,0 +1,145 @@
...
+
+            bool ICollection<TValue>.Contains(TValue item) => ((IList<TValue>)this).IndexOf(item) >= 0;
+
+            void ICollection<TValue>.CopyTo(TValue[] array, int arrayIndex)
+            {
+                if (array == null)
+                {
+                    throw new ArgumentNullException(nameof(array));
+                }
+                if (arrayIndex < 0 || arrayIndex > array.Length)
```
safern: use uint casting

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            Resize(newSize);
+            ++_version;
+            return newSize;
+        }
+
+        public TValue GetAt(int index) => GetAt(index, out _);
+
+        public TValue GetAt(int index, out TKey key)
+        {
+            if (index < 0 || index >= Count)
```
safern: (uint) ?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
safern: Also should we use a ThrowHelper for collections in this experimental package as well? To minimize IL size?

https://github.com/dotnet/coreclr/blob/master/src/System.Private.CoreLib/src/System/ThrowHelper.cs

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/IDictionaryDebugView.cs
TylerBrinkley: That's a copy paste from Corefx.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
danmosemsft: I don't think it's very important at this point.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+                }
+            }
+            return index;
+        }
+
+        private bool TryInsert(int? index, TKey key, TValue value, InsertionBehavior behavior)
+        {
+            int i = IndexOf(key, out int hashCode);
+            Entry[] entries = _entries;
+            ++_version;
```
danmosemsft: You changed all postfix to prefix - personal style? I don't think it matters, although it increases the diff with Dictionary.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+
+        public void Clear()
+        {
+            if (Count > 0)
+            {
+                Array.Clear(_buckets, 0, _buckets.Length);
+                Array.Clear(_entries, 0, Count);
+                Count = 0;
+            }
+            ++_version;
```
danmosemsft: I think `++_version` could be removed here if it's OK that enumeration just terminates. This was already done for Dictionary.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+
+        public void SetAt(int index, TValue value)
+        {
+            if (index < 0 || index >= Count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index), "Index was out of range. Must be non-negative and less than the size of the collection.");
+            }
+
+            _entries[index].Value = value;
+            ++_version;
```
danmosemsft: We could choose to avoid `++_version` for `SetAt()` and for `SetValue()` which ended up updating an existing value, because that will not disrupt enumeration. I am not sure why Dictionary increments it in that case.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
danmosemsft: A Debug.Assert would not add anything because it would throw on the next line.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+
+            // Increment indices >= index;
+            int actualIndex = index ?? count;
+            for (i = actualIndex; i < count; ++i)
+            {
+                UpdateBucket(i, false, 1);
+            }
+            Array.Copy(entries, actualIndex, entries, actualIndex + 1, count - actualIndex);
+
+            Debug.Assert(_buckets.Length > 0);
```
danmosemsft: This assert is not useful because it would throw on the next line.

danmosemsft: Same below

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+        public void RemoveAt(int index)
+        {
+            int count = Count;
+            if (index < 0 || index >= count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index), "Index was out of range. Must be non-negative and less than the size of the collection.");
+            }
+
+            // Remove the entry from the bucket
+            UpdateBucket(index, true);
```
danmosemsft: Please use named parameter style as it saves looking at the callee. Eg `UpdateBucket(index, setToNext:true);`
Same elsewhere.
@stephentoub should that be in the [style guide](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            public int Next; // the index of the next item in the same bucket
+        }
+
+        private int[] _buckets = Array.Empty<int>();
+        private Entry[] _entries = Array.Empty<Entry>();
+        private int _version;
+        private readonly IEqualityComparer<TKey> _comparer;
+        private KeyCollection _keys;
+        private ValueCollection _values;
+
```
danmosemsft: I assume you removed `_freeList` because of the complexity? That's fine.

danmosemsft: Oh, because you're coalescing entries 😄

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            Debug.Assert(_buckets.Length > 0);
+            ref int bucket = ref _buckets[hashCode % _buckets.Length];
+            Entry newEntry = new Entry { HashCode = hashCode, Key = key, Value = value, Next = bucket - 1 };
+            entries[actualIndex] = newEntry;
+            bucket = actualIndex + 1;
+            ++Count;
+            return true;
+        }
+
+        private void UpdateBucket(int i, bool setToNext, int incrementAmount = 0)
```
danmosemsft: This method is hard to read because it is overloaded for two purposes: removing an entry from the bucket, and re-numbering the index in the bucket. I think it would be clearer to have two methods. It might even be faster (fewer branches)

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
```diff
@@ -0,0 +1,966 @@
...
+
+        [Fact]
+        public void Constructor_ThrowsArgumentNullException_WhenCollectionIsNull()
+        {
+            Assert.Throws<ArgumentNullException>(() => new OrderedDictionary<string, int>((IEnumerable<KeyValuePair<string, int>>)null));
+            Assert.Throws<ArgumentNullException>(() => new OrderedDictionary<string, int>(null, StringComparer.OrdinalIgnoreCase));
+        }
+
+        [Fact]
+        public void TestWithStringComparerIgnoreCase()
```
danmosemsft: Do you have tests for hash collision? OrdinalIgnoreCase could be used for a collision of 2 entries (although this test doesn't do that) but it would probably be worth having a custom comparer to test behavior when > 2 entries collide. Eg., repeat some of the general tests using a comparer that returns the same hashcode for everything.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+    {
+        private struct Entry
+        {
+            public int HashCode; // lower 31 bits of the hash code
+            public TKey Key;
+            public TValue Value;
+            public int Next; // the index of the next item in the same bucket
+        }
+
+        private int[] _buckets = Array.Empty<int>();
```
danmosemsft: Any particular reason you chose to initialize by default to capacity 0, compared with `Dictionary<TKey, TValue>` which looks like it will never have arrays smaller than 3? 

Although I wonder whether Dictionary should do the same. It makes empty ones cheaper.

danmosemsft: Correction, I see Dictionary starts at 0 with null arrays, you just exchange a test for null array against a test for empty array. Maybe it makes no difference.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+            }
+
+            IEqualityComparer<TKey> comparer = _comparer;
+            hashCode = (comparer?.GetHashCode(key) ?? key.GetHashCode()) & 0x7FFFFFFF;
+            int[] buckets = _buckets;
+            int index = -1;
+            if (buckets.Length > 0)
+            {
+                int bucket = hashCode % buckets.Length;
+                index = buckets[bucket] - 1;
```
danmosemsft: Consider adding GetBucketIndex and GetEntryIndex (eg like [this](https://github.com/dotnet/corefxlab/pull/2458/files#diff-6f44e9ce636799efd5c8b26382f2248bR39) and [this](https://github.com/dotnet/corefxlab/pull/2458/files#diff-6f44e9ce636799efd5c8b26382f2248bR42)). That hides a lot of the +1 and -1 making the code easier to read, and they will certainly be inlined.

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
danmosemsft: I did this for DictionarySlim, they can share: https://github.com/dotnet/corefxlab/commit/da5f15991f7a9778f65c948bd54a646adebc10a1

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,514 @@
...
+
+        #region Explicit Interface Implementation
+        KeyValuePair<TKey, TValue> IList<KeyValuePair<TKey, TValue>>.this[int index]
+        {
+            get
+            {
+                TValue value = GetAt(index, out TKey key);
+                return new KeyValuePair<TKey, TValue>(key, value);
+            }
+            set => throw new NotSupportedException(); // Should this be supported?
```
TylerBrinkley: Thoughts on this?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
TylerBrinkley: It doesn't seem worth it as it would only be used once here as the other place I do this I use a ref variable.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
TylerBrinkley: Just personal style from learning c++ in my college days. I can change it if need be.

safern: I think it makes sense to be consistent with Dictionary, specially if we're not fixing the other nits because they came from corefx itself. Like the boundary checks by using uint.

safern: Ahh github showed me an old diff. I saw you indeed fixed them :smile: thanks!

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -399,12 +403,13 @@ private bool TryInsert(int? index, TKey key, TValue value, InsertionBehavior beh
...
                         return true;
                     case InsertionBehavior.ThrowOnExisting:
-                        throw new ArgumentException("An item with the same key has already been added.");
+                        throw new ArgumentException(string.Format(Strings.Argument_AddingDuplicateWithKey, key.ToString()));
                     default:
                         return false;
                 }
             }
 
+            ++_version;
```
safern: nit: ++ afterwards.

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
```diff
@@ -962,5 +963,102 @@ public void TestLargeOrderedDictionaryOperations()
...
+            throw new InvalidOperationException();
+        }
+
+        [DebuggerStepThrough]
+        internal Collider C(int val) => new Collider(val);
     }
 }
+
+[DebuggerDisplay("{Key}")]
+internal struct Collider : IEquatable<Collider>, IComparable<Collider>
```
safern: Any reason why this is not part of the tests namespace?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,544 @@
...
+            else
+            {
+                int j = b - 1;
+                do
+                {
+                    ref Entry e = ref entries[j];
+                    if (e.Next == i)
+                    {
+                        e.Next = entry.Next;
+                        j = -1;
```
safern: Should we just call break here? I think it would be clearer than making it -1 and then saying, ahh if it is -1 it will exit the loop.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
danmosemsft: Our general pattern (in corefx anyway) seems to be to avoid both pre and post fix within an expression (eg `foo = bar++` or `foo = ++bar`) therefore the difference does not matter and we generally pick postfix as a convention.

Of course if the code is just clearer then exceptions are fine.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -109,13 +116,13 @@ public OrderedDictionary(IEnumerable<KeyValuePair<TKey, TValue>> collection, IEq
...
-            if (Count > 0)
+            int count = Count;
+            if (count > 0)
             {
                 Array.Clear(_buckets, 0, _buckets.Length);
-                Array.Clear(_entries, 0, Count);
+                Array.Clear(_entries, 0, count);
                 Count = 0;
             }
-            ++_version;
```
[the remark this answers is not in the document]
danmosemsft: Depends what invalid means, if it means "will fail or produce incorrect results" such as in the case where an insert moves an element along in the array, causing the enumerator to potentially return an element more than once or not at all. In the case of Clear, the MoveNext() checks Count every time `if (_index < _orderedDictionary.Count)` so MoveNext() will return false as if it reached the end, which is arguably correct. If Current is already set, it will not be cleared, but we don't check version in that property getter anyway.

It's probably a bit pedantic, but it seems across our collections we never had a solid definition. I think generally we are trying to relax it a bit (eg we now allow Remove during enumeration on Dictionary)

My care level is low though so either is fine.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
danmosemsft: I think it's fine to throw for now, but it should be InvalidOperationException.

danmosemsft: Scratch that, NotSupportedException is probably better.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,544 @@
...
+
+            // Increment indices >= index;
+            int actualIndex = index ?? count;
+            for (i = actualIndex; i < count; ++i)
+            {
+                UpdateBucketIndices(i, 1);
+            }
+            Array.Copy(entries, actualIndex, entries, actualIndex + 1, count - actualIndex);
+
+            ref int bucket = ref _buckets[hashCode % _buckets.Length];
```
danmosemsft: Per the perf measurements in DictionarySlim, it might be faster to just store `hashCode % _buckets.Length` and index into `_buckets` again when you want to set it. Perhaps you already tried though.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,544 @@
...
+            Entry newEntry = new Entry { HashCode = hashCode, Key = key, Value = value, Next = bucket - 1 };
+            entries[actualIndex] = newEntry;
+            bucket = actualIndex + 1;
+            ++Count;
+            return true;
+        }
+
+        private void RemoveEntryFromBucket(int i)
```
danmosemsft: `i` -> `entryIndex` perhaps?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
danmosemsft: Same comment other places. Only measurements can say.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,544 @@
...
+        {
+            Entry[] entries = _entries;
+            Entry entry = entries[i];
+            ref int b = ref _buckets[entry.HashCode % _buckets.Length];
+            if (b == i + 1)
+            {
+                b = entry.Next + 1;
```
danmosemsft: `// Bucket was pointing to removed entry. Update it to point to the next in the chain` ?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
danmosemsft: maybe, just return.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,544 @@
...
+            }
+            else
+            {
+                int j = b - 1;
```
danmosemsft: `// Start at the entry the bucket points to, and walk the chain until we find the entry with the index we want to remove, then fix the chain` ?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,544 @@
...
+        }
+
+        private void UpdateBucketIndices(int i, int incrementAmount)
+        {
+            Entry[] entries = _entries;
+            Entry entry = entries[i];
+            ref int b = ref _buckets[entry.HashCode % _buckets.Length];
+            if (b == i + 1)
+            {
+                b += incrementAmount;
```
danmosemsft: Simliar comments to above?

Thanks for breaking the method into two, now it's clear to me what it does. Maybe there is a way to write it in one method in a clear enough way, but this is fine for now.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/HashHelpers.cs
```diff
@@ -0,0 +1,93 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
```
danmosemsft: Nit, extra newline

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
TylerBrinkley: I just mimicked what @danmosemsft did in the DictionarySlim PR. I can pull it into the namespace or perhaps as a nested type.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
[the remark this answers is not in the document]
safern: Fair enough.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
TylerBrinkley: I'll leave it as is until performance tests are added.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
TylerBrinkley: I ended up implementing this.

TylerBrinkley:
I've added a `Move` and `MoveRange` method as proposed in [corefx#14338](https://github.com/dotnet/corefx/issues/14338).

I've also added support for the IList<KeyValuePair<TKey, TValue>>.this[int index] setter. The logic is as follows.

If the key is not contained in the dictionary then the entry at the specified index is replaced.
If the key is contained within the dictionary at the specified index then the key and value are updated for the entry.
If the key is contained within the dictionary at a different index then an exception is thrown as the method shouldn't affect the indices of other entries.

I've also addressed the PR feedback.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,678 @@
...
+        public TValue GetAt(int index) => GetAt(index, out _);
+
+        public TValue GetAt(int index, out TKey key)
+        {
+            if ((uint)index >= (uint)Count)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index), Strings.ArgumentOutOfRange_Index);
+            }
+
+            Entry entry = _entries[index];
```
omariom: It doesn't check if the entry has been removed? Is it a part of the contract?

TylerBrinkley: When an entry is removed all the entries after it are pushed down so that the entries array is always contiguous.

danmosemsft:
We don't need perf tests in order to merge this into corefxlab. But if you're making changes to tune perf, you might want to have them first, since sometimes the effect of a microoptimization can be counterintuitive. They are easy to create and run and have been very helpful in the DictionarySlim PR.

TylerBrinkley:
Thanks. I'm not making any changes to tune performance for now, just simplifying the code.

TylerBrinkley:
@dotnet-bot retest

TylerBrinkley:
@dotnet-bot test

danmosemsft:
@dotnet-bot test this please

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.KeyCollection.cs
```diff
@@ -0,0 +1,135 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
danmosemsft: Please update header to match
```
// Licensed to the .NET Foundation under one or more agreements.
```

Some of them were wrong a little while ago, maybe you snapped then.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.ValueCollection.cs
```diff
@@ -0,0 +1,144 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
danmosemsft: SAme license

tests/Microsoft.Experimental.Collections.Tests/Microsoft/Collections/Extensions/OrderedDictionaryTests.cs
```diff
@@ -0,0 +1,1492 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
danmosemsft: License

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,703 @@
+﻿// Copyright (c) Microsoft. All rights reserved.
```
danmosemsft: License

danmosemsft:
Good enough to put out for feedback. @TylerBrinkley any more changes planned right now? Mind if I squash?

danmosemsft:
I rebased but it looks like I don't have write permissions to your branch. Please grab it yourself from here: https://github.com/danmosemsft/corefxlab/tree/ordered-dictionary and satisfy yourself I did it correctly 😃

TylerBrinkley:
Looks good to me. How do I apply the squash to my branch? A git merge?

YohDeadfall:
You can do it via `git rebase -i`, so you would be able to choose what commits to squash.

danmosemsft:
I am not a git expert but I would expect you would just add my fork as a remote and fetch it, then reset your branch to it and force push ie
// add my remote and fetch 
git checkout yourbranch
git reset --hard mybranch
git push --force

Does that work? Should be no rebase -I as I did that part.

danmosemsft:
BTW would be good to go through and add XML comments to the public members with a note on their complexity?

YohDeadfall:
@danmosemsft Sorry, haven't read that you squashed commits already. Then yes, this will do the trick.

TylerBrinkley:
Alright, I've grabbed your changes and force pushed them into my branch, makes me think of Star Wars by the way. Now for the XML documentation, do you know where can I find the XML documentation for `Dictionary` so I don't have to recreate the wheel?

danmosemsft:
for core, the sources are at 
https://source.dot.net/#System.Private.CoreLib/shared/System/Collections/Generic/Dictionary.cs,d3599058f8d79be0
or in the corefx repo.

for desktop, either 
https://source.dot.net/#System.Private.CoreLib/shared/System/Collections/Generic/Dictionary.cs
or
C:\Program Files (x86)\Reference Assemblies\Microsoft\Framework\.NETCore\v4.5\System.Collections.xml

TylerBrinkley:
Why is there no XML documentation in the source?

danmosemsft:
In CoreFX, XML docs are used inconsistently. In this case it is because the code started in Desktop, which was inconsistent because it was not used directly for docs or intellisense at that time. There is a PR up in CoreFX to at least validate what there is. It would be valuable to have an XML docs (at least `<summary>` on anything non trivial) on all public/protected members.

Meanwhile I will find out more about how they generate docs.microsoft.com from the sources.

TylerBrinkley:
Thanks, I'll use the System.Collections.xml file for now.

TylerBrinkley:
I've added XML Documentation throughout.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.KeyCollection.cs
```diff
@@ -11,14 +11,27 @@ namespace Microsoft.Collections.Extensions
...
         [DebuggerTypeProxy(typeof(DictionaryKeyCollectionDebugView<,>))]
         [DebuggerDisplay("Count = {Count}")]
         public sealed class KeyCollection : IList<TKey>, IReadOnlyList<TKey>
         {
             private readonly OrderedDictionary<TKey, TValue> _orderedDictionary;
 
+            /// <summary>
+            /// Gets the number of elements contained in the <see cref="OrderedDictionary{TKey, TValue}.KeyCollection" />.
+            /// </summary>
+            /// <returns>The number of elements contained in the <see cref="OrderedDictionary{TKey, TValue}.KeyCollection" />. Retrieving the value of this property is an O(1) operation.</returns>
```
safern: Should `Retrieving the value of this property is an O(1) operation.` be part of the summary tag?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -214,6 +349,18 @@ public void Insert(int index, TKey key, TValue value)
...
+        /// </summary>
+        /// <param name="fromIndex">The zero-based index of the element to move.</param>
+        /// <param name="toIndex">The zero-based index to move the element to.</param>
+        /// <exception cref="ArgumentOutOfRangeException"><paramref name="fromIndex"/> is less than 0.
+        /// -or-
+        /// <paramref name="fromIndex"/> is equal to or greater than <see cref="OrderedDictionary{TKey, TValue}.Count" />
+        /// -or-
+        /// <paramref name="toIndex"/> is less than 0.
+        /// -or-
+        /// <paramref name="toIndex"/> is equal to or greater than <see cref="OrderedDictionary{TKey, TValue}.Count" /></exception>
```
safern: Can </exception> be in a closing tag for this guys with multiple or statements? It seems more readable.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.KeyCollection.cs
TylerBrinkley: That's how it was specified in the Dictionary XML docs for its KeyCollection. I agree it would make more sense there.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,981 @@
...
+        /// <exception cref="ArgumentNullException"><paramref name="key" /> is null.</exception>
+        /// <exception cref="ArgumentException">An element with the same key already exists in the <see cref="OrderedDictionary{TKey, TValue}" />.</exception>
+        public void Add(TKey key, TValue value) => TryInsert(null, key, value, InsertionBehavior.ThrowOnExisting);
+
+        /// <summary>
+        /// Removes all keys and values from the <see cref="OrderedDictionary{TKey, TValue}" />.
+        /// </summary>
+        public void Clear()
+        {
+            if (_count > 0)
```
danmosemsft: BTW, in DictionarySlim I made Clear() reset the capacity also. Dictionary retains the arrays as you have done. I guess it depends on the use case, whether you are likely to need that capacity soon enough again.

BTW did you mean to increment version?

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
```diff
@@ -0,0 +1,981 @@
...
+                throw new KeyNotFoundException(string.Format(Strings.Arg_KeyNotFoundWithKey, key.ToString()));
+            }
+            return _entries[index].Value;
+        }
+
+        /// <summary>
+        /// Returns the zero-based index of the element with the specified key within the <see cref="OrderedDictionary{TKey, TValue}" /> as an O(1) operation.
+        /// </summary>
+        /// <param name="key">The key of the element to locate.</param>
+        /// <returns>The zero-based index of the element with the specified key within the <see cref="OrderedDictionary{TKey, TValue}" />, if found; otherwise, –1.</returns>
```
danmosemsft: `-1` on this line is actually `U+2013	EN DASH` followed by `1`. Can you please use regular ASCII hyphen or else the C# `\u2013`? We use `\u` for anything over 0x7F.

danmosemsft:
@ahsonkhan random termination while running JSON tests - I believe I have seen this a few times before.
https://ci.dot.net/job/dotnet_corefxlab/job/master/job/ubuntu16.04_release_prtest/2824/consoleText

```
Skipped  System.Text.JsonLab.Tests.ReadFromPipeTests.ReadFromPipeUsingSpan
The active test run was aborted. Reason: 

Total tests: Unknown. Passed: 261. Failed: 0. Skipped: 3.
```
Bit fishy

danmosemsft:
@dotnet-bot test this please

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
danmosemsft: I was going to fix it - but your branch is still read only 😱 .  for future reference you can check this box if you want to accept changes to a PR:
image

danmosemsft:
nits above, then I will hit merge!

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
TylerBrinkley: That's weird because the checkbox is checked for me.

src/Microsoft.Experimental.Collections/Microsoft/Collections/Extensions/OrderedDictionary.cs
TylerBrinkley: Good catch. That was an issue in the source I copied from the XML Documentation on Dictionary.

danmosemsft:
Thanks @TylerBrinkley ! Once https://github.com/dotnet/corefxlab/pull/2594 is merged then up to NuGet.

PR #2526, opened 2018-10-08
Remove .NET Core 1.0 from build.sh

joshfree:
Remove .NET Core 1.0 from build scripts.
Add Ubuntu 18.04 to scripts/install-dotnet.sh


build.sh
```diff
@@ -45,13 +45,6 @@ if [ ! -d "dotnetcli" ]; then
     exit -1
   fi
 
-  ./scripts/install-dotnet.sh -Version 1.0.0 -InstallDir "dotnetcli"
```
ahsonkhan: We should remove this from build.cmd as well:
https://github.com/dotnet/corefxlab/blob/3baf21124e6b875cd3b987ea13f6e12202332136/scripts/build.ps1#L28-L32

PR #2534, opened 2018-10-12
Change skip* to advance*.

JeremyKuhne:
Fix moving past empty sequences.
Fix span advance to advancePastDelimiter handling.
Add IsLastSegment.
Add 4x faster Position property.
Add a variety of tests.

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -49,6 +49,11 @@ namespace System.Buffers.Reader
         }
 
         /// <summary>
+        /// Return true if we're in the last segment
+        /// </summary>
+        public bool IsLastSegment => _nextPosition.GetObject() == null;
+
+        /// <summary>
         /// True when there is no more data in the <see cref="Sequence"/>.
         /// </summary>
         public bool End => !_moreData;
@@ -61,7 +66,8 @@ namespace System.Buffers.Reader
         /// <summary>
         /// The current position in the <see cref="Sequence"/>.
         /// </summary>
-        public SequencePosition Position => Sequence.GetPosition(CurrentSpanIndex, _currentPosition);
+        public SequencePosition Position
+            => new SequencePosition(_currentPosition.GetObject(), CurrentSpanIndex + (_currentPosition.GetInteger() & ~(1 << 31)));
 
         /// <summary>
         /// The current segment in the <see cref="Sequence"/>.
@@ -216,6 +222,7 @@ namespace System.Buffers.Reader
                     {
                         CurrentSpan = default;
                         CurrentSpanIndex = 0;
+                        previousNextPosition = _nextPosition;
                     }
                 }
             }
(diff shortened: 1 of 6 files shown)


tests/System.Buffers.ReaderWriter.Tests/ReaderBasicTests.cs
```diff
@@ -395,6 +395,70 @@ public void ReaderIndexIsCorrect()
...
+            {
+                BufferReader<T> reader = new BufferReader<T>(buffer);
+                Assert.True(reader.TryAdvanceTo(inputData[i], advancePastDelimiter: false));
+                Assert.True(reader.TryPeek(out T value));
+                Assert.Equal(inputData[i], value);
+            }
+        }
+
+        [Fact]
+        public void AdvanceTo_AdvancePast()
```
ahsonkhan: Can we merge AdvanceTo and AdvanceTo_AdvancePast tests and pass in the advancePastDelimiter bool? The tests are very similar.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -61,7 +66,8 @@ public BufferReader(in ReadOnlySequence<T> buffer)
         /// <summary>
         /// The current position in the <see cref="Sequence"/>.
         /// </summary>
-        public SequencePosition Position => Sequence.GetPosition(CurrentSpanIndex, _currentPosition);
+        public SequencePosition Position
```
ahsonkhan: @pakrym, is it OK for us to rely on internal implementation details of SequencePosition to construct one (as part of this optimization)?

@JeremyKuhne, will this work regardless of how the ReadOnlySequence was constructed?

JeremyKuhne: I don't plan to leave it this way. I'm going to create an issue- chatting with @pakrym I'm going to suggest that we add a Dangerous* method to do this.

tests/System.Buffers.ReaderWriter.Tests/ReaderBasicTests.cs
JeremyKuhne: I considered it, but it would have been more complicated to put the conditions in than was worth it in my opinion.

ahsonkhan: You do it in other tests like TryReadTo_Sequence (below).

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
pakrym: They have to be part of API review.

JeremyKuhne:
A faster Position showed 5-7% benefits in other benchmarks (such as TryReadUntil). On it's own it is much faster:

**Before:**
```
          Method |     Mean |    Error |   StdDev |
---------------- |---------:|---------:|---------:|
 Reader_Position | 106.9 us | 1.259 us | 1.051 us |
```
**After:**
```
---------------- |---------:|----------:|----------:|
 Reader_Position | 24.84 us | 0.4780 us | 0.5870 us |
```

PR #2537, opened 2018-10-14
Change HttpParser header parsing to span

JeremyKuhne:
- Optimize Header parser BufferReader usage
- Optimize reader TryReadToAny for two items
- Optimize reader TryRead<T>

Goal here is to make sure the reader is working well for the parser.

**Parser before**
```
      Method |      Mean |     Error |    StdDev |
------------ |----------:|----------:|----------:|
 RequestLine |  69.73 ns | 0.3118 ns | 0.2917 ns |
     Headers | 147.80 ns | 0.6158 ns | 0.5459 ns |
 FullRequest | 230.65 ns | 0.6696 ns | 0.6264 ns |
```

**Parser after (no new reader aggressive inlining)**
```
      Method |      Mean |     Error |    StdDev |
------------ |----------:|----------:|----------:|
 RequestLine |  66.01 ns | 0.2759 ns | 0.2581 ns |
     Headers | 151.10 ns | 0.6065 ns | 0.5673 ns |
 FullRequest | 227.00 ns | 0.9968 ns | 0.9324 ns |
```

Slightly slower reading just the headers, but _much_ easier to follow. If the reader's `TryReadToAny` is marked as `AggressiveInlining` it would end up notably faster than it started:

**Parser after (reader TryRead AggressiveInlining)**
```
      Method |      Mean |     Error |    StdDev |
------------ |----------:|----------:|----------:|
 RequestLine |  66.05 ns | 0.4254 ns | 0.3979 ns |
     Headers | 137.62 ns | 0.9707 ns | 0.9080 ns |
 FullRequest | 219.71 ns | 0.6370 ns | 0.5958 ns |
```

At a 10% impact it seems like we should keep the `AggressiveInlining`.

For the binary reader it is another 17% faster than it was:

**Before**
```
                 Method |      Mean |    Error |   StdDev |
----------------------- |----------:|---------:|---------:|
              ReadInt32 |  74.46 us | 1.462 us | 2.746 us |
    ReadInt32_BigEndian | 121.07 us | 1.324 us | 1.238 us |
 ReadInt32_LittleEndian |  89.43 us | 1.756 us | 2.221 us |
```

**After**
```
                 Method |      Mean |     Error |    StdDev |
----------------------- |----------:|----------:|----------:|
              ReadInt32 |  63.58 us | 1.2667 us | 2.0812 us |
    ReadInt32_BigEndian | 112.13 us | 0.7395 us | 0.6917 us |
 ReadInt32_LittleEndian |  75.97 us | 1.4968 us | 1.4001 us |
```

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -40,6 +40,7 @@ namespace System.Buffers
 
         // TODO:
         // Move to helper in ReadOnlySequence
+        // https://github.com/dotnet/corefx/issues/33029
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
         private unsafe static void GetFirstSpan(in ReadOnlySequence<T> buffer, out ReadOnlySpan<T> first, out SequencePosition next)
         {
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -17,34 +17,97 @@ namespace System.Buffers
         /// Create a <see cref="BufferReader" over the given <see cref="ReadOnlySequence{T}"/>./>
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public BufferReader(in ReadOnlySequence<T> buffer)
+        public BufferReader(ReadOnlySequence<T> buffer)
         {
             CurrentSpanIndex = 0;
             Consumed = 0;
             Sequence = buffer;
-            _currentPosition = Sequence.Start;
-            _nextPosition = _currentPosition;
+            _currentPosition = buffer.Start;
 
-            if (buffer.TryGet(ref _nextPosition, out ReadOnlyMemory<T> memory, advance: true))
+            GetFirstSpan(buffer, out ReadOnlySpan<T> first, out _nextPosition);
+            CurrentSpan = first;
+            _moreData = first.Length > 0;
+
+            if (!buffer.IsSingleSegment && !_moreData)
             {
                 _moreData = true;
+                GetNextSpan();
+            }
+        }
 
-                if (memory.Length == 0)
+        private const int FlagBitMask = 1 << 31;
+        private const int IndexBitMask = ~FlagBitMask;
+
+        // TODO:
+        // Move to helper in ReadOnlySequence
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private unsafe static void GetFirstSpan(in ReadOnlySequence<T> buffer, out ReadOnlySpan<T> first, out SequencePosition next)
+        {
+            first = default;
+            next = default;
+            SequencePosition start = buffer.Start;
+            int startIndex = start.GetInteger();
+            object startObject = start.GetObject();
+
+            if (startObject != null)
+            {
+                SequencePosition end = buffer.End;
+                int endIndex = end.GetInteger();
+                bool isMultiSegment = startObject != end.GetObject();
+
+                // A == 0 && B == 0 means SequenceType.MultiSegment
+                if (startIndex >= 0)
                 {
-                    CurrentSpan = default;
-                    // No space in the first span, move to one with space
-                    GetNextSpan();
+                    if (endIndex >= 0)  // SequenceType.MultiSegment
+                    {
+                        ReadOnlySequenceSegment<T> segment = (ReadOnlySequenceSegment<T>)startObject;
+                        next = new SequencePosition(segment.Next, 0);
+                        first = segment.Memory.Span;
+                        if (isMultiSegment)
+                        {
+                            first = first.Slice(startIndex);
+                        }
+                        else
+                        {
+                            first = first.Slice(startIndex, endIndex - startIndex);
+                        }
+                    }
+                    else
+                    {
+                        if (isMultiSegment)
+                            ThrowHelper.ThrowInvalidOperationException_EndPositionNotReached();
+
+                        first = new ReadOnlySpan<T>((T[])startObject, startIndex, (endIndex & IndexBitMask) - startIndex);
+                    }
                 }
                 else
                 {
-                    CurrentSpan = memory.Span;
+                    first = GetFirstSpanSlow(startObject, startIndex, endIndex, isMultiSegment);
                 }
             }
-            else
+        }
+
+        [MethodImpl(MethodImplOptions.NoInlining)]
+        private static ReadOnlySpan<T> GetFirstSpanSlow(object startObject, int startIndex, int endIndex, bool isMultiSegment)
+        {
+            Debug.Assert(startIndex < 0 || endIndex < 0);
+            if (isMultiSegment)
+                ThrowHelper.ThrowInvalidOperationException_EndPositionNotReached();
+
+            // The type == char check here is redundant. However, we still have it to allow
+            // the JIT to see when that the code is unreachable and eliminate it.
+            // A == 1 && B == 1 means SequenceType.String
+            if (typeof(T) == typeof(char) && endIndex < 0)
             {
-                // No space in any spans and at end of sequence
-                _moreData = false;
-                CurrentSpan = default;
+                var memory = (ReadOnlyMemory<T>)(object)((string)startObject).AsMemory();
+
+                // No need to remove the FlagBitMask since (endIndex - startIndex) == (endIndex & ReadOnlySequence.IndexBitMask) - (startIndex & ReadOnlySequence.IndexBitMask)
+                return memory.Span.Slice(startIndex & IndexBitMask, endIndex - startIndex);
+            }
+            else // endIndex >= 0, A == 1 && B == 0 means SequenceType.MemoryManager
+            {
+                startIndex &= IndexBitMask;
+                return ((MemoryManager<T>)startObject).Memory.Slice(startIndex, endIndex - startIndex).Span;
             }
         }
 
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
@@ -42,7 +42,7 @@ namespace System.Buffers
         // Move to helper in ReadOnlySequence
         // https://github.com/dotnet/corefx/issues/33029
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        private unsafe static void GetFirstSpan(in ReadOnlySequence<T> buffer, out ReadOnlySpan<T> first, out SequencePosition next)
+        private static void GetFirstSpan(in ReadOnlySequence<T> buffer, out ReadOnlySpan<T> first, out SequencePosition next)
         {
             first = default;
             next = default;
@@ -108,7 +108,7 @@ namespace System.Buffers
             else // endIndex >= 0, A == 1 && B == 0 means SequenceType.MemoryManager
             {
                 startIndex &= IndexBitMask;
-                return ((MemoryManager<T>)startObject).Memory.Slice(startIndex, endIndex - startIndex).Span;
+                return ((MemoryManager<T>)startObject).Memory.Span.Slice(startIndex, endIndex - startIndex);
             }
         }
 
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
@@ -737,10 +737,11 @@ namespace System.Buffers
                 }
                 return true;
             }
+
+            // Only check the slow path if there wasn't enough to satisfy next
             return unread.Length < next.Length && IsNextSlow(next, advancePast);
         }
 
-        [MethodImpl(MethodImplOptions.NoInlining)]
         private unsafe bool IsNextSlow(ReadOnlySpan<T> next, bool advancePast)
         {
             ReadOnlySpan<T> currentSpan = UnreadSpan;
--- a/src/System.Text.Http/System.Text.Http.csproj
+++ b/src/System.Text.Http/System.Text.Http.csproj
@@ -4,7 +4,6 @@
     <Description>Non-allocating HTTP parser</Description>
     <TargetFramework>netstandard1.3</TargetFramework>
     <PackageTags>HTTP parser parsing .NET non-allocating corefxlab</PackageTags>
-    <LangVersion>7.2</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
   </PropertyGroup>
   <ItemGroup>
--- a/src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
+++ b/src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
@@ -5,6 +5,7 @@
 using System.Buffers;
 using System.Buffers;
 using System.Buffers.Text;
+using System.Diagnostics;
 using System.Numerics;
 using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
@@ -44,7 +45,7 @@ namespace System.Text.Http.Parser
             _showErrorDetails = showErrorDetails;
         }
 
-        public unsafe bool ParseRequestLine<T>(T handler, in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined) where T : IHttpRequestLineHandler
+        public bool ParseRequestLine<T>(T handler, in ReadOnlySequence<byte> buffer, out SequencePosition consumed, out SequencePosition examined) where T : IHttpRequestLineHandler
         {
             consumed = buffer.Start;
             examined = buffer.End;
@@ -75,17 +76,14 @@ namespace System.Text.Http.Parser
                 }
             }
 
-            // Fix and parse the span
-            fixed (byte* data = &MemoryMarshal.GetReference(span))
-            {
-                ParseRequestLine(ref handler, data, span.Length);
-            }
+            // Parse the span
+            ParseRequestLine(ref handler, span);
 
             examined = consumed;
             return true;
         }
 
-        public unsafe bool ParseRequestLine<T>(ref T handler, in ReadOnlySequence<byte> buffer, out int consumed) where T : IHttpRequestLineHandler
+        public bool ParseRequestLine<T>(ref T handler, in ReadOnlySequence<byte> buffer, out int consumed) where T : IHttpRequestLineHandler
         {
             // Prepare the first span
             var span = buffer.First.Span;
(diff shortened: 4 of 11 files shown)


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -321,23 +321,27 @@ public bool TryReadTo(out ReadOnlySequence<T> sequence, T delimiter, T delimiter
         /// <param name="delimiters">The delimiters to look for.</param>
         /// <param name="advancePastDelimiter">True to move past the first found instance of any of the given <paramref name="delimiters"/>.</param>
         /// <returns>True if any of the the <paramref name="delimiters"/> were found.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
JeremyKuhne: Heh, whoops. Left this in. I was intending to pull this, but I'll leave it for now to illustrate what I was talking about.

JeremyKuhne:
Scraped a bit more perf out- updated the comments above.

KrzysztofCwalina:
The parser was heavily optimized in the context of TechEmpower, and was very sensitive to very small changes. @davidfowl, @pakrym, do you think we should merge these changes?

pakrym:
Would be nice to try it with Kestrel and see how it affects plaintext performance.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
```diff
@@ -22,14 +23,12 @@ public static partial class ReaderExtensions
...
-            {
-                value = MemoryMarshal.Read<T>(span);
-                reader.Advance(sizeof(T));
-                return true;
-            }
+            if (span.Length < sizeof(T))
+                return TryReadSlow(ref reader, out value);
 
-            return TryReadSlow(ref reader, out value);
+            value = Unsafe.ReadUnaligned<T>(ref MemoryMarshal.GetReference(span));
```
ahsonkhan: Why is this necessary? I would imagine the redundant length check in MemoryMarshal.Read would be ignored by the JIT.

cc @AndyAyersMS 

https://github.com/dotnet/corefx/blob/71c1371d56fea965c6743bf770c46408dbbdae05/src/Common/src/CoreLib/System/Runtime/InteropServices/MemoryMarshal.cs#L141
```C#
        public static T Read<T>(ReadOnlySpan<byte> source)
            where T : struct
        {
            if (RuntimeHelpers.IsReferenceOrContainsReferences<T>())
            {
                ThrowHelper.ThrowInvalidTypeWithPointersNotSupported(typeof(T));
            }
            if (Unsafe.SizeOf<T>() > source.Length)
            {
                ThrowHelper.ThrowArgumentOutOfRangeException(ExceptionArgument.length);
            }
            return Unsafe.ReadUnaligned<T>(ref GetReference(source));
        }
```

src/System.Text.Http/System.Text.Http.csproj
```diff
@@ -4,7 +4,7 @@
     <Description>Non-allocating HTTP parser</Description>
     <TargetFramework>netstandard1.3</TargetFramework>
     <PackageTags>HTTP parser parsing .NET non-allocating corefxlab</PackageTags>
-    <LangVersion>7.2</LangVersion>
+    <LangVersion>7.3</LangVersion>
```
ahsonkhan: We already have latest LangVersion in common.props. We should just remove this property.

https://github.com/dotnet/corefxlab/blob/6fb9e0e79789d0f46849a73354ea4d962112ddaa/tools/common.props#L16

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -615,12 +493,66 @@ private unsafe int FindEndOfName(byte* headerLine, int length)
...
             {
-                RejectRequestHeader(headerLine, length);
+                RejectRequestHeader(new ReadOnlySpan<byte>(headerLine, length));
             }
 
             return index;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private unsafe void TakeSingleHeader<T>(ReadOnlySpan<byte> headerLine, T handler) where T : IHttpHeadersHandler
```
ahsonkhan: This method doesn't need to be unsafe.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -107,271 +105,151 @@ public HttpParser(bool showErrorDetails)
...
-            }
+            // Parse the span
+            ParseRequestLine(ref handler, span);
 
             consumed = span.Length;
             return true;
         }
 
-        private unsafe void ParseRequestLine<T>(ref T handler, byte* data, int length) where T : IHttpRequestLineHandler
+        private unsafe void ParseRequestLine<T>(ref T handler, ReadOnlySpan<byte> data) where T : IHttpRequestLineHandler
```
ahsonkhan: Similarly, here/elsewhere, if we are using span instead of ptr/length, we should remove the unsafe keyword.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
JeremyKuhne: > Why is this necessary?

Don't know precisely, but it makes a huge difference.

src/System.Text.Http/System.Text.Http.csproj
JeremyKuhne: Fair enough.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
JeremyKuhne: Yep, sorry, missed removing these.

JeremyKuhne:
> The parser was heavily optimized in the context of TechEmpower, and was very sensitive to very small changes.

If we need more perf tests I can add them. There is no way we can make a safe parser as fast as an unsafe one, of course. We can, however, try to make the fastest safe parser possible. Whether I do that with these existing classes or fork another set of classes is something I need guidance on- I don't know what the expectations are here.

Most of what I won here was through smarter logic (at least in the case of the perf tests we have). As stated, it would be even faster if I flipped it all to raw pointers and inlined code. There is just no way around that. 

> Would be nice to try it with Kestrel and see how it affects plaintext performance.

How do I do this?

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -75,17 +76,14 @@ public HttpParser(bool showErrorDetails)
                 }
             }
 
-            // Fix and parse the span
-            fixed (byte* data = &MemoryMarshal.GetReference(span))
-            {
-                ParseRequestLine(ref handler, data, span.Length);
-            }
+            // Parse the span
+            ParseRequestLine(ref handler, span);
```
ahsonkhan: Missed one. `ParseRequestLine<T>` is no longer `unsafe`.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
AndyAyersMS: Other than the redundant length check in the `IG03` block I don't see much difference in the jitted code between the two.
```asm
;; slow version
G_M12736_IG01:
       4883EC38             sub      rsp, 56
       90                   nop      

G_M12736_IG02:
       488D4108             lea      rax, bword ptr [rcx+8]
       4C8B00               mov      r8, bword ptr [rax]
       4C89442430           mov      bword ptr [rsp+30H], r8
       8B4008               mov      eax, dword ptr [rax+8]
       448B4118             mov      r8d, dword ptr [rcx+24]
       443BC0               cmp      r8d, eax
       7774                 ja       SHORT G_M12736_IG10

G_M12736_IG03:
       4C8B4C2430           mov      r9, bword ptr [rsp+30H]
       412BC0               sub      eax, r8d
       4D63C0               movsxd   r8, r8d
       4D03C1               add      r8, r9
       83F804               cmp      eax, 4
       7C56                 jl       SHORT G_M12736_IG08
       83F804               cmp      eax, 4
       7C62                 jl       SHORT G_M12736_IG11

G_M12736_IG04:
       4C89442428           mov      bword ptr [rsp+28H], r8
       488B442428           mov      rax, bword ptr [rsp+28H]
       8B00                 mov      eax, dword ptr [rax]
       8902                 mov      dword ptr [rdx], eax
       488B11               mov      rdx, qword ptr [rcx]
       4883C204             add      rdx, 4
       488911               mov      qword ptr [rcx], rdx
       8B5118               mov      edx, dword ptr [rcx+24]
       4863C2               movsxd   rax, edx
       4C8D4108             lea      r8, bword ptr [rcx+8]
       4D8B08               mov      r9, bword ptr [r8]
       458B4008             mov      r8d, dword ptr [r8+8]
       4D63C0               movsxd   r8, r8d
       4983E804             sub      r8, 4
       4C3BC0               cmp      r8, rax
       7E08                 jle      SHORT G_M12736_IG05
       83C204               add      edx, 4
       895118               mov      dword ptr [rcx+24], edx
       EB0A                 jmp      SHORT G_M12736_IG06

G_M12736_IG05:
       BA04000000           mov      edx, 4
       E83CCAFFFF           call     BufferReader`1:AdvanceToNextSpan(long):this

G_M12736_IG06:
       B801000000           mov      eax, 1

G_M12736_IG07:
       4883C438             add      rsp, 56
       C3                   ret      

G_M12736_IG08:
       E865CAFFFF           call     X:TryReadSlow(byref,byref):bool
       90                   nop      

G_M12736_IG09:
       4883C438             add      rsp, 56
       C3                   ret      

G_M12736_IG10:
       E81A3FE64F           call     ThrowHelper:ThrowArgumentOutOfRangeException()
       CC                   int3     

G_M12736_IG11:
       B919000000           mov      ecx, 25
       E8AF42E64F           call     ThrowHelper:ThrowArgumentOutOfRangeException(int)
       CC                   int3     

```
We really should teach the jit about ref structs . It is being super-paranoid about re-fetching fields. Closest open issue is dotnet/coreclr#19185.

ahsonkhan: > Other than the redundant length check

Presumably, it is this check: `if (Unsafe.SizeOf<T>() > source.Length)`

Can the JIT elide that redundant check if we have already validated that `source.Length >= Unsafe.SizeOf<T>()` before calling the method?

AndyAyersMS: Hard to say for sure -- the jit is currently not very good at removing redundant checks in "straight line" code.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -107,271 +105,151 @@ public HttpParser(bool showErrorDetails)
...
-                }
-                else if (pathStart == -1)
-                {
-                    pathStart = offset;
-                }
+            int escapeIndex = path.IndexOf(BytePercentage);
+            if (escapeIndex == 0)
+            {
+                // Can't start with %
+                RejectRequestLine(data);
```
ahsonkhan: consider `goto Error` for all the error cases.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -107,271 +105,151 @@ public HttpParser(bool showErrorDetails)
...
                 {
                     // else inform HTTP version is unsupported.
-                    RejectUnknownVersion(data + offset, length - offset - 2);
+                    RejectUnknownVersion(data.Slice(offset, data.Length - offset - 2));
                 }
             }
 
             // After version's 8 bytes and CR, expect LF
-            if (data[offset + 8 + 1] != ByteLF)
+            if (version[8 + 1] != ByteLF)
```
ahsonkhan: Hopefully this is valid access and won't throw out of range exception.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -107,271 +105,151 @@ public HttpParser(bool showErrorDetails)
...
+            out SequencePosition examined,
+            out int consumedBytes) where T : IHttpHeadersHandler
         {
             consumed = buffer.Start;
             examined = buffer.End;
             consumedBytes = 0;
 
-            var bufferEnd = buffer.End;
+            BufferReader<byte> reader = new BufferReader<byte>(buffer);
+            ReadOnlySpan<byte> eol = s_Eol;
```
ahsonkhan: Update the static, constant, byte arrays to be ReadOnlySpan properties.

So, s_Eol becomes:
```C#
ReadOnlySpan<byte> Eol => new byte [] {(byte)'\r', (byte)'\n'};
```

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -615,12 +493,66 @@ private unsafe int FindEndOfName(byte* headerLine, int length)
...
             if (index == length || sawWhitespace)
             {
-                RejectRequestHeader(headerLine, length);
+                RejectRequestHeader(new ReadOnlySpan<byte>(headerLine, length));
             }
 
             return index;
         }
 
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ahsonkhan: I don't think we should mark such a large method as aggressive inlining.

JeremyKuhne:
@pakrym, @davidfowl, @stephentoub, @ahsonkhan 

I've updated the pull to use the reader for the request line and also improve the construction of the reader. Getting the first span out of the `ReadOnlySequence` is extremely expensive (like 15-20ns). I've temporarily leveraged the intended API additions that @ahson and I discussed from investigating this.

This essentially gets rid of unsafe, better validates CR/LF pairs, and improves most of the parsing. Parsing the request line itself is slightly slower (5%, but is more correct in validating CR/LF). Reading the rest of the header is significantly faster ( 30%). The full request is about 15% faster- it would be faster still if we would use the same reader to parse the full request.

Here is where I started (updated for the latest bits in master:

```
      Method |      Mean |     Error |    StdDev |
------------ |----------:|----------:|----------:|
 RequestLine |  66.18 ns | 0.4604 ns | 0.4306 ns |
     Headers | 160.09 ns | 1.6956 ns | 1.5861 ns |
 FullRequest | 244.85 ns | 2.6407 ns | 2.4701 ns |

                 Method |     Mean |     Error |   StdDev |   Median |
----------------------- |---------:|----------:|---------:|---------:|
 ConstructSingleSegment | 30.71 ns | 0.6348 ns | 1.177 ns | 31.31 ns |
  ConstructMultiSegment | 25.54 ns | 0.5252 ns | 1.012 ns | 26.17 ns |
```

**After:**

```
      Method |      Mean |     Error |    StdDev |
------------ |----------:|----------:|----------:|
 RequestLine |  69.54 ns | 0.3191 ns | 0.2829 ns |
     Headers | 121.82 ns | 0.9522 ns | 0.8441 ns |
 FullRequest | 211.13 ns | 0.8298 ns | 0.7356 ns |

                 Method |     Mean |     Error |    StdDev |   Median |
----------------------- |---------:|----------:|----------:|---------:|
 ConstructSingleSegment | 14.14 ns | 0.3037 ns | 0.3376 ns | 14.22 ns |
  ConstructMultiSegment | 19.02 ns | 0.4060 ns | 0.8293 ns | 19.50 ns |
```

**IF** we were to create an unsafe reader constructor we could squeeze out another 3 ns each time we construct it. Bad user code that tears the structs in the `ReadOnlySequence` would cause us to violate type safety instead of throwing invalid cast exceptions.

cc: @joshfree, @danmosemsft

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
JeremyKuhne: Tried, it made it worse.

JeremyKuhne:
Here is what passing the reader along instead of `ReadOnlySequence` gets you:

**Before**

```
      Method |      Mean |     Error |    StdDev |
 FullRequest | 244.85 ns | 2.6407 ns | 2.4701 ns |
```

**After**

```
      Method |      Mean |     Error |    StdDev |
------------ |----------:|----------:|----------:|
 FullRequest | 173.62 ns | 1.9351 ns | 1.8101 ns |
```

244.85 -> 173.62. Not too shabby.

pakrym:
> passing the reader along 

What do you mean by this?

JeremyKuhne:
>> passing the reader along

> What do you mean by this?

The current signature takes a `ReadOnlySequence<byte>`, here is how I changed it for the benchmark:

``` C#
[Benchmark]
public void FullRequest()
{
    BufferReader<byte> reader = new BufferReader<byte>(s_plaintextTechEmpowerRequestRos);
    s_parser.ParseRequestLine(request, ref reader);
    s_parser.ParseHeaders(request, ref reader);
}
```

> Were you able to add this to kestrel? Ping me if you need help, would be awesome to see how it behaves in out perf harness.

I've started on it, and I will ping you.  :)

JeremyKuhne:
Made a last pass through to clean out the old code and tweak inlining. Here are the final results, which are a little bit better:

```
Class Handler

      Method |    Before |     After | Diff |
------------ |----------:|----------:|-----:|
 RequestLine |  77.27 ns |  67.31 ns | 0.87 |
     Headers | 175.35 ns | 136.59 ns | 0.78 |
 FullRequest | 266.24 ns | 181.06 ns | 0.68 |

Struct Handler

      Method |    Before |     After | Diff |
------------ |----------:|----------:|-----:|
 RequestLine |  66.54 ns |  54.49 ns | 0.82 |
     Headers | 151.08 ns | 110.16 ns | 0.73 |
 FullRequest | 237.48 ns | 157.77 ns | 0.66 |
```

Pretty happy with this, especially given that the code is much simpler and there is no `byte*`.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -17,34 +17,98 @@ namespace System.Buffers
...
 
-                if (memory.Length == 0)
+        private const int FlagBitMask = 1 << 31;
+        private const int IndexBitMask = ~FlagBitMask;
+
+        // TODO:
+        // Move to helper in ReadOnlySequence
+        // https://github.com/dotnet/corefx/issues/33029
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
+        private unsafe static void GetFirstSpan(in ReadOnlySequence<T> buffer, out ReadOnlySpan<T> first, out SequencePosition next)
```
ahsonkhan: Does this need to be unsafe?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -17,34 +17,98 @@ namespace System.Buffers
...
-                CurrentSpan = default;
+                var memory = (ReadOnlyMemory<T>)(object)((string)startObject).AsMemory();
+
+                // No need to remove the FlagBitMask since (endIndex - startIndex) == (endIndex & ReadOnlySequence.IndexBitMask) - (startIndex & ReadOnlySequence.IndexBitMask)
+                return memory.Span.Slice(startIndex & IndexBitMask, endIndex - startIndex);
+            }
+            else // endIndex >= 0, A == 1 && B == 0 means SequenceType.MemoryManager
+            {
+                startIndex &= IndexBitMask;
+                return ((MemoryManager<T>)startObject).Memory.Slice(startIndex, endIndex - startIndex).Span;
```
ahsonkhan:
```C#
((MemoryManager<T>)startObject).Memory.Span.Slice(startIndex, endIndex - startIndex);
```

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -17,34 +17,98 @@ namespace System.Buffers
...
+
+            // The type == char check here is redundant. However, we still have it to allow
+            // the JIT to see when that the code is unreachable and eliminate it.
+            // A == 1 && B == 1 means SequenceType.String
+            if (typeof(T) == typeof(char) && endIndex < 0)
             {
-                // No space in any spans and at end of sequence
-                _moreData = false;
-                CurrentSpan = default;
+                var memory = (ReadOnlyMemory<T>)(object)((string)startObject).AsMemory();
```
ahsonkhan: Note that once this moves to ReadOnlySequence, it can be re-written as follows:
```C#
ReadOnlySpan<char> spanOfChar =((string)startObject).AsSpan(startIndex & IndexBitMask, endIndex - startIndex);
ReadOnlySpan<T> spanOfT = MemoryMarshal.CreateSpan(ref Unsafe.As<char, T>(ref MemoryMarshal.GetReference(spanOfChar)), spanOfChar.Length);
```

`MemoryMarshal.CreateSpan` is a netcoreapp only API.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -27,7 +24,7 @@ public class HttpParser : IHttpParser
         private const byte BytePercentage = (byte)'%';
         private const long maxRequestLineLength = 1024;
 
-        static readonly byte[] s_Eol = Encoding.UTF8.GetBytes("\r\n");
+        private static ReadOnlySpan<byte> Eol => new byte[] { ByteCR, ByteLF };
```
ahsonkhan: We  could probably make the same change to the other static byte arrays below (if it makes sense): `s_http11`/etc.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -44,714 +41,310 @@ public HttpParser(bool showErrorDetails)
...
 
-            // Fix and parse the span
-            fixed (byte* data = &MemoryMarshal.GetReference(span))
-            {
-                ParseRequestLine(ref handler, data, span.Length);
-            }
+            ParseRequestLine(handler, requestLine);
 
-            consumed = span.Length;
+            consumed = (int)reader.Consumed;
```
ahsonkhan: Can you please add a comment here explaining why this cast is safe here?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: No, I'll make another pass for that.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
JeremyKuhne: I don't think we're using them, but I'll look.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
JeremyKuhne: It isn't really, but we weren't checking for it before.

@pakrym, any thoughts on this in general? I can turn this into a checked cast, but it would be good to understand what our expectations are with >2GB headers.

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -44,714 +41,310 @@ public HttpParser(bool showErrorDetails)
...
 
-            // Skip space
-            offset++;
-
-            byte ch = 0;
-            // Target = Path and Query
-            var pathEncoded = false;
-            var pathStart = -1;
-            for (; offset < length; offset++)
+            if (data.Length < 11)
```
ahsonkhan: `Z * HTTP/1.1` is 12 characters. Why is the check < 11?

Also does this hold true for the newer RFC?
https://tools.ietf.org/html/rfc7230#section-5.3.4

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
```diff
@@ -44,714 +41,310 @@ public HttpParser(bool showErrorDetails)
...
-            if (index == length || sawWhitespace)
-            {
-                RejectRequestHeader(headerLine, length);
-            }
+            var reader = new BufferReader<byte>(buffer);
+            consumedBytes = 0;
 
-            return index;
+            bool success = ParseHeaders(handler, ref reader);
+            if (success == true)
```
ahsonkhan: nit: `if (success)`

src/System.Text.Http/System/Text/Http/Parser/HttpParser.cs
JeremyKuhne: > Z * HTTP/1.1 is 12 characters. Why is the check < 11?

Because I can't count. The real thing we need here is <9 to make sure we can read our known headers out (long and a space).

> Also does this hold true for the newer RFC?

That example is longer. This is just a slight optimization to avoid checking twice for header length when reading out known headers. I'll comment.

src/System.Text.Http/System/Text/Http/Parser/HttpUtilities.cs
```diff
@@ -152,110 +152,65 @@ public static string GetAsciiStringEscaped(this Span<byte> span, int maxChars)
...
-            else if (*(uint*)data == _httpGetMethodInt)
-            {
-                methodLength = 3;
-                return Http.Method.Get;
-            }
-            else if (length < sizeof(ulong))
+            length = 0;
+            method = Http.Method.Custom;
+
+            if (Unsafe.ReadUnaligned<uint>(ref GetReference(span)) == _httpGetMethodInt)
```
ahsonkhan: `MemoryMarshal.Read<uint>` ? Using the Unsafe class doesn't make the method fully safe (in release).

JeremyKuhne: This is just eliding the bounds check. Yes, it is technically unsafe, but I think it is worth going this far.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: Noted

mjsabby:
This is really interesting. Is the plan here to graduate this as a general utility parser? Or is this for benchmarking only?

PR #2539, opened 2018-10-16
Remove SpanExtensions and replace with GetValueAsX instance methods

ahsonkhan:


src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -1529,5 +1530,218 @@ private bool TryGetNumberLookForEnd(ReadOnlySpan<byte> data, out ReadOnlySpan<by
             number = data.Slice(0, i);
             return true;
         }
+
+        public string GetValueAsString()
```
joshfree: Why don't we also have TryGetValueAsNNN alternatives that don't throw InvalidCastException?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
```diff
@@ -1529,5 +1530,218 @@ private bool TryGetNumberLookForEnd(ReadOnlySpan<byte> data, out ReadOnlySpan<by
...
+        {
+            if (Utf8Parser.TryParse(Value, out int intVal, out int bytesConsumed))
+            {
+                if (Value.Length == bytesConsumed)
+                {
+                    return intVal;
+                }
+            }
+
+            if (Utf8Parser.TryParse(Value, out long longVal, out bytesConsumed))
```
joshfree: should this method first try to parse as a long and then iff the value fits in an int, cast and return?

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
ahsonkhan: We already have the `Utf8Parser.TryParse` methods that the caller can use instead. These APIs are meant to be convenience APIs which throw if there is invalid data rather than returning a value based on partial consumption. For example, if the user calls `GetValueAsInt32` for a span containing a decimal, this throws. If they want to truncate instead, `Utf8Parser.TryParse` already does that.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonReader.cs
ahsonkhan: The `TryParse(out long)` API is ~10% slower than the one that returns an int. It is also more likely for the input to contain an int.

PR #2542, opened 2018-10-17
Break out the reader changes from #2537

JeremyKuhne:
See #2537

### DIFF
--- a/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
+++ b/src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
@@ -4,6 +4,7 @@
 
 using System.Buffers.Binary;
 using System.Diagnostics;
+using System.Runtime.CompilerServices;
 using System.Runtime.InteropServices;
 
 namespace System.Buffers.Reader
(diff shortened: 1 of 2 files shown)


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
```diff
@@ -46,7 +45,7 @@ public static partial class ReaderExtensions
                 return false;
             }
 
-            value = MemoryMarshal.Read<T>(tempSpan);
+            value = Unsafe.ReadUnaligned<T>(ref MemoryMarshal.GetReference(tempSpan));
```
ahsonkhan: I am not sure if we should be relying on the Unsafe class for performance here, especially for the slow path.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -321,23 +321,27 @@ public bool TryReadTo(out ReadOnlySequence<T> sequence, T delimiter, T delimiter
         /// <param name="delimiters">The delimiters to look for.</param>
         /// <param name="advancePastDelimiter">True to move past the first found instance of any of the given <paramref name="delimiters"/>.</param>
         /// <returns>True if any of the the <paramref name="delimiters"/> were found.</returns>
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
         public bool TryReadToAny(out ReadOnlySpan<T> span, ReadOnlySpan<T> delimiters, bool advancePastDelimiter = true)
         {
             ReadOnlySpan<T> remaining = UnreadSpan;
-            int index = remaining.IndexOfAny(delimiters);
+            int index = delimiters.Length == 2
```
ahsonkhan: Why optimize specifically for 2 only?

JeremyKuhne: Because it didn't measure any improvements past that.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_binary.cs
JeremyKuhne: We're simply skipping the redundant bounds checks here. While I would prefer not to, I don't know that we have a choice to meet our performance goals for now. Hopefully this sort of thing will be taken care of automagically by the JIT in the future.

JeremyKuhne: In this case the "slow" path isn't _that_ slow, so it definitely measures. The amortization we're talking about is our best guess- while we don't expect that we'll hit the boundary often, in this case the change is relatively benign and will definitely help where we've guessed wrong.

PR #2543, opened 2018-10-17
Allow passing in an optional allocator for Reader

JeremyKuhne:
Also change the namespace to System.Buffers.

This doesn't impact the existing perf tests any. I'm working on some tests to look at impact of managing the array creation and will follow up.


src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -77,10 +77,12 @@ private bool TryReadToSlow(out ReadOnlySpan<T> span, T delimiter, T delimiterEsc
             }
 
             Debug.Assert(sequence.Length > 0);
-            span = sequence.IsSingleSegment ? sequence.First.Span : sequence.ToArray();
+            span = sequence.ToSpan(_arrayPool);
             return true;
         }
 
+
```
ahsonkhan: nit: cleanup extra new lines

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -5,25 +5,28 @@
...
         private SequencePosition _nextPosition;
+        private ArrayPool<T> _arrayPool;
         private bool _moreData;
 
         /// <summary>
         /// Create a <see cref="BufferReader" over the given <see cref="ReadOnlySequence{T}"/>./>
         /// </summary>
         [MethodImpl(MethodImplOptions.AggressiveInlining)]
-        public BufferReader(in ReadOnlySequence<T> buffer)
+        public BufferReader(in ReadOnlySequence<T> buffer, ArrayPool<T> arrayPool = null)
```
stephentoub: What's our current thinking on `ArrayPool<T>`?  Do we expect developers using this API will actually be creating their own pools?  I'm wondering if we should instead just use Shared, and only add such a ctor if there's real demand.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -297,7 +300,7 @@ public ReadOnlySpan<T> Peek(int maxCount)
             }
 
             // Not enough contiguous Ts, allocate and copy what we can get
-            return PeekSlow(new T[maxCount]);
+            return PeekSlow(_arrayPool.Rent(maxCount));
```
stephentoub: I don't see any calls to `.Return` in this PR.  Where is the array being returned to the pool?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/NewArray.cs
```diff
@@ -0,0 +1,22 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Buffers
+{
+    /// <summary>
+    /// Simple ArrayPool that just new(s) up arrays
+    /// </summary>
+    internal sealed class NewArray<T> : ArrayPool<T>
```
ahsonkhan: Why do we need this?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: How do we plan to return the array back to the pool?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
```diff
@@ -427,7 +429,7 @@ public unsafe bool TryReadTo(out ReadOnlySequence<T> sequence, ReadOnlySpan<T> d
             }
             else
             {
-                peekBuffer = new Span<T>(new T[delimiter.Length]);
+                peekBuffer = new Span<T>(_arrayPool.Rent(delimiter.Length));
```
stephentoub: How are you going to return this array if the only reference to it is the one stored inside the span?

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
ahsonkhan: +1

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
```diff
@@ -5,25 +5,28 @@
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 
-namespace System.Buffers.Reader
+namespace System.Buffers
 {
     public ref partial struct BufferReader<T> where T : unmanaged, IEquatable<T>
     {
+        private static NewArray<T> s_newArray = new NewArray<T>();
```
stephentoub: Nit: readonly

stephentoub (dismissed):
I'm not clear on how this is actually supposed to work, given that arrays aren't being returned to the pool.  This will just end up draining the pool for anything other than the `NewArray` one.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: Yes, we expect them to create their own wrappers **if** this matters to them. We can't use the regular array pool as there is no reasonable way to manage lifetime when you don't know if the Span you got back is rented or not.

I'll be creating some concrete examples of how to do this and will be updating this pull with them (and measurements).

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: It isn't being returned as it isn't actually renting in the default implementation and isn't always renting here anyway. See my other comments.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader_search.cs
JeremyKuhne: Not going to return it as there is nothing to return to.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/NewArray.cs
JeremyKuhne: Because we're not actualy renting anything. We can't control lifetime without making the performance terrible for the normal case.

The idea is that there is an out here for those that want the **best** performance but still want to handle allocations in the segment straddling case.

src/System.Buffers.ReaderWriter/System/Buffers/Reader/BufferReader.cs
JeremyKuhne: We don't- see my other comments.

JeremyKuhne:
> I'm not clear on how this is actually supposed to work, given that arrays aren't being returned to the pool. This will just end up draining the pool for anything other than the NewArray one.

It isn't supposed to work without creating a wrapper. E.g. something that tracks rented arrays and wraps our existing pool or creating a pool that simply keeps giving the same buffer. Depends on usage.

One way around all of this is to turn the reader into something that never passes back spans, but passes back lengths and forces you to make another call to a SliceOrCopy method.

Another is to extend all of the APIs to take an output Span and have them also return some tristate that explains whether a copy was made and if there was enough room.

The last option is to just call the ReadOnlySequence APIs if you care about allocation.

None of these options can be done without significant cost to the caller. I naturally went to the ArrayPool<T> as it is an existing "allocation" interface we have. I get that it could be misleading so I'm thinking I'll just update this to take an optional delegate. Perhaps super direct like `public delegate Span<T> Allocator<T>(int length);`.

stephentoub:
> E.g. something that tracks rented arrays

I don't know how you intend to do this.  Please share very clear examples before merging.  I do not think we should be creating APIs that accept an ArrayPool and rent from it without returning to it.

JeremyKuhne:
> I do not think we should be creating APIs that accept an ArrayPool and rent from it without returning to it.

I agree, I've already updated to use a delegate.

> I'm still interested in understanding exactly how this is meant to be used in a sane manner

It's reasonably simple, but still intended for advanced usage. You know when something is created as you create it through your delegate. One very simple example:

``` C#
private const int BufferSize = 4096;
private byte[] _buffer = new byte[BufferSize];
private Span<byte> Allocate(int size)
{
    if (size > BufferSize)
        throw new InvalidOperationException("Unexpected line length");
    return _buffer;
}

public void ProcessLines(ReadOnlySequence<byte> sequence)
{
    BufferReader<byte> reader = new BufferReader<byte>(sequence, Allocate);
    ReadOnlySpan<byte> crlf = s_crlf;

    while (!reader.End)
    {
        if (!reader.TryReadToAny(out ReadOnlySpan<byte> line, crlf, advancePastDelimiter: false))
        {
            // Couldn't find another delimiter
        }

        if (!reader.IsNext(crlf, advancePast: true))
        {
            // Not a good CR/LF pair
        }

        // Good line, handle it
        ProcessLine(line);
    }
}
```

Some may want to follow that pattern above and allocate everytime and throw to handle bad input data (say, 1GB lines).

Writing a disposable wrapper for `ArrayPool` (`ArrayPoolTracker`) that tracks everything you rent and then returns when disposed is another way to leverage this.

This could also be used to instrument when you're crossing boundaries and needing to copy.

At some point any solutions that you would write with this could become very heavy- and likely counter productive. Those cases the overhead of letting the GC handle things (the default) or the overhead of working with `ReadOnlySequence` APIs gets wiped out- so I'm not really concerned about super complicated solutions being written with this.

stephentoub:
> new BufferReader<byte>(sequence, Allocate);

FWIW, this will allocate a delegate for `this.Allocate`.  Avoiding that makes it less simple.

> It's reasonably simple

So if I'm understanding correctly, you're relying on the facts that:
1) `BufferReader` is a ref struct, and thus can't escape
2) The delegate returns a span, and thus can't escape

which when combined allows the consumer of `BufferReader` to know that whatever array was returned from the call to `Allocate` will not still be in use once their code regains control.  Is that right?

This also assumes then that `BufferReader` will not call `Allocate` more than once in any given execution, or if it does, that it will no longer use the previous result, as `Allocate` could end up returning the same object repeatedly.

Why is this design better than, say, just passing in the scratch space you want `BufferReader` to use, with `BufferReader` using it if it's large enough and otherwise allocating or throwing or failing with an error code that indicates larger scratch space is needed, or some such thing?  I agree the flexibility of a delegate/interface/abstraction/etc. is nice, but it also seems to make things more complex, at least to me.  The developer then needs to understand if/when/for what purpose the delegate will be invoked, whereas if they just supply a buffer, there's no question about it.

JeremyKuhne:
> Why is this design better than...

I'm running through a bunch of performance measurements to give a full answer. I'm also digging into what it looks like to pass the Allocator in where it might be used.

KrzysztofCwalina:
> Writing a disposable wrapper for ArrayPool (ArrayPoolTracker) that tracks everything you rent and then returns when disposed is another way to leverage this.

MemoryPool is such wrapper. 
https://github.com/dotnet/corefx/blob/master/src/System.Memory/ref/System.Memory.cs#L126

stephentoub:
> MemoryPool is such wrapper.

That allocates.

PR #2549, opened 2018-10-20
Optimize TryParse out bool extension method by splitting slow path

ahsonkhan:
This type of optimization can be applied in a lot of other APIs.

**~5-10% improvement for the single segment case.**

``` ini

BenchmarkDotNet=v0.11.1, OS=Windows 10.0.17763
.NET Core SDK=3.0.100-alpha1-009667
  DefaultJob : .NET Core 2.1.3 (CoreCLR 4.6.26725.06, CoreFX 4.6.26725.05), 64bit RyuJIT


```
|                                       Method |     Mean |     Error |    StdDev | Scaled | Allocated |
|--------------------------------------------- |---------:|----------:|----------:|-------:|----------:|
|           SingleSegmentBufferReader_original | 5.963 us | 0.0477 us | 0.0446 us |   1.00 |       0 B |
| SingleSegmentBufferReader_original_with_opts | 5.376 us | 0.0449 us | 0.0420 us |   0.90 |       0 B |


ahsonkhan:
Master got better (because of https://github.com/dotnet/corefxlab/pull/2551):

|                             Method |     Mean |     Error |    StdDev | Scaled | Allocated |
|----------------------------------- |---------:|----------:|----------:|-------:|----------:|
| SingleSegmentBufferReader_original | 5.637 us | 0.0911 us | 0.0807 us |   1.00 |       0 B |
|      SingleSegmentBufferReader_new | 5.428 us | 0.0369 us | 0.0308 us |   0.96 |       0 B |

ilmax:
@ahsonkhan out of curiosity, does a go-to is faster than two returns?

ahsonkhan:
> out of curiosity, does a go-to is faster than two returns?

No, not really (at least not visible in a micro-benchmark). However, the disassembly is smaller: https://www.diffchecker.com/93DU926Y

ilmax:
Thanks!

JeremyKuhne:
@AndyAyersMS do you have any comments on this difference? How valuable is the reduction in instruction count here? Is this something we could theoretically get to with JIT improvements and no goto?

> No, not really (at least not visible in a micro-benchmark). However, the disassembly is smaller: https://www.diffchecker.com/93DU926Y

AndyAyersMS:
Unless this is a generic method/class with a large number of potential instantiations (does not seem to be the case here), I would not worry about saving few bytes of code and would favor whatever is more readable or natural looking.

The jit behavior for returns is dependent on target architecture and you may well see different impacts for x64 or arm64. Likely at some point we'll adjust the jit strategy around multiple returns (see for instance dotnet/coreclr#13814) as we know there are some quirks in the current jit approach. Once we tackle that, it generally won't matter how you express this in your code.

JeremyKuhne:
Thanks @AndyAyersMS. @ahsonkhan, given these comments I'd not put `goto` in unless we measure an important difference in an upstream scenario or see it in something really core like `Advance`.

ahsonkhan:
> given these comments I'd not put `goto` in unless we measure an important difference in an upstream scenario or see it in something really core like `Advance`.

OK. Will remove.

PR #2571, opened 2018-10-26
dropped to released version of piplelines

KrzysztofCwalina:
The preview version of pipelines is not in corfxlab myget not on nuget.org. This means adding System.Buffers.ReaderWriter to a project fails, unless the project has pipelines myget feed (which is not common or intuitive). Since ReaderWriter (nor any components in corfxlab) needs preview of pipelines, I am dropping the pipelines dependency to the officially released on nuget.org.


KrzysztofCwalina:

ahsonkhan:
The latest pipelines package from master also has its own dependency issue due to System.Buffers which is blocking https://github.com/dotnet/corefxlab/pull/2550, so dropping to release is fine. However, I don't understand why it is necessary for you.

**Also, remove it from the update-Dependencies project or it will change again when a new build comes out (with auto-PR):**
https://github.com/dotnet/corefxlab/blob/fe117a6a49212d0f93e3ae4387d1fa05b5024c20/scripts/update-dependencies/Program.cs#L104

PR #2577, opened 2018-10-31
 Address JsonReader API review feedback

ahsonkhan:
**Changes:**
- Utf8JsonReader -> JsonUtf8Reader
- Remove the single argument constructors to avoid confusion and improve discoverability of the ctor that requires passing state
- Change JsonReaderOptions to a struct and add an enum called CommentHandling with regular enum values (rather than flags).
- Rename BytePosition to LineBytePosition to avoid confusion.
- Rename Consumed to BytesConsumed.
- Change TokenType to be backed by an int rather than a byte.
- Keep the Position and BytesConsumed properties on both Utf8JsonReader and on JsonReaderState.
- Add GetValueAsBool().
- Remove GetValueAsNumber().
- Change the GetValueAsX APIs to TryGetValueAsX.
- GetValueAsString() should throw when TokenType is not string or property name.


src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
```diff
@@ -28,7 +28,7 @@ public static JsonDynamicObject Parse(ReadOnlySpan<byte> utf8, int expectedNumbe
             var properties = new Dictionary<JsonProperty, JsonValue>(expectedNumberOfProperties);
             stack.Push(new JsonDynamicObject(properties));
 
-            var reader = new Utf8JsonReader(utf8);
```
KrzysztofCwalina: I was late to the review. Where can I find some writeup on the motivation for the name change?

src/System.Text.JsonLab/System/Text/Json/JsonReaderException.cs
```diff
@@ -7,15 +7,15 @@ namespace System.Text.JsonLab
     public class JsonReaderException : Exception
     {
         //TODO: Consider adding line number and position to the message itself
-        public JsonReaderException(string message, long lineNumber, long bytePosition) : base(message)
+        public JsonReaderException(string message, long lineNumber, long lineBytePosition) : base(message)
```
KrzysztofCwalina: Is there a code sample proving that we can write helpers on top of the position APIs that would allow higher level code to compute the information that the callers actually want: the text around the place where the error occurred.

src/System.Text.JsonLab/System/Text/Json/JsonReaderOptions.cs
```diff
@@ -4,11 +4,20 @@
...
-        AllowComments = 0b0001, // Allow comments but don't skip them
-        SkipComments = 0b0011,  // Allow and Skip comments
+        public JsonReaderOptions(CommentHandling commentHandling)
+        {
+            _commentHandling = commentHandling;
+        }
+
+        internal CommentHandling _commentHandling;
+
+        public enum CommentHandling
```
KrzysztofCwalina: Why do we want this to be backed by int and not byte?

src/System.Text.JsonLab/System/Text/Json/Utf8Json.Reader.cs
```diff
@@ -422,7 +422,7 @@ private void EndObject()
                 }
                 else
                 {
-                    _utf8Json._state._inObject = _utf8Json._state._stack.Pop() != JsonTokenType.StartArray;
+                    _utf8Json._state._inObject = (JsonTokenType)_utf8Json._state._stack.Pop() != JsonTokenType.StartArray;
```
KrzysztofCwalina: Should we cache state in a local? i.e. would it improve perf?

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: https://github.com/dotnet/apireviews/pull/73

Alternatives consider (either of which I am fine with): `Utf8JsonReader`, `JsonReaderUtf8`, `JsonReader`

What name would you prefer?

src/System.Text.JsonLab/System/Text/Json/JsonReaderException.cs
ahsonkhan: Not yet, I will write something up. In any case, it requires the caller to have access to all the data so they can seek to the appropriate location and encode the byte position to character position.

src/System.Text.JsonLab/System/Text/Json/JsonReaderOptions.cs
ahsonkhan: It can be backed by a byte. Is there a downside?

src/System.Text.JsonLab/System/Text/Json/Utf8Json.Reader.cs
ahsonkhan: It might, and I can try it.

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
KrzysztofCwalina: It seems like we have Utf8Parser, Utf8Formatter, and so it seems Utf8JsonReader would follow the pattern.

src/System.Text.JsonLab/System/Text/Json/JsonReaderOptions.cs
KrzysztofCwalina: If you use byte, the downside is that you won't be able to have more than 256 options; the upside that JsonReaderOptions might be smaller.

ahsonkhan: We recently changed `JsonTokenType` from a byte-backed enum to be a regular int. In your opinion, should it remain byte-backed as well?

ahsonkhan: I will make this change in an upcoming PR. Thanks!

src/System.Text.JsonLab.Dynamic/System/Text/Json/JsonDynamicObject.cs
ahsonkhan: > But I was mainly asking about the reasons.

- We want all JsonX to sort together. JsonReaderState, JsonUtf8Reader, JsonReaderOptions, JsonUtf8Writer, etc.
- The feature is JSON, not UTF-8, where UTF-8 is a characteristic of the reader. Putting Utf8 up-front emphasizes the incorrect aspect of the type. Arguable, Utf8Parser/Formatter are primarily about UTF-8.

stephentoub: > we have Utf8Parser, Utf8Formatter

Arguably this is a Utf8Reader.  What kind?  A JsonUtf8Reader.

But I don't actually have a strong opinion.

PR #2578, opened 2018-11-01
 Changing the Tensor operators to be internal

tannergooding:
CC. @danmosemsft, @ericstj, @eerhardt


samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.NativeMemory.cs
```diff
@@ -9,23 +9,22 @@ namespace TensorSamples
 {
     partial class WrapNativeMemory
     {
-        public class NativeMemory<T> : OwnedMemory<T>
+        public class NativeMemory<T> : MemoryManager<T>
```
tannergooding: Sample was out of date, needed cleanup. Not sure what all the appropriate changes here are, so I just did the minimal to get it to compile.

src/System.Numerics.Tensors/System/Numerics/Tensors/Tensor.Operations.tt
```diff
@@ -10,14 +10,14 @@ namespace System.Numerics.Tensors
     public static partial class Tensor
     { 
 <# foreach (MethodConfiguration method in methodConfiguration) { #>
-        public static <#= method.GetGenericResultMethodSignature("Tensor", "T")#>
+        internal static <#= method.GetGenericResultMethodSignature("Tensor", "T")#>
```
tannergooding: There were a couple arithmetic things (like `Zero`) that are used all over the place (even in constructors).

I found it easier to just make the operations internal, than to move them out of the assembly entirely.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.NativeMemory.cs
tannergooding: CC. @GrabYourPitchforks, who indicated there may be a gist with the guidance

tannergooding: Gist is here: https://gist.github.com/GrabYourPitchforks/8efb15abbd90bc5b128f64981766e834

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.NativeMemory.cs
```diff
@@ -78,6 +74,11 @@ public override void Retain()
                 Interlocked.Increment(ref refCount);
             }
 
+            public override void Unpin()
+            {
+                throw new NotImplementedException();
```
ericstj: Shouldn't this just be a NOOP?  Native memory is effectively always pinned.

src/System.Numerics.Tensors/System.Numerics.Tensors.csproj
```diff
@@ -1,44 +1,25 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
   <Import Project="..\..\tools\common.props" />
   <PropertyGroup>
-    <Description>Tensor class which represents and extends multi-dimensional arrays</Description>
     <TargetFramework>netstandard1.1</TargetFramework>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Description>Tensor class which represents and extends multi-dimensional arrays</Description>
     <PackageTags>.NET Tensor Array corefxlab</PackageTags>
-    <!-- disable automatic compile item inclusion so that we can set the correct metadata on TextTemplate-generated sources -->
-    <EnableDefaultCompileItems>false</EnableDefaultCompileItems>
```
ericstj: Thanks for cleaning this up.

tests/System.Numerics.Tensors.Tests/System.Numerics.Tensors.Tests.csproj
```diff
@@ -17,10 +18,7 @@
     <ProjectReference Include="..\..\src\System.Numerics.Tensors\System.Numerics.Tensors.csproj" />
   </ItemGroup>
   <ItemGroup>
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="$(SystemCompilerServicesUnsafeVersion)" />
-  </ItemGroup>
-  <!-- register for test discovery in Visual Studio -->
-  <ItemGroup>
+    <!-- register for test discovery in Visual Studio -->
     <Service Include="{82a7f48d-3b50-4b1e-b82e-3ada8210c358}" />
```
ericstj: We may be able to remove this now that https://github.com/Microsoft/vstest/issues/472 is fixed.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.NativeMemory.cs
tannergooding: Yes, I just forgot to delete the code the IDE automatically added.

tannergooding: Fixed.

tests/System.Numerics.Tensors.Tests/System.Numerics.Tensors.Tests.csproj
tannergooding: Removed.

samples/System.Numerics.Tensors.Samples/TensorSamples/WrapNativeMemory.NativeMemory.cs
tannergooding: We had already fixed up `NativeMemory<T>` in the test project, so I just merged the code from there into the sample version.

eerhardt:
Can we remove the unused/dead `internal` code, and only keep the `internal` code that is actually callable from a public API?

One idea @ericstj originally had was that we should move the arithmetic and operations code to the unit tests, to ensure that someone from outside the library can still implement it, if they want.

See https://github.com/dotnet/corefxlab/issues/1798#issuecomment-336947110 for more context.

src/System.Numerics.Tensors/System/Numerics/Tensors/Tensor.Operations.tt
ericstj: Was it too expensive to split out?  Typically I don't like IVT in a case like unless we are planning on eventually shipping arithmetic from here.  I was expecting we'd just move the static methods to the test and make sure they worked over the public API.

tannergooding: There is a lot of coupling and it isn't definitely clear that everything should be removed or duplicated, etc. Things like `Tensor.CreateIdentity` depend on arithmetic, as well as various constructors, getters/setters (including things like `GetValue` and `SetValue`), etc. You also wouldn't be able to get rid of the IVTs due to dependencies on the helper utilities.

tannergooding: (most of the coupling is to `TensorArithmetic.Zero` and `TensorArithmetic.One`)

eerhardt: Can we just leave `TensorArithmetic.Zero` and `TensorArithmetic.One` in the library then? I don't understand why'd we want the other dead code in the library.

tannergooding: It would require splitting the T4 template and generated interfaces. I think that it's probably more effort than its worth for getting the package published initially.

What I would really like to do is spend a bit more time looking at the type and understanding this implementation of Tensors and some of the design decisions behind it (I did talk with @ericstj briefly).  I don't think the design needs to change a lot for V1, but I do think there might be a couple core things that we should probably think more deeply about, especially if we are going to make this first class and be part of CoreFX.

Some initial thoughts are:
* reference types, in general, are a bad choice for any interop (IMO) because they require marshaling.
* It seems like quite a bit of this was also designed before we did a lot of the perf work (such as ref structs, hardware intrinsics, etc)
* for the `DenseTensor` case, we are only holding a small struct; and we are likewise only holding a single reference type field for the `SparseTensor` case. So, having `structs` with backing interfaces that define the contract seems much more appropriate at first glance. If we are going to have this be a first class citizen (this also allows things like generic-specialization, if users care about it).

ericstj: I absolutely want to think more about how to improve this type, but I don't see how any of this blocks us from making the arithmetic change.  For the other ideas you list here: please open up issues and we can have those conversations, but we shouldn't be having them here in this PR.

tannergooding: Split the code apart.

src/System.Numerics.Tensors/System/Numerics/Tensors/Tensor.cs
```diff
@@ -124,224 +124,137 @@ public static Tensor<T> CreateFromDiagonal<T>(Tensor<T> diagonal, int offset)
...
-        internal static void ValidateBinaryArgs<T>(Tensor<T> left, Tensor<T> right)
+    /// <summary>
+    /// Represents a multi-dimensional collection of objects of type T that can be accessed by indices.
+    /// </summary>
+    /// <typeparam name="T">type contained within the Tensor.  Typically a value type such as int, double, float, etc.</typeparam>
+    [DebuggerDisplay("{GetArrayString(false)}")]
+    // When we cross-compile for frameworks that expose ICloneable this must implement ICloneable as well.
+    public abstract class Tensor<T> : IList, IList<T>, IReadOnlyList<T>, IStructuralComparable, IStructuralEquatable
+    {
+        internal static T Zero
```
ericstj: I wonder if we can simplify this by returning default(T), if that returns the same in all cases I'd opt for that.

src/System.Numerics.Tensors/AssemblyInfo.cs
```diff
@@ -0,0 +1,7 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Runtime.CompilerServices;
+
+[assembly: InternalsVisibleTo("System.Numerics.Tensors.Tests, PublicKey=0024000004800000940000000602000000240000525341310004000001000100039ac461fa5c82c7dd2557400c4fd4e9dcdf7ac47e3d572548c04cd4673e004916610f4ea5cbf86f2b1ca1cb824f2a7b3976afecfcf4eb72d9a899aa6786effa10c30399e6580ed848231fec48374e41b3acf8811931343fc2f73acf72dae745adbcb7063cc4b50550618383202875223fc75401351cd89c44bf9b50e7fa3796")]
```
ericstj: Can we remove this now?

tests/System.Numerics.Tensors.Tests/Tensor.Operations.tt
```diff
@@ -0,0 +1,240 @@
...
+<#@ output extension=".cs" #>
+<#@ include file="TensorTemplate.ttinclude" #>// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Numerics.Tensors
+{
+    public static partial class TensorOperations
+    {
+        internal static void ValidateBinaryArgs<T>(Tensor<T> left, Tensor<T> right)
```
eerhardt: Why are these methods in the `.tt` file? This class is `partial`, so can they just be in a separate .cs file, to not clutter up the `.tt`?

src/System.Numerics.Tensors/System/Numerics/Tensors/Tensor.cs
tannergooding: I did that originally, but it doesn't cover the case where `T` is an unsupported type.

src/System.Numerics.Tensors/AssemblyInfo.cs
tannergooding: No, the operations and arithmetic still use internals (like the ArrayUtilities).

tests/System.Numerics.Tensors.Tests/Tensor.Operations.tt
tannergooding: I couldn't think of a good name and it seemed reasonable to just put them in the same file as they were used.

src/System.Numerics.Tensors/System/Numerics/Tensors/Tensor.cs
ericstj: I’d be ok with permitting that.  I think that’d let us do sparse reference types (string)

PR #2584, opened 2018-11-10
Update Benchmark.NET version to use EtwProfiler

danmosemsft:
With this change you can temporarily add `[EtwProfiler]` on your benchmark class to get an ETW file. You may need to add `using BenchmarkDotNet.Diagnostics.Windows.Configs;`

Also enabled symbols on the benchmark project for release. It's already done in Microsoft.Experimental.Collections.csproj

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -239,9 +239,6 @@ Network Trash Folder
 Temporary Items
 .apdisk
 
-#dotnet.exe output
-tools/
-
 *.ncrunchproject
 *.ncrunchsolution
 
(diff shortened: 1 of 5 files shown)


.gitignore
```diff
@@ -239,9 +239,6 @@ Network Trash Folder
 Temporary Items
 .apdisk
 
-#dotnet.exe output
-tools/
```
ahsonkhan: Strange. Why **hasn't** this .gitignore affected past updates to the files under `tools/`?

adamsitnik:
To make the console line argument work without any code change you also should use  `<CopyLocalLockFileAssemblies>true</CopyLocalLockFileAssemblies>` in the .csproj.

Why? When we do `dotnet run` the MSBuild copies to the output folder the dependencies that are somehow consumed in the code. If we don't use any type from `BenchmarkDotNet.Diagnostics.Windows` the `.dll` does not get copied and the  console line argument parsing logic fails to load the `.dll` in a dynamic way.

adamsitnik:
@danmosemsft it looks like the commit with `CopyLocalLockFileAssemblies` change did not get merged into master.

Issue #2592, opened 2018-11-19
DictionarySlim Improvements


TylerBrinkley:
## Potential API Changes

```diff
 namespace Microsoft.Collections.Extensions {
     public class DictionarySlim<TKey, TValue> : IReadOnlyCollection<KeyValuePair<TKey, TValue>> where TKey : IEquatable<TKey> {
         // More ref return methods
+        public ref TValue GetValueRef(TKey key); // throws KeyNotFoundException if not contains key
+        public ref TValue TryGetValueRef(TKey key, out bool found); // returns a null reference if possible or else a reference to a private static dummy field when not found

         // Select API differences from Dictionary<K, V>
+        public DictionarySlim(IEnumerable<KeyValuePair<TKey, TValue>> collection); // convenience constructor
+        public int EnsureCapacity(int capacity);
+        public void TrimExcess();
+        public void TrimExcess(int capacity);

         // Existing API
         public DictionarySlim();
         public DictionarySlim(int capacity);
         public int Count { get; }
         public void Clear();
         public bool ContainsKey(TKey key);
         public bool TryGetValue(TKey key, out TValue value);
         public bool Remove(TKey key);
         public ref TValue GetOrAddValueRef(TKey key);
         public Enumerator GetEnumerator();

         public struct Enumerator : IEnumerator<KeyValuePair<TKey, TValue> {
             public KeyValuePair<TKey, TValue> Current { get; }
             public void Dispose();
             public bool MoveNext();
         }
     }
 }
```

## Updates
* Removed `IDictionary<TKey, TValue>` support.

TylerBrinkley:
@danmosemsft @AnthonyLloyd Thoughts?

danmosemsft:
Would you mind including existing API in the diff above?

TylerBrinkley:
Done.

danmosemsft:
Should there be a constructor over `IDictionary<TKey, TValue>` (as well or instead of `IEnumerator<KeyValuePair<TKey, TValue>>`). That is what Dictionary chooses - not sure why.

Either way, in .NET Core it probes for the concrete Dictionary type, which it treats specially - I guess it could literally copy both the arrays in such a case, although it does not.

danmosemsft:
[...]

PR #2594, opened 2018-11-19
DictionarySlim trim

danmosemsft:
Relates to https://github.com/dotnet/corefxlab/issues/2592

2 commits - 
Remove Key/Value Enumerables to reduce instantiated size of type as discussed above
Remove Capacity to reduce unnecessarily publicness


danmosemsft:
Oh, let me add XML comments too

tests/Microsoft.Experimental.Collections.Tests/DictionarySlimTests.cs
```diff
@@ -723,6 +537,22 @@ public void UsedIEquatable2()
...
+    {
+        // Capacity is not exposed publicly, but is valuable in tests to help
+        // ensure everything is working as expected internally
+        public static int GetCapacity<TKey, TValue>(this DictionarySlim<TKey, TValue> dict) where TKey : IEquatable<TKey>
+        {
+            FieldInfo fi = typeof(DictionarySlim<TKey, TValue>).GetField("_entries", BindingFlags.NonPublic | BindingFlags.Instance);
+            Object entries = fi.GetValue(dict);
+
+            PropertyInfo pi = typeof(Array).GetProperty("Length");
+            int capacity = (int)pi.GetValue(entries);
```
safern: We could just return this?

danmosemsft:
OK now also 
Added XML docs (file temporarily included in the commit as I cannot find a way to consume $(OutputPath) in the project)
Removed Copy to (low value, extra size)

TylerBrinkley:
I think a key and value collection are still both needed just not the nested Enumerator struct. I think that's what @jkotas meant.

danmosemsft:
> I think a key and value collection are still both needed just not the nested Enumerator struct. I think that's what @jkotas meant.

I'm missing something, if we still have `ICollection<TKey>` but no Enumerator struct what would it return from `GetEnumerator<TKey>` ?

TylerBrinkley:
yield return? Though that would create an iterator class. I guess we can just use a LINQ Select clause to replicate the Key and Value collections where needed.

TylerBrinkley:
Looks good to me.

danmosemsft:
> I guess we can just use a LINQ Select clause to replicate the Key and Value collections where needed.

Yes, that's what I did in the tests. As with anything - it is always easier to add than remove API.

danmosemsft:
@safern is going to get the package signed and on NuGet in the next few days. After that is up @AnthonyLloyd are you interested in sending in the updated benchmark game submission? I think you deserve most of the credit.

AnthonyLloyd:
> @safern is going to get the package signed and on NuGet in the next few days. After that is up @AnthonyLloyd are you interested in sending in the updated benchmark game submission? I think you deserve most of the credit.

Yes sure, thanks you all. It may be a good idea to wait until the package has a few downloads. In the benchmarks game they want to see that it's not just a package put up for the benchmark only.

danmosemsft:
@anthonylloyd it already has 50K downloads :https://www.nuget.org/packages/Microsoft.Experimental.Collections/1.0.3-alpha

Issue #2585, opened 2018-11-11
20 warnings building repo


danmosemsft:
```
D:\j\workspace\windows_nt_re---4abe806d\tests\Benchmarks\Benchmarks.csproj : warning NU1701: Package 'codetitans-json 1.8.3' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.1'. This package may not be fully compatible with your project.
D:\j\workspace\windows_nt_re---4abe806d\tests\Benchmarks\Benchmarks.csproj : warning NU1701: Package 'jayrock-json 0.9.16530.1' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.1'. This package may not be fully compatible with your project.
D:\j\workspace\windows_nt_re---4abe806d\tests\Benchmarks\Benchmarks.csproj : warning NU1701: Package 'System.Text.Json 2.0.0.11' was restored using '.NETFramework,Version=v4.6.1' instead of the project target framework '.NETCoreApp,Version=v2.1'. This package may not be fully compatible with your project.
Utf8BufferEncoder.cs(32,23): warning CS0618: 'UrlEncoder.Decode(ReadOnlySpan<byte>, Span<byte>)' is obsolete: 'Use UrlDecoder.Utf8.Decode' [D:\j\workspace\windows_nt_re---4abe806d\src\System.Text.Encodings.Web.Utf8\System.Text.Encodings.Web.Utf8.csproj]
Utf8BufferEncoder.cs(39,23): warning CS0618: 'UrlEncoder.DecodeInPlace(Span<byte>)' is obsolete: 'Use UrlDecoder.Utf8.DecodeInPlace' [D:\j\workspace\windows_nt_re---4abe806d\src\System.Text.Encodings.Web.Utf8\System.Text.Encodings.Web.Utf8.csproj]
Utf8BufferEncoder.cs(14,17): warning CS0618: 'UrlEncoder.TryEncode(ReadOnlySpan<byte>, Span<byte>, out int)' is obsolete: 'Use UrlEncoder.Utf8.Encode' [D:\j\workspace\windows_nt_re---4abe806d\src\System.Text.Encodings.Web.Utf8\System.Text.Encodings.Web.Utf8.csproj]
[...]

PR #2600, opened 2018-11-24
Suppress some warnings

danmosemsft:
Relates to https://github.com/dotnet/corefxlab/issues/2585

I also updated the BDN version so that `[DryRunJob]` works.

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/Native/NativeMemoryPool.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/Native/NativeMemoryPool.cs
@@ -24,7 +24,7 @@ namespace System.Buffers.Native
         int _bufferCount;
         bool[] _rented;
 
-        public static NativeMemoryPool Shared => s_shared;
+        public static new NativeMemoryPool Shared => s_shared;
 
         public NativeMemoryPool(int bufferSize, int bufferCount = 10)
         {
--- a/build.sh
+++ b/build.sh
@@ -91,7 +91,7 @@ fi
 
 echo "Building solution $myFile..."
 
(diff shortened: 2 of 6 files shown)


src/System.Buffers.Experimental/System/Buffers/Native/NativeMemoryPool.cs
```diff
@@ -24,7 +24,7 @@ public unsafe sealed partial class NativeMemoryPool : MemoryPool<byte>
         int _bufferCount;
         bool[] _rented;
 
-        public static NativeMemoryPool Shared => s_shared;
+        public static new NativeMemoryPool Shared => s_shared;
```
ahsonkhan: I don't understand the reasoning for this. Why is the new keyword required here (i.e. why do we get a warning otherwise)?

ahsonkhan:

src/System.Buffers.Experimental/System/Buffers/Native/NativeMemoryPool.cs
danmosemsft: The base class `MemoryPool` has a property with the same name:
```c#
        /// <summary>
        /// Returns a singleton instance of a MemoryPool based on arrays.
        /// </summary>
        public static MemoryPool<T> Shared => s_shared;
```
This one hides it. You are allowed to do that but it will warn unless you indicate using the `new` keyword that you intended to do it. I am not sure why it is not an error.

The advantage of `new` over `virtual`/`overrides` is presumably that there is no virtual dispatch overhead.

danmosemsft:
Thanks @ahsonkhan . We should get your JSON test disabled for Ubuntu so CI is clean.

PR #2596, opened 2018-11-21
Use GenerateDocumentationFile so that the outputpath is calculated by the SDK

safern:

Issue #2390, opened 2018-07-09
API Proposal: FileSystemState


MisinformedDNA:
I decided that `PollingFileSystemWatcher` was not a great fit for my project. First off, my operations department has all sorts of tools for monitoring scheduled jobs, but not so much for continuous jobs. The other problem with continuous jobs is that they will stop at some point and file system changes could be missed. Furthermore, to make `PollingFileSystemWatcher` behave how I wanted it to, I've been jumping through a lot of hoops and asking for [features](https://github.com/dotnet/corefxlab/issues/2222) that may not otherwise be needed.

So I created [FileSystemState](https://github.com/MisinformedDNA/YellowCounter.FileSystemState). (NOTE: I created it in a separate repo since I need it for a project sooner than later.)

`FileSystemState` is mostly just a stripped down version of `PollingFileSystemWatcher`. It gets rid of the timer and eventing features. If you want changes, you ask for it and you get a list back. You can save and load state so you aren't required to have it running all the time.

## Requirements

- Can run as a scheduled job, instead of a continuous job
- All file changes are eventually detected
- Can be used as the basis of `PollingFileSystemWatcher`

## APIs

```csharp
namespace System.IO
{
    public class FileSystemState
    {
        public FileSystemState(string path, string filter = "*", EnumerationOptions options = null);
        public string Path { get; set; }
        public string Filter { get; set; }
        public EnumerationOptions EnumerationOptions { get; set; }

        public void LoadState();
        public void LoadState(Stream stream);
        public void SaveState(Stream stream);
        public FileChangeList GetChanges();

        protected internal virtual void DetermineChange(string directory, ref FileChangeList changes, ref FileSystemEntry file);
        protected internal virtual bool ShouldIncludeEntry(ref FileSystemEntry entry);
        protected internal virtual bool ShouldRecurseIntoEntry(ref FileSystemEntry entry);
    }

    public class FileChangeList : List<FileChange> { }

    public class FileChange
    {
        public string Directory { get; }
        public string FileName { get; }
        public WatcherChangeTypes ChangeType { get; }
    }
}
```

**Notes:**
- Testing is much easier and cleaner without polling and eventing
- `PollingFileSystemWatcher` would inherit from `FileSystemState` and the timer would simply call `GetChanges` and raise an event, in items were returned
- We could remove serialization from `PollingFileSystemWatcher`
[...]

PR #2603, opened 2018-11-27
Disable, remove, or fix JSON tests that consume too much memory.

ahsonkhan:
This should, hopefully, resolve the intermittent test failures we have been seeing in CI (especially for Ubuntu). The Json tests, in total, now allocate < 2 GB (instead of ~7).

Addressing:
https://github.com/dotnet/corefxlab/pull/2600#issuecomment-441402322
https://github.com/dotnet/corefxlab/pull/2596#issuecomment-440608860

As an added bonus they also take less time to complete now (< 2 minutes instead of ~4).

cc @danmosemsft

### DIFF
--- a/src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
+++ b/src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
@@ -97,9 +97,9 @@ namespace System.IO
             {
                 // Ignore `.Current`
             }
-            var changes = enumerator.Changes;
+            FileChangeList changes = enumerator.Changes;
 
-            foreach (var value in _state)
+            foreach (FileState value in _state)
             {
                 if (value._version != _version)
                 {
@@ -144,7 +144,7 @@ namespace System.IO
 
             _state.Values[index]._version = _version;
 
-            var previousState = _state.Values[index];
+            FileState previousState = _state.Values[index];
             if (file.LastWriteTimeUtc != previousState.LastWriteTimeUtc || file.Length != previousState.Length)
             {
                 changes.AddChanged(directory, previousState.Path);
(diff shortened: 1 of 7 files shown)


src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
```diff
@@ -167,20 +167,26 @@ internal void UpdateState(string directory, ref FileChangeList changes, ref File
         /// </summary>
         public void Dispose()
         {
-            _disposed = true;
-            _timer.Dispose();
-            Dispose(true);
-            GC.SuppressFinalize(this);
+            lock (_timer)
```
ahsonkhan: We clearly need some sort of synchronization mechanism here or else we keep getting intermittent test failures here: https://github.com/dotnet/corefxlab/pull/2603/files#diff-bf9f0af07091216198ca86b115fd2777R213

@JeremyKuhne, @MisinformedDNA, does this change of using a lock during Dispose make sense to you?

Otherwise, we should skip all the Polling tests against https://github.com/dotnet/corefxlab/issues/420#issuecomment-441950347.

ahsonkhan: Never mind. That didn't resolve the issue. Will revert.

ahsonkhan:
@dotnet-bot test this please

ahsonkhan:
@dotnet-bot test this please

ahsonkhan:
@dotnet-bot test this please

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
MisinformedDNA: I don't fully understand why it fails at all. I believe it is only certain platforms. With that and many other issues (like lack of extensibility hooks I needed), I [proposed](https://github.com/dotnet/corefxlab/issues/2390) a base class for PollingFileSystemWatcher that would make the unit testing much easier, but the discussion stalled out.

ahsonkhan: > I believe it is only certain platforms.

I was able to repro the failures on windows (x64, debug) for all the tests that I have marked as skip. It is a (or several) race condition(s) somewhere.

Many of the tests fail in < 100 iterations, but some took 1000s to repro, locally.

```text
Unhandled Exception: System.ObjectDisposedException: Cannot access a disposed object.
```

ahsonkhan:
@dotnet-bot test this please

src/System.IO.FileSystem.Watcher.Polling/System/IO/PollingFileSystemWatcher.cs
ahsonkhan: If anyone wants the test projects that I ran to consistently repro the failures (essentially its just separate applications that run a single test in a loop), here it is:
https://github.com/ahsonkhan/corefxlab/tree/PollingTestRepros/repro_test_failures

Run them like any dotnet application.
`SerializableTests\RoundTripAfterStartedTest> dotnet run`

Or you could just add a loop in the test itself and you should be able to see the failure in VS test explorer.

PR #2609, opened 2018-11-30
Few OrderedDictionary Updates

TylerBrinkley:
Removed unnecessary GetAt, GetValue, SetAt, and SetValue methods as the indexers work, even in the case where TKey is an int as you can still call the key indexer by specifying the key parameter name.


ahsonkhan:
> I suppose it's being enforced by an editor config file but am unable to see where.

Which solution? corefxlab.sln?

This is the only editorconfig we have in the repo: https://github.com/dotnet/corefxlab/blob/master/.editorconfig

Have you built from the command line first before opening the solution in VS? I don't see such errors (for example if I do something like: `var temp = ""`).

TylerBrinkley:
Sorry, I think it's due to my own machine settings have them set as a Warning but this project must have changed to specify Warnings as Errors.

danmosemsft:
LGTM

PR #2612, opened 2018-12-07
Update UTF-8 JsonWriter APIs based on previous API feedback

ahsonkhan:
Addressing most of https://github.com/dotnet/apireviews/pull/80 + other changes.

API proposal updated: https://github.com/dotnet/corefx/issues/33552

**TODO:**
- [x] Finish implementation of API skeletons.
- [ ] Add more tests (especially exceptions and edge cases) - **outside this PR.**
- [x] Add BytesWritten support.
- [ ] Figure out how to deal with escaping (either implement from scratch or leverage exising APIs) - cc @GrabYourPitchforks - **outside this PR.**
- [ ] Fixup and remove the TODOs left in code. - **outside this PR.**
- [x] Add state-passing and write sample/test for async support.
- [ ] Re-evaluate size of JsonWriter and maybe reduce it by packing more info in available bits/custom struct layout. - **outside this PR.**
- [ ] XML comments/documentation - **outside this PR.**
- [ ] Consider making the type non-generic with a ctor overload that accepts span (or memory) - **outside this PR.**

cc @steveharter, @stephentoub, @bartonjs, @KrzysztofCwalina


src/System.Text.JsonLab/System/Text/Json/BitStack.cs
```diff
@@ -0,0 +1,169 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.JsonLab
+{
+    public struct BitStack
```
ahsonkhan: Ignore this. It is a copy of the internal API from https://github.com/dotnet/corefx/blob/master/src/System.Text.Json/src/System/Text/Json/BitStack.cs which I just copied for convenience. This will go away once the code moves to corefx.

src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs
```diff
@@ -53,5 +53,19 @@ internal static class JsonConstants
...
+        public const int RemoveFlagsBitMask = 0x7FFFFFFF;
+        public const int MaxDepth = (int.MaxValue - 2_000_001_000) / 2;  // 73_741_323 (to account for double space indentation), leaving 1_000 buffer for "JSONifying"
+        public const int MaxTokenSize = 1_000_000_000; // 1 GB
+        public const int MaxCharacterTokenSize = 1_000_000_000 / 3; // 333 million characters, i.e. 333 MB
+
+        public const int MaximumInt64Length = 20;   // 19 + sign (i.e. -9223372036854775808)
+        public const int MaximumUInt64Length = 20;  // i.e. 18446744073709551615
+        public const int MaximumDoubleLength = 32;  // default (i.e. 'G') TODO: Should it be 22?
+        public const int MaximumSingleLength = 32;  // default (i.e. 'G') TODO: Should it be 13?
+        public const int MaximumDecimalLength = 32; // default (i.e. 'G') TODO: Should it be 31?
```
ahsonkhan: What should this be? @tannergooding 

I don't understand why some of these are so large: https://github.com/dotnet/coreclr/blob/80c0a0ea2446b665d13e1632422802f4bf208ae5/src/System.Private.CoreLib/shared/System/Number.NumberBuffer.cs#L14

src/System.Text.JsonLab/System/Text/Json/Utf8JsonWriter.StaticHelpers.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        // TODO: Either implement the escaping helpers from scratch or leverage the upcoming System.Text.Encodings.Web.TextEncoder APIs
+        public static OperationStatus EscapeString(string value, Span<byte> destination, out int consumed, out int bytesWritten)
+        {
+            throw new NotImplementedException();
+        }
+
+        public static OperationStatus EscapeString(ReadOnlySpan<char> value, Span<byte> destination, out int consumed, out int bytesWritten)
+        {
+            //    // TODO:
```
ahsonkhan: Intentionally commented out for now.

src/System.Text.JsonLab/System/Text/Json/Utf8JsonWriter.WriteProperties.cs
```diff
@@ -0,0 +1,2110 @@
...
+        {
+            Debug.Assert(_options.Formatted || !_options.SkipValidation);
+
+            if (_options.Formatted)
+            {
+                if (!_options.SkipValidation)
+                {
+                    ValidateWritingPropertyWithEncoding(propertyName);
+                }
+                WriteStringFormattedWithEncoding(propertyName, value)
```
ahsonkhan: Any suggestions on reducing code duplication here (or in general)? The code for `Guid`, `DateTime`, and `DateTimeOffset` is essentially identical (all calling into the `Utf8Formatter`).

src/System.Text.JsonLab/System/Text/Json/Utf8JsonWriter.WriteArrayValues.cs
```diff
@@ -0,0 +1,164 @@
...
+            _currentDepth |= 1 << 31;
+            _tokenType = JsonTokenType.EndArray;
+        }
+
+        private void WriteArrayFast(ReadOnlySpan<byte> propertyName, ReadOnlySpan<long> values)
+        {
+            if (values.Length > 0 && values.Length < (int.MaxValue - 2) / (JsonConstants.MaximumInt64Length + 1))
+            {
+                // Calculated based on the following: '[number0,number1,...,numberN]'
+                int bytesNeeded = 2 + values.Length * (1 + JsonConstants.MaximumInt64Length);
```
ahsonkhan: Is there a concern with pessimistically asking `IBufferWriter<byte>` for more bytes than needed (i.e. get the bytes needed for the worst case)?

ahsonkhan: cc @davidfowl, @pakrym - any thoughts on the approach of over-asking for the amount of bytes from IBufferWriter for the worst case? For instance, if we are writing a 4, single digit integers as a JSON array, I ask for 4*20 + 3 + 2 =85 bytes rather than 4 + 3+ 2 = 9 (for example: `[1, 2, 3, 4]`). This way I don't have to count the digits of each element (at most it would be 20 digits, for an Int64).

pakrym: Over-asking a little bit is fine and considering IBufferWriter is free to give you less than requested if it doesn't have enough space.

src/System.Text.JsonLab/System/Text/Json/JsonConstants.cs
tannergooding: The `NumberBufferLength` values are large enough to represent the exact string representation of any given input plus one for a rounding digit and one for the null-terminating character.

For example, the input with the longest exact string for a `double` value is `double.Epsilon`, whose exact representation has 767 significant digits (there are an additional 307 leading zeros, giving 1074 digits total; but we don't need to track leading or trailing zeros). If a user gives us an input string that contains all 1074 digits of this value, we need to be able to parse that. If they provide additional digits, we also have to consider them to determine which direction we round (do we round down to `double.Epsilon` or do we round up to the next representable value above `double.Epsilon`). If the 768th digit is a `5`, then we may be in "midpoint-rounding" and we have to continue processing the rest of the string until we find a non-zero digit to determine if this is actually midpoint (the rest of the digits in the input were 0, in which case we round up or down to the value with the first bit set) or if we are slightly above (one or more of the digits after the 768th were non-zero) in which case we round up.

PR #2629, opened 2019-01-07
Left over fixes for JsonWriter

ahsonkhan:


Tornhoof:
@ahsonkhan Are you aware that not all of your escaping logic is required?
Assuming RFC8259 as the last JSON RFC, then https://tools.ietf.org/html/rfc8259#section-7:
- You need to escape <0x20, 0x22 and 0x5C
- Most serializers also escape 0x2F ('/'), because it has a short escape sequence too and helps with Javascript compatibility, but it's not really required.

All your logic for characters outside of ASCII is not required, you can simply write the character encoded in UTF8 into the JSON (see:  unescaped = %x20-21 / %x23-5B / %x5D-10FFFF)
It is allowed to escape them (because you can escape everything), but it's not required.  Your code looks fine, but I'm not sure if you should activate that by default, because it's probably quite slow, you might want to just extend your switch logic for the short values to handle 0x0-0x1F and write them in \u001F style.

ahsonkhan:
> Are you aware that not all of your escaping logic is required?

Yep :)

> Your code looks fine, but I'm not sure if you should activate that by default, because it's probably quite slow, you might want to just extend your switch logic for the short values to handle 0x0-0x1F and write them in \u001F style.

Here, we are eagerly escaping more than what the spec states, by **default,** as part of SDL requirements. I am following what the [JavaScriptEncoder](https://github.com/dotnet/corefx/blob/master/src/System.Text.Encodings.Web/src/System/Text/Encodings/Web/JavaScriptEncoder.cs) does which forbids more characters to avoid security pitfalls.

I may consider providing options to customize that behavior, if there is a strong need, but I don't think we can get away with skipping escaping those characters.

PR #2631, opened 2019-01-10
Add Signing in order to be able to publish packages to nuget

safern:
Add targets to be able to sign packages and its contents in order to be able to push experimental or preview packages to nuget.org

This is based on arcade's Microsoft.DotNet.SignTool package.

I will test the build definition before merging.

### DIFF
--- a/NuGet.Config
+++ b/NuGet.Config
@@ -8,12 +8,13 @@
     nuget package source please add it to the item group in dir.props instead.
   -->
   <config>
-    <add key="repositoryPath" value="..\packages" />
+    <add key="repositoryPath" value=".nuget\packages" />
   </config>
   <packageSources>
     <add key="dotnet.myget.org dotnet-core" value="https://dotnet.myget.org/F/dotnet-core/api/v3/index.json" protocolVersion="3" />
     <add key="dotnet.myget.org dotnet-buildtools" value="https://dotnet.myget.org/F/dotnet-buildtools/" />
     <add key="dotnet.myget.org dotnet-corefxlab" value="https://dotnet.myget.org/F/dotnet-corefxlab/" />
+    <add key="dotnet blob feed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
     <add key="nuget.org" value="https://api.nuget.org/v3/index.json" protocolVersion="3" />
     <add key="dotnet.myget.org dotnet-cli" value="https://dotnet.myget.org/F/dotnet-cli/" />
     <add key="coreclr-xunit" value="https://www.myget.org/F/coreclr-xunit/api/v2" />
--- a/scripts/package.ps1
+++ b/scripts/package.ps1
@@ -1,7 +1,8 @@
 ﻿Param(
     [string]$Configuration="Debug",
     [string]$ApiKey,
-    [string]$BuildVersion=[System.DateTime]::Now.ToString('preview2-yyMMdd-1'),
+    [string]$BuildVersion=[System.DateTime]::Now.ToString('eyyMMdd-1'),
+    [string]$FilesToSignFilter="*.nupkg",
     [string]$SignType="test",
     [switch]$Sign
 )
@@ -43,7 +44,7 @@ foreach ($file in [System.IO.Directory]::EnumerateFiles("$repoRoot\src", "*.cspr
 if ($Sign)
 {
     Write-Host "** Signing packages with SignType = $SignType **"
-    & $dotnetExePath msbuild "$repoRoot\tools\signing\sign.proj" /p:SignType="$SignType" /p:PackagesPath="$packagesPath"
+    & $dotnetExePath msbuild "$repoRoot\tools\signing\sign.proj" /p:SignType="$SignType" /p:PackagesPath="$packagesPath" /p:FilesToSignFilter="$FilesToSignFilter"
 
     if (!$?) {
         Write-Error "Failed to sign the packages"
--- a/scripts/package.ps1
+++ b/scripts/package.ps1
@@ -43,9 +43,10 @@ foreach ($file in [System.IO.Directory]::EnumerateFiles("$repoRoot\src", "*.cspr
 if ($Sign)
 {
     Write-Host "** Signing packages with SignType = $SignType **"
-    & $dotnetExePath msbuild "$repoRoot\tools\sign.proj" /p:SignType="$SignType" /p:PackagesPath="$packagesPath"
+    & $dotnetExePath msbuild "$repoRoot\tools\signing\sign.proj" /p:SignType="$SignType" /p:PackagesPath="$packagesPath"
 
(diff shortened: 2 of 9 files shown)


tools/signing/sign.targets
```diff
@@ -0,0 +1,61 @@
...
+    <SignType Condition="'$(SignType)' == ''">test</SignType> 
+
+    <_DryRun>true</_DryRun>
+    <_DryRun Condition="'$(BUILD_BUILDNUMBER)' != ''">false</_DryRun>
+    <_TestSign>false</_TestSign>
+    <_TestSign Condition="'$(SignType)' == 'test'">true</_TestSign>
+
+    <_DesktopMSBuildRequired Condition="'$(_DryRun)' != 'true'">true</_DesktopMSBuildRequired>
+
+    <_SigningTempDir Condition="'$(_SigningTempDir)' == ''">$(MSBuildProjectDirectory)\obj\</_SigningTempDir>
```
JohnTortugo: You may want to add a `tmp` directory here. A bunch of folder\files will be created during the signing.

tools/signing/sign.targets
```diff
@@ -0,0 +1,61 @@
...
+  <Import Project="..\dependencies.props" />
+  <PropertyGroup>
+    <NuGetPackageRoot>$(RestorePackagesPath)</NuGetPackageRoot>
+    <NuGetPackageRoot Condition="!HasTrailingSlash('$(NuGetPackageRoot)')">$(NuGetPackageRoot)\</NuGetPackageRoot>
+    <SignType Condition="'$(SignType)' == ''">test</SignType> 
+
+    <_DryRun>true</_DryRun>
+    <_DryRun Condition="'$(BUILD_BUILDNUMBER)' != ''">false</_DryRun>
+    <_TestSign>false</_TestSign>
+    <_TestSign Condition="'$(SignType)' == 'test'">true</_TestSign>
```
JohnTortugo: Maybe you can change this to just use some logic around SignType

safern: Could you elaborate? I thought that is what I was doing.

JohnTortugo: I meant to remove the `_TestSign` property and just use `SignType` instead. But looking again I think that might just make things more complicated anyway..

scripts/package.ps1
```diff
@@ -38,6 +40,16 @@ foreach ($file in [System.IO.Directory]::EnumerateFiles("$repoRoot\src", "*.cspr
     }
 }
 
+if ($Sign)
+{
+    Write-Host "** Signing packages with SignType = $SignType **"
+    & $dotnetExePath msbuild "$repoRoot\tools\sign.proj" /p:SignType="$SignType" /p:PackagesPath="$packagesPath"
+
+    if (!$?) {
+        exit $lastExitCode
```
ahsonkhan: Nit: Write an error message before exiting

`Write-Error "Failed to sign the packages"`

tools/dependencies.props
```diff
@@ -12,5 +12,9 @@
     <TestSdkVersion>15.0.0</TestSdkVersion>
     <XunitVersion>2.2.0</XunitVersion>
     <BenchmarkDotNetVersion>0.11.3</BenchmarkDotNetVersion>
+    <SnVersion>1.0.0</SnVersion>
+    <VSWhereVersion>2.5.2</VSWhereVersion>
+    <MicroBuildCoreVersion>0.2.0</MicroBuildCoreVersion>
+    <MicrosoftDotNetSignToolVersion>1.0.0-beta.19058.5</MicrosoftDotNetSignToolVersion>
```
ahsonkhan: Where are these versions coming from?

tools/signing/sign.targets
```diff
@@ -0,0 +1,61 @@
+<Project DefaultTargets="Sign">
+  <Import Project="..\dependencies.props" />
+  <PropertyGroup>
+    <NuGetPackageRoot>$(RestorePackagesPath)</NuGetPackageRoot>
+    <NuGetPackageRoot Condition="!HasTrailingSlash('$(NuGetPackageRoot)')">$(NuGetPackageRoot)\</NuGetPackageRoot>
```
ahsonkhan: Can you please explain why `HasTrailingSlash` check is required here?

tools/sign.proj
```diff
@@ -0,0 +1,14 @@
+<Project DefaultTargets="Build">
```
ahsonkhan: Can this also go under the signing directory?

tools/dependencies.props
safern: This versions are coming from the arcade repo.

MicrosoftDotNetSignToolVersion is built in arcade.
Sn and VSWhere I just took the versions that arcade SDK was using to do signing.

tools/signing/sign.targets
safern: Just to avoid having a double '\' in the path. I can remove it if desired.

tools/sign.proj
safern: Yup.

NuGet.Config
```diff
@@ -8,12 +8,13 @@
     nuget package source please add it to the item group in dir.props instead.
   -->
   <config>
-    <add key="repositoryPath" value="..\packages" />
+    <add key="repositoryPath" value=".nuget\packages" />
```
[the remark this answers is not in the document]
safern: Nope, but nuget should create it when restoring from what I know.

scripts/package.ps1
```diff
@@ -1,7 +1,9 @@
 ﻿Param(
     [string]$Configuration="Debug",
     [string]$ApiKey,
-    [string]$BuildVersion=[System.DateTime]::Now.ToString('preview2-yyMMdd-1')
+    [string]$BuildVersion=[System.DateTime]::Now.ToString('preview2-yyMMdd-1'),
```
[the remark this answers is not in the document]
safern: Sure, I can do that. Do we want experimental? Or beta?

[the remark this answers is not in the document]
ahsonkhan: They used to just have the `e` prefix to denote experimental (for example: 0.1.0-e160310-1)

https://dotnet.myget.org/feed/dotnet-corefxlab/package/nuget/System.Drawing.Graphics

Changing it back to 'e' would be fine, imo (as long as it doesn't break the sort order and the newest ones show up before the ones listed as preview).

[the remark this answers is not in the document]
safern: I thin preview would sort as latest (because of alphabetical ordering) so then beta would probably work better.

cc: @ericstj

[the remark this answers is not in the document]
ericstj: If you move backwards in prerelease specifier you should bump all the package versions so that "latest" does the right thing.

safern:
@ahsonkhan I fixed the versioning and also added the ability to pass a filter to the signing targets so with that we could only sign a specific package if wanted. PTAL

PR #2633, opened 2019-01-10
Removing S.N.Tensors now that it is part of CoreFX proper

tannergooding:
Removed all the tensor code now that we have moved it back into CoreFX (modulo the samples, which I did remove here but which aren't part of CoreFX).

It looked like, for things that have been moved into CoreFX; the code has been deleted rather than moved into `archived_projects`, but let me know if we would rather archive it instead.

CC. @ericstj, @eerhardt, @danmosemsft


ericstj:
I'm OK with deleting.  For the samples we may want to add a test cases in CoreFx that implement native bindings using the tensor types so that we can fill this gap introduced by their removal.  I think its an important part of the API design to make sure we have a playground for testing out their use in this scenario.

tannergooding:
This issue was moved to dotnet/corefx#34509

PR #2634, opened 2019-01-17
Fix packages to be compliant with nuget's requirements

safern:
PackageLicenseUrl is being deprecated and in all other repos we use PackageLicenseFile and put both, LICENSE and THIRD-PARTY-NOTICES files in our packages.

### DIFF
--- a/tools/common.props
+++ b/tools/common.props
@@ -2,13 +2,13 @@
   <Import Project="dependencies.props" />
   <PropertyGroup>
     <VersionSuffix Condition="$(VersionSuffix) == ''">$([System.DateTime]::Now.ToString(preview2-yyMMdd-1))</VersionSuffix>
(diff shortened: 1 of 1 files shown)


tools/common.props
```diff
@@ -17,5 +17,11 @@
     <LangVersion>latest</LangVersion>
     <!-- We are usually using a preview SDK -->
     <SuppressNETCoreSdkPreviewMessage>true</SuppressNETCoreSdkPreviewMessage>
+    <RepoRoot>$(MSBuildProjectDirectory)\..\..\</RepoRoot>
```
ahsonkhan: Where does `MSBuildProjectDirectory` point to?

safern: The csproj that is being packed.

ericstj: I wouldn't depend on the naming convention of the project paths.  This should be `$(MSBuildThisFileDirectory)\..`

Issue #2635, opened 2019-01-18
[Proposal] Start a project around various numeric types


tannergooding:
We have a number of proposals on CoreFX around exposing various new numeric types. It would likely be a good idea to have a project here to prototype some of these.

Some of the ideas that I think would likely be applicable are the following:

Types which are generally considered ABI primitives:
* The IEEE `binary128` (quad-precision) floating-point type
* The IEEE `decimal32`, `decimal64`, and `decimal128` floating-point types
* The `Int128` and `UInt128` types

Types which are fairly core to machine-learning and graphics scenarios:
* The IEEE `binary16` (half-precision) floating-point type
* The `bfloat16` floating-point type

Types which allow arbitrary precision arithmetic:
* The IEEE `binary{k}` (where `k >= 128` and `(k % 32) == 0`) floating-point type (effectively a `BigFloat` type)
* The IEEE `decimal{k}` (where `k >= 32` and `(k % 32) == 0`) floating-point type (effectively a `BigDecimal` type)

AlgorithmsAreCool:
This is cool, I just had to roll my own BigRational to support exact precision decimal arithmetic for a project.

Would you consider mutable arbitrary precision numbers?
I recall the original discussions around wanting to keep the semantics of BigInteger the same as normal ints.

But when you are doing alot of calculations the copying becomes a big (heh) pain.

It would be really nice to have a BigIntegerBuilder class or something that could accumulate repeated updates and only reallocate buffers when it was actually needed (or rent buffers in smaller segments).

I can provide a more concrete use case if you are open to the concept.

fanoI:
Happy to see this moving forward.

Can I hope the implementations of these new numeric types will be all in managed code?

tannergooding:
> Can I hope the implementations of these new numeric types will be all in managed code?

Yes. Although they likely will take advantage of the new hardware intrinsics feature, where applicable.

tannergooding:
@AlgorithmsAreCool, I think having the various `arbitrary precision types` and corresponding `Builder` types is probably the correct way forward.

This is a pattern that has been used with fairly good success for other data types and `BigInteger` has already shipped as `readonly`, so it would likely be good to keep things consistent moving forward.

Happypig375:
If anyone need this right now, there is https://github.com/peteroupc/Numbers which implements all of this via configurable context settings.

PR #2638, opened 2019-01-22
Adding the beginnings of a binary16 (half-precision) floating-point implementation

tannergooding:
This is work towards https://github.com/dotnet/corefxlab/issues/2635 and starts with a basic implementation for `Half`.

The type exposes the same interfaces (modulo `IConvertible` and `ISpanFormattable`) as `System.Single` and `System.Double`. For `ToString` or `Parse`, the implementations currently throw `NotImplementedException`, since the actual implementation would be a small variant on the existing `NumberToString` and `NumberToDouble/Single` implementations.

The various operators (such as the casting operators as well as `+`, `-`, `*`, `/`, and `%`) still need to be implemented.


ahsonkhan:
Add a test project with at least some basic smoke tests.

tannergooding:
> Add a test project with at least some basic smoke tests.

I plan on it, will likely do so this weekend.

src/System.Numerics.Experimental/System.Numerics.Experimental.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\tools\common.props" />
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+    <Description>Experimental numeric types.</Description>
+    <PackageTags>Numerics corefxlab</PackageTags>
+    <TargetFramework>netcoreapp2.1</TargetFramework>
```
ahsonkhan: Target 3.0?

src/System.Numerics.Experimental/System.Numerics.Experimental.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+  <Import Project="..\..\tools\common.props" />
+  <PropertyGroup>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
```
ahsonkhan: Is this needed?

src/System.Numerics.Experimental/System.Numerics.Experimental.csproj
tannergooding: Fixed.

src/System.Numerics.Experimental/System.Numerics.Experimental.csproj
tannergooding: Not right now. Removed.

src/System.Numerics.Experimental/Half.cs
```diff
@@ -0,0 +1,357 @@
...
+    [StructLayout(LayoutKind.Sequential)]
+    public readonly struct Half : IComparable, IFormattable, IComparable<Half>, IEquatable<Half>
+    {
+        internal const NumberStyles DefaultParseStyle = NumberStyles.Float | NumberStyles.AllowThousands;
+
+        //
+        // Constants for manipulating the internal bit-representation
+        //
+
+        internal const ushort SignMask = 0x8000;
```
danmosemsft: Seems these fields should either be public or private?

danmosemsft: If they are private they still take up space in the IL of course, seems a shame since they are "inlined" by the C# compiler. A case where a preprocessor would be nice.

tannergooding: Made private. We shouldn't have these be public unless we also plan to expose similar constants on `Single`/`Double`. Such constants which might be useful for certain types of low-level manipulation, but the raw types might also not be the appropriate place to expose them.

tannergooding:
> I plan on it, will likely do so this weekend.

It only took three months, but I have added some tests mirroring the existing ones we have for `Single`/`Double`.

This should be good for an initial merge; and then work can continue progressing as time is available.

src/System.Numerics.Experimental/System.Numerics.Experimental.csproj
ahsonkhan: This still needs to be fixed.

tests/System.Numerics.Experimental.Tests/HalfTests.cs
```diff
@@ -0,0 +1,364 @@
...
+
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using Xunit;
+
+namespace System.Numerics.Experimental.Tests
+{
+    public partial class HalfTests
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ahsonkhan: Why add these for tests?

tannergooding: I just copied the existing method template from BitConverter and didnt consider removing it.

tannergooding:
@ahsonkhan, @danmosemsft. Is this good to merge?

src/System.Numerics.Experimental/Properties/Strings.resx
```diff
@@ -0,0 +1,123 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="Arg_MustBeHalf" xml:space="preserve">
+    <value>Object must be of type Half.</value>
```
danmosemsft: Nit, should probably say System.Half

tannergooding: This matches the existing resource string used for single/double.

danmosemsft: ah yes

PR #2639, opened 2019-01-25
Initial prototype for json serialization

steveharter:
Initial prototype of json deserialization to get feedback on API and usage.

This uses the 3.0 corefx preview bits to get the new JsonReader functionality instead of trying to merge into the existing corefxlab project for that (System.Text.JsonLab).

Per https://github.com/dotnet/corefx/issues/34372


src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Api.async.cs
```diff
@@ -0,0 +1,125 @@
...
+            Type returnType,
+            bool isFinalBlock,
+            byte[] buffer,
+            int bytesToRead,
+            JsonConverterSettings options,
+            ref JsonObjectState current,
+            ref List<JsonObjectState> previous,
+            ref int arrayIndex)
+        {
+            Utf8JsonReader reader = new Utf8JsonReader(new ReadOnlySequence<byte>(buffer, 0, bytesToRead), isFinalBlock, readerState);
```
ahsonkhan: I wouldn't create a ReadOnlySequence frm the input buffer. Instead pass it in as a span slice.

```C#
var reader = new Utf8JsonReader(buffer.AsSpan(0, bytesToRead), isFinalBlock, readerState);
```

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Api.async.cs
```diff
@@ -0,0 +1,125 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
```
ahsonkhan: nit: Name the file `JsonConverter.Async.Stream.cs`

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Api.sync.cs
```diff
@@ -0,0 +1,67 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
+    {
+        public static T FromJson<T>(string json, JsonConverterSettings options = null)
+        {
+            return (T)FromJson(json, typeof(T), options);
```
ahsonkhan: check for null string?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Api.sync.cs
```diff
@@ -0,0 +1,67 @@
...
+        }
+
+        public static object FromJson(string json, Type returnType, JsonConverterSettings options = null)
+        {
+            byte[] jsonBytes = Encoding.UTF8.GetBytes(json);
+            return FromJson((ReadOnlySpan<byte>)jsonBytes, returnType, options);
```
ahsonkhan: nit: `jsonBytes.AsSpan()` reads better than the cast, imo

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
```diff
@@ -0,0 +1,170 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
```
ahsonkhan: Name this file `JsonConverter.Helpers.cs` given it looks like it contains private static methods, or  `JsonConverter.Common.cs`

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Api.sync.cs
```diff
@@ -0,0 +1,67 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers;
+using System.Collections.Generic;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
```
ahsonkhan: Then this can be `JsonConverter.Sync.cs` and once we add PipeReader/Writer support, those can be `JsonConverter.Async.PipeReader.cs`.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
```diff
@@ -0,0 +1,170 @@
...
+            ref Utf8JsonReader reader,
+            JsonConverterSettings options,
+            Type returnType,
+            ref JsonObjectState current,
+            ref List<JsonObjectState> previous,
+            ref int arrayIndex)
+        {
+            while (reader.Read())
+            {
+                JsonTokenType tokenType = reader.TokenType;
```
ahsonkhan: nit: consider using switch?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
```diff
@@ -0,0 +1,170 @@
...
+                    {
+                        return true;
+                    }
+                }
+                else if (tokenType == JsonTokenType.PropertyName)
+                {
+                    Debug.Assert(current.ReturnValue != default);
+                    Debug.Assert(current.ClassInfo != default);
+
+                    ReadOnlySpan<byte> propertyName = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
```
ahsonkhan: If your `Utf8JsonReader` was known to be constructed from a byte[] or span (rather than `ReadOnlySequence`), we can get rid of the `reader.HasValueSequence` check.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
```diff
@@ -0,0 +1,170 @@
...
+            if (current.ClassInfo == null)
+            {
+                Debug.Assert(current.PropertyInfo == null);
+                JsonClassInfo returnClassInfo = options.GetOrAddClass(returnType);
+                current.ClassInfo = returnClassInfo;
+                current.PropertyInfo = returnClassInfo.GetPolicyProperty();
+            }
+        }
+
+        private static void GetPreviousState(ref List<JsonObjectState> previous, ref JsonObjectState state, int index)
```
ahsonkhan: Can we return the `JsonObjectState` rather than passing it by ref, esp given the method name is `Get...`?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
```diff
@@ -0,0 +1,250 @@
...
+using System.Linq;
+using System.Reflection;
+
+namespace System.Text.Json.Serialization
+{
+    public class JsonConverterSettings
+    {
+        private volatile JsonMemberBasedClassMaterializer _classMaterializerStrategy;
+        private JsonClassMaterializer _classMaterializer;
+        private int _defaultBufferSize = 16 * 1024;
```
ahsonkhan: Create a named constant for the initial default buffer size?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
```diff
@@ -0,0 +1,250 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+
+namespace System.Text.Json.Serialization
+{
+    public class JsonConverterSettings
+    {
+        private volatile JsonMemberBasedClassMaterializer _classMaterializerStrategy;
```
ahsonkhan: Why `volatile`?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
```diff
@@ -0,0 +1,250 @@
...
+                if (value < 0)
+                {
+                    throw new ArgumentException("todo");
+                }
+
+                _maxDepth = value;
+            }
+        }
+
+        //public static void AddAttribute(ICustomAttributeProvider type, Attribute attribute)
```
ahsonkhan: Remove commented out code?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/CamelCasingPolicy.cs
```diff
@@ -0,0 +1,31 @@
...
+    {
+        public static string FromJson(string value)
+        {
+            if (value.Length == 0 || char.IsUpper(value[0]))
+                return value;
+
+            if (value.Length == 1)
+                return value.ToUpperInvariant();
+
+            return char.ToUpperInvariant(value[0]) + value.Substring(1);
```
ahsonkhan: Can we use span/slicing to avoid unnecessary substring allocations?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/DateConverter.cs
```diff
@@ -0,0 +1,24 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers.Text;
+
+namespace System.Text.Json.Serialization.Policies
+{
+    internal class DateConverter : IUtf8ValueConverter<DateTime>
```
ahsonkhan: Would this still be needed after https://github.com/dotnet/corefx/issues/34690?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/EnumConverter.cs
```diff
@@ -0,0 +1,47 @@
...
+        {
+            string stringVal;
+            if (_treatAsString)
+            {
+                stringVal = value.ToString();
+            }
+            else
+            {
+                object objVal = Convert.ChangeType(value, Enum.GetUnderlyingType(value.GetType()));
+                stringVal = value.ToString();
```
ahsonkhan: @stephentoub, can `Enum` implement `ISpanFormattable`?

tests/System.Text.JsonLab.Serialization.Tests/SerializeAsyncTests.cs
```diff
@@ -0,0 +1,41 @@
...
+        [Fact]
+        public static async Task SerializePrimitivesAsync()
+        {
+            MemoryStream stream = new MemoryStream(Encoding.UTF8.GetBytes(@"1"));
+            JsonConverterSettings settings = new JsonConverterSettings
+            {
+                DefaultBufferSize = 1
+            };
+
+            System.Diagnostics.Debugger.Break();
```
ahsonkhan: remove debugging code

tests/System.Text.JsonLab.Serialization.Tests/SerializeTests.NonObjectTests.cs
```diff
@@ -0,0 +1,112 @@
...
+            List<List<int>> result = JsonConverter.FromJson<List<List<int>>>(Encoding.UTF8.GetBytes(@"[[1,2],[3,4]]"));
+
+            Assert.Equal(1, result[0][0]);
+            Assert.Equal(2, result[0][1]);
+            Assert.Equal(3, result[1][0]);
+            Assert.Equal(4, result[1][1]);
+        }
+
+        //[Fact]
+        //public static void SerializeHeterogeneousArray()
```
ahsonkhan: Why is this test commented out?

tests/System.Text.JsonLab.Serialization.Tests/System.Text.JsonLab.Serialization.Tests.csproj
```diff
@@ -0,0 +1,46 @@
...
+    <PackageIconUrl></PackageIconUrl>
+    <PackageProjectUrl></PackageProjectUrl>
+    <PackageLicenseUrl></PackageLicenseUrl>
+    <PackageRequireLicenseAcceptance>False</PackageRequireLicenseAcceptance>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="$(TestSdkVersion)" />
+    <PackageReference Include="xunit" Version="$(XunitVersion)" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="$(XunitVersion)" />
+    <PackageReference Include="Newtonsoft.Json" Version="10.0.2" />
```
ahsonkhan: Remove unused references to `Newtonsoft` and `Utf8Json`.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.HandleNumber.cs
```diff
@@ -0,0 +1,80 @@
...
+
+            // More common types are listed first
+
+            if (propType == typeof(int))
+            {
+                if (!reader.TryGetInt32Value(out int value))
+                {
+                    throw new FormatException();
+                }
+                return SetValueAsPrimitive(ref current, value);
```
ahsonkhan: Use the new `GetInt32` APIs that already throw FormatException.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.HandleNumber.cs
```diff
@@ -0,0 +1,80 @@
...
+            if (propType == typeof(float))
+            {
+                if (!reader.TryGetSingleValue(out float value))
+                {
+                    throw new FormatException();
+                }
+                return SetValueAsPrimitive(ref current, value);
+            }
+
+            //todo: add support for unsigned types (UInt, etc): https://github.com/dotnet/corefx/issues/33320
```
ahsonkhan: This issue is closed. We have the `GetUInt32` APIs now

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
```diff
@@ -0,0 +1,170 @@
...
+                    Debug.Assert(current.ReturnValue != default);
+                    Debug.Assert(current.ClassInfo != default);
+
+                    ReadOnlySpan<byte> propertyName = reader.HasValueSequence ? reader.ValueSequence.ToArray() : reader.ValueSpan;
+                    current.PropertyInfo = current.ClassInfo.GetProperty(current.ReturnValue.GetType(), propertyName, current.PropertyIndex);
+                    current.PropertyIndex++;
+                }
+                else if (tokenType == JsonTokenType.StartObject)
+                {
+                    HandleStartObject(options, returnType, ref current, ref previous, ref arrayIndex);
```
ahsonkhan: Why don't we need to call `CheckForNonObjectReturn` for this case? We do it for all the others.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/EnumConverter.cs
stephentoub: > can Enum implement ISpanFormattable?

It could with a little work, but how would that help here?  ISpanFormattable isn't public, and it's char-based.

Separately, is this code correct?  It's currently doing `stringValue = value.ToString()`, but should that instead be `objVal.ToString()`?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
```diff
@@ -0,0 +1,260 @@
...
+                }
+            }
+
+            if (propertyInfo != null)
+            {
+                string propertyName = jsonInfo.NameConverter == null ? propertyInfo.Name : jsonInfo.NameConverter.ToJson(propertyInfo.Name);
+                var bytes = (ReadOnlySpan<byte>)Encoding.Default.GetBytes(propertyName);
+                if (bytes.Length > 6)
+                {
+                    jsonInfo.Key = bytes.Slice(6).ToArray();
```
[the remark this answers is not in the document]
steveharter: The first 6 bytes are encoded into the key, so redundant

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonObjectState.cs
```diff
@@ -0,0 +1,163 @@
...
+
+        // Support System.Array and other types that don't implement IList
+        public List<object> TempEnumerableValues;
+
+        // Cached values used to determine if the current property value or current return value is enumerable.
+        private bool _isEnumerable;
+        private bool _isPropertyEnumerable;
+
+        // For performance, we order the properties by the first usage and this index helps find the right slot quicker.
+        public int PropertyIndex;
```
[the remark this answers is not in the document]
steveharter: Fields are faster than properties, especially with primitives. I've measured just raw sets\gets on x64, release build. I added properties for when there is some logic. However I can measure again to see if it's noticeable in real-world with serialization.

tests/System.Text.JsonLab.Serialization.Tests/SerializeAsyncTests.cs
steveharter: Thanks. Missed that

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
steveharter: Because its an Object :) ... The check is to initialize the "class" that has the metadata for the deseserialize and when returning primitives and arrays that isn't set up yet.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
```diff
@@ -0,0 +1,260 @@
...
+                    Type genericPropertyType = typeof(JsonPropertyInfo<>).MakeGenericType(propertyType);
+                    Type collectionElementType = GetElementType(propertyType);
+                    jsonInfo = (JsonPropertyInfo)Activator.CreateInstance(genericPropertyType, new object[] { classType, propertyType, propertyInfo, options, collectionElementType });
+                }
+            }
+
+            if (propertyInfo != null)
+            {
+                string propertyName = jsonInfo.NameConverter == null ? propertyInfo.Name : jsonInfo.NameConverter.ToJson(propertyInfo.Name);
+                var bytes = (ReadOnlySpan<byte>)Encoding.Default.GetBytes(propertyName);
```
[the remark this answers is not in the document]
steveharter: Done

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
```diff
@@ -0,0 +1,260 @@
...
+            }
+
+            info = null;
+            return false;
+        }
+
+        private static ulong GetKey(ReadOnlySpan<byte> propertyName)
+        {
+            int length = propertyName.Length;
+            ulong key = propertyName[0];
```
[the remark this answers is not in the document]
steveharter: Yes since based on a property name which has to be at least one character. I'll add an Assert

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
```diff
@@ -0,0 +1,260 @@
...
+                    }
+                }
+            }
+            key |= (ulong)propertyName.Length << 48;
+            return key;
+        }
+
+        public JsonPropertyInfo GetPolicyProperty()
+        {
+            return _property_refs[0]._info;
```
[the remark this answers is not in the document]
steveharter: As-is (based on callers) it can't. I'll add an Assert however.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
steveharter: I'll measure perf; switch will change IL generation from elseif to an "optimized" version that may or may not be faster since currently with elseif the ordering is from common to less common.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
steveharter: Hmm not sure if passing that info down is any better\quicker

ahsonkhan: What info would you have to pass down? If this is a general purpose work-horse which needs to work for any instance of `Utf8JsonReader`, **then keep the code as is**. What you have now is a more robust implementation, so I am fine with keeping it.

If on the other hand, we only every construct it from spans (or single-segment ReadOnlySequence), the value will always be contained in `ValueSpan` so checking and calling ToArray on `ValueSequence` is pointless. I was just pointing out that we could leverage the info about how `Utf8JsonReader` was constructed since we control it.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonObjectState.cs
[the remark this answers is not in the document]
ahsonkhan: Alright. The only reason why I asked is due to the coding guideline:

> Public fields should be used **sparingly** and should use PascalCasing with no prefix when used.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
steveharter: Not as-is. As an optimization, `previous` is null initially (for simple classes that don't have nested objects)

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
steveharter: Done

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
steveharter: It is created lazily and can be accessed by more than one thread

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,301 @@
...
+                        }
+                    }
+                }
+            }
+
+            return attr;
+        }
+    }
+
+    internal class JsonPropertyInfo<TValue> : JsonPropertyInfo
```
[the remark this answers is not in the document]
steveharter: Done

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/CamelCasingPolicy.cs
```diff
@@ -0,0 +1,31 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Text.Json.Serialization.Policies
+{
+    internal static class CamelCasingPolicy
+    {
+        public static string FromJson(string value)
+        {
+            if (value.Length == 0 || char.IsUpper(value[0]))
```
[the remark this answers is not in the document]
steveharter: Done

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/CamelCasingPolicy.cs
steveharter: Not in this case as I want to return a string

YohDeadfall: You are able to return a string, but without allocation of two additional strings which are concatenated to ptoduce the result. Use the `string.Create<TState>(Int32, TState, SpanAction<Char, TState>)`method instead and pass `value` as a state object. This approach allows you to remove the second condition from the method.

ahsonkhan: Current:
Allocations: 88 bytes
Runtime: 39 ns

You could use the SpanAction, which is faster (though it allocates more):
```C#
string result = string.Create(value.Length, first, (Span<char> chars, char ch) =>
{
    chars[0] = char.ToUpperInvariant(ch);
    value.AsSpan(1).CopyTo(chars.Slice(1));
});
```
Allocations: 120 bytes
Runtime: 31 ns

OR, if we really care, use unsafe code (which is much faster with no allocations):
```C#
ReadOnlySpan<char> span = value.AsSpan();
Span<char> writableSpan = MemoryMarshal.CreateSpan(ref MemoryMarshal.GetReference(span), span.Length);
writableSpan[0] = char.ToUpperInvariant(first);
return new string(writableSpan);
```
Allocations: 0 bytes
Runtime: 4 ns

If we want to only use safe code, we could leverage stackalloc span for small lengths to avoid the allocation:

```C#
public static string FromJsonNew(string value)
{
    if (value.Length == 0)
        return value;

    char first = value[0];
    if (char.IsUpper(first))
        return value;

    if (value.Length == 1)
        return value.ToUpperInvariant();

    if (value.Length < 256)
    {
        Span<char> span = stackalloc char[value.Length];
        span[0] = char.ToUpperInvariant(first);
        value.AsSpan(1).CopyTo(span.Slice(1));
        return new string(span);
    }
    else
    {
        return char.ToUpperInvariant(first) + value.Substring(1);
    }
}
```
Allocations: 32 bytes
Runtime: 33 ns

stephentoub: When targeting .NET Core 3, it can be:
```C#
char c = char.ToUpperInvariant(first);
return string.Concat(MemoryMarshal.CreateReadOnlySpan(ref c, 1), value.AsSpan(1));
```

stephentoub: > You could use the SpanAction, which is faster (though it allocates more):

It doesn't need to allocate beyond the returned string (other than the one-time allocation for the compiler-cached delegate):
```C#
return string.Create(value.Length, (ch, value), (chars, args) =>
{
    chars[0] = char.ToUpperInvariant(args.ch);
    args.value.AsSpan(1).CopyTo(chars.Slice(1));
});
```
You just need to make sure not to capture anything into the closure.

ahsonkhan: `string.Concat` isn't the best for perf here and it ends up using unsafe code, so if we are using `MemoryMarshal`, maybe we shouldn't use `string.Concat` and just use the span-based string ctor.

I am surprised the string ctor doesn't allocate at all.

stephentoub: > string.Concat isn't the best for perf here and it ends up using unsafe code

How so?  It should be the fastest of all the options.  It's also not doing anything unsafe; the only thing bordering on that is the MemoryMarshal.CreateReadOnlySpan, and the only thing there that needs to be accurate is the 1 passed in as the length.

> I am surprised the unsafe sample which uses the string ctor doesn't allocate at all.

It allocates the result string like all the others.  What other allocations were you expecting?

ahsonkhan: > Please don't do this: you're mutating the original string

Lol, you are right :)

Yes, we shouldn't do this. 

> It allocates the result string like all the others. What other allocations were you expecting?

The benchmark showed 0 allocations, because we are mutating the original string and the ctor ends up not allocating.

> How so? It should be the fastest of all the options.

Let me try to call the method in a loop, but the SpanAction-based impl took 19 ns, while string.Concat took 30 ns.

stephentoub: > but the SpanAction-based impl took 19 ns, while string.Concat took 30 ns.

Sounds like we should improve string.Concat then.  It's probably suffering from doing a Span.CopyTo on the 1-char first span.

stephentoub: > The benchmark showed 0 allocations, because we are mutating the original string and the ctor ends up not allocating.

The `return new string(writableSpan);` would still end up allocating.  You're saying it somehow wasn't?

ahsonkhan:
``` ini

BenchmarkDotNet=v0.11.3, OS=Windows 10.0.17763.253 (1809/October2018Update/Redstone5)
.NET Core SDK=3.0.100-preview-010174
  [Host]     : .NET Core 3.0.0-preview-27324-5 (CoreCLR 4.6.27322.0, CoreFX 4.7.19.7311), 64bit RyuJIT
  DefaultJob : .NET Core 3.0.0-preview-27324-5 (CoreCLR 4.6.27322.0, CoreFX 4.7.19.7311), 64bit RyuJIT


```
|       Method |     Mean |     Error |    StdDev | Ratio | RatioSD | Gen 0/1k Op | Gen 1/1k Op | Gen 2/1k Op | Allocated Memory/Op |
|------------- |---------:|----------:|----------:|------:|--------:|------------:|------------:|------------:|--------------------:|
|   SpanAction | 20.88 us | 0.2818 us | 0.2353 us |  0.67 |    0.02 |      9.3689 |           - |           - |            38.28 KB |
| StringConcat | 30.78 us | 0.6499 us | 0.7981 us |  1.00 |    0.00 |      9.3384 |           - |           - |            38.28 KB |

stephentoub: > Yes.

What is the implementation of UnsafeMutateString?  It's doing `return new string(writeableSpan)` and it's not allocating?  I don't believe it.

ahsonkhan: > I don't believe it.

> each subsequent call to the method returns the same (unchanged) string so it doesn't allocate
> We never end up calling the string ctor.

You were right not to :)

The results make sense now.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.cs
```diff
@@ -0,0 +1,170 @@
...
+using System.Buffers;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
+    {
+        private static readonly JsonConverterSettings s_default_options = new JsonConverterSettings();
```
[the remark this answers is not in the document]
steveharter: I'll use "settings" for now. Fixed some naming conventions

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Api.async.cs
```diff
@@ -0,0 +1,125 @@
...
+        {
+            return FromJsonAsync<T>(source, typeof(T), options, cancellationToken);
+        }
+
+        public static Task<object> FromJsonAsync(this Stream source, Type returnType, JsonConverterSettings options = null, CancellationToken cancellationToken = default)
+        {
+            return FromJsonAsync<object>(source, returnType, options, cancellationToken);
+        }
+
+        private static async Task<T> FromJsonAsync<T>(this Stream source, Type returnType, JsonConverterSettings options = null, CancellationToken cancellationToken = default)
```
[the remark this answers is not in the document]
steveharter: I didn't add property validation yet.... I'll add it

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/DateConverter.cs
steveharter: Perhaps if someone has a special encoding of date (there are many)

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/EnumConverter.cs
steveharter: I refactored that code... That was for serialization which wasn't added\tested yet (i.e. I added the method but couldn't test it yet)

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonObjectState.cs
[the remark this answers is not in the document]
steveharter: None of these are public to the outside (they are public on internal class)

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
```diff
@@ -0,0 +1,250 @@
...
+
+using System.Collections.Concurrent;
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Linq;
+using System.Reflection;
+
+namespace System.Text.Json.Serialization
+{
+    public class JsonConverterSettings
```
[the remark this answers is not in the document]
steveharter: Possible yes but I think there will be lots of accidental copy-by-value semantics and confusion. Plus it contains a cache and thus not a simple primitive type.

tests/System.Text.JsonLab.Serialization.Tests/SerializeTests.NonObjectTests.cs
steveharter: I removed it; case covered by SerializePrimitiveArrayFail

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Api.sync.cs
steveharter: Apparently there is an implicit cast operator from byte[] to ReadOnlySpan, so I can just remove that cast.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.HandleNumber.cs
steveharter: Done

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/CamelCasingPolicy.cs
steveharter: I used the AsSpan \ CopyTo approach

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.HandleNumber.cs
steveharter: Unfortunately I am using Preview1; I changed it in my private branch off of corefx and added extension methods in the corefxlab branch so I can share source.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Async.Stream.cs
```diff
@@ -0,0 +1,140 @@
...
+            }
+            finally
+            {
+                ArrayPool<byte>.Shared.Return(buffer, clearArray: true);
+            }
+
+            throw new InvalidOperationException("todo");
+        }
+
+
```
ahsonkhan: nit: extra new lines

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonEnumConverterAttribute.cs
```diff
@@ -0,0 +1,43 @@
...
+        }
+
+        public JsonEnumConverterAttribute()
+        {
+            PropertyType = typeof(Enum);
+        }
+
+        public override bool TryGetFromJson(ReadOnlySpan<byte> span, Type type, out object value)
+        {
+            System.Enum temp;
```
ahsonkhan: Why fully qualify the type name here?

ahsonkhan: nit: can we inline the out variable declaration?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
```diff
@@ -0,0 +1,89 @@
...
+
+            return FromJsonInternal(json, returnType, options);
+        }
+
+        private static object FromJsonInternal(string json, Type returnType, JsonConverterSettings options = null)
+        {
+            if (options == null)
+                options = s_DefaultSettings;
+
+            byte[] jsonBytes = s_utf8Encoding.GetBytes(json);
```
ahsonkhan: Can we instead use span-based transcoding APIs and write the UTF-8 bytes to a stackalloc span for json.Length <= 512 and borrow from the array pool for > 512, so we can avoid these intermediary allocations?

We can return the array back to the pool (and clear it) before returning the result.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
```diff
@@ -0,0 +1,273 @@
...
+                }
+                else if (propertyType == typeof(ushort))
+                {
+                    jsonInfo = new JsonPropertyInfo<ushort>(classType, propertyType, propertyInfo, options);
+                }
+                else if(propertyType == typeof(uint))
+                {
+                    jsonInfo = new JsonPropertyInfo<uint>(classType, propertyType, propertyInfo, options);
+                }
+                else if(propertyType == typeof(ulong))
```
ahsonkhan: nit: inconsistent spacing/formatting
`else if (propertyType == typeof(ulong))`

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
```diff
@@ -0,0 +1,273 @@
...
+
+        private static ulong GetKey(ReadOnlySpan<byte> propertyName)
+        {
+            Debug.Assert(propertyName.Length > 0);
+
+            int length = propertyName.Length;
+
+            // Embed the propertyName in the first 6 bytes of the key.
+            ulong key = propertyName[0];
+            if (length > 1)
```
ahsonkhan: Use `MemoryMarshal.Read` with fewer nested branches (20%+ faster)

```C#
private static ulong GetKey(ReadOnlySpan<byte> propertyName)
{
    Debug.Assert(propertyName.Length > 0);

    ulong key;
    int length = propertyName.Length;

    if (length > 3)
    {
        key = MemoryMarshal.Read<uint>(propertyName);
        if (length > 4)
        {
            key |= (ulong)propertyName[4] << 32;
        }
        if (length > 5)
        {
            key |= (ulong)propertyName[5] << 40;
        }
    }
    else if (length > 1)
    {
        key = MemoryMarshal.Read<ushort>(propertyName);
        if (length > 2)
        {
            key |= (ulong)propertyName[2] << 16;
        }
    }
    else
    {
        key = propertyName[0];
    }
           
    // Embed the propertyName length in the last two bytes.
    key |= (ulong)propertyName.Length << 48;
    return key;
}```

File: src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs [hunk 37]
```diff
@@ -0,0 +1,273 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Reflection;
+
+namespace System.Text.Json.Serialization
+{
+    internal class JsonClassInfo
+    {
+        // The length of the property name embedded in the key (in bytes).
+        private const int PropertyNameKeyLength = 6;
```
ahsonkhan: What's the rationale behind choosing 6 for the key?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/PropertyRef.cs
```diff
@@ -0,0 +1,18 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Text.Json.Serialization
+{
+    internal struct PropertyRef
```
YohDeadfall: Should it be `readonly struct`?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonEnumConverterAttribute.cs
steveharter: I cleaned these up - remnants from a previous iteration with generic methods.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
steveharter: I'll add a perf todo for now. How important do you think string-based scenarios are?

I'll be sure to add some API doc that the UTF8 approach is preferred for performance - meaning be sure to specify UTF8 content encoding.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
steveharter: I was hoping there was a better way to copy six bytes[] into a ulong. I'll give this a try but hoping there's something better.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/PropertyRef.cs
steveharter: Yes. Thanks. The next commit will include readonly.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
steveharter: The underlying propertyname->info "hashtable" is a List<ulong> and 6 bytes (the first 6 characters of the property name) are reserved for the key and 2 bytes for the length. This allows a high-performance iteration since just ulongs are compared for the key. Once a hit is made based on key, the remaining bytes are searched if >6. Once a match is found then a secondary array is accessed with the same index that returns the cached "info" structure for the property.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
```diff
@@ -0,0 +1,235 @@
...
+            get
+            {
+                return _defaultBufferSize;
+            }
+            set
+            {
+                if (value < 1)
+                {
+                    throw new ArgumentException("todo");
+                }
```
YohDeadfall: Which style do you use here and in some other places? I see that somewhere you have brackets around one line statements, but not everywhere. Maybe remove them where needed and keep code cleaner?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
```diff
@@ -0,0 +1,235 @@
...
+                        _classMaterializerStrategy = new JsonReflectionEmitMaterializer();
+                    }
+                    else if (ClassMaterializer == JsonClassMaterializer.Reflection)
+                    {
+                        _classMaterializerStrategy = new JsonReflectionMaterializer();
+                    }
+                    else
+                    {
+                        throw new InvalidOperationException("todo");
+                    }
```
YohDeadfall: This could be converted to `switch` since it's small enough.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+        public Type ElementType { get; private set; }
+
+        public abstract Type TempListType { get; }
+
+        public abstract object GetValueAsObject(object obj);
+        public abstract void SetValueAsObject(object obj, object value);
+
+        public abstract PropertyValueConverterAttribute GetValueConverter();
+
+        public virtual void GetPolicies()
```
YohDeadfall: It's a bad name for the method which returns nothing, but binds fields.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+            {
+                // Then class type
+                attr = Options.GetAttributes<TAttribute>(ClassType, inherit: true).FirstOrDefault();
+
+                if (attr == null)
+                {
+                    // Then declaring assembly
+                    attr = Options.GetAttributes<TAttribute>(ClassType.Assembly).FirstOrDefault();
+                }
+            }
```
YohDeadfall: This can be simplified to:

```csharp
TAttribute attr = PropertyInfo != null
    ? Options.GetAttributes<TAttribute>(PropertyInfo).FirstOrDefault()
    : null;
return attr
    ?? Options.GetAttributes<TAttribute>(ClassType, inherit: true).FirstOrDefault();
    ?? Options.GetAttributes<TAttribute>(ClassType.Assembly).FirstOrDefault();
```

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+
+            if (attr == null)
+            {
+                value = default;
+            }
+            else
+            {
+                value = selector(attr);
+            }
```
YohDeadfall: Same here.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+                    attr = Options.GetAttributes<JsonPropertyAttribute>(ClassType.Assembly).FirstOrDefault();
+
+                    if (attr == null)
+                    {
+                        value = default;
+                    }
+                    else
+                    {
+                        value = selector(attr);
+                    }
```
YohDeadfall: And here.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+        protected EnumerableConverterAttribute GetEnumerableConverter()
+        {
+            Debug.Assert(ElementType != null);
+
+            Type enumerableType;
+            if (PropertyType.IsGenericType)
+            {
+                enumerableType = PropertyType.GetGenericTypeDefinition();
+            }
+            else if(PropertyType.IsArray)
```
YohDeadfall: nit: Lost whitespace after `if`.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+                    // Then declaring assembly
+                    attr = Options.GetAttributes<EnumerableConverterAttribute>(ClassType.Assembly).Where(a => a.EnumerableType == enumerableType).FirstOrDefault();
+
+                    if (attr == null)
+                    {
+                        // Then default
+                        if (enumerableType == typeof(Array))
+                        {
+                            attr = s_arrayConverterAttibute;
+                        }
```
YohDeadfall: Oh, how many things could be simplified.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+        }
+
+        public override PropertyValueConverterAttribute GetValueConverter()
+        {
+            return ValueConverter;
+        }
+
+        public override void GetPolicies()
+        {
+            // ValueConverter
```
YohDeadfall: What is this comment for?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonPropertyInfo.cs
```diff
@@ -0,0 +1,300 @@
...
+                    {
+                        // Then default
+                        if (propertyType == typeof(DateTime))
+                        {
+                            attr = s_dateConverterAttibute;
+                        }
+                        else if (propertyType == typeof(Enum))
+                        {
+                            attr = s_enumConverterAttibute;
+                        }
```
YohDeadfall: This code could be simplified too. If you would like I can submit a PR on the weekend to rewrite such things.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonReflectionEmitMaterializer.cs
```diff
@@ -0,0 +1,102 @@
...
+
+            ILGenerator generator = dynamicMethod?.GetILGenerator();
+            if (generator == null)
+                throw new InvalidOperationException();
+
+            generator.Emit(OpCodes.Newobj, realMethod);
+            generator.Emit(OpCodes.Ret);
+
+            var result = (JsonClassInfo.ConstructorDelegate)dynamicMethod.CreateDelegate(typeof(JsonClassInfo.ConstructorDelegate));
+            return result;
```
YohDeadfall: Any reason to store `result` in a variable?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonReflectionEmitMaterializer.cs
```diff
@@ -0,0 +1,102 @@
...
+            ILGenerator generator = dynamicMethod?.GetILGenerator();
+            if (generator == null)
+                throw new InvalidOperationException();
+
+            generator.Emit(OpCodes.Ldarg_0);
+            generator.EmitCall(OpCodes.Callvirt, realMethod, null);
+            generator.Emit(OpCodes.Ret);
+
+            var result = (JsonPropertyInfo<TValue>.GetterDelegate)dynamicMethod.CreateDelegate(typeof(JsonPropertyInfo<TValue>.GetterDelegate));
+            return result;
```
YohDeadfall: Same here.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonReflectionEmitMaterializer.cs
```diff
@@ -0,0 +1,102 @@
...
+            if (propertyInfo.PropertyType.IsEnum)
+            {
+                generator.Emit(OpCodes.Unbox_Any, propertyInfo.PropertyType);
+            }
+
+            generator.EmitCall(OpCodes.Callvirt, realMethod, null);
+            generator.Emit(OpCodes.Ret);
+
+            var result = (JsonPropertyInfo<TValue>.SetterDelegate)dynamicMethod.CreateDelegate(typeof(JsonPropertyInfo<TValue>.SetterDelegate));
+            return result;
```
YohDeadfall: Same here.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/ArrayConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+namespace System.Text.Json.Serialization.Policies
+{
+    internal static class ArrayConverter
+    {
+        public static IEnumerable CreateFromList(Type elementType, IList sourceList)
+        {
+            Array array;
+            Array probe;
+
+            if (sourceList.Count > 0 && (probe = sourceList[0] as Array) != null)
```
YohDeadfall: You don't need such a complex expression here. Use pattern matching.

```csharp
if (sourceList.Count > 0 && sourceList[0] is Array probe)
```

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/ArrayConverter.cs
```diff
@@ -0,0 +1,36 @@
...
+            {
+                array = Array.CreateInstance(probe.GetType(), sourceList.Count);
+
+                int i = 0;
+                foreach (Array childArray in sourceList.OfType<Array>())
```
YohDeadfall: Is it performance critical?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/DateConverter.cs
```diff
@@ -0,0 +1,25 @@
...
+using System.Buffers.Text;
+
+namespace System.Text.Json.Serialization.Policies
+{
+    internal static class DateConverter
+    {
+        public static bool TryGetFromJson(ReadOnlySpan<byte> span, Type type, out object value)
+        {
+            DateTime temp;
+            bool success = Utf8Parser.TryParse(span, out temp, out int bytesConsumed, 'O') && span.Length == bytesConsumed;
```
YohDeadfall: Declare an `out` variable instead.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/TempExtensionMethods.cs
```diff
@@ -0,0 +1,81 @@
...
+    // Temporary extension methods until we get the API additions from Preview2
+    internal static class TempExtensionMethods
+    {
+        public static int GetInt32(this Utf8JsonReader reader)
+        {
+            if (!reader.TryGetInt32Value(out int value))
+            {
+                throw new FormatException();
+            }
+            return value;
```
YohDeadfall: Invert the condition here and below to remove the negation operator.

tests/System.Text.JsonLab.Serialization.Tests/NameAttribute.cs
```diff
@@ -0,0 +1,122 @@
...
+        }
+
+        [Fact]
+        public static void CamelCaseAttributeDesignTime()
+        {
+            SimpleTestClassWithCamelCase obj  =JsonConverter.FromJson<SimpleTestClassWithCamelCase>(SimpleTestClassWithCamelCase.s_data);
+            Assert.Equal(obj.MyInt16, 1);
+        }
+
+
```
YohDeadfall: nit: An extra new line.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
ahsonkhan: Ah gotcha. Makes sense. Thanks for the explanation.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
ahsonkhan: The problem is, our `MemoryMarshal.Read` APIs expect the length to be > sizeof(T), so we can't pass a smaller than 8 byte span to it and expect the result to work.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
ahsonkhan: Since usability burden on the serializer/deserializer is higher than the jsonreader/writer (and used a lot more), I would imagine people to use what they are familiar with (i.e. strings). If we can make it faster for them as well, that would be awesome.

PR #2641, opened 2019-01-31
Update to latest dotnet cli and multi-target netstandard2.0 with TODOs

ahsonkhan:
Some clean-up and TODOs which throw NotImplementedException for netstandard2.0.


src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/EnumConverter.cs
```diff
@@ -9,12 +9,16 @@ internal class EnumConverter
         //todo: support [Flags]?
         public static bool TryGetFromJson(ReadOnlySpan<byte> span, Type type, out Enum value)
         {
-            string enumString = JsonConverter.s_utf8Encoding.GetString(span);
+            string enumString = TempExtensionMethods.TranscodeHelper(span);
+#if BUILDING_INBOX_LIBRARY
```
YohDeadfall: Move the variable into the conditional compilation block. Otherwise, it's unused.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/EnumConverter.cs
```diff
@@ -9,12 +9,16 @@ internal class EnumConverter
...
             if (Enum.TryParse(type, enumString, out object objValue))
             {
                 value = (Enum)objValue;
                 return true;
             }
+#else
+            throw new NotImplementedException("TODO: EnumConverter is not yet supported on .NET Standard 2.0.");
+#endif
 
             throw new InvalidOperationException($"todo:could not convert value to string-based Enum {type}");
```
YohDeadfall: And this one too. Probably it would be better to revert the condition to have `NotImplementedException` at the start.

BrennanConroy:
👀

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverterSettings.cs
```diff
@@ -51,7 +51,7 @@ internal JsonClassInfo GetOrAddClass(Type classType)
             return result;
         }
 
-        public JsonReaderOptions ReaderOptions { get; set; }
+        internal JsonReaderOptions ReaderOptions { get; set; }
```
ahsonkhan: I just marked it internal for now since this is consuming the source package on netstandard2.0, where ideally it would be in the package it self. We should revert this at that point and make it public again.

PR #2644, opened 2019-02-06
Add System.Text.Json serialization prototype support

steveharter:
Another iteration for early users to experiment with. Adds support for the non-async version of the ToJson\Serialization methods as specified in https://github.com/dotnet/corefx/issues/34372. Previously only the FromJson\Deserialize methods were available.

In addition, support for Nullable<T> and null object references was added.

The next release will add async PipeReader and Stream.

Disclaimer: this is a work-in-progress commit; there is pending performance and API validation\review work.

_Implementation note: the code is structured to support async by manually holding the call stack and object graph state so we can call flush on an async Stream or PipeReader. This is designed so we can flush at every property value if the buffer is small enough. This makes the code somewhat convoluted compared to a simple call-stack synchronous approach. Pending research, I may end up splitting the code base in this area for sync vs. async in order to reach higher performance numbers for synchronous. Also we should discuss whether per-property granularity for async support is overkill, and perhaps a per-object is good enough, which would help with code readability as well._


ahsonkhan:
> Disclaimer: this is a work-in-progress commit; there is pending performance and API validation\review work.

Should I hold off on reviewing it for now?

steveharter:
@ahsonkhan I was more referring to the surface area and functionality - API and perf. However, you could hold off since I have a least one more large refactoring to add more extensibility while avoid boxing, and perhaps one more after that if we decide to fork the implementation for async vs sync.

At a minimum please take a look at ToJsonInternal() using your prototype implementation of IBufferWriter.

About half of the code you reviewed last time is the same, although some of it moved.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonClassInfo.cs
```diff
@@ -193,12 +134,39 @@ public JsonPropertyInfo GetProperty(Type classType, ReadOnlySpan<byte> propertyN
...
+        }
+
         private static bool TryIsPropertyRefEqual(List<PropertyRef> list, ReadOnlySpan<byte> propertyName, ulong key, int index, out JsonPropertyInfo info)
         {
             if (key == list[index].Key)
             {
                 if (propertyName.Length <= PropertyNameKeyLength ||
-                    propertyName.Slice(PropertyNameKeyLength).SequenceEqual((ReadOnlySpan<byte>)list[index].Info.Key))
+                    // todo: is it really any faster to do slices here to avoid comparing the first 6 bytes?
+                    propertyName.Slice(PropertyNameKeyLength).SequenceEqual(((ReadOnlySpan<byte>)list[index].Info.Name).Slice(PropertyNameKeyLength)))
```
ahsonkhan: Slicing spans is generally quite cheap, but something we should measure, especially taking into account common property name lengths.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Async.Stream.cs
```diff
@@ -15,42 +15,50 @@ public static partial class JsonConverter
...
+
+            var readerState = new JsonReaderState(settings.MaxDepth, settings.ReaderOptions);
+            List<FromJsonObjectState> previous = null;
             int arrayIndex = 0;
 
             int bytesRemaining = 0;
             int bytesRead;
 
-            byte[] buffer = ArrayPool<byte>.Shared.Rent(options.DefaultBufferSize);
+            // todo: switch to IBufferWriter implementation.
```
ahsonkhan: Which implementation?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.FromJson.Null.cs
```diff
@@ -0,0 +1,45 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ahsonkhan: Given the method size, I am not sure if we should mark it with AI. How many calls to it do we have?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.FromJson.cs
```diff
@@ -0,0 +1,121 @@
...
+            JsonConverterSettings settings,
+            Type returnType,
+            ref FromJsonObjectState current,
+            ref List<FromJsonObjectState> previous,
+            ref int arrayIndex)
+        {
+            while (reader.Read())
+            {
+                JsonTokenType tokenType = reader.TokenType;
+                if (tokenType == JsonTokenType.String || tokenType == JsonTokenType.Number || tokenType == JsonTokenType.True || tokenType == JsonTokenType.False)
```
ahsonkhan: We can make this faster by doing a single range comparison between `JsonTokenType.String` and `JsonTokenType.False`

```C#
tokenType >= JsonTokenType.String && tokenType <= JsonTokenType.False
```

along with a trick like:
```C#
public static bool IsInRangeInclusive(uint value, uint lowerBound, uint upperBound)
            => (value - lowerBound) <= (upperBound - lowerBound);
```

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Helpers.cs
```diff
@@ -0,0 +1,50 @@
...
+
+using System.Collections.Generic;
+using System.Diagnostics;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
+    {
+        private static void GetPreviousState<T>(ref List<T> previous, ref T state, int index) where T:new()
```
ahsonkhan: nit: spacing

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
```diff
@@ -3,88 +3,181 @@
...
+            if (settings == null)
+                settings = s_DefaultSettings;
 
             // todo: use an array pool here for smaller requests to avoid the alloc. Also doc the API that UTF8 is preferred for perf. 
             byte[] jsonBytes = s_utf8Encoding.GetBytes(json);
-            JsonReaderState state = new JsonReaderState(options.DefaultBufferSize, options.ReaderOptions);
-            Utf8JsonReader reader = new Utf8JsonReader(jsonBytes, true, state);
-            return FromJsonInternal(reader, returnType, options);
+            var state = new JsonReaderState(settings.DefaultBufferSize, settings.ReaderOptions);
+            var reader = new Utf8JsonReader(jsonBytes, true, state);
```
ahsonkhan: nit: use named parameter for `isFinalBlock`

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
```diff
@@ -3,88 +3,181 @@
...
         }
+
+        public static string ToJsonString(object value, JsonConverterSettings settings = null)
+        {
+            if (value == null)
+                throw new ArgumentNullException(nameof(value));
+
+            Span<byte> jsonBytes = ToJsonInternal(value, settings);
+#if BUILDING_INBOX_LIBRARY
+            string stringJson = s_utf8Encoding.GetString(jsonBytes);
```
ahsonkhan: This ifdef'd code is used quite often. Move to a static helper method to avoid duplication?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
```diff
@@ -3,88 +3,181 @@
...
+
+        public static Span<byte> ToJson(object value, JsonConverterSettings settings = null)
+        {
+            if (value == null)
+                throw new ArgumentNullException(nameof(value));
+
+            return ToJsonInternal(value, settings);
+        }
+
+        private static Span<byte> ToJsonInternal(object value, JsonConverterSettings settings = null)
```
ahsonkhan: Remove the optional argument in private method?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
```diff
@@ -3,88 +3,181 @@
...
+
+                ToJson(ref writer, settings, ref current, ref previous, ref arrayIndex);
+
+                int byteCount;
+                checked
+                {
+                    byteCount = (int)writer.BytesWritten;
+                }
+
+                Span<byte> json = output.GetSpan().Slice(0, byteCount);
```
ahsonkhan: `GetSpan` gives you the span to write more data into, not what has already been written.

What you should use is `output.Formatted` instead.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
```diff
@@ -3,88 +3,181 @@
...
+            byte[] result;
+
+            using (var output = new ArrayBufferWriter(settings.DefaultBufferSize))
+            {
+                var writer = new Utf8JsonWriter(output, state);
+
+                ToJson(ref writer, settings, ref current, ref previous, ref arrayIndex);
+
+                int byteCount;
+                checked
```
ahsonkhan: Is this local and checked cast to int necessary?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.ToJson.Enumerable.cs
```diff
@@ -0,0 +1,77 @@
...
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
+    {
+        [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
ahsonkhan: This method should definitely not be marked with AI. I doubt it helps.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.ToJson.cs
```diff
@@ -0,0 +1,56 @@
...
+    {
+        private static bool ToJson(
+            ref Utf8JsonWriter writer,
+            JsonConverterSettings settings,
+            ref ToJsonObjectState current,
+            ref List<ToJsonObjectState> previous,
+            ref int arrayIndex)
+        {
+            bool writeMore = true;
+            while (writeMore)
```
ahsonkhan: Can this instead be a do/while with `(while writer.CurrentDepth > 0)`?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.ToJson.cs
```diff
@@ -0,0 +1,56 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace System.Text.Json.Serialization
+{
+    public static partial class JsonConverter
+    {
+        private static bool ToJson(
```
ahsonkhan: What does returning true/false here mean?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonEnumConverterAttribute.cs
```diff
@@ -17,26 +18,91 @@ public sealed class JsonEnumConverterAttribute : PropertyValueConverterAttribute
...
+        internal override object GetFromJson(ref Utf8JsonReader reader, Type propertyType)
+#endif
         {
-            bool success = EnumConverter.TryGetFromJson(span, type, out Enum temp);
-            value = temp;
-            return success;
+            if (TreatAsString)
+            {
+#if !BUILDING_INBOX_LIBRARY 
+            throw new NotImplementedException("TODO: EnumConverter is not yet supported on .NET Standard 2.0.");
```
ahsonkhan: nit: spacing

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/CamelCasingPolicy.cs
```diff
@@ -25,23 +25,7 @@ public static string FromJson(string value)
...
-                    {
-                        return new string(ptr, 0, span.Length);
-                    }
-                }
-            }
-            else
-            {
-                return char.ToUpperInvariant(firstChar) + value.Substring(1);
-            }
+            return char.ToUpperInvariant(value[0]) + value.Substring(1);
```
ahsonkhan: use the cached `firstChar` value

Also, not worth using stackalloc to avoid extra allocs in the common case?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/PropertyValueConverterAttribute.cs
```diff
@@ -9,8 +9,15 @@ public abstract class PropertyValueConverterAttribute : Attribute
...
-        public abstract bool TryGetFromJson(ReadOnlySpan<byte> span, Type type, out object value);
+        // todo: these must to be boxed to object (performance); we need to refactor to add converter interfaces with <T>
 
-        public abstract bool TrySetToJson(object value, out Span<byte> span);
+#if BUILDING_INBOX_LIBRARY
+        public abstract object GetFromJson(ref Utf8JsonReader reader, Type propertyType);
+        public abstract void SetToJson(ref Utf8JsonWriter writer, ReadOnlySpan<byte> name, object value);
+#else
+        //todo: ns20
+        internal abstract object GetFromJson(ref Utf8JsonReader reader, Type propertyType);
```
ahsonkhan: Why are these ifdef'd for ns2.0?

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/TempExtensionMethods.cs
```diff
@@ -1,43 +0,0 @@
...
-// See the LICENSE file in the project root for more information.
-
-namespace System.Text.Json.Serialization
-{
-    // Temporary extension methods until we move to corefx
-    internal static class TempExtensionMethods
-    {
-        // TODO: Replace with existing static helper in corefx
-        // System/Text/Json/Reader/JsonReaderHelper.Unescaping.cs#L42
-        public static string TranscodeHelper(ReadOnlySpan<byte> utf8Unescaped)
```
ahsonkhan: Why remove this?

tests/System.Text.JsonLab.Serialization.Tests/NameAttribute.FromJsonTests.cs
```diff
@@ -27,7 +27,7 @@ public static void CamelCaseAttributeDesignTimeFail()
         [Fact]
         public static void CamelCaseAttributeDesignTime()
         {
-            SimpleTestClassWithCamelCase obj = JsonConverter.FromJson<SimpleTestClassWithCamelCase>(SimpleTestClassWithCamelCase.s_data);
+            SimpleTestClassWithCamelCase obj  =JsonConverter.FromJson<SimpleTestClassWithCamelCase>(SimpleTestClassWithCamelCase.s_data);
```
ahsonkhan: nit: spacing

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/ArrayBufferWriter.cs
```diff
@@ -0,0 +1,87 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Buffers;
+using System.Runtime.CompilerServices;
+
+namespace System.Text.Json.Serialization
+{
+    // Note: this is an internal class that will be replaced with a shared version.
+    internal class ArrayBufferWriter : IBufferWriter<byte>, IDisposable
```
ahsonkhan: Maybe use the implementation we are actually shipping and get rid of ResizableArray: https://github.com/dotnet/corefx/blob/4e715d0ba37c3775e3f10ac4741264f556b84684/src/System.Buffers/src/System/Buffers/ArrayBufferWriter.cs

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Converters/JsonPropertyInfoChar.cs
```diff
@@ -0,0 +1,30 @@
...
+        { }
+
+        public char FromJson(ref Utf8JsonReader reader)
+        {
+            return reader.GetString()[0];
+        }
+
+        public void ToJson(ref Utf8JsonWriter writer, char value)
+        {
+            writer.WriteStringValue(value.ToString());
```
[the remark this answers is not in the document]
davidfowl: You don't have to stackalloc, there are 2 options:
- `MemoryMarshel.CreateSpan(ref char, 1)` (netcoreapp3.0 only)
- `new Span<char>(&c, 1)` (requires unsafe code)

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/ClassType.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace System.Text.Json.Serialization
+{
+    /// <summary>
+    /// Determines how a given class is treated when it is (de)serialized.
+    /// </summary>
+    internal enum ClassType
```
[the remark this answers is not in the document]
steveharter: Do you know if we have a policy or measurements for this? For raw CPU performance a four-byte int will beat a byte AFAIK due to register size and memory alignment.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Converters/JsonPropertyInfoBooleanNullable.cs
```diff
@@ -0,0 +1,30 @@
...
+{
+    internal class JsonPropertyInfoBooleanNullable : JsonPropertyInfo<bool?>, IJsonConverterInternal<bool?>
+    {
+        public JsonPropertyInfoBooleanNullable(Type classType, Type propertyType, PropertyInfo propertyInfo, JsonConverterSettings settings) :
+            base(classType, propertyType, propertyInfo, settings)
+        { }
+
+        public bool? FromJson(ref Utf8JsonReader reader)
+        {
+            return reader.GetBoolean();
```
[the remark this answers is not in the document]
steveharter: During deserialization a True or False token type has already been found so this will never be null. It's a nullable to avoid boxing.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Converters/JsonPropertyInfoSingle.Nullable.cs
```diff
@@ -0,0 +1,33 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Reflection;
+
+namespace System.Text.Json.Serialization.Converters
+{
+    internal class JsonPropertyInfoSingleNullable : JsonPropertyInfo<float?>, IJsonConverterInternal<float?>
```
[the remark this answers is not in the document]
steveharter: Oops. Copy-paste failed me. Or I failed it. Thanks

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Converters/JsonPropertyInfoChar.cs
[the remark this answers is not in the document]
steveharter: Thanks. I implemented the MemoryMarshal.Create span approach for netcoreapp3.0; kept the original for netstandard.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Converters/JsonPropertyInfoSingle.cs
```diff
@@ -0,0 +1,33 @@
...
+{
+    internal class JsonPropertyInfoSingle : JsonPropertyInfo<float>, IJsonConverterInternal<float>
+    {
+        public JsonPropertyInfoSingle(Type classType, Type propertyType, PropertyInfo propertyInfo, JsonConverterSettings settings) :
+            base(classType, propertyType, propertyInfo, settings)
+        { }
+
+        public float FromJson(ref Utf8JsonReader reader)
+        {
+            checked
```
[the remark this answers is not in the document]
steveharter: It doesn't. Removed it. Copy-paste issue again. Thanks

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Converters/JsonPropertyInfoSingle.Nullable.cs
```diff
@@ -0,0 +1,33 @@
...
+    {
+        public JsonPropertyInfoSingleNullable(Type classType, Type propertyType, PropertyInfo propertyInfo, JsonConverterSettings settings) :
+            base(classType, propertyType, propertyInfo, settings)
+        { }
+
+        public float? FromJson(ref Utf8JsonReader reader)
+        {
+            checked
+            {
+                return reader.GetSingle();
```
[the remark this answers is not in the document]
steveharter: All of the token checks for null already happened earlier in a common location. If there is a null token then either nothing happens (the default behavior) or there is a setting (DeserializeNullValues) to force the property to be set to null.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Async.Stream.cs
steveharter: The prototype implementation of IBufferWriter that you added - ArrayBufferWriter. If it helps remove the allocation and pool logic here.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.FromJson.Null.cs
steveharter: There is only one call, but it's from the "main" loop so called for every null token.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.FromJson.cs
steveharter: Sounds good although the helper method probably wouldn't be any faster, so just doing the range with a Debug.Assert for readability

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
steveharter: It is used twice here. I re-added the helper method but renamed the helper class to JsonReaderHelper to match corefx.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.ToJson.Enumerable.cs
steveharter: It is only called once, but often, so a candidate for inlining.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.ToJson.cs
steveharter: Currently not used until we add async. True=continue. False=done. The callee then if True was returned will write to the steam and async flush

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/CamelCasingPolicy.cs
steveharter: Yes I will change to use firstChar. However I am going to hold off on micro-optimizations for netstandard like unsafe stackalloc for now at least here because the camel-casing name is cached.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/PropertyValueConverterAttribute.cs
steveharter: I'll change the comment... They have to be internal for now in corefxlab due to a compile error regarding increasing visibility

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/TempExtensionMethods.cs
steveharter: See previous - I added it aback but renamed the class to match corefx.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/ArrayBufferWriter.cs
steveharter: OK thanks I'll use the current code.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
steveharter: Removed when changed to new API

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
steveharter: Is this called `WrittenMemory` in the newer version of ArrayBufferWriter?

steveharter:
Addressed comments; merging.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/ClassType.cs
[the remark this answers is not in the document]
ahsonkhan: It depends. For structs (like `Ut8fJsonReader`), I used byte backed enums (e.g. `TokenType`) to reduce struct size.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.FromJson.cs
ahsonkhan: The helper method converts 2 comparisons and an and to a single comparison (i.e. fewer branches).

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
ahsonkhan: Yes.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.ToJson.Enumerable.cs
ahsonkhan: I doubt the JIT can inline it though, even if you ask. cc @jkotas, @AndyAyersMS

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/Policies/PropertyValueConverterAttribute.cs
ahsonkhan: Oh right, good point.

src/System.Text.JsonLab.Serialization/System/Text/Json/Serialization/JsonConverter.Sync.cs
```diff
@@ -153,19 +138,14 @@ private static Span<byte> ToJsonInternal(object value, JsonConverterSettings set
...
 
-                int byteCount;
-                checked
-                {
-                    byteCount = (int)writer.BytesWritten;
-                }
-
-                Span<byte> json = output.GetSpan().Slice(0, byteCount);
+                writer.Flush(isFinalBlock: true);
+                ReadOnlyMemory<byte> json = output.WrittenMemory;
```
ahsonkhan: Can we avoid calling ToArray on the written memory some how? For the string case, we should just transcode and return the string to avoid the extra allocation.

Span is trickier. As an aside, one thing I noticed, shouldn't we return a `ReadOnlySpan<byte>` instead?

PR #2656, opened 2019-04-09
Data frame binary operations

pgovind:
This is work towards [#26854](https://github.com/dotnet/corefx/issues/26845) to add a DataFrame type to .NET Core

This patch implements the basic infrastructure, indexing and binary operations on a dataframe. I thought this was a good spot to get an initial review. More patches will follow on top of this to support computations, aggregations, joins, merges, group by etc

@eerhardt @stephentoub @ericstj


YohDeadfall:
Could you reference an issue or describe the goal?

/cc @roji

src/Microsoft.Data/BaseDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+        {
+            Length = length;
+            Name = name;
+        }
+        public long Length { get; protected set; }
+        public long NullCount { get; protected set; }
+        public string Name;
+
+        public virtual object this[long rowIndex] { get { throw new NotImplementedException(); } set { throw new NotImplementedException(); } }
+        public virtual object this[long startIndex, int length] { get { throw new NotImplementedException(); } }
```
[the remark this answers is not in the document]
pgovind: They are overridden in DataFrameColumn<T> at the moment?. Do you mean they should occur only in DataFrameColumn<T> and not in the base class? I'm not completely convinced about that yet. At the moment, the DataFrame class has a DataFrameTable which has an IList<BaseDataFrameColumn> called columns i.e. the DataFrameTable does not know the real type of the columns it holds => therefore, the approach at the moment is to add APIs on BaseDataFrameColumn and override them in the derived columns. It is possible that this will change down the line as I add more features, but for the moment I think this looks reasonable. Thoughts?

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
```diff
@@ -0,0 +1,69 @@
...
+            if (values.Count != NumColumns)
+            {
+                throw new ArgumentException($"values.Count {values.Count} must match the number of columns in the table", nameof(values));
+            }
+<# } #>
+            var newDataFrame = new DataFrame();
+            for (int ii = 0; ii < NumColumns; ii++)
+            {
+                DataFrameColumn<T> column = _table.Column(ii) as DataFrameColumn<T>;
+                if (column != null)
```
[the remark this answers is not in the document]
pgovind: I'm not opposed to the style but I do have some questions here for the larger crowd:

If I had a `DataFrameColumn<int>` and called `DataFrameColumn<int>.Add(5.0)`,
The lines 
`DataFrameColumn<T> column = _table.Column(ii) as DataFrameColumn<T>; 
                if (column != null)`

mean that nothing will happen since the dynamic cast will fail. I'm leaning towards checking the type of the resulting operation and creating a new column with that type?
For ex: DataFrameColumn<int>.Add(5.0) will result in a new DataFrameColumn<double>, similar to what `1+5.0` does in C#.

For cases such as DataFrameColumn<int> + string, we just don't perform the operation instead of throwing?
Thoughts?

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
```diff
@@ -0,0 +1,69 @@
...
+
+namespace Microsoft.Data
+{
+    public partial class DataFrame
+    {
+        #region Binary Operations
+
+<# foreach (MethodConfiguration method in methodConfiguration) { #>
+<# if (method.MethodType == MethodType.BinaryScalar || method.MethodType == MethodType.ComparisonScalar) {#>
+        public DataFrame <#=method.MethodName#><T>(T value)
```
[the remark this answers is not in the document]
pgovind: Right, I don't think T4TT engine does a great job with the indentations. I did try that before deciding to stay with this scheme

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
```diff
@@ -0,0 +1,69 @@
...
+<# } else if (method.MethodType == MethodType.BinaryInt) { #>
+        public DataFrame <#=method.MethodName#><T>(int value)
+<# } else { #>
+        public DataFrame <#=method.MethodName#><T>(IList<T> values)
+<# } #>
+            where T : struct
```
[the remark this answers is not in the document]
pgovind: Because we want this to be a DataFrame of primitive types. Since there's no way at the moment to specify primitive types alone, I put in value types. A DataFrame is used for data wrangling with primitive types, similar to Pandas. Is there a way to add reference types to Pandas/DataFrame that makes sense? For ex, what does it mean to say DataFrameColumn<someClass>.Max() or DataFrameColumn<someClass>.Add(5)?

src/Microsoft.Data/Dependencies/netcoreapp2.1/Apache.Arrow.deps.json
```diff
@@ -0,0 +1,80 @@
+{
```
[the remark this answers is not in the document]
pgovind: My bad. I removed the Arrow dependency, but VS still keeps the folders around apparently

corefxlab.sln
```diff
@@ -106,6 +106,13 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Text.JsonLab.Seriali
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Text.JsonLab.Serialization.Tests", "tests\System.Text.JsonLab.Serialization.Tests\System.Text.JsonLab.Serialization.Tests.csproj", "{544D4C8C-B5C6-4C3C-9763-E4CB6AF9A90C}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data", "src\Microsoft.Data\Microsoft.Data.csproj", "{9542C4FF-BC8F-47E3-916D-D2FF38003823}"
```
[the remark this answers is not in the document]
voltcode: is there going to be more than one DataFrame? What about System.Data,Frame?

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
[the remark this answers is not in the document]
voltcode: you should be able to store strings in a dataframe column.

src/Microsoft.Data/DataFrameColumn.cs
```diff
@@ -0,0 +1,73 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Data
+{
+    public class DataFrameColumn<T> : BaseDataFrameColumn
+        where T : struct
```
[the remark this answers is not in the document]
pgovind: No, I think the `T : struct` restriction is required. If there was a restriction that truly only allowed primitive types, I would've used that. The DataFrame can only hold primitive types. No ref types and ideally no user defined structs/value types

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,96 @@
...
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Data
+{
+    public class DataFrameBuffer<T>
+        where T : struct
+    {
+        public Memory<byte> Memory { get; private set; }
```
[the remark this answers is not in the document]
pgovind: Because I was going for similarity with Arrow. This way, the `Memory` can just be wrapped in an Arrow Buffer with no conversion overhead

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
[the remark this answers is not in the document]
mungojam: I would expect it to throw if I try an operation that isn't supported, but why wouldn't that one follow the same rule as number + string with scalars? i.e. convert to string

src/Microsoft.Data/DataFrameColumn.cs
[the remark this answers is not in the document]
YohDeadfall: What is a reason of that restriction?

corefxlab.sln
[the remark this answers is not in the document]
eerhardt: I don't think there will be more than one DataFrame (at least not owned/maintained/worked on by the `corefx` team at Microsoft).

We could consider putting this in `System`, but I don't think the type would ever be in the .NET Core shared framework or in the `netstandard` specification. I imagine this type to always be available through a NuGet package where we can add more APIs without having to rely on a new .NET Core version/TFM. With that in mind, it is unclear to me what should go into a `System` namespace and what shouldn't.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTableTests.cs
```diff
@@ -0,0 +1,126 @@
...
+using Microsoft.Data;
+using Xunit;
+
+namespace Microsoft.Data.Tests
+{
+    public class DataFrameTableTests
+    {
+        static public DataFrame MakeTestTableWithTwoColumns(int length)
+        {
+            BaseDataFrameColumn dataFrameColumn1 = new PrimitiveDataFrameColumn<int>("Int1", Enumerable.Range(0, length).Select(x => x));
```
mungojam: PrimitiveDataFrameColumn feels very verbose for such a common thing. It would be good to have some shortcuts to create them, like some static functions that could at least infer the generic type and maybe just be called Column()

corefxlab.sln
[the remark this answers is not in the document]
voltcode: DataFrame is yet another abstraction over data in general and I think putting it in System is better than putting it in Microsoft. namespace, when other data-related abstractions are in System.Data, etc. It is not windows-specific, and should not be seen as such to encourage wider adoption.

src/Microsoft.Data/DataFrameBuffer.cs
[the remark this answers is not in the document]
voltcode: Is this going to be compatible with Arrow? I've seen the dependency commented out, but I'm not sure whether Arrow is going to be supported as the engine behind the interface.

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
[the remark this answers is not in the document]
pgovind: Ugh, my bad. I typed in a hurry. I meant to write "DataFrameColumn - string"(Basically whenever we have incompatible operands). Edited it now.

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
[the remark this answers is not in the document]
pgovind: I agree. At the moment, my plan is to store strings as `byte`s so I can still keep the struct constraint

src/Microsoft.Data/DataFrameBuffer.cs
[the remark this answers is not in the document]
pgovind: Yes, this is going to be compatible with Arrow but, if I understand you right, not exactly in the way you mention. The C# Arrow implementation as it exists now requires that the buffer storing the data is immutable. No APIs exist to mutate the contents of an ArrowBuffer. I had a discussion in the Arrow repo a month or so back about removing this restriction, but a lot of stuff has already been built with immutability in mind, so I think it's safe to say that that is a no-go.

By compatible here I mean that a `DataFrameBuffer` can be wrapped in an `ArrowBuffer` easily with no overhead. That way the DataFrameBuffer can be mutated by the DataFrame, and anyone who consumes an `ArrowBuffer` can also get to the data.

src/Microsoft.Data/DataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        public IList<T> this[long startIndex, int length]
+        {
+            get
+            {
+                var ret = new List<T>();
+                long endIndex = startIndex + length;
+                int arrayIndex = GetArrayContainingRowIndex(ref startIndex);
+                bool temp = Buffers[arrayIndex][(int)startIndex, length, ret];
```
ahsonkhan: Are we fine with this downcasting from long to int?

src/Microsoft.Data/DataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+            set
+            {
+                int arrayIndex = GetArrayContainingRowIndex(ref rowIndex);
+                Buffers[arrayIndex][(int)rowIndex] = value;
+            }
+        }
+
+        public override string ToString()
+        {
+            string ret = "";
```
ahsonkhan: Maybe use a string builder.

src/Microsoft.Data/DataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+                    ret += "...";
+                    break;
+                }
+            }
+            return ret;
+        }
+        public PrimitiveDataFrameColumnContainer<T> Clone()
+        {
+            var ret = new PrimitiveDataFrameColumnContainer<T>();
+            foreach (var buffer in Buffers)
```
ahsonkhan: nit: avoid var here.

src/Microsoft.Data/DataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+            {
+                DataFrameBuffer<T> newBuffer = new DataFrameBuffer<T>();
```
ahsonkhan: nit: But can use var here.

src/Microsoft.Data/DataFrameColumnContainerBinaryOperations.cs
```diff
@@ -0,0 +1,172 @@
...
+        where T : struct
+    {
+       public PrimitiveDataFrameColumnContainer<T> Add(PrimitiveDataFrameColumnContainer<T> right)
+        {
+            DataFrameColumnArithmetic<T>.Instance.Add(this, right);
+            return this;
+        }
+       public PrimitiveDataFrameColumnContainer<T> Add(T scalar)
+        {
+
```
ahsonkhan: nit: unnecessary new line, and indenting is off.

src/Microsoft.Data/DataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        internal PrimitiveDataFrameColumnContainer<bool> CreateBoolContainerForCompareOps()
+        {
+            var ret = new PrimitiveDataFrameColumnContainer<bool>();
+            foreach(var buffer in Buffers)
+            {
+                DataFrameBuffer<bool> newBuffer = new DataFrameBuffer<bool>();
+                ret.Buffers.Add(newBuffer);
+                var span = buffer.Span;
```
ahsonkhan: This `span` isn't used anywhere.

tests/Microsoft.Data.DataFrame.Tests/Microsoft.DataFrames.Tests.csproj
```diff
@@ -0,0 +1,26 @@
...
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="xunit" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\..\src\Microsoft.Data\Microsoft.DataFrames.csproj" />
+  </ItemGroup>
+
+  <!--<ItemGroup>
```
eerhardt: This can be deleted.

src/Microsoft.Data/DataFrameBuffer.cs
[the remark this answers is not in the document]
eerhardt: The issue is below:

```C#
public Span<T> Span
        {
            get => MemoryMarshal.Cast<byte, T>(Memory.Span);
        }
```

This isn't a free operation. If someone calls this many times in a hot path, this Cast is going to take cycles, when if it was just in terms of `T` to begin with, there is no need for a cast.

Similarly, you wouldn't need to hold onto `_size` and do division for things like Capacity and MaxCapacity.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTableTests.cs
```diff
@@ -0,0 +1,126 @@
+// Licensed to the .NET Foundation under one or more agreements.
```
eerhardt: We may need a little more tests in this PR. This doesn't seem like enough given the functionality in the library with this PR.

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,96 @@
...
+                return true;
+            }
+        }
+        public override string ToString()
+        {
+            string ret = "";
+            var span = Span;
+            for (int ii = 0; ii < Length; ii++)
+            {
+                ret += span[ii] + " ";
```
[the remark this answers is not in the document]
pgovind: Yup, I think so. I'm overriding this mainly for usage from an interactive shell where df["columnName"] should print the values in a column. Similar to what pandas does at the moment

src/Microsoft.Data/DataFrameColumn.cs
```diff
@@ -0,0 +1,73 @@
...
+                return _columnContainer[startIndex, length];
+            }
+        }
+
+        // This method involves boxing
+        public override object this[long rowIndex]
+        {
+            get
+            {
+                return _columnContainer[rowIndex];
```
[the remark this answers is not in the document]
pgovind: _columnContainer implements the bounds check

src/Microsoft.Data/DataFrameColumnContainer.cs
pgovind: Yup yup. This is on purpose

src/Microsoft.Data/DataFrameColumnArithmeticTemplate.cs
```diff
@@ -0,0 +1 @@
+﻿
```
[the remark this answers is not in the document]
pgovind: No. Deleting this file

tests/Microsoft.Data.DataFrame.Tests/DataFrameTableTests.cs
pgovind: I sort of agree. With this patch, the only ops allowed are things like floatColumn + float, so the scope is pretty narrow. Would you consider waiting till the next patch(will be up in an hour) where I've added a couple more tests for binary ops? That patch tests different functions and more combinations such as int + bool/double/decimal

eerhardt: This is OK for now. Let's be sure to add adequate tests in future PRs.

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+namespace Microsoft.Data
+{
+    public partial class PrimitiveDataFrameColumnContainer<T>
+        where T : struct
+    {
+        public IList<DataFrameBuffer<T>> Buffers = new List<DataFrameBuffer<T>>();
+        public PrimitiveDataFrameColumnContainer(T[] values)
+        {
+            values = values ?? throw new ArgumentNullException(nameof(values));
+            long ii = 0;
```
eerhardt: The variable can be in the `for` loop on line 30.

BTW - standard convention is just to use `i` in for loops (or a full named variable).  See [https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/for#structure-of-the-for-statement](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/for#structure-of-the-for-statement)

```C#
for (int i = 0; i < 5; i++)
{
    Console.WriteLine(i);
}
```

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+                curArrayIndex++;
+            }
+            return curArrayIndex;
+        }
+
+        public IList<T> this[long startIndex, int length]
+        {
+            get
+            {
+                var ret = new List<T>();
```
eerhardt: You should initialize this list to the length capacity. That way it doesn't have to grow while you are adding elements to it.

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+                long endIndex = startIndex + length;
+                int arrayIndex = GetArrayContainingRowIndex(ref startIndex);
+                bool temp = Buffers[arrayIndex][(int)startIndex, length, ret];
+                while (ret.Count < length)
+                {
+                    long nextRowIndex = startIndex + ret.Count + 1;
+                    arrayIndex++;
+                    temp = Buffers[arrayIndex][(int)nextRowIndex, length - ret.Count, ret];
```
eerhardt: Even though this is internal, this API usage/pattern is a little unusual:

```[(int)nextRowIndex, length - ret.Count, ret];```

Specifically, passing in an object `ret` to get filled up. It feels like it should just be a standard method call.

Comment (eerhardt) [inline src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs]: Is casting `nextRowIndex` to `int`, the right thing here? Same for above where we are casting `startIndex` to an `int`. Why take in a `long` if you are casting it to `int`?

File: src/Microsoft.Data/DataFrameColumnContainerBinaryOperations.cs [hunk 24]
```diff
@@ -0,0 +1,162 @@
+﻿
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+namespace Microsoft.Data
+{
+    public partial class PrimitiveDataFrameColumnContainer<T>
```
eerhardt: Can you keep the file names and the type that is contained in the file the same? That way it is easier to find classes when navigating the file system.

For partial classes, a typical pattern is use put a `.` in the file name to separate type name from the category of things in it. So this file could be called `PrimitiveDataFrameColumnContainer.BinaryOperations.cs`.

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -0,0 +1,130 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Data
+{
+    public class DataFrameTable
```
eerhardt: What's the difference between a `DataFrame` and a `DataFrameTable`?

Maybe it would be good to have high-level class summary comments describing each public class.  You can write them like this:


```
    /// <summary>
    /// High level summary comment
    /// </summary>
   public class Foo
```

src/Microsoft.Data/Microsoft.DataFrames.csproj
```diff
@@ -0,0 +1,67 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
+    <DefaultItemExcludes>$(DefaultItemExcludes);Dependencies\**\*</DefaultItemExcludes>
+  </PropertyGroup>
+
+  <!--<ItemGroup>
```
eerhardt: These commented out lines can be deleted.

src/Microsoft.Data/Microsoft.DataFrames.csproj
```diff
@@ -0,0 +1,67 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
+    <DefaultItemExcludes>$(DefaultItemExcludes);Dependencies\**\*</DefaultItemExcludes>
```
eerhardt: This line can be deleted.

src/Microsoft.Data/Microsoft.DataFrames.csproj
```diff
@@ -0,0 +1,67 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0;netcoreapp2.1</TargetFrameworks>
```
eerhardt: I don't see us needing multi-targeting (yet).  This can change to just be:

```
<TargetFramework>netstandard2.0</TargetFramework>
```

tests/Microsoft.Data.DataFrame.Tests/Microsoft.DataFrames.Tests.csproj
```diff
@@ -0,0 +1,22 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+
+    <IsPackable>false</IsPackable>
```
eerhardt: I don't think you need this

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
```diff
@@ -0,0 +1,73 @@
...
+    {
+        internal PrimitiveDataFrameColumnContainer<T> _columnContainer;
+        public Type DataType = typeof(T);
+        public PrimitiveDataFrameColumn(string name, PrimitiveDataFrameColumnContainer<T> column) : base(name, column.Length)
+        {
+            _columnContainer = column;
+        }
+        public PrimitiveDataFrameColumn(string name, T[] columnData) : base(name)
+        {
+            throw new NotImplementedException();
```
eerhardt: Maybe let's delete this constructor until it is implemented...

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
```diff
@@ -0,0 +1,158 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    public partial class PrimitiveDataFrameColumnContainer<T>
```
eerhardt: From a public (or a user's) perspective, what is the difference between PrimitiveDataFrameColumn and PrimitiveDataFrameColumnContainer?

src/Microsoft.Data/DataFrameColumnArithmeticTemplate.ttinclude
```diff
@@ -0,0 +1,228 @@
...
+            return $"void {MethodName}({arguments})";
+        }
+
+        public string GetSingleArgumentMethodSignature(string columnType, string genericType)
+        {
+            var arguments = GetSingleParameterMethodArguments(columnType, genericType);
+            return $"PrimitiveDataFrameColumnContainer<T> {MethodName}({arguments})";
+        }
+        
+        public string GetMethodArguments(string tensorType, string genericType)
```
eerhardt: `tensorType`?

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
pgovind: Right. I don't like that I have to do this either. Basically, this is a slice of length "length" that can begin in the middle of some buffer and spill over into the next buffer. I was trying to get by with passing just 1 return list "ret" to both calls to avoid creating more objects. That's why the return type of the indexer is a bool. Ideally, if there was API that made a List from a Span efficiently, or that allowed me to append to a List from a Span, I would've used that. Unless there's a less ugly way to do this? Definitely open to suggestions

pgovind: nextRowIndex should've been 0. That's a bug. Fixing it.

I'm taking a long because the df can hold long number of elements. Each buffer can hold only intMax number of elements though, so I'm casting once I get to the right buffer

src/Microsoft.Data/DataFrameTable.cs
pgovind: A DataFrameTable is just a convenient store to hold the columns. The `DataFrame` class uses it to manipulate the columns and implement its algorithms.

src/Microsoft.Data/Microsoft.DataFrames.csproj
```diff
@@ -0,0 +1,67 @@
...
+      <HintPath>Dependencies\netcoreapp2.1\Apache.Arrow.dll</HintPath>
+    </Reference>
+    
+    <Reference Include="Apache.Arrow" Condition="'$(TargetFramework)' == 'net472'">
+      <HintPath>Dependencies\netstandard1.3\Apache.Arrow.dll</HintPath>
+    </Reference>
+  </ItemGroup>-->
+
+  <ItemGroup>
+    <PackageReference Include="CsvHelper" Version="12.1.2" />
```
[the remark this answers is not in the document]
pgovind: Good catch. I put this in a long time ago to write unit tests

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
pgovind: A `PrimitiveDataFrameColumnContainer` is just an internal store(changed the modifier to `internal` now) to hold the column data. `PrimitiveDataFrameColumn` defines the APIs that manipulate the container

pgovind:
Just a note: I'm hoping to get this approved soon so I can put up the patches I have for StringColumn support and BinaryOperator overloads

src/Microsoft.Data/Microsoft.DataFrames.csproj
```diff
@@ -0,0 +1,60 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
ahsonkhan: Generally, the project name and the namespace match. Should the namespace be `Microsoft.DataFrames` instead of `Microsoft.Data` or should the project file be changed to `Microsoft.Data.csproj`.

tests/Microsoft.Data.DataFrame.Tests/Microsoft.DataFrames.Tests.csproj
```diff
@@ -0,0 +1,24 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
ahsonkhan: Similarly here, should it be either `Microsoft.Data.Tests` or `Microsoft.DataFrame.Tests`? If we want it to be `Microsoft.Data.DataFrame`, then let's be consistent in the source project/namespace as well. I don't have a strong opinion one way or the other.

danmosemsft:
At some point you will want to check the code coverage for your tests.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTableTests.cs
```diff
@@ -0,0 +1,128 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System.Linq;
+using System.Collections.Generic;
+using Microsoft.Data;
+using Xunit;
+
+namespace Microsoft.Data.Tests
+{
+    public class DataFrameTableTests
```
eerhardt: This should just be `DataFrameTests` right?  `DataFrameTable` is an internal type and we generally write unit tests at the public API surface.

src/Microsoft.Data/Microsoft.DataFrames.csproj
```diff
@@ -0,0 +1,60 @@
...
+
+  <PropertyGroup>
+    <TargetFrameworks>netstandard2.0</TargetFrameworks>
```
eerhardt: Since this is only 1 TFM, you can use the `TargetFramework` property, not `TargetFrameworks`. You only need the plural when you have more than 1.

src/Microsoft.Data/BinaryOperations/DataFrameBinaryOperations.tt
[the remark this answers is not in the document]
pgovind: I filed an issue to supports string columns in the DataFrame. Working on a patch for it at the moment

src/Microsoft.Data/DataFrameColumn.cs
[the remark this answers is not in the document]
pgovind: This has been renamed to PrimitiveColumn, so the T : struct remains. Strings will be supported with a new column type

src/Microsoft.Data/DataFrameBuffer.cs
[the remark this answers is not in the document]
pgovind: I created an issue for this

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
eerhardt: I think we can just make this a method and not use an indexer. Honestly, I don't think I've ever seen someone use a C# indexer this way.

eerhardt:
```C#
        public IList<T> this[long startIndex, int length]
        {
            get
            {
                var ret = new List<T>(length);
                long endIndex = startIndex + length;
                int arrayIndex = GetArrayContainingRowIndex(ref startIndex);
                bool temp = Buffers[arrayIndex][(int)startIndex, length, ret];
```

Here we take a `long startIndex` and then are casting it `(int)startIndex`. 

I see that `GetArrayContainingRowIndex` is decreasing the number, which honestly is not intuitive.  Maybe make `GetArrayContainingRowIndex` return an `out int newStartIndex` (or a better name). Have `GetArrayContainingRowIndex` ensure that the number is less than `int.MaxValue`, then the 3 calling places don't need to cast here. If it ever fails, I'm not sure what would even happen with the current code.

src/Microsoft.Data/Microsoft.DataFrames.csproj
pgovind: I changed it back to `Microsoft.Data` for now. I'll come back to this in the next patch and change the assembly and namespace names. I have other patches built on top of this are getting harder and harder to rebase with these .csproj changes.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTableTests.cs
pgovind: Done

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -0,0 +1,130 @@
...
+                ret.Add(Column(ii)[rowIndex]);
+            }
+            return ret;
+        }
+
+        public void InsertColumn<T>(int columnIndex, IEnumerable<T> column, string columnName)
+            where T : struct
+        {
+            column = column ?? throw new ArgumentNullException(nameof(column));
+            if (columnIndex < 0 || columnIndex > _columns.Count)
```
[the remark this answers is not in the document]
pgovind: Great suggestions! I have another big patch coming that enables things like intColumn + decimalColumn. I'll address all this feedback in that patch. Its getting to be too much of a pain to keep rebasing(and reordering) my patches on this with all the csproj names and other changes

src/Microsoft.Data/Microsoft.DataFrames.csproj
ahsonkhan: I am going to push-back **slightly** on renaming the assembly after the initial check-in since it would end up publishing a NuGet package to our feed (with the same name) which would become out-dated and have to be cleaned up (after the rename). That said, **it's generally fine if we do that,** but if we know that a rename is required up front, it might be better to just settle on the name now.

Although, I understand your rebase/patch work that is being done on top of this PR, so I don't have any significant pushback form keeping it as is and renaming it later :)

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
```diff
@@ -0,0 +1,160 @@
...
+            set
+            {
+                int arrayIndex = GetArrayContainingRowIndex(ref rowIndex);
+                Buffers[arrayIndex][(int)rowIndex] = value;
+            }
+        }
+
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder("");
```
danmosemsft: No need for `""`

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.cs
```diff
@@ -0,0 +1,160 @@
...
+            foreach (var buffer in Buffers)
+            {
+                sb.Append(buffer.ToString());
+                // Can this run out of memory? Just being safe here
```
danmosemsft: StringBuilder can hold int.MaxValue chars unless otherwise specified. ToString is really for debugging, so a limited size string is nice from that point of view..

src/Microsoft.Data/DataFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            _table = new DataFrameTable();
+        }
+
+        public long RowCount => _table.RowCount;
+        public int ColumnCount => _table.ColumnCount;
+        public IList<string> Columns()
+        {
+            var ret = new List<string>();
+            for (int ii = 0; ii < ColumnCount; ii++)
```
danmosemsft: Any particular reason you use `ii` -- is this standard in this domain? We usually use `i`, `j` .. or something descriptive

pgovind: Nah. I'll change that now. I started doing that a while back to make the loop variable more unique so that I didn't get as many hits when I used "Find"

src/Microsoft.Data/DataFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+        public DataFrame()
+        {
+            _table = new DataFrameTable();
+        }
+
+        public long RowCount => _table.RowCount;
+        public int ColumnCount => _table.ColumnCount;
+        public IList<string> Columns()
+        {
+            var ret = new List<string>();
```
danmosemsft: `var ret - new List<string>(ColumnCount);`

src/Microsoft.Data/DataFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            return ret;
+        }
+        public BaseDataFrameColumn Column(int index) => _table.Column(index);
+        public void InsertColumn(int columnIndex, BaseDataFrameColumn column) => _table.InsertColumn(columnIndex, column);
+        public void SetColumn(int columnIndex, BaseDataFrameColumn column) => _table.SetColumn(columnIndex, column);
+        public void RemoveColumn(int columnIndex) => _table.RemoveColumn(columnIndex);
+        public void RemoveColumn(string columnName) => _table.RemoveColumn(columnName);
+        
+        public object this[long rowIndex, int columnIndex]
```
danmosemsft: Use expression bodied form for oneliners
https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/properties#expression-body-definitions
eg
``c#
        public object this[long rowIndex, int columnIndex]
        {
            get => _table.Column(columnIndex)[rowIndex];
            set =>_table.Column(columnIndex)[rowIndex] = value;
        }
```
easier to read/shorter

File: src/Microsoft.Data/DataFrame.cs (hunk 42 above)
Comment (danmosemsft) [reply src/Microsoft.Data/DataFrame.cs]: generally when you know the capacity you'll need you would do this to avoid it realocating/copying. Same below

File: src/Microsoft.Data/DataFrame.cs [hunk 44]
```diff
@@ -0,0 +1,92 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// A DataFrame
```
danmosemsft: It would be nice to write what such a thing is and what it does, and any interesting comments about its behavior/perf characteristics. I tend to do that realtively early as it helps me think...

src/Microsoft.Data/DataFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// A DataFrame 
+    /// </summary>
+    public partial class DataFrame
+    {
+        internal DataFrameTable _table;
```
danmosemsft: Please avoid internal fields ... fields are pretty much always private, and properties otherwise, except special cases

src/Microsoft.Data/BaseDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+namespace Microsoft.Data
+{
+    public abstract class BaseDataFrameColumn
+    {
+        public BaseDataFrameColumn(string name, long length = 0)
+        {
+            Length = length;
+            Name = name;
+        }
+        public long Length { get; protected set; }
```
danmosemsft: Check for >= 0 ? Since it's not got a private setter.

src/Microsoft.Data/BaseDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+        public long NullCount { get; protected set; }
+        public string Name;
+
+        public virtual object this[long rowIndex] { get { throw new NotImplementedException(); } set { throw new NotImplementedException(); } }
```
danmosemsft: This is temporary code, but long term it should be on several lines (using => syntax if it's oneliners)

src/Microsoft.Data/DataFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+        public DataFrame()
+        {
+            _table = new DataFrameTable();
+        }
+
+        public long RowCount => _table.RowCount;
+        public int ColumnCount => _table.ColumnCount;
+        public IList<string> Columns()
```
danmosemsft: Although general OOP principle is to return the base type (or interface) for future flexibility, I am not sure we generally do that in API design. There is some perf impact (although probably immeasurable in most cases). Anyway it doesn't matter - easy to adjust during API review - but if you're interested, ask Immo

src/Microsoft.Data/DataFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+        {
+            var ret = new List<string>();
+            for (int ii = 0; ii < ColumnCount; ii++)
+            {
+                ret.Add(_table.Column(ii).Name);
+            }
+            return ret;
+        }
+        public BaseDataFrameColumn Column(int index) => _table.Column(index);
```
danmosemsft: blank lines between each please and blank line after `}` unless followed by `}` is our pattern I think

src/Microsoft.Data/DataFrame.cs
```diff
@@ -0,0 +1,92 @@
...
+            }
+            //TODO?: set?
+        }
+
+        public object this[string columnName]
+        {
+            get
+            {
+                int columnIndex = _table.GetColumnIndex(columnName);
+                if (columnIndex == -1) throw new ArgumentException($"{columnName} does not exist");
```
danmosemsft: Should `GetColumnIndex` throw instead? Or is asking for an unknown column a reasonable thing to do?

src/Microsoft.Data/DataFrameBinaryOperations.cs
```diff
@@ -0,0 +1,584 @@
...
+                }
+            }
+            return newDataFrame;
+        }
+
+        public DataFrame Add<T>(T value)
+            where T : struct
+        {
+            var newDataFrame = new DataFrame();
+            for (int ii = 0; ii < ColumnCount; ii++)
```
danmosemsft: Extract common code ..?

src/Microsoft.Data/DataFrameBinaryOperations.cs
```diff
@@ -0,0 +1,584 @@
...
+            {
+                PrimitiveDataFrameColumn<T> column = _table.Column(ii) as PrimitiveDataFrameColumn<T>;
+                if (column != null)
```
danmosemsft: is it correct that it should jsut return silently if this is null (if it's not the expected type)?

src/Microsoft.Data/DataFrameBinaryOperations.cs
danmosemsft: and others below

danmosemsft: it would be nice if this loop appeared exacctly once in this file, which would mean calling Add, Xor etc through delegates. That would be nice unless these were relatively hot calls, which I imagine we expect them to be. A call through a delegate is never going to be quite as fast.

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,108 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Data
+{
+    public class DataFrameBuffer<T>
```
danmosemsft: Some <summary> blurb would be nice for the major types, even if only for code reviewers

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,108 @@
...
+        public Memory<byte> Memory { get; private set; }
+        private readonly int _size;
+
+        private int Capacity => Memory.Length / _size;
+
+        public int MaxCapacity => Int32.MaxValue / _size;
+
+        public Span<T> Span
+        {
+            [MethodImpl(MethodImplOptions.AggressiveInlining)]
```
danmosemsft: Generally we try to avoid inlining attributes unless we know for sure we need them. ie., we know it will not get inlined (for some bad reason) and we know inlining is important to get the performance we want. In this case I would trust it will get inlined, and when we do perf work later, if it's not, we can do appropriate things (which might include modifying the jit)

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,108 @@
...
+
+        public DataFrameBuffer(int numberOfValues = 8)
+        {
+            _size = Unsafe.SizeOf<T>();
+            if ((long)numberOfValues * _size > MaxCapacity)
+            {
+                throw new ArgumentException($"{numberOfValues} exceeds buffer capacity", nameof(numberOfValues));
+            }
+            Memory = new byte[numberOfValues * _size];
+            Length = 0;
```
danmosemsft: This is C#, fields are initialized to their defaults for free

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,96 @@
...
+                var memory = new Memory<byte>(new byte[newCapacity]);
+                Memory.CopyTo(memory);
+                Memory = memory;
+            }
+        }
+        internal T this[int index]
+        {
+            get
+            {
+                if (index > Length) throw new ArgumentOutOfRangeException(nameof(index));
```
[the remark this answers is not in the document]
danmosemsft: Note you only do this when it is definitely hot path

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,108 @@
...
+                    returnList.Add(Span[ii]);
+                }
+                return true;
+            }
+        }
+        public override string ToString()
+        {
+            string ret = "";
+            Span<T> span = Span;
+            for (int ii = 0; ii < Length; ii++)
```
danmosemsft: Unless you know Length is never more than 3 or 4 or so, use a StringBuilder. Also if it can be really large, crop it.

src/Microsoft.Data/DataFrameColumnArithmeticTemplate.cs
danmosemsft: Does templating require the empty file be in the repo for some reason?

src/Microsoft.Data/DataFrameColumnArithmeticTemplate.ttinclude
```diff
@@ -0,0 +1,228 @@
...
+        new MethodConfiguration("GreaterThanOrEqual", MethodType.ComparisonScalar, ">=", isNumeric:true),
+        new MethodConfiguration("LessThanOrEqual", MethodType.Comparison, "<=", isNumeric:true),
+        new MethodConfiguration("LessThanOrEqual", MethodType.ComparisonScalar, "<=", isNumeric:true),
+        new MethodConfiguration("GreaterThan", MethodType.Comparison, ">", isNumeric:true),
+        new MethodConfiguration("GreaterThan", MethodType.ComparisonScalar, ">", isNumeric:true),
+        new MethodConfiguration("LessThan", MethodType.Comparison, "<", isNumeric:true),
+        new MethodConfiguration("LessThan", MethodType.ComparisonScalar, "<", isNumeric:true),
+    };
+
+
```
danmosemsft: Nit, double line

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -0,0 +1,133 @@
...
+{
+    /// <summary>
+    /// A DataFrameTable is just a container that holds a number of DataFrameColumns. It mainly acts as a convenient store to allow DataFrame to implement its algorithms
+    /// </summary>
+    internal class DataFrameTable
+    {
+        private IList<BaseDataFrameColumn> _columns;
+
+        private List<string> _columnNames = new List<string>();
+        private Dictionary<string, int> _columnNameToIndexDictionary = new Dictionary<string, int>();
```
danmosemsft: When constructing a dictionary with string keys, it's nice to always specify a comparer. Since half the time there isn't one, you intended case insensitive, or you intended culture invariant. It makes it explicit

src/Microsoft.Data/DataFrameBinaryOperations.cs
pgovind: It's on the next patch :)

src/Microsoft.Data/DataFrameBinaryOperations.cs
pgovind: Yup, this is wrong. It's hard to support binary ops on generic columns at the moment. My next patch deals with this pretty well though

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -0,0 +1,133 @@
...
+        public int ColumnCount { get; private set; } = 0;
+        public DataFrameTable()
+        {
+            _columns = new List<BaseDataFrameColumn>();
+        }
+
+        public DataFrameTable(IList<BaseDataFrameColumn> columns)
+        {
+            columns = columns ?? throw new ArgumentNullException(nameof(columns));
+            _columns = columns;
```
danmosemsft: I don't know what the scenario is for this ctor, but it should be such that the caller has full knowledge that you're going to take ownership of the list and they presumably are to not touch it again. In general it is a bit unusual for a non trivial type to be passed a field to use directly as a private field. It's not really very private. But of course depends on the semantics you're looking for

src/Microsoft.Data/DataFrameTable.cs
danmosemsft: I'm guessing you want STringComparer.CurrentCulture or Ordinal, without IgnoreCase, ?

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -0,0 +1,133 @@
...
+                throw new ArgumentException($"Table already contains a column called {column.Name}");
+            }
+            _columnNameToIndexDictionary.Remove(_columnNames[columnIndex]);
+            _columnNames[columnIndex] = column.Name;
+            _columnNameToIndexDictionary[column.Name] = columnIndex;
+            _columns[columnIndex] = column;
+        }
+        public void RemoveColumn(int columnIndex)
+        {
+            _columnNameToIndexDictionary.Remove(_columnNames[columnIndex]);
```
danmosemsft: Should you validate for -1

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
```diff
@@ -0,0 +1,4138 @@
+﻿
```
danmosemsft: Nit, blank lines

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
```diff
@@ -0,0 +1,4138 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+
+namespace Microsoft.Data
+{
+    internal interface IPrimitiveDataFrameColumnArithmetic<T>
+        where T : struct
+    {
+       void Add(PrimitiveDataFrameColumnContainer<T> left, PrimitiveDataFrameColumnContainer<T> right);
```
danmosemsft: Consider sorting. It's at least a stable ordering..

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
```diff
@@ -0,0 +1,4138 @@
...
+    {
+        public static IPrimitiveDataFrameColumnArithmetic<T> Instance => PrimitiveDataFrameColumnArithmetic.GetArithmetic<T>();
+    }
+
+    internal static class PrimitiveDataFrameColumnArithmetic
+    {
+        public static IPrimitiveDataFrameColumnArithmetic<T> GetArithmetic<T>()
+            where T : struct
+        {
+            if (typeof(T) == typeof(bool))
```
danmosemsft: why not use a `switch` statement on the type?

danmosemsft: you could also set a return value then return `return (IPrimitiveDataFrameColumnArithmetic<T>)value`  which will save all these pasted casts

src/Microsoft.Data/DataFrameBuffer.cs
pgovind: Got it. This will go away soon enough though for https://github.com/dotnet/corefxlab/issues/2659

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
```diff
@@ -0,0 +1,4138 @@
...
+            {
+                return (IPrimitiveDataFrameColumnArithmetic<T>)new UShortArithmetic();
+            }
+            return null;
+        }
+    }
+
+    internal class BoolArithmetic : IPrimitiveDataFrameColumnArithmetic<bool>
+    {
+        public void Add(PrimitiveDataFrameColumnContainer<bool> left, PrimitiveDataFrameColumnContainer<bool> right)
```
danmosemsft: VS has a magic refactoring that will implement an interface for you. When it does this it uses NotImplementedEXception. You should probably use that.

NotImplementedException --> I am not done here
NotSupportedException --> It is not meaningful to call this (usually when something really should be abstract)
PlatformNotSupportedException --> Works on other platforms, not this one

danmosemsft: ie in theory in CoreFX there are zero NotImplementedException, but in practice I see we weren't consistent

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.BinaryOperations.cs
```diff
@@ -0,0 +1,162 @@
+﻿
```
danmosemsft: Nit, newline

src/Microsoft.Data/DataFrameColumnArithmeticTemplate.cs
pgovind: Removed. I had to rebase on upstream/master and it got generated this time.

src/Microsoft.Data/DataFrameTable.cs
pgovind: Nope I think. _columnName[columnIndex] will throw the exception I want

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
pgovind: I was able to remove 1 of the blank lines in my next patch, but for the life of me, I am not able to find out where the first blank line is coming from. All I know is that it is from some random space in the .ttinclude file

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
pgovind: Next patch. This'll cause too much pain when I rebase

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
pgovind: Oh, I just took this from some of the Tensor code we have checked in, so I didn't pay too much attention to it. Will consider it for my next patch

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
pgovind: NotSupportedException is the right thing here I think. It doesn't make sense to call any arithmetic ops on a bool

src/Microsoft.Data/PrimitiveDataFrameColumnContainer.BinaryOperations.cs
pgovind: Same as the prev comment. I really wish I knew where in the ttinclude file this was coming from

src/Microsoft.Data/DataFrame.cs
pgovind: I'm not throwing yet because I want to support things like:

df["Int3"] = df["Int1"] + df["Int2"];

where df["Int3"] will be created if it doesn't exist. So, I'm not throwing yet. I don't like returning -1 either, so if I can write the setter (when I get around to it) without the -1, I'll add code to throw then.

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -0,0 +1,96 @@
...
+        }
+        public int Length { get; internal set; }
+
+        public DataFrameBuffer(int numberOfValues = 8)
+        {
+            _size = Unsafe.SizeOf<T>();
+            Memory = new byte[numberOfValues * _size];
+            Length = 0;
+        }
+        public void Append(T value)
```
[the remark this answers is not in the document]
pgovind: Note: As part of perf work

src/Microsoft.Data/DataFrameBuffer.cs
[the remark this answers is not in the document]
pgovind: Note: As part of perf work

src/Microsoft.Data/Microsoft.DataFrames.csproj
pgovind: Hmmm, if I name the assembly Microsoft.DataFrame, the type is also called DataFrame, so usage now needs to be qualified as DataFrame.DataFrame, which seems silly. Microsoft.DataFrames is misleading considering there is likely only going to be 1 DataFrame. What about Microsoft.Data.DataFrame with a namespace of Microsoft.Data?

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
gfoidl: How would you specify the `case` where a constant value is required, i.e.
```c#
switch (typeof(T))
{
    case typeof(int):
// ...
```
Won't compile (CS0150) (with C# 7.3 and any version SharpLab offers).
Or does pattern matching provide a syntax I'm not aware of?

eerhardt: I'm not an expert on this, but my understanding of this pattern was that the JIT was smart enough to eliminate chunks of code. When JIT'ing for `T` == `bool`, it can see that `if (typeof(T) == typeof(bool))` always holds, similarly `else if (typeof(T) == typeof(int))` never holds, so it can completely remove non-`bool` code paths.

src/Microsoft.Data/DataFrameTable.cs
YohDeadfall: It's a bad practice not to check parameters and delegate their validation.

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -0,0 +1,142 @@
...
+            int columnIndex = GetColumnIndex(columnName);
+            if (columnIndex != -1)
+            {
+                RemoveColumn(columnIndex);
+            }
+        }
+
+        public int GetColumnIndex(string columnName)
+        {
+            if (_columnNameToIndexDictionary.TryGetValue(columnName, out int columnIndex) )
```
YohDeadfall: nit: Space between closing parenthesis.

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -0,0 +1,142 @@
...
+
+        public DataFrameTable()
+        {
+            _columns = new List<BaseDataFrameColumn>();
+        }
+
+        private DataFrameTable(IList<BaseDataFrameColumn> columns)
+        {
+            columns = columns ?? throw new ArgumentNullException(nameof(columns));
+            _columns = columns;
```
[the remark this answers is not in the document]
pgovind: I don't understand?? I'm assigning to `_columns`? `columns` without the "_" is the parameter. Technically I'm not 100% convinced with this code yet. I'm still not certain if I should be making a copy of the `columns` being passed in, so I'll wait for more usage scenarios before I make the choice

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -0,0 +1,114 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp2.2</TargetFramework>
```
eerhardt: Can this not be `netstandard2.0`?

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -0,0 +1,114 @@
...
+      <LastGenOutput>PrimitiveDataFrameColumnContainer.BinaryOperations.cs</LastGenOutput>
+    </None>
+  </ItemGroup>
+
+  <ItemGroup>
+    <Service Include="{508349b6-6b84-4df5-91f0-309beebad82d}" />
+  </ItemGroup>
+
+  <ItemGroup>
+    <Compile Update="BaseColumn.BinaryOperations.cs">
```
eerhardt: I don't see this file in your PR. Are you sure all of these are still necessary?

src/Microsoft.Data/BaseDataFrameColumn.cs
```diff
@@ -0,0 +1,41 @@
...
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// The base column type. All APIs should have atleast a stub here first
```
eerhardt: `atleast` is two words.

src/Microsoft.Data/DataFrameTable.cs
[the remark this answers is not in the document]
YohDeadfall: You're assigning `columns` to `columns` in line 32 to check for a null value. Just check the value while assigning it to the field:

```suggestion
            _columns = columns ?? throw new ArgumentNullException(nameof(columns));
```

src/Microsoft.Data/PrimitiveDataFrameColumnArithmetic.cs
gfoidl: @eerhardt correct.

The JIT does this by dead code elimination, as `typeof` is a JIT-time-constant, so all other branches will fall away.

danmosemsft: Interesting.

Issue #2659, opened 2019-04-16
DataFrame: DataFrameBuffer should hold a Memory<T> directly instead of Memory<byte>


pgovind:
`Buffer.Span` is a cast at the moment for each call. Having a Memory<T> as the backing store would avoid the cast.

Creating this to address feedback from [2656](https://github.com/dotnet/corefxlab/pull/2656)

pgovind:
Nope. Storing a Memory<T> would mean Arrow interop is not zero-copy anymore. To offset the repeated cost of Buffer.Span calls, DataFrame now offers access to the underlying Memory<byte> itself, so users can cast the Memory<byte> -> Memory<T> once.

PR #2660, opened 2019-04-23
Shorten names. Implement string columns and computations

pgovind:
More work towards [#26854](https://github.com/dotnet/corefx/issues/26845) to add a DataFrame type to .NET Core

This patch:
Improves binary operations on primitive columns
Creates a StringColumn to support columns of strings. It also implements some meaningful binary ops on string columns
Implements basic computations on primitive columns

I added another patch that implements "DataFrame.Sort" too. The changes are mostly new and incremental. Apologies for putting it in this patch and making the review harder. I'm on app building for the rest of the week, so I wanted to get this out soon


eerhardt:
> using System;

need a copyright header on every code file.

---
Refers to: src/Microsoft.Data/StringColumn.cs:1 in 63bd555. [](commit_id = 63bd555aef1a9cab7753b6d9b36d50abde3eded4, deletion_comment = False)

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    public partial class StringColumn : BaseColumn
+    {
+        internal IList<IList<string>> stringBuffers = new List<IList<string>>(); // To store more than intMax number of strings
```
eerhardt: why is this field internal? Typically fields should be private, and if they need to be exposed, expose them through a property.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,107 @@
...
+            }
+            return curArrayIndex;
+        }
+
+        public override object this[long rowIndex]
+        {
+            get
+            {
+                int bufferIndex = GetBufferIndexContainingRowIndex(ref rowIndex);
+                return stringBuffers[bufferIndex][(int)rowIndex];
```
eerhardt: I still dislike this pattern of casting long to an int. I think `GetBufferIndexContainingRowIndex` should ensure the value is an `int` and return the `int rowIndex` in an `out` parameter.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,107 @@
...
+            set
+            {
+                if (value is string)
+                {
+                    int bufferIndex = GetBufferIndexContainingRowIndex(ref rowIndex);
+                    stringBuffers[bufferIndex][(int)rowIndex] = (string)value;
+                }
+                else
+                {
+                    throw new ArgumentException(nameof(value));
```
eerhardt: This should be a better exception - it should tell the user what about the argument caused the exception - basically that it wasn't a string.

src/Microsoft.Data/StringColumn.cs
eerhardt: instance fields should start with an underscore in their name. `_stringBuffers`.

src/Microsoft.Data/StringColumn.BinaryOperations.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+            return ret;
+        }
+
+        public override BaseColumn Equals(BaseColumn column)
+        {
+            // TODO: Using indexing is VERY inefficient here. Each indexer call will find the "right" buffer and then return the value
+            if (Length != column.Length)
+            {
+                throw new ArgumentException($"Column lengths are mismatched", nameof(column));
```
eerhardt: Should we be adding the exception messages (and any other localizable strings) to a .resx file from the start? It will be easier to add them now, then go through all the strings and try to do it after the fact.

src/Microsoft.Data/StringColumn.BinaryOperations.cs
```diff
@@ -0,0 +1,103 @@
...
+            }
+            PrimitiveColumn<bool> ret = new PrimitiveColumn<bool>(Name);
```
eerhardt: Should we initialize the capacity of this column to `Length`, we know how big it should be from the start.

eerhardt: Apply this in all cases where we know the capacity up front.

---
In reply to: [278586504](https://github.com/dotnet/corefxlab/pull/2660#discussion_r278586504) [](ancestors = 278586504)

src/Microsoft.Data/StringColumn.BinaryOperations.cs
```diff
@@ -0,0 +1,103 @@
...
+
+        public override BaseColumn Equals<T>(T value)
+        {
+            PrimitiveColumn<bool> ret = new PrimitiveColumn<bool>(Name);
+            string valString = value.ToString();
+            for (int i = 0; i < stringBuffers.Count; i++)
+            {
+                IList<string> buffer = stringBuffers[i];
+                int bufferLen = buffer.Count;
+                for (int j = 0; j < bufferLen; j++)
```
eerhardt: Why does the `Equals` method above use a single `for` loop over `Length` and the index, but here we are doing 2 `for` loops over the `stringBuffers`?

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,107 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    public partial class StringColumn : BaseColumn
```
eerhardt: public classes should at least have an xml `summary` doc comment explaining what they are at a high level.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -0,0 +1,362 @@
...
+            BaseColumn ulongColumn = new PrimitiveColumn<ulong>("Ulong", Enumerable.Range(0, length).Select(x => (ulong)x));
+            BaseColumn ushortColumn = new PrimitiveColumn<ushort>("Ushort", Enumerable.Range(0, length).Select(x => (ushort)x));
+
+            DataFrame dataFrame = new DataFrame(new List<BaseColumn> { byteColumn, charColumn, decimalColumn, doubleColumn, floatColumn, intColumn, longColumn, sbyteColumn, shortColumn, uintColumn, ulongColumn, ushortColumn});
+
+            return dataFrame;
+        }
+
+        [Fact]
+        public void TestIndexer()
```
eerhardt: We need to write some more tests. I don't think these tests cover the amount of code that is in the library.

src/Microsoft.Data/PrimitiveColumnComputations.cs
```diff
@@ -0,0 +1,1864 @@
...
+        void Min(PrimitiveColumnContainer<T> column, out T ret);
+        void Prod(PrimitiveColumnContainer<T> column, out T ret);
+        void Sum(PrimitiveColumnContainer<T> column, out T ret);
+        void Round(PrimitiveColumnContainer<T> column);
+    }
+
+    internal static class PrimitiveColumnComputation<T>
+        where T : struct
+    {
+        public static IPrimitiveColumnComputation<T> Instance => PrimitiveColumnComputation.GetComputation<T>();
```
eerhardt: This will new up the Computation object every time it is called. That seems wasteful. Instead, we should cache it in a field.

src/Microsoft.Data/PrimitiveColumnComputations.cs
```diff
@@ -0,0 +1,1864 @@
...
+
+namespace Microsoft.Data
+{
+    internal interface IPrimitiveColumnComputation<T>
+        where T : struct
+    {
+        void Abs(PrimitiveColumnContainer<T> column);
+        void All(PrimitiveColumnContainer<T> column, out bool ret);
+        void Any(PrimitiveColumnContainer<T> column, out bool ret);
+        void Cummax(PrimitiveColumnContainer<T> column);
```
eerhardt: I think we should spell out these words - or at least make them more ".NET"-ish.

`CumulativeMax` would be a better name IMO.

src/Microsoft.Data/PrimitiveColumnComputations.cs
```diff
@@ -0,0 +1,1864 @@
...
+        void Cummin(PrimitiveColumnContainer<T> column);
+        void Cumprod(PrimitiveColumnContainer<T> column);
+        void Cumsum(PrimitiveColumnContainer<T> column);
+        void Max(PrimitiveColumnContainer<T> column, out T ret);
+        void Min(PrimitiveColumnContainer<T> column, out T ret);
+        void Prod(PrimitiveColumnContainer<T> column, out T ret);
```
eerhardt: `Prod` => `Product`

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -0,0 +1,110 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// A column to hold primitive values such as int, float etc. Other value types are not really supported
```
eerhardt: What does `Other value types are not really supported` mean?

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -0,0 +1,110 @@
...
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public partial class PrimitiveColumn<T> : BaseColumn
+        where T : struct
```
eerhardt: Do you think we can use `where T : unmanaged` instead?

[https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/where-generic-type-constraint](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/where-generic-type-constraint)

> The unmanaged constraint limits the type parameter to types known as unmanaged types. An unmanaged type is a type that isn't a reference type and doesn't contain reference type fields at any level of nesting. The unmanaged constraint makes it easier to write low-level interop code in C#. This constraint enables reusable routines across all unmanaged types.

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -0,0 +1,110 @@
...
+            {
+                if (startIndex > Length )
+                {
+                    throw new ArgumentException($"Indexer arguments exceed Length {Length} of the Column");
+                }
+                return _columnContainer[startIndex, length];
+            }
+        }
+
+        // This method involves boxing
```
eerhardt: Do we want to have `T` based indexers as well? That way we can avoid boxing in certain situations.

eerhardt:
> <#@ template debug="false" hostspecific="false" language="C#" #>

What's the difference between this file and `PrimitiveColumnComputations.tt`?

---
Refers to: src/Microsoft.Data/PrimitiveColumn.Computations.tt:1 in 63bd555. [](commit_id = 63bd555aef1a9cab7753b6d9b36d50abde3eded4, deletion_comment = False)

eerhardt:
>         #region Binary Operations

Do we need this region? Everything in this file is a binary operation.

---
Refers to: src/Microsoft.Data/PrimitiveColumn.BinaryOperations.cs:16 in 63bd555. [](commit_id = 63bd555aef1a9cab7753b6d9b36d50abde3eded4, deletion_comment = False)

eerhardt:
>                 case Type boolType when boolType == typeof(bool):

You can just switch on `column` and use `case PrimitiveColumn<bool> boolColumn`, can't you?

If not, you should be able to use pattern matching by simply saying `if (column is PrimitiveColumn<bool> boolColumn)`

---
Refers to: src/Microsoft.Data/PrimitiveColumn.BinaryOperations.cs:21 in 63bd555. [](commit_id = 63bd555aef1a9cab7753b6d9b36d50abde3eded4, deletion_comment = False)

eerhardt:
>             switch (typeof(T))

Another possible pattern that could be used here is the same approach that ML.NET uses for conversions.

[https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Data/Conversion.cs#L65-L67](https://github.com/dotnet/machinelearning/blob/master/src/Microsoft.ML.Data/Data/Conversion.cs#L65-L67)

Where you have a map of `(Type, Type)` tuples that map to Delegates that perform the function. So here we would have a delegate that would take in 2 `int`s and divide them. Then another that would take an `int` and a `short` and divide them. Two `double`s. A `double` and a `float`, etc. etc.

---
Refers to: src/Microsoft.Data/PrimitiveColumn.BinaryOperations.cs:985 in 63bd555. [](commit_id = 63bd555aef1a9cab7753b6d9b36d50abde3eded4, deletion_comment = False)

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -24,10 +24,10 @@ internal class DataFrameTable
 
         public DataFrameTable()
         {
-            _columns = new List<BaseDataFrameColumn>();
+            _columns = new List<BaseColumn>();
         }
 
-        private DataFrameTable(IList<BaseDataFrameColumn> columns)
+        public DataFrameTable(IList<BaseColumn> columns)
```
eerhardt: Why is this now public?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -92,5 +111,10 @@ public IList<IList<object>> Tail(int numberOfRows)
         }
         // TODO: Add strongly typed versions of these APIs
         #endregion
+
+        public override bool Equals(object obj)
+        {
+            return ReferenceEquals(this, obj);
```
eerhardt: I don't believe this is necessary. This Equals override can be removed.

eerhardt:
Is the `ColumnArithmeticTemplate.cs` file necessary? It is empty.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -9,14 +9,15 @@
 namespace Microsoft.Data
 {
     /// <summary>
-    /// The base column type. All APIs should have atleast a stub here first
+    /// The base column type. All APIs should have a stub here first
```
eerhardt: What does `All APIs should have a stub here first` mean?

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -34,8 +35,23 @@ protected set
 
         public string Name;
 
-        public virtual object this[long rowIndex] { get { throw new NotImplementedException(); } set { throw new NotImplementedException(); } }
+        public readonly Type DataType;
```
eerhardt: This should be a property. All these exposed fields should be properties instead - e.g. Name.

In general, values are exposed as properties in C#, not fields. Only on rare occasions are fields exposed outside of a class or struct.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -34,8 +35,23 @@ protected set
...
+            set => throw new NotImplementedException();
+        }
+
+        public virtual object this[long startIndex, int length]
+        {
+            get => throw new NotImplementedException();
+        }
+
+
+        public override bool Equals(object obj)
```
eerhardt: I don't think this is necessary.

corefxlab.sln
```diff
@@ -112,7 +112,7 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Numerics.Experimenta
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.DataFrame", "src\Microsoft.Data\Microsoft.Data.DataFrame.csproj", "{AD22AAD4-FCB0-4DB4-BC38-AB6ACD153899}"
 EndProject
-Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.DataFrame.Tests", "tests\Microsoft.Data.Tests\Microsoft.Data.DataFrame.Tests.csproj", "{485FC567-4AEC-4335-B767-283173F64C42}"
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Data.DataFrame.Tests", "tests\Microsoft.Data.DataFrame.Tests\Microsoft.Data.DataFrame.Tests.csproj", "{9C354CA7-5CDF-44F9-A884-E5840736DF36}"
```
eerhardt: Why the change in GUID here?

`FAE04EC0-301F-11D3-BF4B-00C04F79EFBC` is the "old style" project GUID, I believe. This should stay as `9A19103F-16F7-4668-BE54-9A1E7A4F7556`

src/Microsoft.Data/BaseColumn.cs
danmosemsft: I actually find it tough to think of any situation where we would expose a field from a class.

src/Microsoft.Data/DataFrameTable.cs
danmosemsft: What criteria do we use to choose between public and internal visibility on members, on a public class? Assuming the member is not impelmenting an interface.

src/Microsoft.Data/PrimitiveColumnComputations.cs
```diff
@@ -0,0 +1,1864 @@
...
+                var buffer = column.Buffers[bb];
+                for (int i = 0; i < buffer.Length; i++)
+                {
+                    ret = (sbyte)(buffer.Span[i] * ret);
+                }
+            }
+        }
+        public void Sum(PrimitiveColumnContainer<sbyte> column, out sbyte ret)
+        {
+
```
danmosemsft: Remove blank lines after the {

src/Microsoft.Data/PrimitiveColumnComputations.cs
```diff
@@ -0,0 +1,1864 @@
...
+        }
+        public void All(PrimitiveColumnContainer<ushort> column, out bool ret)
+        {
+            throw new NotSupportedException();
+        }
+        public void Any(PrimitiveColumnContainer<ushort> column, out bool ret)
+        {
+            throw new NotSupportedException();
+        }
+        public void Cummax(PrimitiveColumnContainer<ushort> column)
```
danmosemsft: Nit, blank lines between methods

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,107 @@
...
+                    bufferIndex++;
+                    startIndex = 0;
+                }
+                return ret;
+            }
+        }
+
+        public StringColumn Clone()
+        {
+            StringColumn ret = new StringColumn(Name, Length);
```
danmosemsft: you can use `var`.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,107 @@
...
+
+        public override object this[long startIndex, int length]
+        {
+            get
+            {
+                var ret = new List<string>();
+                int bufferIndex = GetBufferIndexContainingRowIndex(ref startIndex);
+                while (ret.Count < length && bufferIndex < stringBuffers.Count)
+                {
+                    for (int i = (int)startIndex; ret.Count < length && i < stringBuffers[bufferIndex].Count; i++)
```
danmosemsft: what if `startIndex` is greater than `int.MaxValue` ? The cast will wrap. And if it can never be, why is it a long?

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -160,5 +170,41 @@ internal PrimitiveDataFrameColumnContainer<bool> CreateBoolContainerForCompareOp
...
+        {
+            var ret = new PrimitiveColumnContainer<decimal>();
+            foreach(var buffer in Buffers)
+            {
+                ret.Length += buffer.Length;
+                DataFrameBuffer<decimal> newBuffer = new DataFrameBuffer<decimal>();
+                ret.Buffers.Add(newBuffer);
+                newBuffer.EnsureCapacity(buffer.Length);
+                var span = buffer.Span;
+                for (int ii = 0; ii < buffer.Length; ii++)
```
danmosemsft: More of your `ii`' s going on.. 🙀

src/Microsoft.Data/PrimitiveColumnArithmetic.cs
```diff
@@ -4,155 +4,157 @@
 // The .NET Foundation licenses this file to you under the MIT license.
 // See the LICENSE file in the project root for more information.
 
+// Generated from PrimitiveColumnArithmetic.tt. Do not modify directly
```
danmosemsft: Some of the VS generators have a banner like
```
//------------------------------------------------------------------------------
// <auto-generated>
//     This code was generated by a tool.
//     Runtime Version:2.0.50727.42
//
//     Changes to this file may cause incorrect behavior and will be lost if
//     the code is regenerated.
// </auto-generated>
//------------------------------------------------------------------------------
```
I vaguely remember that the `<autogenerated>` part has some effect - on VS maybe - to prevent you writing to it? If so it is nice to put it in this form. I don't know though.

src/Microsoft.Data/PrimitiveColumnContainer.cs
ahsonkhan: ii -> eye eye :p

src/Microsoft.Data/DataFrame.cs
pgovind: Deleted

src/Microsoft.Data/DataFrameTable.cs
pgovind: Good question: In this case it doesn't matter if this constructor public or internal. I just chose public, there was no particular reason to do so.

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: I wanted to say that only primitive types are really supported. A user defined struct for example will throw if some binary op is called. I removed the last sentence and changed it to now say "A column to hold primitive types such as int, float etc"

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: First thing I tried. I don't think generic indexers are supported in C# yet. There's an issue for it here:

https://github.com/dotnet/csharplang/issues/2054

src/Microsoft.Data/StringColumn.BinaryOperations.cs
pgovind: Oh good idea. I'll do that later this week, once I'm done with app building

src/Microsoft.Data/StringColumn.BinaryOperations.cs
pgovind: Done

src/Microsoft.Data/StringColumn.BinaryOperations.cs
pgovind: I wrote that initially thinking the 2 for loop version may be more optimized since I'm avoiding indexing. I changed it now to use the single for loop though. When I fix https://github.com/dotnet/corefxlab/issues/2659 though, the single loop version should be super efficient.

src/Microsoft.Data/StringColumn.cs
pgovind: Changed

pgovind: Done

src/Microsoft.Data/StringColumn.cs
pgovind: startIndex is long. After the call to GetBufferIndexContainingRowIndex, it is guaranteed to be less than int.MaxValue. 

startIndex is long because a column can hold more than int.MaxValue values. A `List` can only hold int.MaxValue items, so `StringColumn` holds a list of lists(called `stringBuffers`) to get around this. 

I will be making a simplifying change soon and change the pattern here in my next patch. I think it is already not possible to have 2 non-maxed out lists in `stringBuffers`. So, the indexer here can be improved to eliminate the call to `GetBufferIndexContaingRowIndex` and this whole business of casting long to int

corefxlab.sln
pgovind: Hmm, I wonder if this happened when I rebased on master cherry-picked a commit. Any idea on how I should go about fixing this? I'm not super good with csproj/sln stuff. If I revert the changes to this file alone, would it break anything else?

src/Microsoft.Data/PrimitiveColumn.cs
eerhardt: `PrimitiveColumn<T>` is already generic. So you can return `T` here. Just like we do in `List<T>`.

https://github.com/dotnet/corefx/blob/a26fd9bb1b17510d51a9b5a2ce7aadc1a5b631ee/src/Common/src/CoreLib/System/Collections/Generic/List.cs#L145-L167

The csharplang feature you referred to is for when you want to have a generic indexer that isn't in a generic class (or the indexer returns a different type than what the class is generic for).

corefxlab.sln
danmosemsft: Just hand edit this sln to revert this one guid (FAE.. to 9A19..). Reopen in VS, if you then hit save on it and VS changes it back - that is odd...

src/Microsoft.Data/PrimitiveColumnArithmetic.cs
danmosemsft: I don't think this matters and you can ignore.

src/Microsoft.Data/PrimitiveColumnComputations.cs
pgovind: Really? Because this is static, shouldn't this new up a computation object once per `T` and reuse that for next time?

eerhardt: >  shouldn't this new up a computation object once per T and reuse that for next time?

I would think that is the behavior we want, but the code you've written doesn't do that.

```C#
public static IPrimitiveColumnComputation<T> Instance => PrimitiveColumnComputation.GetComputation<T>();
```

That is short-hand for:

```C#
public static IPrimitiveColumnComputation<T> Instance
{
    get { return PrimitiveColumnComputation.GetComputation<T>(); }
}
```

Which means every time the property is invoked/called it will call `GetComputation<T>()`, which in turn returns a new object every time it is invoked.

YohDeadfall: `=>` means execute the following code each time. `=` means assignment. You need the last one to initialize the member only once when the static constructor is called.

```suggestion
        public static IPrimitiveColumnComputation<T> Instance = PrimitiveColumnComputation.GetComputation<T>();
```

eerhardt: However, the suggested change is turning `Instance` from a property into a field. And as written above, allows the field to be publicly settable.

In general, we shouldn't be exposing fields publicly unless there is a strong reason for it. We should keep it a read-only property. But it should reuse the same object each time it is invoked, so it doesn't create unnecessary allocations.

YohDeadfall: Yep, fixed. Thanks!

Should this rule be applied for internal classes?

eerhardt: I missed that it was an internal class. That does change the "property or field" decision.

It is probably personal preference, but I generally tend to keep all fields (static or instance) private by default and exposed through properties. And only expose them as fields if there are specific reasons for it.

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: Right, but I'm overriding a virtual method from `BaseColumn` though? Unless I'm not seeing something, I would have to make `BaseColumn`'s indexer generic and then deal with type conversions in the derived class? I thought boxing to return a single value seemed acceptable in this instance

eerhardt: What I am thinking is adding a `new` indexer on the derived class that is typed as `T`.

At first I was thinking:

```C#
    public class BaseColumn
    {
        public virtual object this[long rowIndex]
        {
            get => throw new NotImplementedException();
            set => throw new NotImplementedException();
        }
    }

    public class PrimitiveColumn<T> : BaseColumn
       where T : struct
    {
        public override object this[long rowIndex]
        {
            get => base[rowIndex];
            set => base[rowIndex] = value;
        }

        public new T this[long rowIndex]
        {
            get => default;
            set => base[rowIndex] = value;
        }
    }
```

But that doesn't compile:

> CS0111	Type 'PrimitiveColumn<T>' already defines a member called 'this' with the same parameter types


So instead, we could do:

```C#
    public class BaseColumn
    {
        public object this[long rowIndex]
        {
            get => GetValue(rowIndex);
            set => SetValue(rowIndex, value);
        }

        protected virtual object GetValue(long rowIndex) => throw new NotImplementedException();
        protected virtual void SetValue(long rowIndex, object value) => throw new NotImplementedException();
    }

    public class PrimitiveColumn<T> : BaseColumn
       where T : struct
    {
        protected override object GetValue(long rowIndex)
        {
            return base.GetValue(rowIndex);
        }

        protected override void SetValue(long rowIndex, object value)
        {
            base.SetValue(rowIndex, value);
        }

        public new T this[long rowIndex]
        {
            get => default;
            set => base[rowIndex] = value;
        }
    }
```

This allows us to not box/unbox when the column is casted to a `PrimitiveColumn<T>`.

pgovind:
`PrimitiveColumnComputations.tt` contains the implementations for the computation functions(All, Any, Sum etc). `PrimitiveColumn.Computations.tt` is just the API that calls into the implementations

I like the `switch (typeof(T))` pattern in ML.NET. I'll likely go with that for the element-wise function application APIs

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -0,0 +1,507 @@
...
+using System.Linq;
+using System.Collections.Generic;
+using Microsoft.Data;
+using Xunit;
+
+namespace Microsoft.Data.Tests
+{
+    public class DataFrameTests
+    {
+        static public DataFrame MakeDataFrameWithTwoColumns(int length)
```
eerhardt: (nit) the accessibility comes first

`public static ...`

See number 5 here: [https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md](https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md)

> We always specify the visibility, even if it's the default (e.g. private string _foo not string _foo). Visibility should be the first modifier (e.g. public abstract not abstract public).

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -0,0 +1,507 @@
...
+            df["Double"].Round();
+            Assert.Equal(100.0, df["Double"][0]);
+
+            // Test that none of the numeric column types throw
+            for (int i = 0; i < df.ColumnCount; i++)
+            {
+                BaseColumn column = df.Column(i);
+                if (column.DataType == typeof(bool) || column.DataType == typeof(string))
+                {
+                    continue;
```
eerhardt: Maybe test that these DO throw?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -0,0 +1,507 @@
...
+            dtype = dfm.Column(0).DataType;
+            Assert.True(dtype == typeof(decimal));
+
+            // int + bool should throw
+            Assert.Throws<System.NotSupportedException>(() => df.Add(true));
+
+            var dataFrameColumn1 = new PrimitiveColumn<double>("Double1", Enumerable.Range(0, 10).Select(x => (double)x));
+            df.SetColumn(0, dataFrameColumn1);
+            // Double + comparison ops should throw
+            Assert.Throws<System.NotSupportedException>(() => df.And(true));
```
eerhardt: You should be able to `using System;` at the top and then you don't need the full `System.NotSupportedException` here.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,199 @@
...
+            if (!(mapIndices is null))
+            {
+                if (mapIndices.DataType != typeof(long)) throw new ArgumentException($"Expected sortIndices to be a PrimitiveColumn<long>");
+                if (mapIndices.Length != Length) throw new ArgumentException(strings.MismatchedColumnLengths, nameof(mapIndices));
+                return _Clone(mapIndices as PrimitiveColumn<long>, invertMapIndices);
+            }
+            return _Clone();
+        }
+
+        public StringColumn _Clone(PrimitiveColumn<long> mapIndices = null, bool invertMapIndex = false)
```
eerhardt: Why does this public method start with an underscore? We don't use underscores in public member names.

src/Microsoft.Data/BaseColumn.Computations.cs
```diff
@@ -0,0 +1,78 @@
...
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual bool Any()
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual void CumulativeMax()
```
eerhardt: Do these Cumulative functions not need to return results?

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -0,0 +1,157 @@
...
+            {
+                _columnContainer[rowIndex] = (T)value;
+            }
+            else
+            {
+                throw new ArgumentException(nameof(value));
+            }
+        }
+
+        // This method involves boxing
```
eerhardt: This comment should no longer be true. This shouldn't involve boxing.

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -0,0 +1,157 @@
...
+
+        protected override void SetValue(long rowIndex, object value)
+        {
+            if (value.GetType() == typeof(T))
+            {
+                _columnContainer[rowIndex] = (T)value;
+            }
+            else
+            {
+                throw new ArgumentException(nameof(value));
```
eerhardt: Probably should say what about the argument caused the exception.

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -0,0 +1,157 @@
...
+            _columnContainer = new PrimitiveColumnContainer<T>(values);
+            Length = _columnContainer.Length;
+        }
+
+        public PrimitiveColumn(string name, long length = 0, bool isNullable = true) : base(name, length, typeof(T))
+        {
+            _columnContainer = new PrimitiveColumnContainer<T>(length);
+        }
+
+        public override object this[long startIndex, int length] {
```
eerhardt: (nit) braces should go on a new line.

src/Microsoft.Data/PrimitiveColumnComputations.cs
```diff
@@ -0,0 +1,1924 @@
...
+        }
+
+        public void Round(PrimitiveColumnContainer<double> column)
+        {
+            for (int b = 0; b < column.Buffers.Count; b++)
+            {
+                var buffer = column.Buffers[b];
+                for (int i = 0; i < buffer.Length; i++)
+                {
+                    buffer.Span[i] = (double)(Math.Round((decimal)buffer.Span[i]));
```
eerhardt: why the cast to `decimal` here? Math.Round can take a `double`.

src/Microsoft.Data/PrimitiveColumnComputations.cs
```diff
@@ -0,0 +1,1924 @@
...
+        }
+
+        public void Round(PrimitiveColumnContainer<float> column)
+        {
+            for (int b = 0; b < column.Buffers.Count; b++)
+            {
+                var buffer = column.Buffers[b];
+                for (int i = 0; i < buffer.Length; i++)
+                {
+                    buffer.Span[i] = (float)(Math.Round((decimal)buffer.Span[i]));
```
eerhardt: We should either be using MathF here, or casting to a `double` instead of decimal.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -0,0 +1,338 @@
...
+        /// <param name="mapIndices"></param>
+        /// <param name="invertMapIndices"></param>
+        /// <returns></returns>
+        public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
+
+        public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
+
+        public virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
+
+        protected delegate long GetBufferSortIndex(int bufferIndex, int sortIndex);
```
eerhardt: You may want to make all these `internal`, as I don't think we'd want this implementation detail leaking in the public contract of this class.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -0,0 +1,338 @@
...
+            n >>= 2;
+            while (n > 0)
+            {
+                ++result;
+                n >>= 1;
+            }
+            return result;
+        }
+
+        protected void IntrospectiveSort<T>(
```
eerhardt: `internal`. More implementation detail we don't want in the external API.

eerhardt: Why isn't this static?

---
In reply to: [283058030](https://github.com/dotnet/corefxlab/pull/2660#discussion_r283058030) [](ancestors = 283058030)

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,199 @@
...
+    public partial class StringColumn : BaseColumn
+    {
+        private List<List<string>> _stringBuffers = new List<List<string>>(); // To store more than intMax number of strings
+
+        public StringColumn(string name, long length) : base(name, length, typeof(string))
+        {
+            _stringBuffers.Add(new List<string>());
+        }
+
+        public StringColumn(string name, IEnumerable<string> values) : base(name, 0, typeof(string)) {
```
eerhardt: braces go on a new line.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -0,0 +1,338 @@
...
+            }
+        }
+
+        public long NullCount { get; protected set; }
+
+        public string Name { get; set; }
+
+        public Type DataType { get; }
+
+        public virtual object this[long rowIndex]
```
eerhardt: This shouldn't be `virtual`.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,199 @@
...
+            int numBuffers = _stringBuffers.Count;
+            while (curArrayIndex < numBuffers && rowIndex > _stringBuffers[curArrayIndex].Count)
+            {
+                rowIndex -= _stringBuffers[curArrayIndex].Count;
+                curArrayIndex++;
+            }
+            return curArrayIndex;
+        }
+
+        public override object this[long rowIndex]
```
eerhardt: Why don't we have a `string` typed indexer like we do for `PrimitiveColumn<T>`?

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,199 @@
...
+            return Clone(columnSortIndices, !ascending);
+        }
+
+        public override BaseColumn GetAscendingSortIndices()
+        {
+            _GetSortIndices(Comparer<string>.Default, out PrimitiveColumn<long> columnSortIndices);
+            return columnSortIndices;
+
+        }
+        private void _GetSortIndices(Comparer<string> comparer, out PrimitiveColumn<long> columnSortIndices)
```
eerhardt: (nit) even private methods shouldn't have underscores in their names - especially at the beginning.

src/Microsoft.Data/StringColumn.BinaryOperations.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    public partial class StringColumn : BaseColumn
+    {
+        public override BaseColumn Add(BaseColumn column)
+        {
+            // TODO: Using indexing is VERY inefficient here. Each indexer call will find the "right" buffer and then return the value
```
eerhardt: Can we log issues for TODOs? What is the plan to address them?

src/Microsoft.Data/StringColumn.BinaryOperations.cs
```diff
@@ -0,0 +1,98 @@
...
+using System.Collections.Generic;
+using System.Globalization;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    public partial class StringColumn : BaseColumn
+    {
+        public override BaseColumn Add(BaseColumn column)
```
eerhardt: Do we want strongly typed versions of these methods? That way if I have a `StringColumn`, and I call one of these methods, I should get back a `StringColumn`, not?

src/Microsoft.Data/DecimalConverter.cs
```diff
@@ -0,0 +1,173 @@
...
+{
+    internal interface IDecimalConverter<T>
+        where T : struct
+    {
+        decimal GetDecimal(T value);
+    }
+    internal static class DecimalConverter<T>
+        where T : struct
+    {
+        public static IDecimalConverter<T> Instance => DecimalConverter.GetDecimalConverter<T>();
```
eerhardt: This and DoubleConverter suffer from the same pattern that `PrimitiveColumnArithmetic` did - every time you invoke this property it allocates a new object.

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -69,46 +98,74 @@
...
+    <Compile Update="strings.Designer.cs">
       <DesignTime>True</DesignTime>
       <AutoGen>True</AutoGen>
-      <DependentUpon>PrimitiveDataFrameColumnContainer.BinaryOperations.tt</DependentUpon>
+      <DependentUpon>strings.resx</DependentUpon>
     </Compile>
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="strings.resx">
```
eerhardt: Any reason `strings.resx` is lower cased? Why not `Strings.resx`? It seems more usual.

src/Microsoft.Data/PrimitiveColumnArithmetic.tt
```diff
@@ -5,50 +5,52 @@
...
+       <#= method.GetMethodSignature("PrimitiveColumnContainer", "T")#>;
 <# } #>
     }
 
-    internal static class PrimitiveDataFrameColumnArithmetic<T>
+    internal static class PrimitiveColumnArithmetic<T>
         where T : struct
     {
-        public static IPrimitiveDataFrameColumnArithmetic<T> Instance => PrimitiveDataFrameColumnArithmetic.GetArithmetic<T>();
+        public static IPrimitiveColumnArithmetic<T> Instance = PrimitiveColumnArithmetic.GetArithmetic<T>();
```
eerhardt: How come this one is a field, but `PrimitiveColumnComputations` is a property?

ahsonkhan:
FYI, this build is failing. GitHub just isn't showing the results properly from the CI legs: https://dev.azure.com/dnceng/public/_build/results?buildId=183617

src/Microsoft.Data/StringColumn.cs
pgovind: Oversight while refactoring. Fixed

src/Microsoft.Data/BaseColumn.Computations.cs
pgovind: Pandas does not, not that we need to follow Pandas. Arguably this is inconsistent with Max(), which returns the Max value in the column. I would rather consider this [here](https://github.com/dotnet/corefxlab/issues/2680)

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: Checked in an editorconfig file with the latest 2 commits. Shouldn't happen again

src/Microsoft.Data/PrimitiveColumnComputations.cs
pgovind: Math.Round(int) was ambiguous since Round takes both decimal and double. This file is template generated and I wanted to keep the template file simple, so I just cast every type to decimal.

src/Microsoft.Data/PrimitiveColumnComputations.cs
pgovind: Similar here. I wanted to keep the template file simple

src/Microsoft.Data/BaseColumn.cs
pgovind: Fixed

src/Microsoft.Data/StringColumn.cs
pgovind: Fixed

src/Microsoft.Data/StringColumn.cs
pgovind: Fixed. Had to change PrimitiveColumn.BinaryOperations.cs too.

src/Microsoft.Data/StringColumn.BinaryOperations.cs
pgovind: Filed [this](https://github.com/dotnet/corefxlab/issues/2681) to address this

src/Microsoft.Data/PrimitiveColumnComputations.cs
eerhardt: Can you file an issue for this?  I don't think we should be doing the cast to decimal, as there could be some precision/truncation issues/changes.

src/Microsoft.Data/StringColumn.BinaryOperations.cs
pgovind: Hmmm, so at the moment, I've taken the less strongly typed approach and since one of the main use cases is notebooks, it pairs well with returning a copy everytime(since I don't know the return type until runtime). In the case of a `StringColumn`, `Add` will always returns a `StringColumn`, but for `intColumn + 5.0`, the result will be a `PrimitiveColumn<double>` for example. To enable the strong typed versions, I would have to define APIs for all possible combinations of all columns. How about I consider that in a separate issue? Since these methods are general, it's easy enough to build the strong APIs on top of this?

src/Microsoft.Data/DecimalConverter.cs
pgovind: Fixed

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
pgovind: Fixed

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -0,0 +1,347 @@
...
+        }
+
+        protected virtual object GetValue(long rowIndex) => throw new NotImplementedException();
+        protected virtual object GetValue(long startIndex, int length) => throw new NotImplementedException();
+
+        protected virtual void SetValue(long rowIndex, object value) => throw new NotImplementedException();
+
+        public object this[long startIndex, int length]
+        {
+            get => GetValue(startIndex, length);
```
pgovind: Just as a note: On a notebook, most calls such as the following will still box at the moment:
`
var intColumn = df.Column('intColumn'); //intColumn is a PrimitiveColumn<int>, but is returned as a BaseColumn
var thisWillBox = intColumn[2];
`

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -20,37 +29,57 @@
...
     </None>
     <None Update="PrimitiveColumn.BinaryOperations.tt">
       <Generator>TextTemplatingFileGenerator</Generator>
       <LastGenOutput>PrimitiveColumn.BinaryOperations.cs</LastGenOutput>
     </None>
+    <None Update="PrimitiveColumn.Computations.tt">
+      <Generator>TextTemplatingFileGenerator</Generator>
+      <LastGenOutput>PrimitiveColumn.Computations.cs</LastGenOutput>
+    </None>
+    <None Update="PrimitiveColumnArithmetic - Copy.tt">
```
pgovind: Delete

src/Microsoft.Data/PrimitiveColumn.Sort.cs
```diff
@@ -0,0 +1,61 @@
...
+    {
+        public override BaseColumn Sort(bool ascending = true)
+        {
+            PrimitiveColumn<long> sortIndices = GetAscendingSortIndices() as PrimitiveColumn<long>;
+            return _Clone(sortIndices, !ascending);
+        }
+
+        internal override BaseColumn GetAscendingSortIndices()
+        {
+            // Is Comparer<T>.Default guaranteed to sort in ascending order?
```
pgovind: @eerhardt

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -0,0 +1,158 @@
...
+        }
+
+        public PrimitiveColumn(string name, long length = 0, bool isNullable = true) : base(name, length, typeof(T))
+        {
+            _columnContainer = new PrimitiveColumnContainer<T>(length);
+        }
+
+        public new T this[long startIndex, int length]
+        {
+            get => (T)GetValue(startIndex, length);
```
pgovind: This will still box. Need to replace this with the code inside GetValue itself

src/Microsoft.Data/StringColumn.BinaryOperations.cs
eerhardt: Yes, a separate issue for this sounds good to me.

As for every combination of type, I don't think I'd go that far.  But for the case where I have a strongly-typed `string` or `Int32` column, if I add a `string` or `Int32` to that column, I know I get back a `string` or `Int32`.

The analogy here is this code:

```C#
string a = "foo";
string b = "bar";

??? c = a + b;
````

It's pretty obvious that `c` should be a `string`, not `object`.

File: src/Microsoft.Data/PrimitiveColumn.Sort.cs (hunk 50 above)
Comment (eerhardt) [reply src/Microsoft.Data/PrimitiveColumn.Sort.cs]: This is guaranteed for ALL `IComparer` implementations.

See https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.icomparer-1.compare?view=netframework-4.8

> Returns: A signed integer that indicates the relative values of x and y, as shown in the following table.

Value | Meaning
-- | --
Less than zero | x is less than y.
Zero | x equals y.
Greater than zero | x is greater than y.

File: src/Microsoft.Data/PrimitiveColumn.cs (hunk 51 above)
Comment (eerhardt) [reply src/Microsoft.Data/PrimitiveColumn.cs]: Yes - that is the point of these "new" strongly typed members. No boxing should be necessary.

File: src/Microsoft.Data/StringColumn.BinaryOperations.cs (hunk 44 above)
Comment (pgovind) [reply src/Microsoft.Data/StringColumn.BinaryOperations.cs]: Filed [here](https://github.com/dotnet/corefxlab/issues/2684)

File: src/Microsoft.Data/PrimitiveColumn.Sort.cs (hunk 50 above)
Comment (eerhardt) [reply src/Microsoft.Data/PrimitiveColumn.Sort.cs]: You can remove this comment (in the next PR if you want)

---
In reply to: [284780397](https://github.com/dotnet/corefxlab/pull/2660#discussion_r284780397) [](ancestors = 284780397)

File: src/Microsoft.Data/DataFrame.cs [hunk 52]
```diff
@@ -92,5 +113,55 @@ public IList<IList<object>> Tail(int numberOfRows)
...
+            {
+                newColumns.Add(Column(i).Clone());
+            }
+            return new DataFrame(newColumns);
+        }
+
+        public DataFrame Sort(string columnName, bool ascending = true)
+        {
+            BaseColumn column = this[columnName];
+            BaseColumn sortIndices = column.GetAscendingSortIndices();
```
eerhardt: we should only call public BaseColumn methods so this works with external derived classes. As it exists today -  this will throw an exception because there is no way for an external column to override an internal method.

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -69,46 +94,74 @@
...
+    <Compile Update="Strings.Designer.cs">
       <DesignTime>True</DesignTime>
       <AutoGen>True</AutoGen>
-      <DependentUpon>PrimitiveDataFrameColumnContainer.BinaryOperations.tt</DependentUpon>
+      <DependentUpon>Strings.resx</DependentUpon>
     </Compile>
   </ItemGroup>
 
+  <ItemGroup>
+    <EmbeddedResource Update="Strings.resx">
```
eerhardt: I think the casing needs to match the actual file casing - is it possible to rename the file?

See [https://stackoverflow.com/questions/1793735/change-case-of-a-file-on-windows](https://stackoverflow.com/questions/1793735/change-case-of-a-file-on-windows) for how to do this on Windows.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,225 @@
...
+        protected override void SetValue(long rowIndex, object value)
+        {
+            if (value is string)
+            {
+                int bufferIndex = GetBufferIndexContainingRowIndex(ref rowIndex);
+                _stringBuffers[bufferIndex][(int)rowIndex] = (string)value;
+            }
+            else
+            {
+                throw new ArgumentException(Strings.MismatchedValueType + " string", nameof(value));
```
eerhardt: When throwing exceptions with a  message that comes from a resource, you shouldn't concat strings. Instead use formatted strings - `{0}`. The reason is because when translating to other languages, the hard-coded type name may need to move in the sentence to make sense in the language.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -0,0 +1,225 @@
...
+            }
+        }
+
+        public new string this[long rowIndex]
+        {
+            get => (string)GetValue(rowIndex);
```
eerhardt: Is it possible to implement this without the cast?

PR #2662, opened 2019-04-24
Move CI builds to azure devops

safern:
cc: @ahsonkhan @chcosta


safern:
> Seems simple enough that it could just be one file rather than using a template. Your call though.

Once I add official builds I will only need the windows job and some custom steps. That is why I split it, to avoid much duplication.

PR #2663, opened 2019-04-24
Add official build support into yml file

safern:


.azure-pipelines.yml
```diff
@@ -13,21 +13,29 @@ jobs:
...
-      name: Hosted VS2017
 
-- template: /pipelines/corefxlab-base.yml
-  parameters:
-    name: MacOS
-    displayName: Innerloop OSX
-    targetOS: OSX
-    pool:
-      name: Hosted macOS
+    ${{ if and(ne(variables['System.TeamProject'], 'public'), notin(variables['Build.Reason'], 'PullRequest')) }}:
```
ahsonkhan: nit: Is there a way to save such a condition "locally" in yml and re-use it/!it (not it) to avoid duplication/typos between the negation?

pipelines/corefxlab-base.yml
```diff
@@ -3,25 +3,58 @@ parameters:
...
           _BuildConfig: Release
 
     steps:
-      - script: $(_buildScript) $(_BuildConfig)
+      - ${{ if eq(parameters.isOfficialBuild, 'true') }}:
+        - task: MicroBuildSigningPlugin@2
+          displayName: Install MicroBuild plugin
+          inputs:
+            signType: $(_SignType)
+            zipSources: false
```
ahsonkhan: **Question:** What does this do?

pipelines/corefxlab-base.yml
```diff
@@ -3,25 +3,58 @@ parameters:
...
+            feedSource: https://dnceng.pkgs.visualstudio.com/_packaging/MicroBuildToolset/nuget/v3/index.json
+          env:
+            TeamName: $(_TeamName)
+          condition: and(in(variables['_SignType'], 'real', 'test'), eq(variables['Agent.Os'], 'Windows_NT'))
+
+      - script: $(_buildScript) $(_buildArguments)
         displayName: Build and Test
```
ahsonkhan: Should we specify in the display name whether its official or not, or do you think its not necessary?

pipelines/corefxlab-base.yml
safern: Do you mean the zipSources parameter? I have no idea, I just copied what arcade does.

.azure-pipelines.yml
safern: I heard they were going to add support through importing a file that defines constants, but I haven't heard of that support being added, that is why we need to pass it down as a parameter to the template, that is the only way I know of caching the value.

cc: @chcosta do you know if this is now supported?

pipelines/corefxlab-base.yml
safern: No, I think it makes sense to keep it like this for both builds since we build and test on both builds.

safern:
> Looks good. Is removing this next on the docket?
> https://github.com/dotnet/dotnet-ci/blob/master/data/repolist.txt#L53

Correct, also deleting netci.groovy from the repo.

safern:
Merging as the build shape looks good. The only thing pending is to get signing approval in the new build definition.

PR #2666, opened 2019-04-25
Fix build number format for correct package versions

safern:
cc: @ahsonkhan

### DIFF
--- a/.azure-pipelines.yml
+++ b/.azure-pipelines.yml
@@ -1,3 +1,7 @@
+# This will assign the build version on this format.
+# The package version we produce depend on this number.
+name: $(Date:eyyMMdd)$(Rev:-r)
+
 trigger:
   batch: true
   branches:


.azure-pipelines.yml
```diff
@@ -1,3 +1,7 @@
+# This will assign the build version on this format.
+# The package version we produce depend on this number.
+name: $(Date:eyyMMdd)$(Rev:-r)
```
ahsonkhan: Will the major.minor version remain in-tact (`0.1.1`)?

Essentially, will this be: `0.1.1-eyyMMdd-r`?

ahsonkhan: Also, where are we using this name?

safern: > Will the major.minor version remain in-tact (0.1.1)?

Yes, that is defined by the package projects.

> Also, where are we using this name?

This is used by azure pipelines to define the BuildNumber.

safern: https://docs.microsoft.com/en-us/azure/devops/pipelines/yaml-schema?view=azure-devops&tabs=example#pipeline

ahsonkhan: `name: $(Date:yyyyMMdd).$(Rev:.r)`

It looks like they placed two dots `.` (one before `Rev` and one before `r`). Do we need a dot/hyphen between the eyyMMdd and Rev as well?

safern: No, if we put those dots then the version would be

`1.1.0-.eyyMMdd.-r` the dots are just separating the concatenation. `$()` encloses a variable and were just concaneting 2 of those without any separators. We could change it to:
`$(Date:eyyMMdd)-$(Rev:r)` if that makes it more clear.

ahsonkhan: > `1.1.0-.eyyMMdd.-r`

By that logic, wouldn't the example shown in the link result in 2 dots?
`yyyyMMdd..r`

safern: Correct. That seems like an error in their example itself 😄 

Validated with today's official build and it produced the right version as we want it:

```
Pushing package F:\workspace\_work\1\s\scripts\..\packages\System.Time.0.1.1-e190426-1.nupkg to MyGet...
Pushing System.Time.0.1.1-e190426-1.nupkg to 'https://dotnet.myget.org/F/dotnet-corefxlab/api/v2/package'...
  PUT https://dotnet.myget.org/F/dotnet-corefxlab/api/v2/package/
  Created https://dotnet.myget.org/F/dotnet-corefxlab/api/v2/package/ 8120ms
Your package was pushed.
```

ahsonkhan: > Validated with today's official build and it produced the right version as we want it

Great.

> That seems like an error in their example itself

That makes a lot more sense then :)

PR #2673, opened 2019-05-10
Implement conversion operators for Half-precision floating point data type (IEEE 754 binary16)

Gnbrkm41:
Related: #2635, dotnet/corefx#25702, dotnet/corefx#17267

This PR includes various conversion operators from primitive numeric types to Half-precision FP type and the other way around, ported from [Berkeley SoftFloat library Release 3e](https://github.com/ucb-bar/berkeley-softfloat-3). https://github.com/dotnet/corefxlab/blob/d260677424aea4f5c4294d9bd6f8640a47a4f3a7/src/System.Numerics.Experimental/Half.cs#L13

In addition, this also includes unary sign operators (`+num, -num`) as well. 

cc @tannergooding


Gnbrkm41:
The reason I'm leaving this as a draft PR is because I'm using members from `BitOperations` class from Core 3.0, this will not build properly without modifications to the build script (Results in `BitOperations does not exist in the current context...`). Commenting out this line did appear to allow the use of the type, though it is a temporary solution. https://github.com/dotnet/corefxlab/blob/master/scripts/build.ps1#L52

However, even after that, it appears that types from `System.Text.JsonLab.Serialization` and `System.Text.Json.Serialization` are clashing, causing the build on my local PC to fail, with errors shown below.
```
Error	CS0436	The type 'JsonSerializerOptions' in 'C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializerOptions.cs' conflicts with the imported type 'JsonSerializerOptions' in 'System.Text.Json, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'. Using the type defined in 'C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializerOptions.cs'.	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Read.cs	16	Active
Error	CS1739	The best overload for 'Flush' does not have a parameter named 'isFinalBlock'	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Write.cs	32	Active
Error	CS1061	'Utf8JsonWriter' does not contain a definition for 'GetCurrentState' and no accessible extension method 'GetCurrentState' accepting a first argument of type 'Utf8JsonWriter' could be found (are you missing a using directive or an assembly reference?)	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Write.cs	33	Active
Error	CS1501	No overload for method 'Flush' takes 1 arguments	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Write.Helpers.cs	37	Active
Error	CS1061	'Utf8JsonWriter' does not contain a definition for 'BytesWritten' and no accessible extension method 'BytesWritten' accepting a first argument of type 'Utf8JsonWriter' could be found (are you missing a using directive or an assembly reference?)	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Write.cs	63	Active
Error	CS1739	The best overload for 'Flush' does not have a parameter named 'isFinalBlock'	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Write.Helpers.cs	77	Active
Error	CS0246	The type or namespace name 'JsonWriterState' could not be found (are you missing a using directive or an assembly reference?)	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Write.Helpers.cs	46	Active
Error	CS0246	The type or namespace name 'JsonWriterState' could not be found (are you missing a using directive or an assembly reference?)	System.Text.JsonLab.Serialization (netcoreapp3.0)	C:\Users\gotos\Source\Repos\corefxlab\src\System.Text.JsonLab.Serialization\System\Text\Json\Serialization\JsonSerializer.Write.Pipe.cs	37	Active
Error	CS0433	The type 'JsonSerializer' exists in both 'System.Text.JsonLab.Serialization, Version=0.1.1.0, Culture=neutral, PublicKeyToken=adb9793829ddae60' and 'System.Text.Json, Version=4.0.0.0, Culture=neutral, PublicKeyToken=cc7b13ffcd2ddd51'	System.Text.JsonLab.Serialization.Tests	C:\Users\gotos\Source\Repos\corefxlab\tests\System.Text.JsonLab.Serialization.Tests\Array.ReadTests.cs	14	Active
```

Gnbrkm41:
You might notice that there are a few `// TODO:` comments here and there. Part of them is because some of them might require design discussions (e.g. should `int` -> `Half` conversion be implicit, even though it cannot represent the full magnitude of the type?), and some are just because I had a few doubts while porting from C codes.

ahsonkhan:
> However, even after that, it appears that types from `System.Text.JsonLab.Serialization` and `System.Text.Json.Serialization` are clashing, causing the build on my local PC to fail, with errors shown below.

Yep, we should clean up the corefxlab implementation of the serializer now to avoid the clash.

cc @steveharter 

> I'm using members from `BitOperations` class from Core 3.0, this will not build properly without modifications to the build script (Results in `BitOperations does not exist in the current context...`). 

Regarding that, we need to update the version of cli we are targeting: https://github.com/dotnet/corefxlab/blob/master/DotnetCLIVersion.txt#L1

We are still referencing an older version of the SDK because our auto-updater doesn't work with arcade (cc @safern): https://github.com/dotnet/corefxlab/pull/2643

I think we can manually update it to latest, for now, and that should unblock you.

src/System.Numerics.Experimental/Half.cs
```diff
@@ -353,5 +356,402 @@ public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan
...
+        public static implicit operator Half(int i)
+        {
+            bool sign = i < 0;
+            Half h = (uint)(sign ? -i : i); // Math.Abs but doesn't throw exception, because we cast it to uint anyway
+            return sign ? new Half((ushort)(h.m_value | SignMask)) : h;
+        }
+
+        public static implicit operator Half(uint i)
+        {
+            if (i == 0)
```
tannergooding: Would be good to keep this as part of the `if (shiftDist >= 0)` case. Having it here will cause a mispredicted branch for all non-zero inputs and will force 2 branch checks for any non-zero result.

Having it as part of the `shiftDist` check will result in 1 or 2 branch checks, depending on the input and can put `0` on the mispredict path.

src/System.Numerics.Experimental/Half.cs
```diff
@@ -353,5 +356,402 @@ public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan
...
+        // -----------------------Start of to-half conversions-------------------------
+
+        public static implicit operator Half(int i)
+        {
+            bool sign = i < 0;
+            Half h = (uint)(sign ? -i : i); // Math.Abs but doesn't throw exception, because we cast it to uint anyway
+            return sign ? new Half((ushort)(h.m_value | SignMask)) : h;
+        }
+
+        public static implicit operator Half(uint i)
```
tannergooding: nit: would be nice to name these `value` instead of `i`

src/System.Numerics.Experimental/Half.cs
```diff
@@ -353,5 +356,402 @@ public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan
...
+                return new Half(false, (ushort)(0x18 - shiftDist), (ushort)(l << shiftDist));
+
+            shiftDist += 4;
+            ushort sig = (ushort)(shiftDist < 0 ? ShiftRightJam(l, -shiftDist) : l << shiftDist);
+            return new Half(RoundPackToHalf(false, (short)(0x1C - shiftDist), sig));
+        }
+
+        public static implicit operator Half(short s)
+        {
+            return (int) s;
```
tannergooding: note: These could probably be micro-optimized since we know the value will always fit (`byte`/`sbyte`) or that only a limited set of values will overflow (`short`/`ushort`)

src/System.Numerics.Experimental/Half.cs
```diff
@@ -353,5 +356,402 @@ public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan
...
+        {
+            return (uint) b;
+        }
+
+        public static implicit operator Half(sbyte b)
+        {
+            return (int) b;
+        }
+
+        private const uint SingleSignMask = 0x80000000;
```
tannergooding: (thinking out loud) Maybe these should be put in a helper class since they will also be needed for `Binary128` and potentially others...

It might also just be worth exposing them on the actual types, but that needs an API proposal

tannergooding: nit: Would be good to keep these at the top with the other constants

Gnbrkm41: Some methods in the Utility region as well, I guess; the original implementation seemed to share parts of them (e.g. Create*NaN). Definitely worth moving them out into a helper class.

src/System.Numerics.Experimental/Half.cs
```diff
@@ -353,5 +356,402 @@ public bool TryFormat(Span<char> destination, out int charsWritten, ReadOnlySpan
...
+            return h;
+        }
+
+        #region Utilities
+
+        // TODO: Worth bringing the `ShortShiftRightJam`? looks like some perf difference only
+        // If any bits are lost by shifting, "jam" them into the LSB.
+        // if dist > bit count, Will be 1 or 0  depending on i 
+        // (unlike bitwise operators that masks the lower 5 bits)
+        private static uint ShiftRightJam(uint i, int dist)
```
tannergooding: note: this basically looks like a rotate function. but with different functionality if the distance is larger than the source bit count.

tannergooding:
Overall looks good and to match the reference implementation. Probably some code cleanup and micro-optimizing to do (to take advantage of branch predictions/inlining/etc).

Thanks for doing this!

safern:
> We are still referencing an older version of the SDK because our auto-updater doesn't work with arcade (cc @safern): #2643

I think @riarenas was working on making core-sdk consumable. However, we could potentially instead use the Microsoft.NETCore.App package from core-setup. No need to use the SDK just to get new APIs.

Gnbrkm41:
Will open after fixing all tests... soon:tm:

Gnbrkm41:
Oh, I meant once tests pass on local... there was some dumb issues I haven't spotted before (because I couldn't run the tests), one of it being infinite recursion causing StackOverflowException 😅 

Almost done, though.

Gnbrkm41:
*Oh no, I didn't expect the tests to run again!*

I see 34 tests failing on ToDouble (16) / FromDouble (9) / FromSingle (9) tests on any negative numbers (excl. MinValue); Trying to figure out what's wrong with it.

Gnbrkm41:
And the answer is yes, turns out that I was missing a pair of parentheses and a cast....

src/System.Numerics.Experimental/Half.cs
Gnbrkm41: Moved them to `Ieee754Helpers` class with some other methods (and potentially useful ones?)

tannergooding:
@Gnbrkm41, any other changes you were planning on here? If not, I'll give it a final review pass and I think we should be good to merge.

src/System.Numerics.Experimental/Half.cs
```diff
@@ -341,17 +353,317 @@ public override int GetHashCode()
...
+        {
+            bool sign = value < 0;
+            Half h = (uint)(sign ? -value : value); // Math.Abs but doesn't throw exception, because we cast it to uint anyway
+            return sign ? new Half((ushort)(h.m_value | SignMask)) : h;
+        }
+
+        public static implicit operator Half(uint value)
+        {
+            int shiftDist = BitOperations.LeadingZeroCount(value) - 21;
+            if (shiftDist >= 0)
```
tannergooding: nit: Using braces makes the scope much clearer

src/System.Numerics.Experimental/Half.cs
```diff
@@ -341,17 +353,317 @@ public override int GetHashCode()
...
+                return value != 0?
+                    new Half(false, (ushort)(0x18 - shiftDist), (ushort) (value << shiftDist)):
```
tannergooding: nit: The generaly codestyle is no space after the closing parentheses

tannergooding: same goes below for cases like `return (int) value`

src/System.Numerics.Experimental/Half.cs
Gnbrkm41: I'm that 'no brace if possible' guy... but sure. 😁

Do you want me to do it for every `if`s in general, or only for these multi-line statements?

src/System.Numerics.Experimental/Half.cs
Gnbrkm41: Ah, yes. R# had some weird default settings on... thought I've got rid of all of it.

src/System.Numerics.Experimental/Half.cs
tannergooding: The only hard rule (see https://github.com/dotnet/corefx/blob/master/Documentation/coding-guidelines/coding-style.md and quoted below), is:
> A single line statement block can go without braces but the block must be properly indented on its own line and must not be nested in other statement blocks that use braces (See issue 381 for examples).

My personal preference is to always use braces (I think it helps readability and can make diffs smaller when adding code/asserts/etc), but I won't block the PR on it 😄

Gnbrkm41:
@tannergooding Nothing at the moment. Feel free to do so

src/System.Numerics.Experimental/Ieee754Helpers.cs
```diff
@@ -0,0 +1,97 @@
...
+        public const int SingleSignificandShift = 0;
+
+        public const ushort HalfSignMask = 0x8000;
+        public const int HalfSignShift = 15;
+        public const ushort HalfExponentMask = 0x7C00;
+        public const int HalfExponentShift = 10;
+        public const ushort HalfSignificandMask = 0x03FF;
+        public const int HalfSignificandShift = 0;
+
+        public static double CreateDouble(ulong value)
```
Gnbrkm41: @tannergooding, Any opinion on the names of these methods (`Create***(*** value)`, `To****(*** value)`)? They don't really match the names on `BitConverter` class, but I thought it would be okay to omit the type being converted from, because the type can be inferred from the size of the type. It's just an internal class after all (which is another way of 'I was lazy typing all these `BitConverter.***ToInt**Bits`' 😅 ), but I'm happy to change them if you have any objections on it.

tannergooding: I think they are fine for now, especially if the helper is internal.

I still need to open a proposal for exposing `SingleToUInt32Bits`, etc....

Gnbrkm41: That'd be lovely, converting back and forth to signed/unsigned types were pretty annoying!

src/System.Numerics.Experimental/Half.cs
Gnbrkm41: Braces everywhere it is.

Gnbrkm41:
I think that's it unless you have any other opinions on any of these, happy for this to get merged.

That is, of course, if you allow it to be so :^)

src/System.Numerics.Experimental/Half.cs
```diff
@@ -341,17 +353,358 @@ public override int GetHashCode()
...
+        public static implicit operator Half(long value)
+        {
+            bool sign = value < 0;
+            Half h = (ulong)(sign ? -value : value); // Math.Abs but doesn't throw exception, because we cast it to ulong anyway
+            return sign ? new Half((ushort)(h.m_value | SignMask)) : h;
+        }
+
+        public static implicit operator Half(ulong value)
+        {
+            int shiftDist = BitOperations.LeadingZeroCount(value) - 53;
```
tannergooding: Just a note: other than the `shiftDist`, this and the `uint` handling is identical.

Gnbrkm41: Guess I could just redirect the int->Half conversion to long->Half conversion operator, yeah. Would it be worth changing?

Gnbrkm41: ...Or alternatively, redirect the long->Half conversion to int->Half. Max/MinValue is 65504/-65504 anyway, so....

tannergooding: I think it's just a case were we could share code in the future. It doesn't need to be done now.

tannergooding:
Thanks so much for doing this! 🎉

PR #2676, opened 2019-05-14
Update CLI version, remove JSON serializer APIs and rename Encodings to TextEncodings to avoid conflict.

ahsonkhan:
This helps unblock @Gnbrkm41: https://github.com/dotnet/corefxlab/pull/2673#issuecomment-491459711

cc @steveharter, @GrabYourPitchforks


tests/System.Text.JsonLab.Tests/JsonObjectTests.cs
```diff
@@ -212,8 +212,9 @@ public void ParseJson(bool compactData, TestCaseType type, string jsonString)
 
                 if (type == TestCaseType.Json400KB)
                 {
-                    expectedString = ReadJson400KB(jtoken);
-                    actualString = ReadJson400KB(obj);
+                    // TODO: Investigate why 1 + 75486/(double)100000 == 1.7548599999999999 instead of just 1.75846
```
ahsonkhan: @tannergooding, is this expected behavior? Shows up in 3.0 only (not in netfx/.NET Core 2.1):
```C#
double x = 1 + 75486/(double)100000;
Console.WriteLine(x);   // 1.7548599999999999
```

ahsonkhan:
@safern, the CI isn't running all the legs. Do you know why?

safern:
> @safern, the CI isn't running all the legs. Do you know why?

I know there is an Azure DevOps bug with the github webhooks we have a thread going on. Don’t know if this is what is hitting but looking now.

safern:
The build is running but for some reason GH it is not showing it:

https://dev.azure.com/dnceng/public/_build/results?buildId=188138

ahsonkhan:
Since the build was actually successful, merging to unblock other contributors (https://dev.azure.com/dnceng/public/_build/results?buildId=188146). Will keep an eye out for any official build failures, just in case.

tests/System.Text.JsonLab.Tests/JsonObjectTests.cs
tannergooding: Yes. The raw bits for `1 + 75486 / (double)100000` are `0x3FFC13E81450EFDC`
The raw bits for `1.75486` are `0x3FFC13E81450EFDD`

In full framework, `ToString()` returned `1.75486` for both of them, which meant you couldn't differentiate.

In core, we return the shortest roundtrippable string, which allows them to always be differentiated.

PR #2682, opened 2019-05-15
Testing the DotnetBot AutoMerge Feature

Anipik:


NickCraver:
Tip: you can push a commit without any files for this sort of testing as to not impact history, like this:
```bash
git commit --allow-empty -m "Empty commit"
```

ahsonkhan:
Nice tip @NickCraver.

Additionally, if we ever have to modify a file (for whatever reason), we should stick to one to minimize the impact (looks like you had picked `/src/System.Azure.Experimental/System/Azure/Key.cs` previously, so I wouldn't pick a different file).

PR #2686, opened 2019-05-22
Implement support for Null values in columns

pgovind:
Implements support for null values in columns. 
Updated the unit tests to include null values and added some new ones to test the more corner cases

### DIFF
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -37,7 +37,6 @@ namespace Microsoft.Data
         public abstract long NullCount
         {
             get;
-            protected set;
         }
 
         public string Name { get; set; }
@@ -68,7 +67,7 @@ namespace Microsoft.Data
         /// <returns></returns>
         public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
 
-        internal virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
+        public virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
 
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
 
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -34,10 +34,10 @@ namespace Microsoft.Data
             }
         }
 
-        public virtual long NullCount
+        public abstract long NullCount
         {
-            get => throw new NotImplementedException();
-            protected set => throw new NotImplementedException();
+            get;
+            protected set;
         }
 
         public string Name { get; set; }
@@ -50,10 +50,10 @@ namespace Microsoft.Data
             set => SetValue(rowIndex, value);
         }
 
-        protected virtual object GetValue(long rowIndex) => throw new NotImplementedException();
-        protected virtual object GetValue(long startIndex, int length) => throw new NotImplementedException();
+        protected abstract object GetValue(long rowIndex);
+        protected abstract object GetValue(long startIndex, int length);
 
-        protected virtual void SetValue(long rowIndex, object value) => throw new NotImplementedException();
+        protected abstract void SetValue(long rowIndex, object value);
 
         public object this[long startIndex, int length]
         {
@@ -75,9 +75,9 @@ namespace Microsoft.Data
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
         internal delegate long GetBufferSortIndex(int bufferIndex, int sortIndex);
-        internal delegate Tuple<T, int> GetValueAndBufferSortIndexAtBuffer<T>(int bufferIndex, int valueIndex);
+        internal delegate ValueTuple<T, int> GetValueAndBufferSortIndexAtBuffer<T>(int bufferIndex, int valueIndex);
         internal delegate int GetBufferLengthAtIndex(int bufferIndex);
-        internal void PopulateColumnSortIndicesWithHeap<T>(SortedDictionary<T, List<Tuple<int, int>>> heapOfValueAndListOfTupleOfSortAndBufferIndex,
+        internal void PopulateColumnSortIndicesWithHeap<T>(SortedDictionary<T, List<ValueTuple<int, int>>> heapOfValueAndListOfTupleOfSortAndBufferIndex,
                                                             PrimitiveColumn<long> columnSortIndices,
                                                             GetBufferSortIndex getBufferSortIndex,
                                                             GetValueAndBufferSortIndexAtBuffer<T> getValueAndBufferSortIndexAtBuffer,
@@ -85,9 +85,9 @@ namespace Microsoft.Data
         {
             while (heapOfValueAndListOfTupleOfSortAndBufferIndex.Count > 0)
             {
-                KeyValuePair<T, List<Tuple<int, int>>> minElement = heapOfValueAndListOfTupleOfSortAndBufferIndex.ElementAt(0);
-                List<Tuple<int, int>> tuplesOfSortAndBufferIndex = minElement.Value;
-                Tuple<int, int> sortAndBufferIndex;
+                KeyValuePair<T, List<ValueTuple<int, int>>> minElement = heapOfValueAndListOfTupleOfSortAndBufferIndex.ElementAt(0);
+                List<ValueTuple<int, int>> tuplesOfSortAndBufferIndex = minElement.Value;
+                (int sortIndex, int bufferIndex) sortAndBufferIndex;
                 if (tuplesOfSortAndBufferIndex.Count == 1)
                 {
                     heapOfValueAndListOfTupleOfSortAndBufferIndex.Remove(minElement.Key);
@@ -98,18 +98,18 @@ namespace Microsoft.Data
                     sortAndBufferIndex = tuplesOfSortAndBufferIndex[tuplesOfSortAndBufferIndex.Count - 1];
                     tuplesOfSortAndBufferIndex.RemoveAt(tuplesOfSortAndBufferIndex.Count - 1);
                 }
-                int sortIndex = sortAndBufferIndex.Item1;
-                int bufferIndex = sortAndBufferIndex.Item2;
+                int sortIndex = sortAndBufferIndex.sortIndex;
+                int bufferIndex = sortAndBufferIndex.bufferIndex;
                 long bufferSortIndex = getBufferSortIndex(bufferIndex, sortIndex);
                 columnSortIndices.Append(bufferSortIndex);
                 if (sortIndex + 1 < getBufferLengthAtIndex(bufferIndex))
                 {
                     int nextSortIndex = sortIndex + 1;
-                    Tuple<T, int> nextValueAndBufferSortIndex = getValueAndBufferSortIndexAtBuffer(bufferIndex, nextSortIndex);
-                    T nextValue = nextValueAndBufferSortIndex.Item1;
+                    (T value, int bufferSortIndex) nextValueAndBufferSortIndex = getValueAndBufferSortIndexAtBuffer(bufferIndex, nextSortIndex);
+                    T nextValue = nextValueAndBufferSortIndex.value;
                     if (nextValue != null)
                     {
-                        heapOfValueAndListOfTupleOfSortAndBufferIndex.Add(nextValue, new List<Tuple<int, int>>() { new Tuple<int, int>((int)nextValueAndBufferSortIndex.Item2, bufferIndex) });
+                        heapOfValueAndListOfTupleOfSortAndBufferIndex.Add(nextValue, new List<ValueTuple<int, int>>() { (nextValueAndBufferSortIndex.bufferSortIndex, bufferIndex) });
                     }
                 }
             }
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -34,7 +34,11 @@ namespace Microsoft.Data
             }
         }
 
-        public long NullCount { get; protected set; }
+        public virtual long NullCount
+        {
+            get => throw new NotImplementedException();
+            protected set => throw new NotImplementedException();
+        }
 
         public string Name { get; set; }
 
@@ -64,17 +68,19 @@ namespace Microsoft.Data
         /// <returns></returns>
         public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
 
+        internal virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
+
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
 
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
         internal delegate long GetBufferSortIndex(int bufferIndex, int sortIndex);
-        internal delegate T GetValueAtBuffer<T>(int bufferIndex, int valueIndex);
+        internal delegate Tuple<T, int> GetValueAndBufferSortIndexAtBuffer<T>(int bufferIndex, int valueIndex);
         internal delegate int GetBufferLengthAtIndex(int bufferIndex);
         internal void PopulateColumnSortIndicesWithHeap<T>(SortedDictionary<T, List<Tuple<int, int>>> heapOfValueAndListOfTupleOfSortAndBufferIndex,
                                                             PrimitiveColumn<long> columnSortIndices,
                                                             GetBufferSortIndex getBufferSortIndex,
-                                                            GetValueAtBuffer<T> getValueAtBuffer,
+                                                            GetValueAndBufferSortIndexAtBuffer<T> getValueAndBufferSortIndexAtBuffer,
                                                             GetBufferLengthAtIndex getBufferLengthAtIndex)
         {
             while (heapOfValueAndListOfTupleOfSortAndBufferIndex.Count > 0)
@@ -99,8 +105,12 @@ namespace Microsoft.Data
                 if (sortIndex + 1 < getBufferLengthAtIndex(bufferIndex))
                 {
                     int nextSortIndex = sortIndex + 1;
-                    T nextValue = getValueAtBuffer(bufferIndex, nextSortIndex);
-                    heapOfValueAndListOfTupleOfSortAndBufferIndex.Add(nextValue, new List<Tuple<int, int>>() { new Tuple<int, int>((int)nextSortIndex, bufferIndex) });
+                    Tuple<T, int> nextValueAndBufferSortIndex = getValueAndBufferSortIndexAtBuffer(bufferIndex, nextSortIndex);
+                    T nextValue = nextValueAndBufferSortIndex.Item1;
+                    if (nextValue != null)
+                    {
+                        heapOfValueAndListOfTupleOfSortAndBufferIndex.Add(nextValue, new List<Tuple<int, int>>() { new Tuple<int, int>((int)nextValueAndBufferSortIndex.Item2, bufferIndex) });
+                    }
                 }
             }
 
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -4,6 +4,7 @@
 
 using System;
 using System.Collections.Generic;
+using System.Diagnostics;
 using System.Text;
 
 namespace Microsoft.Data
@@ -131,7 +132,9 @@ namespace Microsoft.Data
             List<BaseColumn> newColumns = new List<BaseColumn>(ColumnCount);
             for (int i = 0; i < ColumnCount; i++)
             {
-                var newColumn = Column(i).Clone(sortIndices, !ascending);
+                BaseColumn oldColumn = Column(i);
+                BaseColumn newColumn = oldColumn.CloneAndAppendNulls(sortIndices, !ascending);
+                Debug.Assert(newColumn.NullCount == oldColumn.NullCount);
                 newColumns.Add(newColumn);
             }
             return new DataFrame(newColumns);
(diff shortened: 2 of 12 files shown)


src/Microsoft.Data/BaseColumn.cs
```diff
@@ -64,17 +68,19 @@ protected set
...
 
+        internal virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
+
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
 
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
         internal delegate long GetBufferSortIndex(int bufferIndex, int sortIndex);
-        internal delegate T GetValueAtBuffer<T>(int bufferIndex, int valueIndex);
+        internal delegate Tuple<T, int> GetValueAndBufferSortIndexAtBuffer<T>(int bufferIndex, int valueIndex);
```
gfoidl: Can this be a value-tuple? Would safe the allocation for `Tuple<T1, T2>`.

src/Microsoft.Data/PrimitiveColumn.Sort.cs
```diff
@@ -36,26 +37,43 @@ private void GetSortIndices(IComparer<T> comparer, out PrimitiveColumn<long> col
                 bufferSortIndices.Add(sortIndices);
             }
             // Simple merge sort to build the full column's sort indices
+            Tuple<T, int> GetFirstNonNullValueAndBufferIndexStartingAtIndex(int bufferIndex, int startIndex)
```
gfoidl: Value-tuple?

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+            lastBuffer.Append(value ?? default);
+            SetBit(Length, value.HasValue ? true : false);
             Length++;
         }
 
+        public bool IsValid(long index) => NullCount == 0 || GetBit(index);
+
+        private void SetBit(long index, bool value)
+        {
+            if (index < 0 || index > Length)
```
```suggestion
            if ((ulong)index > Length)
```
Saves a explicit comparison. Negative indexed casted to `ulong` results in values > long.MaxValue.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
```
gfoidl: Maybe use a ThrowHelper? Altough this method is quite long, so won't be inlined. Still the JITed code will be smaller, as the creation of the exception is moved out of this method.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+            }
+            // First find the right bitMapBuffer
+            int bitMapIndex = (int)(index / Buffers[0].MaxCapacity);
+            Debug.Assert(NullBitMapBuffers.Count > bitMapIndex);
+            DataFrameBuffer<byte> bitMapBuffer = NullBitMapBuffers[bitMapIndex];
+
+            // Set the bit
+            index -= bitMapIndex * Buffers[0].MaxCapacity;
+            int bitMapBufferIndex = (int)index / 8;
```
```suggestion
            int bitMapBufferIndex = (int)index >> 3;    // index / 8
```
The bit-shift will produce [better code](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAbgFgAoYgZkoCZyBhcgb0fIGVWASwB2GcgFkAFABsIogObkxAExgIAlOQC8APnLSxGTWo3kA9OQAc9BoPL9BLFePIA5OQuVmtug0bipqLqCOR6Bsx2AL5AA=)

JIT will substitute division by a power of two only for unsigned ints / longs. So instead of the bit-shift it [could also be](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAbgFgAoYgZkoCZyBhcgb0fIGVWASwB2GcgFkAFABsIogObkxAExgIAlOQC8APnLSxGTWo3kA9OQAc9BoPL9BLFePIA5OQuVmtug0bipqLqCOR6Bsx2Dk4CLsZSVF5KrqHa+obGmtLSAK5ZvpY2mnYAvkA===)
```suggestion
            int bitMapBufferIndex = (int)((uint)index / 8);
```

You can take whatever you prefer :wink:
I would take the latter.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+            Debug.Assert(bitMapBuffer.Length >= bitMapBufferIndex);
+            if (bitMapBuffer.Length == bitMapBufferIndex)
+                bitMapBuffer.Append(0);
+            byte curBitMap = bitMapBuffer[bitMapBufferIndex];
+            byte newBitMap;
+            if (value)
+            {
+                newBitMap = (byte)(curBitMap | (byte)(1 << (int)(index % 8)));
```
```suggestion
                newBitMap = (byte)(curBitMap | (byte)(1 << (int)((uint)index % 8)));
```
Boils down to [index & 7](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAbgFgAoYgZkoCZyBhcgb0fIGVWASwB2GcgFkAFABsIogObkxAExgIAlOQC8APnLSxGTUdHqE5AKTkAHJvoNB5foJYrx5AHJyFytRra+obGptIArqEBljb2jgC+QA).

Or you write the bit-hack yourself as `(int)(index & 7)`.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+                if ((curBitMap >> ((int)(index % 8)) & 1) == 0 && index < Length && NullCount > 0)
```
```suggestion
                if ((curBitMap >> ((int)index & 1) == 0 && index < Length && NullCount > 0)
```
`index % 8` <=> `index & 7`, so `& 1` will mask the rest, just `& 1` remains.

JIT [won't optimize this](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABABgAJiBGAbgFgAoYgZkoCZyBhcgb0fIGVWASwB2GcgFkAFABsIogObkxAExgIAlOQC8APnLTpYjJuOj1CcgFJyADk3aAZOVr9B7gSxXjyAOTkFZTUNbX1Dc1MjAFcTTRCrWwdnV3oGQXJPIR8JSSpApR9LMINIswTyFwB2FLcGAF8gA=).

Same below in else-branch.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+                { 
+                    // old value was NOT null and new value is null
+                    NullCount++;
+                }
+                else if (index == Length)
+                {
+                    // New entry from an append
+                    NullCount++;
+                }
+                newBitMap = (byte)(curBitMap & (byte)~(1 << (int)(index % 8)));
```
```suggestion
                newBitMap = (byte)(curBitMap & (byte)~(1 << (int)((uint)index % 8)));
```

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+            }
+            bitMapBuffer[bitMapBufferIndex] = newBitMap;
+        }
+
+        private bool GetBit(long index)
+        {
+            if (index < 0 || index > Length)
```
```suggestion
            if ((ulong)index > Length)
```
See above. Also for ThrowHelper-thoughts.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+            // First find the right bitMapBuffer
+            int bitMapIndex = (int)(index / Buffers[0].MaxCapacity);
+            Debug.Assert(NullBitMapBuffers.Count > bitMapIndex);
+            DataFrameBuffer<byte> bitMapBuffer = NullBitMapBuffers[bitMapIndex];
+
+            // Get the bit
+            index -= bitMapIndex * Buffers[0].MaxCapacity;
+            int bitMapBufferIndex = (int)index / 8;
```
```suggestion
            int bitMapBufferIndex = (int)(uint)index / 8;
```

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+            Debug.Assert(bitMapBuffer.Length > bitMapBufferIndex);
+            byte curBitMap = bitMapBuffer[bitMapBufferIndex];
+            return ((curBitMap >> ((int)index % 8)) & 1) != 0;
```
```suggestion
            return ((curBitMap >> ((int)index & 1) != 0;
```
See above.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -124,47 +138,64 @@ protected override void SetValue(long rowIndex, object value)
...
+                {
                     sortIndices[i] = i;
+                }
                 // TODO: Refactor the sort routine to also work with IList?
                 string[] array = buffer.ToArray();
                 IntrospectiveSort(array, array.Length, sortIndices, comparer);
                 bufferSortIndices.Add(sortIndices);
             }
             // Simple merge sort to build the full column's sort indices
+            Tuple<string, int> GetFirstNonNullValueStartingAtIndex(int stringBufferIndex, int startIndex)
```
gfoidl: Valuetuple?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -15,6 +15,8 @@ public static DataFrame MakeDataFrameWithTwoColumns(int length)
         {
             BaseColumn dataFrameColumn1 = new PrimitiveColumn<int>("Int1", Enumerable.Range(0, length).Select(x => x));
             BaseColumn dataFrameColumn2 = new PrimitiveColumn<int>("Int2", Enumerable.Range(10, length).Select(x => x));
+            dataFrameColumn1[length/2] = null;
```
gfoidl: Nit: spacing

```suggestion
            dataFrameColumn1[length / 2] = null;
```
Same on next line.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -34,7 +34,11 @@ protected set
             }
         }
 
-        public long NullCount { get; protected set; }
+        public virtual long NullCount
```
eerhardt: Why not `abstract`?  I guess the same could be said for GetValue and SetValue below.  (Should have said that last time)

src/Microsoft.Data/DataFrame.cs
```diff
@@ -131,7 +132,9 @@ public DataFrame Sort(string columnName, bool ascending = true)
             List<BaseColumn> newColumns = new List<BaseColumn>(ColumnCount);
             for (int i = 0; i < ColumnCount; i++)
             {
-                var newColumn = Column(i).Clone(sortIndices, !ascending);
+                BaseColumn oldColumn = Column(i);
+                BaseColumn newColumn = oldColumn.CloneAndAppendNulls(sortIndices, !ascending);
```
eerhardt: `CloneAndAppendNulls` is internal. How could someone derive their own column type from BaseColumn, and then use this Sort method?

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -34,7 +34,11 @@ protected set
...
+        {
+            get => throw new NotImplementedException();
+            protected set => throw new NotImplementedException();
```
eerhardt: Who would ever `set` this property? Should it be `get` only?

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -106,13 +122,23 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
             {
                 if (mapIndices.DataType != typeof(long))
                     throw new ArgumentException(Strings.MismatchedValueType + " PrimitiveColumn<long>", nameof(mapIndices));
-                if (mapIndices.Length != Length)
+                if (mapIndices.Length >= Length)
                     throw new ArgumentException(Strings.MismatchedColumnLengths, nameof(mapIndices));
```
eerhardt: Does `MismatchedColumnLengths` provide a good error message here now? Before it was just different lengths, but now the check is only for greater than or equals.

Same applies to the below usages of MismatchedColumnLengths.

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -106,13 +122,23 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
...
             }
             return Clone();
         }
 
+        internal override BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false)
+        {
+            PrimitiveColumn<T> ret = Clone(mapIndices, invertMapIndices) as PrimitiveColumn<T>;
+            for (long i = 0; i < NullCount; i++)
+            {
+                ret.Append(null);
```
eerhardt: Instead of appending `null` multiple times, what do you think about a method that will do it all at once?  Like `AppendMany(value, count)`.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
             }
-            lastBuffer.Append(value);
+            lastBuffer.Append(value ?? default);
+            SetBit(Length, value.HasValue ? true : false);
             Length++;
         }
 
+        public bool IsValid(long index) => NullCount == 0 || GetBit(index);
+
+        private void SetBit(long index, bool value)
```
eerhardt: Is there a better name than `SetBit` for this method? I don't think it is very descriptive of what it does.  maybe `SetNullIndex`? or something linking it to `Null` support.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -25,6 +31,7 @@ public PrimitiveColumnContainer(T[] values)
             {
                 curBuffer = new DataFrameBuffer<T>();
                 Buffers.Add(curBuffer);
+                NullBitMapBuffers.Add(new DataFrameBuffer<byte>());
```
eerhardt: Is it possible to lazy initialize the NullBitMapBuffers? For example, if I don't have any null values, there should be no needed to create these objects.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -36,8 +43,10 @@ public PrimitiveColumnContainer(T[] values)
                 {
                     curBuffer = new DataFrameBuffer<T>();
                     Buffers.Add(curBuffer);
+                    NullBitMapBuffers.Add(new DataFrameBuffer<byte>());
                 }
                 curBuffer.Append(values[i]);
+                SetBit(Length, true);
```
eerhardt: Why are we doing this? It shouldn't be possible to get or set the null value at index `Length`. That should be out of range.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
+            // Get the bit
+            index -= bitMapIndex * Buffers[0].MaxCapacity;
+            int bitMapBufferIndex = (int)index / 8;
+            Debug.Assert(bitMapBuffer.Length > bitMapBufferIndex);
+            byte curBitMap = bitMapBuffer[bitMapBufferIndex];
+            return ((curBitMap >> ((int)index % 8)) & 1) != 0;
+        }
+
         public long Length;
         //TODO:
```
eerhardt: Can you remove the TODO?

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +81,180 @@ public PrimitiveColumnContainer(long length = 0)
...
-                int arrayIndex = GetArrayContainingRowIndex(ref startIndex);
-                bool temp = Buffers[arrayIndex][(int)startIndex, length, ret];
-                while (ret.Count < length)
+                long endIndex = Math.Min(Length, startIndex + length);
+                for (long i = startIndex; i < endIndex; i++)
                 {
-                    arrayIndex++;
-                    temp = Buffers[arrayIndex][0, length - ret.Count, ret];
+                    T? value = this[i];
+                    if (value.HasValue)
```
eerhardt: Is this correct - to skip any nulls? Would the caller expect that passing in a `length` of `10` be returned with a list of `8` values?

src/Microsoft.Data/StringColumn.cs
```diff
@@ -18,7 +16,17 @@ public partial class StringColumn : BaseColumn
...
-            _stringBuffers.Add(new List<string>());
+            int numberOfBuffersRequired = Math.Max((int)(length / int.MaxValue), 1);
+            for (int i = 0; i < numberOfBuffersRequired; i++)
+            {
+                long bufferLen = length - _stringBuffers.Count * int.MaxValue;
+                List<string> buffer = new List<string>((int)Math.Min(int.MaxValue, bufferLen));
+                _stringBuffers.Add(buffer);
+                for (int j = 0; j < bufferLen; j++)
+                {
+                    buffer.Add(default);
```
eerhardt: Is this necessary? If yes - is this the most performant way of appending `bufferLen` `nulls` to the list?

src/Microsoft.Data/StringColumn.cs
```diff
@@ -85,10 +91,18 @@ protected override object GetValue(long startIndex, int length)
...
+            if (value == null || value is string)
             {
                 int bufferIndex = GetBufferIndexContainingRowIndex(ref rowIndex);
+                var oldValue = this[rowIndex];
                 _stringBuffers[bufferIndex][(int)rowIndex] = (string)value;
+                if (oldValue != (string)value)
+                {
+                    if (value == null)
+                        _nullCount++;
+                    if (oldValue == null && _nullCount > 0)
```
eerhardt: Can the `oldValue` be `null` AND `_nullCount <= 0`? Maybe that should be an `Assert`?

eerhardt:
> using System;

copyright

---
Refers to: tests/Microsoft.Data.DataFrame.Tests/BufferTests.cs:1 in 9ce1ada. [](commit_id = 9ce1ada5149427c535e2d8b80aab01784efa747c, deletion_comment = False)

eerhardt:
Maybe a good test to have would be that you can take an `Apache.Arrow.RecordBatch` and wrap a `DataFrame` around it without any copies. I'm wondering if the "null" support you are adding is compatible with the Arrow "null" format.

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: The `& 1` should take place after the ` >> ` :)

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Same reply. The right shift should happen before the And

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
pgovind: Damn, I wonder why the .editorconfig did not find this

src/Microsoft.Data/PrimitiveColumnContainer.cs
gfoidl: Ups... :smile:

src/Microsoft.Data/DataFrame.cs
pgovind: Good catch!

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: Yup, good idea. Done

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: How about `SetValidityBit`? Unfortunately in Arrow, a `1` implies a valid value and a `0` implies a null value.

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: I did consider lazy initialization, but the C# Arrow implementation seems to always maintain a null bit map, so I went with this.

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Right, `GetValidityBit` does throw an exception when `index >= Length`. The set here is still valid because the current length is `Length + 1`, after the `curBuffer.Append` call just above. Line 50(the next line) is `Length++` and updates the length.

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Yup, this was wrong. Fixed it

src/Microsoft.Data/StringColumn.cs
pgovind: Yes for the necessary question. This seemed more efficient to me than appending and doubling the size everytime? 
I don't see a more performant way to appending nulls to a list. If I had an array, I could've just initialized it. I could use LINQ, but I'm not so sure that that is more performant?

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -64,24 +67,26 @@ protected set
         /// <returns></returns>
         public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
 
+        public virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
```
eerhardt: Can `appendNulls` just be an option to the existing `Clone` method?

eerhardt:
> Oh also, how do folks reply to comments like this on GitHub? I don't see a "Reply" option, just an "Edit" option

Typically, you reply to comments like this on GitHub at the bottom, and using a `>` quote of the comment you are replying to.

src/Microsoft.Data/BaseColumn.cs
pgovind: It could. I'll put it in https://github.com/dotnet/corefxlab/pull/2687

pgovind:
Thanks for the review @gfoidl  and @eerhardt

PR #2687, opened 2019-05-30
Implement "Join"

pgovind:
This should go in after the [Null](https://github.com/dotnet/corefxlab/pull/2686) patch

More work towards a [DataFrame](https://github.com/dotnet/corefx/issues/26845) type. "Merge" will follow after this patch.

### DIFF
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -65,9 +65,7 @@ namespace Microsoft.Data
         /// <param name="mapIndices"></param>
         /// <param name="invertMapIndices"></param>
         /// <returns></returns>
-        public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
-
-        public virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
+        public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
 
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
 
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -151,12 +151,13 @@ namespace Microsoft.Data
         private void SetSuffixForDuplicatedColumnNames(DataFrame dataFrame, BaseColumn column, string leftSuffix, string rightSuffix)
         {
             int index = dataFrame._table.GetColumnIndex(column.Name);
-            if (index != -1)
+            while (index != -1)
             {
                 // Pre-existing column. Change name
                 BaseColumn existingColumn = dataFrame.Column(index);
                 dataFrame._table.SetColumnName(existingColumn, existingColumn.Name + leftSuffix);
                 column.Name += rightSuffix;
+                index = dataFrame._table.GetColumnIndex(column.Name);
             }
         }
 
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -9,6 +9,14 @@ using System.Text;
 
 namespace Microsoft.Data
 {
+    public enum JoinAlgorithm
+    {
+        Left,
+        Right,
+        Outer,
+        Inner
+    }
+
     /// <summary>
     /// A DataFrame to support indexing, binary operations, sorting, selection and other APIs. This will eventually also expose an IDataView for ML.NET
     /// </summary>
@@ -115,12 +123,12 @@ namespace Microsoft.Data
         // TODO: Add strongly typed versions of these APIs
         #endregion
 
-        private DataFrame Clone()
+        private DataFrame Clone(BaseColumn mapIndices = null, bool invertMapIndices = false)
         {
             List<BaseColumn> newColumns = new List<BaseColumn>(ColumnCount);
             for (int i = 0; i < ColumnCount; i++)
             {
-                newColumns.Add(Column(i).Clone());
+                newColumns.Add(Column(i).Clone(mapIndices, invertMapIndices));
             }
             return new DataFrame(newColumns);
         }
@@ -133,11 +141,122 @@ namespace Microsoft.Data
             for (int i = 0; i < ColumnCount; i++)
             {
                 BaseColumn oldColumn = Column(i);
-                BaseColumn newColumn = oldColumn.CloneAndAppendNulls(sortIndices, !ascending);
+                BaseColumn newColumn = oldColumn.Clone(sortIndices, !ascending, oldColumn.NullCount);
                 Debug.Assert(newColumn.NullCount == oldColumn.NullCount);
                 newColumns.Add(newColumn);
             }
             return new DataFrame(newColumns);
+        }
+
+        private void SetSuffixForDuplicatedColumnNames(DataFrame dataFrame, BaseColumn column, string leftSuffix, string rightSuffix)
+        {
+            int index = dataFrame._table.GetColumnIndex(column.Name);
+            if (index != -1)
+            {
+                // Pre-existing column. Change name
+                BaseColumn existingColumn = dataFrame.Column(index);
+                dataFrame._table.SetColumnName(existingColumn, existingColumn.Name + leftSuffix);
+                column.Name += rightSuffix;
+            }
+        }
+
+        public DataFrame Join(DataFrame other, string leftSuffix = "_left", string rightSuffix = "_right", JoinAlgorithm joinAlgorithm = JoinAlgorithm.Left)
+        {
+            DataFrame ret = new DataFrame();
+            if (joinAlgorithm == JoinAlgorithm.Left)
+            {
+                for (int i = 0; i < ColumnCount; i++)
+                {
+                    BaseColumn newColumn = Column(i).Clone();
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+                long minLength = Math.Min(RowCount, other.RowCount);
+                PrimitiveColumn<long> mapIndices = new PrimitiveColumn<long>("mapIndices", minLength);
+                for (long i = 0; i < minLength; i++)
+                {
+                    mapIndices[i] = i;
+                }
+                for (int i = 0; i < other.ColumnCount; i++)
+                {
+                    BaseColumn newColumn;
+                    if (other.RowCount < RowCount)
+                    {
+                        newColumn = other.Column(i).Clone(numberOfNullsToAppend: RowCount - other.RowCount);
+                    }
+                    else
+                    {
+                        newColumn = other.Column(i).Clone(mapIndices);
+                    }
+                    SetSuffixForDuplicatedColumnNames(ret, newColumn, leftSuffix, rightSuffix);
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+            }
+            else if (joinAlgorithm == JoinAlgorithm.Right)
+            {
+                long minLength = Math.Min(RowCount, other.RowCount);
+                PrimitiveColumn<long> mapIndices = new PrimitiveColumn<long>("mapIndices", minLength);
+                for (long i = 0; i < minLength; i++)
+                {
+                    mapIndices[i] = i;
+                }
+                for (int i = 0; i < ColumnCount; i++)
+                {
+                    BaseColumn newColumn;
+                    if (RowCount < other.RowCount)
+                    {
+                        newColumn = Column(i).Clone(numberOfNullsToAppend: other.RowCount - RowCount);
+                    }
+                    else
+                    {
+                        newColumn = Column(i).Clone(mapIndices);
+                    }
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+                for (int i = 0; i < other.ColumnCount; i++)
+                {
+                    BaseColumn newColumn = other.Column(i).Clone();
+                    SetSuffixForDuplicatedColumnNames(ret, newColumn, leftSuffix, rightSuffix);
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+            }
+            else if (joinAlgorithm == JoinAlgorithm.Outer)
+            {
+                long newRowCount = Math.Max(RowCount, other.RowCount);
+                long numberOfNulls = newRowCount - RowCount;
+                for (int i = 0; i < ColumnCount; i++)
+                {
+                    BaseColumn newColumn = Column(i).Clone(numberOfNullsToAppend: numberOfNulls);
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+                numberOfNulls = newRowCount - other.RowCount;
+                for (int i = 0; i < other.ColumnCount; i++)
+                {
+                    BaseColumn newColumn = other.Column(i).Clone(numberOfNullsToAppend: numberOfNulls);
+                    SetSuffixForDuplicatedColumnNames(ret, newColumn, leftSuffix, rightSuffix);
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+            }
+            else if (joinAlgorithm == JoinAlgorithm.Inner)
+            {
+                long newRowCount = Math.Min(RowCount, other.RowCount);
+                PrimitiveColumn<long> mapIndices = new PrimitiveColumn<long>("mapIndices", newRowCount);
+                for (long i = 0; i < newRowCount; i++)
+                {
+                    mapIndices[i] = i;
+                }
+                for (int i = 0; i < ColumnCount; i++)
+                {
+                    BaseColumn newColumn = Column(i).Clone(mapIndices);
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+                for (int i = 0; i < other.ColumnCount; i++)
+                {
+                    BaseColumn newColumn = other.Column(i).Clone(mapIndices);
+                    SetSuffixForDuplicatedColumnNames(ret, newColumn, leftSuffix, rightSuffix);
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+            }
+            return ret;
         }
 
         public override string ToString()
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -13,7 +13,7 @@ namespace Microsoft.Data
     {
         Left,
         Right,
-        Outer,
+        FullOuter,
         Inner
     }
 
@@ -220,7 +220,7 @@ namespace Microsoft.Data
                     ret.InsertColumn(ret.ColumnCount, newColumn);
                 }
             }
-            else if (joinAlgorithm == JoinAlgorithm.Outer)
+            else if (joinAlgorithm == JoinAlgorithm.FullOuter)
             {
                 long newRowCount = Math.Max(RowCount, other.RowCount);
                 long numberOfNulls = newRowCount - RowCount;
--- a/src/Microsoft.Data/DataFrameTable.cs
+++ b/src/Microsoft.Data/DataFrameTable.cs
@@ -57,6 +57,16 @@ namespace Microsoft.Data
             return ret;
         }
 
+        public void SetColumnName(BaseColumn column, string newName)
+        {
+            string currentName = column.Name;
+            int currentIndex = _columnNameToIndexDictionary[currentName];
+            column.Name = newName;
+            _columnNameToIndexDictionary.Remove(currentName);
+            _columnNameToIndexDictionary.Add(newName, currentIndex);
+        }
+
+
         public void InsertColumn<T>(int columnIndex, IEnumerable<T> column, string columnName)
             where T : unmanaged
         {
--- a/src/Microsoft.Data/DataFrameTable.cs
+++ b/src/Microsoft.Data/DataFrameTable.cs
@@ -66,7 +66,6 @@ namespace Microsoft.Data
             _columnNameToIndexDictionary.Add(newName, currentIndex);
         }
 
-
         public void InsertColumn<T>(int columnIndex, IEnumerable<T> column, string columnName)
             where T : unmanaged
         {
--- a/src/Microsoft.Data/PrimitiveColumn.Sort.cs
+++ b/src/Microsoft.Data/PrimitiveColumn.Sort.cs
@@ -14,7 +14,7 @@ namespace Microsoft.Data
         public override BaseColumn Sort(bool ascending = true)
         {
             PrimitiveColumn<long> sortIndices = GetAscendingSortIndices() as PrimitiveColumn<long>;
-            return CloneAndAppendNulls(sortIndices, !ascending);
+            return Clone(sortIndices, !ascending, NullCount);
         }
 
         internal override BaseColumn GetAscendingSortIndices()
--- a/src/Microsoft.Data/PrimitiveColumn.cs
+++ b/src/Microsoft.Data/PrimitiveColumn.cs
@@ -118,24 +118,24 @@ namespace Microsoft.Data
             return $"{Name}: {_columnContainer.ToString()}";
         }
 
-        public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false)
+        public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0)
         {
+            PrimitiveColumn<T> clone;
             if (!(mapIndices is null))
             {
                 if (mapIndices.DataType != typeof(long))
                     throw new ArgumentException(Strings.MismatchedValueType + " PrimitiveColumn<long>", nameof(mapIndices));
                 if (mapIndices.Length > Length)
                     throw new ArgumentException(Strings.MapIndicesExceedsColumnLenth, nameof(mapIndices));
-                return Clone(mapIndices as PrimitiveColumn<long>, invertMapIndices);
+                clone = Clone(mapIndices as PrimitiveColumn<long>, invertMapIndices);
             }
-            return Clone();
-        }
-
-        public override BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false)
-        {
-            PrimitiveColumn<T> ret = Clone(mapIndices, invertMapIndices) as PrimitiveColumn<T>;
-            ret.AppendMany(null, NullCount);
-            return ret;
+            else
+            {
+                clone = Clone();
+            }
+            Debug.Assert(!ReferenceEquals(clone, null));
+            clone.AppendMany(null, numberOfNullsToAppend);
+            return clone;
         }
 
         public PrimitiveColumn<T> Clone(PrimitiveColumn<long> mapIndices = null, bool invertMapIndices = false)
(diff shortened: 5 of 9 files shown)


src/Microsoft.Data/BaseColumn.cs
```diff
@@ -64,24 +67,26 @@ protected set
...
                                                             GetBufferLengthAtIndex getBufferLengthAtIndex)
         {
             while (heapOfValueAndListOfTupleOfSortAndBufferIndex.Count > 0)
             {
-                KeyValuePair<T, List<Tuple<int, int>>> minElement = heapOfValueAndListOfTupleOfSortAndBufferIndex.ElementAt(0);
-                List<Tuple<int, int>> tuplesOfSortAndBufferIndex = minElement.Value;
-                Tuple<int, int> sortAndBufferIndex;
+                KeyValuePair<T, List<ValueTuple<int, int>>> minElement = heapOfValueAndListOfTupleOfSortAndBufferIndex.ElementAt(0);
+                List<ValueTuple<int, int>> tuplesOfSortAndBufferIndex = minElement.Value;
+                (int sortIndex, int bufferIndex) sortAndBufferIndex;
```
gfoidl: Deconstruct the tuple, so later usage is easier and the locals from L100, L101 are not needed.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -92,15 +97,19 @@ protected set
...
+                int sortIndex = sortAndBufferIndex.sortIndex;
+                int bufferIndex = sortAndBufferIndex.bufferIndex;
                 long bufferSortIndex = getBufferSortIndex(bufferIndex, sortIndex);
                 columnSortIndices.Append(bufferSortIndex);
                 if (sortIndex + 1 < getBufferLengthAtIndex(bufferIndex))
                 {
                     int nextSortIndex = sortIndex + 1;
-                    T nextValue = getValueAtBuffer(bufferIndex, nextSortIndex);
-                    heapOfValueAndListOfTupleOfSortAndBufferIndex.Add(nextValue, new List<Tuple<int, int>>() { new Tuple<int, int>((int)nextSortIndex, bufferIndex) });
+                    (T value, int bufferSortIndex) nextValueAndBufferSortIndex = getValueAndBufferSortIndexAtBuffer(bufferIndex, nextSortIndex);
```
gfoidl: Deconstruct?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -131,12 +132,125 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
+                BaseColumn newColumn = oldColumn.CloneAndAppendNulls(oldColumn.NullCount, sortIndices, !ascending);
+                Debug.Assert(newColumn.NullCount == oldColumn.NullCount);
                 newColumns.Add(newColumn);
             }
             return new DataFrame(newColumns);
         }
 
+        public enum JoinAlgorithm
+        {
+            LEFT,
```
gfoidl: Just usual _PascalCase_ for enum members. So
```c#
Left,
Right,
...
```

src/Microsoft.Data/PrimitiveColumn.Sort.cs
```diff
@@ -36,26 +37,43 @@ private void GetSortIndices(IComparer<T> comparer, out PrimitiveColumn<long> col
...
+                return (value, startIndex);
+            }
+            SortedDictionary<T, List<ValueTuple<int, int>>> heapOfValueAndListOfTupleOfSortAndBufferIndex = new SortedDictionary<T, List<ValueTuple<int, int>>>(comparer);
             IList<DataFrameBuffer<T>> buffers = _columnContainer.Buffers;
             for (int i = 0; i < buffers.Count; i++)
             {
                 DataFrameBuffer<T> buffer = buffers[i];
-                T value = buffer[bufferSortIndices[i][0]];
-                if (heapOfValueAndListOfTupleOfSortAndBufferIndex.ContainsKey(value))
+                ValueTuple<T, int> valueAndBufferIndex = GetFirstNonNullValueAndBufferIndexStartingAtIndex(i, 0);
```
gfoidl: Deconstruct so later usage can be done without `Item1` and `Item2`.

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -106,13 +124,20 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
...
+                if (mapIndices.Length > Length)
+                    throw new ArgumentException(Strings.MapIndicesExceedsColumnLenth, nameof(mapIndices));
                 return Clone(mapIndices as PrimitiveColumn<long>, invertMapIndices);
             }
             return Clone();
         }
 
+        public override BaseColumn CloneAndAppendNulls(long numberOfNulls, BaseColumn mapIndices = null, bool invertMapIndices = false)
+        {
+            PrimitiveColumn<T> ret = Clone(mapIndices, invertMapIndices) as PrimitiveColumn<T>;
```
gfoidl: Add `Debug.Assert(ret != null);` just to be on the safe side?

And rename `ret` to `clone`?

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +84,224 @@ public PrimitiveColumnContainer(long length = 0)
...
             Length++;
         }
 
+        public void AppendMany(T? value, long count)
+        {
+            if (!value.HasValue)
+            {
+                NullCount += count;
+            }
+            while (count > 0)
```
gfoidl: Nit: empty line  between `if` and `while`.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +84,224 @@ public PrimitiveColumnContainer(long length = 0)
...
+                    NullBitMapBuffers.Add(new DataFrameBuffer<byte>());
+                }
+                DataFrameBuffer<T> lastBuffer = Buffers[Buffers.Count - 1];
+                if (lastBuffer.Length == lastBuffer.MaxCapacity)
+                {
+                    lastBuffer = new DataFrameBuffer<T>();
+                    Buffers.Add(lastBuffer);
+                    NullBitMapBuffers.Add(new DataFrameBuffer<byte>());
+                }
+                int allocatable = (int)Math.Min(count, lastBuffer.MaxCapacity);
```
gfoidl: Is this safe for overflow?
(I don't remember the type of MaxCapacity).

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +84,224 @@ public PrimitiveColumnContainer(long length = 0)
...
+                lastBuffer.EnsureCapacity(allocatable);
+                lastBuffer.Span.Slice(lastBuffer.Length, allocatable).Fill(value ?? default);
+                lastBuffer.Length += allocatable;
+                Length += allocatable;
+
+                DataFrameBuffer<byte> lastNullBitMapBuffer = NullBitMapBuffers[NullBitMapBuffers.Count - 1];
+                int nullBitMapAllocatable = (int)Math.Ceiling(allocatable / 8.0);
```
```suggestion
                int nullBitMapAllocatable = (allocatable + 7) / 8;
```
No need for ceil and floatint point.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +84,224 @@ public PrimitiveColumnContainer(long length = 0)
...
+        /// A null value has an unset bit
+        /// A NON-null value has a set bit
+        /// </summary>
+        /// <param name="index"></param>
+        /// <param name="value"></param>
+        private void SetValidityBit(long index, bool value)
+        {
+            if ((uint)index > Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
```
gfoidl: ThrowHelper?

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +84,224 @@ public PrimitiveColumnContainer(long length = 0)
...
+
+        /// <summary>
+        /// A null value has an unset bit
+        /// A NON-null value has a set bit
+        /// </summary>
+        /// <param name="index"></param>
+        /// <param name="value"></param>
+        private void SetValidityBit(long index, bool value)
+        {
+            if ((uint)index > Length)
```
```suggestion
            if ((ulong)index > Length)
```
or is it safe with `uint`?

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +84,224 @@ public PrimitiveColumnContainer(long length = 0)
...
+            index -= bitMapIndex * Buffers[0].MaxCapacity;
+            int bitMapBufferIndex = (int)((uint)index / 8);
+            Debug.Assert(bitMapBuffer.Length >= bitMapBufferIndex);
+            if (bitMapBuffer.Length == bitMapBufferIndex)
+                bitMapBuffer.Append(0);
+            byte curBitMap = bitMapBuffer[bitMapBufferIndex];
+            byte newBitMap;
+            if (value)
+            {
+                newBitMap = (byte)(curBitMap | (byte)(1 << (int)(index % 8)));
```
```suggestion
                newBitMap = (byte)(curBitMap | (byte)(1 << (int)(index & 7)));   // bithack for % 8
```

The same on other places, did not comment them explicitely. Search for `% 8`.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -69,91 +84,224 @@ public PrimitiveColumnContainer(long length = 0)
...
+                    NullCount++;
+                }
+                newBitMap = (byte)(curBitMap & (byte)~(1 << (int)((uint)index % 8)));
+            }
+            bitMapBuffer[bitMapBufferIndex] = newBitMap;
+        }
+
+        private bool GetValidityBit(long index)
+        {
+            if ((uint)index >= Length)
```
gfoidl: Same question as above, also for ThrowHelper.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -124,47 +139,64 @@ protected override void SetValue(long rowIndex, object value)
...
+            }
+
+            SortedDictionary<string, List<ValueTuple<int, int>>> heapOfValueAndListOfTupleOfSortAndBufferIndex = new SortedDictionary<string, List<ValueTuple<int, int>>>(comparer);
             List<List<string>> buffers = _stringBuffers;
             for (int i = 0; i < buffers.Count; i++)
             {
                 List<string> buffer = buffers[i];
-                string value = buffer[bufferSortIndices[i][0]];
-                if (heapOfValueAndListOfTupleOfSortAndBufferIndex.ContainsKey(value))
+                ValueTuple<string, int> valueAndBufferSortIndex = GetFirstNonNullValueStartingAtIndex(i, 0);
```
gfoidl: Deconstruct?

src/Microsoft.Data/StringColumn.cs
```diff
@@ -175,47 +207,50 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
...
+                if (mapIndices.Length > Length)
+                    throw new ArgumentException(Strings.MapIndicesExceedsColumnLenth, nameof(mapIndices));
                 return Clone(mapIndices as PrimitiveColumn<long>, invertMapIndices);
             }
             return Clone();
         }
 
+        public override BaseColumn CloneAndAppendNulls(long numberOfNulls, BaseColumn mapIndices = null, bool invertMapIndices = false)
+        {
+            StringColumn ret = Clone(mapIndices, invertMapIndices) as StringColumn;
```
gfoidl: `Debug.Assert(ret != null)`? Rename `ret` to `clone`?

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Yup. MaxCapacity returns an int

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: I'm no expert on compilers/runtimes, but I wonder why the compiler doesn't already do this

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Yup. Just not in this patch. Which namespace has ThrowHelper though? When I look at source.dot.net, it seems to be `internal` to System?

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Good catch. It should be ulong. Copy-paste error

src/Microsoft.Data/PrimitiveColumnContainer.cs
gfoidl: Good question, which I can't answer.
But neither GCC nor Clang do this optimization.

src/Microsoft.Data/PrimitiveColumnContainer.cs
gfoidl: ThrowHelper is more a pattern than a concrete class. The motivation for that use is best described in [this comment from corefx](https://github.com/dotnet/corefx/blob/a1940826e89c4c333bca45459b25ba009e869ff6/src/Common/src/CoreLib/System/ThrowHelper.cs#L6-L36). Note: it makes only sense in lower-level APIs, not in user-code.
To see the effect of the ThrowHelper-pattern see [this SharpLab-demo](https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIGYACMhgYQYG8aHunGBLAHYYGAWQAUg4QDdsAGwCuMAJRcenaj00M+AMwZiZCmAwA8DAAwqNWmxgAWUCAHcGAmC4CCUAObz8MIQB5eQxAnQAlbAFvGABRBDAYAAcMPggBMQFsfwgdAzlFJSUAblUbMq1iAHYGQ0VS624AXwqK+m0hBgA5CU665Qr1LQBIXX1+0wsrGxsAFQdnAAkYWSSYKAA6eccnL19/IPkdMMjomDFtpZW1zfjElLSBPb8AjA3+korNL55q2oKYA1NC1qCCaO1iOQkExSAxLk5lqt1jQhr9GJDocQUHCFrsfC9DscIlEYmI7slUulngdhNh8TTpjMALwAPgY9h2rncDGpr2CoWJZ3JD3SYjp+1eWwgAGUMFBBN4xEUgTw2owAn4GMLKU96a9Bl9hhNuBUQU0gA==) (compare method `M` with `N` in the dissambly and see that `N` is straightforward, whilst `M` has all the exception code in it).

In essence this pattern moves the exception code out of the method (and possible out of hot-paths), thus resulting in  more compact and faster code.

For your question this means, if this project doesn't have a `ThrowHelper`-class you can create one.
Note also that a distinct class isn't needed to take advantage of moving out the exception code from "worker-code", as this can be done with any method (even local methods). But with a distinct class all exception-code is in one place and can be reused.

At the JIT-level this advantage is based on the fact that the JIT recognises methods that won't return (as throw-codepath will not return), so it doesn't inline these methods and moves it out of the hot-path.

src/Microsoft.Data/BaseColumn.cs
pgovind: I don't mind doing it, but does it really produce better code?

gfoidl: Not necessarily better code -- it should result in the same assembly code that the JIT produces.
But IMHO more readable C#-code. 

E.g. this can be avoided, as it is already done by the deconstruction.
https://github.com/dotnet/corefxlab/blob/e9f2bccdc570c2757da6e33c7ac913ecfb183062/src/Microsoft.Data/BaseColumn.cs#L100-L101

Or `nextValueAndBufferSortIndex.bufferSortIndex` looks quite long, in contrast to just `bufferSortIndex`.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -64,24 +67,26 @@ protected set
...
-                                                            GetValueAtBuffer<T> getValueAtBuffer,
+                                                            GetValueAndBufferSortIndexAtBuffer<T> getValueAndBufferSortIndexAtBuffer,
                                                             GetBufferLengthAtIndex getBufferLengthAtIndex)
         {
             while (heapOfValueAndListOfTupleOfSortAndBufferIndex.Count > 0)
             {
-                KeyValuePair<T, List<Tuple<int, int>>> minElement = heapOfValueAndListOfTupleOfSortAndBufferIndex.ElementAt(0);
-                List<Tuple<int, int>> tuplesOfSortAndBufferIndex = minElement.Value;
-                Tuple<int, int> sortAndBufferIndex;
+                KeyValuePair<T, List<ValueTuple<int, int>>> minElement = heapOfValueAndListOfTupleOfSortAndBufferIndex.ElementAt(0);
```
gfoidl:
```suggestion
                KeyValuePair<T, List<(int, int)>> minElement = heapOfValueAndListOfTupleOfSortAndBufferIndex.ElementAt(0);
```
Seems more readable to me. Other places as well.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -67,7 +67,7 @@ protected set
         /// <returns></returns>
         public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
 
-        public virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
+        public virtual BaseColumn CloneAndAppendNulls(long numberOfNulls, BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
```
pgovind: I'll update this as an argument to the `Clone` method above

src/Microsoft.Data/DataFrame.cs
```diff
@@ -133,13 +133,124 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
                 BaseColumn oldColumn = Column(i);
-                BaseColumn newColumn = oldColumn.CloneAndAppendNulls(sortIndices, !ascending);
+                BaseColumn newColumn = oldColumn.CloneAndAppendNulls(oldColumn.NullCount, sortIndices, !ascending);
                 Debug.Assert(newColumn.NullCount == oldColumn.NullCount);
                 newColumns.Add(newColumn);
             }
             return new DataFrame(newColumns);
         }
 
+        public enum JoinAlgorithm
```
eerhardt: This shouldn't be a nested enum. [https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/nested-types)

src/Microsoft.Data/DataFrame.cs
```diff
@@ -133,13 +133,124 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
+        {
+            Left,
+            Right,
+            Outer,
+            Inner
+        }
+
+        public DataFrame Join(DataFrame other, string leftSuffix = "_left", string rightSuffix = "_right", JoinAlgorithm joinAlgorithm = JoinAlgorithm.Left)
```
eerhardt: Don't you need something to join "on"? For example, Join on `left.CustomerId == right.Id`.

eerhardt: I see in Pandas, the default is blank for left and right suffix. Do we always want a suffix?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -133,13 +133,124 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
+        {
+            DataFrame ret = new DataFrame();
+            if (joinAlgorithm == JoinAlgorithm.Left)
+            {
+                for (int i = 0; i < ColumnCount; i++)
+                {
+                    BaseColumn newColumn = Column(i).Clone();
+                    newColumn.Name += leftSuffix;
```
eerhardt: Shouldn't the suffix only be applied if the names conflict between the two data frames?

src/Microsoft.Data/DataFrame.cs
pgovind: Nope. This is a straight index on index join in Pandas speak. Maybe you are thinking of `Merge`?

pgovind: Yup. Because pandas will fail if you try to join without the suffix and there are duplicate column names

src/Microsoft.Data/DataFrame.cs
pgovind: Oh you're right. I'll change that

src/Microsoft.Data/DataFrame.cs
eerhardt: https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.join.html

> on : str, list of str, or array-like, optional
Column or index level name(s) in the caller to join on the index in other, otherwise joins index-on-index. If multiple values given, the other DataFrame must have a MultiIndex. Can pass an array as the join key if it is not already contained in the calling DataFrame. Like an Excel VLOOKUP operation.

pgovind: Right, I should've explained better. So, we can't support `on` yet because the `DataFrame`  does not understand `index` yet. In Pandas, each dataframe has an index associated with it. So I can do this:
`df = pd.DataFrame({'key': ['K0', 'K1', 'K2', 'K3', 'K4', 'K5'],
...                    'A': ['A0', 'A1', 'A2', 'A3', 'A4',
other = pd.DataFrame({'key': ['K0', 'K3', 'K5'],
...                       'B': ['B0', 'B3', 'B5']})
df.join(other.set_index('key'), on='key')`
where it would use the `key` column in `other` as its index. For our prototype here(as a v1.0), I thought using the row numbers as an implicit index is good enough? Thoughts?

src/Microsoft.Data/DataFrame.cs
pgovind: Done

src/Microsoft.Data/DataFrame.cs
pgovind: Done

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -65,9 +65,7 @@ protected set
         /// <param name="mapIndices"></param>
         /// <param name="invertMapIndices"></param>
         /// <returns></returns>
-        public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
-
-        public virtual BaseColumn CloneAndAppendNulls(BaseColumn mapIndices = null, bool invertMapIndices = false) => throw new NotImplementedException();
+        public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
```
eerhardt: Not necessarily needs to be done in this PR, but you may want to consider creating a `CloneOptions` class that holds all these options. That way we can add future options without breaking the API.

Unless you think this is all the options `Clone` will ever need.

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -57,6 +57,16 @@ public IList<object> GetRow(long rowIndex)
...
 
+        public void SetColumnName(BaseColumn column, string newName)
+        {
+            string currentName = column.Name;
+            int currentIndex = _columnNameToIndexDictionary[currentName];
+            column.Name = newName;
+            _columnNameToIndexDictionary.Remove(currentName);
+            _columnNameToIndexDictionary.Add(newName, currentIndex);
+        }
+
```
eerhardt: (nit) double blank lines. One of them can be removed.

src/Microsoft.Data/BaseColumn.cs
eerhardt: Can `numberOfNullsToAppend` be inferred by the column being cloned? Or does it always have to be specified?

src/Microsoft.Data/DataFrame.cs
eerhardt: If we need to support `index` before implementing `on`, then let's continue as-is for now without the `on` support.

eerhardt: Do we want to match the Pandas behavior of throwing if there are conflicts by default? Or is defaulting to `_left` and `_right` suffixes better behavior?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -598,5 +598,66 @@ public void TestPrimitiveColumnSort()
...
+        {
+            DataFrame left = MakeDataFrameWithAllColumnTypes(10);
+            DataFrame right = MakeDataFrameWithAllColumnTypes(5);
+
+            // Tests with right.RowCount < left.RowCount
+            // Left join
+            DataFrame join = left.Join(right);
+            Assert.Equal(join.RowCount, left.RowCount);
+            Assert.Equal(join.ColumnCount, left.ColumnCount + right.ColumnCount);
+            Assert.Null(join["Int_right"][6]);
```
eerhardt: I would have expected better validation here than just checking a single value. Shouldn't we be validating that the whole DataFrame is correct?

src/Microsoft.Data/BaseColumn.cs
pgovind: Nope, it cannot be inferred. That was my first instinct as well. Sometimes(like in joins and merges), it depends on the number of rows in another table.

src/Microsoft.Data/DataFrame.cs
pgovind: Personally I think defaulting to `_left` and `_right` is better behavior. I don't have a strong opinion about this though. It just feels like a place where we can do better than throwing an exception. I did however just realize that I had to change the `if` to a `while` though.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
pgovind: Done.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -9,6 +9,14 @@
 
 namespace Microsoft.Data
 {
+    public enum JoinAlgorithm
+    {
+        Left,
+        Right,
+        Outer,
```
eerhardt: This might be my old T-SQL experience hindering me, but "Left" and "Right" are also "Outer" joins, so having an "Outer" option confuses me. Would it be better if we called this `FullOuter`?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -133,13 +141,125 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+                for (int i = 0; i < other.ColumnCount; i++)
+                {
+                    BaseColumn newColumn = other.Column(i).Clone(mapIndices);
+                    SetSuffixForDuplicatedColumnNames(ret, newColumn, leftSuffix, rightSuffix);
+                    ret.InsertColumn(ret.ColumnCount, newColumn);
+                }
+            }
+            return ret;
```
eerhardt: If someone specifies an invalid `joinAlgorithm` value, we just return an empty DataFrame. Is that intended behavior?

src/Microsoft.Data/DataFrame.cs
pgovind: Sure. I just went with "Outer" because Pandas calls it that

src/Microsoft.Data/DataFrame.cs

PR #2692, opened 2019-07-01
Implementations for Merge, IDataView and ReadCsv

pgovind:
Should go in after https://github.com/dotnet/corefxlab/pull/2687

More work towards https://github.com/dotnet/corefx/issues/26845

### DIFF
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -7,6 +7,7 @@ using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
 using System.Text;
+using Microsoft.Collections.Extensions;
 
 namespace Microsoft.Data
 {
@@ -60,6 +61,12 @@ namespace Microsoft.Data
         }
 
         /// <summary>
+        /// Called internally from Merge
+        /// </summary>
+        /// <param name="length"></param>
+        public virtual void Resize(long length) => throw new NotImplementedException();
+
+        /// <summary>
         /// Clone column to produce a copy potentially changing the order by supplying mapIndices and an invert flag
         /// </summary>
         /// <param name="mapIndices"></param>
@@ -68,6 +75,8 @@ namespace Microsoft.Data
         public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
 
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
+
+        public virtual MultiValueDictionary<TKey, long> HashColumnValues<TKey>() => throw new NotImplementedException();
 
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
--- a/src/Microsoft.Data/DataFrame.IO.cs
+++ b/src/Microsoft.Data/DataFrame.IO.cs
@@ -1,0 +1,261 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Linq;
+using System.Text;
+
+namespace Microsoft.Data
+{
+
+    public partial class DataFrame
+    {
+        private static Type GuessKind(int col, List<string[]> read)
+        {
+            Type res = typeof(string);
+            int nbline = 0;
+            foreach (var line in read)
+            {
+                if (col >= line.Length)
+                    throw new FormatException(string.Format(Strings.LessColumnsThatExpected, nbline + 1)); // Will this really work?
+
+                var val = line[col];
+                try
+                {
+                    bool.Parse(val);
+                    res = DetermineType(nbline == 0, typeof(bool), res);
+                    ++nbline;
+                    continue;
+                }
+                catch (Exception)
+                {
+                    if (string.IsNullOrEmpty(val))
+                    {
+                        res = DetermineType(nbline == 0, typeof(bool), res);
+                        continue;
+                    }
+                }
+                try
+                {
+                    float.Parse(val);
+                    res = DetermineType(nbline == 0, typeof(float), res);
+                    ++nbline;
+                    continue;
+                }
+                catch (Exception)
+                {
+                    if (string.IsNullOrEmpty(val))
+                    {
+                        res = DetermineType(nbline == 0, typeof(float), res);
+                        continue;
+                    }
+                }
+                res = DetermineType(nbline == 0, typeof(string), res);
+                ++nbline;
+            }
+            return res;
+        }
+
+        private static Type DetermineType(bool first, Type suggested, Type previous)
+        {
+            if (first)
+                return suggested;
+            else
+                return MaxKind(suggested, previous);
+        }
+
+        private static Type MaxKind(Type a, Type b)
+        {
+            if (a == typeof(string) || b == typeof(string))
+                return typeof(string);
+            if (a == typeof(float) || b == typeof(float))
+                return typeof(float);
+            if (a == typeof(bool) || b == typeof(bool))
+                return typeof(bool);
+            return typeof(string);
+        }
+
+        /// <summary>
+        /// Reads a text file as a IDataView.
+        /// Follows pandas API.
+        /// </summary>
+        /// <param name="filename">filename</param>
+        /// <param name="sep">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="names">column names (can be empty)</param>
+        /// <param name="dtypes">column types (can be empty)</param>
+        /// <param name="nrows">number of rows to read</param>
+        /// <param name="guess_rows">number of rows used to guess types</param>
+        /// <param name="encoding">text encoding</param>
+        /// <param name="index">add one column with the row index</param>
+        /// <returns>DataFrame</returns>
+        public static DataFrame ReadCsv(string filename,
+                                char sep = ',', bool header = true,
+                                string[] names = null, Type[] dtypes = null,
+                                int nrows = -1, int guess_rows = 10,
+                                Encoding encoding = null, bool index = false)
+        {
+            return ReadStream(() => new StreamReader(filename, encoding ?? Encoding.ASCII),
+                              sep: sep, header: header, names: names, dtypes: dtypes, nrows: nrows,
+                              guess_rows: guess_rows, index: index);
+        }
+
+        private delegate StreamReader FunctionCreateStreamReader();
+
+        /// <summary>
+        /// Reads a text file as a IDataView.
+        /// Follows pandas API.
+        /// </summary>
+        /// <param name="createStream">function which creates a stream</param>
+        /// <param name="sep">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="names">column names (can be empty)</param>
+        /// <param name="dtypes">column types (can be empty)</param>
+        /// <param name="nrows">number of rows to read</param>
+        /// <param name="guess_rows">number of rows used to guess types</param>
+        /// <param name="index">add one column with the row index</param>
+        /// <returns>DataFrame</returns>
+        private static DataFrame ReadStream(FunctionCreateStreamReader createStream,
+                                char sep = ',', bool header = true,
+                                string[] names = null, Type[] dtypes = null,
+                                long nrows = -1, int guess_rows = 10, bool index = false)
+        {
+            var lines = new List<string[]>();
+            long rowline = 0;
+
+            // First pass: schema and number of rows.
+            using (var st = createStream())
+            {
+                string line = st.ReadLine();
+                int nbline = 0;
+                while (line != null && (nrows == -1 || rowline < nrows))
+                {
+                    var spl = line.Split(sep);
+                    if (header && nbline == 0)
+                    {
+                        if (names == null)
+                            names = spl;
+                    }
+                    else
+                    {
+                        ++rowline;
+                        if (lines.Count < guess_rows)
+                            lines.Add(spl);
+                    }
+                    ++nbline;
+                    line = st.ReadLine();
+                }
+            }
+
+            if (lines.Count == 0)
+                throw new FormatException(Strings.EmptyFile);
+            int numCol = lines.Select(c => c.Length).Max();
+            var df = new DataFrame();
+
+            // Guesses types and adds columns.
+            for (int i = 0; i < numCol; ++i)
+            {
+                Type kind = GuessKind(i, lines);
+                switch (kind)
+                {
+                    case Type boolType when boolType == typeof(bool):
+                        BaseColumn boolColumn = new PrimitiveColumn<bool>(names[i]);
+                        df.InsertColumn(i, boolColumn);
+                        break;
+                    case Type byteType when byteType == typeof(byte):
+                        BaseColumn byteColumn = new PrimitiveColumn<byte>(names[i]);
+                        df.InsertColumn(i, byteColumn);
+                        break;
+                    case Type charType when charType == typeof(char):
+                        BaseColumn charColumn = new PrimitiveColumn<char>(names[i]);
+                        df.InsertColumn(i, charColumn);
+                        break;
+                    case Type decimalType when decimalType == typeof(decimal):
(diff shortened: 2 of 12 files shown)


src/Microsoft.Data/BaseColumn.cs
```diff
@@ -59,18 +60,20 @@ protected set
             get => GetValue(startIndex, length);
         }
 
+        public virtual void Resize(long length) => throw new NotImplementedException();
```
pgovind: Add a comment about why this is needed

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+    public partial class DataFrame
+    {
+        private static Type GuessKind(int col, List<string[]> read)
+        {
+            Type res = typeof(string);
+            int nbline = 0;
+            foreach (var line in read)
+            {
+                if (col >= line.Length)
+                    throw new FormatException(string.Format(Strings.LessColumnsThatExpected, nbline + 1)); // Will this really work?
```
pgovind: The question is w.r.t localization

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+            if (lines.Count == 0)
+                throw new FormatException(Strings.EmptyFile);
+            int numCol = lines.Select(c => c.Length).Max();
+            var df = new DataFrame();
+
+            // Guesses types and adds columns.
+            for (int i = 0; i < numCol; ++i)
+            {
+                Type kind = GuessKind(i, lines);
+                switch (kind)
```
pgovind: Support exists for now to create the different types of columns, but `GuessKind` itself only returns one of `string, bool, float` at the moment.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -260,6 +261,305 @@ public DataFrame Join(DataFrame other, string leftSuffix = "_left", string right
...
+                BaseColumn column = Column(i).Clone(emptyMap);
+                ret.InsertColumn(ret.ColumnCount, column);
+            }
+
+            for (int i = 0; i < other.ColumnCount; i++)
+            {
+                // Create empty columns
+                BaseColumn column = other.Column(i).Clone(emptyMap);
+                int index = ret._table.GetColumnIndex(column.Name);
+                if (index != -1)
```
pgovind: Refactor to use the SetSuffix... method

src/Microsoft.Data/StringColumn.cs
```diff
@@ -45,6 +45,17 @@ public StringColumn(string name, IEnumerable<string> values) : base(name, 0, typ
...
+            if (length < Length)
+                throw new ArgumentException(Strings.CannotResizeDown, nameof(length));
+
+            for (long i = Length; i < length; i++)
+            {
+                Append(null);
+            }
+        }
+
         public void Append(string value)
```
pgovind: Implementation of `HashColumnValues` for StringColumn is coming in the next patch

eerhardt:
@pgovind - can you rebase or merge this with the current master? I don't want to re-review the Join code.

pgovind:
> @pgovind - can you rebase or merge this with the current master? I don't want to re-review the Join code.

Done

src/Microsoft.Data/DataFrame.IO.cs
eerhardt: Yes, you can have placeholders in resources `{0}`.

---
In reply to: [299212735](https://github.com/dotnet/corefxlab/pull/2692#discussion_r299212735) [](ancestors = 299212735)

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+        /// <param name="sep">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="names">column names (can be empty)</param>
+        /// <param name="dtypes">column types (can be empty)</param>
+        /// <param name="nrows">number of rows to read</param>
+        /// <param name="guess_rows">number of rows used to guess types</param>
+        /// <param name="encoding">text encoding</param>
+        /// <param name="index">add one column with the row index</param>
+        /// <returns>DataFrame</returns>
+        public static DataFrame ReadCsv(string filename,
```
eerhardt: I almost wonder if these methods should be external to the `DataFrame`. That way we can ensure anyone can take any data (like from SQL, or elsewhere) and materialize a DataFrame.

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+                return suggested;
+            else
+                return MaxKind(suggested, previous);
+        }
+
+        private static Type MaxKind(Type a, Type b)
+        {
+            if (a == typeof(string) || b == typeof(string))
+                return typeof(string);
+            if (a == typeof(float) || b == typeof(float))
```
eerhardt: What about `double` or `int`, etc?

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -18,6 +18,11 @@
     <PackageReference Include="System.Memory" Version="4.5.1" />
     <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.1" />
     <PackageReference Include="System.Text.Encoding" Version="4.3.0" />
+    <PackageReference Include="Microsoft.ML" Version="1.1.0" />
```
eerhardt: Don't reference the whole `Microsoft.ML`.  But instead, `Microsoft.ML.DataView`.

I wonder if that functionality should be split into a separate assembly. That way the core DataFrame library doesn't need to depend on Microsoft.ML.

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -18,6 +18,11 @@
...
+  </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Experimental.Collections\Microsoft.Experimental.Collections.csproj" />
```
eerhardt: Do we really need to be dependent on this experimental library? Is it planning on shipping?

/cc @safern

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +744,68 @@ public void TestJoin()
             Assert.Equal(join["Int_right"][2], right["Int"][2]);
             VerifyJoin(join, left, right, JoinAlgorithm.Inner);
         }
+
+        [Fact]
```
eerhardt: We need tests for ReadCsv , IDataView.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +744,68 @@ public void TestJoin()
...
+        {
+            DataFrame left = MakeDataFrameWithAllColumnTypes(10);
+            DataFrame right = MakeDataFrameWithAllColumnTypes(5);
+
+            // Tests with right.RowCount < left.RowCount
+            // Left merge
+            DataFrame merge = left.Merge<int>(right, "Int", "Int");
+            Assert.Equal(10, merge.RowCount);
+            Assert.Equal(merge.ColumnCount, left.ColumnCount + right.ColumnCount);
+            Assert.Null(merge["Int_right"][6]);
```
eerhardt: We should be validating the whole resultant dataframe's values.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -45,6 +45,17 @@ public StringColumn(string name, IEnumerable<string> values) : base(name, 0, typ
...
         public override long NullCount => _nullCount;
 
+        public override void Resize(long length)
+        {
+            if (length < Length)
+                throw new ArgumentException(Strings.CannotResizeDown, nameof(length));
+
+            for (long i = Length; i < length; i++)
+            {
+                Append(null);
```
eerhardt: Is doing this one at a time the best approach?

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -147,5 +147,82 @@ public int GetColumnIndex(string columnName)
...
+            for (int i = 0; i < values.Length; i++)
+            {
+                string value = values[i];
+                BaseColumn column = Column(i);
+                bool parse;
+                switch (column.DataType)
+                {
+                    case Type boolType when boolType == typeof(bool):
+                        parse = bool.TryParse(value, out bool boolAppend);
+                        (column as PrimitiveColumn<bool>).Append(parse ? boolAppend : default);
```
eerhardt: It feels wrong to me that the DataFrameTable is responsible for parsing these values. It feels more tied to the CSV reading implementation.

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+                                string[] names = null, Type[] dtypes = null,
+                                int nrows = -1, int guess_rows = 10,
+                                Encoding encoding = null, bool index = false)
+        {
+            return ReadStream(() => new StreamReader(filename, encoding ?? Encoding.ASCII),
+                              sep: sep, header: header, names: names, dtypes: dtypes, nrows: nrows,
+                              guess_rows: guess_rows, index: index);
+        }
+
+        private delegate StreamReader FunctionCreateStreamReader();
```
eerhardt: Why not just use a `Func<StreamReader>`?

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+
+        /// <summary>
+        /// Reads a text file as a IDataView.
```
eerhardt: `as a IDataView`?

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+            int nbline = 0;
+            foreach (var line in read)
+            {
+                if (col >= line.Length)
+                    throw new FormatException(string.Format(Strings.LessColumnsThatExpected, nbline + 1)); // Will this really work?
+
+                var val = line[col];
+                try
+                {
+                    bool.Parse(val);
```
eerhardt: `TryParse` instead of `Parse` and an exception. Throwing exceptions in happy paths will cause perf degradation.

eerhardt: Are you planning on profiling the CSV parsing and seeing where any bottlenecks are?

---
In reply to: [299702122](https://github.com/dotnet/corefxlab/pull/2692#discussion_r299702122) [](ancestors = 299702122)

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+                        BaseColumn stringColumn = new StringColumn(names[i], 0);
+                        df.InsertColumn(i, stringColumn);
+                        break;
+                    default:
+                        throw new NotSupportedException();
+                }
+            }
+
+            // Fills values.
+            using (var st = createStream())
```
eerhardt: We already have ALL the values in memory, right? Why read them from the stream again?

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,261 @@
...
+                        ++rowline;
+                    }
+                    ++nbline;
+                    line = st.ReadLine();
+                }
+            }
+
+            if (index)
+            {
+                // TODO: Bug in InsertColumn. Can only insert at the end. Should support inserting in the middle
```
eerhardt: What's the plan for handling this TODO?

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
safern: I think he depends on MultiValueDictionary... the plan for experimental collections is that if a collection gets popular enough and might be useful for partners or the framework, we will move that collection into the product.

danmosemsft: I think MultiValueDictionary already got formally API reviewed (as part of Ian's internship). That was before my time, but assuming so then most likely it would only need a brief re-review. And I think it should likely go into CoreFX for 5.0. It seems a fairly basic, uncontroversial type.

eerhardt: Is the plan for MultiValueDictionary to support `netstandard2.0` or lower?

eerhardt: IMHO - for this case, using `Dictionary<TKey, ICollection<long>>` (or similar) would suffice, and it would alleviate the need for an added dependency. Making `DataFrame` dependent on a non-stable library affects how and when we could ship it in the future.

src/Microsoft.Data/DataFrame.IO.cs
pgovind: By external, do you mean an API such as 

`ReadCsv(filename, out DataFrame dataFrame)`?

src/Microsoft.Data/DataFrame.IO.cs
pgovind: Right, `MaxKind` only has 1 caller `DetermineType` and `DetermineKind` always passes in one of string, float or bool. Or do you mean to add support now to make it easier to parse `double`/`int` in the future?

src/Microsoft.Data/DataFrame.IO.cs
pgovind: Good catch. Left over from scratch work I was doing

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
pgovind: Ah, I was worried this would be the case. I can change to using a `Dictionary`.

src/Microsoft.Data/DataFrame.IO.cs
eerhardt: I mean not on the `DataFrame` class. The reasoning is that other code outside of the current assembly should be able to load `DataFrame` objects from different sources - SQL, Parquet, in-memory objects, etc. If `ReadCsv` is on the `DataFrame` class itself, it has access to `private` fields/methods that other code won't be able to use. This means it gets `special` treatment. If someone else wanted to write a "create a DataFrame from some data source" method, would they be able to do it?

src/Microsoft.Data/DataFrame.IO.cs
eerhardt: Maybe the question is - why aren't we supporting `double` and `int` types? Why just `float`?

src/Microsoft.Data/DataFrame.IO.cs
pgovind: Yup. Sometime soon. There's a whole bunch of perf tests I want to do.

src/Microsoft.Data/DataFrame.IO.cs
pgovind: Ah, that's because we only read in the first 10 lines in the csv file and try to guess the types from it. Since the first 10 lines of column A could be `int`, but everything else after that a `float`, I chose `float` as the common numeric type. At some point, it'd probably be worth it to have a strongly typed API that specifies the column types

src/Microsoft.Data/DataFrame.IO.cs
pgovind: No, we don't. `lines` only has a length of `guess_rows` (defaults to 10 rows)

src/Microsoft.Data/DataFrame.IO.cs
pgovind: Filed https://github.com/dotnet/corefxlab/issues/2695

pgovind:
Considering the comments, this will likely take me a bit of time to refactor. I'm going to close this PR for now and make 3 separate ones addressing the comments for Merge, IDataView and ReadCsv respectively.

PR #2694, opened 2019-07-08
WIP: Roundtrip to/from RecordBatch through a DataFrame [WIP]

pgovind:
Do Not Merge: Just to get eyes from @eerhardt

### DIFF
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -6,6 +6,9 @@ using System;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
+using Apache.Arrow;
+using Apache.Arrow.Types;
+using Microsoft.Collections.Extensions;
 
 namespace Microsoft.Data
 {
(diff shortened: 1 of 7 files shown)


src/Microsoft.Data/DataFrame.cs
```diff
@@ -6,6 +6,9 @@
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Text;
+using Apache.Arrow;
+using Apache.Arrow.Types;
+using Microsoft.Collections.Extensions;
```
pgovind: Ignore this. VS keeps adding these for some reason

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +36,273 @@ public DataFrame(IList<BaseColumn> columns)
...
+                        ReadOnlyMemory<byte> ulongNullBitMapBuffer = (arrowArray as PrimitiveArray<ulong>).NullBitmapBuffer.Memory;
+                        dataFrameMemory = new Memory<byte>(new byte[ulongValueBuffer.Length]);
+                        nullBitMapMemory = new Memory<byte>(new byte[ulongNullBitMapBuffer.Length]);
+                        copied = ulongValueBuffer.TryCopyTo(dataFrameMemory) && ulongNullBitMapBuffer.TryCopyTo(nullBitMapMemory);
+                        if (copied)
+                        {
+                            dataFrameColumn = new PrimitiveColumn<ulong>(field.Name, dataFrameMemory, nullBitMapMemory, arrowArray.Length);
+                        }
+                        break;
+                    case ArrowTypeId.Binary:
```
pgovind: I *think* this case is wrong. Binary has 3 buffers in Arrow, so I need to look into this further

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +36,273 @@ public DataFrame(IList<BaseColumn> columns)
...
+                fieldIndex++;
+            }
+        }
+
+        public IEnumerable<RecordBatch> AsArrowRecordBatch()
+        {
+            Apache.Arrow.Schema.Builder schemaBuilder = new Apache.Arrow.Schema.Builder();
+            Field.Builder fieldBuilder = new Field.Builder();
+            // TODO: Sanity check. Some column types are not Arrow Compatible
+            // Derived column types will NOT be supported
```
pgovind: Ignore this comment. Derived types will be supported

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +36,273 @@ public DataFrame(IList<BaseColumn> columns)
...
+                            arrays.Add(ulongColumn.AsArrowArray(n));
+                            break;
+                        case PrimitiveColumn<ushort> ushortColumn:
+                            arrays.Add(ushortColumn.AsArrowArray(n));
+                            break;
+                        default:
+                            throw new NotImplementedException(nameof(column.DataType));
+                    }
+                }
+                yield return new RecordBatch(schema, arrays, 0);
```
pgovind: Ignore the 0 here. I didn't spend much time here thinking about how to split up each `PrimitiveColumn<T>`'s `DataFrameBuffer`s such that there is a 1x1 match with their `NullBitMapBuffer`s. For this test, each column in the DataFrame becomes an Arrow `Array`

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -15,10 +15,22 @@
 
   <ItemGroup>
     <PackageReference Include="System.Buffers" Version="4.5.0" />
-    <PackageReference Include="System.Memory" Version="4.5.1" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.1" />
+    <PackageReference Include="System.Memory" Version="4.5.2" />
```
pgovind: Ignore the version changes

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -15,10 +15,22 @@
...
+    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.2" />
     <PackageReference Include="System.Text.Encoding" Version="4.3.0" />
+    <PackageReference Include="Microsoft.ML" Version="1.1.0" />
   </ItemGroup>
+
+  <ItemGroup>
+    <ProjectReference Include="..\Microsoft.Experimental.Collections\Microsoft.Experimental.Collections.csproj" />
+    <ProjectReference Include="C:\Users\prgovi\Desktop\Work\arrow\csharp\src\Apache.Arrow\Apache.Arrow.csproj" />
```
pgovind: Ignore this. This was written before 0.14 came out on Nuget

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -77,6 +77,13 @@ public PrimitiveColumnContainer(IEnumerable<T> values)
             }
         }
 
+        internal PrimitiveColumnContainer(Memory<byte> memory, Memory<byte> nullBitMap, int length)
+        {
+            Buffers.Add(new DataFrameBuffer<T>(memory, length));
+            NullBitMapBuffers.Add(new DataFrameBuffer<byte>(nullBitMap, length));
+            Length += length;
```
eerhardt: This is a constructor, right? `Length` should be `0` to start, so this should be:

`Length = length`.

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -17,6 +19,47 @@ public partial class PrimitiveColumn<T> : BaseColumn
...
+            Debug.Assert(i < _columnContainer.Buffers.Count);
+            DataFrameBuffer<T> values = _columnContainer.Buffers[i];
+            DataFrameBuffer<byte> nulls = _columnContainer.NullBitMapBuffers[i];
+            ArrowBuffer valueBuffer = new ArrowBuffer(values.Memory);
+            // The following line is wrong, but C# Arrow does not have good null support yet
+            ArrowBuffer nullBuffer = nulls.Length == 0 ? ArrowBuffer.Empty : new ArrowBuffer(nulls.Memory);
+            switch (this)
+            {
+                case PrimitiveColumn<bool> boolColumn:
+                    return new BooleanArray(valueBuffer, nullBuffer, values.Length, -1, 0);
```
eerhardt: Why `-1` for NullCount?

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -44,6 +44,14 @@ public DataFrameBuffer(int numberOfValues = 8)
             Memory = new byte[numberOfValues * _size];
         }
 
+        public DataFrameBuffer(Memory<byte> memory, int length)
+        {
+            _size = Unsafe.SizeOf<T>();
+            //Debug.Assert(length * _size == memory.Length); // Wont work for the nullBitMapBuffer
```
eerhardt: Should this just be deleted then?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +36,273 @@ public DataFrame(IList<BaseColumn> columns)
...
+                Memory<byte> dataFrameMemory, nullBitMapMemory;
+                bool copied;
+                switch (fieldType.TypeId)
+                {
+                    case ArrowTypeId.Boolean:
+                        ReadOnlyMemory<byte> valueBuffer = (arrowArray as BooleanArray).ValueBuffer.Memory;
+                        ReadOnlyMemory<byte> nullBitMapBuffer = (arrowArray as BooleanArray).NullBitmapBuffer.Memory;
+                        dataFrameMemory = new Memory<byte>(new byte[valueBuffer.Length]);
+                        nullBitMapMemory = new Memory<byte>(new byte[nullBitMapBuffer.Length]);
+                        copied = valueBuffer.TryCopyTo(dataFrameMemory) && nullBitMapBuffer.TryCopyTo(nullBitMapMemory);
```
eerhardt: I would have hoped we could do this without a copy....

src/Microsoft.Data/DataFrameBuffer.cs
pgovind: Yup yup, when this patch is really ready to be merged, I'll delete this

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: Because we only store the NullCount for an entire column, we don't know the NullCount in each `DataFrameBuffer`. Similar to the behavior we get in Arrow when we `Slice`.(https://issues.apache.org/jira/browse/ARROW-5634)

src/Microsoft.Data/DataFrame.cs
pgovind: Yup, I do plan to support that soon(https://github.com/dotnet/corefxlab/issues/2697). The problem is that Arrow exposes only a ReadOnlyMemory<byte>

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +36,273 @@ public DataFrame(IList<BaseColumn> columns)
...
+            // Assuming that all columns are Arrow compatible 
+            int numberOfRecordBatches = 0;
+            for (int i = 0; i < numberOfColumns; i++)
+            {
+                BaseColumn column = Column(i);
+                switch (column)
+                {
+                    case PrimitiveColumn<bool> boolColumn:
+                        numberOfRecordBatches = boolColumn.NumberOfBuffers;
+                        Field boolField = fieldBuilder.Name(boolColumn.Name).Nullable(true).DataType(BooleanType.Default).Build();
```
eerhardt: Why are they always `Nullable`? What if this DataFrame doesn't contain any nulls in that column? Should it still be Nullable?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +36,273 @@ public DataFrame(IList<BaseColumn> columns)
...
+                        schemaBuilder.Field(boolField);
+                        break;
+                    case PrimitiveColumn<byte> byteColumn:
+                    case PrimitiveColumn<char> charColumn:
+                    case PrimitiveColumn<decimal> decimalColumn:
+                    case PrimitiveColumn<sbyte> sbyteColumn:
+                        throw new NotImplementedException(nameof(byteColumn.DataType));
+                    case PrimitiveColumn<double> doubleColumn:
+                        numberOfRecordBatches = doubleColumn.NumberOfBuffers;
+                        Field doubleField = fieldBuilder.Name(doubleColumn.Name).Nullable(true).DataType(DoubleType.Default).Build();
```
eerhardt: I would think a simple `new Field(....)` would suffice here. Why go through the overhead/complexity of using the Field.Builder?

src/Microsoft.Data/DataFrame.cs
pgovind: It need not be `Nullable`. I was just hacking in this patch and testing Null support in C# Arrow

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +36,273 @@ public DataFrame(IList<BaseColumn> columns)
...
+                        Field boolField = fieldBuilder.Name(boolColumn.Name).Nullable(true).DataType(BooleanType.Default).Build();
+                        schemaBuilder.Field(boolField);
+                        break;
+                    case PrimitiveColumn<byte> byteColumn:
+                    case PrimitiveColumn<char> charColumn:
+                    case PrimitiveColumn<decimal> decimalColumn:
+                    case PrimitiveColumn<sbyte> sbyteColumn:
+                        throw new NotImplementedException(nameof(byteColumn.DataType));
+                    case PrimitiveColumn<double> doubleColumn:
+                        numberOfRecordBatches = doubleColumn.NumberOfBuffers;
```
eerhardt: The fact that you need to cast to a `PrimitiveColumn<double>` in order to access `NumberOfBuffers` kind of smells here.

Maybe `BaseColumn` should have a `ToArrow` method?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +746,87 @@ public void TestJoin()
...
+            Field fInt = new Field.Builder().Name("Int").DataType(Int32Type.Default).Build();
+            Field fFloat = new Field.Builder().Name("Float").DataType(FloatType.Default).Build();
+            Schema schema = new Schema.Builder().Field(fInt).Field(fFloat).Build();
+
+            var intArrayBuilder = new Int32Array.Builder().Append(5).Append(10).Append(5).Append(10);
+            Int32Array intArray = intArrayBuilder.Build();
+            var floatArrayBuilder = new FloatArray.Builder().Append(5.0f).Append(10.0f).Append(5.0f).Append(10.0f);
+            FloatArray floatArray = floatArrayBuilder.Build();
+            List<IArrowArray> arrowArrays = new List<IArrowArray>() { intArray, floatArray };
+            RecordBatch recordBatch = new RecordBatch(schema, arrowArrays, 4);
```
eerhardt: (nit) you can use the RecordBatch.Builder API here. It would make this a little more concise.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +746,87 @@ public void TestJoin()
...
+
+            DataFrame df = new DataFrame(recordBatch);
+            DataFrame df2 = new DataFrame(recordBatch);
+
+            IEnumerable<RecordBatch> recordBatches = df2.AsArrowRecordBatch();
+
+            foreach (RecordBatch batch in recordBatches)
+            {
+                Assert.Equal(batch.ColumnCount, recordBatch.ColumnCount);
```
eerhardt: Maybe we could lift the comparison code from the Arrow library?

[https://github.com/apache/arrow/blob/master/csharp/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs#L46](https://github.com/apache/arrow/blob/master/csharp/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs#L46)

Or possibly another way of verifying would be to write the RecordBatch instances to a stream, and then compare that the streams are identical.  Although, debugging issues in that way would be hard.

src/Microsoft.Data/DataFrame.cs
pgovind: Yup yup, I agree. I realized that too late and since this was just a hack patch, I just put it up.

src/Microsoft.Data/DataFrame.cs
pgovind: Oh, yea you're right.

pgovind:
Closing this. Have enough data to make a clean patch now.

PR #2696, opened 2019-07-09
GroupBy implementation

pgovind:

### DIFF
--- a/src/Microsoft.Data/BaseColumn.Computations.cs
+++ b/src/Microsoft.Data/BaseColumn.Computations.cs
@@ -33,7 +33,17 @@ namespace Microsoft.Data
             throw new NotImplementedException();
         }
 
+        public virtual void CumulativeMax(IEnumerable<long> rows)
+        {
+            throw new NotImplementedException();
+        }
+
         public virtual void CumulativeMin()
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual void CumulativeMin(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
@@ -43,7 +53,17 @@ namespace Microsoft.Data
             throw new NotImplementedException();
         }
 
+        public virtual void CumulativeProduct(IEnumerable<long> rows)
+        {
+            throw new NotImplementedException();
+        }
+
         public virtual void CumulativeSum()
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual void CumulativeSum(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
@@ -53,7 +73,17 @@ namespace Microsoft.Data
             throw new NotImplementedException();
         }
 
+        public virtual object Max(IEnumerable<long> rows)
+        {
+            throw new NotImplementedException();
+        }
+
         public virtual object Min()
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual object Min(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
@@ -63,7 +93,17 @@ namespace Microsoft.Data
             throw new NotImplementedException();
         }
 
+        public virtual object Product(IEnumerable<long> rows)
+        {
+            throw new NotImplementedException();
+        }
+
         public virtual object Sum()
+        {
+            throw new NotImplementedException();
+        }
+
+        public virtual object Sum(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
--- a/src/Microsoft.Data/BaseColumn.Computations.cs
+++ b/src/Microsoft.Data/BaseColumn.Computations.cs
@@ -12,107 +12,165 @@ namespace Microsoft.Data
 {
     public abstract partial class BaseColumn
     {
-        #region Computations
+        /// <summary>
+        /// Updates each numeric element with its absolute numeric value
+        /// </summary>
         public virtual void Abs()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns whether all the elements are True
+        /// </summary>
         public virtual bool All()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns whether any element is True
+        /// </summary>
         public virtual bool Any()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Updates each element with its cumulative maximum
+        /// </summary>
         public virtual void CumulativeMax()
         {
             throw new NotImplementedException();
         }
 
-        public virtual void CumulativeMax(IEnumerable<long> rows)
+        /// <summary>
+        /// Updates column values at rowIndices with its cumulative rowIndices maximum
+        /// </summary>
+        public virtual void CumulativeMax(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Updates each element with its cumulative minimum
+        /// </summary>
         public virtual void CumulativeMin()
         {
             throw new NotImplementedException();
         }
 
-        public virtual void CumulativeMin(IEnumerable<long> rows)
+        /// <summary>
+        /// Updates column values at rowIndices with its cumulative rowIndices minimum
+        /// </summary>
+        public virtual void CumulativeMin(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Updates each element with its cumulative product
+        /// </summary>
         public virtual void CumulativeProduct()
         {
             throw new NotImplementedException();
         }
 
-        public virtual void CumulativeProduct(IEnumerable<long> rows)
+        /// <summary>
+        /// Updates column values at rowIndices with its cumulative rowIndices product
+        /// </summary>
+        public virtual void CumulativeProduct(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Updates each element with its cumulative sum
+        /// </summary>
         public virtual void CumulativeSum()
         {
             throw new NotImplementedException();
         }
 
-        public virtual void CumulativeSum(IEnumerable<long> rows)
+        /// <summary>
+        /// Updates column values at rowIndices with its cumulative rowIndices sum
+        /// </summary>
+        public virtual void CumulativeSum(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the maximum of the values in the column
+        /// </summary>
         public virtual object Max()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the maximum of the values at rowIndices
+        /// </summary>
         public virtual object Max(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the minimum of the values in the column
+        /// </summary>
         public virtual object Min()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the minimum of the values at the rowIndices
+        /// </summary>
         public virtual object Min(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the product of the values in the column
+        /// </summary>
         public virtual object Product()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the product of the values at the rowIndices
+        /// </summary>
         public virtual object Product(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the sum of the values in the column
+        /// </summary>
         public virtual object Sum()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the sum of the values at the rowIndices
+        /// </summary>
         public virtual object Sum(IEnumerable<long> rows)
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Calls Math.Round on each value in a column
+        /// </summary>
         public virtual void Round()
         {
             throw new NotImplementedException();
         }
 
-        #endregion
     }
 }
--- a/src/Microsoft.Data/BaseColumn.Computations.cs
+++ b/src/Microsoft.Data/BaseColumn.Computations.cs
@@ -111,7 +111,7 @@ namespace Microsoft.Data
         /// <summary>
         /// Returns the maximum of the values at rowIndices
         /// </summary>
-        public virtual object Max(IEnumerable<long> rows)
+        public virtual object Max(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
@@ -127,7 +127,7 @@ namespace Microsoft.Data
         /// <summary>
         /// Returns the minimum of the values at the rowIndices
         /// </summary>
-        public virtual object Min(IEnumerable<long> rows)
+        public virtual object Min(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
@@ -143,7 +143,7 @@ namespace Microsoft.Data
         /// <summary>
         /// Returns the product of the values at the rowIndices
         /// </summary>
-        public virtual object Product(IEnumerable<long> rows)
+        public virtual object Product(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
@@ -159,7 +159,7 @@ namespace Microsoft.Data
         /// <summary>
         /// Returns the sum of the values at the rowIndices
         /// </summary>
-        public virtual object Sum(IEnumerable<long> rows)
+        public virtual object Sum(IEnumerable<long> rowIndices)
         {
             throw new NotImplementedException();
         }
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -3,7 +3,6 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
@@ -20,6 +19,16 @@ namespace Microsoft.Data
             Length = length;
             Name = name;
             DataType = type;
+        }
+
+        private DataFrame _dataFrame;
+        internal DataFrame DataFrame
+        {
+            get => _dataFrame;
+            set
+            {
+                _dataFrame = value;
+            }
         }
 
         private long _length;
@@ -75,7 +84,9 @@ namespace Microsoft.Data
 
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
 
-        public virtual Dictionary<TKey, ICollection<long>> HashColumnValues<TKey>() => throw new NotImplementedException();
+        public virtual Dictionary<TKey, ICollection<long>> GroupColumnValues<TKey>() => throw new NotImplementedException();
+
+        public virtual GroupBy GroupBy(int columnIndex) => throw new NotImplementedException();
 
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -7,7 +7,6 @@ using System.Collections;
 using System.Collections.Generic;
 using System.Diagnostics;
 using System.Linq;
-using System.Text;
 
 namespace Microsoft.Data
 {
@@ -61,6 +60,12 @@ namespace Microsoft.Data
         }
 
         /// <summary>
+        /// Called internally from Merge
+        /// </summary>
+        /// <param name="length"></param>
+        public virtual void Resize(long length) => throw new NotImplementedException();
+
+        /// <summary>
         /// Clone column to produce a copy potentially changing the order by supplying mapIndices and an invert flag
         /// </summary>
         /// <param name="mapIndices"></param>
@@ -70,6 +75,8 @@ namespace Microsoft.Data
         public virtual BaseColumn Clone(IEnumerable<long> mapIndices, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
 
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
+
+        public virtual Dictionary<TKey, ICollection<long>> HashColumnValues<TKey>() => throw new NotImplementedException();
 
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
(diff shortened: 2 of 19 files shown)


tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -23,40 +23,80 @@ public static DataFrame MakeDataFrameWithTwoColumns(int length)
             return dataFrame;
         }
 
-        public static DataFrame MakeDataFrameWithAllColumnTypes(int length)
+        public static DataFrame MakeDataFrameForGroupBy(int length) => MakeDataFrameWithAllColumnTypes(length, true);
```
pgovind: Just adding this for now. As I add more functions to the GroupBy class, this may set up a DataFrame with edge case-y data

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -196,6 +230,31 @@ internal PrimitiveColumn<decimal> CloneAsDecimalColumn()
...
+                for (long i = 0; i < Length; i++)
+                {
+                    bool containsKey = multimap.TryGetValue(this[i] ?? default, out ICollection<long> values);
+                    if (containsKey)
+                    {
+                        values.Add(i);
+                    }
+                    else
+                    {
+                        multimap.Add(this[i] ?? default, new List<long>() { i });
```
pgovind: This would mean `Null` values in primitive columns will be equivalent to `0`. I *think* this is ok

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -59,16 +59,25 @@ protected set
             get => GetValue(startIndex, length);
         }
 
+        /// <summary>
+        /// Called internally from Merge and GroupBy
```
pgovind: Merge does not exist yet, but it will soon

src/Microsoft.Data/DataFrame.cs
```diff
@@ -260,6 +261,77 @@ public DataFrame Join(DataFrame other, string leftSuffix = "_left", string right
             return ret;
         }
 
+        public GroupBy GroupBy(string columnName)
+        {
+            int columnIndex = _table.GetColumnIndex(columnName);
+            if (columnIndex == -1)
+                throw new ArgumentException($"{columnName} does not exist");
```
pgovind: use Strings here

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
eerhardt: Let's not introduce dead code into our tests. I don't see this method called anywhere.

Also having a `bool forGroupByTests` doesn't feel very sustainable. When the next test that comes along and needs some special test data, adding a new bool to these methods isn't going to work well.
Instead, it may be better to have common code in small reusable methods that are called from higher-level methods.

---
In reply to: [301800494](https://github.com/dotnet/corefxlab/pull/2696#discussion_r301800494) [](ancestors = 301800494)

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,432 @@
...
+                    count++;
+                }
+            });
+
+            EnumerateColumnsWithRows(groupByColumnDelegate, columnDelegate);
+            ret.SetTableRowCount(firstColumn.Length);
+            return ret;
+        }
+
+        private delegate object columnFunc(IEnumerable<long> rows);
```
eerhardt: Is this used?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +784,44 @@ public void TestJoin()
             Assert.Equal(join["Int_right"][2], right["Int"][2]);
             VerifyJoin(join, left, right, JoinAlgorithm.Inner);
         }
+
+        [Fact]
+        public void TestGroupBy()
+        {
+            DataFrame df = MakeDataFrameWithNumericAndBoolColumns(10);
+            DataFrame count = df.GroupBy("Bool").Count();
```
eerhardt: Should we test grouping by something other than `Bool`?

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -99,6 +99,11 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>BaseColumn.Computations.tt</DependentUpon>
     </Compile>
+    <Compile Update="ColumnArithmeticTemplate.cs">
```
eerhardt: I don't think this file exists.  Typically `.ttinclude` files don't generate their own `.cs` file. (At least in my experience.)
This change can be reverted.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -260,6 +261,77 @@ public DataFrame Join(DataFrame other, string leftSuffix = "_left", string right
...
+
+            BaseColumn column = _table.Column(columnIndex);
+
+            switch (column)
```
eerhardt: How about instead of a massive switch like this (and not really being able to support new, extension column types), why don't we have a public abstract method on `BaseColumn` that gets called here?

src/Microsoft.Data/BaseColumn.cs
eerhardt: That isn't a very good public description of what the method does. Can we give it a better summary?

---
In reply to: [301802932](https://github.com/dotnet/corefxlab/pull/2696#discussion_r301802932) [](ancestors = 301802932)

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -59,16 +59,25 @@ protected set
...
+        public virtual void Resize(long length) => throw new NotImplementedException();
+
         /// <summary>
         /// Clone column to produce a copy potentially changing the order by supplying mapIndices and an invert flag
         /// </summary>
         /// <param name="mapIndices"></param>
         /// <param name="invertMapIndices"></param>
         /// <returns></returns>
         public virtual BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
+        public virtual BaseColumn Clone(IEnumerable<long> mapIndices, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
```
eerhardt: Do we need 2 `virtual` Clone methods? Is one a more general version of the other? It would be nice if we could figure out a way to only have a single `virtual Clone` method that takes the appropriate parameters.

src/Microsoft.Data/BaseColumn.Computations.cs
```diff
@@ -33,41 +33,81 @@ public virtual void CumulativeMax()
             throw new NotImplementedException();
         }
 
+        public virtual void CumulativeMax(IEnumerable<long> rows)
```
eerhardt: Is there a better name for this parameter other than `rows`? These aren't really the rows themselves - these are the rowIndices, right?

eerhardt: Since these are all public methods, we should have XML comments on them.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -260,6 +261,77 @@ public DataFrame Join(DataFrame other, string leftSuffix = "_left", string right
...
+                    Dictionary<object, ICollection<long>> dictionary = column.HashColumnValues<object>();
+                    return new GroupBy<object>(this, columnIndex, dictionary);
+            }
+        }
+
+        // In a GroupBy call, columns get resized. We need to set the RowCount to reflect the true Length of the DataFrame. Internal only. Should not be exposed
+        internal void SetTableRowCount(long rowCount)
+        {
+            // Even if current RowCount == rowCount, do the validation
+            int numberOfColumns = ColumnCount;
```
eerhardt: (nit) Why a new variable here? Why not just use `ColumnCount` in the loop?

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,432 @@
...
+            });
+
+            EnumerateColumnsWithRows(groupByColumnDelegate, columnDelegate);
+            ret.SetTableRowCount(firstColumn.Length);
+
+            return ret;
+        }
+
+
+        public override DataFrame First()
```
eerhardt: (nit) should just be single new lines between methods.

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,432 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// A GroupBy class that is typically the product of a DataFrame.GroupBy call
```
eerhardt: (minor) instead of saying `the product of a DataFrame.GroupBy call`, I would go with `the result of a DataFrame.GroupBy call`.

Also, add a period to the end of the sentence.

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,432 @@
...
+    public class GroupBy<TKey> : GroupBy
+    {
+        private int _groupByColumnIndex;
+        private Dictionary<TKey, ICollection<long>> _keyValuePairs;
+        private DataFrame _dataFrame;
+        private PrimitiveColumn<long> _mapIndices;
+
+        public GroupBy(DataFrame dataFrame, int groupByColumnIndex, Dictionary<TKey, ICollection<long>> keyValuePairs)
+        {
+            _groupByColumnIndex = groupByColumnIndex;
```
eerhardt: Since this is a public constructor - we should be verifying the parameters.

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,432 @@
...
+            _keyValuePairs = keyValuePairs;
+            _dataFrame = dataFrame;
+        }
+
+        public override DataFrame All()
+        {
+            throw new NotImplementedException();
```
eerhardt: Should we not add this method until it is actually implemented?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +784,44 @@ public void TestJoin()
...
+            Assert.Equal(2, count.RowCount);
+            Assert.Equal((long)5, count["Int"][0]);
```
eerhardt: Should verify the 2nd row.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +784,44 @@ public void TestJoin()
...
+
+            DataFrame first = df.GroupBy("Bool").First();
+            Assert.Equal(2, first.RowCount);
```
eerhardt: Should verify the values.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +784,44 @@ public void TestJoin()
...
+
+            DataFrame head = df.GroupBy("Bool").Head(3);
+            Assert.Equal(6, head.RowCount);
```
eerhardt: Should verify the values. Repeated below.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +784,44 @@ public void TestJoin()
...
+
+            DataFrame max = df.GroupBy("Bool").Max();
+            Assert.Equal(2, max.RowCount);
+            Assert.Equal(8, max["Int"][0]);
+            Assert.Equal(9, max["Int"][1]);
+
+            DataFrame min = df.GroupBy("Bool").Min();
+            Assert.Equal(2, min.RowCount);
+            Assert.Equal(0, min["Int"][0]);
+            Assert.Equal(0, min["Int"][1]); // Because of the null value
```
eerhardt: I'm not sure I get this comment. Can you elaborate more?

Also, if the behavior is surprising to warrant a comment, is there a bug here that should be addressed to make the behavior understandable?

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,432 @@
...
+        public abstract DataFrame Max();
+        public abstract DataFrame Min();
+        public abstract DataFrame Product();
+        public abstract DataFrame Sum();
+    }
+
+    public class GroupBy<TKey> : GroupBy
+    {
+        private int _groupByColumnIndex;
+        private Dictionary<TKey, ICollection<long>> _keyValuePairs;
```
eerhardt: `_keyValuePairs` doesn't seem like the best name for this variable. Does `_keysToRowIndices` describe it better?

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,432 @@
...
+        }
+
+        public override DataFrame All()
+        {
+            throw new NotImplementedException();
+        }
+
+        internal delegate void ColumnDelegate(int columnIndex, long rowIndex, IEnumerable<long> rows, TKey key, bool firstGroup);
+        internal delegate void GroupByColumnDelegate(long rowNumber, TKey key);
+        internal void EnumerateColumnsWithRows(GroupByColumnDelegate groupByColumnDelegate, ColumnDelegate columnDelegate)
```
eerhardt: Does this and the above delegates need to be `internal`? I think only this class uses them.

pgovind:
Filed https://github.com/dotnet/corefxlab/issues/2700 as a follow up. I don't want to add to this patch

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
pgovind: I filed this https://github.com/dotnet/corefxlab/issues/2700 as a follow up.

Consider a column of ints like [0, 1, 0, 1, null]. At the moment, this forms the groups : [0] and [1] instead of [0], [1] and [null]. `null` is considered equivalent to `default`. This is because the Dictionary used internally for grouping uses the following for it's key:
`dictionary.Add(column[rowIndex] ?? default, ...`

Once I fix https://github.com/dotnet/corefxlab/issues/2700, it should becomes more intuitive.

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: Filed https://github.com/dotnet/corefxlab/issues/2700 to address this. `null` should be it's own group

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -744,5 +772,156 @@ public void TestJoin()
...
+                    BaseColumn productColumn = product[originalColumn.Name];
+                    Assert.Equal("0", productColumn[r].ToString());
+
+                    BaseColumn sumColumn = sum[originalColumn.Name];
+                    Assert.Equal("20", sumColumn[r].ToString());
+                }
+            }
+        }
+
+        [Fact]
```
pgovind: The template types are the ones changing here. Is there a way to use Theory and InlineData for this?

src/Microsoft.Data/DataFrame.cs
eerhardt: >  I was just trying to avoid boxing 

What would be boxed? `BaseColumn` and `GropuBy` are classes and thus reference types. Calling a an abstract method on it won't box anything.

>  the default case is doing that right?

No, the default case is 

```C#
default:
                    Dictionary<object, ICollection<long>> dictionary = column.HashColumnValues<object>();
                    return new GroupBy<object>(this, columnIndex, dictionary);
```

Which means the dictionary is always a `Dictionary<object, ...>`.  If you made a method on BaseColumn that looked like 

```C#
public abstract GroupBy GroupBy(int columnIndex);
```

Then each column can decide how to create the `GroupBy` object and its underlying Dictionary.

src/Microsoft.Data/BaseColumn.Computations.cs
```diff
@@ -12,67 +12,165 @@ namespace Microsoft.Data
...
+        /// </summary>
         public virtual object Max()
         {
             throw new NotImplementedException();
         }
 
+        /// <summary>
+        /// Returns the maximum of the values at rowIndices
+        /// </summary>
+        public virtual object Max(IEnumerable<long> rows)
```
eerhardt: The parameter here is still called `rows`, which is inconsistent with the summary comment and the methods above.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
eerhardt: One option would be to use `[Theory]` and `[MemberData]`. Something like:

```

public static IEnumerable<object[]> GroupByDataFrames()
{
    MakeDataFrame<byte, bool>(10, false), 
    MakeDataFrame<char, bool>(10, false),
    ....
}

[Theory]
[MemberData(nameof(GroupByDataFrames))]
public void TestGoupByDifferentColumnTypes(DataFrame df)
{
    DataFrame grouped = df.GroupBy("Column1").Count();
    Assert.Equal(2, grouped.RowCount);
}

```

I'm not sure if it is significantly better than what you have here, but that would be a way to do it with a Theory.

---
In reply to: [304140369](https://github.com/dotnet/corefxlab/pull/2696#discussion_r304140369) [](ancestors = 304140369)

src/Microsoft.Data/StringColumn.cs
```diff
@@ -255,5 +256,30 @@ private StringColumn Clone(PrimitiveColumn<long> mapIndices = null, bool invertM
             }
             return ret;
         }
+
+        public override Dictionary<TKey, ICollection<long>> HashColumnValues<TKey>()
```
eerhardt: `HashColumnValues` => what does "hash" have to do with this method? My understanding of the method is that it builds a map of values => rowIndices.  It doesn't really "hash" the values.

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,421 @@
...
+                firstColumn.Resize(rowIndex + 1);
+                firstColumn[rowIndex] = key;
+            });
+            ColumnDelegate columnDelegate = new ColumnDelegate((int columnIndex, long rowIndex, IEnumerable<long> rowEnumerable, TKey key, bool firstGroup) =>
+            {
+                if (columnIndex == _groupByColumnIndex)
+                    return;
+                BaseColumn column = _dataFrame.Column(columnIndex);
+                long count = 0;
+                IEnumerator<long> rows = rowEnumerable.GetEnumerator();
```
eerhardt: Why not use `foreach` here?

src/Microsoft.Data/GroupBy.cs
```diff
@@ -0,0 +1,421 @@
...
+        /// <summary>
+        /// Return the first value in each group
+        /// </summary>
+        /// <returns></returns>
+        public abstract DataFrame First();
+
+        /// <summary>
+        /// Returns the first numberOfRows rows of each group
+        /// </summary>
+        /// <param name="numberOfRowsInEachGroup"></param>
```
eerhardt: This param name doesn't match the actual parameter. You should be getting warnings for these kinds of mismatches.

src/Microsoft.Data/strings.resx
```diff
@@ -117,16 +117,37 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CannotResizeDown" xml:space="preserve">
+    <value>Cannot resize down</value>
+  </data>
   <data name="ColumnIndexOutOfRange" xml:space="preserve">
     <value>Index cannot be greater than the Column's Length</value>
   </data>
+  <data name="EmptyFile" xml:space="preserve">
```
eerhardt: This isn't being used. Can you ensure all the resource strings are being used?

src/Microsoft.Data/DataFrame.cs
pgovind: > What would be boxed? BaseColumn and GropuBy are classes and thus reference types. Calling a an abstract method on it won't box anything.

Right right, but that's not what I meant. So, the way it is set up at the moment, let's say I have a `PrimitiveColumn<int> intColumn`. The `GroupBy` method will call `intColumn.HashColumnValues<int>` to produce a `Dictionary<int...`. That's ok, no boxing involved. If instead, I had only the `default` case of `column.HashColumnValues<object>`(where `column` is really `intColumn` and 'object' is really `int`), that would involve boxing when going in and out of the dictionary right? Anyway, I set it up this way because I didn't want the `BaseColumn` objects to have a reference to their parent `DataFrame` object. With your API suggestion, maybe I can still do that. Let me see

src/Microsoft.Data/StringColumn.cs
pgovind: Oh, I guess I associate dictionary with hashing. I changed it to "GroupColumnValues" now. Open to other names too.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -31,6 +32,10 @@ public DataFrame()
         public DataFrame(IList<BaseColumn> columns)
         {
             _table = new DataFrameTable(columns);
+            foreach (BaseColumn column in columns)
+            {
+                column.DataFrame = this;
```
pgovind: Is this really safe? `this` is being constructed here, so I'm not sure.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -69,6 +84,10 @@ protected set
 
         public virtual BaseColumn Sort(bool ascending = true) => throw new NotImplementedException();
 
+        public virtual Dictionary<TKey, ICollection<long>> GroupColumnValues<TKey>() => throw new NotImplementedException();
```
eerhardt: You probably don't need this public method anymore, now that `GroupBy` is a public method on column.

src/Microsoft.Data/DataFrame.cs
eerhardt: Maybe instead of having `BaseColumn` hold onto the `DataFrame` for the lifetime, maybe passing in the `DataFrame` to `GroupBy()` would work better?

That way you could have a `Column` object without a DataFrame.

---
In reply to: [305510252](https://github.com/dotnet/corefxlab/pull/2696#discussion_r305510252) [](ancestors = 305510252)

src/Microsoft.Data/BaseColumn.cs
pgovind: Right, I thought of that and decided to keep it around for the `Merge` PR. I'm ok deleting the API on BaseColumn for now and revisiting later

pgovind:
Thanks for the review and stamp @eerhardt

PR #2698, opened 2019-07-10
Mutable and Immutable buffers

pgovind:
This should go in after https://github.com/dotnet/corefxlab/pull/2696

81922733597f017531bca4c8a475c34f9025c646 is the only commit to review. It also might be an easier review after GroupBy goes in.

### DIFF
--- a/src/Microsoft.Data/DataFrameBuffer.cs
+++ b/src/Microsoft.Data/DataFrameBuffer.cs
@@ -11,22 +11,24 @@ using System.Text;
 namespace Microsoft.Data
 {
     /// <summary>
-    /// A basic store to hold values in a DataFrame column. Supports wrapping with an ArrowBuffer
+    /// A basic immutable store to hold values in a DataFrame column. Supports wrapping with an ArrowBuffer
     /// </summary>
     /// <typeparam name="T"></typeparam>
     public class DataFrameBuffer<T>
         where T : struct
     {
         // TODO: Change this to Memory<T>
-        public Memory<byte> Memory { get; private set; }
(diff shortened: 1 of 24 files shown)


src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -11,22 +11,24 @@
...
     /// </summary>
     /// <typeparam name="T"></typeparam>
     public class DataFrameBuffer<T>
         where T : struct
     {
         // TODO: Change this to Memory<T>
-        public Memory<byte> Memory { get; private set; }
 
-        private readonly int _size;
+        protected Memory<byte> _memory;
```
pgovind: This is changing to a ReadOnlyMemory<byte> in the next patch(Arrow integration)

src/Microsoft.Data/MutableDataFrameBuffer.cs
```diff
@@ -0,0 +1,92 @@
...
+using System.Runtime.InteropServices;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// A basic mutable store to hold values in a DataFrame column. Supports wrapping with an ArrowBuffer
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class MutableDataFrameBuffer<T> : DataFrameBuffer<T>
```
pgovind: I'll make this internal in the next patch perhaps.

eerhardt:
@pgovind - can you update this PR with the latest code?

PR #2699, opened 2019-07-15
Implementation for in-proc proxies with AssemblyLoadContext

cablooshe:
Implementation follow up to #2689. There's currently a bug preventing user-defined types from being put into the constructor, but otherwise everything else works.

/cc @jeffschwMSFT @sbomer @elinor-fung @AaronRobinsonMSFT @jkoritzinsky @vitek-karas @lpereira @swaroop-sridhar @tlakollo


src/ALCProxy.Communication/ALCProxy.Communication.csproj
```diff
@@ -0,0 +1,10 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup> 
+   <Description>Non-allocating collections</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks> 
+   <PackageTags>collections, non-allocating, corefxlab</PackageTags>
```
sbomer: nit: fix indentation. Looks like other projects are using 2 spaces.

PackageTags should only be relevant if we're going to publish this as a package. Seems like the repo packaging logic may pick up all projects: https://github.com/dotnet/corefxlab/blob/7076498fd58574e01bbc6747a788a235bbcc9217/scripts/package.ps1 - we should make sure this doesn't get packaged unless it's intentional.

corefxlab.sln
```diff
@@ -111,6 +111,14 @@ Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.DataFrame",
 EndProject
 Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.DataFrame.Tests", "tests\Microsoft.Data.DataFrame.Tests\Microsoft.Data.DataFrame.Tests.csproj", "{485FC567-4AEC-4335-B767-283173F64C42}"
 EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ALCProxy.Communication", "src\ALCProxy.Communication\ALCProxy.Communication.csproj", "{53180841-8CB6-4A84-8BA7-3D85F0E94879}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ALCProxy.Proxy", "src\ALCProxy.Proxy\ALCProxy.Proxy.csproj", "{1EFD0A3F-C73F-4319-887A-D0CF4F30E14D}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "ALCProxy.Tests", "tests\ALCProxy.Tests\ALCProxy.Tests.csproj", "{D76469F4-3694-463A-8444-187080ECD752}"
+EndProject
+Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "System.Text.CaseFolding", "src\System.Text.CaseFolding\System.Text.CaseFolding.csproj", "{5AD1F0B7-6279-45CE-BE5A-D35CCD198BAC}"
```
sbomer: Why is this project being added?

tests/Benchmarks/ALCProxy/ALCBenchmark.cs
```diff
@@ -0,0 +1,185 @@
...
+using System.Reflection;
+using System.Runtime.Loader;
+using ALCProxy.Proxy;
+using BenchmarkDotNet.Attributes;
+
+namespace Benchmarks.ALCProxy
+{
+    public interface ITest
+    {
+        string PrintContext();
```
sbomer: Maybe call this "GetContextName" or similar?

tests/ALCProxy.Tests/ALCProxyTests.cs
```diff
@@ -0,0 +1,268 @@
...
+
+        public int SimpleMethod()
+        {
+            return tt.testValue;
+        }
+    }
+    public class ALCProxyTests
+    {
+        [Fact]
+        public void TestBasicContextLoading()
```
sbomer: I think it's common practice for xunit to make the test name a statement about what it's asserting. For example https://github.com/dotnet/corefxlab/blob/master/tests/System.Buffers.Primitives.Tests/Tests.cs#L19.

tests/ALCProxy.Tests/ALCProxyTests.cs
```diff
@@ -0,0 +1,268 @@
...
+            this.bb = b;
+            tt = new Test2();
+        }
+        public TestObjectParamConst(int a, string b, Test2 t)
+        {
+            this.aa = a;
+            this.bb = b;
+            tt = t;
+        }
+        public int DoThing2(int a, List<string> list)
```
sbomer: I'd recommend naming these methods after what they're testing. This will make the tests easier to read.

tests/Benchmarks/ALCProxy/ALCBenchmark.cs
```diff
@@ -0,0 +1,185 @@
...
+        public int test;
+        public Test2()
+        {
+            test = 5;
+        }
+        public Test2(int start)
+        {
+            test = start;
+        }
+        public void DoThingy()
```
sbomer: Maybe "MutateMember"

src/ALCProxy.Communication/ICommObjects.cs
```diff
@@ -0,0 +1,257 @@
...
+        }
+        private Type FindTypeInAssembly(string typeName, Assembly a)
+        {
+            //find the type we're looking for
+            Type t = null;
+            foreach (Type ty in a.GetTypes())
+            {
+                if (ty.Name.Equals(typeName) || (ty.Name.StartsWith(typeName) && ty.Name.Contains("`"))) // will happen for non-generic types, generics we need to find the additional "`1" afterwards
+                {
+                    t = ty;
```
AaronRobinsonMSFT: Return immediately,then always throw at the end of the function.

src/ALCProxy.Communication/ICommObjects.cs
```diff
@@ -0,0 +1,257 @@
...
+            _intType = interfaceType;
+            _serverTypeName = serverName;
+        }
+        private Type FindTypeInAssembly(string typeName, Assembly a)
+        {
+            //find the type we're looking for
+            Type t = null;
+            foreach (Type ty in a.GetTypes())
+            {
+                if (ty.Name.Equals(typeName) || (ty.Name.StartsWith(typeName) && ty.Name.Contains("`"))) // will happen for non-generic types, generics we need to find the additional "`1" afterwards
```
AaronRobinsonMSFT: >generics we need to find the additional "`1" afterwards

What about generic types with more than 1 parameter? I would just clarify.

src/ALCProxy.Communication/CommObjects.cs
```diff
@@ -0,0 +1,65 @@
...
+        protected override object SerializeParameter(object param, Type paramType)
+        {
+            MemoryStream stream = new MemoryStream();
+            //Serialize the Record object to a memory stream using DataContractSerializer.  
+            DataContractSerializer serializer = new DataContractSerializer(paramType);
+            serializer.WriteObject(stream, param);
+            return stream;
+
+        }
+        protected override object DecryptReturnType(object s, Type returnType)
```
sbomer: I'd suggest `Deserialize` instead of `Decrypt`

src/ALCProxy.Communication/ICommObjects.cs
```diff
@@ -0,0 +1,257 @@
...
+            Assembly aa = alc.LoadFromAssemblyPath(Assembly.GetAssembly(typeof(ClientDispatch)).CodeBase.Substring(8));
+            Type serverType = FindTypeInAssembly(_serverTypeName, aa);
+            //Set up all the generics to allow for the serverDispatch to be created correctly
+            Type constructedType = serverType.MakeGenericType(_intType);
+            //Give the client its reference to the server
+            SerializeParameters(constructorParams, out IList<object> serializedConstArgs, out IList<Type> argTypes);
+            _server = constructedType.GetConstructor(
+                new Type[] { typeof(Type), typeof(Type[]), typeof(IList<object>), typeof(IList<Type>) })
+                .Invoke(new object[] { objType, genericTypes, serializedConstArgs.ToList(), argTypes });
+            _callMethod = _server.GetType().GetMethod("CallObject");
```
sbomer: I think we should be consistent about how we specify the server name and method - we shouldn't hard-code "CallObject" when the server type is passed into the ctor.

src/ALCProxy.Communication/CommObjects.cs
```diff
@@ -0,0 +1,65 @@
...
+        {
+            MemoryStream stream = (MemoryStream)s;
+            stream.Position = 0;
+            //Deserialize the Record object back into a new record object.  
+            DataContractSerializer newSerializer = new DataContractSerializer(returnType);
+            return newSerializer.ReadObject(stream);
+        }
+    }
+    public class ServerDispatch<InterfaceType> : ALCServer<InterfaceType>
```
sbomer: Try to be consistent in `I`/`InterfaceType` between different classes - I think generally we prefer shorter names for type parameters when it's relatively clear from context what they are.

src/ALCProxy.Communication/ICommObjects.cs
```diff
@@ -0,0 +1,257 @@
...
+        public ALCClient(Type interfaceType, string serverName)
+        {
+            _intType = interfaceType;
+            _serverTypeName = serverName;
+        }
+        private Type FindTypeInAssembly(string typeName, Assembly a)
+        {
+            //find the type we're looking for
+            Type t = null;
+            foreach (Type ty in a.GetTypes())
```
sbomer: Wondering if we can use `Assembly.GetType` instead of iterating through all types and matching ourselves.

src/ALCProxy.Communication/ICommObjects.cs
```diff
@@ -0,0 +1,257 @@
...
+        /// <summary>
+        /// Creates the link between the client and the server, while also passing in all the information to the server for setup
+        /// </summary>
+        /// <param name="alc">The target AssemblyLoadContext</param>
+        /// <param name="typeName">Name of the proxied type</param>
+        /// <param name="assemblyPath">path of the assembly to the type</param>
+        /// <param name="genericTypes">any generics that we need the proxy to work with</param>
+        public void SetUpServer(AssemblyLoadContext alc, string typeName, string assemblyPath, object[] constructorParams, Type[] genericTypes)
+        {
+            Assembly a = alc.LoadFromAssemblyPath(assemblyPath);
```
jeffschwMSFT: Ditto on the parameter validation, particularly for public methods.

src/ALCProxy.Communication/ICommObjects.cs
```diff
@@ -0,0 +1,257 @@
...
+        }
+        /// <summary>
+        /// Create the instance of the object we want to proxy
+        /// </summary>
+        /// <param name="instanceType">the type of the object we want</param>
+        /// <param name="constructorTypes">The list of types that the constructor of the object takes in as an argument</param>
+        /// <param name="constructorArgs">The physical objects that are the parameters to the constructor</param>
+        protected void SetInstance(Type instanceType, Type[] constructorTypes, object[] constructorArgs)
+        {
+            ConstructorInfo ci = instanceType.GetConstructor(constructorTypes);//GrabConstructor(constructorTypes, instanceType);//instanceType.GetConstructor(constructorTypes);
```
jeffschwMSFT: nit: feel free to use var when the type is obvious from the right hand expression

tests/ALCProxy.Tests/ALCProxyTests.cs
```diff
@@ -0,0 +1,268 @@
...
+            return tt.testValue;
+        }
+    }
+    public class ALCProxyTests
+    {
+        [Fact]
+        public void TestBasicContextLoading()
+        {
+            AssemblyLoadContext alc = new AssemblyLoadContext("TestBasicContextLoading", isCollectible: true);
+            ITest t = ProxyBuilder<ITest>.CreateInstanceAndUnwrap(alc, Assembly.GetExecutingAssembly().CodeBase.Substring(8), "Test");
```
jeffschwMSFT: Do you have any tests where the interface is defined in a separate assembly?  I feel that is the main customer scenario, where the plugin designer defines an exchange interface.  The interface would likely then be loaded in the both the child and parent ALC.  The hope is that the plug-in or the host do not have to do anything special to be able to communicate through the interface.

src/ALCProxy.Proxy/ProxyBuilder.cs
```diff
@@ -0,0 +1,61 @@
...
+            T obj = ALCDispatch<T>.CreateGeneric(alc, assemblyPath, typeName, constructorParams, genericTypes);
+            return obj;
+        }
+        public static T CreateGenericInstanceAndUnwrap(AssemblyLoadContext alc, string assemblyPath, string typeName, Type[] genericTypes)
+        {
+            return CreateGenericInstanceAndUnwrap(alc, assemblyPath, typeName, new object[] { }, genericTypes);
+        }
+
+    }
+    public class ALCDispatch<I> : System.Reflection.DispatchProxy //T is the TargetObject type, I is the specific client you want to use.
```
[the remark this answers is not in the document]
cablooshe: That was leftover from an old version. It's been removed.

src/ALCProxy.Communication/CommObjects.cs
cablooshe: Fixed.

src/ALCProxy.Communication/ICommObjects.cs
```diff
@@ -0,0 +1,257 @@
...
+// The .NET Foundation licenses this file to you under the MIT license. 
+// See the LICENSE file in the project root for more information. 
+
+using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Reflection;
+using System.Runtime.Loader;
+
+namespace ALCProxy.Communication
```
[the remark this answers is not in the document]
cablooshe: I wasn't exactly what to name the file at the time, since it holds both the code for servers and clients and they both fall under the section of the proxies that handles communication between ALCs. Would it make more sense to split up those two into separate files and rename the files to match with each?

src/ALCProxy.Communication/ICommObjects.cs
cablooshe: `Assembly.GetType` works, but only if we put the full name of the assembly (ex. entering "ALCProxy.Communication.ServerDispatch\`1" vs entering "ServerDispatch\`1") in the path, which the user may not do. I'll add ways for users to use both options and change this to use `Assembly.GetType` no matter which version of the type name you've used.

src/ALCProxy.Communication/ICommObjects.cs
cablooshe: I've removed this part when getting the method to work with `Assembly.GetType`

src/ALCProxy.Communication/IProxyClient.cs
```diff
@@ -0,0 +1,127 @@
...
+        /// <param name="typeName">Name of the proxied type</param>
+        /// <param name="assemblyPath">path of the assembly to the type</param>
+        /// <param name="genericTypes">any generics that we need the proxy to work with</param>
+        public void SetUpServer(AssemblyLoadContext alc, string typeName, string assemblyPath, object[] constructorParams, Type[] genericTypes)
+        {
+            Assembly a = alc.LoadFromAssemblyPath(assemblyPath);
+            //find the type we're going to proxy inside the loaded assembly
+            Type objType = FindTypeInAssembly(typeName, a);
+            //Load *this* (ALCProxy.Communication) assembly into the ALC so we can get the server into the ALC
+            Assembly aa = alc.LoadFromAssemblyPath(Assembly.GetAssembly(typeof(ServerDispatch<>)).CodeBase.Substring(8));
```
vitek-karas: I don't understand the need for this - this will cause the assembly to be loaded twice which incurs non-trivial cost to the runtime. Among other things we can't reuse Ready2Run images, so it will be JITed again.
Since we own the code in this assembly we can make it ALC aware and thus it should not matter where it comes from - meaning it should he shared.
Going forward, if this would ever become part of the framework it would be shared almost by default anyway.

src/ALCProxy.Communication/IProxyServer.cs
```diff
@@ -0,0 +1,157 @@
...
+        public object CallObject(MethodInfo targetMethod, IList<object> streams, IList<Type> argTypes)
+        {
+            //Turn the memstreams into their respective objects
+            argTypes = argTypes.Select(x => ConvertType(x)).ToList();
+            object[] args = DeserializeParameters(streams, argTypes);
+            MethodInfo[] methods = instance.GetType().GetMethods();
+            MethodInfo m = FindMethod(methods, targetMethod, argTypes.ToArray());
+            if (m.ContainsGenericParameters)
+            {
+                //While this may work without the conversion, we want it to uphold the type-load boundary, don't let the passed in method use anything from outside the target ALC
```
vitek-karas: I don't understand this requirement. I think it would be good to split the responsibilities here.
* The communication classes should not make a decision which assembly gets loaded where
* The ALC itself should make the final decisions where to load assemblies from
* The communication classes just need to make sure to load any types "by name" through the target ALC (so that it can make the right decisions).

src/ALCProxy.Communication/IProxyServer.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// </summary>
+        protected Type ConvertType(Type toConvert)
+        {
+            string assemblyPath = Assembly.GetAssembly(toConvert).CodeBase.Substring(8);
+            if (toConvert.IsPrimitive || assemblyPath.Contains("System.Private.CoreLib")) //Can't load/dont want to load extra types from System.Private.CoreLib
+            {
+                return toConvert;
+            }
+            //TODO can we look for the existence of the loaded assembly here, instead of just doing the load?
+            return currentLoadContext.LoadFromAssemblyPath(assemblyPath).GetType(toConvert.FullName);
```
vitek-karas: The runtime will not load anything twice into the same ALC - it would return the already loaded assembly anyway.
But in general I don't think this code should be here - see my other comments.

src/ALCProxy.Communication/IProxyServer.cs
```diff
@@ -0,0 +1,157 @@
...
+        /// Sends a message to the server to proc the method call, and return the result
+        /// </summary>
+        /// <param name="method">the method that needs to be called</param>
+        /// <param name="streams">The parameters for the given method, converted into a serialized object by the client that now need to be deserialized</param>
+        /// <param name="types">The types of each stream, so the server knows how to decode the streams</param>
+        /// <returns></returns>
+        object CallObject(MethodInfo method, IList<object> streams, IList<Type> types);
+    }
+
+    public abstract class ALCServer<I> : IProxyServer
```
vitek-karas: I could not find where the `I` type parameter is used... I guess I missed something

src/ALCProxy.Communication/IProxyServer.cs
```diff
@@ -0,0 +1,157 @@
...
+            }
+            throw new MissingMethodException("Error in ALCProxy: Method Not found for " + instance.ToString() + ": " + methodName);
+        }
+        /// <summary>
+        /// If a parameter of a function isn't the direct type that we've passed in, this function should find that the type we've passed is correct.
+        /// </summary>
+        private bool RecursivelyCheckForTypes(Type sentParameterType, Type toCompare)
+        {
+            Type[] interfaces = sentParameterType.GetInterfaces();
+            if (sentParameterType.Equals(toCompare))
```
vitek-karas: This seems to be using `Type.Equals` which will likely return false if the assembly is loaded twice - or they are slightly different assemblies. Basically we would need a fully string based comparisons throughout...

src/ALCProxy.Communication/IProxyServer.cs
cablooshe: You're right, it seems like a holdover from an older version of the server implementation I was trying. I've removed the use of `I` for the server.

src/ALCProxy.Communication/ALCClient.cs
```diff
@@ -57,13 +59,14 @@ public void SetUpServer(AssemblyLoadContext alc, string typeName, string assembl
             _server = constructedType.GetConstructor(
                 new Type[] { typeof(Type), typeof(Type[]), typeof(IList<object>), typeof(IList<Type>) })
                 .Invoke(new object[] { objType, genericTypes, serializedConstArgs.ToList(), argTypes });
-            _callMethod = _server.GetType().GetMethod("CallObject");
+            _serverDelegate = Delegate.CreateDelegate(typeof(ServerCall), _server, constructedType.GetMethod("CallObject"));
```
jkoritzinsky: You should make the `_serverDelegate` field be of type `ServerCall` and just call the delegate below instead of using `DynamicInvoke`. That way you'll reap all of the performance gains of using delegates instead of just `MethodInfo.Invoke`.

src/ALCProxy.Communication/ALCClient.cs
```diff
@@ -18,19 +19,24 @@ public abstract class ALCClient : IProxyClient
         protected object _server;
         protected string _serverTypeName;
         protected Type _intType;
-        protected Delegate _serverDelegate;
+        internal ServerCall _serverDelegate;
+
+        public StackTrace _stackTrace;
```
AaronRobinsonMSFT: make this `private` - the debugger can see all :)

src/ALCProxy.Communication/IProxyClient.cs
cablooshe: For now, I was building most of the implementation in such a way that we can make sure that no shared assemblies are required to get plugins working. That said, if we wanted to enforce that plugins need some reference to the main program, and thus this API, we could probably run with shared assemblies and prevent the extra load cost. Is that what you're thinking we should be doing with the API here to prevent the double-load?

src/ALCProxy.Communication/IProxyServer.cs
cablooshe: Since both parameters to this method should've gone through the `ConvertType` method, `Type.Equals` should be working as long as the ALC can't load two of the same type (like you said previously above). That said if we change how that load works, then yeah, we'll definitely need to change this.

src/ALCProxy.Communication/IProxyServer.cs
cablooshe: I'm a little confused on what you're trying to say here. Is this like the examples in the docs of using AssemblyLoadResolver to get the assemblies instead? The current implementation relies on making sure that any types that are required by the proxied object are loaded in from the proxy API, from the same assemblies that we're using for the default ALC where the proxy is created. I'm not sure how we can make sure everything will work without doing any direct loads from this API, but I may just be misunderstanding what you're suggesting...

src/ALCProxy.Communication/IProxyClient.cs
vitek-karas: As noted later on, I think the proxies should not make any assumptions about the ALC and how it loads assemblies. They should always load "through" that ALC, but they should not assume in which ALC the assembly will actually get loaded.
There are benefits to sharing and ALCs should be free to make that choice.

Note that this does not limit unloadability - the proxies should have the clear isolation regardless of where the assemblies ends up loaded. And so if they do get loaded into a separate ALC, then they will get unloaded. If they don't, then we only added potentially unnecessary cost to calling to it, but nothing will break due to it.

I get why you made it this way - it is relatively easy to see that we isolate fully I guess. That said I would still make the proxy classes itself an exception to this rule. But I guess it doesn't matter much.

src/ALCProxy.Communication/IProxyServer.cs
vitek-karas: I see - well ConvertType should not rely on where the assemblies are loaded - it just needs to load them through the right ALC. Then it will keep on working regardless of the actual assembly load layout.

src/ALCProxy.Communication/IProxyServer.cs
vitek-karas: We need to load *through* the right ALC, but it doesn't matter where the actual assembly ends up being loaded to.
So in this case instead of calling LoadAssemblyByPath, simply call Load with a simple name, It's up to the ALC to decide how to load the assembly. That way you get to avoid the problem of not loading CoreLib twice (as ALCs will typically handle that correctly). It will also typically make sure that all framework types are shared, so for example if you would have an XElement as a parameter, right now you would try to load System.Private.Xml twice... which is generally not a good idea (it would probably work though).

The only visible change would be that the initial proxy creation would only specify assembly name (not a full path) and it would be up to the ALC to find the file. So you would probably need to sublass the custom ALC and implement some logic in Load - which most custom ALCs will do anyway.

Note that this will give you the easy "differences" - as each ALC may decide to load from different place. So for example you IPlugin may come from different places and thus have different version. But that is the desired behavior in the end.

src/ALCProxy.Communication/IProxyClient.cs
cablooshe: That makes a lot of sense to consider, and yeah, performance would be a lot better using shared assemblies there. I'll add a comment on how we want to do this to the design doc, since the issue from the current implementation does seem pretty important.

vitek-karas: Sounds great!

src/ALCProxy.Communication/IProxyServer.cs
cablooshe: I've got the proxy system working now when you pass in an `AssemblyName`. Do you think it may be work to keep the file path option working to allow for people to use non-subclassed `AssemblyLoadContext`s for proxies, or will that just lead to more confusion?

tests/ALCProxy.Tests/ALCProxyTests.cs
```diff
@@ -0,0 +1,353 @@
...
+            Assert.Equal(test.testValue, t2.SimpleMethodReturnsInt());
+        }
+        //As a note, to test these, we need to manually build any of the external classes before running tests, 
+        //or you will error out with a TypeLoadException when the creation of the proxy is attempted
+        [Fact]
+        public void CanLoadOustideAssemblyWithSharedInterface()
+        {
+            string pathOfCurrentString = "ALCProxy.Tests";
+            string assemblyLocation = Assembly.GetExecutingAssembly().Location;
+            string newPath = assemblyLocation.Substring(0, assemblyLocation.IndexOf(pathOfCurrentString)) + "ALCProxy.TestAssembly\\bin\\" + dbgString + "\\netcoreapp3.0\\ALCProxy.TestAssembly.dll";
```
sbomer: Use `Path.Combine` or `Path.DirectorySeparatorChar`

sbomer: Use `Path.Combine` or `Path.DirectorySeparatorChar`, and `Path.GetDirectoryName`

src/ALCProxy.Communication/IProxyServer.cs
vitek-karas: I would remove it - beside the potential confusion, I'm no aware of any other place where we explicitly load files into an ALC in a similar manner - so it's not a common patter either. I prefer simplicity in this case. If there's a need for something else we can extend it later on.

src/ALCProxy.Communication/ALCClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            }
+            return t;
+        }
+        /// <summary>
+        /// Creates the link between the client and the server, while also passing in all the information to the server for setup
+        /// </summary>
+        /// <param name="alc">The target AssemblyLoadContext</param>
+        /// <param name="typeName">Name of the proxied type</param>
+        /// <param name="assemblyName">path of the assembly to the type</param>
+        /// <param name="genericTypes">any generics that we need the proxy to work with</param>
```
ahsonkhan: If the plan is to get public APIs documented with xml comments, make sure to add exception tags as well to document what exceptions are thrown and when.

src/ALCProxy.Communication/ALCClient.cs
```diff
@@ -0,0 +1,174 @@
...
+    {
+        //Can't make this an IServerObject directly due to the type-loading barrier
+        protected object _server;
+        protected string _serverTypeName;
+        protected Type _intType;
+        internal ServerCall _serverDelegate;
+        protected Type _serverType;
+
+        private StackTrace _stackTrace;
+        public ALCClient(Type interfaceType, string serverName, Type serverType)
```
ahsonkhan: public surface area missing docs.

src/ALCProxy.Communication/ALCClient.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Runtime.Loader;
+using System.Reflection.Emit;
+using System.IO;
+
+namespace ALCProxy.Communication
+{
+    internal delegate object ServerCall(MethodInfo info, IList<object> args, IList<Type> types);
+    /// <summary>
+    /// This currently is designed to only work in-process
+    /// TODO: set up to allow for construction of out-of-proc proxies
```
ahsonkhan: Move such TODO out of the xml comments (regular comment outside the `<summary>` is fine).

src/ALCProxy.Communication/ALCClient.cs
```diff
@@ -0,0 +1,174 @@
...
+            //Give the client its reference to the server
+            SerializeParameters(constructorParams, out IList<object> serializedConstArgs, out IList<Type> argTypes);
+            ConstructorInfo ci = serverType.GetConstructor(
+                new Type[] { typeof(Type), typeof(Type[]), typeof(IList<object>), typeof(IList<Type>) });
+            _server = ci.Invoke(new object[] { objType, genericTypes, serializedConstArgs.ToList(), argTypes });
+            _serverDelegate = (ServerCall)Delegate.CreateDelegate(typeof(ServerCall), _server, serverType.GetMethod("CallObject"));
+            //Attach to the unloading event
+            alc.Unloading += UnloadClient;
+        }
+        /// <summary>
```
ahsonkhan: nit: add line break between methods.

src/ALCProxy.Communication/ALCProxy.Communication.csproj
```diff
@@ -0,0 +1,9 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <Description>Non-allocating collections</Description>
```
ahsonkhan: Description doesn't match.

src/ALCProxy.Communication/ALCServer.cs
```diff
@@ -0,0 +1,148 @@
...
+                for (int i = 0; i < parameterTypes.Length; i++)
+                {
+                    if (!RecursivelyCheckForTypes(parameterTypes[i], m.GetParameters()[i].ParameterType))
+                    {
+                        methodParamsAlligned = false;
+                        break;
+                    }
+                }
+                if (!methodParamsAlligned)
+                    continue;
```
ahsonkhan: Add braces whenever in a nested context (even for single line branches). Generally, no braces at the root scope or at the beginning of the method for arg validation is fine.

src/ALCProxy.Proxy/ALCProxy.Proxy.csproj
```diff
@@ -0,0 +1,16 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+
+  <PropertyGroup>
+    <Description>Non-allocating collections</Description>
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
+  </PropertyGroup>
+
+
```
ahsonkhan: nit: too much white space

tests/ALCProxy.TestAssembly/ALCProxy.TestAssembly.csproj
```diff
@@ -0,0 +1,14 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
```
ahsonkhan: Import common.props?

https://github.com/dotnet/corefxlab/blob/58b32dabce427f7ef37a7ff9bdbb3c64321229d5/tests/System.Buffers.Experimental.Tests/System.Buffers.Experimental.Tests.csproj#L2

tests/ALCProxy.TestInterface/IExternalClass.cs
```diff
@@ -0,0 +1,16 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections.Generic;
+
+namespace ALCProxy.TestInterface
```
ahsonkhan: Maybe this is acceptable, why do we have small test projects with single types in them? TestInterface, TestInterfaceUpdated? Could they be in one instead?

tests/ALCProxy.Tests/ALCProxy.Tests.csproj
```diff
@@ -0,0 +1,28 @@
...
+
+    <GeneratePackageOnBuild>false</GeneratePackageOnBuild>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="Microsoft.Diagnostics.Tracing.TraceEvent" Version="2.0.30" />
+    <PackageReference Include="Microsoft.NET.Test.Sdk" Version="15.9.0" />
+    <PackageReference Include="xunit" Version="2.4.0" />
+    <PackageReference Include="xunit.performance.api" Version="1.0.0-beta-build0020" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0">
```
ahsonkhan: Generally, we import common.prop/dependencies.props and use the version specified there: https://github.com/dotnet/corefxlab/blob/58b32dabce427f7ef37a7ff9bdbb3c64321229d5/tools/dependencies.props#L2-L19

tests/ALCProxy.TestInterface/IExternalClass.cs
cablooshe: The test here is to check if a plugin that was built off an older interface (`TestInterface`) can be proxied with a newer version of the interface and treated like one (`TestInterfaceUpdated`). Keeping them in separate assemblies is to simulate the main ALC where the proxy is created not having any access to the older interface, but I could probably merge them into one assembly to organize it a bit better.

cablooshe: Actually whoops, my mistake. Since they're technically the same type I'm not sure we could merge them into one assembly.

tests/ALCProxy.Tests/ALCProxy.Tests.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <LangVersion>8.0</LangVersion>
+  </PropertyGroup>
+  <ItemGroup>
+    <PackageReference Include="xunit" Version="2.4.0" />
+    <PackageReference Include="xunit.runner.visualstudio" Version="2.4.0" >
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+  </ItemGroup>  <ItemGroup>
```
ahsonkhan: spacing is off

Issue #2700, opened 2019-07-16
Improve support for Null values in DataFrame GroupBy


pgovind:
At the moment, `null` values are considered equivalent to `default`. Instead, `null` should be considered its own group

pgovind:
https://github.com/dotnet/corefxlab/pull/2701 implements an immutable arrow string compatible column. The GroupBy method needs to be fixed there too

pgovind:
Fixing this will also allow for optimizations in GroupBy.cs. At the moment, there are checks in some of the functions to account for the `null`.

PR #2701, opened 2019-07-24
Immutable DataFrameBuffers and ArrowStringColumn

pgovind:

This patch implements mutable and immutable DataFrameBuffers to allow for true zero copy Arrow integration. It also implements a new immutable ArrowStringColumn column type compatible with Arrow's StringColumn


src/Microsoft.Data/strings.Designer.cs
```diff
@@ -78,6 +78,33 @@ public class Strings {
...
+        ///   Looks up a localized string similar to Bool operations expect bool columns.
+        /// </summary>
+        public static string ExpectedBoolColumn {
+            get {
+                return ResourceManager.GetString("ExpectedBoolColumn", resourceCulture);
+            }
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to ArrowStringColumn is immutable.
```
eerhardt: Where are the changes to the .resx file for this?

src/Microsoft.Data/StringColumn.cs
```diff
@@ -9,7 +9,7 @@
 namespace Microsoft.Data
 {
     /// <summary>
-    /// A column to hold strings
+    /// A mutable column to hold strings. Is NOT Arrow compatible
```
eerhardt: Something like `Is NOT Arrow compatible` is more appropriate for a `<remarks>` section, not a summary.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -34,6 +35,132 @@ public DataFrame(IList<BaseColumn> columns)
...
+            IEnumerable<IArrowArray> arrowArrays = recordBatch.Arrays;
+            foreach (IArrowArray arrowArray in arrowArrays)
+            {
+                Field field = arrowSchema.GetFieldByIndex(fieldIndex);
+                IArrowType fieldType = field.DataType;
+                BaseColumn dataFrameColumn = null;
+                switch (fieldType.TypeId)
+                {
+                    case ArrowTypeId.Boolean:
+                        ReadOnlyMemory<byte> valueBuffer = (arrowArray as BooleanArray).ValueBuffer.Memory;
```
eerhardt: Instead of a pattern like:

`(someObj as SomeType).SomeProperty`

in general it is better to cast:

`((SomeType)someObj).SomeProperty`

The reasoning is because if the cast doesn't succeed in the first case, you get a NullReferenceException. In the second case, you get a InvalidCastException. It is easier to diagnose the error using the second pattern.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+                for (long i = 0; i < Length; i++)
+                {
+                    bool containsKey = multimap.TryGetValue(this[i] ?? default, out ICollection<long> values);
+                    if (containsKey)
+                    {
+                        values.Add(i);
+                    }
+                    else
+                    {
+                        multimap.Add(this[i] ?? default, new List<long>() { i });
```
eerhardt: Doing `this[i]` is actually a pretty expensive operation, right? It will decode the UTF8 bytes into a UTF16 .NET string. We should probably only do it once for each index.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+using System.Text;
+using Apache.Arrow;
+using Apache.Arrow.Types;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// An immutable to hold Arrow style strings
+    /// </summary>
+    public partial class ArrowStringColumn : BaseColumn
```
eerhardt: I'm not sure I like using the name `ArrowStringColumn`. My first inclination was to call it `UTF8StringColumn` or similar.

However, I'm not sure I like my suggestion. So I'm just raising this thought here to see if you have any ideas on what could be a better name.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+    {
+        private IList<DataFrameBuffer<byte>> _dataBuffers;
+        private IList<DataFrameBuffer<int>> _offsetsBuffers;
+        private IList<DataFrameBuffer<byte>> _nullBitMapBuffers;
+
+        private readonly Encoding _encoding = Encoding.UTF8;
```
eerhardt: Since this is an "Arrow" string, and those are always UTF8, I don't think we need to capture this in a readonly field. We can just use Encoding.UTF8 inline where we need it.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+
+        // Need a way to differentiate between columns initialized with default values and those with null values in SetValidityBit
+        internal bool _modifyNullCountWhileIndexing = true;
```
eerhardt: Is this always `true`? I don't see it being set to false. Also, it shouldn't be exposed outside of the class unless absolutely necessary. And if it is exposed outside of the class, it shouldn't be named with an `_` prefix.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+        public bool IsValid(long index) => NullCount == 0 || GetValidityBit(index);
+
+        public bool GetValidityBit(long index)
+        {
+            if ((ulong)index > (ulong)Length)
+            {
+                throw new ArgumentOutOfRangeException(nameof(index));
+            }
+            // First find the right bitMapBuffer
+            int bitMapIndex = GetBufferIndexContainingRowIndex(ref index);
```
eerhardt: I still don't like this pattern of a `ref long` that I know I can cast to an `int` later. I think the method should pass back 2 ints.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+                curArrayIndex++;
+            }
+            return curArrayIndex;
+        }
+
+        private ReadOnlySpan<byte> GetBytes(long index)
+        {
+            int offsetsBufferIndex = GetBufferIndexContainingRowIndex(ref index);
+            int currentOffset = _offsetsBuffers[offsetsBufferIndex].ReadOnlySpan[(int)index];
+            int nextOffset = _offsetsBuffers[offsetsBufferIndex].ReadOnlySpan[(int)index + 1];
```
eerhardt: Should only call `_offsetsBuffers[offsetsBufferIndex].ReadOnlySpan` once here.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+                    return _encoding.GetString(data, bytes.Length);
+            }
+        }
+
+        protected override object GetValue(long startIndex, int length)
+        {
+            var ret = new List<string>();
+            while (ret.Count < length)
+            {
+                ret.Add((string)GetValue((int)startIndex));
```
eerhardt: Don't you need to increment startIndex here?

BTW - do we have tests for this method?

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+        }
+
+        protected override void SetValue(long rowIndex, object value)
+        {
+            throw new NotSupportedException(Strings.ImmutableStringColumn);
+        }
+
+        public new string this[long rowIndex]
+        {
+            get => (string)GetValue(rowIndex);
```
eerhardt: This is backwards. The weakly typed API (the one returning `object`) should call the strongly-typed API (the one returning a `string`). Then no casts are necessary.

src/Microsoft.Data/ArrowStringColumn.cs
eerhardt: This method could just `return this[startIndex, length];` right?

---
In reply to: [307378768](https://github.com/dotnet/corefxlab/pull/2701#discussion_r307378768) [](ancestors = 307378768)

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+        }
+
+        public new List<string> this[long startIndex, int length]
+        {
+            get
+            {
+                var ret = new List<string>();
+                while (ret.Count < length)
+                {
+                    ret.Add((string)GetValue(startIndex++));
```
eerhardt: Call the strongly-typed method here so you don't need to cast.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+                }
+                return ret;
+            }
+        }
+
+        public override Field Field => new Field(Name, StringType.Default, NullCount != 0);
```
eerhardt: In general, this isn't a good pattern. A property should act more or less like a field, it should be quick to access it. Here you are creating a new object every time the property is accessed. If this is intended, it should be a method.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+            Debug.Assert(startIndex <= Int32.MaxValue);
+            return _offsetsBuffers[offsetsBufferIndex].Length - (int)startIndex;
+        }
+
+        private int GetNullCount(long startIndex, int numberOfRows)
+        {
+            int nullCount = 0;
+            for (long i = startIndex; i < numberOfRows; i++)
+            {
+                if (!IsValid(i))
```
eerhardt: (not necessarily for this PR) - I wonder if we could speed this up by using PopCount: [https://docs.microsoft.com/en-us/dotnet/api/system.numerics.bitoperations.popcount?view=netcore-3.0](https://docs.microsoft.com/en-us/dotnet/api/system.numerics.bitoperations.popcount?view=netcore-3.0)

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+                    nullCount++;
+            }
+            return nullCount;
+        }
+
+
+        public override Apache.Arrow.Array AsArrowArray(long startIndex, int numberOfRows)
+        {
+            int offsetsBufferIndex = GetBufferIndexContainingRowIndex(ref startIndex);
+            ArrowBuffer dataBuffer = numberOfRows == 0 ? ArrowBuffer.Empty : new ArrowBuffer(_dataBuffers[offsetsBufferIndex].ReadOnlyMemory);
```
eerhardt: What happens if `numberOfRows` crosses over 2 buffers?

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+                clone = Clone();
+            }
+            for (long i = 0; i < numberOfNullsToAppend; i++)
+            {
+                clone.Append(null);
+            }
+            return clone;
+        }
+
+        private ArrowStringColumn Clone(PrimitiveColumn<long> mapIndices = null, bool invertMapIndex = false)
```
eerhardt: Do you really need default values on this private method? Isn't it only called in 1 place?

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+            }
+            else
+            {
+                if (mapIndices.Length > Length)
+                    throw new ArgumentException(Strings.MapIndicesExceedsColumnLenth, nameof(mapIndices));
+                if (invertMapIndex == false)
+                {
+                    for (long i = 0; i < mapIndices.Length; i++)
+                    {
+                        ret.Append(this[(long)mapIndices[i]]);
```
eerhardt: Why is the cast to `long` here necessary?

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,348 @@
...
+                    for (long i = 0; i < mapIndices.Length; i++)
+                    {
+                        ret.Append(this[(long)mapIndices[mapIndicesLengthIndex - i]]);
+                    }
+                }
+            }
+            return ret;
+        }
+
+        public Dictionary<TKey, ICollection<long>> HashColumnValues<TKey>()
```
eerhardt: Who calls this method?

eerhardt: Don't you want to override `GroupColumnValues` instead?

---
In reply to: [307387981](https://github.com/dotnet/corefxlab/pull/2701#discussion_r307387981) [](ancestors = 307387981)

src/Microsoft.Data/DataFrame.cs
eerhardt: Should only need to cast the arrowArray one time in the method

src/Microsoft.Data/DataFrame.cs
```diff
@@ -34,6 +35,132 @@ public DataFrame(IList<BaseColumn> columns)
...
+                        ReadOnlyMemory<byte> doubleValueBuffer = (arrowArray as PrimitiveArray<double>).ValueBuffer.Memory;
+                        ReadOnlyMemory<byte> doubleNullBitMapBuffer = (arrowArray as PrimitiveArray<double>).NullBitmapBuffer.Memory;
+                        dataFrameColumn = new PrimitiveColumn<double>(field.Name, doubleValueBuffer, doubleNullBitMapBuffer, arrowArray.Length, arrowArray.NullCount);
+                        break;
+                    case ArrowTypeId.Float:
+                        ReadOnlyMemory<byte> floatValueBuffer = (arrowArray as PrimitiveArray<float>).ValueBuffer.Memory;
+                        ReadOnlyMemory<byte> floatNullBitMapBuffer = (arrowArray as PrimitiveArray<float>).NullBitmapBuffer.Memory;
+                        dataFrameColumn = new PrimitiveColumn<float>(field.Name, floatValueBuffer, floatNullBitMapBuffer, arrowArray.Length, arrowArray.NullCount);
+                        break;
+                    case ArrowTypeId.Int8:
```
eerhardt: Does Int8 and Int16 arrays really cast to `PrimitiveArray<int>`? I don't think this will work.

Do we have a test for it?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -34,6 +35,132 @@ public DataFrame(IList<BaseColumn> columns)
...
+            }
+
+            Schema schema = schemaBuilder.Build();
+            List<Apache.Arrow.Array> arrays = new List<Apache.Arrow.Array>();
+
+            int recordBatchLength = Int32.MaxValue;
+            int numberOfRowsInThisRecordBatch = (int)Math.Min(recordBatchLength, RowCount);
+            long numberOfRowsProcessed = 0;
+
+            while (numberOfRowsProcessed < RowCount || RowCount == 0)
```
eerhardt: You may want to add a comment for the `RowCount == 0` clause. I don't think it is obvious why this is here.

src/Microsoft.Data/MutableDataFrameBuffer.cs
```diff
@@ -0,0 +1,91 @@
...
+using System.Runtime.CompilerServices;
+using System.Runtime.InteropServices;
+
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// A basic mutable store to hold values in a DataFrame column. Supports wrapping with an ArrowBuffer
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    internal class MutableDataFrameBuffer<T> : DataFrameBuffer<T>
```
eerhardt: Why is DataFrameBuffer public but this one is internal? Do either need to be public?

eerhardt: I would go the opposite route with naming. We have `Span` and `ReadOnlySpan`, and `Memory` and `ReadOnlyMemory`. We should follow the same naming pattern.

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -11,83 +11,57 @@
...
+            if ((long)numberOfValues * Size > MaxCapacity)
             {
                 throw new ArgumentException($"{numberOfValues} exceeds buffer capacity", nameof(numberOfValues));
             }
-            Memory = new byte[numberOfValues * _size];
+            _readOnlyMemory = new byte[numberOfValues * Size];
         }
 
-        public void Append(T value)
+        public DataFrameBuffer(ReadOnlyMemory<byte> buffer, int length)
```
eerhardt: Why do we need a separate `length`. The buffer already has a Length, why can't we use it?

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -14,10 +15,9 @@
   </ItemGroup>
 
   <ItemGroup>
-    <PackageReference Include="System.Buffers" Version="4.5.0" />
-    <PackageReference Include="System.Memory" Version="4.5.1" />
-    <PackageReference Include="System.Runtime.CompilerServices.Unsafe" Version="4.5.1" />
-    <PackageReference Include="System.Text.Encoding" Version="4.3.0" />
+    <PackageReference Include="Apache.Arrow" Version="0.14.0" />
```
eerhardt: Use `0.14.1`. It has a few critical bug fixes in it.

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -34,6 +35,99 @@ public PrimitiveColumn(string name, long length = 0) : base(name, length, typeof
...
         }
 
+        public PrimitiveColumn(string name, ReadOnlyMemory<byte> buffer, ReadOnlyMemory<byte> nullBitMap, int length = 0, int nullCount = 0) : base(name, length, typeof(T))
+        {
+            _columnContainer = new PrimitiveColumnContainer<T>(buffer, nullBitMap, length, nullCount);
+        }
+
+        private IArrowType GetArrowType()
+        {
+            switch (typeof(T))
```
eerhardt: I think a simple `if-elseif` pattern is much simpler here:

```C#
if (typeof(T) == typeof(bool))
{
   return BooleanType.Default;
}
else if (typeof(T) == typeof(double))
{
   return DoubleType.Default;
}
...
```

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -34,6 +35,99 @@ public PrimitiveColumn(string name, long length = 0) : base(name, length, typeof
...
+        }
+
+        public override Apache.Arrow.Array AsArrowArray(long startIndex, int numberOfRows)
+        {
+            ArrowBuffer valueBuffer = numberOfRows == 0 ? ArrowBuffer.Empty : new ArrowBuffer(_columnContainer.GetValueBuffer(startIndex));
+            ArrowBuffer nullBuffer = numberOfRows == 0 ? ArrowBuffer.Empty : new ArrowBuffer(_columnContainer.GetNullBuffer(startIndex));
+            int nullCount = GetNullCount(startIndex, numberOfRows);
+            int arrayIndex = numberOfRows == 0 ? 0 :_columnContainer.GetArrayContainingRowIndex(startIndex);
+            int offset = (int)(startIndex - arrayIndex * DataFrameBuffer<T>.MaxCapacity);
+            switch (this)
```
eerhardt: Again, I think an `if-else if` pattern would be more simple here.

tests/Microsoft.Data.DataFrame.Tests/ArrowIntegrationTests.cs
```diff
@@ -0,0 +1,235 @@
...
+using System.Linq;
+using Apache.Arrow;
+using Apache.Arrow.Ipc;
+using Xunit;
+
+namespace Microsoft.Data.Tests
+{
+    public class ArrowIntegrationTests
+    {
+        private class ArrayComparer :
```
eerhardt: I think it may be easier to read if these arrow comparison methods were split into their own dedicated class. Then they could also be used from multiple tests in the future as well.

danmosemsft:
I suggest that at some point you set up time with @grabyourpitchforks to discuss how you are handling UTF8. He owns plans around Utf8String etc and you'll want to understand whether it would be useful here in future.

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Believe me, I spent a while just thinking about the name here and decided to just call it `ArrowStringColumn` for now and rename once we have consensus. My inclination was to call it `ImmutableStringColumn`, but that gives an impression that it is related to the already existing `StringColumn`, when they are not really related. OTOH, it does make the immutability constraint obvious. I actually do like `UTF8StringColumn`. It makes the difference with `StringColumn` obvious.

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Good catch on the startIndex. I added a test for this.

No, this is the implementation of `this[startIndex, length]`.

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Both `GetValue` and `new string this[long rowIndex]` are overriding weakly typed methods in `BaseColumn`. If I made `GetValue` call `this[rowIndex]`, it *may* work, but it'd rather not rely on method hiding like that. I did make a new `GetValueImplementation` method though that is strongly typed, so there are no casts now.

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Oh yeah, good idea. I didn't even know this existed. Filed this:
https://github.com/dotnet/corefxlab/issues/2704

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Fixed.

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Fixed. My bad.

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: > Don't you want to override GroupColumnValues instead?

Yup. I missed renaming this somehow. Fixed

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Yup. Fixed. Added a temp patch to fix a crash in the process too that I overlooked. Added a comment to https://github.com/dotnet/corefxlab/issues/2700

src/Microsoft.Data/DataFrameBuffer.cs
pgovind: Because `length` is the Length of the `DataFrameBuffer`, which can be different from `buffer.Length`. This happens because `DataFrameBuffer` stores everything as `Memory<byte>`. So for ex, the test case `TestArrowStringColumnWithNulls` in BufferTests.cs declares a `DataFrameBuffer<int>` like this:

            Memory<byte> offsetMemory = new byte[] { 0, 0, 0, 0, 3, 0, 0, 0, 3, 0, 0, 0, 7, 0, 0, 0, 7, 0, 0, 0 };
            DataFrameBuffer<int> offsetBuffer = new DataFrameBuffer<int>(offsetMemory, 5);

src/Microsoft.Data/MutableDataFrameBuffer.cs
pgovind: I left the read only version public for now to write unit tests. Once Arrow's `StringArray.Builder.Append` can accept nulls(I'll file a bug separately for that), I'll change both the buffers to internal

src/Microsoft.Data/DataFrameBuffer.cs
pgovind: It also occured to me that I could directly wrap an `ArrowBuffer` here instead of deconstructing the `ReadOnlyMemory` from the `ArrowBuffer`. That would remove the issue with specifying a `length` too. Does that seem better?

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,378 @@
...
+        private IList<ReadOnlyDataFrameBuffer<byte>> _nullBitMapBuffers;
+
+        public ArrowStringColumn(string name) : base(name, 0, typeof(string))
+        {
+            _dataBuffers = new List<ReadOnlyDataFrameBuffer<byte>>();
+            _offsetsBuffers = new List<ReadOnlyDataFrameBuffer<int>>();
+            _nullBitMapBuffers = new List<ReadOnlyDataFrameBuffer<byte>>();
+        }
+
+        public ArrowStringColumn(string name, ReadOnlyDataFrameBuffer<byte> dataBuffer, ReadOnlyDataFrameBuffer<int> offsetBuffer, ReadOnlyDataFrameBuffer<byte> nullBitMapBuffer, int length, int nullCount) : base(name, length, typeof(string))
```
pgovind: I'll remove this once Arrow's `StringColumn.Builder.Append` supports appending null values. Leaving this around for the unit tests now

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -0,0 +1,378 @@
...
+        }
+
+        public override Dictionary<TKey, ICollection<long>> GroupColumnValues<TKey>()
+        {
+            if (typeof(TKey) == typeof(string))
+            {
+                Dictionary<string, ICollection<long>> multimap = new Dictionary<string, ICollection<long>>(EqualityComparer<string>.Default);
+                for (long i = 0; i < Length; i++)
+                {
+                    string str = this[i] ?? "__null__";
```
pgovind: This will go away when I fix https://github.com/dotnet/corefxlab/issues/2700

src/Microsoft.Data/DataFrame.cs
```diff
@@ -34,6 +35,140 @@ public DataFrame(IList<BaseColumn> columns)
...
+            int fieldIndex = 0;
+            IEnumerable<IArrowArray> arrowArrays = recordBatch.Arrays;
+            foreach (IArrowArray arrowArray in arrowArrays)
+            {
+                Field field = arrowSchema.GetFieldByIndex(fieldIndex);
+                IArrowType fieldType = field.DataType;
+                BaseColumn dataFrameColumn = null;
+                switch (fieldType.TypeId)
+                {
+                    case ArrowTypeId.Boolean:
```
pgovind: Added unit tests for all the supported types.

tests/Microsoft.Data.DataFrame.Tests/ArrayComparer.cs
```diff
@@ -0,0 +1,169 @@
...
+            Assert.Equal(expectedArray.Length, actualArray.Length);
+            Assert.Equal(expectedArray.NullCount, actualArray.NullCount);
+            Assert.Equal(expectedArray.Offset, actualArray.Offset);
+
+            Assert.True(expectedArray.NullBitmapBuffer.Span.SequenceEqual(actualArray.NullBitmapBuffer.Span));
+            Assert.True(expectedArray.Values.Slice(0, expectedArray.Length).SequenceEqual(actualArray.Values.Slice(0, actualArray.Length)));
+        }
+    }
+
+    internal class FieldComparer
```
eerhardt: This and the classes below should all be `static` classes since they only have static methods.

tests/Microsoft.Data.DataFrame.Tests/BufferTests.cs
```diff
@@ -102,5 +121,45 @@ public void TestAppendMany()
...
+        }
+
+        [Fact]
+        public void TestArrowStringColumnWithNulls()
+        {
+            string data = "joemark";
+            byte[] bytes = Encoding.UTF8.GetBytes(data);
+            Memory<byte> dataMemory = new Memory<byte>(bytes);
+            ReadOnlyDataFrameBuffer<byte> dataBuffer = new ReadOnlyDataFrameBuffer<byte>(dataMemory, 7);
+            Memory<byte> nullMemory = new byte[] { 13 };
```
eerhardt: It may be more readable/understandable if the `13` was written in either hex or binary. You could do `0x0D` or `0b1101`.

tests/Microsoft.Data.DataFrame.Tests/BufferTests.cs
```diff
@@ -102,5 +121,45 @@ public void TestAppendMany()
...
+            ReadOnlyDataFrameBuffer<byte> nullBuffer = new ReadOnlyDataFrameBuffer<byte>(nullMemory, 4);
+            Memory<byte> offsetMemory = new byte[] { 0, 0, 0, 0, 3, 0, 0, 0, 3, 0, 0, 0, 7, 0, 0, 0, 7, 0, 0, 0 };
+            ReadOnlyDataFrameBuffer<int> offsetBuffer = new ReadOnlyDataFrameBuffer<int>(offsetMemory, 5);
+            ArrowStringColumn stringColumn = new ArrowStringColumn("String", dataBuffer, offsetBuffer, nullBuffer, 4, 1);
+
+            //StringArray strArray = new StringArray.Builder().Append("joe").Append("null").Append("mark").Append("").Build();
```
eerhardt: Should we delete this?

src/Microsoft.Data/ReadOnlyDataFrameBuffer.cs
```diff
@@ -0,0 +1,94 @@
...
+            get
+            {
+                if (index > Length)
+                    throw new ArgumentOutOfRangeException(nameof(index));
+                return ReadOnlySpan[index];
+            }
+            set => throw new NotSupportedException();
+        }
+
+        internal bool this[int startIndex, int length, IList<T> returnList]
```
eerhardt: I don't think this should be an indexer. I think it fits more like a normal method. Why does it return a bool? It only ever returns true.

src/Microsoft.Data/ReadOnlyDataFrameBuffer.cs
```diff
@@ -0,0 +1,94 @@
...
+        {
+            get
+            {
+                if (startIndex > Length)
+                    throw new ArgumentOutOfRangeException(nameof(startIndex));
+                long endIndex = Math.Min(Length, startIndex + length);
+                for (int i = startIndex; i < endIndex; i++)
+                {
+                    returnList.Add(ReadOnlySpan[i]);
```
eerhardt: Should cache the span outside of the loop.

src/Microsoft.Data/ReadOnlyDataFrameBuffer.cs
```diff
@@ -0,0 +1,94 @@
...
+namespace Microsoft.Data
+{
+    /// <summary>
+    /// A basic immutable store to hold values in a DataFrame column. Supports wrapping with an ArrowBuffer
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class ReadOnlyDataFrameBuffer<T>
+        where T : struct
+    {
+        // TODO: Change this to Memory<T>
```
eerhardt: ?? Why this TODO?  This is a readonly version.

src/Microsoft.Data/ReadOnlyDataFrameBuffer.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        private ReadOnlyMemory<byte> _readOnlyMemory;
```
eerhardt: should be `private readonly ReadOnlyMemory<byte> _readOnlyMemory;`.  (That's a lot of readonlys)

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -78,6 +79,14 @@ protected set
 
         public virtual GroupBy GroupBy(int columnIndex, DataFrame parent) => throw new NotImplementedException();
 
+        // Arrow related APIs
+        public virtual Field Field() => throw new NotImplementedException();
```
eerhardt: I think things like this should be `protected internal`, like I made them in my IDataView PR. Normal public consumers shouldn't need to see/call these.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -34,6 +35,150 @@ public DataFrame(IList<BaseColumn> columns)
...
+            }
+
+            Schema schema = schemaBuilder.Build();
+            List<Apache.Arrow.Array> arrays = new List<Apache.Arrow.Array>();
+
+            int recordBatchLength = Int32.MaxValue;
+            int numberOfRowsInThisRecordBatch = (int)Math.Min(recordBatchLength, RowCount);
+            long numberOfRowsProcessed = 0;
+
+            // Sometime Spark passes in an empty DataFrame
```
eerhardt: I think this comment could use some work. It isn't understandable in its current form to someone new in the code.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -34,6 +35,150 @@ public DataFrame(IList<BaseColumn> columns)
...
+
+            Schema schema = schemaBuilder.Build();
+            List<Apache.Arrow.Array> arrays = new List<Apache.Arrow.Array>();
+
+            int recordBatchLength = Int32.MaxValue;
+            int numberOfRowsInThisRecordBatch = (int)Math.Min(recordBatchLength, RowCount);
+            long numberOfRowsProcessed = 0;
+
+            // Sometime Spark passes in an empty DataFrame 
+            while (numberOfRowsProcessed < RowCount || RowCount == 0)
```
eerhardt: What do you think of using a `do-while` here instead?

src/Microsoft.Data/DataFrameBuffer.cs
```diff
@@ -3,45 +3,39 @@
...
+        }
 
         public Span<T> Span
         {
             [MethodImpl(MethodImplOptions.AggressiveInlining)]
             get => MemoryMarshal.Cast<byte, T>(Memory.Span);
         }
 
-        public int Length { get; internal set; }
+        public DataFrameBuffer(int numberOfValues = 8) : base(numberOfValues) { }
```
eerhardt: Why have a default value?

src/Microsoft.Data/MutableDataFrameBuffer.cs
eerhardt: I almost wonder if it would be better to use reflection or `InternalsVisibleTo` instead. That way we have the right public API.

src/Microsoft.Data/DataFrameBuffer.cs
eerhardt: I think it would be best to limit the places that were dependent on Arrow here. If we can keep these buffers general purpose, I think that would be good.

src/Microsoft.Data/DataFrame.cs
pgovind: Yup, that looks better. Good idea

src/Microsoft.Data/DataFrameBuffer.cs
pgovind: Hmm, I guess I greedily allocated some space upfront. I'll investigate if I have some logic that depends on this somewhere, but I'd rather do that in a separate patch and get this in. Sound ok? If you agree, I'll open an issue

src/Microsoft.Data/ReadOnlyDataFrameBuffer.cs
pgovind: Oh, that was a copy paste error. Deleted now

src/Microsoft.Data/MutableDataFrameBuffer.cs
pgovind: Oh yeah I'm good with that. I thought using `InternalsVisibleTo` was frowned upon? But I guess this is temporary

src/Microsoft.Data/DataFrameBuffer.cs
pgovind: Ok, done.

eerhardt:
I think you need to sync up with my IDataView change. With that, the ArrowStringColumn should override the new virtual methods for IDataView.

pgovind:
> I think you need to sync up with my IDataView change. With that, the ArrowStringColumn should override the new virtual methods for IDataView.

Sure. I was hoping to do that right after this went in, but I don't mind doing it now

src/Microsoft.Data/DataFrame.cs
```diff
@@ -33,6 +35,153 @@ public DataFrame(IList<BaseColumn> columns)
...
+                }
+                for (int i = 0; i < columnCount; i++)
+                {
+                    BaseColumn column = Column(i);
+                    arrays.Add(column.AsArrowArray(numberOfRowsProcessed, numberOfRowsInThisRecordBatch));
+                }
+                numberOfRowsProcessed += numberOfRowsInThisRecordBatch;
+                yield return new RecordBatch(schema, arrays, numberOfRowsInThisRecordBatch);
+                if (RowCount == 0)
+                    break;
```
eerhardt: Why do you need this check here? The whole loop should suffice.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.IDataView.cs
```diff
@@ -15,11 +15,11 @@ public partial class DataFrameTests
         public void TestIDataView()
         {
             // IDataView doesn't support null values
-            IDataView dataView = MakeDataFrameWithAllColumnTypes(10, withNulls: false);
+            IDataView dataView= MakeDataFrameWithAllColumnTypes(10, withNulls: false);
```
eerhardt: Can this be reverted?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.IDataView.cs
```diff
@@ -76,6 +76,10 @@ public void TestIDataView()
             Assert.Equal("Bool", preview.ColumnView[13].Column.Name);
             Assert.Equal(true, preview.ColumnView[13].Values[0]);
             Assert.Equal(false, preview.ColumnView[13].Values[1]);
+
+            Assert.Equal("ArrowString", preview.ColumnView[14].Column.Name);
+            Assert.Equal("foo".ToString(), preview.ColumnView[14].Values[0].ToString());
```
eerhardt: I would do AsMemory, like above with the normal string.

pgovind: For some reason, that just fails. I don't understand why. The test explorer says this:

```
Message: 
    Assert.Equal() Failure
    Expected: foo
    Actual:   foo
```

eerhardt: That's odd - did you try debugging it? Why would it work for the normal string?

PR #2705, opened 2019-07-29
Implement IDataView on DataFrame.

eerhardt:
Implement a simple IDataView implementation on top of DataFrame. This allows for DataFrame to be passed into training and scoring using ML.NET.

### DIFF
--- a/src/Microsoft.Data/DataFrame.IDataView.cs
+++ b/src/Microsoft.Data/DataFrame.IDataView.cs
@@ -16,11 +16,11 @@ namespace Microsoft.Data
         bool IDataView.CanShuffle => false;
 
         private DataViewSchema _schema;
-        DataViewSchema IDataView.Schema
+        private DataViewSchema DataViewSchema
         {
             get
             {
-                if (_schema != null && _schema.Count == ColumnCount)
+                if (_schema != null)
                 {
                     return _schema;
                 }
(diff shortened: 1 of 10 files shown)


src/Microsoft.Data/DataFrame.IDataView.cs
```diff
@@ -0,0 +1,138 @@
...
+        bool IDataView.CanShuffle => false;
+
+        private DataViewSchema _schema;
+        DataViewSchema IDataView.Schema
+        {
+            get
+            {
+                if (_schema != null && _schema.Count == ColumnCount)
+                {
+                    return _schema;
```
pgovind: Maybe set this to null in the `InsertColumn`, `SetColumn` and `RemoveColumn` methods? Presumably, someone can run an ML algo=>change the dataframe(thereby changing the schema)=>run ML again and so on.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -287,5 +289,18 @@ public override GroupBy GroupBy(int columnIndex, DataFrame parent)
...
+            builder.AddColumn(Name, TextDataViewType.Instance);
+        }
+
+        protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
+        {
+            return CreateValueGetterDelegate(cursor);
+        }
+
+        private ValueGetter<ReadOnlyMemory<char>> CreateValueGetterDelegate(DataViewRowCursor cursor) =>
+            (ref ReadOnlyMemory<char> value) => value = this[cursor.Position].AsMemory();
```
pgovind: This might be interesting once `ArrowStringColumn` goes in.

src/Microsoft.Data/DataFrame.IDataView.cs
eerhardt: Good call.  I'll invalidate this when the DataFrame changes.

src/Microsoft.Data/StringColumn.cs
eerhardt: Yes, when our 2 changes come in, one will have to merge with the other.

danmosemsft:
Was a test needed for the invalidation case?

eerhardt:
> Was a test needed for the invalidation case?

Done. https://github.com/dotnet/corefxlab/pull/2706

PR #2709, opened 2019-08-07
Implement IEnumerable<T> on PrimitiveColumn.

eerhardt:
This allows for things like XPlot to be able to show charts using code like:

```C#
    BaseColumn column = ...;

    Chart.Plot(
        new Graph.Histogram()
        {
            x = column,
            nbinsx = bins
        },
        new Layout.Layout()
        {
            title = column.Name
        }
    ).Display();
```

### DIFF
--- a/tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
+++ b/tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Text;
 using Xunit;
 
 namespace Microsoft.Data.Tests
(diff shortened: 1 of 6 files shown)


tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -988,7 +989,85 @@ void GroupCountAndAssert(DataFrame df)
...
+                if (value == null)
+                {
+                    actualStrings.Append("<null>");
+                }
+                else
+                {
+                    actualStrings.Append(value);
+                }
+            }
+            Assert.Equal("01234<null>6789", actualStrings.ToString());
```
pgovind: Nice!

PR #2710, opened 2019-08-08
Implementations for Clip, Filter, Sample, Merge, ReadCsv and Description

pgovind:
Implementations for Clip, Filter, Sample, Merge, ReadCsv and Description


src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,282 @@
+﻿using System;
+using System.Collections.Generic;
+using System.IO;
+using System.Text;
+
+namespace Microsoft.Data
+{
+
+    public static class Utilities
```
pgovind: I named this Utilities for now, maybe it should be `IO` instead? I don't have a strong opinion either way

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,282 @@
...
+        /// <param name="encoding">text encoding</param>
+        /// <param name="index">add one column with the row index</param>
+        /// <returns>DataFrame</returns>
+        public static DataFrame ReadCsv(string filename,
+                                char sep = ',', bool header = true,
+                                string[] names = null, Type[] dtypes = null,
+                                int nrows = -1, int guess_rows = 10,
+                                Encoding encoding = null, bool index = false)
+        {
+            return ReadStream(() => new StreamReader(filename, encoding ?? Encoding.ASCII),
```
pgovind: Should this be `Encoding.Default` instead?

src/Microsoft.Data/StringColumn.cs
```diff
@@ -223,11 +234,15 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
             StringColumn clone;
             if (!(mapIndices is null))
             {
-                if (mapIndices.DataType != typeof(long))
-                    throw new ArgumentException(Strings.MismatchedValueType + " PrimitiveColumn<long>", nameof(mapIndices));
+                if (mapIndices.DataType != typeof(long) && mapIndices.DataType != typeof(bool))
+                    throw new ArgumentException(String.Format("{0} {1} {2} {3}", Strings.MismatchedValueType, "${typeof(long)}", Strings.Or, "${typeof(bool)}", nameof(mapIndices)));
```
eerhardt: In general, writing localizable strings this way is not preferred because it doesn't give the localizer a chance to change any word ordering.

It's better to put a full format string in the .resx and just have `{0}` placeholders for specific arguments.

So in this case, the string:

```
"Expected value to be of type {0} or {1}."
```

Should be in the .resx.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1069,5 +1115,171 @@ public void TestIEnumerable()
...
         }
+
+        [Fact]
+        public void TestColumnClip()
+        {
+            DataFrame df = MakeDataFrameWithNumericColumns(10);
+            BaseColumn clipped = df["Int"].Clip(3, 7);
+            Assert.Equal(3, clipped[0]);
+            Assert.Equal(3, clipped[1]);
+            Assert.Equal(3, clipped[2]);
```
eerhardt: Should also assert that the values inside are still the same (4, 5, 6)

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1069,5 +1115,171 @@ public void TestIEnumerable()
...
+            Assert.Equal(9, merge.RowCount);
+            Assert.Equal(merge.ColumnCount, left.ColumnCount + right.ColumnCount);
+            Assert.Equal(merge["Int_right"][2], right["Int"][2]);
+            VerifyMerge(merge, left, right, JoinAlgorithm.Inner);
+        }
+
+        [Fact]
+        public void TestDescription()
+        {
+            DataFrame left = MakeDataFrameWithAllMutableColumnTypes(10);
```
eerhardt: (nit) `left` ?

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -303,6 +307,20 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
...
 
+        private ArrowStringColumn Clone(PrimitiveColumn<bool> boolColumn)
+        {
+            if (boolColumn.Length > Length)
+                throw new ArgumentException(Strings.MapIndicesExceedsColumnLenth, nameof(boolColumn));
+            ArrowStringColumn ret = new ArrowStringColumn(Name);
+            for (long i = 0; i < boolColumn.Length; i++)
+            {
+                bool? value = boolColumn[i];
+                if (value.HasValue && value.Value == true)
```
eerhardt: Should just be able to say `if (value == true)`.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -303,6 +307,20 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
...
+                    ret.Append(this[i]);
```
eerhardt: This is less than ideal, since you are going from UTF8 => UTF16 => UTF8. Probably want to optimize this to not have to encode/decode the characters.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -41,7 +41,17 @@ protected set
             get;
         }
 
-        public string Name { get; set; }
+        private string _name;
+        public string Name => _name;
+
+        public void SetName(string newName, DataFrame dataFrame = null)
+        {
+            if (!ReferenceEquals(dataFrame, null))
```
eerhardt: Why are you sometimes using `x is null` and sometimes using `ReferenceEquals(x, null)`. We should be consistent in how we are checking for null.

src/Microsoft.Data/BaseColumn.DataFrameFunctions.cs
```diff
@@ -0,0 +1,27 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data
+{
+    public abstract partial class BaseColumn
```
eerhardt: I'm not sure I understand the name of this file - `BaseColumn.DataFrameFunctions`. Aren't all the functions in BaseColumn "data frame" functions?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1069,5 +1115,171 @@ public void TestIEnumerable()
...
+            DataFrame clipped = df.Clip(3, 7);
+            Assert.Equal(df.ColumnCount, clipped.ColumnCount);
+            for (int c = 0; c < df.ColumnCount; c++)
+            {
+                BaseColumn column = clipped.Column(c);
+                for (int i = 0; i < 3; i++)
+                {
+                    Assert.Equal("3", column[i].ToString());
+                }
+                for (int i = 8; i < 10; i++)
```
eerhardt: Same as above - you need to assert all the values are as expected.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -301,116 +352,62 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
-                    ret.InsertColumn(ret.ColumnCount, newColumn);
-                }
-                for (int i = 0; i < other.ColumnCount; i++)
-                {
-                    BaseColumn newColumn = other.Column(i).Clone();
-                    SetSuffixForDuplicatedColumnNames(ret, newColumn, leftSuffix, rightSuffix);
-                    ret.InsertColumn(ret.ColumnCount, newColumn);
-                }
+                BaseColumn column = Column(i);
+                _table.SetColumnName(column, prefix + column.Name);
```
eerhardt: You're going to need to call `OnColumnsChanged()` after these operations, so the IDataView Schema gets updated properly.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -301,116 +352,62 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
-                    BaseColumn newColumn = Column(i).Clone(mapIndices);
-                    ret.InsertColumn(ret.ColumnCount, newColumn);
-                }
-                for (int i = 0; i < other.ColumnCount; i++)
-                {
-                    BaseColumn newColumn = other.Column(i).Clone(mapIndices);
-                    SetSuffixForDuplicatedColumnNames(ret, newColumn, leftSuffix, rightSuffix);
-                    ret.InsertColumn(ret.ColumnCount, newColumn);
-                }
+                indices[i] = rand.Next((int)Math.Min(Int32.MaxValue, RowCount));
```
eerhardt: You probably don't need to call `(int)Math.Min(Int32.MaxValue, RowCount)` each time through the loop, do you?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -423,8 +420,8 @@ public GroupBy GroupBy(string columnName)
             return column.GroupBy(columnIndex, this);
         }
 
-        // In a GroupBy call, columns get resized. We need to set the RowCount to reflect the true Length of the DataFrame. Internal only. Should not be exposed
-        internal void SetTableRowCount(long rowCount)
+        // In GroupBy and ReadCsv calls, columns get resized. We need to set the RowCount to reflect the true Length of the DataFrame. This does internal validation
+        public void SetTableRowCount(long rowCount)
```
eerhardt: Why are we exposing this now?

src/Microsoft.Data/DataFrame.IO.cs
eerhardt: If the user passes in `null`, I would say we should use the default `StreamReader` constructor - 

[https://source.dot.net/#System.Private.CoreLib/shared/System/IO/StreamReader.cs,180](https://source.dot.net/#System.Private.CoreLib/shared/System/IO/StreamReader.cs,180)

That way `detectEncodingFromByteOrderMarks` gets set as `true`.

---
In reply to: [312262614](https://github.com/dotnet/corefxlab/pull/2710#discussion_r312262614) [](ancestors = 312262614)

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -0,0 +1,238 @@
...
+        /// <param name="names">column names (can be empty)</param>
+        /// <param name="dtypes">column types (can be empty)</param>
+        /// <param name="nrows">number of rows to read</param>
+        /// <param name="guess_rows">number of rows used to guess types</param>
+        /// <param name="index">add one column with the row index</param>
+        /// <returns>DataFrame</returns>
+        public static DataFrame ReadStream(Func<StreamReader> createStream,
+                                char sep = ',', bool header = true,
+                                string[] names = null, Type[] dtypes = null,
+                                long nrows = -1, int guess_rows = 10, bool index = false)
```
eerhardt: We should be using standard .NET parameter names. `nrows` and `guess_rows` do not follow the guidelines - 

[https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-parameters](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-parameters)
[https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/parameter-design](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/parameter-design)

src/Microsoft.Data/DataFrame.cs
```diff
@@ -286,6 +284,59 @@ public IList<IList<object>> Tail(int numberOfRows)
...
+                newColumns.Add(Column(i).Clone(mapIndices, invertMapIndices));
+            }
+            return new DataFrame(newColumns);
+        }
+
+        public void SetColumnName(BaseColumn column, string newName) => _table.SetColumnName(column, newName);
+
+        public DataFrame Description()
+        {
+            bool IsNonNumericColumn(BaseColumn column)
```
eerhardt: This isn't how this should work. What happens if someone wants to make an "Image" column in the future?

Instead, the column should be responsible for telling the DataFrame what it can and can't do.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -301,116 +352,62 @@ public DataFrame Sort(string columnName, bool ascending = true)
...
-            while (index != -1)
+            DataFrame ret = new DataFrame();
+            for (int i = 0; i < ColumnCount; i++)
             {
-                // Pre-existing column. Change name
-                BaseColumn existingColumn = dataFrame.Column(index);
-                dataFrame._table.SetColumnName(existingColumn, existingColumn.Name + leftSuffix);
-                column.Name += rightSuffix;
-                index = dataFrame._table.GetColumnIndex(column.Name);
+                ret.InsertColumn(0, Column(i).Clip(lower, upper));
```
eerhardt: Won't this result in the result data frame having columns in the reversed order?  You are going from 0 .. ColumnCount, and inserting each column at index `0`, which means the last column will be first.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1069,5 +1115,171 @@ public void TestIEnumerable()
...
+            Assert.Equal(3, clipped[1]);
+            Assert.Equal(3, clipped[2]);
+            Assert.Equal(7, clipped[8]);
+            Assert.Equal(7, clipped[9]);
+        }
+
+        [Fact]
+        public void TestDataFrameClip()
+        {
+            DataFrame df = MakeDataFrameWithNumericColumns(10);
```
eerhardt: What happens when I call `df.Clip` on a DataFrame that has a string column?

src/Microsoft.Data/PrimitiveColumn.DataFrameFunctions.cs
```diff
@@ -0,0 +1,69 @@
...
+
+        public override DataFrame Description()
+        {
+            DataFrame ret = new DataFrame();
+            StringColumn stringColumn = new StringColumn("Description", 0);
+            stringColumn.Append("Length");
+            stringColumn.Append("Max");
+            stringColumn.Append("Min");
+            stringColumn.Append("Mean");
+            float max = (float)Convert.ChangeType(Max(), typeof(float));
```
eerhardt: is using `float` an issue when the column is of type `double`?

src/Microsoft.Data/PrimitiveColumn.DataFrameFunctions.cs
```diff
@@ -0,0 +1,69 @@
...
+            return ret;
+        }
+
+        private PrimitiveColumn<T> _Clip(T lower, T upper)
+        {
+            PrimitiveColumn<T> ret = Clone() as PrimitiveColumn<T>;
+            Comparer<T> comparer = Comparer<T>.Default;
+            for (long i = 0; i < Length; i++)
+            {
+                if (!ret.IsValid(i))
```
eerhardt: IMO - this naming doesn't make for readable code. `IsValid` really means `IsNotNull` right?  Maybe we should use the term `null` in the name instead?

Or even better - if we say 

```C#
T? value = ret[i];

if (value == null)
    continue;

...

```

I think it would be even more readable.

src/Microsoft.Data/DataFrame.IO.cs
eerhardt: I think we should leave `ReadCsv` on DataFrame itself. But we should ensure we aren't using internal/private functionality, so other libraries can performantly create DataFrames from other data sources.

src/Microsoft.Data/BaseColumn.cs
pgovind: That's because I overloaded the `==` operator on DataFrame to mean element wise equality. I've since thought many times that writing `ReferenceEquals` to check for null doesn't feel right, but `==` does exist in Pandas

src/Microsoft.Data/BaseColumn.DataFrameFunctions.cs
pgovind: Yup, they are. I just wanted to organize these functions in a different file, similar to `BaseColumn.Computations` and `BaseColumn.BinaryOperations`. The idea was to add all new functions to this file. I didn't really pay much attention to the name here. I'm open to changing it. I was more interested in avoiding a big file like `DataFrame.cs` is now

src/Microsoft.Data/BaseColumn.cs
eerhardt: You can use ‘x is null’ in all cases and it will skip the == operator.

src/Microsoft.Data/BaseColumn.DataFrameFunctions.cs
eerhardt: I’m not sure I like this reasoning.

1. It doesn’t solve the problem because as you add more functions to this file, it is going to fill up as well. 
2. People who are trying to get a grasp on the functionality need to jump back and forth between multiple files just to understand what this class does.

src/Microsoft.Data/DataFrame.cs
pgovind: Because when we read in the csv file, each `BaseColumn` appends values one by one. The columns don't have a reference to the `DataFrame` they are a part of though, so `DataFrame.RowCount` will still return 0. That's why I had to expose this method. I know this is not an ideal design, but I don't have a much better alternative? The `SetTableRowCount` method will do internal validation though and throw, so randomly setting a `RowCount` will fail for example. This method will only set the `RowCount` if all the `BaseColumn`s have the same `RowCount`.

eerhardt: What if we built up all the columns (or even buffers) first, and then when we were done create a DataFrame around them?

eerhardt: BTW - don't let this block you. You can log an issue and follow up on it later, if you want.

src/Microsoft.Data/DataFrame.cs
pgovind: Yup, that was a typo. Nice catch!

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
pgovind: It'll throw a NotImplemented exception. I changed it to a NotSupportedException now

src/Microsoft.Data/PrimitiveColumn.DataFrameFunctions.cs
pgovind: Not that I have seen yet. We will lose some accuracy sure, but I figured it wouldn't matter much for a `Description` method.

src/Microsoft.Data/DataFrame.cs
eerhardt: (there should be a whisper in the back of your head, "add a test....")

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
eerhardt: Why doesn't it just skip the string columns?

pgovind: Yours was a better idea. I'm doing that in the latest commit now

eerhardt:
>             if (value == null)

I don't think a `ReadOnlySpan` can be `null`.

---
Refers to: src/Microsoft.Data/ArrowStringColumn.cs:134 in 6e45c5c. [](commit_id = 6e45c5c640d6bb32d528ebd1ec23a6518a695db0, deletion_comment = False)

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -138,25 +137,28 @@ private void Append(string value)
...
+                value.CopyTo(mutableDataBuffer.Span.Slice(mutableDataBuffer.Length));
+                mutableDataBuffer.Length += value.Length;
+                mutableOffsetsBuffer.Append(mutableOffsetsBuffer[mutableOffsetsBuffer.Length - 1] + value.Length);
             }
             SetValidityBit(Length - 1, value == null ? true : false);
+
         }
 
+        // This is an immutable column, however this method exists to support Clone(). Keep this method private
+        private void Append(string value) => Append(value == null ? null : Encoding.UTF8.GetBytes(value));
```
eerhardt: I'm not 100% sure, but is this only called with `null`? In general I don't think we should be using this method. Can you remove it somehow, to ensure no one calls it in the future, since it will cause needless memory consumption?

(note this can be done in a follow up PR)

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -110,6 +121,25 @@ protected set
...
+        /// <typeparam name="U"></typeparam>
+        /// <param name="lower">Minimum value. All values below this threshold will be set to it</param>
+        /// <param name="upper">Maximum value. All values above this threshold will be set to it</param>
+        public virtual BaseColumn Clip<U>(U lower, U upper) => throw new NotImplementedException();
+
+        /// <summary>
+        /// Determines if the column is of a numeric type
+        /// </summary>
+        /// <remarks>Used by the Description method to filter out descriptons from non-numeric columns</remarks>
+        public virtual bool IsNumericColumn() => false;
```
eerhardt: I don't think this is the right design. Your `remarks` even suggests why naming this `IsNumericColumn` is a bad idea: `Used by the Description method to filter out descriptons from non-numeric columns` - why couldn't a non-numeric column want a description?

Instead, I would think this should be `HasDescription()` or `SupportsDescription()`. It could even be a property instead of a method.

src/Microsoft.Data/DataFrame.IO.cs
eerhardt: Even `nRows` and `dTypes` still don't follow the naming guidelines.

Typically this would be `rowCount` and `types`.

---
In reply to: [313009266](https://github.com/dotnet/corefxlab/pull/2710#discussion_r313009266) [](ancestors = 313009266)

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -30,16 +30,10 @@ public DataFrameTable()
         public DataFrameTable(IList<BaseColumn> columns)
         {
             columns = columns ?? throw new ArgumentNullException(nameof(columns));
-            _columns = columns;
-            ColumnCount = columns.Count;
-            if (columns.Count > 0)
+            _columns = new List<BaseColumn>();
```
eerhardt: Why duplicate the list now? Why not just hold onto the List that was passed in?

src/Microsoft.Data/DataFrame.cs
eerhardt: Was there a test/verification added to ensure the columns were in the correct order?

---
In reply to: [313542565](https://github.com/dotnet/corefxlab/pull/2710#discussion_r313542565) [](ancestors = 313542565)

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Removed

src/Microsoft.Data/BaseColumn.cs
pgovind: Sure, this is just the first pass at the `Description` method and I was planning on adding a minor improvement it in a follow up PR. The reason I went with `IsNumericColumn` is because `DataFrame.Description()` will call each `BaseColumn`'s `Description` method and then perform a `Merge`. The numeric columns have properties such as `Length, Max` etc, but the string columns don't have anything yet => The `Merge` will just produce an empty `DataFrame`, since there won't be any common properties. I could instead perform an `OuterMerge` which will preserve all the `Description`s, but I think we can inspect what columns are in the DataFrame and do something a little more intelligent here. I changed it to `HasDescription` for now.

src/Microsoft.Data/DataFrameTable.cs
pgovind: 2 reasons:
1) I didn't want any external references to the list that the DataFrame is holding onto. This is mostly so that no columns can be added or removed without going through the DataFrame APIs. Considering that the list is only holding references to the columns, I figured this would not be too expensive. It's still not perfect though. The columns themselves could be changed independently, but this would not break the internal book-keeping for the most part
2) `InsertColumn` does some internal validation that I wanted to re-use.

pgovind:
> I don't think a ReadOnlySpan can be null.

I thought so too, but https://source.dot.net/#System.Private.CoreLib/shared/System/ReadOnlySpan.Fast.cs,42 says otherwise? I added a test case too to make sure and it behaves fine. Passing a null returns a default ReadOnlySpan that has the right behavior for null checks.

tests/Microsoft.Data.DataFrame.Tests/DataFrame.IOTests.cs
```diff
@@ -0,0 +1,34 @@
...
+using System.Runtime.CompilerServices;
+using System.Text;
+using Xunit;
+
+namespace Microsoft.Data.Tests
+{
+    public partial class DataFrameTests
+    {
+        [Fact]
+        public void TestReadCsv()
```
eerhardt: We probably need way more Read Csv tests than this.... but this is fine for now. Can you log an issue to add more tests in this area?

src/Microsoft.Data/strings.resx
```diff
@@ -139,7 +145,13 @@
     <value>rowCount differs from Column length for Column </value>
   </data>
   <data name="MismatchedValueType" xml:space="preserve">
-    <value>Expected value to be of type</value>
+    <value>Expected value to be of type </value>
+  </data>
+  <data name="MultipleMismatchedValueType" xml:space="preserve">
+    <value>Expected value to be of type {0} or {1}</value>
+  </data>
+  <data name="Or" xml:space="preserve">
```
eerhardt: Is this still used?

eerhardt:
>> I don't think a ReadOnlySpan can be null.

> I thought so too, but https://source.dot.net/#System.Private.CoreLib/shared/System/ReadOnlySpan.Fast.cs,42 says otherwise? I added a test case too to make sure and it behaves fine. Passing a null returns a default ReadOnlySpan that has the right behavior for null checks.

Ok, I understand what is happening now - there is an implicit operator from an array to a Span, and `null` can be an array.

I think to make this make more sense, you should replace the `null` usages with `default` (both in the check in Append, and all the places that call `Append(null)`). Then we should have some good comments in `Append` about the subtle difference between appending a `null` string (which will set the null bitmap bit) and appending an empty string (which won't set the null bitmap bit, but will make a new offsets entry with `length == 0`).

ahsonkhan:
> Ok, I understand what is happening now - there is an implicit operator from an array to a Span, and `null` can be an array.

Yep. The hidden implicit cast from T[] to Span has tripped up some folks in the past when they set a ref struct to null. Span itself cannot be "null".

tests/Microsoft.Data.DataFrame.Tests/DataFrame.IOTests.cs
pgovind: Yup, I agree. Filed https://github.com/dotnet/corefxlab/issues/2713

src/Microsoft.Data/strings.resx
pgovind: Nope. Good catch. Removed

pgovind:
> I think to make this make more sense, you should replace the null usages with default (both in the check in Append, and all the places that call Append(null)).

Done

PR #2712, opened 2019-08-15
 .NET Core 3.0 Preview8 IntelliSense (English) 

carlossanlop:
These are the IntelliSense files obtained from the latest Docs repo Live build, generated for Preview8.
Some files are gone: Ahson and I investigated each one of them and all the removals seem to be expected.
The APIs did not disappear, they were moved to a different assembly.


carlossanlop:
One of the reasons why some files were gone is because this Live build was generated using the latest batch of reference assemblies that @safern sent to @mairaw. The previous batch was sent to the Docs team about a month ago, so the changes are expected.

ahsonkhan:
Looking more closely, a lot of files seem to be getting removed now. Can you spot check a couple (maybe System.xml and System.Security.xml) and make sure those APIs still exist/are documented elsewhere? I remember we saw System.Xml.xml and found the documentation/types were already within System.Xml.XmlReaderWriter.xml, but I don't see that file here.

PR #2714, opened 2019-08-19
Filter and mean

pgovind:

This rounds out all the APIs I'd consider essential for a DataFrame


src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -323,37 +325,57 @@ private ArrowStringColumn Clone(PrimitiveColumn<bool> boolColumn)
             return ret;
         }
 
-        private ArrowStringColumn Clone(PrimitiveColumn<long> mapIndices, bool invertMapIndex)
+        private ArrowStringColumn CloneImplementation(BaseColumn mapIndices, Func<long, long?> getIndex, bool invertMapIndices = false)
```
pgovind: This is all just refactor

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -181,6 +182,32 @@ protected override void SetValue(long rowIndex, object value)
             }
         }
 
+        public override double Median()
+        {
+            // Not the most efficient implementation. Using a selection algorithm here would be O(n) instead of O(nLogn)
```
pgovind: I'm not very inclined to write an efficient selection method. Primarily because I want efficient stats to be provided by an external lib. LinqStatistics for example

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -374,17 +452,12 @@ public override GroupBy GroupBy(int columnIndex, DataFrame parent)
             }
         }
 
-        public void ApplyElementwise(Func<T, T> func)
+        public void ApplyElementwise(Func<T?, long, T?> func)
```
pgovind: Is there a typical way to inform users that this function is not thread safe?

eerhardt:
Looks like the CI is failing.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -128,11 +140,23 @@ public void SetName(string newName, DataFrame dataFrame = null)
...
+        /// <param name="lower"></param>
+        /// <param name="upper"></param>
+        public virtual BaseColumn Filter<U>(U lower, U upper) => throw new NotImplementedException();
+
         /// <summary>
         /// Determines if the column is of a numeric type
         /// </summary>
         public virtual bool IsNumericColumn() => false;
 
+        public virtual double Mean() => throw new NotImplementedException();
```
gfoidl: Are the xml-comments intentionally left out?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1297,11 +1312,62 @@ public void TestDescription()
             {
                 BaseColumn column = description.Column(i);
                 Assert.Equal(4, column.Length);
-                Assert.Equal((float)10, column[0]);
+                Assert.Equal((float)9, column[0]);
```
eerhardt: Why did this change?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -3,14 +3,23 @@
...
 using System.Diagnostics;
 using System.Text;
+using System.Threading.Tasks;
 using Apache.Arrow;
 using Apache.Arrow.Types;
 
 namespace Microsoft.Data
 {
+
+    public enum DropNullOptions
```
eerhardt: xml comments

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -145,10 +147,9 @@ protected override object GetValue(long startIndex, int length)
             return _columnContainer[startIndex, length];
         }
 
-        protected override object GetValue(long rowIndex)
-        {
-            return _columnContainer[rowIndex];
-        }
+        internal T? GetTypedValue(long rowIndex) => _columnContainer[rowIndex];
```
eerhardt: What's the difference between this method and the `T? this[long rowIndex]` indexer?

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -222,6 +249,37 @@ public override bool IsNumericColumn()
             return ret;
         }
 
+        public override BaseColumn FillNulls(object value, bool inPlace = false)
+        {
+            T TValue = (T)Convert.ChangeType(value, typeof(T));
```
eerhardt: `TValue` - variables should be camel cased, so readers don't get confused that `TValue` is a generic type.

src/Microsoft.Data/PrimitiveColumn.cs
eerhardt: Instance methods are never considered thread safe, unless they are explicitly stated to be.

---
In reply to: [315443932](https://github.com/dotnet/corefxlab/pull/2714#discussion_r315443932) [](ancestors = 315443932)

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -374,17 +452,12 @@ public override GroupBy GroupBy(int columnIndex, DataFrame parent)
             }
         }
 
-        public void ApplyElementwise(Func<T, T> func)
+        public void ApplyElementwise(Func<T?, long, T?> func)
         {
-            foreach (ReadOnlyDataFrameBuffer<T> buffer in _columnContainer.Buffers)
+            Parallel.For(0, Length, index =>
```
eerhardt: Why use a `Parallel.For` here? Have you done benchmarks that show this actually makes a performance improvement?

src/Microsoft.Data/BaseColumn.cs
pgovind: Fixed

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
pgovind: Because `Description()` now returns `Length - NullCount` instead of just `Length`. In a notebook, this will help identify which columns have nulls.

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: None. I initially thought it wasn't worth refactoring since it's just 1 line. The indexer now calls `GetTypedValue`. `GetTypedValue` itself exists because I want to pass the method as a delegate to other functions

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: Benchmarks will follow soon. Also, based on the benchmarks I just ran in a private branch, I'm going to revert this change. `Parallel.For` is about 50% slower than the regular for loop on my machine. It was on a large 50000 element column too, so not something I expected. I'll revert the change.

pgovind: Oh, the `Parallel.For` loop is 50% faster than a regular for loop, but 50% slower than the unrolled version on the left. The unrolled version drills down to the Span first and then starts applying `func`, whereas the regular for loop would involve indexing calls for each loop iteration in addition to applying `func`. I bet cache effects are in play here. I wonder if the generated IL code makes it obvious, but that investigation is for later this week.

eerhardt: That is what I expected, I've seen this in the past. If you are doing a relatively small amount of work inside the loop, using a `Parallel.For` will slow your program down because of the overhead due to switching and coordinating between threads.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -182,6 +182,22 @@ public void AppendMany(T? value, long count)
             }
         }
 
+        public void ApplyElementwise(Func<T?, long, T?> func)
+        {
+            for (int b = 0; b < Buffers.Count; b++)
+            {
+                ReadOnlyDataFrameBuffer<T> buffer = Buffers[b];
+                long prevLength = Buffers[0].Length * b;
```
eerhardt: You probably want a `checked` here, to ensure this doesn't overflow.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -3,14 +3,28 @@
...
+using System.Threading.Tasks;
 using Apache.Arrow;
 using Apache.Arrow.Types;
 
 namespace Microsoft.Data
 {
+
+    /// <summary>
+    /// Options for DropNull(). 
+    /// "Any" drops a row if any of the row values are null.
```
eerhardt: These usually go on the enum value itself. Example:


```C#
    /// <summary>
    /// Options for DropNull(). 
    /// </summary>
    public enum DropNullOptions
    {
        /// <summary>Drops a row if any of the row values are null. </summary>
        Any,
        /// <summary>Drops a row when all of the row values are null. </summary>
        All
    }
```

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -93,8 +93,20 @@ public void SetName(string newName, DataFrame dataFrame = null)
...
+        public virtual DataFrame ValueCounts() => throw new NotImplementedException();
+
         public virtual GroupBy GroupBy(int columnIndex, DataFrame parent) => throw new NotImplementedException();
 
+        /// <summary>
+        /// Returns a new column with nulls replaced by value
+        /// </summary>
+        /// <remarks>Tries to convert value to the column's DataType</remarks>
+        /// <param name="value"></param>
+        public virtual BaseColumn FillNulls(object value, bool inPlace = false) => throw new NotImplementedException();
```
eerhardt: I don't see this method implemented for `StringColumn`.  Also, I would think that `ArrowStringColumn` should override this and throw a `NotSupportedException` instead.

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -93,8 +93,20 @@ public void SetName(string newName, DataFrame dataFrame = null)
...
+        /// </summary>
+        public virtual DataFrame ValueCounts() => throw new NotImplementedException();
+
         public virtual GroupBy GroupBy(int columnIndex, DataFrame parent) => throw new NotImplementedException();
 
+        /// <summary>
+        /// Returns a new column with nulls replaced by value
+        /// </summary>
+        /// <remarks>Tries to convert value to the column's DataType</remarks>
+        /// <param name="value"></param>
```
ahsonkhan: Fyi, if you want completeness when it comes to documentation, the triple slash comments should contain all the params being documented, along with a return tag to describe what's being returned.

You also would want to document the exception behavior of the API (at least the ones that have been implemented), for example argumentnullexception/etc.

**If complete documentation is being deferred for later, that is fine.** Just something to remember. When initially implementing/documenting the JSON APIs, not all of the "required" elements (such as `<returns>`) were initially commented, and that lead to us having to go back and add the elements to the docs repo later.

E.g.:  https://github.com/dotnet/dotnet-api-docs/blob/7fff15ca4397fef8fddce8c38cfa1ae3d14a885e/xml/System.Text.Json/JsonElement.xml#L70-L76

cc @carlossanlop

PR #2717, opened 2019-09-05
Perf improvements in Clone, Sort, Merge and ReadCsv

pgovind:
Perf work towards dotnet/corefx#26845. 

Most of the changes are related to caching the Span<T> instead of randomly accessing Memory<T> every iteration. Some algorithms(for ex: Merge) need random access, so there's not much we can do there.

### DIFF
--- a/src/Microsoft.Data/DataFrame.IO.cs
+++ b/src/Microsoft.Data/DataFrame.IO.cs
@@ -127,23 +127,28 @@ namespace Microsoft.Data
             using (var st = createStream())
             {
                 string line = st.ReadLine();
-                int nbline = 0;
-                while (line != null && (numRows == -1 || rowline < numRows))
+                long nbline = 0;
+                while (line != null)
                 {
-                    var spl = line.Split(separator);
-                    if (header && nbline == 0)
+                    if ((numRows == -1) || rowline < numRows)
                     {
-                        if (columnNames == null)
-                            columnNames = spl;
-                    }
-                    else
-                    {
-                        ++rowline;
                         if (linesForGuessType.Count < guessRows)
                         {
-                            linesForGuessType.Add(spl);
-                            numberOfColumns = Math.Max(numberOfColumns, spl.Length);
+                            var spl = line.Split(separator);
+                            if (header && nbline == 0)
+                            {
+                                if (columnNames == null)
+                                    columnNames = spl;
+                            }
+                            else
+                            {
+                                ++rowline;
+                                linesForGuessType.Add(spl);
+                                numberOfColumns = Math.Max(numberOfColumns, spl.Length);
+                            }
                         }
+                        else
+                            ++rowline;
                     }
                     ++nbline;
                     line = st.ReadLine();
@@ -161,17 +166,17 @@ namespace Microsoft.Data
                 Type kind = GuessKind(i, linesForGuessType);
                 if (kind == typeof(bool))
                 {
-                    BaseColumn boolColumn = new PrimitiveColumn<bool>(columnNames[i]);
+                    BaseColumn boolColumn = new PrimitiveColumn<bool>(columnNames[i], rowline);
                     columns.Add(boolColumn);
                 }
                 else if (kind == typeof(float))
                 {
-                    BaseColumn floatColumn = new PrimitiveColumn<float>(columnNames[i]);
+                    BaseColumn floatColumn = new PrimitiveColumn<float>(columnNames[i], rowline);
                     columns.Add(floatColumn);
                 }
                 else if (kind == typeof(string))
                 {
-                    BaseColumn stringColumn = new StringColumn(columnNames[i], 0);
+                    BaseColumn stringColumn = new StringColumn(columnNames[i], rowline);
                     columns.Add(stringColumn);
                 }
                 else
@@ -218,7 +223,6 @@ namespace Microsoft.Data
             for (int i = 0; i < columns.Count; i++)
             {
                 BaseColumn column = columns[i];
-                column.Resize(rowIndex + 1);
                 string val = values[i];
                 Type dType = column.DataType;
                 if (dType == typeof(bool))
--- a/src/Microsoft.Data/DataFrame.IO.cs
+++ b/src/Microsoft.Data/DataFrame.IO.cs
@@ -97,7 +97,7 @@ namespace Microsoft.Data
                                 bool addIndexColumn = false)
         {
             return ReadStream(() => new StreamReader(filename),
-                              separator: separator, header: header, columnNames: columnNames, dataTypes: dataTypes, numRows: numRows,
+                              separator: separator, header: header, columnNames: columnNames, dataTypes: dataTypes, numberOfRowsToRead: numRows,
                               guessRows: guessRows, addIndexColumn: addIndexColumn);
         }
 
@@ -110,47 +110,48 @@ namespace Microsoft.Data
         /// <param name="header">has a header or not</param>
         /// <param name="columnNames">column names (can be empty)</param>
         /// <param name="dataTypes">column types (can be empty)</param>
-        /// <param name="numRows">number of rows to read</param>
+        /// <param name="numberOfRowsToRead">number of rows to read not including the header(if present)</param>
         /// <param name="guessRows">number of rows used to guess types</param>
         /// <param name="addIndexColumn">add one column with the row index</param>
         /// <returns>DataFrame</returns>
         public static DataFrame ReadStream(Func<StreamReader> createStream,
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
-                                long numRows = -1, int guessRows = 10, bool addIndexColumn = false)
+                                long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false)
         {
             var linesForGuessType = new List<string[]>();
             long rowline = 0;
             int numberOfColumns = 0;
 
+            if (header == true && numberOfRowsToRead != -1)
+                numberOfRowsToRead++;
+
             // First pass: schema and number of rows.
             using (var st = createStream())
             {
                 string line = st.ReadLine();
-                long nbline = 0;
                 while (line != null)
                 {
-                    if ((numRows == -1) || rowline < numRows)
+                    if ((numberOfRowsToRead == -1) || rowline < numberOfRowsToRead)
                     {
                         if (linesForGuessType.Count < guessRows)
                         {
                             var spl = line.Split(separator);
-                            if (header && nbline == 0)
+                            if (header && rowline == 0)
                             {
                                 if (columnNames == null)
                                     columnNames = spl;
                             }
                             else
                             {
-                                ++rowline;
                                 linesForGuessType.Add(spl);
                                 numberOfColumns = Math.Max(numberOfColumns, spl.Length);
                             }
                         }
-                        else
-                            ++rowline;
                     }
-                    ++nbline;
+                    ++rowline;
+                    if (rowline == numberOfRowsToRead)
+                        break;
                     line = st.ReadLine();
                 }
             }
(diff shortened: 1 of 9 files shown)


pgovind:
Current Benchmark numbers:

          Method |        Mean |     Error |     StdDev |
---------------- |------------:|----------:|-----------:|
   ColumnIndexer |  7,848.6 us | 102.60 us |  90.953 us |
      ColumnSort | 24,445.9 us | 451.15 us | 422.007 us |
            Sort | 35,689.9 us | 306.84 us | 272.005 us |
            Clip |  5,140.9 us |  21.64 us |  19.187 us |
         GroupBy |    666.4 us |  10.62 us |   9.934 us |
 ColumnFillNulls |  4,726.6 us |  31.59 us |  28.002 us |
        LeftJoin | 10,519.7 us |  96.58 us |  85.615 us |
       RightJoin | 14,755.4 us | 212.28 us | 188.179 us |
       OuterJoin |  7,698.4 us |  77.77 us |  68.945 us |
       InnerJoin | 22,329.7 us | 432.96 us | 515.414 us |
       LeftMerge | 51,758.0 us | 455.56 us | 426.135 us |
      RightMerge | 37,633.9 us | 352.28 us | 329.522 us |
      OuterMerge | 55,329.4 us | 597.64 us | 499.054 us |
      InnerMerge | 44,772.9 us | 645.16 us | 571.917 us |


New Benchmark numbers with this patch:
          Method |        Mean |     Error |    StdDev |
---------------- |------------:|----------:|----------:|
   ColumnIndexer |  8,059.5 us | 159.74 us | 202.03 us |
      ColumnSort | 19,115.4 us | 379.31 us | 654.29 us |
            Sort | 20,351.7 us | 172.31 us | 161.18 us |
            Clip |  5,618.0 us |  70.38 us |  65.84 us |
         GroupBy |    751.9 us |  11.85 us |  11.09 us |
 ColumnFillNulls |  2,414.2 us |  24.26 us |  22.69 us |
        LeftJoin | 11,071.9 us |  75.09 us |  58.63 us |
       RightJoin |  6,078.1 us |  79.59 us |  74.45 us |
       OuterJoin |  7,822.8 us |  71.70 us |  59.87 us |
       InnerJoin |  5,012.9 us |  37.73 us |  35.29 us |
       LeftMerge | 24,607.5 us | 205.79 us | 182.42 us |
      RightMerge | 21,471.6 us | 383.92 us | 320.59 us |
      OuterMerge | 27,069.6 us | 536.02 us | 617.28 us |
      InnerMerge | 17,966.3 us | 296.75 us | 263.06 us |

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -292,8 +293,6 @@ public override BaseColumn Clone(BaseColumn mapIndices = null, bool invertMapInd
                 Type dataType = mapIndices.DataType;
                 if (dataType != typeof(long) && dataType != typeof(int) && dataType != typeof(bool))
                     throw new ArgumentException(String.Format(Strings.MultipleMismatchedValueType, typeof(long), typeof(int), typeof(bool)), nameof(mapIndices));
-                if (mapIndices.Length > Length)
```
eerhardt: Why is this being removed?

src/Microsoft.Data/PrimitiveColumn.Sort.cs
```diff
@@ -39,16 +42,33 @@ private void GetSortIndices(IComparer<T> comparer, out PrimitiveColumn<long> col
...
+                int index = bufferSortIndices[bufferIndex][startIndex];
+                T value;
+                ReadOnlyMemory<byte> buffer = _columnContainer.Buffers[bufferIndex].ReadOnlyMemory;
+                ReadOnlyMemory<T> typedBuffer = Unsafe.As<ReadOnlyMemory<byte>, ReadOnlyMemory<T>>(ref buffer);
+                if (!typedBuffer.IsEmpty)
+                {
+                    bool isArray = MemoryMarshal.TryGetArray(typedBuffer, out ArraySegment<T> array);
+                    if (isArray)
+                    {
+                        value = array.Array[index];
```
eerhardt: You need to use the ArraySegment.Offset here.  The ArraySegment might point to somewhere inside the array, and not at the beginning.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -273,26 +271,76 @@ private StringColumn Clone(PrimitiveColumn<bool> boolColumn)
...
+                        setBuffer[(int)index] = null;
+                        ret._nullCount++;
+                        return mapIndex;
+                    }
+                    setBuffer[(int)index] = getBuffer[mapIndex.Value];
+                    return mapIndex;
+                });
             }
+            else
+                throw new NotImplementedException(nameof(mapIndices.DataType));
```
eerhardt: Should probably give a better exception and message here. Maybe an ArgumentException since `mapIndices` wasn't a valid argument.

Can a customer in the public API ever hit this? If not, maybe add an assert here.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -273,26 +271,76 @@ private StringColumn Clone(PrimitiveColumn<bool> boolColumn)
...
+                        index = longMapIndices.Length - 1 - index;
+                    if (index < setBufferMinRange || index >= setBufferMaxRange)
+                    {
+                        int bufferIndex = (int)(index / maxCapacity);
+                        setBuffer = ret._stringBuffers[bufferIndex];
+                        setBufferMinRange = bufferIndex * maxCapacity;
+                        setBufferMaxRange = (bufferIndex + 1) * maxCapacity;
+                    }
+                    index -= setBufferMinRange;
+                    if (mapIndex == null)
```
eerhardt: Should you check for this first? No reason to do the above if `mapIndex` is null.

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -127,23 +127,28 @@ private static Type MaxKind(Type a, Type b)
             using (var st = createStream())
             {
                 string line = st.ReadLine();
-                int nbline = 0;
-                while (line != null && (numRows == -1 || rowline < numRows))
+                long nbline = 0;
```
eerhardt: What does `nbline` do? Can you just use `rowLine` instead? Also, it should probably have a better, more readable name.

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -127,23 +127,28 @@ private static Type MaxKind(Type a, Type b)
...
+                while (line != null)
```
eerhardt: You are now looping over all the rows twice, correct?

eerhardt: I wonder if we could use a `ReadOnlySpan<char>` to parse the file, instead of creating a new string for every line.

---
In reply to: [321825779](https://github.com/dotnet/corefxlab/pull/2717#discussion_r321825779) [](ancestors = 321825779)

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -188,55 +189,53 @@ public void ApplyElementwise(Func<T?, long, T?> func)
...
+                    T? value = func(isValid ? span[i] : default(T?), curIndex);
+                    span[i] = value.GetValueOrDefault();
+                    SetValidityBit(nullBitMapSpan, i, value != null);
+                }
             }
         }
 
-        public bool IsValid(long index) => NullCount == 0 || GetValidityBit(index);
+        // Faster to use when we already have a span since it avoids indexing
+        public bool IsValid(ref Span<byte> bitMapBufferSpan, int index)
```
eerhardt: You shouldn't need this overload - you should just need the one that takes a ReadOnlySpan.  Spans will implicitly convert to ReadOnlySpans.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -188,55 +189,53 @@ public void ApplyElementwise(Func<T?, long, T?> func)
...
 
-        /// <summary>
-        /// A null value has an unset bit
-        /// A NON-null value has a set bit
-        /// </summary>
-        /// <param name="index"></param>
-        /// <param name="value"></param>
-        private void SetValidityBit(long index, bool value)
+        // Faster to use when we already have a span since it avoids indexing
+        public bool IsValid(ref ReadOnlySpan<byte> bitMapBufferSpan, int index)
```
eerhardt: `bitMapBufferSpan` shouldn't be `ref ReadOnlySpan`, but instead just a regular `ReadOnlySpan` parameter.

src/Microsoft.Data/DataFrame.IO.cs
pgovind: > You are now looping over all the rows twice, correct?

Yup.

> I wonder if we could use a ReadOnlySpan<char> to parse the file, instead of creating a new string for every line.

Hmm, maybe? This ReadCsv method is just a method that I put together so we can work on real files on disk. At the moment, the `Split` calls on each line are taking up a sizeable chunk of time. Maybe I should file an issue to revamp this method now that we have decent capabilities in the DataFrame

src/Microsoft.Data/PrimitiveColumn.Sort.cs
pgovind: Hmm, I don't understand, array.Array will return the original `Array` right? Shouldn't it be safe to then index into the original `Array`?

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: For the new `Merge` algorithm in this patch. Because processing column-wise is more efficient, `Merge` with left, right and outer joins pass in a `mapIndices` larger than `Length`. So now, for ex: when the current column's length is say 5, but a right join produces a column of length 10, it is passed in a `mapIndices` that looks like [0, 1, 2, 3, 4, null, null, null, null, null]. Each null mapIndex appends a null in the resulting column

src/Microsoft.Data/PrimitiveColumn.Sort.cs
eerhardt:
```C#
MemoryMarshal.TryGetArray(typedBuffer, out ArraySegment<T> array);
```

`TryGetArray` and naming `ArraySegment<T>` as `array` are confusing you. `ArraySegment` is a section of an array. `typedBuffer` is a `ReadOnlyMemory`. A `Memory` instance may be pointing into the middle of an array. So say the whole backing array might look like:

```
[5, 4, 3, 2, 1]
```

`typedBuffer` might be pointing at just the `4, 3, 2` part of the total backing array.

That's why this method returns an `ArraySegment` - a section of the array. In the example above, you'd get the original `[5, 4, 3, 2 , 1]` array back, but the `ArraySegment.Offset` will be `1`, meaning the segment of the array that the buffer is pointing to starts at element `1` of the array.

In your code, `index` is the index into `typedBuffer`. And `typedBuffer` starts at `ArraySegment.Offset` into `ArraySegment.Array`. So you need to add `index` and `ArraySegment.Offset` together to get the total index into the backing array.

src/Microsoft.Data/DataFrame.IO.cs
eerhardt: > At the moment, the Split calls on each line are taking up a sizeable chunk of time.

Right, this is because new `string` instances are being created, and their contents are being copied out, for each section in the file. Using a span to parse will make it faster because you don't need to create all those intermediate `string` instances.

src/Microsoft.Data/StringColumn.cs
pgovind: Unfortunately no, I did try. The next line depends on `index` which depends on `setBufferMinRange` which is set in the lines above.

src/Microsoft.Data/StringColumn.cs
pgovind: Public API never hits this. Added a Debug assert

src/Microsoft.Data/PrimitiveColumn.Sort.cs
pgovind: Ah, now I see why you said that. So, in this case, `typedBuffer` would always be the full buffer, never a slice. So `ArraySegment.Offset` would always be 0. But, I get the point. I'll change it to `ArraySegment.Offset + index` just in case `typedBuffer` stops being the full buffer in the future.

eerhardt: > in this case, typedBuffer would always be the full buffer, never a slice.

How are you guaranteeing that? What if the data came from Apache Arrow?  There, we might have taken a giant `byte[]` and sliced it all up to form the individual Arrow columns.

src/Microsoft.Data/DataFrame.IO.cs
pgovind: Filed https://github.com/dotnet/corefxlab/issues/2720

src/Microsoft.Data/PrimitiveColumn.Sort.cs
ahsonkhan: Just a general note of caution (as I am passing by). Try to avoid the use of "unsafe" types/classes like `MemoryMarshal` (basically anything in the `S.R.InteropServices` namespace) and `S.R.CS.Unsafe`. I would default to use the safe APIs and only use the unsafe/advance APIs in limited/targeted scenarios where it is clear that the perf wins are noticeable and worth the maintenance burden (and potential correctness risk). That said, I haven't looked deeply into its usage in the context of this PR/feature, so treat this note as just a heads up to scrutinize usage of such APIs more closely during code review (I got the sense that a bunch of type-safety/etc. was being forgone here).

eerhardt: > (I got the sense that a bunch of type-safety/etc. was being forgone here)

In general, the Arrow format forces a bunch of type-safety to be forgone. At its heart, Arrow takes a collection of bytes, and reinterprets them as different types depending on the type information stored in the Arrow metadata. So while we have a `Memory<byte>` underneath, we need to reinterpret it as a `Memory<int|float|etc>` later in order to read/write integers, floats, etc from the data.

PR #2718, opened 2019-09-05
WIP: Just to get eerhardt's thoughts

pgovind:
Do not merge. This is just to get some eyes on this idea.

### DIFF
--- a/src/Microsoft.Data/PrimitiveColumn.cs
+++ b/src/Microsoft.Data/PrimitiveColumn.cs
@@ -19,7 +19,7 @@ namespace Microsoft.Data
     /// A column to hold primitive types such as int, float etc.
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    public partial class PrimitiveColumn<T> : BaseColumn, IEnumerable<T?>
+    public partial class PrimitiveColumn<T> : BaseColumn, IEnumerable<ReadOnlyMemory<T>>
         where T : unmanaged
     {
         private PrimitiveColumnContainer<T> _columnContainer;
@@ -234,7 +234,7 @@ namespace Microsoft.Data
 
         public bool IsValid(long index) => _columnContainer.IsValid(index);
 
-        public IEnumerator<T?> GetEnumerator() => _columnContainer.GetEnumerator();
+        public IEnumerator<ReadOnlyMemory<T>> GetEnumerator() => _columnContainer.GetEnumerator();
 
         protected override IEnumerator GetEnumeratorCore() => GetEnumerator();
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -14,7 +16,7 @@ namespace Microsoft.Data
     /// PrimitiveDataFrameColumnContainer is just a store for the column data. APIs that want to change the data must be defined in PrimitiveDataFrameColumn
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    internal partial class PrimitiveColumnContainer<T> : IEnumerable<T?>
+    internal partial class PrimitiveColumnContainer<T> : IEnumerable<ReadOnlyMemory<T>>
```
pgovind: I'm not sure if this should be a `IEnumerable<ReadOnlyMemory<T>>` or a `IEnumerable<Memory<T>>`. A `PrimitiveColumnContainer` could hold either `ReadOnlyMemory<T>` or `Memory<T>` in its `Buffers`. `IEnumerable<ReadOnlyMemory<T>>` would work for either, but it'd be immutable. If we instead always expose an `IEnumerable<Memory<T>>`, it'd involve an unnecessary copy for simple enumeration use cases. 

Another option is to not touch the interface but add 2 new APIs that return a `IEnumerable<ReadOnlyMemory<T>>` and an `IEnumerable<Memory<T>>`. That way, the users of the API have options.

src/Microsoft.Data/StringColumn.cs
```diff
@@ -14,7 +14,7 @@ namespace Microsoft.Data
     /// A mutable column to hold strings
     /// </summary>
     /// <remarks> Is NOT Arrow compatible </remarks>
-    public partial class StringColumn : BaseColumn, IEnumerable<string>
+    public partial class StringColumn : BaseColumn, IEnumerable<List<string>>
```
pgovind: Same alternative here. We could leave the interface as it is and make a new API that returned a `List<string>`. We'd have to deal with keeping the `NullCount` in sync, but that is easily solved

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -19,7 +19,7 @@ namespace Microsoft.Data
     /// A column to hold primitive types such as int, float etc.
     /// </summary>
     /// <typeparam name="T"></typeparam>
-    public partial class PrimitiveColumn<T> : BaseColumn, IEnumerable<T?>
```
eerhardt: You need to still keep `IEnumerable<T?>`. This way things like XPlot charting still work by just passing it a Column, and it enumerates over the data in the column.

src/Microsoft.Data/PrimitiveColumnContainer.cs
eerhardt: Honestly, I think a new API (or two) is called for here. Something like `GetBuffers()` and `GetReadOnlyBuffers()`?

---
In reply to: [321517883](https://github.com/dotnet/corefxlab/pull/2718#discussion_r321517883) [](ancestors = 321517883)

PR #2725, opened 2019-09-17
Enable Strong Naming in DataFrame library

eerhardt:
This is necessary so other strong named libraries can consume the DataFrame library.


ahsonkhan:
Look's good. Did we add any other csprojs recently that don't import common.props or was this the only one? We should verify and add it to all of them.

eerhardt:
IMO - We should use Directory.Build.props instead. But I didn’t have time to do it here since I needed this fixed ASAP for a customer. I’ll see if I can get something for tomorrow.

PR #2726, opened 2019-09-18
Use Directory.Build.props instead of tools\common.props.

eerhardt:
This ensures new projects get the global settings by default.

### DIFF
--- a/tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
+++ b/tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
@@ -1020,9 +1020,9 @@ namespace Microsoft.Data.Tests
         [Fact]
         public void TestGoupByDifferentColumnTypes()
         {
-            void GroupCountAndAssert(DataFrame df)
+            void GroupCountAndAssert(DataFrame frame)
             {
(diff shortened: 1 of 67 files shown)


tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1020,9 +1020,9 @@ public void TestGroupBy()
         [Fact]
         public void TestGoupByDifferentColumnTypes()
         {
-            void GroupCountAndAssert(DataFrame df)
+            void GroupCountAndAssert(DataFrame frame)
```
ahsonkhan: Lol. Unrelated parameter renames got snuck into this PR :)

eerhardt: These changes were related. Before, this project didn’t have warnings as errors turned on. With directory build props, now it is on for all projects. These places were causing errors because the names conflicted with variables in the outer scope.

ahsonkhan: Ah! I didn't notice that. `string data` used twice within the same method in different scopes. Nice. Thanks for clarifying.

PR #2728, opened 2019-09-19
Inplace computations and binary ops

pgovind:
Making the API more consistent around whether it operates in place or not. Should be an easy review.

### DIFF
--- a/src/Microsoft.Data/GroupBy.cs
+++ b/src/Microsoft.Data/GroupBy.cs
@@ -17,13 +17,13 @@ namespace Microsoft.Data
         /// Compute the number of non-null values in each group 
         /// </summary>
         /// <returns></returns>
-        public abstract DataFrame Count();
+        public abstract DataFrame Count(params string[] columnNames);
 
         /// <summary>
         /// Return the first value in each group
         /// </summary>
         /// <returns></returns>
-        public abstract DataFrame First();
+        public abstract DataFrame First(params string[] columnNames);
 
         /// <summary>
         /// Returns the first numberOfRows rows of each group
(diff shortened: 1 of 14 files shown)


src/Microsoft.Data/PrimitiveColumn.BinaryOperations.tt
```diff
@@ -77,7 +101,11 @@ namespace Microsoft.Data
                     throw new NotSupportedException();
 <# } else { #>
                     PrimitiveColumn<<#=type.TypeName#>> <#=type.TypeName#>Column = this as PrimitiveColumn<<#=type.TypeName#>>;
-                    var new<#=type.TypeName#>Column = <#=type.TypeName#>Column.Clone();
+                    PrimitiveColumn<<#=type.TypeName#>> new<#=type.TypeName#>Column;
+                    if (inPlace)
```
gfoidl: My previous suggestion should be applied here (I know you know this...).

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -541,17 +544,29 @@ public void TestComputations()
 
             // Test the computation results
             df["Double"][0] = 100.0;
-            df["Double"].CumulativeMax();
+            BaseColumn doubleColumn = df["Double"].CumulativeMax();
+            Assert.Equal(100.0, doubleColumn[9]);
```
eerhardt: Should we be verifying more than just the 9th value?

src/Microsoft.Data/GroupBy.cs
```diff
@@ -17,13 +17,13 @@ public abstract class GroupBy
         /// Compute the number of non-null values in each group 
         /// </summary>
         /// <returns></returns>
-        public abstract DataFrame Count();
+        public abstract DataFrame Count(params string[] columnNames);
```
eerhardt: These changes seem unrelated. Do we have tests covering this (passing in columnNames)?

pgovind: Yup. The last parts of TestGroupBy

pgovind:
Thanks for the review folks!

Issue #2719, opened 2019-09-09
DataFrame and BaseColumn should have easier APIs to append new rows


eerhardt:
We should have some APIs that allow a developer to easily append new values to DataFrame and to BaseColumn.

BaseColumn today has `Resize(Length + 1)` and then `SetValue(Length, value)`, but this isn't intuitive to use.

DataFrame should have an InsertRow or AppendRow method that takes a collection of objects to insert.

cc @pgovind

pgovind:
[2730](https://github.com/dotnet/corefxlab/pull/2730) implements `AppendRow` on the DataFrame.

I'm not super convinced that we should have an `Append` on `BaseColumn`. This is mostly because `DataFrameTable` holds a list of `BaseColumn` and does book keeping of `RowCount`. `DataFrameTable` also enforces that each `BaseColumn` has the same `RowCount`. If `BaseColumn` could append data to itself, nothing will break at the moment but it may not behave the way folks expect. Consider the following for ex:
`C#
BaseColumn intColumn = df["IntColumn"];
intColumn.Append(5);
df.AppendRow({100, 2.0}); // appends 100 to intColumn and 2.0 to floatColumn
`
When someone asks for the row containing [100, 2.0], they might actually get [5, 2.0] now. An easy way to fix this would be to append nulls to the remaining columns whenever a `BaseColumn.Append` is called. That's exactly what `DataFrame.AppendRow` does, but the intent is more obvious with `AppendRow`. So, do we still want an `Append` on `BaseColumn`?

Note: What I described above can happen already. `PrimitiveColumn` and `StringColumn` have an `Append` on them. They exist so a user can initially populate a column with data or because they are being used without a DataFrame. Starting from a DataFrame though, a user would have to explicitly cast a `BaseColumn` to one of those to call `Append`. At that point, maybe we can assume that users won't be surprised by this behavior? Thoughts?

eerhardt:
> Note: What I described above can happen already. PrimitiveColumn and StringColumn have an Append on them.
...
 a user would have to explicitly cast a BaseColumn to one of those to call Append. At that point, maybe we can assume that users won't be surprised by this behavior? 

I don't understand why casting from a `BaseColumn` down to `PrimitiveColumn` would change the expectation on behavior.

> Another Note: Pandas avoid this by not appending in place. All appends to a column/Series return a new column/Series.

I think that behavior may make sense. As long as there is an `AppendRange` type of method so I can append many values without creating intermediate column/Series objects.

PR #2730, opened 2019-09-23
AppendRow and fix NullCount bugs

pgovind:
Implements one part of [2719](https://github.com/dotnet/corefxlab/issues/2719). I've added some discussions about adding an `AppendRow` API to BaseColumn. 
This patch also fixes some bugs around `NullCount`.  Because C# Arrow is essentially immutable, it doesn't have to deal with keeping `NullCount` in sync on mutation. Arrow can pass in empty data and/or null buffers. DataFrame prefers to always populate both the data and null buffers, so this patch ensures that when empty data/null buffers are encounterd, they are initialized with the right default values. 
Note: This is still zero copy. We only allocate and initialize when empty buffers are present.

### DIFF
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -79,7 +79,7 @@ namespace Microsoft.Data
         /// Called internally from Merge and GroupBy. Resizes the column to the specified length to allow setting values by indexing
         /// </summary>
         /// <param name="length"></param>
-        public virtual void Resize(long length) => throw new NotImplementedException();
+        protected internal virtual void Resize(long length) => throw new NotImplementedException();
 
         /// <summary>
         /// Clone column to produce a copy potentially changing the order by supplying mapIndices and an invert flag
@@ -164,6 +164,23 @@ namespace Microsoft.Data
         public virtual double Median() => throw new NotImplementedException();
 
         /// <summary>
+        /// Returns a DataFrame with a concise summary of this column
+        /// </summary>
+        public virtual DataFrame Info()
+        {
+            DataFrame ret = new DataFrame();
+            StringColumn infoColumn = new StringColumn("Info", 0);
+            infoColumn.Append("DataType");
+            infoColumn.Append("Length");
+            StringColumn dataColumn = new StringColumn(Name, 0);
+            dataColumn.Append(DataType.ToString());
+            dataColumn.Append((Length - NullCount).ToString() + " non-null values");
+            ret.InsertColumn(0, infoColumn);
+            ret.InsertColumn(1, dataColumn);
+            return ret;
+        }
+
+        /// <summary>
         /// Used to exclude columns from the Description method
         /// </summary>
         public virtual bool HasDescription() => false;
@@ -172,6 +189,10 @@ namespace Microsoft.Data
         /// Returns a DataFrame with statistics that describe the column
         /// </summary>
         public virtual DataFrame Description() => throw new NotImplementedException();
+
+        public override bool Equals(object obj) => ReferenceEquals(this, obj);
+
+        public override int GetHashCode() => base.GetHashCode();
 
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
--- a/src/Microsoft.Data/BaseColumn.cs
+++ b/src/Microsoft.Data/BaseColumn.cs
@@ -166,18 +166,12 @@ namespace Microsoft.Data
         /// <summary>
         /// Returns a DataFrame with a concise summary of this column
         /// </summary>
-        public virtual DataFrame Info()
+        public virtual void Info(DataFrame dataFrame)
         {
-            DataFrame ret = new DataFrame();
-            StringColumn infoColumn = new StringColumn("Info", 0);
-            infoColumn.Append("DataType");
-            infoColumn.Append("Length");
-            StringColumn dataColumn = new StringColumn(Name, 0);
-            dataColumn.Append(DataType.ToString());
-            dataColumn.Append((Length - NullCount).ToString() + " non-null values");
-            ret.InsertColumn(0, infoColumn);
-            ret.InsertColumn(1, dataColumn);
-            return ret;
+            StringColumn dataColumn = new StringColumn(Name, 2);
+            dataColumn[0] = DataType.ToString();
+            dataColumn[1] = (Length - NullCount).ToString() + " non-null values";
+            dataFrame.InsertColumn(dataFrame.ColumnCount, dataColumn);
         }
 
         /// <summary>
@@ -189,10 +183,6 @@ namespace Microsoft.Data
         /// Returns a DataFrame with statistics that describe the column
         /// </summary>
         public virtual DataFrame Description() => throw new NotImplementedException();
-
-        public override bool Equals(object obj) => ReferenceEquals(this, obj);
-
-        public override int GetHashCode() => base.GetHashCode();
 
         internal virtual BaseColumn GetAscendingSortIndices() => throw new NotImplementedException();
 
--- a/src/Microsoft.Data/DataFrame.IO.cs
+++ b/src/Microsoft.Data/DataFrame.IO.cs
@@ -239,7 +239,7 @@ namespace Microsoft.Data
                             column[rowIndex] = null;
                             continue;
                         }
-                        throw new ArgumentException(Strings.MismatchedValueType + typeof(bool), nameof(val));
+                        throw new ArgumentException(string.Format(Strings.MismatchedValueType, typeof(bool)), nameof(val));
                     }
                 }
                 else if (dType == typeof(float))
@@ -257,7 +257,7 @@ namespace Microsoft.Data
                             column[rowIndex] = null;
                             continue;
                         }
-                        throw new ArgumentException(Strings.MismatchedValueType + typeof(float), nameof(val));
+                        throw new ArgumentException(string.Format(Strings.MismatchedValueType, typeof(float)), nameof(val));
                     }
                 }
                 else if (dType == typeof(string))
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -207,6 +207,19 @@ namespace Microsoft.Data
             }
         }
 
+        public IList<string> ColumnTypes
+        {
+            get
+            {
+                var ret = new List<string>(ColumnCount);
+                for (int i = 0; i < ColumnCount; i++)
+                {
+                    ret.Add(_table.Column(i).DataType.ToString());
+                }
+                return ret;
+            }
+        }
+
         public BaseColumn Column(int index) => _table.Column(index);
 
         public void InsertColumn(int columnIndex, BaseColumn column)
@@ -327,6 +340,41 @@ namespace Microsoft.Data
         public void SetColumnName(BaseColumn column, string newName) => _table.SetColumnName(column, newName);
 
         /// <summary>
+        /// Generates a concise summary of each column
+        /// </summary>
+        public DataFrame Info()
+        {
+            DataFrame ret = new DataFrame();
+            if (ColumnCount == 0)
+                return ret;
+
+            for (int i = 0; i < ColumnCount; i++)
+            {
+                BaseColumn column = Column(i);
+                DataFrame columnInfo = column.Info();
+                if (i == 0)
+                    ret = columnInfo;
+                else
+                    ret = ret.Merge<string>(columnInfo, "Info", "Info", "_left", "_right", joinAlgorithm: JoinAlgorithm.Inner);
+                RemoveRightAndRenameLeftColumn(ret, "Info", "_left", "_right");
+            }
+            return ret;
+        }
+
+        private void RemoveRightAndRenameLeftColumn(DataFrame dataFrame, string columnName, string left, string right)
+        {
+            string leftColumnName = columnName + left;
+            string rightColumnName = columnName + right;
+            int leftMergeColumn = dataFrame._table.GetColumnIndex(leftColumnName);
+            int rightMergeColumn = dataFrame._table.GetColumnIndex(rightColumnName);
+            if (leftMergeColumn != -1 && rightMergeColumn != -1)
+            {
+                dataFrame.RemoveColumn(rightColumnName);
+                dataFrame._table.SetColumnName(dataFrame[leftColumnName], columnName);
+            }
+        }
+
+        /// <summary>
         /// Generates descriptive statistics that summarize each numeric column
         /// </summary>
         public DataFrame Description()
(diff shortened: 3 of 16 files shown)


src/Microsoft.Data/BaseColumn.cs
```diff
@@ -173,6 +190,10 @@ public void SetName(string newName, DataFrame dataFrame = null)
         /// </summary>
         public virtual DataFrame Description() => throw new NotImplementedException();
 
+        public override bool Equals(object obj) => ReferenceEquals(this, obj);
+
+        public override int GetHashCode() => base.GetHashCode();
```
gfoidl: Can be removed? Would point to `base.GetHashCode` anyway.

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -138,6 +138,76 @@ public void RemoveColumn(string columnName)
...
+            IEnumerator<BaseColumn> columnEnumerator = _columns.GetEnumerator();
+            bool columnMoveNext = columnEnumerator.MoveNext();
+            if (row != null)
+            {
+                IEnumerator<object> rowEnumerator = row.GetEnumerator();
+                bool rowMoveNext = rowEnumerator.MoveNext();
+                while (columnMoveNext && rowMoveNext)
+                {
+                    BaseColumn column = columnEnumerator.Current;
+                    object value = rowEnumerator.Current ?? null;
```
gfoidl:
```suggestion
                    object value = rowEnumerator.Current;
```
I think the `?? null` is not necessary.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -81,8 +81,43 @@ public PrimitiveColumnContainer(IEnumerable<T> values)
...
+            ReadOnlyDataFrameBuffer<T> dataBuffer;
+            if (buffer.IsEmpty)
+            {
+                DataFrameBuffer<T> mutableBuffer = new DataFrameBuffer<T>();
+                mutableBuffer.EnsureCapacity(length);
+                mutableBuffer.RawSpan.Fill(default(T));
+                dataBuffer = mutableBuffer;
+            }
+            else
+                dataBuffer = new ReadOnlyDataFrameBuffer<T>(buffer, length);
```
```suggestion
            {
                dataBuffer = new ReadOnlyDataFrameBuffer<T>(buffer, length);
            }
```
Add braces.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -81,8 +81,43 @@ public PrimitiveColumnContainer(IEnumerable<T> values)
...
+            Buffers.Add(dataBuffer);
+            int bitMapBufferLength = (int)Math.Ceiling(length / 8.0);
```
```suggestion
            int bitMapBufferLength = (length + 7) / 8;
```
No need for floating points here. 

With `(uint)`s the JIT will emit a bitshift instead of the division -- cf. L217 where we applyed that trick.
But here this will be too much micro-optimization, so plain integer arithmetic will do a good job.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -106,11 +141,13 @@ public PrimitiveColumnContainer(long length = 0)
                 int allocatable = (int)Math.Min(length, ReadOnlyDataFrameBuffer<T>.MaxCapacity);
                 lastBuffer.EnsureCapacity(allocatable);
                 DataFrameBuffer<byte> lastNullBitMapBuffer = (DataFrameBuffer<byte>)(NullBitMapBuffers[NullBitMapBuffers.Count - 1]);
-                lastNullBitMapBuffer.EnsureCapacity((int)Math.Ceiling(allocatable / 8.0));
+                int nullBufferAllocatable = (int)Math.Ceiling(allocatable / 8.0);
```
```suggestion
                int nullBufferAllocatable = (allocatable + 7) / 8;
```

src/Microsoft.Data/BaseColumn.cs
eerhardt: Why are these Equals and GetHashCode changes necessary for `AppendRow and to fix NullCount bugs`?

src/Microsoft.Data/BaseColumn.cs
```diff
@@ -163,6 +163,23 @@ public void SetName(string newName, DataFrame dataFrame = null)
         /// </summary>
         public virtual double Median() => throw new NotImplementedException();
 
+        /// <summary>
+        /// Returns a DataFrame with a concise summary of this column
+        /// </summary>
+        public virtual DataFrame Info()
+        {
+            DataFrame ret = new DataFrame();
+            StringColumn infoColumn = new StringColumn("Info", 0);
```
eerhardt: You know you are going to append 2 values to these columns, you should initialize them with `length: 2`.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -207,6 +207,19 @@ public IList<string> Columns
             }
         }
 
+        public IList<string> ColumnTypes
```
eerhardt: I don't think this is the right API to expose to people. We shouldn't be dealing with types as strings.

Also, we should have some class/struct that represents the Schema that contains columns with names and types, and potentially other information.

eerhardt: Also, returning a new List on every call to a property is an anti-pattern IMO.

---
In reply to: [327715603](https://github.com/dotnet/corefxlab/pull/2730#discussion_r327715603) [](ancestors = 327715603)

src/Microsoft.Data/DataFrame.cs
```diff
@@ -326,6 +339,41 @@ private DataFrame Clone(BaseColumn mapIndices = null, bool invertMapIndices = fa
 
         public void SetColumnName(BaseColumn column, string newName) => _table.SetColumnName(column, newName);
 
+        /// <summary>
+        /// Generates a concise summary of each column
+        /// </summary>
+        public DataFrame Info()
+        {
+            DataFrame ret = new DataFrame();
+            if (ColumnCount == 0)
```
eerhardt: Don't need this code. The `for` loop will do the same thing.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -326,6 +339,41 @@ private DataFrame Clone(BaseColumn mapIndices = null, bool invertMapIndices = fa
...
+                return ret;
+
+            for (int i = 0; i < ColumnCount; i++)
+            {
+                BaseColumn column = Column(i);
+                DataFrame columnInfo = column.Info();
+                if (i == 0)
+                    ret = columnInfo;
```
eerhardt: Why create a `ret = new DataFrame()` above, just to discard it here?

Also, this feels like a lot of allocations (a new DataFrame for every column) and then they get thrown away. Why not pass the DataFrame into `column.Info()` and have the column add to the current DataFrame?

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -239,7 +239,7 @@ private static void AppendRow(List<BaseColumn> columns, long rowIndex, string[]
                             column[rowIndex] = null;
                             continue;
                         }
-                        throw new ArgumentException(Strings.MismatchedValueType + typeof(bool), nameof(val));
+                        throw new ArgumentException(string.Format(Strings.MismatchedValueType, typeof(bool)), nameof(val));
```
eerhardt: :thumbsup:

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -138,6 +138,76 @@ public void RemoveColumn(string columnName)
...
+            long length = column.Length;
+            column.Resize(length + 1);
+            column[length] = value;
+        }
+
+        public void Append(IEnumerable<object> row)
+        {
+            IEnumerator<BaseColumn> columnEnumerator = _columns.GetEnumerator();
+            bool columnMoveNext = columnEnumerator.MoveNext();
+            if (row != null)
```
eerhardt: Do we want to throw if the user passed in `null`?

src/Microsoft.Data/DataFrameTable.cs
```diff
@@ -138,6 +138,76 @@ public void RemoveColumn(string columnName)
...
+            {
+                IEnumerator<object> rowEnumerator = row.GetEnumerator();
+                bool rowMoveNext = rowEnumerator.MoveNext();
+                while (columnMoveNext && rowMoveNext)
```
eerhardt: What happens when I give too many values and there aren't enough columns? Do we just silently ignore that?

src/Microsoft.Data/PrimitiveColumn.cs
```diff
@@ -479,14 +460,14 @@ public override BaseColumn Filter<U>(U lower, U upper)
...
-                throw new ArgumentException(Strings.MismatchedValueType + typeof(T).ToString(), nameof(U));
+                throw new ArgumentException(string.Format(Strings.MismatchedValueType, typeof(T)), typeof(U).ToString());
         }
 
         public override DataFrame Description()
         {
             DataFrame ret = new DataFrame();
             StringColumn stringColumn = new StringColumn("Description", 0);
-            stringColumn.Append("Length");
+            stringColumn.Append("Length(non null values)");
```
eerhardt: For the future - we should probably put these strings into a resource. End-users will see these strings.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -81,8 +81,43 @@ public PrimitiveColumnContainer(IEnumerable<T> values)
...
+            }
+            else
+                dataBuffer = new ReadOnlyDataFrameBuffer<T>(buffer, length);
+            Buffers.Add(dataBuffer);
+            int bitMapBufferLength = (int)Math.Ceiling(length / 8.0);
+            ReadOnlyDataFrameBuffer<byte> nullDataFrameBuffer;
+            if (nullBitMap.IsEmpty)
+            {
+                if (nullCount != 0)
+                    throw new ArgumentNullException(nameof(nullBitMap));
```
eerhardt: Probably need a better exception message here.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -81,8 +81,43 @@ public PrimitiveColumnContainer(IEnumerable<T> values)
...
+                if (!buffer.IsEmpty)
+                {
+                    // Create a new bitMap with all the bits set
+                    var bitMap = new byte[bitMapBufferLength];
+                    for (int i = 0; i < bitMapBufferLength - 1; i++)
```
eerhardt: `Span.Fill`?

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -81,8 +81,43 @@ public PrimitiveColumnContainer(IEnumerable<T> values)
...
+                        bitMap[i] = 255;
+                    for (int i = 0; i < length - (bitMapBufferLength - 1) * 8; i++)
+                    {
+                        bitMap[bitMapBufferLength - 1] = SetBit(bitMap[bitMapBufferLength - 1], i, true);
```
eerhardt: (minor) you could probably do this a bit fancier/faster. Since you know how many elements past a multiple of 8 you have, one thought is to take `2^(elementsLeft + 1) - 1`, which will give you a byte with that many bits set. Or another way is described here: [https://stackoverflow.com/a/12416681/3680432](https://stackoverflow.com/a/12416681/3680432).  

> The usual way is to take a 1, and shift it left n bits. That will give you something like: 00100000. Then subtract one from that, which will clear the bit that's set, and set all the less significant bits, so in this case we'd get: 00011111.

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -275,7 +321,7 @@ internal void SetValidityBit(long index, bool value)
             Debug.Assert(bitMapBuffer.Length >= bitMapBufferIndex);
             if (bitMapBuffer.Length == bitMapBufferIndex)
                 bitMapBuffer.Append(0);
-            SetValidityBit(bitMapBuffer.RawSpan, (int)index, value);
+            SetValidityBit(bitMapBuffer.Span, (int)index, value);
```
eerhardt: Is there a reason for this change?

src/Microsoft.Data/DataFrame.cs

src/Microsoft.Data/DataFrameTable.cs
pgovind: I don't think so. It seems like a convenient way to append a row of `null`s to a DataFrame?

src/Microsoft.Data/DataFrameTable.cs
pgovind: Yup, they get ignored.

src/Microsoft.Data/PrimitiveColumn.cs
pgovind: https://github.com/dotnet/corefxlab/issues/2736

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Ah, so obvious now that I've seen it. Nice!

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Yup. `Span` is the right(and more conservative) API to call. This was a bad fix(for a subtle bug). The `Span` property really returns `bitMapBuffer.Span.Slice(0, Length)`. Before this patch, when we encountered empty null bitmaps from Arrow, `Span` would throw. While debugging, I didn't realize that `Length = 0` was the bug. Instead I always looked at the `NullBitMapBuffer` and because they are always default initialized and the bitmaps encode 8 indices into 1 byte, I thought calling `Span.Slice(0, Length)` was the bug and called `RawSpan` instead. It worked because say `RawSpan[3]` really exists and doesn't throw, and because the `NullCount` was 0, the unit tests passed. That didn't address the `Length = 0` bug at all. With this patch, we always initialize the buffers and **set** the bits. So `Length` is now set to the length of the column and `Span` doesn't throw. The new test cases in `TestEmptyArrowColumns` explicitly test for this.

src/Microsoft.Data/DataFrameTable.cs
eerhardt: Seems like this behavior should be documented then, if it is intended as the way to append a row of `null`s to the DataFrame.

src/Microsoft.Data/DataFrameTable.cs
eerhardt: Is that the best approach? It seems like a mistake by the user, and they would want to know about it.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -350,13 +385,7 @@ public DataFrame Description()
...
                 DataFrame columnDescription = column.Description();
                 ret = ret.Merge<string>(columnDescription, "Description", "Description", "_left", "_right", JoinAlgorithm.Inner);
-                int leftMergeColumn = ret._table.GetColumnIndex("Description" + "_left");
-                int rightMergeColumn = ret._table.GetColumnIndex("Description" + "_right");
-                if (leftMergeColumn != -1 && rightMergeColumn != -1)
-                {
-                    ret.RemoveColumn("Description" + "_right");
-                    ret._table.SetColumnName(ret["Description_left"], "Description");
-                }
+                RemoveRightAndRenameLeftColumn(ret, "Description", "_left", "_right");
```
eerhardt: Does this need to be changed anymore?
Also, could we follow the same pattern here as `Info`, where the DataFrame is passed into the `Description()` method? Then we wouldn't have to merge and unmangle the resultant DataFrames.

pgovind:
Thanks for the initial review folks. I split out the NullCount changes(and relevant feedback) into https://github.com/dotnet/corefxlab/pull/2770. I'll put up the `AppendRow` changes after 2770 goes in.

Issue #2727, opened 2019-09-19
DataFrame Description() returns the columns in the reversed order


eerhardt:
Notice the columns are reversed in these two cells:

![image](https://user-images.githubusercontent.com/8291187/65284712-ff860680-daff-11e9-812c-bd4e546be417.png)
/cc @pgovind

pgovind:
Fixed with #2731

PR #2731, opened 2019-09-24
Description: Column ordering fix

pgovind:
Fix for https://github.com/dotnet/corefxlab/issues/2727


tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1397,6 +1397,7 @@ public void TestDescription()
             for (int i = 1; i < description.ColumnCount; i++)
```
eerhardt: Can we also verify the first column is what we expect?

PR #2743, opened 2019-10-08
Empty ArrowStringColumn->Arrow column

pgovind:
Fix a bug in `ArrowStringColumn.AsArrowArray`.
Required for the Spark work.

### DIFF
--- a/src/Microsoft.Data/ArrowStringColumn.cs
+++ b/src/Microsoft.Data/ArrowStringColumn.cs
@@ -312,14 +312,16 @@ namespace Microsoft.Data
 
         protected internal override Apache.Arrow.Array AsArrowArray(long startIndex, int numberOfRows)
         {
+            if (numberOfRows == 0)
+                return new StringArray.Builder().Build();
             int offsetsBufferIndex = GetBufferIndexContainingRowIndex(startIndex, out int indexInBuffer);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -312,14 +312,16 @@ private int GetNullCount(long startIndex, int numberOfRows)
 
         protected internal override Apache.Arrow.Array AsArrowArray(long startIndex, int numberOfRows)
         {
+            if (numberOfRows == 0)
```
pgovind: With this check, `GetBufferIndexContainingRowIndex` won't throw anymore.

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -312,14 +312,16 @@ private int GetNullCount(long startIndex, int numberOfRows)
...
+                return new StringArray.Builder().Build();
```
eerhardt: Instead of using `StringArray.Builder()`, which will allocate a bunch of new array, how about just:

```C#
return new StringArray(numberOfRows, ArrowBuffer.Empty , ArrowBuffer.Empty , ArrowBuffer.Empty);
```

which won't allocate a bunch of unnecessary memory.

Issue #2745, opened 2019-10-08
Flaky unit test on ALCProxy: OSX


pgovind:
The OSX - InnerLoop CI fails quite often. I'll update the issue with the failing unit test the next time I see it. 

@cablooshe

cablooshe:
I'm looking at the failing test [here](https://dev.azure.com/dnceng/public/_build/results?buildId=381572&view=logs&jobId=7b614801-ed38-51b7-97d8-b2f3e799aa24), seems to be something failing with the delegate to the server. I'll take a closer look later on why that's failing on OSX specifically.

cablooshe:
I'm not 100% sure exactly why this is happening, but the delegate between client and server sometimes doesn't seem to exist for the OSX release. I can PR a patch that uses the old Reflection.Invoke as a backup in case the delegate fails, which doesn't exactly solve the problem but should hopefully prevent more random failures.

PR #2748, opened 2019-10-14
Fix for flaky OSX tests from ALCProxy

cablooshe:
Fixes to some flaky tests causing failures on the OSX release build coming from ALCProxy.

See #2745 for details.

### DIFF
--- a/src/ALCProxy.Communication/ALCClient.cs
+++ b/src/ALCProxy.Communication/ALCClient.cs
@@ -150,13 +150,21 @@ namespace ALCProxy.Communication
         {
             if (method == null)
                 throw new ArgumentNullException();
-            if (_serverDelegate == null) //We've called the ALC unload, so the proxy has been cut off
+            if (_server == null) //We've called the ALC unload, so the proxy has been cut off
                 throw new InvalidOperationException("Error in ALCClient: Proxy has been unloaded, or communication server was never set up correctly");
             if (args == null)
                 args = new object[] { };
 
             SerializeParameters(args, out IList<object> streams, out IList<Type> argTypes);
-            object encryptedReturn = _serverDelegate( method, streams, argTypes );
+            object encryptedReturn;
+            try
+            {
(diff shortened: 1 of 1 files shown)


src/ALCProxy.Communication/ALCClient.cs
```diff
@@ -150,13 +150,21 @@ public object SendMethod(MethodInfo method, object[] args)
...
                 args = new object[] { };
 
             SerializeParameters(args, out IList<object> streams, out IList<Type> argTypes);
-            object encryptedReturn = _serverDelegate( method, streams, argTypes );
+            object encryptedReturn;
+            try
+            {
+                encryptedReturn = _serverDelegate(method, streams, argTypes);
+            }
+            catch (NullReferenceException) //Some minor failures with OSX release builds using the serverDelegate, if it fails, the backup is to call the method using Invoke directly
```
ahsonkhan: Do we generally catch exceptions like `NullReferenceException`? Doesn't that mean there is a null ref bug that needs to be fixed?

ahsonkhan:
cc @sbomer @AaronRobinsonMSFT @jkoritzinsky

src/ALCProxy.Communication/ALCClient.cs
ahsonkhan: Ok, reading https://github.com/dotnet/corefxlab/issues/2745#issuecomment-541801971, sure.

> I'm not 100% sure exactly why this is happening, but the delegate between client and server sometimes doesn't seem to exist for the OSX release. I can PR a patch that uses the old Reflection.Invoke as a backup in case the delegate fails, which doesn't exactly solve the problem but should hopefully prevent more random failures.

ahsonkhan:
Re-running CI again just to spot check it passes consistently.

src/ALCProxy.Communication/ALCClient.cs
danmosemsft: Might this represent a product bug (that needs an issue open0

Issue #2724, opened 2019-09-17
Fix warnings in Microsoft.Data.DataFrame


eerhardt:
Today we have to turn off warnings as errors, because we have the following warnings in the DataFrame library

```
BaseColumn.BinaryOperations.cs(13,35): error CS0660: 'BaseColumn' defines operator == or operator != but does not override Object.Equals(object o) [/home/eerhardt/git/corefxlab/src/Microsoft.Data/Microsoft.Data.DataFrame.csproj]
BaseColumn.BinaryOperations.cs(13,35): error CS0661: 'BaseColumn' defines operator == or operator != but does not override Object.GetHashCode() [/home/eerhardt/git/corefxlab/src/Microsoft.Data/Microsoft.Data.DataFrame.csproj]
DataFrame.BinaryOperations.cs(13,26): error CS0660: 'DataFrame' defines operator == or operator != but does not override Object.Equals(object o) [/home/eerhardt/git/corefxlab/src/Microsoft.Data/Microsoft.Data.DataFrame.csproj]
DataFrame.BinaryOperations.cs(13,26): error CS0661: 'DataFrame' defines operator == or operator != but does not override Object.GetHashCode() [/home/eerhardt/git/corefxlab/src/Microsoft.Data/Microsoft.Data.DataFrame.csproj]
```

We should fix these warnings and turn back on warnings as errors.

/cc @pgovind

PR #2750, opened 2019-10-15
Remove comparison operators on DataFrame and *Column

pgovind:
Also modifies `And`, `Or`, and `Xor` to take only bool params
Also fixes https://github.com/dotnet/corefxlab/issues/2724 as a result


src/Microsoft.Data/BaseColumn.BinaryOperators.cs
```diff
@@ -371,67 +311,6 @@ public abstract partial class BaseColumn
...
-        {
-            return column.And(value);
-        }
-
-        public static BaseColumn operator &(BaseColumn column, ushort value)
-        {
-            return column.And(value);
-        }
-
         public static BaseColumn operator |(BaseColumn left, BaseColumn right)
```
pgovind: Maybe this should change to take `PrimitiveColumn<bool>` as its parameters and also return a `PrimitiveColumn<bool>`?

src/Microsoft.Data/BaseColumn.BinaryOperators.tt
```diff
@@ -20,34 +20,27 @@ namespace Microsoft.Data
...
         {
             return column.<#=method.MethodName#>(value);
         }
-
-<# } #>
-<# } else if (method.MethodType == MethodType.ComparisonScalar) { #>
+<# } else { #>
 <# foreach (TypeConfiguration type in typeConfiguration) { #>
-        public static PrimitiveColumn<bool> operator <#=method.Operator#>(BaseColumn column, <#=type.TypeName#> value)
+        public static BaseColumn operator <#=method.Operator#>(BaseColumn column, <#=type.TypeName#> value)
```
pgovind: The diff algorithm is just doing a bad job here. They are independent changes

src/Microsoft.Data/DataFrame.BinaryOperations.cs
```diff
@@ -332,7 +326,7 @@ public DataFrame RightShift(int value, bool inPlace = false)
             }
             return retDataFrame;
         }
-        public DataFrame Equals<T>(IReadOnlyList<T> values)
+        public DataFrame PairwiseEquals<T>(IReadOnlyList<T> values)
```
pgovind: Does `Elementwise` sound better than `Pairwise`?

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -4,8 +4,6 @@
     <TargetFramework>netstandard2.0</TargetFramework>
     <LangVersion>7.3</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
-    <!-- TODO: Remove this https://github.com/dotnet/corefxlab/issues/2724 -->
```
pgovind: Now that the `==` operators are being removed, we can delete this

eerhardt: Let's close that issue when this gets checked in then. Now that warnings as errors are turned on, we shouldn't add any more warnings to this library.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -1624,7 +1572,7 @@ public void TestClone()
         public void TestColumnCreationFromExisitingColumn()
         {
             DataFrame df = MakeDataFrameWithAllColumnTypes(10);
-            PrimitiveColumn<bool> bigInts = new PrimitiveColumn<bool>("BigInts", df["Int"] > 5);
+            PrimitiveColumn<bool> bigInts = new PrimitiveColumn<bool>("BigInts", df["Int"].PairwiseGreaterThan(5));
```
eerhardt: Do you have any thought about `Elementwise` vs. `Pairwise`?

For some reason, I think I lean towards `Elementwise`, because when we do operators (like add, multiply) those will be named ElementwiseMultiply.

eerhardt: In Pandas docs, they say element-wise:

[https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.lt.html](https://pandas.pydata.org/pandas-docs/stable/reference/api/pandas.DataFrame.lt.html)

---
In reply to: [335140660](https://github.com/dotnet/corefxlab/pull/2750#discussion_r335140660) [](ancestors = 335140660)

src/Microsoft.Data/BaseColumn.BinaryOperations.cs
```diff
@@ -72,8 +72,7 @@ public virtual BaseColumn And(BaseColumn column, bool inPlace = false)
             throw new NotImplementedException();
         }
 
-        public virtual BaseColumn And<T>(T value, bool inPlace = false)
-            where T : unmanaged
+        public virtual BaseColumn And(bool value, bool inPlace = false)
```
eerhardt: Can I only get back a `PrimitiveColumn<bool>` from this method?

pgovind: Yes, it only returns a `PrimitiveColumn<bool>`. You are thinking about marking the return type PrimitiveColumn<bool> correct?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
pgovind: I had the same question. Change it to `Elementwise` now

src/Microsoft.Data/BaseColumn.BinaryOperations.cs
eerhardt: I think that makes sense. Especially since we are changing the input type to `bool`.

src/Microsoft.Data/BaseColumn.BinaryOperators.cs
pgovind: Unfortunately no, because we can't specialize this template. If I changed all the `BaseColumn` here to `PrimitiveColumn<bool>`, the compiler will complain that the overload should contain at least one defining type. Also, `PrimitiveColumn<T>` is a different type than `PrimitiveColumn<bool>`, so I can't overload this in `PrimitiveColumn.cs` either. I don't think I can define an extension method that accepts a `PrimitiveColumn<bool>` either, but I'm not certain yet.

src/Microsoft.Data/ColumnArithmeticTemplate.ttinclude
```diff
@@ -114,18 +114,18 @@
...
-        new MethodConfiguration("NotEquals", MethodType.ComparisonScalar, "!="),
-        new MethodConfiguration("GreaterThanOrEqual", MethodType.Comparison, ">=", isNumeric:true),
-        new MethodConfiguration("GreaterThanOrEqual", MethodType.ComparisonScalar, ">=", isNumeric:true),
-        new MethodConfiguration("LessThanOrEqual", MethodType.Comparison, "<=", isNumeric:true),
-        new MethodConfiguration("LessThanOrEqual", MethodType.ComparisonScalar, "<=", isNumeric:true),
-        new MethodConfiguration("GreaterThan", MethodType.Comparison, ">", isNumeric:true),
-        new MethodConfiguration("GreaterThan", MethodType.ComparisonScalar, ">", isNumeric:true),
-        new MethodConfiguration("LessThan", MethodType.Comparison, "<", isNumeric:true),
-        new MethodConfiguration("LessThan", MethodType.ComparisonScalar, "<", isNumeric:true),
+        new MethodConfiguration("ElementwiseEquals", MethodType.Comparison, "=="),
```
eerhardt: (minor) do we still need the operator here now that we are removing it?

src/Microsoft.Data/DataFrame.BinaryOperations.cs
eerhardt: I think so.... but honestly, I'm not super excited about either. `Elementwise` is probably best for now.

---
In reply to: [335084385](https://github.com/dotnet/corefxlab/pull/2750#discussion_r335084385) [](ancestors = 335084385)

src/Microsoft.Data/StringColumn.BinaryOperations.cs
```diff
@@ -42,9 +42,8 @@ public override BaseColumn Add<T>(T value, bool inPlace = false)
             return ret;
         }
 
-        public override PrimitiveColumn<bool> Equals(BaseColumn column)
+        public override PrimitiveColumn<bool> ElementwiseEquals(BaseColumn column)
```
eerhardt: Does the ArrowStringColumn not have this?

src/Microsoft.Data/ColumnArithmeticTemplate.ttinclude
pgovind: Removed

src/Microsoft.Data/StringColumn.BinaryOperations.cs
pgovind: Not yet. Filed https://github.com/dotnet/corefxlab/issues/2754

PR #2752, opened 2019-10-15
Operator overloads

pgovind:
For https://github.com/dotnet/spark/pull/277


src/Microsoft.Data/BaseColumn.BinaryOperators.cs
```diff
@@ -92,66 +157,131 @@ public abstract partial class BaseColumn
...
+        }
+
         public static BaseColumn operator -(BaseColumn column, byte value)
         {
             return column.Subtract(value);
         }
 
+        public static BaseColumn operator -(byte value, BaseColumn column)
+        {
+            return column.Subtract(value);
```
pgovind: This is wrong.

PR #2753, opened 2019-10-16
More operator overloads

pgovind:
Enable expressions such as `1 + df`
For dotnet/spark#277

Should be a straightforward review

### DIFF
--- a/src/Microsoft.Data/BaseColumn.BinaryOperations.cs
+++ b/src/Microsoft.Data/BaseColumn.BinaryOperations.cs
@@ -23,12 +23,24 @@ namespace Microsoft.Data
             throw new NotImplementedException();
         }
 
+        internal virtual BaseColumn Add<T>(T value, bool inPlace = false, bool reverseOrderOfOperations = false)
+            where T : unmanaged
+        {
+            throw new NotImplementedException();
+        }
+
         public virtual BaseColumn Subtract(BaseColumn column, bool inPlace = false)
         {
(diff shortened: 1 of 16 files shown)


src/Microsoft.Data/BaseColumn.BinaryOperations.cs
```diff
@@ -23,6 +23,12 @@ public virtual BaseColumn Add<T>(T value, bool inPlace = false)
             throw new NotImplementedException();
         }
 
+        internal virtual BaseColumn Add<T>(T value, bool inPlace = false, bool reverseOrderOfOperations = false)
```
eerhardt: 2 concerns about this:

1. I don't really like the `reverseOrderOfOperations` parameter. It feels like this might be better served as 2 static functions: one that takes `Column, T` and another that takes `T, Column`.
2. Since this is internal, the only public way of calling this function is through the operator, which is against [https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/operator-overloads](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/operator-overloads)

> ✓ CONSIDER providing methods with friendly names that correspond to each overloaded operator.
Many languages do not support operator overloading. For this reason, it is recommended that types that overload operators include a secondary method with an appropriate domain-specific name that provides equivalent functionality.

pgovind:
Closing this for now. It might require a bigger change, and I'd rather do it after I address the first round of API review feedback

PR #2755, opened 2019-10-17
Rename BaseColumn to DataFrameColumn

pgovind:
Rename BaseColumn to DataFrameColumn
Rename PrimitiveColumn to PrimitiveDataFrameColumn

I'm not renaming `ArrowStringColumn` -> `ArrowStringDataFrameColumn` because that's just too big IMO.

This patch is pure renaming, contains no new code. Should be an easy stamp.


src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -18,7 +18,7 @@ namespace Microsoft.Data
     /// <summary>
     /// An immutable to hold Arrow style strings
     /// </summary>
-    public partial class ArrowStringColumn : BaseColumn, IEnumerable<string>
+    public partial class ArrowStringColumn : DataFrameColumn, IEnumerable<string>
```
eerhardt: All other classes now have the `DataFrameColumn` suffix, but this one doesn't. Is there a reason for that?

src/Microsoft.Data/ArrowStringColumn.cs
```diff
@@ -446,7 +446,7 @@ public override GroupBy GroupBy(int columnIndex, DataFrame parent)
             }
         }
 
-        public override BaseColumn FillNulls(object value, bool inPlace = false) => throw new NotSupportedException();
+        public override DataFrameColumn FillNulls(object value, bool inPlace = false) => throw new NotSupportedException();
```
eerhardt: (not related to this PR) But ArrowStringColumn should be able to `FillNulls` when `inPlace` is `false`, right? Maybe we should log an issue for that.

eerhardt:
>                     return OrImplementation(column as PrimitiveDataFrameColumn<bool>, inPlace);

This is an existing issue, so you don't need to address it in this PR, but you don't need this extra cast here.

```C#
switch (column)
{
    case PrimitiveDataFrameColumn<bool> boolColumn:
        return OrImplementation(boolColumn, inPlace);
```

Will work correctly without the double type check.


---
Refers to: src/Microsoft.Data/PrimitiveDataFrameColumn.BinaryOperations.cs:249 in 8703944. [](commit_id = 870394406ed3d97a003ce7975d5e3230411a0842, deletion_comment = False)

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Filed an issue for this

src/Microsoft.Data/ArrowStringColumn.cs
pgovind: Renamed to `ArrowStringDataFrameColumn` now

PR #2757, opened 2019-10-18
Refactor the Arrow related APIs

pgovind:
This is the first cut. There could be other things we can do here that I'm missing.


src/Microsoft.Data/DataFrame.Arrow.cs
```diff
@@ -0,0 +1,165 @@
...
+                    default:
+                        throw new NotImplementedException(nameof(fieldType.Name));
+                }
+                ret.InsertColumn(ret.ColumnCount, dataFrameColumn);
+                fieldIndex++;
+            }
+            return ret;
+        }
+
+        public IEnumerable<RecordBatch> ToArrowRecordBatches()
```
pgovind: Not sure if I have a strong reason to make this a static method or an extension

src/Microsoft.Data/DataFrameColumn.cs
```diff
@@ -108,12 +108,12 @@ public void SetName(string newName, DataFrame dataFrame = null)
...
 
         // Arrow related APIs
-        protected internal virtual Field Field() => throw new NotImplementedException();
+        protected internal virtual Field GetArrowField() => throw new NotImplementedException();
         /// <summary>
         /// Returns the max number of values that are contiguous in memory
         /// </summary>
-        protected internal virtual int MaxRecordBatchLength(long startIndex) => 0;
-        protected internal virtual Apache.Arrow.Array AsArrowArray(long startIndex, int numberOfRows) => throw new NotImplementedException();
+        protected internal virtual int GetMaxArrowRecordBatchLength(long startIndex) => 0;
```
eerhardt: Does this really need to be prefixed with `Arrow`? It could be generally useful for other things, not?

src/Microsoft.Data/DataFrame.Arrow.cs
```diff
@@ -0,0 +1,165 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using Apache.Arrow;
+using Apache.Arrow.Types;
+
+namespace Microsoft.Data
+{
+    public partial class DataFrame
+    {
+        public static DataFrame FromArrowRecordBatch(RecordBatch recordBatch)
```
eerhardt: Can we add XML doc on these two public methods?

src/Microsoft.Data/DataFrame.Arrow.cs
eerhardt: I think it is fine like this.

src/Microsoft.Data/DataFrameColumn.cs
pgovind: Potentially. Nobody else is using it now, but you're right that it doesn't need the `Arrow` moniker.

PR #2758, opened 2019-10-21
.NET Core 3.1 Preview2 IntelliSense (English)

carlossanlop:
@ahsonkhan @safern please review.

Explanation:
- [Here is the actual PR ](https://github.com/dotnet/corefx/pull/41951) that bumps the IntelliSense in corefx 3.1 using the nupkg file uploaded to the blob (the PR that matters).
- This PR's purpose is to simply keep a change history for our IntelliSense files, so people can know what is included in the nupkg from the other PR.
- Eventually this branch will go away and we will automate the process. There won't be a need to do this.

CC @mairaw


carlossanlop:
@ahson should I have named this nupkg "Preview1", since we are already one version behind?

ahsonkhan:
> @ahson should I have named this nupkg "Preview1", since we are already one version behind?

Yes please. Also, you tagged the wrong person :p

standalone/Microsoft.Private.Intellisense/xmldocs/netcoreapp/1033/System.Text.Json.xml
```diff
@@ -543,7 +543,7 @@ The current <paramref name="reader" /> token does not start or represent a value
       <param name="utf8Text">The UTF-8 encoded text to compare against.</param>
       <returns>
         <see langword="true" /> if the string value of this element has the same UTF-8 encoding as
-    <paramref name="utf8Text" />; otherwise, <see langword="false" />.</returns>
+<paramref name="utf8Text" />; otherwise, <see langword="false" />.</returns>
```
ahsonkhan: This spacing change is weird. Where did it come from.

carlossanlop: Downloaded from Docs. I don't touch the files.

carlossanlop: The unexpected endline may have been ported like that with the docs porting tool some time ago, because some programmers tend to write their triple slash comments with lines that are no longer than 80 characters. It used to be a restriction in older languages, but not on C#, so I guess people still do it to reduce the need to scroll horizontally.

The docs porting tool used to port these messages untouched, but I fixed that bug some time ago: any unexpected endline that is not preceded by a symbol (.,?!*:) is removed automatically before porting.

So maybe this unexpected endline is a remnant of how the docs porting tool used to port strings before I fixed that.

I wouldn't worry. I noticed there are a lot of these, and some of these precede the docs porting tool.

PR #2759, opened 2019-10-21
Refactor Columns and add ColumnNames

pgovind:
Address feedback around the `Column` APIs from the first API review

### DIFF
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -45,18 +45,9 @@ namespace Microsoft.Data
 
         public int ColumnCount => _table.ColumnCount;
 
-        public IList<string> Columns
-        {
-            get
-            {
-                var ret = new List<string>(ColumnCount);
-                for (int i = 0; i < ColumnCount; i++)
-                {
-                    ret.Add(_table.Column(i).Name);
-                }
-                return ret;
-            }
-        }
+        public IReadOnlyList<DataFrameColumn> Columns => _table.Columns;
+
+        public IReadOnlyList<string> ColumnNames => _table.ColumnNames;
 
         public DataFrameColumn Column(int index) => _table.Column(index);
 
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -47,7 +47,7 @@ namespace Microsoft.Data
 
         public IReadOnlyList<DataFrameColumn> Columns => _table.Columns;
 
-        public IReadOnlyList<string> ColumnNames => _table.ColumnNames;
+        public IReadOnlyList<string> GetColumnNames() => _table.GetColumnNames();
 
         public DataFrameColumn Column(int index) => _table.Column(index);
 
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -43,7 +43,7 @@ namespace Microsoft.Data
 
         public long RowCount => _table.RowCount;
 
-        public int ColumnCount => _table.ColumnCount;
+        //public int Columns.Count => _table.Columns.Count;
 
         public IReadOnlyList<DataFrameColumn> Columns => _table.Columns;
 
@@ -123,7 +123,7 @@ namespace Microsoft.Data
                 newColumn.SetName(columnName);
                 if (columnIndex == -1)
                 {
-                    _table.InsertColumn(ColumnCount, newColumn);
+                    _table.InsertColumn(Columns.Count, newColumn);
                 }
                 else
                 {
@@ -156,8 +156,8 @@ namespace Microsoft.Data
 
         private DataFrame Clone(DataFrameColumn mapIndices = null, bool invertMapIndices = false)
         {
-            List<DataFrameColumn> newColumns = new List<DataFrameColumn>(ColumnCount);
-            for (int i = 0; i < ColumnCount; i++)
+            List<DataFrameColumn> newColumns = new List<DataFrameColumn>(Columns.Count);
+            for (int i = 0; i < Columns.Count; i++)
             {
                 newColumns.Add(Columns[i].Clone(mapIndices, invertMapIndices));
             }
@@ -172,7 +172,7 @@ namespace Microsoft.Data
         public DataFrame Description()
         {
             DataFrame ret = new DataFrame();
-            if (ColumnCount == 0)
+            if (Columns.Count == 0)
                 return ret;
             int i = 0;
             while (!Columns[i].HasDescription())
@@ -181,7 +181,7 @@ namespace Microsoft.Data
             }
             ret = Columns[i].Description();
             i++;
-            for (; i < ColumnCount; i++)
+            for (; i < Columns.Count; i++)
             {
                 DataFrameColumn column = Columns[i];
                 if (!column.HasDescription())
@@ -206,8 +206,8 @@ namespace Microsoft.Data
         {
             DataFrameColumn column = this[columnName];
             DataFrameColumn sortIndices = column.GetAscendingSortIndices();
-            List<DataFrameColumn> newColumns = new List<DataFrameColumn>(ColumnCount);
-            for (int i = 0; i < ColumnCount; i++)
+            List<DataFrameColumn> newColumns = new List<DataFrameColumn>(Columns.Count);
+            for (int i = 0; i < Columns.Count; i++)
             {
                 DataFrameColumn oldColumn = Columns[i];
                 DataFrameColumn newColumn = oldColumn.Clone(sortIndices, !ascending, oldColumn.NullCount);
@@ -227,7 +227,7 @@ namespace Microsoft.Data
         {
             DataFrame ret = inPlace ? this : Clone();
 
-            for (int i = 0; i < ret.ColumnCount; i++)
+            for (int i = 0; i < ret.Columns.Count; i++)
             {
                 DataFrameColumn column = ret.Columns[i];
                 if (column.IsNumericColumn())
@@ -242,7 +242,7 @@ namespace Microsoft.Data
         public DataFrame AddPrefix(string prefix, bool inPlace = false)
         {
             DataFrame df = inPlace ? this : Clone();
-            for (int i = 0; i < df.ColumnCount; i++)
+            for (int i = 0; i < df.Columns.Count; i++)
             {
                 DataFrameColumn column = df.Columns[i];
                 df._table.SetColumnName(column, prefix + column.Name);
@@ -257,7 +257,7 @@ namespace Microsoft.Data
         public DataFrame AddSuffix(string suffix, bool inPlace = false)
         {
             DataFrame df = inPlace ? this : Clone();
-            for (int i = 0; i < df.ColumnCount; i++)
+            for (int i = 0; i < df.Columns.Count; i++)
             {
                 DataFrameColumn column = df.Columns[i];
                 df._table.SetColumnName(column, column.Name + suffix);
@@ -297,7 +297,7 @@ namespace Microsoft.Data
         internal void SetTableRowCount(long rowCount)
         {
             // Even if current RowCount == rowCount, do the validation
-            for (int i = 0; i < ColumnCount; i++)
+            for (int i = 0; i < Columns.Count; i++)
             {
                 if (Columns[i].Length != rowCount)
                     throw new ArgumentException(String.Format("{0} {1}", Strings.MismatchedRowCount, Columns[i].Name));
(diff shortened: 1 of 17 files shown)


src/Microsoft.Data/DataFrame.cs
```diff
@@ -45,18 +45,9 @@ public DataFrame(IList<DataFrameColumn> columns)
...
-                for (int i = 0; i < ColumnCount; i++)
-                {
-                    ret.Add(_table.Column(i).Name);
-                }
-                return ret;
-            }
-        }
+        public IReadOnlyList<DataFrameColumn> Columns => _table.Columns;
+
+        public IReadOnlyList<string> GetColumnNames() => _table.GetColumnNames();
```
eerhardt: I thought we were going to leave this to the "Schema" property.

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
```diff
@@ -153,7 +153,7 @@
 
   <ItemGroup>
     <EmbeddedResource Update="Strings.resx">
-      <Generator>PublicResXFileCodeGenerator</Generator>
+      <Generator>ResXFileCodeGenerator</Generator>
```
eerhardt: What's the story here?

pgovind: Changes the `Strings.resx` modifier to private

src/Microsoft.Data/DataFrame.cs
```diff
@@ -45,18 +45,9 @@ public DataFrame(IList<DataFrameColumn> columns)
 
         public int ColumnCount => _table.ColumnCount;
```
eerhardt: Do we still need `ColumnCount` when we have `.Columns.Count`?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -45,18 +45,9 @@ public DataFrame(IList<DataFrameColumn> columns)
...
-                    ret.Add(_table.Column(i).Name);
-                }
-                return ret;
-            }
-        }
+        public IReadOnlyList<DataFrameColumn> Columns => _table.Columns;
+
+        public IReadOnlyList<string> GetColumnNames() => _table.GetColumnNames();
 
         public DataFrameColumn Column(int index) => _table.Column(index);
```
eerhardt: Do we still need this when we have `.Columns[i]`?

eerhardt:
>         public void InsertColumn(int columnIndex, DataFrameColumn column)

Should we continue to have these, or instead would it be more natural to make the `.Columns` property a mutable collection (of some sort)?

---
Refers to: src/Microsoft.Data/DataFrame.cs:54 in 23bc683. [](commit_id = 23bc6837fb745194ace20bcbee1ba310fee305b5, deletion_comment = False)

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
eerhardt: It didn't do that in this PR....

src/Microsoft.Data/DataFrame.cs
pgovind: Right, but without this method, there's be no way to examine the column names in the DataFrame. IDataView's `Schema` doesn't expose a way to get the column names(except by enumerating the `Schema's Columns`).

src/Microsoft.Data/DataFrame.cs
pgovind: Since `RowCount` exists, I felt that a user might also expect to find `ColumnCount`.

src/Microsoft.Data/DataFrame.cs
eerhardt: > IDataView's Schema doesn't expose a way to get the column names(except by enumerating the Schema's Columns).

That `except` is exactly how you get the column names. You inspect the Schema. Having 2 ways of getting the column names is duplication. Why do we need this API when we already have a way of getting the column names (and even more information)?

src/Microsoft.Data/DataFrame.cs
eerhardt: Well, to play devils advocate: We have a `Columns` property, but we don't have a `Rows` property.

src/Microsoft.Data/DataFrame.cs
pgovind: So, my reasoning was, in a notebook, if I wanted to see the column names, I'd have to write a foreach loop instead of 1 function call. I see now that executing `(IDataView)dataFrame.Schema` in a cell prints the column names and types. So, I'll pull this out.

src/Microsoft.Data/DataFrame.cs
pgovind: Alright, reasonably fair observation. Deleted `ColumnCount`

src/Microsoft.Data/Microsoft.Data.DataFrame.csproj
pgovind: Reverted this

pgovind:
> Should we continue to have these, or instead would it be more natural to make the .Columns property a mutable collection (of some sort)?

How do you feel about delaying this until we absolutely need it? We already have an `OnColumnsChanged` "event", but other than that, I don't see a strong reason to make a new type?

eerhardt:
> How do you feel about delaying this until we absolutely need it?

Ask yourself a different question. Why would we have an `IReadOnlyList` of `Columns`, but then have `InsertColumn`, `RemoveColumn`, etc. methods on the `DataFrame` class? What would be a more natural API? Modifying the collection itself, or asking the owner of the collection to modify it?

eerhardt:
> I don't see a strong reason to make a new type

It's fairly simple to make this type. See all the derived types from https://docs.microsoft.com/en-us/dotnet/api/system.collections.objectmodel.collection-1. Specifically the ones that are in corefx:

* System.Net.Mail
    - AttachmentCollection
    - LinkedResourceCollection
    - MailAddressCollection
* CngPropertyCollection
* SyndicationElementExtensionCollection

Doing so also follows our [Guidelines for Collections](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/guidelines-for-collections)

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -4,47 +4,54 @@
...
         private List<string> _columnNames = new List<string>();
 
         private Dictionary<string, int> _columnNameToIndexDictionary = new Dictionary<string, int>(StringComparer.Ordinal);
 
         public long RowCount { get; internal set; }
 
-        public int ColumnCount { get; private set; }
-
-        public DataFrameTable()
+        public DataFrameColumnCollection(Action columnsChanged) : base()
```
eerhardt: Do we want/need to have public constructors on this type? I kind of think it is only intended to be created from a `DataFrame`. So I don't think external callers should construct it.

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -4,47 +4,54 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 
 namespace Microsoft.Data
 {
     /// <summary>
     /// A DataFrameTable is just a container that holds a number of DataFrameColumns. It mainly acts as a convenient store to allow DataFrame to implement its algorithms
```
eerhardt: This summary needs to be updated.

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -4,47 +4,54 @@
...
+    public class DataFrameColumnCollection : Collection<DataFrameColumn>
     {
-        private IList<DataFrameColumn> _columns;
+        private Action ColumnsChanged;
 
         private List<string> _columnNames = new List<string>();
 
         private Dictionary<string, int> _columnNameToIndexDictionary = new Dictionary<string, int>(StringComparer.Ordinal);
 
         public long RowCount { get; internal set; }
```
eerhardt: `RowCount` on a `DataFrameColumnCollection` doesn't seem right to me.

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -4,47 +4,54 @@
...
+        {
+            var ret = new List<string>(Count);
+            for (int i = 0; i < Count; i++)
+            {
+                ret.Add(this[i].Name);
+            }
+            return ret;
+        }
 
         public IList<object> GetRow(long rowIndex)
```
eerhardt: Why would we have this public API when it is on `DataFrame`?

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -4,47 +4,54 @@
...
+            ColumnsChanged = columnsChanged;
             for (int i = 0; i < columns.Count; i++)
             {
-                InsertColumn(i, columns[i]);
+                Add(columns[i]);
             }
         }
 
-        public DataFrameColumn Column(int columnIndex) => _columns[columnIndex];
+        public IReadOnlyList<string> GetColumnNames()
```
eerhardt: Do we really need this public API? Someone can easily write this code to get the column Names themselves.

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -88,21 +88,22 @@ public void InsertColumn(int columnIndex, DataFrameColumn column)
...
                 _columnNameToIndexDictionary[_columnNames[i]]++;
             }
-            _columns.Insert(columnIndex, column);
-            ColumnCount++;
+            base.InsertItem(columnIndex, column);
+            ColumnsChanged?.Invoke();
         }
 
-        public void SetColumn(int columnIndex, DataFrameColumn column)
+        public void Set(int columnIndex, DataFrameColumn column)
```
eerhardt: Why have this public API when a caller can just use the indexer?

src/Microsoft.Data/DataFrameColumnCollection.cs
pgovind: Nope. Made it internal.

src/Microsoft.Data/DataFrameColumnCollection.cs
pgovind: `GroupBy` needs it. I made it internal though.

src/Microsoft.Data/DataFrameColumnCollection.cs
pgovind: Made it internal

src/Microsoft.Data/DataFrameColumnCollection.cs
eerhardt: Yep that sounds good. We can have internal helpers, but they shouldn't necessarily be public.

src/Microsoft.Data/DataFrameColumnCollection.cs
pgovind: Made it internal

src/Microsoft.Data/DataFrameColumnCollection.cs
pgovind: Removed.

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -4,47 +4,54 @@
 
 using System;
 using System.Collections.Generic;
+using System.Collections.ObjectModel;
 
 namespace Microsoft.Data
 {
     /// <summary>
-    /// A DataFrameTable is just a container that holds a number of DataFrameColumns. It mainly acts as a convenient store to allow DataFrame to implement its algorithms
+    /// A DataFrameColumnCollection is just a container that holds a number of DataFrameColumns. It mainly acts as a convenient store to allow DataFrame to implement its algorithms
```
eerhardt: > It mainly acts as a convenient store to allow DataFrame to implement its algorithms

Is this something users should know/care about? To users, this is a collection of DataFrameColumn instances.

src/Microsoft.Data/DataFrameColumnCollection.cs
```diff
@@ -88,21 +88,22 @@ public void InsertColumn(int columnIndex, DataFrameColumn column)
...
                 _columnNameToIndexDictionary[_columnNames[i]]++;
             }
-            _columns.Insert(columnIndex, column);
-            ColumnCount++;
+            base.InsertItem(columnIndex, column);
+            ColumnsChanged?.Invoke();
         }
 
-        public void SetColumn(int columnIndex, DataFrameColumn column)
+        internal void Set(int columnIndex, DataFrameColumn column)
```
eerhardt: Do we need this method at all? Can't any callers just use the indexer?

PR #2760, opened 2019-10-21
Minor changes to the API

pgovind:
Mostly cosmetic changes.

### DIFF
--- a/src/Microsoft.Data/DataFrame.IO.cs
+++ b/src/Microsoft.Data/DataFrame.IO.cs
@@ -98,14 +98,14 @@ namespace Microsoft.Data
         {
             using (Stream fileStream = new FileStream(filename, FileMode.Open))
             {
-                return LoadStream(fileStream,
+                return LoadCsv(fileStream,
                                   separator: separator, header: header, columnNames: columnNames, dataTypes: dataTypes, numberOfRowsToRead: numRows,
                                   guessRows: guessRows, addIndexColumn: addIndexColumn);
             }
         }
 
         /// <summary>
-        /// Reads a text file as a DataFrame.
+        /// Reads a seekable stream of CSV data into a DataFrame.
         /// Follows pandas API.
         /// </summary>
         /// <param name="csvStream">stream of CSV data to be read in</param>
@@ -117,7 +117,7 @@ namespace Microsoft.Data
         /// <param name="guessRows">number of rows used to guess types</param>
         /// <param name="addIndexColumn">add one column with the row index</param>
         /// <returns>DataFrame</returns>
-        public static DataFrame LoadStream(Stream csvStream,
+        public static DataFrame LoadCsv(Stream csvStream,
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
                                 long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false)
@@ -135,7 +135,7 @@ namespace Microsoft.Data
             List<DataFrameColumn> columns;
             long streamStart = csvStream.Position;
             // First pass: schema and number of rows.
-            using (var streamReader = new StreamReader(csvStream))
+            using (var streamReader = new StreamReader(csvStream, Encoding.UTF8, detectEncodingFromByteOrderMarks: true, bufferSize: 1024, leaveOpen: true))
             {
                 string line = streamReader.ReadLine();
                 while (line != null)
--- a/src/Microsoft.Data/DataFrame.IO.cs
+++ b/src/Microsoft.Data/DataFrame.IO.cs
@@ -90,13 +90,14 @@ namespace Microsoft.Data
         /// <param name="guessRows">number of rows used to guess types</param>
         /// <param name="addIndexColumn">add one column with the row index</param>
         /// <returns>DataFrame</returns>
-        public static DataFrame ReadCsv(string filename,
+        public static DataFrame LoadCsv(string filename,
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
                                 int numRows = -1, int guessRows = 10,
                                 bool addIndexColumn = false)
         {
-            return ReadStream(() => new StreamReader(filename),
+            Stream fileStream = new FileStream(filename, FileMode.Open);
+            return LoadStream(fileStream,
                               separator: separator, header: header, columnNames: columnNames, dataTypes: dataTypes, numberOfRowsToRead: numRows,
                               guessRows: guessRows, addIndexColumn: addIndexColumn);
         }
@@ -105,7 +106,7 @@ namespace Microsoft.Data
         /// Reads a text file as a DataFrame.
         /// Follows pandas API.
         /// </summary>
-        /// <param name="createStream">function which creates a stream</param>
+        /// <param name="csvStream">stream of CSV data to be read in</param>
         /// <param name="separator">column separator</param>
         /// <param name="header">has a header or not</param>
         /// <param name="columnNames">column names (can be empty)</param>
@@ -114,11 +115,14 @@ namespace Microsoft.Data
         /// <param name="guessRows">number of rows used to guess types</param>
         /// <param name="addIndexColumn">add one column with the row index</param>
         /// <returns>DataFrame</returns>
-        public static DataFrame ReadStream(Func<StreamReader> createStream,
+        public static DataFrame LoadStream(Stream csvStream,
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
                                 long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false)
         {
+            if (!csvStream.CanSeek)
+                throw new ArgumentException(Strings.NonSeekableStream, nameof(csvStream));
+
             var linesForGuessType = new List<string[]>();
             long rowline = 0;
             int numberOfColumns = 0;
@@ -126,8 +130,11 @@ namespace Microsoft.Data
             if (header == true && numberOfRowsToRead != -1)
                 numberOfRowsToRead++;
 
+            List<DataFrameColumn> columns;
+            long streamStart = csvStream.Position;
+            StreamReader streamReader = new StreamReader(csvStream);
             // First pass: schema and number of rows.
-            using (var st = createStream())
+            using (var st = streamReader)
             {
                 string line = st.ReadLine();
                 while (line != null)
@@ -154,40 +161,41 @@ namespace Microsoft.Data
                         break;
                     line = st.ReadLine();
                 }
-            }
 
-            if (linesForGuessType.Count == 0)
-                throw new FormatException(Strings.EmptyFile);
+                if (linesForGuessType.Count == 0)
+                    throw new FormatException(Strings.EmptyFile);
 
-            List<DataFrameColumn> columns = new List<DataFrameColumn>(numberOfColumns);
+                columns = new List<DataFrameColumn>(numberOfColumns);
 
-            // Guesses types and adds columns.
-            for (int i = 0; i < numberOfColumns; ++i)
-            {
-                Type kind = GuessKind(i, linesForGuessType);
-                if (kind == typeof(bool))
+                // Guesses types and adds columns.
+                for (int i = 0; i < numberOfColumns; ++i)
                 {
-                    DataFrameColumn boolColumn = new PrimitiveDataFrameColumn<bool>(columnNames == null ? "Column" + i.ToString() : columnNames[i], header == true ? rowline - 1 : rowline);
-                    columns.Add(boolColumn);
+                    Type kind = GuessKind(i, linesForGuessType);
+                    if (kind == typeof(bool))
+                    {
+                        DataFrameColumn boolColumn = new PrimitiveDataFrameColumn<bool>(columnNames == null ? "Column" + i.ToString() : columnNames[i], header == true ? rowline - 1 : rowline);
+                        columns.Add(boolColumn);
+                    }
+                    else if (kind == typeof(float))
+                    {
+                        DataFrameColumn floatColumn = new PrimitiveDataFrameColumn<float>(columnNames == null ? "Column" + i.ToString() : columnNames[i], header == true ? rowline - 1 : rowline);
+                        columns.Add(floatColumn);
+                    }
+                    else if (kind == typeof(string))
+                    {
+                        DataFrameColumn stringColumn = new StringDataFrameColumn(columnNames == null ? "Column" + i.ToString() : columnNames[i], header == true ? rowline - 1 : rowline);
+                        columns.Add(stringColumn);
+                    }
+                    else
+                        throw new NotSupportedException(nameof(kind));
                 }
-                else if (kind == typeof(float))
-                {
-                    DataFrameColumn floatColumn = new PrimitiveDataFrameColumn<float>(columnNames == null ? "Column" + i.ToString() : columnNames[i], header == true ? rowline - 1 : rowline);
-                    columns.Add(floatColumn);
-                }
-                else if (kind == typeof(string))
-                {
-                    DataFrameColumn stringColumn = new StringDataFrameColumn(columnNames == null ? "Column" + i.ToString() : columnNames[i], header == true ? rowline - 1 : rowline);
-                    columns.Add(stringColumn);
-                }
-                else
-                    throw new NotSupportedException(nameof(kind));
-            }
 
-            // Fills values.
-            using (StreamReader st = createStream())
-            {
-                string line = st.ReadLine();
+                line = null;
+                streamReader.DiscardBufferedData();
+                streamReader.BaseStream.Seek(streamStart, SeekOrigin.Begin);
+
+                // Fills values.
+                line = st.ReadLine();
                 rowline = 0;
                 while (line != null && (numberOfRowsToRead == -1 || rowline < numberOfRowsToRead))
                 {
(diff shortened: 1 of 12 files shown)


src/Microsoft.Data/DataFrameColumn.cs
```diff
@@ -87,9 +87,15 @@ public void SetName(string newName, DataFrame dataFrame = null)
         /// <param name="mapIndices"></param>
         /// <param name="invertMapIndices"></param>
         /// <returns></returns>
-        public virtual DataFrameColumn Clone(DataFrameColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
+        public virtual DataFrameColumn Clone(DataFrameColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => CloneImplementation(mapIndices, invertMapIndices, numberOfNullsToAppend);
 
-        public virtual DataFrameColumn Sort(bool ascending = true) => throw new NotImplementedException();
+        protected virtual DataFrameColumn CloneImplementation(DataFrameColumn mapIndices, bool invertMapIndices, long numberOfNullsToAppend) => throw new NotImplementedException();
```
pgovind: This is so `PrimitiveDataFrameColumn` can hide the base class' `Clone` method with `new`. Similar pattern for `FillNulls`, 'Clip` and `Filter`

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
```diff
@@ -170,13 +170,19 @@ protected internal override Apache.Arrow.Array ToArrowArray(long startIndex, int
...
             if (startIndex > Length)
             {
                 throw new ArgumentOutOfRangeException(nameof(startIndex));
             }
-            return _columnContainer[startIndex, length];
+            var ret = new List<object>(length);
+            long endIndex = Math.Min(Length, startIndex + length);
+            for (long i = startIndex; i < endIndex; i++)
+            {
+                ret.Add(this[i]);
```
eerhardt: I'm not certain this is an improvement. With this new change, you are going to box every value when adding it to this list. And then callers will need to unbox the values to get the underlying `T` values.

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
```diff
@@ -278,7 +284,7 @@ public override bool IsNumericColumn()
             return ret;
         }
 
-        public override DataFrameColumn FillNulls(object value, bool inPlace = false)
+        public new PrimitiveDataFrameColumn<T> FillNulls(object value, bool inPlace = false)
```
eerhardt: You could change this to:

```C#
public new PrimitiveDataFrameColumn<T> FillNulls(T value, bool inPlace = false)
```

And then you wouldn't need to change the type below.

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
```diff
@@ -477,7 +493,12 @@ public override DataFrameColumn Clip<U>(U lower, U upper, bool inPlace = false)
                 throw new ArgumentException(Strings.MismatchedValueType + typeof(T).ToString(), nameof(U));
         }
 
-        public override DataFrameColumn Filter<U>(U lower, U upper)
+        protected override DataFrameColumn ClipImplementation<U>(U lower, U upper, bool inPlace)
+        {
+            return Clip(lower, upper, inPlace);
+        }
+
+        public new PrimitiveDataFrameColumn<T> Filter<U>(U lower, U upper)
```
eerhardt: Could be 

```C#
public new PrimitiveDataFrameColumn<T> Filter(T lower, T upper)
```

src/Microsoft.Data/DataFrameColumn.cs
```diff
@@ -173,7 +185,7 @@ public void SetName(string newName, DataFrame dataFrame = null)
         /// </summary>
         public virtual DataFrame Description() => throw new NotImplementedException();
 
-        internal virtual DataFrameColumn GetAscendingSortIndices() => throw new NotImplementedException();
+        public virtual PrimitiveDataFrameColumn<long> GetAscendingSortIndices() => throw new NotImplementedException();
```
eerhardt: Does this need to be public? Or should it be `protected`?

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -114,7 +114,7 @@ private static Type MaxKind(Type a, Type b)
         /// <param name="guessRows">number of rows used to guess types</param>
         /// <param name="addIndexColumn">add one column with the row index</param>
         /// <returns>DataFrame</returns>
-        public static DataFrame ReadStream(Func<StreamReader> createStream,
+        public static DataFrame LoadStream(Func<StreamReader> createStream,
```
eerhardt: Does this need to be `public`? I thought this was just an implementation detail of `LoadCsv`

src/Microsoft.Data/DataFrame.cs
```diff
@@ -143,24 +144,19 @@ public void RemoveColumn(string columnName)
             }
         }
 
-        public IList<IList<object>> Head(int numberOfRows)
+        public DataFrame Head(int numberOfRows)
```
eerhardt: We should have XML docs for public APIs.

src/Microsoft.Data/DataFrameColumn.cs
pgovind: I initially thought it would be a hook to override in the `sort` method. It doesn't offer much value though. I'm ok with it being `protected` .

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
pgovind: It's definitely not an improvement. I was forced to do this because I can't return an `IList<T?>`(returned by `PrimitiveColumnContainer`) as an `IList<object>` directly. However, I can instead return this as `return _columnContainer[startIndex, length] as dynamic` which is returning `IList<T?> -> IList<object>`. That would result in no boxing when the list is being returned. Know of a better/less ugly way?

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
pgovind: But that is not equivalent to what the code is doing now right? `Convert.ChangeType` will convert `value` to `T`. Let's say I have a `PrimitiveDataFrameColumn<double> doubleCol`. If I called `doubleCol.FillNulls(0.2f)`, `Convert.ChangeType` will convert the float to a double.

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
pgovind: Same explanation as the `FillNulls` comment. If `U` can be converted to `T`, I'd like to let that through.

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
eerhardt: `float` has an implicit conversion to `double. So the C# compiler will do the conversion.

It is not equivalent, but it is better. That's why we have strongly typed APIs. If you want the weakly typed `FillNulls` call the base class's `FillNulls` that takes an `object`.

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
eerhardt: User's can cast, or they can use implicit conversions, if they exist. Our strongly-typed APIs shouldn't be in the business of doing type conversion.

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
pgovind: Returning a IReadOnlyList<T?> may work here. https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/concepts/covariance-contravariance/variance-in-generic-interfaces

pgovind: Sigh, returning `IReadOnlyList<T?>` doesn't work, because variance in generic interfaces is restricted to reference types and `T` here is unmanaged constrained

eerhardt: Maybe, let's take a step back. What is the value of this API? How do we expect users to use it?

pgovind: This is meant to be a weakly typed way to get _n_ rows in a `DataFrameColumn`. So, something like this to return the values between indices 20 and 29:
`var column = dataFrame["SomeColumnName"]
column[20, 10]` 
Without this method, there'd be no way to get more than 1 row at a time from the base `DataFrameColumn`. On the derived types, the indexers are strongly typed, so this is not an issue.

eerhardt:
@pgovind - can you sync and resolve with master?

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -90,13 +90,14 @@ private static Type MaxKind(Type a, Type b)
...
         /// <returns>DataFrame</returns>
-        public static DataFrame ReadCsv(string filename,
+        public static DataFrame LoadCsv(string filename,
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
                                 int numRows = -1, int guessRows = 10,
                                 bool addIndexColumn = false)
         {
-            return ReadStream(() => new StreamReader(filename),
+            Stream fileStream = new FileStream(filename, FileMode.Open);
```
eerhardt: You need to dispose this stream when you are done. So do a `using Stream fileStream = new ....`.

src/Microsoft.Data/DataFrameColumn.cs
```diff
@@ -87,9 +87,15 @@ public void SetName(string newName, DataFrame dataFrame = null)
         /// <param name="mapIndices"></param>
         /// <param name="invertMapIndices"></param>
         /// <returns></returns>
-        public virtual DataFrameColumn Clone(DataFrameColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => throw new NotImplementedException();
+        public virtual DataFrameColumn Clone(DataFrameColumn mapIndices = null, bool invertMapIndices = false, long numberOfNullsToAppend = 0) => CloneImplementation(mapIndices, invertMapIndices, numberOfNullsToAppend);
 
-        public virtual DataFrameColumn Sort(bool ascending = true) => throw new NotImplementedException();
+        protected virtual DataFrameColumn CloneImplementation(DataFrameColumn mapIndices, bool invertMapIndices, long numberOfNullsToAppend) => throw new NotImplementedException();
+
+        public virtual DataFrameColumn Sort(bool ascending = true)
```
eerhardt: xml doc?

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
```diff
@@ -170,13 +170,21 @@ protected internal override Apache.Arrow.Array ToArrowArray(long startIndex, int
...
-        protected override object GetValue(long startIndex, int length)
+        protected override IReadOnlyList<object> GetValues(long startIndex, int length)
         {
             if (startIndex > Length)
             {
                 throw new ArgumentOutOfRangeException(nameof(startIndex));
             }
-            return _columnContainer[startIndex, length];
+
+            //return _columnContainer[startIndex, length];
```
eerhardt: This commented out code can be deleted.

src/Microsoft.Data/PrimitiveDataFrameColumn.Sort.cs
```diff
@@ -13,13 +13,13 @@ namespace Microsoft.Data
     public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn
         where T : unmanaged
     {
-        public override DataFrameColumn Sort(bool ascending = true)
+        public new PrimitiveDataFrameColumn<T> Sort(bool ascending = true)
         {
-            PrimitiveDataFrameColumn<long> sortIndices = GetAscendingSortIndices() as PrimitiveDataFrameColumn<long>;
-            return Clone(sortIndices, !ascending, NullCount);
+            PrimitiveDataFrameColumn<long> sortIndices = GetAscendingSortIndices();
+            return Clone(sortIndices, !ascending, NullCount) as PrimitiveDataFrameColumn<T>;
```
eerhardt: Do you need the `as` here? Don't we have a Clone method that returns the strong type?

src/Microsoft.Data/StringDataFrameColumn.cs
```diff
@@ -418,7 +425,7 @@ public override GroupBy GroupBy(int columnIndex, DataFrame parent)
             }
         }
 
-        public override DataFrameColumn FillNulls(object value, bool inPlace = false)
+        public new StringDataFrameColumn FillNulls(object value, bool inPlace = false)
```
eerhardt: `object value` => `string value`

pgovind:
Cherry-picked on top of master.

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -90,22 +90,25 @@ private static Type MaxKind(Type a, Type b)
...
+            using (Stream fileStream = new FileStream(filename, FileMode.Open))
+            {
+                return LoadStream(fileStream,
+                                  separator: separator, header: header, columnNames: columnNames, dataTypes: dataTypes, numberOfRowsToRead: numRows,
+                                  guessRows: guessRows, addIndexColumn: addIndexColumn);
+            }
         }
 
         /// <summary>
         /// Reads a text file as a DataFrame.
```
eerhardt: This isn't necessarily a text file. It is any seekable stream.

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -114,22 +117,27 @@ private static Type MaxKind(Type a, Type b)
         /// <param name="guessRows">number of rows used to guess types</param>
         /// <param name="addIndexColumn">add one column with the row index</param>
         /// <returns>DataFrame</returns>
-        public static DataFrame ReadStream(Func<StreamReader> createStream,
+        public static DataFrame LoadStream(Stream csvStream,
```
eerhardt: I think this should be named `LoadCsv`. `LoadStream` doesn't describe this method enough.

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -114,22 +117,27 @@ private static Type MaxKind(Type a, Type b)
...
             int numberOfColumns = 0;
 
             if (header == true && numberOfRowsToRead != -1)
                 numberOfRowsToRead++;
 
+            List<DataFrameColumn> columns;
+            long streamStart = csvStream.Position;
             // First pass: schema and number of rows.
-            using (var st = createStream())
+            using (var streamReader = new StreamReader(csvStream))
```
eerhardt: I believe the `StreamReader` will dispose the underlying stream when the `StreamReader` is disposed.

Is it expected that calling `DataFrame.LoadStream(someStream)` is going to dispose the `someStream`?

src/Microsoft.Data/StringDataFrameColumn.cs
```diff
@@ -426,7 +433,7 @@ public override GroupBy GroupBy(int columnIndex, DataFrame parent)
             }
         }
 
-        public override DataFrameColumn FillNulls(object value, bool inPlace = false)
+        public StringDataFrameColumn FillNulls(string value, bool inPlace = false)
         {
             if (value.GetType() != typeof(string) || value == null)
```
eerhardt: no longer need `value.GetType() != typeof(string)`.

Especially because if `value` is null, this would have NRE.

src/Microsoft.Data/StringDataFrameColumn.cs
```diff
@@ -441,6 +448,14 @@ public override DataFrameColumn FillNulls(object value, bool inPlace = false)
             return column;
         }
 
+        protected override DataFrameColumn FillNullsImplementation(object value, bool inPlace)
+        {
+            if (value is string)
+                return FillNulls((string)value, inPlace);
```
eerhardt: Instead of doing 2 type checks, you can just do one if you write the code like this:

```C#
            if (value is string valueString)
                return FillNulls(valueString, inPlace);
```

src/Microsoft.Data/DataFrame.IO.cs
```diff
@@ -114,22 +117,27 @@ private static Type MaxKind(Type a, Type b)
...
             int numberOfColumns = 0;
 
             if (header == true && numberOfRowsToRead != -1)
                 numberOfRowsToRead++;
 
+            List<DataFrameColumn> columns;
+            long streamStart = csvStream.Position;
             // First pass: schema and number of rows.
-            using (var st = createStream())
+            using (var streamReader = new StreamReader(csvStream, Encoding.UTF8, detectEncodingFromByteOrderMarks: true, bufferSize: 1024, leaveOpen: true))
```
eerhardt: I believe the right defaults for the other parameters should be:

```C#
StreamReader(Stream stream, [NullableAttribute(2)] Encoding? encoding = null, bool detectEncodingFromByteOrderMarks = true, int bufferSize = -1, bool leaveOpen = false);
```

At least that is what `netcoreapp3.0` has for defaults on these parameters.

pgovind: Weird, I don't see the constructor with the default values in VS, but I see it on source.dot.net. I'm on netstandard 2.0

eerhardt: The constructor with the default values isn't on `netstandard2.0`. It is only on `netcoreapp3.0`.

The reason I brought it up is because you are forced to put default values here, and you should be using the same defaults as we added for `netcoreapp3.0`.

PR #2761, opened 2019-10-21
Update all projects to netstandard 2.0 minimum

GrabYourPitchforks:
We're not shipping anything that targets frameworks prior to `netstandard2.0`, so no point to compiling corefxlab projects for those frameworks.

This PR also updates the SDK used by corefxlab to .NET Core 3.0 RTM, which also means that System.Text.Json is now built-in and all of the old JsonLab tests and dependent projects are failing due to API changes and type name collisions. Since we've already shipped System.Text.Json, I've removed JsonLab and related projects entirely as part of this PR.


tests/Benchmarks/System.Text.JsonLab/JsonParserPerf.cs
```diff
@@ -1,347 +0,0 @@
...
-using System.Collections.Generic;
-using System.IO;
-using System.Text.Utf8;
-
-namespace System.Text.JsonLab.Benchmarks
-{
-    // Since there are 120 tests here (8 * 15), setting low values for the warmupCount and targetCount
-    [SimpleJob(-1, 3, 5)]
-    [MemoryDiagnoser]
-    public class JsonParserPerf
```
ahsonkhan: I believe most of the relevant benchmarks from here got moved to dotnet/performance.

If we missed some interesting tests, we can always look back at the history and grab some :)

cc @Jozkee

GrabYourPitchforks: Yup. Looks like it got moved to https://github.com/dotnet/performance/blob/master/src/benchmarks/micro/corefx/System.Text.Json/Utf8JsonReader/Perf.Reader.cs.

tests/System.Text.JsonLab.Tests/ReadFromPipeTests.cs
```diff
@@ -1,232 +0,0 @@
...
-using System.Buffers;
-using System.IO;
-using System.IO.Pipelines;
-using System.Text.JsonLab.Tests.Resources;
-using System.Threading.Tasks;
-using Xunit;
-
-namespace System.Text.JsonLab.Tests
-{
-    public class ReadFromPipeTests : IDisposable
```
ahsonkhan: FYI @tdykstra
This is roughly what reading from Pipes would look like. We'd want to re-write this against the final API shape that we shipped with (this test in corefxlab was written against the prototype code/API). I will re-write this and share the sample this week (just a heads up in case you were interested in giving it a shot).

The write side is in WriteToPipTests.cs.

PR #2763, opened 2019-10-22
Remove RangeLab type as BCL introduced Range type

### LINKED ISSUE #2628: Update and replace uses or RangeLab with System.Range within corefxlab
.NET Core 3.0 contains the new `Range` type https://github.com/dotnet/corefx/blob/4378851bc5fd0e3c5043a6d0a3ac835e838fcf7d/src/Common/src/CoreLib/System/Range.cs#L9

We need to remove the RangeLab type from corefxlab and update its uses with the new `Range`.

cc @tarekgh

tarekgh:
Fixes #2628


tarekgh:
@ahsonkhan could you please have a look?

tests/Benchmarks/System.Buffers.Experimental/RangeEnumeration.cs
```diff
@@ -1,48 +0,0 @@
-﻿// Licensed to the .NET Foundation under one or more agreements.
-// The .NET Foundation licenses this file to you under the MIT license.
-// See the LICENSE file in the project root for more information.
-
-using BenchmarkDotNet.Attributes;
-
-namespace System.Buffers.Experimental.Benchmarks
-{
-    public class RangeEnumeration
```
ahsonkhan: Should we migrate these perf tests to dotnet/performance?

cc @adamsitnik

adamsitnik: @ahsonkhan yes, thanks for pointing this! I've created https://github.com/dotnet/performance/issues/1002 to track it

PR #2765, opened 2019-10-24
Add strongly typed indexers

pgovind:
Should be an easy stamp

### DIFF
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -201,11 +201,6 @@ namespace Microsoft.Data
         }
 
         /// <summary>
-        /// Updates <paramref name="column"/>'s name with <paramref name="newName"/>
-        /// </summary>
-        public void SetColumnName(DataFrameColumn column, string newName) => _columnCollection.SetColumnName(column, newName);
-
-        /// <summary>
         /// Generates descriptive statistics that summarize each numeric column
         /// </summary>
         public DataFrame Description()
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -64,17 +64,73 @@ namespace Microsoft.Data
         }
 
         /// <summary>
-        /// Returns a new DataFrame using the rows or true values in filterColumn
+        /// Returns a new DataFrame using the boolean values in <paramref name="filter"/>
         /// </summary>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data/DataFrame.cs
```diff
@@ -64,17 +64,73 @@ public DataFrame(IList<DataFrameColumn> columns)
...
 
         /// <summary>
-        /// Returns a new DataFrame using the rows or true values in filterColumn
+        /// Returns a new DataFrame using the boolean values in <paramref name="filter"/>
         /// </summary>
-        /// <param name="filterColumn">A column of rows/bools</param>
-        /// <remarks>filterColumn must be of type long, int or bool</remarks>
-        public DataFrame this[DataFrameColumn filterColumn] => Clone(filterColumn);
+        /// <param name="filter">A column of booleans</param>
+        public DataFrame Filter(PrimitiveDataFrameColumn<bool> filter) => Clone(filter);
```
eerhardt: Why have both indexers and `Filter` methods?

src/Microsoft.Data/DataFrame.cs
```diff
@@ -136,6 +200,9 @@ private DataFrame Clone(DataFrameColumn mapIndices = null, bool invertMapIndices
             return new DataFrame(newColumns);
         }
 
+        /// <summary>
+        /// Updates <paramref name="column"/>'s name with <paramref name="newName"/>
```
eerhardt: I thought this method was moved to the `Columns` collection in your previous PR, but I guess it wasn't.

What do you think about removing this and leaving the `dataFrame.Columns.SetColumnName` method? We currently have 2 public functions that do the same thing - one on DataFrame, and one on ColumnsCollection.

src/Microsoft.Data/DataFrame.cs
pgovind: We don't HAVE to have it. I added it to aid discoverability that we have a "Filter" method. Otherwise it'd be hidden in the indexers, and I'm not certain users will find it easily.

src/Microsoft.Data/DataFrame.cs

src/Microsoft.Data/DataFrame.cs
eerhardt: Ok, we can always change it later before making this a stable API if we get feedback either way on it.

PR #2766, opened 2019-10-24
IDataView to DataFrame

pgovind:
Implements https://github.com/dotnet/corefxlab/issues/2741
Also fixes https://github.com/dotnet/corefxlab/issues/2754

### DIFF
--- a/src/Microsoft.Data/DataFrame.IDataView.cs
+++ b/src/Microsoft.Data/DataFrame.IDataView.cs
@@ -136,5 +136,86 @@ namespace Microsoft.Data
                 return _position < _dataFrame.RowCount;
             }
         }
+
+        public static DataFrame FromIDataView(IDataView dataView)
+        {
+            DataViewSchema schema = dataView.Schema;
+            List<DataFrameColumn> columns = new List<DataFrameColumn>(schema.Count);
+
+            List<DataViewSchema.Column> activeColumns = new List<DataViewSchema.Column>();
+            foreach (DataViewSchema.Column column in schema)
+            {
+                if (column.IsHidden)
+                {
+                    continue;
+                }
+                activeColumns.Add(column);
+                DataViewType type = column.Type;
+                if (type == BooleanDataViewType.Instance)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<bool>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.Byte)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<byte>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.Double)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<double>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.Single)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<float>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.Int32)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<int>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.Int64)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<long>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.SByte)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<sbyte>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.Int16)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<short>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.UInt32)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<uint>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.UInt64)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<ulong>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == NumberDataViewType.UInt16)
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<ushort>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == TextDataViewType.Instance)
+                {
+                    columns.Add(new StringDataFrameColumn(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else
+                {
+                    throw new NotSupportedException(nameof(type));
+                }
+            }
+
+            DataFrame ret = new DataFrame(columns);
+            DataViewRowCursor cursor = dataView.GetRowCursor(activeColumns);
+            while (cursor.MoveNext())
+            {
+                foreach (var column in activeColumns)
+                {
+                    columns[column.Index].AddValueUsingCursor(cursor, column);
+                }
+            }
+
+            return ret;
+        }
     }
 }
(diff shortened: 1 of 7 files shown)


src/Microsoft.Data/StringDataFrameColumn.cs
```diff
@@ -445,5 +445,24 @@ protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
...
             (ref ReadOnlyMemory<char> value) => value = this[cursor.Position].AsMemory();
+
+        protected internal override void AddValueUsingCursor(DataViewRowCursor cursor, DataViewSchema.Column schemaColumn)
+        {
+            long row = cursor.Position;
+            ReadOnlyMemory<char> value = default;
+            cursor.GetGetter<ReadOnlyMemory<char>>(schemaColumn)(ref value);
+            if (Length > row)
+            {
+                this[row] = new string(value.ToArray());
```
pgovind: Not sure if there's a way to avoid this copy

src/Microsoft.Data/StringDataFrameColumn.BinaryOperations.cs
```diff
@@ -42,54 +42,74 @@ public override DataFrameColumn Add<T>(T value, bool inPlace = false)
             return ret;
         }
 
-        public override PrimitiveDataFrameColumn<bool> ElementwiseEquals(DataFrameColumn column)
+        internal static PrimitiveDataFrameColumn<bool> ElementwiseEqualsImplementation(DataFrameColumn left, DataFrameColumn column)
```
pgovind: Let me know if you want me to put this into its own PR. The test case I added for the IDataView->DataFrame unit test failed because this method was missing on `ArrowStringDataFrameColumn` and it was easy enough to refactor

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.IDataView.cs
```diff
@@ -224,5 +224,19 @@ public void TestIDataViewWithNulls()
             Assert.Equal("", preview.ColumnView[14].Values[5].ToString()); // null row
             Assert.Equal("foo", preview.ColumnView[14].Values[6].ToString());
         }
+
+        [Fact]
+        public void TestDataFrameFromIDataView()
+        {
+            DataFrame df = MakeDataFrameWithAllColumnTypes(10, withNulls: false);
+            IDataView dfAsIDataView = df;
+            DataFrame newDf = DataFrame.FromIDataView(dfAsIDataView);
```
eerhardt: What do you think about making an extension method on `IDataView` that returns the `DataFrame`?

```C#
IDataView values = ...;
ITransformer mlModel = ...;

DataFrame df = mlModel.Transform(values).ToDataFrame();
```

vs.

```C#
IDataView values = ...;
ITransformer mlModel = ...;

DataFrame df = DataFrame.FromIDataView(mlModel.Transform(values));
```

src/Microsoft.Data/PrimitiveDataFrameColumn.cs
```diff
@@ -652,5 +652,24 @@ protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
 
         private static ValueGetter<double> CreateDecimalValueGetterDelegate(DataViewRowCursor cursor, PrimitiveDataFrameColumn<decimal> column) =>
             (ref double value) => value = (double?)column[cursor.Position] ?? double.NaN;
+
+        protected internal override void AddValueUsingCursor(DataViewRowCursor cursor, DataViewSchema.Column column)
+        {
+            long row = cursor.Position;
+            T value = default;
+            cursor.GetGetter<T>(column)(ref value);
```
eerhardt: The `Getter` delegate should be cached outside of the "loop" of cursoring over an IDataView. Getting this for each "cell" in the IDataView is going to be super slow.

Check out [https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewImplementation.md#why-getgetter](https://github.com/dotnet/machinelearning/blob/master/docs/code/IDataViewImplementation.md#why-getgetter)

src/Microsoft.Data/DataFrame.IDataView.cs
```diff
@@ -136,5 +136,86 @@ public override bool MoveNext()
                 return _position < _dataFrame.RowCount;
             }
         }
+
+        public static DataFrame FromIDataView(IDataView dataView)
```
eerhardt: We should also add the ability to limit the number of rows, and which columns. See the description in the issue [https://github.com/dotnet/corefxlab/issues/2741](https://github.com/dotnet/corefxlab/issues/2741)

> And we can provide optional arguments to limit the number of rows, and to limit which columns are selected.

src/Microsoft.Data/StringDataFrameColumn.BinaryOperations.cs
```diff
@@ -42,54 +42,74 @@ public override DataFrameColumn Add<T>(T value, bool inPlace = false)
...
+
+        internal static PrimitiveDataFrameColumn<bool> ElementwiseNotEqualsImplementation<T>(DataFrameColumn left, T value)
+        {
+            PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(left.Name, left.Length);
             string valString = value.ToString();
-            for (long i = 0; i < Length; i++)
+            for (long i = 0; i < left.Length; i++)
             {
-                ret[i] = (string)this[i] != valString;
+                ret[i] = (string)left[i] != valString;
```
eerhardt: For "Arrow strings", wouldn't it be faster to convert `T value` to UTF8, and then do the comparison against the UTF8 values? That way we wouldn't need to materialize all the values as UTF16 strings.

src/Microsoft.Data/StringDataFrameColumn.cs
eerhardt: call `value.ToString()`. 

[https://source.dot.net/#System.Private.CoreLib/shared/System/ReadOnlyMemory.cs,156](https://source.dot.net/#System.Private.CoreLib/shared/System/ReadOnlyMemory.cs,156)

[https://source.dot.net/#System.Private.CoreLib/shared/System/String.Manipulation.cs,1661](https://source.dot.net/#System.Private.CoreLib/shared/System/String.Manipulation.cs,1661)

If the value is the whole backing string, it will return the whole string without copying.

---
In reply to: [338807430](https://github.com/dotnet/corefxlab/pull/2766#discussion_r338807430) [](ancestors = 338807430)

src/Microsoft.Data/DataFrame.IDataView.cs
```diff
@@ -136,5 +136,86 @@ public override bool MoveNext()
...
+                {
+                    columns.Add(new PrimitiveDataFrameColumn<ushort>(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else if (type == TextDataViewType.Instance)
+                {
+                    columns.Add(new StringDataFrameColumn(column.Name, dataView.GetRowCount() ?? 0));
+                }
+                else
+                {
+                    throw new NotSupportedException(nameof(type));
```
eerhardt: One issue we are going to hit immediately is that IDataView supports a "Vector" column. So in a single cell, you can store a vector of floats, for example.

We will probably need a new column type for this.

src/Microsoft.Data/StringDataFrameColumn.cs
pgovind: Interesting. Intellisense says this: `Returns the fully qualified type name of this instance` but I see that the method has different documentation on source.dot.net. I wonder why...

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.IDataView.cs
pgovind: Yup, you brought it up before I had a chance to. I'm in agreement.

src/Microsoft.Data/DataFrame.IDataView.cs
pgovind: The rest of the feedback here is easy to fix. This one will involve some work though, since a new column has to support a bunch of functions. I'm still debating the priority here with respect to shipping a preview

eerhardt: Feel free to log a tracking issue for this.

pgovind:
Closing this for now. I'll reopen once we get a preview out

PR #2767, opened 2019-10-24
Sort fix

pgovind:
Fix and unit test for a bug that Maryam hit a while back.

### DIFF
--- a/src/Microsoft.Data/PrimitiveDataFrameColumn.Sort.cs
+++ b/src/Microsoft.Data/PrimitiveDataFrameColumn.Sort.cs
@@ -33,7 +33,7 @@ namespace Microsoft.Data
             for (int b = 0; b < _columnContainer.Buffers.Count; b++)
             {
                 ReadOnlyDataFrameBuffer<T> buffer = _columnContainer.Buffers[b];
-                ReadOnlySpan<byte> nullBitMapSpan = _columnContainer.NullBitMapBuffers[b].ReadOnlySpan;
+                ReadOnlySpan<byte> nullBitMapSpan = _columnContainer.NullBitMapBuffers[b].RawReadOnlySpan;
                 int[] sortIndices = new int[buffer.Length];
                 for (int i = 0; i < buffer.Length; i++)
                     sortIndices[i] = i;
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data/PrimitiveDataFrameColumn.Sort.cs
```diff
@@ -33,7 +33,7 @@ private void GetSortIndices(IComparer<T> comparer, out PrimitiveDataFrameColumn<
             for (int b = 0; b < _columnContainer.Buffers.Count; b++)
             {
                 ReadOnlyDataFrameBuffer<T> buffer = _columnContainer.Buffers[b];
-                ReadOnlySpan<byte> nullBitMapSpan = _columnContainer.NullBitMapBuffers[b].ReadOnlySpan;
+                ReadOnlySpan<byte> nullBitMapSpan = _columnContainer.NullBitMapBuffers[b].RawReadOnlySpan;
```
pgovind: Each `NullBitMapBuffer` stores flags for 8 indices in 1 byte. However, `NullBitMapBuffer.Length` is set to numberOfBytes * 8 to match it's data buffer's `Length`. `NullBitMapBuffer.ReadOnlySpan` internally calls `Slice(start, Length)`, but that could fail with an `IndexOutOfRangeException`. Instead, when dealing with NullBitMapBuffers, `NullBitMapBuffer.RawReadOnlySpan` should be used to get the span of bytes. I missed 1 spot that still used `ReadOnlySpan`.

src/Microsoft.Data/StringDataFrameColumn.cs
```diff
@@ -316,7 +316,11 @@ private StringDataFrameColumn CloneImplementation<U>(PrimitiveDataFrameColumn<U>
                         getBufferMaxRange = (bufferIndex + 1) * maxCapacity;
                     }
                     int bufferLocalMapIndex = (int)(mapIndex - getBufferMinRange);
-                    setBuffer[(int)index] = getBuffer[bufferLocalMapIndex];
+                    string value = getBuffer[bufferLocalMapIndex];
+                    setBuffer[(int)index] = value;
+                    if (value == null)
+                        ret._nullCount++;
```
pgovind: The existing unit tests didn't catch this `NullCount` mismatch because of a subtle difference in behavior in `Filter` vs `Sort`. Added a new unit test for this too now.

PR #2768, opened 2019-10-28
Add string specific APIs to StringDataFrameColumn

pgovind:
Add StringDataFrameColumn specific APIs. I found these last week while trying to explore a dataset

### DIFF
--- a/src/Microsoft.Data/StringDataFrameColumn.BinaryOperations.cs
+++ b/src/Microsoft.Data/StringDataFrameColumn.BinaryOperations.cs
@@ -33,7 +33,17 @@ namespace Microsoft.Data
 
         public static StringDataFrameColumn operator+(string value, StringDataFrameColumn column)
         {
-            return column.Add(value);
+            StringDataFrameColumn ret = column.Clone();
+            for (int i = 0; i < ret._stringBuffers.Count; i++)
+            {
+                IList<string> buffer = ret._stringBuffers[i];
+                int bufferLen = buffer.Count;
+                for (int j = 0; j < bufferLen; j++)
+                {
+                    buffer[j] = value + buffer[j];
+                }
+            }
+            return ret;
         }
 
         public StringDataFrameColumn Add(string value, bool inPlace = false)
@@ -65,7 +75,7 @@ namespace Microsoft.Data
             PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(Name, Length);
             for (long i = 0; i < Length; i++)
             {
-                ret[i] = (string)this[i] == column[i]?.ToString();
+                ret[i] = this[i] == column[i]?.ToString();
             }
             return ret;
         }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data/StringDataFrameColumn.BinaryOperations.cs
```diff
@@ -26,22 +26,36 @@ public override DataFrameColumn Add(DataFrameColumn column, bool inPlace = false
...
 
-        public override DataFrameColumn Add<T>(T value, bool inPlace = false)
+        public static StringDataFrameColumn operator+(StringDataFrameColumn column, string value)
+        {
+            return column.Add(value);
+        }
+
+        public static StringDataFrameColumn operator+(string value, StringDataFrameColumn column)
+        {
+            return column.Add(value);
```
eerhardt: I would think that:

```
"myString" + someColumn
```

Would result in `myString` being prepended to each column value. Here it appears that the order of operations is not important, but to a user I think it is.

src/Microsoft.Data/StringDataFrameColumn.BinaryOperations.cs
```diff
@@ -56,13 +70,27 @@ public override PrimitiveDataFrameColumn<bool> ElementwiseEquals(DataFrameColumn
             return ret;
         }
 
+        public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
+        {
+            PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(Name, Length);
+            for (long i = 0; i < Length; i++)
+            {
+                ret[i] = (string)this[i] == value;
```
eerhardt: Why is this cast necessary? `this[i]` should return a `string`.

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -442,9 +446,26 @@ public void TestBinaryOperationsOnStringColumn()
...
             Assert.Equal(false, newCol[5]);
             Assert.Equal(false, newCol[0]);
+
+            StringDataFrameColumn typedStringColumn = stringColumn as StringDataFrameColumn;
+            newCol = typedStringColumn.Add("suffix");
+            for (int i = 0; i < newCol.Length; i++)
+            {
+                Assert.Equal(newCol[i], typedStringColumn[i] + "suffix");
+            }
+            DataFrameColumn addString = typedStringColumn + "suffix";
```
eerhardt: Need a test for `"prefix" + typedStringColumn`

src/Microsoft.Data/StringDataFrameColumn.BinaryOperations.cs
```diff
@@ -26,22 +26,46 @@ public override DataFrameColumn Add(DataFrameColumn column, bool inPlace = false
             return ret;
         }
 
-        public override DataFrameColumn Add<T>(T value, bool inPlace = false)
+        public static StringDataFrameColumn operator+(StringDataFrameColumn column, string value)
+        {
+            return column.Add(value);
+        }
+
+        public static StringDataFrameColumn operator+(string value, StringDataFrameColumn column)
```
eerhardt: See [https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/operator-overloads](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/operator-overloads)

> ✓ CONSIDER providing methods with friendly names that correspond to each overloaded operator.

> Many languages do not support operator overloading. For this reason, it is recommended that types that overload operators include a secondary method with an appropriate domain-specific name that provides equivalent functionality.

PR #2769, opened 2019-10-29
APIs for reversed binary operators

pgovind:
To enable statements such as `1 - intColumn`

### DIFF
--- a/src/Microsoft.Data/PrimitiveDataFrameColumn.ReversedBinaryOperations.cs
+++ b/src/Microsoft.Data/PrimitiveDataFrameColumn.ReversedBinaryOperations.cs
@@ -1,0 +1,394 @@
+﻿
+// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+// Generated from PrimitiveDataFrameColumn.ReversedBinaryOperations.tt. Do not modify directly
+
+using System;
+using System.Collections.Generic;
+
+namespace Microsoft.Data
+{
+    public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn
+        where T : unmanaged
+    {
+
+        public override DataFrameColumn ReverseAdd<U>(U value, bool inPlace = false)
+        {
+            switch (this)
+            {
+                case PrimitiveDataFrameColumn<bool> boolColumn:
+                    throw new NotSupportedException();
+                case PrimitiveDataFrameColumn<decimal> decimalColumn:
+                    if (typeof(U) == typeof(bool))
+                    {
+                        throw new NotSupportedException();
+                    }
+                    if (typeof(U) == typeof(T))
+                    {
+                        // No conversions
+                        PrimitiveDataFrameColumn<U> primitiveColumn = this as PrimitiveDataFrameColumn<U>;
+                        PrimitiveDataFrameColumn<U> newColumn = inPlace ? primitiveColumn : primitiveColumn.Clone();
+                        newColumn._columnContainer.ReverseAdd(value);
+                        return newColumn;
+                    }
+                    else
+                    {
+                        PrimitiveDataFrameColumn<decimal> clonedDecimalColumn = CloneAsDecimalColumn();
+                       clonedDecimalColumn._columnContainer.ReverseAdd(DecimalConverter<U>.Instance.GetDecimal(value));
+                        return clonedDecimalColumn;
+                    }
+                case PrimitiveDataFrameColumn<byte> byteColumn:
+                case PrimitiveDataFrameColumn<char> charColumn:
+                case PrimitiveDataFrameColumn<double> doubleColumn:
+                case PrimitiveDataFrameColumn<float> floatColumn:
+                case PrimitiveDataFrameColumn<int> intColumn:
+                case PrimitiveDataFrameColumn<long> longColumn:
+                case PrimitiveDataFrameColumn<sbyte> sbyteColumn:
+                case PrimitiveDataFrameColumn<short> shortColumn:
+                case PrimitiveDataFrameColumn<uint> uintColumn:
+                case PrimitiveDataFrameColumn<ulong> ulongColumn:
+                case PrimitiveDataFrameColumn<ushort> ushortColumn:
+                    if (typeof(U) == typeof(bool))
+                    {
+                        throw new NotSupportedException();
+                    }
+                    if (typeof(U) == typeof(T))
+                    {
+                        // No conversions
+                        PrimitiveDataFrameColumn<U> primitiveColumn = this as PrimitiveDataFrameColumn<U>;
+                        PrimitiveDataFrameColumn<U> newColumn = inPlace ? primitiveColumn : primitiveColumn.Clone();
+                        newColumn._columnContainer.ReverseAdd(value);
+                        return newColumn;
+                    }
+                    else
+                    {
+                        if (typeof(U) == typeof(decimal))
+                        {
+                            PrimitiveDataFrameColumn<decimal> decimalColumn = CloneAsDecimalColumn();
+                            decimalColumn._columnContainer.ReverseAdd(DecimalConverter<U>.Instance.GetDecimal(value));
+                            return decimalColumn;
+                        }
+                        else
+                        {
+                            PrimitiveDataFrameColumn<double> clonedDoubleColumn = CloneAsDoubleColumn();
+                            clonedDoubleColumn._columnContainer.ReverseAdd(DoubleConverter<U>.Instance.GetDouble(value));
+                            return clonedDoubleColumn;
+                        }
+                    }
+                default:
+                    throw new NotSupportedException();
+            }
+        }
+        public override DataFrameColumn ReverseSubtract<U>(U value, bool inPlace = false)
+        {
+            switch (this)
+            {
+                case PrimitiveDataFrameColumn<bool> boolColumn:
+                    throw new NotSupportedException();
+                case PrimitiveDataFrameColumn<decimal> decimalColumn:
+                    if (typeof(U) == typeof(bool))
+                    {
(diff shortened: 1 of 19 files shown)


src/Microsoft.Data/DataFrame.BinaryOperations.cs
```diff
@@ -26,12 +26,15 @@ public DataFrame Add<T>(IReadOnlyList<T> values, bool inPlace = false)
             for (int i = 0; i < Columns.Count; i++)
             {
                 DataFrameColumn baseColumn = _columnCollection[i];
+                baseColumn._reverseOrderOfBinaryOperations = _reverseOrderOfBinaryOperations;
```
pgovind: Adding internal state to `DataFrame` and `PrimitiveDataFrameColumn` seemed to be a cleaner than adding parameters to each binary op.

pgovind:
Ok, can be reviewed now.

src/Microsoft.Data/DataFrame.BinaryOperations.cs
eerhardt: I think I disagree. If an exception is thrown between here and line 35, this state is going to be left to this old value.

---
In reply to: [340205425](https://github.com/dotnet/corefxlab/pull/2769#discussion_r340205425) [](ancestors = 340205425)

eerhardt: Also, how could someone write their own derived column class? They won't be able to get access to this internal state.

Can we find some other way to enable this?

---
In reply to: [340322971](https://github.com/dotnet/corefxlab/pull/2769#discussion_r340322971) [](ancestors = 340322971,340205425)

pgovind: For the 2nd question, I thought of that and put the flag on the base class :)

I think I'm going to make new `Radd`, `Rdivide` etc virtual methods. That should fix all our concerns

eerhardt: The method names should follow the .NET guidelines though - 

https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/naming-guidelines

src/Microsoft.Data/DataFrame.BinaryOperations.cs
```diff
@@ -506,6 +504,123 @@ public DataFrame ElementwiseLessThan<T>(T value)
             }
             return retDataFrame;
         }
-        #endregion
+
+        public DataFrame RAdd<T>(T value, bool inPlace = false)
```
pgovind: All these methods need documentation

src/Microsoft.Data/DataFrameColumn.BinaryOperations.cs
```diff
@@ -23,6 +76,12 @@ public virtual DataFrameColumn Add<T>(T value, bool inPlace = false)
             throw new NotImplementedException();
         }
 
+        public virtual DataFrameColumn RAdd<T>(T value, bool inPlace = false)
```
pgovind: Needs documentation

src/Microsoft.Data/DataFrame.BinaryOperations.cs
pgovind: `RAdd` follows our naming guidelines, but `ROr` sounds a little weird. I don't think we should name it `ReverseAdd` though

pgovind:
Can be reviewed. This is much better than the previous version.

pgovind:
https://github.com/dotnet/corefxlab/issues/2749 can be fixed only after this goes in.

src/Microsoft.Data/DataFrame.BinaryOperations.cs
```diff
@@ -506,6 +552,147 @@ public DataFrame ElementwiseLessThan<T>(T value)
             }
             return retDataFrame;
         }
-        #endregion
+
+        /// <summary>
+        /// Performs a reversed element wise addition on each column
+        /// </summary>
+        public DataFrame RAdd<T>(T value, bool inPlace = false)
```
eerhardt: `RAdd` doesn't follow our naming guidelines. See [https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions](https://docs.microsoft.com/en-us/dotnet/standard/design-guidelines/general-naming-conventions).

src/Microsoft.Data/DataFrame.BinaryOperators.cs
```diff
@@ -16,339 +16,685 @@ public partial class DataFrame
...
+        }
 
         public static DataFrame operator &(DataFrame df, bool value)
         {
             return df.And(value);
         }
+
+        public static DataFrame operator &(bool value, DataFrame df)
+        {
+            DataFrame ret = df.RAnd(value, inPlace: false);
```
eerhardt: Why not just `return df.RAnd(value, inPlace: false);`?

src/Microsoft.Data/DataFrameColumn.BinaryOperations.cs
```diff
@@ -12,96 +12,248 @@ namespace Microsoft.Data
 {
     public abstract partial class DataFrameColumn
     {
+        public static DataFrameColumn Add<T>(T value, DataFrameColumn right)
```
eerhardt: I'm confused why we have both public static `Add(T, DataFrameColumn)` and `DataFrameColumn.RAdd(T)`? Why do they both need to be public?

src/Microsoft.Data/DataFrame.BinaryOperations.cs
pgovind: Hmmm, I thought `RAdd` followed our guidelines. Are you looking at the "favor readability over brevity" one? I could live with `ReverseAdd` I guess. This is my comment from before:

> `RAdd` follows our naming guidelines, but `ROr` sounds a little weird. `ReverseAdd/ReverseOr` sound better?

eerhardt: ✓ DO favor readability over brevity.

> The property name CanScrollHorizontally is better than ScrollableX (an obscure reference to the X-axis).

What is `R` in these methods? Just like the "obscure reference to the X-axis" above, it is hard to guess what is meant by a single letter.

And yes:

X DO NOT use abbreviations or contractions as part of identifier names.

Also applies here.

eerhardt: I would spell out `Reverse` and keep it before the operation name.

`ReverseAdd`, `ReverseSubtract`, `ReverseDivide`, etc.

src/Microsoft.Data/DataFrameColumn.BinaryOperations.cs
pgovind: I decided to get rid of the static methods. The static operators now call the appropriate instance methods directly.

src/Microsoft.Data/PrimitiveColumnContainer.BinaryOperations.cs
```diff
@@ -190,5 +190,45 @@ public PrimitiveColumnContainer<T> ElementwiseLessThan(T scalar, PrimitiveColumn
             return this;
        }
 
+       public PrimitiveColumnContainer<T> RAdd(T scalar)
```
pgovind: This should change to Reverse

pgovind:
Rebased on top of master

src/Microsoft.Data/DataFrame.BinaryOperators.cs
```diff
@@ -16,339 +16,682 @@ public partial class DataFrame
         {
             return df.Add(value);
         }
+        
+        public static DataFrame operator +(bool value, DataFrame df)
+        {
+            return df.ReverseAdd(value);
```
eerhardt: I probably should have caught this months ago, but we have a chance of a null reference exception here if `df` is null. This also is possible in the existing code above. Maybe we can just log a separate bug for it and fix it later.

ex.

```
DataFrame df = null;
df2 = 1 + df;
```

I'm not entirely sure on what the behavior here should be. If we should throw an exception, or return `null`. Thoughts?

src/Microsoft.Data/PrimitiveDataFrameColumn.ReversedBinaryOperations.tt
```diff
@@ -0,0 +1,110 @@
...
+                        // No conversions
+                        PrimitiveDataFrameColumn<U> primitiveColumn = this as PrimitiveDataFrameColumn<U>;
+                        PrimitiveDataFrameColumn<U> newColumn = inPlace ? primitiveColumn : primitiveColumn.Clone();
+                        newColumn._columnContainer.Reverse<#=method.MethodName#>(value);
+                        return newColumn;
+                    }
+                    else
+                    {
+                        PrimitiveDataFrameColumn<decimal> clonedDecimalColumn = CloneAsDecimalColumn();
+                       clonedDecimalColumn._columnContainer.Reverse<#=method.MethodName#>(DecimalConverter<U>.Instance.GetDecimal(value));
```
eerhardt: (nit) white space doesn't line up here.

src/Microsoft.Data/PrimitiveDataFrameColumn.ReversedBinaryOperations.cs
```diff
@@ -0,0 +1,394 @@
+﻿
+// Licensed to the .NET Foundation under one or more agreements.
```
eerhardt: Oops - the rebase left these in the old folder.

pgovind: Yup, I see it. Fixing it

src/Microsoft.Data/DataFrame.BinaryOperators.cs
pgovind: We should throw an exception. Filed https://github.com/dotnet/corefxlab/issues/2776

PR #2770, opened 2019-10-29
Fix NullCount bugs

pgovind:
Fixes issues around NullCount when we encounter Arrow data(C# Arrow doesn't populate it's NullBitMapBuffer if NullCount is 0)

This patch contains half the changes from https://github.com/dotnet/corefxlab/pull/2730/.

### DIFF
--- a/src/Microsoft.Data/PrimitiveColumnContainer.cs
+++ b/src/Microsoft.Data/PrimitiveColumnContainer.cs
@@ -99,7 +99,7 @@ namespace Microsoft.Data
                 }
                 if (!buffer.IsEmpty)
                 {
-                    // Create a new bitMap with all the bits set
+                    // Create a new bitMap with all the bits upto length set
                     var bitMap = new byte[bitMapBufferLength];
                     bitMap.AsSpan().Slice(0, bitMapBufferLength - 1).Fill(255);
                     int lastByte = 1 << (length - (bitMapBufferLength - 1) * 8); 
--- a/src/Microsoft.Data/PrimitiveColumnContainer.cs
+++ b/src/Microsoft.Data/PrimitiveColumnContainer.cs
@@ -75,8 +75,47 @@ namespace Microsoft.Data
 
         public PrimitiveColumnContainer(ReadOnlyMemory<byte> buffer, ReadOnlyMemory<byte> nullBitMap, int length, int nullCount)
         {
-            Buffers.Add(new ReadOnlyDataFrameBuffer<T>(buffer, length));
-            NullBitMapBuffers.Add(new ReadOnlyDataFrameBuffer<byte>(nullBitMap, length));
+            ReadOnlyDataFrameBuffer<T> dataBuffer;
+            if (buffer.IsEmpty)
+            {
+                DataFrameBuffer<T> mutableBuffer = new DataFrameBuffer<T>();
+                mutableBuffer.EnsureCapacity(length);
+                mutableBuffer.Length = length;
+                mutableBuffer.RawSpan.Fill(default(T));
+                dataBuffer = mutableBuffer;
+            }
+            else
+            {
+                dataBuffer = new ReadOnlyDataFrameBuffer<T>(buffer, length);
+            }
+            Buffers.Add(dataBuffer);
+            int bitMapBufferLength = (length + 7) / 8;
+            ReadOnlyDataFrameBuffer<byte> nullDataFrameBuffer;
+            if (nullBitMap.IsEmpty)
+            {
+                if (nullCount != 0)
+                {
+                    throw new ArgumentNullException(Strings.InconsistentNullBitMapAndNullCount, nameof(nullBitMap));
+                }
+                if (!buffer.IsEmpty)
+                {
+                    // Create a new bitMap with all the bits set
+                    var bitMap = new byte[bitMapBufferLength];
+                    bitMap.AsSpan().Slice(0, bitMapBufferLength - 1).Fill(255);
+                    int lastByte = 1 << (length - (bitMapBufferLength - 1) * 8); 
+                    bitMap[bitMapBufferLength - 1] = (byte)(lastByte - 1);
+                    nullDataFrameBuffer = new DataFrameBuffer<byte>(bitMap, bitMapBufferLength);
+                }
+                else
+                {
+                    nullDataFrameBuffer = new DataFrameBuffer<byte>();
+                }
+            }
+            else
+            {
+                nullDataFrameBuffer = new ReadOnlyDataFrameBuffer<byte>(nullBitMap, bitMapBufferLength);
+            }
+            NullBitMapBuffers.Add(nullDataFrameBuffer);
             Length = length;
             NullCount = nullCount;
         }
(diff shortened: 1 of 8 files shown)


src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -75,8 +75,47 @@ public PrimitiveColumnContainer(IEnumerable<T?> values)
...
+            {
+                if (nullCount != 0)
+                {
+                    throw new ArgumentNullException(Strings.InconsistentNullBitMapAndNullCount, nameof(nullBitMap));
+                }
+                if (!buffer.IsEmpty)
+                {
+                    // Create a new bitMap with all the bits set
+                    var bitMap = new byte[bitMapBufferLength];
+                    bitMap.AsSpan().Slice(0, bitMapBufferLength - 1).Fill(255);
```
gfoidl: Even though the last byte is set just below, this will result in less code, and the end-result will be the same.
```suggestion
                    bitMap.AsSpan().Fill(0xFF);
```
Whether `0xFF` or `255` doesn't matter to me.

Aside: `bitMap.AsSpan(0, bitMapBufferLength - 1)` would do the same as the AsSpan and Slice in your code.

src/Microsoft.Data/strings.resx
```diff
@@ -132,6 +132,9 @@
   <data name="ImmutableColumn" xml:space="preserve">
     <value>Column is immutable</value>
   </data>
+  <data name="InconsistentNullBitMapAndNullCount" xml:space="preserve">
+    <value>Inconsisten Null bitmaps and NullCounts</value>
```
gfoidl:
```suggestion
    <value>Inconsistent Null bitmaps and NullCounts</value>
```

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Hmmm, I don't think so? `Slice` takes a start and a length. With `bitMapBufferLength - 1`, for an array of length 5, I'm filling the first 4 bytes. The next line sets an appropriate number of bits in the 5th byte. If I instead did `bitMap.AsSpan().Fill(255)`, that fills all the bytes, including all the bits in the last byte. 

As an aside: I'm trying to be conservative here. You're probably right that I can set all the bits, but I'm not 100% certain that it's safe. For ex: The set method looks up the existing bit and updates the `NullCount`. If all the bits are set, we _may_(not certain) update the `NullCount` wrongly.

gfoidl: > that fills all the bytes, including all the bits in the last byte.

Yep, but in L106 the last byte gets overwritten with the content of `lastByte`. So it doesn't matter which value was there before.
After L106 the content in `bitMap` is the correct one.

The difference is effiency of the code. 
We know that `bitMapBufferLength` is a multiple of 8, and we know that `Span.Fill` is optimized for blocks of length 2, 4, 8, 16 (vectorized), so that matches and `Span.Fill` will be as efficient as possible.
If we would just fill up to `bitMapBufferLength - 1`, then the last block can't be processed efficient, as it's length will be 7.
For larger bitmaps this won't show up, but for smaller sizes there may be a difference.

> not 100% certain that it's safe. For ex: The set method looks up the existing bit and updates the NullCount. If all the bits are set, we may(not certain) update the NullCount wrongly.

All the bits are set, but the next LOC overrides the contents of the last byte.
Let's say it different: with `new byte[]` no bits are set (zero-initialized), the we overwrite the content with actual values. What's the difference to initializing with all bits set?

One only has to take caution when binary operators like or, and, etc. are used, as there already bits set. But here it is an assignment...

If you don't feel confident with this, so change it at least to `bitMap.AsSpan(0, bitMapBufferLength - 1)` :wink:

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -122,14 +163,16 @@ public void Append(T? value)
...
+            ReadOnlyDataFrameBuffer<T> lastBuffer = Buffers[bufferIndex];
             if (lastBuffer.Length == ReadOnlyDataFrameBuffer<T>.MaxCapacity)
             {
                 lastBuffer = new DataFrameBuffer<T>();
                 Buffers.Add(lastBuffer);
                 NullBitMapBuffers.Add(new DataFrameBuffer<byte>());
             }
             DataFrameBuffer<T> mutableLastBuffer = DataFrameBuffer<T>.GetMutableBuffer(lastBuffer);
+            Buffers[bufferIndex] = mutableLastBuffer;
             mutableLastBuffer.Append(value ?? default);
```
eerhardt: Do you need to increment `NullCount` if `value` is null?

eerhardt:
>                     SetValidityBit(rowIndex, true);

Do you also need to adjust `NullCount` here?

---
Refers to: src/Microsoft.Data/PrimitiveColumnContainer.cs:420 in b209994. [](commit_id = b2099940694aaa10656792b14fbcce2bde891b2f, deletion_comment = False)

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -75,8 +75,47 @@ public PrimitiveColumnContainer(IEnumerable<T?> values)
...
+            ReadOnlyDataFrameBuffer<byte> nullDataFrameBuffer;
+            if (nullBitMap.IsEmpty)
+            {
+                if (nullCount != 0)
+                {
+                    throw new ArgumentNullException(Strings.InconsistentNullBitMapAndNullCount, nameof(nullBitMap));
+                }
+                if (!buffer.IsEmpty)
+                {
+                    // Create a new bitMap with all the bits upto length set
```
eerhardt: type-o `upto`

src/Microsoft.Data/PrimitiveColumnContainer.cs
```diff
@@ -75,8 +75,47 @@ public PrimitiveColumnContainer(IEnumerable<T?> values)
...
+                    nullDataFrameBuffer = new DataFrameBuffer<byte>(bitMap, bitMapBufferLength);
+                }
+                else
+                {
+                    nullDataFrameBuffer = new DataFrameBuffer<byte>();
+                }
+            }
+            else
+            {
+                nullDataFrameBuffer = new ReadOnlyDataFrameBuffer<byte>(nullBitMap, bitMapBufferLength);
```
eerhardt: Should we check if `nullBitMap` is at least `bitMapBufferLength` long?

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Ugh, sorry, you're absolutely right. Good call! For some reason I thought I was not overwriting the last bit. Fixed it :)

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: The next line `SetValidityBit` does it. This change is just updating `Buffers[bufferIndex]` with a mutable `DataFrameBuffer`

src/Microsoft.Data/PrimitiveColumnContainer.cs
pgovind: Yup. Good idea. Added a unit test too.

pgovind:
> Do you also need to adjust NullCount here

`SetValidityBit` updates the `NullCount` internally as part of it's book keeping.

PR #2771, opened 2019-10-30
Implement Append on DataFrame

pgovind:
Implements DataFrame.Append.

This is the second half of #2730 and also fixes one part of https://github.com/dotnet/corefxlab/issues/2719

### DIFF
--- a/src/Microsoft.Data/DataFrame.cs
+++ b/src/Microsoft.Data/DataFrame.cs
@@ -424,7 +424,7 @@ namespace Microsoft.Data
                 // Go through row first to make sure there are no data type incompatibilities
                 IEnumerator<object> rowEnumerator = row.GetEnumerator();
                 bool rowMoveNext = rowEnumerator.MoveNext();
-                Dictionary<string, object> cachedObjectConversions = new Dictionary<string, object>();
+                List<object> cachedObjectConversions = new List<object>();
                 while (columnMoveNext && rowMoveNext)
                 {
                     DataFrameColumn column = columnEnumerator.Current;
@@ -437,7 +437,7 @@ namespace Microsoft.Data
                             throw new ArgumentException(string.Format(Strings.MismatchedValueType, column.DataType), value.GetType().ToString());
                         }
                     }
-                    cachedObjectConversions.Add(column.Name, value);
+                    cachedObjectConversions.Add(value);
                     columnMoveNext = columnEnumerator.MoveNext();
                     rowMoveNext = rowEnumerator.MoveNext();
                 }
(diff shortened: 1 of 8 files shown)


src/Microsoft.Data/DataFrame.cs
```diff
@@ -402,6 +402,127 @@ public DataFrame FillNulls(IList<object> values, bool inPlace = false)
...
+        public void Append(IEnumerable<object> row = null)
+        {
+            IEnumerator<DataFrameColumn> columnEnumerator = Columns.GetEnumerator();
+            bool columnMoveNext = columnEnumerator.MoveNext();
+            if (row != null)
+            {
+                // Go through row first to make sure there are no data type incompatibilities
+                IEnumerator<object> rowEnumerator = row.GetEnumerator();
+                bool rowMoveNext = rowEnumerator.MoveNext();
+                while (columnMoveNext && rowMoveNext)
```
pgovind: Unfortunately I have to loop over `row` twice to make sure `DataFrameColumn.Set` doesn't throw later with a type mismatch. Without this check upfront, half the columns may have appended a value before an exception is thrown.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -402,6 +402,131 @@ public DataFrame FillNulls(IList<object> values, bool inPlace = false)
...
+                {
+                    DataFrameColumn column = columnEnumerator.Current;
+                    object value = rowEnumerator.Current;
+                    if (value != null)
+                    {
+                        value = Convert.ChangeType(value, column.DataType);
```
eerhardt: Do you think it would make sense to cache this conversion? That way we wouldn't have to do the conversion twice.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -402,6 +402,131 @@ public DataFrame FillNulls(IList<object> values, bool inPlace = false)
...
+                        {
+                            throw new ArgumentException(string.Format(Strings.MismatchedValueType, column.DataType), value.GetType().ToString());
+                        }
+                    }
+                    columnMoveNext = columnEnumerator.MoveNext();
+                    rowMoveNext = rowEnumerator.MoveNext();
+                }
+                if (rowMoveNext)
+                {
+                    throw new ArgumentException(string.Format(Strings.ExceedsNumberOfColumns, Columns.Count), nameof(row));
```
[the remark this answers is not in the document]
pgovind: Already doing it in TestAppendRow :)

src/Microsoft.Data/DataFrame.cs
pgovind: Done. I wasn't sure how caching and a dictionary lookup compares to a `ChangeType` call initially.

src/Microsoft.Data/DataFrame.cs
```diff
@@ -402,6 +402,129 @@ public DataFrame FillNulls(IList<object> values, bool inPlace = false)
...
+        public void Append(IEnumerable<object> row = null)
+        {
+            IEnumerator<DataFrameColumn> columnEnumerator = Columns.GetEnumerator();
+            bool columnMoveNext = columnEnumerator.MoveNext();
+            if (row != null)
+            {
+                // Go through row first to make sure there are no data type incompatibilities
+                IEnumerator<object> rowEnumerator = row.GetEnumerator();
+                bool rowMoveNext = rowEnumerator.MoveNext();
+                Dictionary<string, object> cachedObjectConversions = new Dictionary<string, object>();
```
eerhardt: You don't need a dictionary here. You just need a `List<object>`. they will be in the right order.

pgovind: Right right, I'm not done yet. I just rebased onto current upstream/master in the middle of doing this

PR #2772, opened 2019-11-01
Add ElementwiseEquals for ArrowStringDataFrameColumn

pgovind:
Fix for https://github.com/dotnet/corefxlab/issues/2754

### DIFF
--- a/src/Microsoft.Data/ArrowStringDataFrameColumn.cs
+++ b/src/Microsoft.Data/ArrowStringDataFrameColumn.cs
@@ -472,7 +472,7 @@ namespace Microsoft.Data
 
         public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
         {
-            ReadOnlySpan<byte> bytes = Encoding.UTF8.GetBytes(value);
+            ReadOnlySpan<byte> bytes = value != null ? Encoding.UTF8.GetBytes(value) : default(ReadOnlySpan<byte>);
             PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(Name, Length);
             for (long i = 0; i < Length; i++)
             {
@@ -508,7 +508,7 @@ namespace Microsoft.Data
 
         public PrimitiveDataFrameColumn<bool> ElementwiseNotEquals(string value)
         {
-            ReadOnlySpan<byte> bytes = Encoding.UTF8.GetBytes(value);
+            ReadOnlySpan<byte> bytes = value != null ? Encoding.UTF8.GetBytes(value) : default(ReadOnlySpan<byte>);
             PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(Name, Length);
             for (long i = 0; i < Length; i++)
             {
--- a/src/Microsoft.Data/ArrowStringDataFrameColumn.cs
+++ b/src/Microsoft.Data/ArrowStringDataFrameColumn.cs
@@ -469,5 +469,77 @@ namespace Microsoft.Data
 
         private ValueGetter<ReadOnlyMemory<char>> CreateValueGetterDelegate(DataViewRowCursor cursor) =>
             (ref ReadOnlyMemory<char> value) => value = this[cursor.Position].AsMemory();
+
+        public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
+        {
+            ReadOnlySpan<byte> bytes = Encoding.UTF8.GetBytes(value);
+            PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(Name, Length);
+            for (long i = 0; i < Length; i++)
+            {
+                var strBytes = GetBytes(i);
+                if (strBytes.Length != bytes.Length)
+                {
+                    ret[i] = false;
+                    continue;
+                }
+                bool isEqual = true;
+                for (int j = 0; j < bytes.Length; j++)
+                {
+                    isEqual = strBytes[j] == bytes[j];
+                    if (!isEqual)
+                    {
+                        break;
+                    }
+                }
+                ret[i] = isEqual;
+            }
+            return ret;
+        }
+
+        public override PrimitiveDataFrameColumn<bool> ElementwiseEquals<T>(T value)
+        {
+            return ElementwiseEquals(value.ToString());
+        }
+
+        public override PrimitiveDataFrameColumn<bool> ElementwiseEquals(DataFrameColumn column)
+        {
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data/ArrowStringDataFrameColumn.cs
```diff
@@ -469,5 +469,77 @@ protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
...
             (ref ReadOnlyMemory<char> value) => value = this[cursor.Position].AsMemory();
+
+        public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
+        {
+            ReadOnlySpan<byte> bytes = value != null ? Encoding.UTF8.GetBytes(value) : default(ReadOnlySpan<byte>);
+            PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(Name, Length);
+            for (long i = 0; i < Length; i++)
+            {
+                var strBytes = GetBytes(i);
+                if (strBytes.Length != bytes.Length)
```
pgovind: Unfortunately, ReadOnlySpan<T> operator == does not do an element wise value comparison

eerhardt: Use `SequenceEquals` instead.

---
In reply to: [341730642](https://github.com/dotnet/corefxlab/pull/2772#discussion_r341730642) [](ancestors = 341730642)

src/Microsoft.Data/ArrowStringDataFrameColumn.cs
```diff
@@ -469,5 +469,77 @@ protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
...
+        }
+
+        public override PrimitiveDataFrameColumn<bool> ElementwiseEquals<T>(T value)
+        {
+            return ElementwiseEquals(value.ToString());
+        }
+
+        public override PrimitiveDataFrameColumn<bool> ElementwiseEquals(DataFrameColumn column)
+        {
+            return StringDataFrameColumn.ElementwiseEqualsImplementation(this, column);
```
eerhardt: This is going to convert every string to a UTF16 string, right? It may be fine for now to get the functionality working, but I think we should log an issue to avoid this when possible. For example, if I call `ElementwiseEquals` on 2 Arrow String columns.

src/Microsoft.Data/ArrowStringDataFrameColumn.cs
```diff
@@ -469,5 +469,77 @@ protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
 
         private ValueGetter<ReadOnlyMemory<char>> CreateValueGetterDelegate(DataViewRowCursor cursor) =>
             (ref ReadOnlyMemory<char> value) => value = this[cursor.Position].AsMemory();
+
+        public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
```
eerhardt: We should have XML doc comments for all public APIs.

src/Microsoft.Data/ArrowStringDataFrameColumn.cs
```diff
@@ -469,5 +469,77 @@ protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
...
+        {
+            ReadOnlySpan<byte> bytes = value != null ? Encoding.UTF8.GetBytes(value) : default(ReadOnlySpan<byte>);
```
eerhardt: If `value` is `null`, we want to check the `null` bit map, right?

tests/Microsoft.Data.DataFrame.Tests/DataFrameTests.cs
```diff
@@ -422,6 +422,55 @@ public void TestBinaryOperationsOnBoolColumn()
...
+
+            ArrowStringDataFrameColumn stringColumn = new ArrowStringDataFrameColumn("String", strArray.ValueBuffer.Memory, strArray.ValueOffsetsBuffer.Memory, strArray.NullBitmapBuffer.Memory, strArray.Length, strArray.NullCount);
+            df.Columns.Insert(0, stringColumn);
+
+            DataFrameColumn newCol = stringColumn.ElementwiseEquals(4);
+            Assert.Equal(true, newCol[4]);
+            Assert.Equal(false, newCol[0]);
+            Assert.Equal(false, newCol[5]);
+
+            newCol = (stringColumn as ArrowStringDataFrameColumn).ElementwiseEquals("4");
```
eerhardt: `stringColumn` is already casted as a `ArrowStringDataFrameColumn`. Why the extra cast here?

src/Microsoft.Data/ArrowStringDataFrameColumn.cs
pgovind: Right. I'll log an issue to add an API that takes in an ArrowStringDataFrameColumn as a parameter to avoid the UTF conversions

pgovind: https://github.com/dotnet/corefxlab/issues/2775

src/Microsoft.Data/ArrowStringDataFrameColumn.cs
```diff
@@ -469,5 +469,55 @@ protected internal override Delegate GetDataViewGetter(DataViewRowCursor cursor)
...
+        /// Returns a boolean column that is the result of an elementwise equality comparison of each value in the column with <paramref name="value"/>
+        /// </summary>
+        public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
+        {
+            ReadOnlySpan<byte> bytes = value != null ? Encoding.UTF8.GetBytes(value) : default(ReadOnlySpan<byte>);
+            PrimitiveDataFrameColumn<bool> ret = new PrimitiveDataFrameColumn<bool>(Name, Length);
+            for (long i = 0; i < Length; i++)
+            {
+                var strBytes = GetBytes(i);
+                ret[i] = value == null ? !IsValid(i) : strBytes.SequenceEqual(bytes);
```
eerhardt: I would re-write this with 2 separate loops. There is no need to call `GetBytes(i)` when value is null. And there is no need to keep re-checking if value is null inside of the loop.

```C#
if (value == null)
{
    for (long i = 0; i < Length; i++)
    {
        ret[i] = !IsValid(i);
    }
}
else
{
    ReadOnlySpan<byte> bytes = Encoding.UTF8.GetBytes(value);
    for (long i = 0; i < Length; i++)
    {
        var strBytes = GetBytes(i);
        ret[i] = strBytes.SequenceEqual(bytes);
    }
}
```

PR #2778, opened 2019-11-08
Throw for incompatible inPlace

pgovind:
Binary operations throw when types don't match and `inPlace` is set 

Fixes https://github.com/dotnet/corefxlab/issues/2749


tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -645,6 +645,16 @@ public void TestBinaryOperatorsOnBoolColumns()
             }
         }
 
+        [Fact]
+        public void TestBinaryOperationsOnColumns()
+        {
+            PrimitiveDataFrameColumn<int> column = new PrimitiveDataFrameColumn<int>("Int", Enumerable.Range(0, 10));
+            Assert.ThrowsAny<ArgumentException>(() => column.Add(5.5, inPlace: true));
+            Assert.ThrowsAny<ArgumentException>(() => column.ReverseAdd(5.5, inPlace: true));
+        }
```
eerhardt: (nit) only need 1 blank line between methods.

src/Microsoft.Data.Analysis/strings.resx
```diff
@@ -153,6 +153,9 @@
   <data name="MismatchedColumnLengths" xml:space="preserve">
     <value>Column lengths are mismatched</value>
   </data>
+  <data name="MismatchedColumnValueType" xml:space="preserve">
+    <value>Expected column to hold values of type {0}</value>
```
eerhardt: I think this message could use some work.

```C#
PrimitiveDataFrameColumn<int> column = new PrimitiveDataFrameColumn<int>("Int", Enumerable.Range(0, 10));
column.Add(5.5, inPlace: true);
```

Throwing a message saying:

`Expected column to hold values of type System.Single`

might not make sense to users.

pgovind: Your example would use Strings.MismatchedValueType (`Expected value to be of type System.Single`) :). `Strings.MismatchedColumnValueType` is used when you do `column.Add(floatColumn, inPlace: true)`

eerhardt: makes sense.

Issue #2784, opened 2019-11-13
Add DateTime supports and tests to DataFrame


eerhardt:
We currently don't have explicit tests for `DateTime` columns. We should add some tests and support in areas, ex. Arrow and IDataView integrations.

pgovind:
Got to fix this ASAP. The `Description` method fails with a `DateTime` column because it tries to calculate `DateTimeColumn.Max()`

PR #2786, opened 2019-11-21
Fix the Description method to not throw

pgovind:
Adds an Info method

The Description method was always calling `Max` on `PrimitiveDataFrameColumn<T>`. If `T` was a struct (such as `DateTime` where `Max` may not be well defined), we return `null` instead of throwing.

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -204,7 +204,6 @@ namespace Microsoft.Data.Analysis
             return new DataFrame(newColumns);
         }
 
-
         /// <summary>
         /// Generates a concise summary of each column in the DataFrame
         /// </summary>
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -204,37 +204,54 @@ namespace Microsoft.Data.Analysis
             return new DataFrame(newColumns);
         }
 
+
+        /// <summary>
+        /// Generates a concise summary of each column in the DataFrame
+        /// </summary>
+        public DataFrame Info()
+        {
+            DataFrame ret = new DataFrame();
+
+            for (int i = 0; i < Columns.Count; i++)
+            {
+                DataFrameColumn column = Columns[i];
+                if (i == 0)
+                {
+                    StringDataFrameColumn strColumn = new StringDataFrameColumn("Info", 2);
+                    strColumn[0] = "DataType";
+                    strColumn[1] = Strings.DescriptionMethodLength;
+                    ret.Columns.Add(strColumn);
+                }
+                column.Info(ret);
+            }
+            return ret;
+        }
+
         /// <summary>
         /// Generates descriptive statistics that summarize each numeric column
         /// </summary>
         public DataFrame Description()
         {
             DataFrame ret = new DataFrame();
-            if (Columns.Count == 0)
-                return ret;
-            int i = 0;
-            while (!Columns[i].HasDescription())
+
+            bool firstDescriptionColumn = true;
+            foreach (DataFrameColumn column in Columns)
             {
-                i++;
-            }
-            ret = Columns[i].Description();
-            i++;
-            for (; i < Columns.Count; i++)
-            {
-                DataFrameColumn column = Columns[i];
                 if (!column.HasDescription())
                 {
                     continue;
                 }
-                DataFrame columnDescription = column.Description();
-                ret = ret.Merge<string>(columnDescription, "Description", "Description", "_left", "_right", JoinAlgorithm.Inner);
-                int leftMergeColumn = ret._columnCollection.IndexOf("Description" + "_left");
-                int rightMergeColumn = ret._columnCollection.IndexOf("Description" + "_right");
-                if (leftMergeColumn != -1 && rightMergeColumn != -1)
+                if (firstDescriptionColumn)
                 {
-                    ret.Columns.Remove("Description" + "_right");
-                    ret._columnCollection.SetColumnName(ret["Description_left"], "Description");
+                    firstDescriptionColumn = false;
+                    StringDataFrameColumn stringColumn = new StringDataFrameColumn("Description", 0);
+                    stringColumn.Append(Strings.DescriptionMethodLength);
+                    stringColumn.Append("Max");
+                    stringColumn.Append("Min");
+                    stringColumn.Append("Mean");
+                    ret.Columns.Add(stringColumn);
                 }
+                column.Description(ret);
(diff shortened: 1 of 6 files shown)


src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -204,37 +204,54 @@ private DataFrame Clone(DataFrameColumn mapIndices = null, bool invertMapIndices
             return new DataFrame(newColumns);
         }
 
+
```
pgovind: Extra line

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -5,9 +5,12 @@
 using System;
 using System.Collections.Generic;
 using System.Linq;
+using System.Runtime.InteropServices.ComTypes;
```
pgovind: Remove these

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1678,6 +1687,42 @@ public void TestDescription()
                 Assert.Equal((float)0, column[2]);
                 Assert.Equal((float)4, column[3]);
             }
+
+            // Explicitly check the dateTimes column
+            DataFrameColumn dateTimeColumn = description.Columns[description.Columns.Count - 1];
+            Assert.Equal(dateTimeColumn.Name, dateTimes.Name);
+            Assert.Equal(4, dateTimeColumn.Length);
+            Assert.Equal((float)10, dateTimeColumn[0]);
+            Assert.Null(dateTimeColumn[1]);
```
eerhardt: Why aren't we displaying `Max` and `Min` for DateTime?

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
```diff
@@ -551,25 +551,41 @@ protected override DataFrameColumn FilterImplementation<U>(U lower, U upper)
...
-            DataFrame ret = new DataFrame();
-            StringDataFrameColumn stringColumn = new StringDataFrameColumn("Description", 0);
-            stringColumn.Append("Length");
-            stringColumn.Append("Max");
-            stringColumn.Append("Min");
-            stringColumn.Append("Mean");
-            float max = (float)Convert.ChangeType(Max(), typeof(float));
-            float min = (float)Convert.ChangeType(Min(), typeof(float));
-            float mean = (float)Convert.ChangeType(Sum(), typeof(float)) / Length;
+        public override void Description(DataFrame dataFrame)
```
eerhardt: Why not have this method return a column instead? And then the caller can add it to a DataFrame or do whatever they want with it.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -205,36 +205,52 @@ private DataFrame Clone(DataFrameColumn mapIndices = null, bool invertMapIndices
...
+        }
+
+        /// <summary>
+        /// Generates descriptive statistics that summarize each numeric column
+        /// </summary>
+        public DataFrame Description()
+        {
+            DataFrame ret = new DataFrame();
+
+            bool firstDescriptionColumn = true;
```
eerhardt: (minor) why the difference in patterns here between Description and Info? Above you use a `for` loop and check for `i==0`. But here you use a `foreach` and keep track of a `firstDescriptionColumn` bool?

pgovind: Because I wrote them on different days :). It doesn't really mean anything. If I'd written `Description` first and then `Info`, I probably would've used the same pattern.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: Because I don't have a great answer to "does `Min`/`Max` make sense for `DateTime`"? Couple reasons:
1. I know we can convert each `DateTime` value to `ticks` and find the `Min`/`Max` that way, but is that the generally accepted method everywhere? I figured users can run a LINQ query if they really want to find the max/min or write their own for loop. 
2. `Max` and `Min` are part of `DataFrameColumn.Computations`. There's other API here such as `Sum`, `Product` etc. I didn't think they made sense for `DateTime` values.

We can revisit this as part of #2784 . For now, this let's users call `Description` without throwing an exception so they can see other useful information about the `DataFrame`.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
pgovind: Done.

src/Microsoft.Data.Analysis/DataFrame.cs
pgovind: Fixed it with the latest commit

pgovind:
I'm going to spend the rest of the afternoon working on my blog post. I put up this PR so I can use the `Description` method in the samples. What do you think about making a 0.2.0 package once this goes in? I can use `Description` in the sample notebook then.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -205,41 +205,57 @@ private DataFrame Clone(DataFrameColumn mapIndices = null, bool invertMapIndices
...
+
+            bool firstColumn = true;
+            foreach (DataFrameColumn column in Columns)
             {
-                i++;
+                if (firstColumn)
+                {
+                    firstColumn = false;
+                    StringDataFrameColumn strColumn = new StringDataFrameColumn("Info", 2);
+                    strColumn[0] = "DataType";
```
eerhardt: Why does "Length" go in a .resx but `DataType` doesn't?

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -185,9 +185,20 @@ public virtual DataFrameColumn Sort(bool ascending = true)
...
+            StringDataFrameColumn dataColumn = new StringDataFrameColumn(Name, 2);
+            dataColumn[0] = DataType.ToString();
+            dataColumn[1] = (Length - NullCount).ToString();
+            return dataColumn;
+        }
+
+        /// <summary>
+        /// Returns a <see cref="PrimitiveDataFrameColumn{T}"/> with statistics that describe the column
+        /// </summary>
+        public virtual PrimitiveDataFrameColumn<float> Description() => throw new NotImplementedException();
```
eerhardt: I wonder if the signature shouldn't be `DataFrameColumn` instead of a concrete type.

PR #2790, opened 2019-11-25
Enable xml docs for Data.Analysis

pgovind:


src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -10,6 +10,9 @@
     <Description>This package contains easy-to-use and high-performance libraries for data analysis and transformation.</Description>
     <PackageReleaseNotes>Initial preview of robust and extensible types and algorithms for manipulating structured data that supports aggregations, statistical funtions, sorting, grouping, joins, merges, handling missing values and more. </PackageReleaseNotes>
     <PackageTags>ML.NET ML Machine Learning Data Science DataFrame Preparation DataView Analytics Exploration</PackageTags>
+    <GenerateDocumentationFile>true</GenerateDocumentationFile>
+    <!-- Documentation warnings -->
+    <NoWarn>$(NoWarn);1591</NoWarn>
```
pgovind: I haven't written documentation for public methods that override base class methods because the base class methods have documentation on them. Suppressing 1591 is not ideal, this disables warnings on all public methods, but I verified that only the overriden methods produce this error. I could open an issue to get rid of this suppresion later. For now, this will let us ship a 0.2.0 with all the documentation we already have.

ahsonkhan: Could you use `#pragma disable warning` only around the affected areas rather than applying it globally to the whole project?

Also, how much effort is it to document the base classes too?

@carlossanlop - any suggestions for the docs warning with base classes? Do we know if there's a solution where the default comments will get used if one isn't provided?

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -42,19 +42,36 @@ public DataFrame(IList<DataFrameColumn> columns)
...
 
         internal IReadOnlyList<string> GetColumnNames() => _columnCollection.GetColumnNames();
 
         #region Operators
+        /// <summary>
+        /// Indexer to get/set values
+        /// </summary>
+        /// <param name="rowIndex"></param>
+        /// <param name="columnIndex"></param>
+        /// <returns></returns>
```
ahsonkhan: Left empty intentionally?

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -111,10 +112,11 @@ public virtual DataFrameColumn Sort(bool ascending = true)
         public virtual GroupBy GroupBy(int columnIndex, DataFrame parent) => throw new NotImplementedException();
 
         /// <summary>
-        /// Returns a new column with nulls replaced by value
+        /// Returns a new column with nulls replaced by <paramref name="value"/>
         /// </summary>
         /// <remarks>Tries to convert value to the column's DataType</remarks>
         /// <param name="value"></param>
+        /// <param name="inPlace">Indicates if the operation should be performed in place</param>
```
ahsonkhan: Any exceptions?

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
carlossanlop: If the base class already has documentation, but you are going to override a method, then you should copy the comments for the base method and make the necessary changes to adapt it to this class that overrides it.

For example, Stream has basic documentation of what we expect a method to do, and the classes that inherit from Stream and override the base methods also have documentation, very similar but with small adaptations to describe the functionality of the method specifically for that class.

> _Do we know if there's a solution where the default comments will get used if one isn't provided?_

You could also use `inheritdoc`, which may get rid of the warning, although I am not entirely sure that it works (meaning, I don't think the comments get inherited when we build and generate the documentation xmls, it may be a bug on our side).

pgovind: I'll look into `inheritdoc` tomorrow. I made sure that all the base class methods have documentation on them. There are some template generated methods that I can cheaply get documentation on. With these 2 steps, I'm guessing most of the errors will go away.

src/Microsoft.Data.Analysis/DataFrame.cs
pgovind: No. I should delete the returns tag here. The indexer returns different things based on if its a set or a get method.

src/Microsoft.Data.Analysis/DataFrameColumn.cs
pgovind: Oh, there are. There's undocumented exceptions in a bunch of other methods too. I wonder if there's a static analysis tool that can look at all possible code paths and generate all possible exceptions in the documentation.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
eerhardt: `inheritdoc` is your friend here.

> Suppressing 1591 is not ideal, this disables warnings on all public methods

We shouldn't suppress 1591 globally because any new undocumented public member won't cause an error.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -42,19 +42,36 @@ public DataFrame(IList<DataFrameColumn> columns)
             _columnCollection = new DataFrameColumnCollection(columns, OnColumnsChanged);
         }
 
+        /// <summary>
+        /// Returns the number of rows in this <seealso cref="DataFrame"/>
```
eerhardt: `seealso cref` => `see cref`. This probably applies everywhere.

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -133,10 +133,10 @@ public IEnumerable<ReadOnlyMemory<byte>> GetReadOnlyDataBuffers()
         }
 
         /// <summary>
-        /// Returns an enumerable of immutable ReadOnlyMemory<byte> buffers representing null values in the Apache Arrow format
+        /// Returns an enumerable of immutable <seealso cref="ReadOnlyMemory{Byte}"/> buffers representing null values in the Apache Arrow format
```
carlossanlop:
```suggestion
        /// Returns an enumerable of immutable <see cref="ReadOnlyMemory{Byte}"/> buffers representing null values in the Apache Arrow format
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -133,10 +133,10 @@ public IEnumerable<ReadOnlyMemory<byte>> GetReadOnlyDataBuffers()
...
         /// </summary>
-        /// <remarks>Each ReadOnlyMemory<byte> encodes the indices of null values in its corresponding Data buffer</remarks>
-        /// <returns>IEnumerable<ReadOnlyMemory<byte>></returns>
+        /// <remarks>Each <seealso cref="ReadOnlyMemory{Byte}"/> encodes the indices of null values in its corresponding Data buffer</remarks>
```
carlossanlop:
```suggestion
        /// <remarks>Each <see cref="ReadOnlyMemory{Byte}"/> encodes the indices of null values in its corresponding Data buffer</remarks>
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -122,7 +122,7 @@ private void SetValidityBit(long index, bool value)
         /// </summary>
         /// <remarks>Null values are encoded in the buffers returned by GetReadOnlyNullBitmapBuffers in the Apache Arrow format</remarks>
         /// <remarks>The offsets buffers returned by GetReadOnlyOffsetBuffers can be used to delineate each value</remarks>
-        /// <returns>IEnumerable<ReadOnlyMemory<byte>></returns>
+        /// <returns>IEnumerable<see cref="ReadOnlyMemory{Byte}"/></returns>
```
carlossanlop:
```suggestion
        /// <returns>An enumeration of <see cref="ReadOnlyMemory{Byte}"/> objects.</returns>
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -122,7 +122,7 @@ private void SetValidityBit(long index, bool value)
         /// </summary>
         /// <remarks>Null values are encoded in the buffers returned by GetReadOnlyNullBitmapBuffers in the Apache Arrow format</remarks>
```
carlossanlop:
```suggestion
        /// <remarks><see langword="null" /> values are encoded in the buffers returned by GetReadOnlyNullBitmapBuffers in the Apache Arrow format.</remarks>
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -122,7 +122,7 @@ private void SetValidityBit(long index, bool value)
...
         /// <remarks>The offsets buffers returned by GetReadOnlyOffsetBuffers can be used to delineate each value</remarks>
```
carlossanlop:
```suggestion
        /// <remarks>The offsets buffers returned by GetReadOnlyOffsetBuffers can be used to delineate each value.</remarks>
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -133,10 +133,10 @@ public IEnumerable<ReadOnlyMemory<byte>> GetReadOnlyDataBuffers()
         }
 
         /// <summary>
-        /// Returns an enumerable of immutable ReadOnlyMemory<byte> buffers representing null values in the Apache Arrow format
+        /// Returns an enumerable of immutable <seealso cref="ReadOnlyMemory{Byte}"/> buffers representing null values in the Apache Arrow format
         /// </summary>
-        /// <remarks>Each ReadOnlyMemory<byte> encodes the indices of null values in its corresponding Data buffer</remarks>
-        /// <returns>IEnumerable<ReadOnlyMemory<byte>></returns>
+        /// <remarks>Each <seealso cref="ReadOnlyMemory{Byte}"/> encodes the indices of null values in its corresponding Data buffer</remarks>
+        /// <returns>IEnumerable<see cref="ReadOnlyMemory{Byte}"/></returns>
```
carlossanlop:
```suggestion
        /// <returns>An enumeration of <see cref="ReadOnlyMemory{Byte}"/> objects.</returns>
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -147,10 +147,10 @@ public IEnumerable<ReadOnlyMemory<byte>> GetReadOnlyNullBitMapBuffers()
         }
 
         /// <summary>
-        /// Returns an enumerable of immutable ReadOnlyMemory<int> representing offsets into its corresponding Data buffer.
+        /// Returns an enumerable of immutable <see cref="ReadOnlyMemory{Int32}"/> representing offsets into its corresponding Data buffer.
         /// The Apache Arrow format specifies how the offset buffer encodes the length of each value in the Data buffer
         /// </summary>
-        /// <returns>IEnumerable<ReadOnlyMemory<int>></returns>
+        /// <returns>IEnumerable<see cref="ReadOnlyMemory{Int32}"/></returns>
```
carlossanlop:
```suggestion
        /// <returns>An enumeration of  <see cref="ReadOnlyMemory{Int32}"/> objects.</returns>
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -147,10 +147,10 @@ public IEnumerable<ReadOnlyMemory<byte>> GetReadOnlyNullBitMapBuffers()
         }
 
         /// <summary>
-        /// Returns an enumerable of immutable ReadOnlyMemory<int> representing offsets into its corresponding Data buffer.
+        /// Returns an enumerable of immutable <see cref="ReadOnlyMemory{Int32}"/> representing offsets into its corresponding Data buffer.
```
carlossanlop: I've seen "enumeration" used more than "enumerable":
```suggestion
        /// Returns an enumeration of immutable <see cref="ReadOnlyMemory{Int32}"/> representing offsets into its corresponding Data buffer.
```

src/Microsoft.Data.Analysis/DataFrame.cs
carlossanlop:
```suggestion
        /// Returns the number of rows in this <see cref="DataFrame"/>.
```

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -42,19 +42,36 @@ public DataFrame(IList<DataFrameColumn> columns)
...
+        /// </summary>
         public long RowCount => _columnCollection.RowCount;
 
+        /// <summary>
+        /// Returns the columns contained in the <seealso cref="DataFrame"/> as a <seealso cref="DataFrameColumnCollection"/>
```
carlossanlop:
```suggestion
        /// Returns the columns contained in the <see cref="DataFrame"/> as a <see cref="DataFrameColumnCollection"/>.
```

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -42,19 +42,36 @@ public DataFrame(IList<DataFrameColumn> columns)
...
+        /// </summary>
         public DataFrameColumnCollection Columns => _columnCollection;
 
         internal IReadOnlyList<string> GetColumnNames() => _columnCollection.GetColumnNames();
 
         #region Operators
+        /// <summary>
+        /// Indexer to get/set values
```
carlossanlop:
```suggestion
        /// An indexer to get or set values.
```

src/Microsoft.Data.Analysis/DataFrame.cs
carlossanlop: +1. Please fill out the two params and the returns.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -42,19 +42,36 @@ public DataFrame(IList<DataFrameColumn> columns)
...
+        /// <param name="columnIndex"></param>
+        /// <returns></returns>
         public object this[long rowIndex, int columnIndex]
         {
             get => _columnCollection[columnIndex][rowIndex];
             set => _columnCollection[columnIndex][rowIndex] = value;
         }
 
+        /// <summary>
+        /// Returns a row in this <seealso cref="DataFrame"/>
```
carlossanlop:
```suggestion
        /// Returns a row in this <seealso cref="DataFrame"/>.
```

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -111,10 +112,11 @@ public virtual DataFrameColumn Sort(bool ascending = true)
         public virtual GroupBy GroupBy(int columnIndex, DataFrame parent) => throw new NotImplementedException();
 
         /// <summary>
-        /// Returns a new column with nulls replaced by value
+        /// Returns a new column with nulls replaced by <paramref name="value"/>
```
carlossanlop:
```suggestion
        /// Returns a new column with <see langword="null" /> elements replaced by <paramref name="value"/>.
```

carlossanlop:
Update: I just noticed this change is in corefxlab. @eerhardt @ahsonkhan how strict do we want to be with the documentation rules here?

ahsonkhan:
> how strict do we want to be with the documentation rules here?

Depends on what the end goal/shipping vehicle/quality of these types is, and what stage of development we are at (still prototyping, or API/ship ready?), which I don't have context on. Up to @eerhardt, @pgovind to chime in for that (along with what the bar should be/when should that bar be applied). If the goal is to use these XML comments to generate the docs on the API docs website at some point, and the APIs are relatively stable, then we probably should apply some scrutiny on quality of comments.

ahsonkhan:
> - Use `<see cref="..." />` instead of "<seealso="..." />".

As an aside, do, `seealso` tags ever show up in docs? I had used them in S.T.Json intentionally (where applicable) in hopes to get a `See also section` on the site:
https://github.com/dotnet/corefx/search?q=seealso&unscoped_q=seealso

Like:
https://github.com/dotnet/corefx/blob/21fd4d2a70447064aacb6536278f6e8cc1c9e2d5/src/System.Text.Json/src/System/Text/Json/Reader/Utf8JsonReader.TryGet.cs#L23-L26

Do we need to do some extra/manual work to make that happen?

pgovind:
> how strict do we want to be with the documentation rules here?

I don't think we want to be too strict with the documentation yet here. This is still a preview package. I went ahead and brought down the number of APIs with missing docs from 500+ to 100+ though. Template generation and `inheritdoc` helped. For a 0.2.0 preview, I think we have enough documentation with this patch. Let's see what @eerhardt thinks.

eerhardt:
> how strict do we want to be with the documentation rules here?

I think the bar is lower in this package than it is in corefx, as this is a "preview" package. But I don't think it isn't that much lower. We don't want to build up debt here by not doing the correct documentation up-front.

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -53,11 +66,19 @@ public ArrowStringDataFrameColumn(string name, ReadOnlyMemory<byte> values, Read
         }
 
         private long _nullCount;
+        /// <summary>
+        /// <inheritdoc/>
```
eerhardt: I don't think you need `<summary>` around these. I think you simply say:

```C#
/// <inheritdoc/>
public override long NullCount => _nullCount;
```

See [https://github.com/aspnet/Scaffolding/blob/9efe0f7e05e51bb43c878be49087bdba7456e36d/src/VS.Web.CG.Contracts/ProjectModel/CommonProjectContext.cs#L10](https://github.com/aspnet/Scaffolding/blob/9efe0f7e05e51bb43c878be49087bdba7456e36d/src/VS.Web.CG.Contracts/ProjectModel/CommonProjectContext.cs#L10) for an example.

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -118,11 +139,11 @@ private void SetValidityBit(long index, bool value)
...
 
         /// <summary>
-        /// Returns an enumerable of immutable buffers representing the underlying values in the Apache Arrow format
+        /// Returns an enumeration of immutable buffers representing the underlying values in the Apache Arrow format
         /// </summary>
-        /// <remarks>Null values are encoded in the buffers returned by GetReadOnlyNullBitmapBuffers in the Apache Arrow format</remarks>
+        /// <remarks><see langword="null" /> values are encoded in the buffers returned by GetReadOnlyNullBitmapBuffers in the Apache Arrow format</remarks>
         /// <remarks>The offsets buffers returned by GetReadOnlyOffsetBuffers can be used to delineate each value</remarks>
-        /// <returns>IEnumerable<ReadOnlyMemory<byte>></returns>
+        /// <returns>An enumeration of <see cref="ReadOnlyMemory{Byte}"/> objects</returns>
```
eerhardt: In the Linq docs, we don't use this pattern.

[https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.select?view=netcore-3.0](https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.select?view=netcore-3.0)

```
Returns
IEnumerable<TResult>
An IEnumerable<T> whose elements are the result of invoking the transform function on each element of source.
```

Should we follow these established patterns?

Also, it would be nice to say what these object represent. Like in the link above `whose elements are the result of invoking the transform function on each element of source.`

So:

`An IEnumerable<ReadOnlyMemory<byte>> whose elements are the raw data buffers for the UTF8 string values.`

(obviously escaping the `<` and `>` characters however XML docs need it)

src/Microsoft.Data.Analysis/DataFrame.BinaryOperations.cs
```diff
@@ -12,6 +12,7 @@ namespace Microsoft.Data.Analysis
 {
     public partial class DataFrame
     {
+        #pragma warning disable 1591
```
eerhardt: Do you need to update the .tt file for this change?

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -17,6 +17,12 @@ namespace Microsoft.Data.Analysis
     /// </summary>
     public abstract partial class DataFrameColumn : IEnumerable
     {
+        /// <summary>
+        /// The base <see cref="DataFrameColumn"/> constructor.
+        /// </summary>
+        /// <param name="name">The name of this column.</param>
+        /// <param name="length">The length of this column.</param>
+        /// <param name="type">The type of data this column holds.</param>
         public DataFrameColumn(string name, long length, Type type)
```
eerhardt: (existing issue) - since this is an abstract class, the constructor should be marked `protected`.

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -162,6 +235,12 @@ public virtual DataFrameColumn Sort(bool ascending = true)
         /// <param name="upper"></param>
         public virtual DataFrameColumn Filter<U>(U lower, U upper) => FilterImplementation(lower, upper);
 
+        /// <summary>
+        /// Returns a new column filtered by the lower and upper bounds
+        /// </summary>
+        /// <typeparam name="U"></typeparam>
+        /// <param name="lower"></param>
```
eerhardt: Should these be filled out?

src/Microsoft.Data.Analysis/GroupBy.cs
```diff
@@ -37,7 +37,7 @@ public abstract class GroupBy
         /// </summary>
         /// <param name="numberOfRowsInEachGroup"></param>
         /// <returns></returns>
-        public abstract DataFrame Tail(int numberOfRows);
+        public abstract DataFrame Tail(int numberOfRowsInEachGroup);
```
eerhardt: We should update `Head` as well. `Tail` and `Head` should be in sync.

eerhardt:
>     public class GroupBy<TKey> : GroupBy

Does this public class need XML docs?

---
Refers to: src/Microsoft.Data.Analysis/GroupBy.cs:73 in b0706ab. [](commit_id = b0706abeb7929d4c360b74b1624bd189567da4d4, deletion_comment = False)

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
carlossanlop: Correct, `<summary></summary>` is not required.

Note: I don't think inheritdoc works properly. When I have seen it being used, the xmls we generate in artifacts when we build, end up not including any inherited documentation.
This prevents me from automatically porting documentation to dotnet-api-docs. So my suggestion would be to add proper documentation instead of using inheritdoc.

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
carlossanlop: Also don't forget the dot at the end of the sentence 😄

src/Microsoft.Data.Analysis/DataFrame.Join.cs
```diff
@@ -34,6 +34,14 @@ private void SetSuffixForDuplicatedColumnNames(DataFrame dataFrame, DataFrameCol
             }
         }
 
+        /// <summary>
+        /// Joins columns of another <see cref="DataFrame"/>
+        /// </summary>
+        /// <param name="other">The other <see cref="DataFrame"/></param>
```
carlossanlop:
```suggestion
        /// <param name="other">The other <see cref="DataFrame"/> to join.</param>
```

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
eerhardt: Asp.net uses it. How come it works for them?

It saves a ton of duplication and things getting out of sync, since it is only defined in one spot.

pgovind:
Will revisit this for 0.3.0

pgovind:
https://github.com/dotnet/corefxlab/pull/2882 takes this forward now.

PR #2791, opened 2019-11-26
Use dataTypes if it passed in to LoadCsv

pgovind:
At the moment, we ignore dataTypes if it passed in to `LoadCsv`. This patch lets us use `dataTypes` and also read in the input file just once when the data types are known. If dataTypes is `null`, we don't read the full file twice now. Instead, we rely on the `DataFrame.Append` method to append values as we read them. 

`dataTypes` can still have structs or user defined types. We just throw in such cases.

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -286,59 +286,5 @@ namespace Microsoft.Data.Analysis
             }
             return ret;
         }
-
-        private static void AppendRow(List<DataFrameColumn> columns, long rowIndex, string[] values)
-        {
-            for (int i = 0; i < columns.Count; i++)
-            {
-                DataFrameColumn column = columns[i];
-                string val = values[i];
-                Type dType = column.DataType;
-                if (dType == typeof(bool))
-                {
-                    bool boolParse = bool.TryParse(val, out bool boolResult);
-                    if (boolParse)
-                    {
-                        column[rowIndex] = boolResult;
-                        continue;
-                    }
-                    else
-                    {
-                        if (string.IsNullOrEmpty(val))
-                        {
-                            column[rowIndex] = null;
-                            continue;
-                        }
-                        throw new ArgumentException(string.Format(Strings.MismatchedValueType, typeof(bool)), nameof(val));
-                    }
-                }
-                else if (dType == typeof(float))
-                {
-                    bool floatParse = float.TryParse(val, out float floatResult);
-                    if (floatParse)
-                    {
-                        column[rowIndex] = floatResult;
-                        continue;
-                    }
-                    else
-                    {
-                        if (string.IsNullOrEmpty(val))
-                        {
-                            column[rowIndex] = null;
-                            continue;
-                        }
-                        throw new ArgumentException(string.Format(Strings.MismatchedValueType, typeof(float)), nameof(val));
-                    }
-                }
-                else if (dType == typeof(string))
-                {
-                    column[rowIndex] = values[i];
-                }
-                else
-                {
-                    throw new NotImplementedException();
-                }
-            }
-        }
     }
 }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -127,70 +197,63 @@ private static Type MaxKind(Type a, Type b)
...
+                                }
+                                else
+                                {
+                                    linesForGuessType.Add(spl);
+                                    numberOfColumns = Math.Max(numberOfColumns, spl.Length);
+                                }
                             }
                         }
+                        ++rowline;
+                        if (rowline == guessRows)
```
pgovind: This line ensures that we read in up to `guessRows` number of rows to perform our type inferencing, no more. Previously, this could have been the entire file.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1881,5 +1881,15 @@ public void TestAppendRow()
...
             Assert.Equal(6, df.Columns[1].NullCount);
         }
+
+        [Fact]
+        public void TestAppendEmptyValue()
+        {
+            DataFrame df = MakeDataFrame<int, bool>(10);
+            df.Append(new List<object> { "", true });
+            Assert.Equal(11, df.RowCount);
+            Assert.Equal(2, df.Columns[0].NullCount);
```
eerhardt: An empty string is not `null`. It should be treated separately.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -127,70 +197,63 @@ private static Type MaxKind(Type a, Type b)
 
             var linesForGuessType = new List<string[]>();
             long rowline = 0;
-            int numberOfColumns = 0;
+            int numberOfColumns = dataTypes != null ? dataTypes.Length : 0;
```
eerhardt: A more compact way of writing this is:

```C#
int numberOfColumns = dataTypes?.Length ?? 0;
```

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -127,70 +197,63 @@ private static Type MaxKind(Type a, Type b)
...
 
             if (header == true && numberOfRowsToRead != -1)
                 numberOfRowsToRead++;
 
+            DataFrame ret;
```
eerhardt: You can move this declaration down to where the variable is initialized. No need to put it here.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: I agree that an empty string should not be `null` in a `StringDataFrameColumn`, but what about an empty string in a primitive column? Our options are either to treat it as `null` or throw. I didn't think we should throw because many csv files seem to have empty strings and we still want to be able to read them. With `null`s, users could set them to whatever value they want.

eerhardt: I agree it should be `null` for non-string columns like `int`, etc.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -429,6 +429,10 @@ public void Append(IEnumerable<object> row = null)
                 {
                     DataFrameColumn column = columnEnumerator.Current;
                     object value = rowEnumerator.Current;
+                    if (value is string stringValue && string.IsNullOrEmpty(stringValue) && column.DataType != typeof(string))
```
eerhardt: Should probably add a comment here on why we are doing this - and why not for strings.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
```diff
@@ -58,5 +57,50 @@ Stream GetStream(string streamData)
...
+                {
+                    Assert.Equal(1, column.NullCount);
+                }
+                else
+                {
+                    Assert.Equal(0, column.NullCount);
+                }
+            }
+            var nullRow = df[3];
+            Assert.Null(nullRow[1]);
```
eerhardt: Add asserts for empty strings for columns 0 and 5.

PR #2794, opened 2019-12-02
Creating a `Rows` property, similar to `Columns`

pgovind:


src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System.Collections;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data.Analysis
+{
+
```
gfoidl: Nit: remove blank line here

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+    /// <summary>
+    /// Represents the rows of a <see cref="DataFrame"/>
+    /// </summary>
+    public class DataFrameRowCollection : IEnumerable<DataFrameRow>
+    {
+        private long _currentIndex = 0;
```
gfoidl: Nit:
```suggestion
        private long _currentIndex;
```
Zero-init is done by default. Though I have no strong opinion on this.

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+            {
+                List<object> row = GetRow(index);
+                return new DataFrameRow(row);
+            }
+        }
+
+        internal List<object> GetRow(long rowIndex)
+        {
+            var ret = new List<object>();
+            var columns = _dataFrame.Columns;
```
gfoidl: Initialize the list with a count?
```c#
var columns = _dataFrame.Columns;
var ret = new List<object>(columns.Count);
```
`columns` should haven explicit type instead of `var` too?

Or maybe use Linq for this:
```c#
internal List<object> GetRow(long rowIndex) 
    => _dataFrame.Columns
        .Select(column => column[rowIndex])
        .ToList();
```

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+        {
+            _row = row;
+        }
+
+        /// <summary>
+        /// Returns an enumerator of the values in this row.
+        /// </summary>
+        public IEnumerator<object> GetEnumerator()
+        {
+            foreach (object value in _row)
```
gfoidl: This may throw a `NullReferenceException` when `_row = null`. Either init `_row` with an empty collection in the ctor if `null` is given, or add a explicit null-check here with a `yield break`. I would do the latter
```c#
public IEnumerator<object> GetEnumerator()
{
    if (_row is null) yield break;

    foreach (object value in _row)
    {
        yield return value;
    }
}
```

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+        /// <summary>
+        /// An indexer to return the value at <paramref name="index"/>.
+        /// </summary>
+        /// <param name="index">The index of the value to return</param>
+        /// <returns>The value at this <paramref name="index"/>.</returns>
+        public object this[int index]
+        {
+            get
+            {
+                return _row[index];
```
gfoidl: Same here, handle the `_row = null`-case. Can `null` be returned? If so, then
```suggestion
                return _row?[index];
```

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+
+        /// <summary>
+        /// The number of values in this row.
+        /// </summary>
+        public int Count => _row.Count;
```
gfoidl:
```suggestion
        public int Count => _row?.Count ?? 0;
```
(for `_row = null`).

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        /// <summary>
+        /// A simple string representation of the values in this row
+        /// </summary>
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder();
```
gfoidl: Nit: can use `var` here
```suggestion
            var sb = new StringBuilder();
```

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+            foreach (object value in _row)
```
gfoidl:
```suggestion
            foreach (object value in this)
```
So the `_row = null`-case is handled by the `GetEnumetor`-method. Or 
```c#
if (_row is null) return string.Empty;
```
I would use the first approach, as `_row = null` won't be that common.

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+
+    /// <summary>
+    /// A DataFrameRow is a collection of values that represent a row in a <see cref="DataFrame"/>.
+    /// </summary>
+    public class DataFrameRow : IEnumerable<object>
+    {
+        internal List<object> _row;
+        internal DataFrameRow(List<object> row)
```
gfoidl: Maybe?
```suggestion
        internal DataFrameRow(List<object> row = null)
```
Then the above use doesn't need to pass `null` to the ctor explicitely.

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+    {
+        private long _currentIndex = 0;
+        private DataFrame _dataFrame;
+
+        /// <summary>
+        /// Initializes a <see cref="DataFrameRowCollection"/>.
+        /// </summary>
+        internal DataFrameRowCollection(DataFrame dataFrame)
+        {
+            _dataFrame = dataFrame;
```
gfoidl: Needs this a null-check? Or at least a `Debug.Assert(dataFrame != null);`? Otherwise there may be `NullReferenceExceptions` in the instance-methods / properties.

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
gfoidl: Anyway this doesn't need to be a field (see comment below).

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+        public long Length => _dataFrame.RowCount;
+
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+    }
+
+
+    /// <summary>
+    /// A DataFrameRow is a collection of values that represent a row in a <see cref="DataFrame"/>.
+    /// </summary>
+    public class DataFrameRow : IEnumerable<object>
```
gfoidl: Move this class to a separate file?

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+
+        /// <summary>
+        /// Returns an enumerator of <see cref="DataFrameRow"/> objects
+        /// </summary>
+        public IEnumerator<DataFrameRow> GetEnumerator()
+        {
+            DataFrameRow _currentRow = new DataFrameRow(null);
+            while (_currentIndex < _dataFrame.RowCount)
+            {
+                _currentRow._row = GetRow(_currentIndex);
```
gfoidl: You are mutating the rerference of `_row` here, so `_currentRow` (which gets yielded) will be updated in each iteration. (aside: `_currentRow` as local should be named `currentRow`)

To mitigate this
```c#
public IEnumerator<DataFrameRow> GetEnumerator()
{
    for (long currentIndex = 0; currentIndex < _dataFrame.RowCount; ++currentIndex)
    {
        yield return new DataFrameRow(GetRow(currentIndex));
    }
}
```

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,130 @@
...
+        IEnumerator IEnumerable.GetEnumerator() => GetEnumerator();
+    }
+
+
+    /// <summary>
+    /// A DataFrameRow is a collection of values that represent a row in a <see cref="DataFrame"/>.
+    /// </summary>
+    public class DataFrameRow : IEnumerable<object>
+    {
+        internal List<object> _row;
```
eerhardt: Instead of copying all the values out into this object, I wonder if it would be better if `DataFrameRow` was more like a "view" into the current values of the `DataFrame`. That way it could also be editable.

To do this, `DataFrameRow` would get a reference to the `DataFrame` and a `rowIndex`. Any time someone asked for a value or to set a value, it would just forward that to the correct underlying value inside the actual `DataFrame`. Thoughts?

gfoidl: I like this idea 👍 In this case _View_ should get part of the name, to make this explicit -> `DataFrameRowView`.

Just thinking about use cases where one really wants to have a copy, so that modifications don't "flow" back to the actual dataframe. Here I have some heuristics in mind, that manipulate the data, get the score and if it's worse, switch back to previous values.

So maybe it will be good to provide both? I.e. like here with the copy, and one with the view?

eerhardt: > In this case View should get part of the name, to make this explicit -> DataFrameRowView.

I am not sure. We don't have the word View in `DataFrameColumn`. My thinking is that `DataFrameRow` and `DataFrameColumn` would work similarly.

> So maybe it will be good to provide both?

For the copy case - couldn't the user copy the data out themselves?

gfoidl: > `DataFrameRow` and `DataFrameColumn` would work similarly.

That's fine and the least astonishing.

> couldn't the user copy the data out themselves?

You're right. That will be pretty trivial.

pgovind: I agree with the idea here. Making `DataFrameRow` a view into the `DataFrame` sounds like a good idea. Updated the PR.

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,58 @@
...
+            {
+                yield return new DataFrameRow(_dataFrame, _currentRowIndex);
+                _currentRowIndex++;
+            }
+        }
+
+        /// <summary>
+        /// The number of rows in this <see cref="DataFrame"/>.
+        /// </summary>
+        public long Length => _dataFrame.RowCount;
```
eerhardt: A couple problems with this API:

1. `Length` vs `Count`. This is a terrible .NET-ism, but array-like things use "Length", while collection like things (ICollection, etc) use "Count".
2. Why have both DataFrame.RowCount and have DataFrame.Rows.Length?  We should have just 1 way of getting the number of rows.

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Collections.Generic;
+
+namespace Microsoft.Data.Analysis
+{
+    /// <summary>
+    /// Represents the rows of a <see cref="DataFrame"/>
+    /// </summary>
+    public class DataFrameRowCollection : IEnumerable<DataFrameRow>
+    {
+        private long _currentRowIndex;
```
eerhardt: This field should be removed.

src/Microsoft.Data.Analysis/DataFrameRow.cs
```diff
@@ -0,0 +1,73 @@
...
+using System.Text;
+
+namespace Microsoft.Data.Analysis
+{
+    /// <summary>
+    /// A DataFrameRow is a collection of values that represent a row in a <see cref="DataFrame"/>.
+    /// </summary>
+    public class DataFrameRow : IEnumerable<object>
+    {
+        private DataFrame _dataFrame;
```
eerhardt: `readonly` for both of these fields.

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,58 @@
...
+        private DataFrame _dataFrame;
```
eerhardt: `readonly`

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1869,6 +1869,23 @@ public void TestColumns()
 
         }
 
+        [Fact]
+        public void TestRows()
```
eerhardt: Are there more tests that need to be added here? Ex. setting a value through a DataFrameRow, or enumerating a DataFrameRow.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -32,20 +32,28 @@ public enum DropNullOptions
...
             _columnCollection = new DataFrameColumnCollection(columns, OnColumnsChanged);
+            _rowCollection = new DataFrameRowCollection(this);
         }
 
         public long RowCount => _columnCollection.RowCount;
 
         public DataFrameColumnCollection Columns => _columnCollection;
 
+        /// <summary>
+        /// Returns a <see cref="DataFrameRowCollection"/> that contains the rows in this <see cref="DataFrame"/>
```
gfoidl:
```suggestion
        /// Returns a <see cref="DataFrameRowCollection"/> that contains a view to the rows in this <see cref="DataFrame"/>
```
?

BTW: for other members the xml-comments are missing.

src/Microsoft.Data.Analysis/DataFrameRow.cs
```diff
@@ -0,0 +1,73 @@
...
+        private long _rowIndex;
+        internal DataFrameRow(DataFrame df, long rowIndex)
+        {
+            _dataFrame = df;
+            _rowIndex = rowIndex;
```
gfoidl: Should there be an assert / check that `rowIndex` is within bounds? Or is it OK to handle this later / bail on error?

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,58 @@
...
+                return new DataFrameRow(_dataFrame, index);
+            }
+        }
+
+        /// <summary>
+        /// Returns an enumerator of <see cref="DataFrameRow"/> objects
+        /// </summary>
+        public IEnumerator<DataFrameRow> GetEnumerator()
+        {
+            while (_currentRowIndex < _dataFrame.RowCount)
```
gfoidl: Use a plain for-loop instead? (See outdated comment).
No need to have the `_currentRowIndex` as field. That would be needed if the iterator is implemented by hand, not via `yield return` (i.e. by the compiler).

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
pgovind: Does the return type matter for `Length` vs `Count`? That's the reason I chose `Length`. `DataFrameRowCollection` is a collection, but it can have `long` number of rows, which `Count` usually returns int.

2) Hmm, I added `Length` here as an afterthought. I wanted a way for users to query the number of rows in the DataFrame from a `DataFrameRowCollection` object. To be consistent with how `Column` works, I'll keep the `Length` here and delete the `RowCount` API from `DataFrame`

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
pgovind: Hmmm, alright. I wanted to support returning a subset of the rows in the `DataFrame` in the future. Guess we can always add this back in later

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
eerhardt: I don't think the return type matters, I think it is a "what does this class represent and how is the word used?" question.

Since this class is named a XXXCollection, I think the term `Count` makes more sense. "Length" implies that they are logically contiguous. This is the "Length" of an array, string, or a stream. But if you had a set of numbers, or a dictionary, you wouldn't use the term "Length".

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: There is now, after we moved to a view. Added them.

src/Microsoft.Data.Analysis/DataFrameRow.cs
pgovind: It's an internal constructor, I doubt someone would pass in an invalid rowIndex. I added one any way though.

src/Microsoft.Data.Analysis/DataFrame.Join.cs
```diff
@@ -44,7 +44,7 @@ public DataFrame Join(DataFrame other, string leftSuffix = "_left", string right
                     DataFrameColumn newColumn = Columns[i].Clone();
                     ret.Columns.Insert(ret.Columns.Count, newColumn);
                 }
-                long minLength = Math.Min(RowCount, other.RowCount);
+                long minLength = Math.Min(Rows.Count, other.Rows.Count);
```
pgovind: All of these are simple refactors

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
```diff
@@ -0,0 +1,56 @@
...
+            for (long i = 0; i < Count; i++)
+            {
+                yield return new DataFrameRow(_dataFrame, i);
+            }
+        }
+
+        /// <summary>
+        /// The number of rows in this <see cref="DataFrame"/>.
+        /// </summary>
+        public long Count => _dataFrame.Columns.RowCount;
```
pgovind: It might be weird that `DataFrameColumnCollection` stores `RowCount`. It started off that way because I needed to know how many rows are in the `DataFrame` and `DataFrameColumnCollection` didn't have a reference to `DataFrame`. We can move `RowCount` to `DataFrameRowCollection`, but it doesn't give us any benefit yet, so I'd vote to keep it this way for now until we have reason to do the refactoring work.

src/Microsoft.Data.Analysis/DataFrameRow.cs
```diff
@@ -0,0 +1,75 @@
...
+
+        /// <summary>
+        /// A simple string representation of the values in this row
+        /// </summary>
+        public override string ToString()
+        {
+            StringBuilder sb = new StringBuilder();
+            foreach (object value in this)
+            {
+                sb.Append(value != null ? value.ToString() : "null").Append(" ");
```
eerhardt: Could be simplified to:

```C#
sb.Append(value?.ToString() ?? "null").Append(" ");
```

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -186,15 +186,15 @@ public void TestIndexer()
             Assert.Equal("Int1", dataFrame.Columns[0].Name);
 
             var headList = dataFrame.Head(5);
-            Assert.Equal(14, (int)headList[4][1]);
+            Assert.Equal(14, (int)headList.Rows[4][1]);
```
eerhardt: Should this be `headList[4, 1]`?

pgovind: It could be. It's the same thing :)

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1869,46 +1869,94 @@ public void TestColumns()
...
+        public void TestRows()
+        {
+            DataFrame df = MakeDataFrameWithAllColumnTypes(10);
+            DataFrameRowCollection rows = df.Rows;
+            Assert.Equal(10, rows.Count);
+            DataFrameRow firstRow = rows[0];
+            object firstValue = firstRow[0];
+            Assert.Equal(df[0, 0], firstValue);
+            long rowCount = 0;
+            foreach (DataFrameRow row in rows)
```
eerhardt: How about testing enumerating through the objects in a `DataFrameRow`?

src/Microsoft.Data.Analysis/DataFrameRow.cs
```diff
@@ -0,0 +1,75 @@
...
+            set
+            {
+                _dataFrame.Columns[index][_rowIndex] = value;
+            }
+        }
+
+        /// <summary>
+        /// The number of values in this row.
+        /// </summary>
+        public int Count => _dataFrame.Columns.Count;
```
eerhardt: `DataFrameRow.Count` is a little odd to me. I would if this should be `ValuesCount` or something. 

Thoughts on a different name here?

src/Microsoft.Data.Analysis/DataFrameRowCollection.cs
eerhardt: I agree is is weird that DataFrameColumnCollection stores RowCount. However, I don't see a benefit in moving it at this point. So I'm fine the way it is, for now.

src/Microsoft.Data.Analysis/DataFrameRow.cs
pgovind: I can't think of a strong choice here. `ValuesCount` sounds good to me. I checked to see if IDataView.DataViewRow has an equivalent, it doesn't.

eerhardt: Or maybe we drop `Count` here, and the user can go to the `Columns.Count` instead?

eerhardt: https://apisof.net/catalog/System.Data.DataRow also doesn't have a `Count` property. How about we drop it for now, and add it later if we see a need.

pgovind: Hmmm, that's not a bad idea. If they include Linq, they'll be able to get the number of values anyway.

pgovind: Yup, dropping.

pgovind: It would mean it's not straightforward to mutate the values in a `DataFrameRow` though. An indexer exists on `DataFrameRow`, but to use the right index, users would need to know the number of values in the row. The foreach enumeration is readonly.

eerhardt: the user can go to the Columns.Count instead

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: Done.

pgovind:
Thanks @eerhardt  and @gfoidl!

PR #2797, opened 2019-12-04
DataFrame.LoadCsv throws an exception on projects targeting < netcore3.0

### LINKED ISSUE #2783: DataFrame.LoadCsv throws an exception on projects targeting < netcore3.0
LoadCsv calls `StreamReader(..encoding: null...)` which throws an exception on netcore 2.1 and 2.2. The fix is probably to expose an encoding parameter on the `LoadCsv` API

eerhardt:
Fixing by passing in an encoding and a default buffer size.

Also, get our tests running on .NET Framework.

Fix #2783

### DIFF
--- a/tests/Microsoft.Data.Analysis.Tests/BufferTests.cs
+++ b/tests/Microsoft.Data.Analysis.Tests/BufferTests.cs
@@ -188,6 +188,7 @@ namespace Microsoft.Data.Analysis.Tests
                 Assert.Null(clone[i]);
         }
 
+#if !NETFRAMEWORK // https://github.com/dotnet/corefxlab/issues/2796
         [Fact]
         public void TestPrimitiveColumnGetReadOnlyBuffers()
         {
@@ -275,5 +276,6 @@ namespace Microsoft.Data.Analysis.Tests
                 }
             }
         }
+#endif //!NETFRAMEWORK
     }
 }
--- a/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
+++ b/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
@@ -169,7 +169,7 @@ namespace Microsoft.Data.Analysis.Tests
         {
             IEnumerable<int> randomIndices = Enumerable.Range(0, (int)input.RowCount);
             IEnumerable<int> trainIndices = randomIndices.Take((int)(input.RowCount * testRatio));
-            IEnumerable<int> testIndices = randomIndices.TakeLast((int)(input.RowCount * (1 - testRatio)));
+            IEnumerable<int> testIndices = randomIndices.Skip((int)(input.RowCount * testRatio));
             Test = input[testIndices];
             return input[trainIndices];
         }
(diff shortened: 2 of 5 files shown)


tests/Microsoft.Data.Analysis.Tests/BufferTests.cs
```diff
@@ -188,6 +188,7 @@ public void TestArrowStringColumnClone()
                 Assert.Null(clone[i]);
         }
 
+#if !NETFRAMEWORK // https://github.com/dotnet/corefxlab/issues/2796
```
pgovind: I'm curious, how did you discover this? We don't have CI running NETFramework right?

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -169,7 +169,7 @@ public DataFrame SplitTrainTest(DataFrame input, float testRatio, out DataFrame
         {
             IEnumerable<int> randomIndices = Enumerable.Range(0, (int)input.RowCount);
             IEnumerable<int> trainIndices = randomIndices.Take((int)(input.RowCount * testRatio));
-            IEnumerable<int> testIndices = randomIndices.TakeLast((int)(input.RowCount * (1 - testRatio)));
```
pgovind: Is this a netcore only API?

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1540,38 +1540,38 @@ public void TestPrefixAndSuffix()
 
             DataFrame prefix = df.AddPrefix("Prefix_");
             IEnumerable<DataViewSchema.Column> prefixNames = ((IDataView)prefix).Schema;
-            foreach ((DataViewSchema.Column First, DataViewSchema.Column Second) in columnNames.Zip(((IDataView)df).Schema))
+            foreach ((DataViewSchema.Column First, DataViewSchema.Column Second) in columnNames.Zip(((IDataView)df).Schema, (e1, e2) => (e1, e2)))
```
pgovind: What does the `(e1, e2)` delegate do? I don't see it being used anywhere

tests/Microsoft.Data.Analysis.Tests/Microsoft.Data.Analysis.Tests.csproj
```diff
@@ -1,7 +1,8 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.0</TargetFramework>
+    <TargetFrameworks>net461;netcoreapp3.0</TargetFrameworks>
+    <TargetFrameworks Condition="'$(OS)' != 'Windows_NT'">netcoreapp3.0</TargetFrameworks>
```
pgovind: Just so I understand, on a Mac, this would define both net461 and netcoreapp3.0 in line 4 and then be overwritten to just netcoreapp3.0 in line 5?

tests/Microsoft.Data.Analysis.Tests/BufferTests.cs
eerhardt: With this PR we are now running tests against .NET Framework. See https://github.com/dotnet/corefxlab/pull/2797/files#diff-08cf3efcb8a6af2b5b7400915d600676R4

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
eerhardt: Correct - it isn't available on .NET Framework.

https://apisof.net/catalog/System.Linq.Enumerable.TakeLast%3CTSource%3E(IEnumerable%3CTSource%3E,Int32)

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
eerhardt: It is taking in 2 parameters, `e1` and `e2`, and then returning the two parameters as a Tuple `(e1, e2)`. This is what it was doing before, except the `Zip` overload that doesn't take a selector and returns a Tuple is only in .NET Core 3.0.

https://apisof.net/catalog/System.Linq.Enumerable.Zip%3CTFirst,TSecond%3E(IEnumerable%3CTFirst%3E,IEnumerable%3CTSecond%3E)

tests/Microsoft.Data.Analysis.Tests/Microsoft.Data.Analysis.Tests.csproj
eerhardt: Correct, that's how MSBuild properties get interpreted - top to bottom. Doing it this way works around some VS bugs that inspect the TargetFrameworks node:

Example:
https://github.com/dotnet/project-system/issues/4149

PR #2800, opened 2019-12-05
Params constructor on DataFrame

pgovind:
Very simple PR

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -39,7 +39,13 @@ namespace Microsoft.Data.Analysis
             _rowCollection = new DataFrameRowCollection(this);
         }
 
-        public DataFrame(IList<DataFrameColumn> columns)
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -39,7 +39,13 @@ public DataFrame()
...
         }
 
-        public DataFrame(IList<DataFrameColumn> columns)
+        public DataFrame(IEnumerable<DataFrameColumn> columns)
+        {
+            _columnCollection = new DataFrameColumnCollection(columns, OnColumnsChanged);
+            _rowCollection = new DataFrameRowCollection(this);
+        }
+
+        public DataFrame(params DataFrameColumn[] columns)
```
eerhardt: Do we still need the empty constructor now that we have a `params` overload?

pgovind: We don't. Deleting it.

PR #2801, opened 2019-12-06
Remove `T : unmanaged` constraint from DataFrameColumn.BinaryOperations

pgovind:


src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.tt
```diff
@@ -54,9 +55,29 @@ namespace Microsoft.Data.Analysis
...
         {
 <# if (method.MethodType == MethodType.BinaryScalar) {#>
+<# if (method.IsBitwise == true) { #>
             return <#=method.MethodName#>Implementation(value, inPlace);
+<# } else { #>
+            switch (typeof(U))
+            {
+<# foreach (TypeConfiguration type in typeConfiguration) { #>
+                case Type <#=type.TypeName#>Type when <#=type.TypeName#>Type == typeof(<#=type.TypeName#>):
+                    return <#=method.MethodName#>Implementation(Unsafe.As<U, <#=type.TypeName#>>(ref value), inPlace);
```
pgovind: Had to use Unsafe.As unfortunately. Definitely open to other/better ways.

ahsonkhan: The safe way to do this would probably be something like this (but that boxes):
```C#
return (bool)(object)(value);
```
```suggestion
                    return <#=method.MethodName#>Implementation((<#=type.TypeName#>)(object)(value), inPlace);
```

pgovind: Right, I was trying to avoid that :). Boxing is already showing up in the perf profiles

gfoidl: Is it really boxing or is the JIT able to "undo" the boxing?
In most cases JIT is smart enough -- https://sharplab.io/#v2:EYLgxg9gTgpgtADwGwBYA0AXEBDAzgWwB8ABAJgEYBYAKGIAYACAZQAtsoAHAGW2ADoASgFcAdhgCW+GAG4aNYgGYGZBgGEGAbxoMdDbboDaAKXEYA4jBEwo4sAAoMATw4wIAMzvAIEADYBKPwBdfR1FBi9fBgsMADVsHyEYAB4AFQA+OxSGADd4xL8GAF40hk9vfzsIYAArGDAMP1yEmRoAXyA=

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.cs
```diff
@@ -49,7 +50,37 @@ public override DataFrameColumn Add(DataFrameColumn column, bool inPlace = false
         }
         public override DataFrameColumn Add<U>(U value, bool inPlace = false)
         {
-            return AddImplementation(value, inPlace);
+            switch (typeof(U))
```
eerhardt: We should change this to `if (typeof(U) == typeof(bool))`.

1. It is simpler than this switch case.
2. The JIT can specialize the generic so the whole `if` statements go away when it is JIT'ing for a specific type.

eerhardt:
Should we add a test or two for this? That way we ensure it works for non `unmanaged` types now and in the future.

pgovind:
> Should we add a test or two for this? That way we ensure it works for non unmanaged types now and in the future.

We already do have tests for this as part of testing `StringDataFrameColumn`(since it's the only column dealing with references). I added a test to `PrimitiveDataFrameColumn` now

eerhardt:
> We already do have tests for this as part of testing StringDataFrameColumn

If we already had tests for this, then those tests would be failing without this change.

src/Microsoft.Data.Analysis/StringDataFrameColumn.BinaryOperations.cs
```diff
@@ -101,7 +101,7 @@ public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
             return ret;
         }
 
-        public override PrimitiveDataFrameColumn<bool> ElementwiseEquals<T>(T value)
+        public override PrimitiveDataFrameColumn<bool> ElementwiseValueEquals<T>(T value)
```
eerhardt: How come it is `AddValue` but here it is `ValueEquals`?

eerhardt: It feels like it should be `EqualsValue`, because it is a method on this column, and you are asking if this column equals a value. Not "value equals".

---
In reply to: [356132434](https://github.com/dotnet/corefxlab/pull/2801#discussion_r356132434) [](ancestors = 356132434)

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.ReversedBinaryOperations.tt
```diff
@@ -106,7 +105,7 @@ namespace Microsoft.Data.Analysis
                         }
                     }
                 default:
-                    throw new NotSupportedException();
+                    throw new NotImplementedException();
```
eerhardt: This should stay `NotSupportedException`. `NotImplementedException` is for things that we will eventually implement, but haven't gotten a chance yet. `NotSupported` means we don't have plans to support it.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.tt
```diff
@@ -54,9 +55,27 @@ namespace Microsoft.Data.Analysis
...
+<# if (method.IsBitwise == true) { #>
             return <#=method.MethodName#>Implementation(value, inPlace);
+<# } else { #>
+<# foreach (TypeConfiguration type in typeConfiguration) { #>
+            if (typeof(U) == typeof(<#=type.TypeName#>))
+            {
+                return <#=method.MethodName#>Implementation(Unsafe.As<U, <#=type.TypeName#>>(ref value), inPlace);
+            }
+<# } #>
+            throw new NotImplementedException();
```
eerhardt: `NotSupported`

src/Microsoft.Data.Analysis/StringDataFrameColumn.BinaryOperations.cs
pgovind: Yea I debated between the names. I know they are inconsistent but they seemed to make more sense to me. How does `ElementwiseNotEqualsValue` sound? I thought `ElementwiseValueNotEquals` sounded better. Similarly, `AddValue` sounded better than `ValueAdd`.

eerhardt: Take "elementwise" out of the method name and just think about "Equals".

`myColumn.Equals(5)` is what we want to write.

Why does `myColumn.ValueEquals(5)` sound better than `myColumn.EqualsValue(5)`?  Especially if you think `AddValue` sounds better than `ValueAdd`.

pgovind: I think for the `Equals` method, `Value` can go in the front or the back and it still reads nicely. For `NotEquals` though, `ElementwiseValueNotEquals` almost sounds like a proper sentence. `ElementwiseNotEqualsValue` sounds like `Value` has been added to the end as an afterthought? Even if you disregard `Elementwise`, `ValueNotEquals` reads "better" than `NotEqualsValue` right?

eerhardt: I'm not sure I agree. Because this is a method on a column. The code reads like English - "column not equals value 5" - `column.NotEqualsValue(5)`. I wouldn't say "column value not equals 5". Just like I wouldn't say "column value add 5". I would say "column add value 5".

eerhardt: If instead we would pick the other overload (the one that takes a column), how would we do the naming?

Would we name it `column.ElementwiseColumnNotEquals(otherColumn)`?  Or would we name it `column.ElementwiseNotEqualsColumn(otherColumn)`.

In the end, I really don't want `Value` in the method name at all. But if we need to add it, I want it to be consistent on where/how we add `Value` so it makes sense as a whole. `AddValue`, `SubtractValue`, `NotEqualsValue`, etc.

pgovind: I agree that `Value` shouldn't be in the method name either. I modified the PR to remove all the `Value` additions to the APIs. For all the overloaded methods, we now check if the value passed in is a `DataFrameColumn` and call the right overload if it is.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.tt
```diff
@@ -54,9 +55,35 @@ namespace Microsoft.Data.Analysis
...
+            {
+                return <#=method.MethodName#>(column);
+            }
+<# foreach (TypeConfiguration type in typeConfiguration) { #>
+            if (typeof(U) == typeof(<#=type.TypeName#>))
+            {
+                return <#=method.MethodName#>Implementation(Unsafe.As<U, <#=type.TypeName#>>(ref value));
+            }
+<# } #>
+            throw new NotImplementedException();
```
pgovind: NotSupported?

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -497,6 +497,11 @@ public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
 
         public override PrimitiveDataFrameColumn<bool> ElementwiseEquals<T>(T value)
         {
+            DataFrameColumn column = value as DataFrameColumn;
```
eerhardt: Can use the new pattern matching in C#:

```
if (value is DataFrameColumn column)
{
    return ElementwiseEquals(column);
}
```

eerhardt:
>             where U : unmanaged

Why does U need to be unmanaged here? Can't you remove that restriction here, and then you don't need the big type switch in the callsite?

---
Refers to: src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.cs:1186 in 163b47b. [](commit_id = 163b47b5461342d1e76f42fe6d079f0c936dc213, deletion_comment = False)

src/Microsoft.Data.Analysis/StringDataFrameColumn.BinaryOperations.cs
```diff
@@ -103,6 +103,11 @@ public PrimitiveDataFrameColumn<bool> ElementwiseEquals(string value)
 
         public override PrimitiveDataFrameColumn<bool> ElementwiseEquals<T>(T value)
         {
+            DataFrameColumn column = value as DataFrameColumn;
+            if (column != null)
+            {
+                return ElementwiseEquals(column);
```
eerhardt: Are we certain this will bind to the `DataFrameColumn` overload? and it won't turn into infinite recursion? Make sure we have tests for all these to ensure passing in a DataFrameColumn works here.

pgovind: Yup, first thing I verified. The compiler is able to find the `DataFrameColumn` overload. It goes back to the generic overload only when a derived type is passed in, for ex: `StringDataFrameColumn`. Added some tests for this too.

pgovind:
> Why does U need to be unmanaged here? Can't you remove that restriction here, and then you don't need the big type switch in the callsite?

Because when I first wrote this code, I wanted to reduce the number of `Unsafe.As` uses. Removed the restriction now. We do have more `Unsafe.As` calls now, but I don't have an issue with the way I'm using them.

Issue #2805, opened 2019-12-17
An ApplyElementWise method that returns a column 


zHaytam:
I'm trying out the `Microsoft.Data.Analysis` and was looking for a method that applies a `Func<,>` but returns the result as an array of values (maybe not efficient) or a column, just like in Pandas.

Here's what I'm currently doing:
```csharp
public static PrimitiveDataFrameColumn<TResult> Apply<T, TResult>(this PrimitiveDataFrameColumn<T> column, 
    Func<T, TResult> func) 
    where T : unmanaged
    where TResult : unmanaged
{
    var resultColumn = new PrimitiveDataFrameColumn<TResult>(string.Empty, 0);

    foreach (var row in column)
        resultColumn.Append(func(row.Value));

    return resultColumn;
}
```

Example usage:
```csharp
var birthdayColumn = df["Birthday"] as PrimitiveDataFrameColumn<DateTime>;
var currentYear = DateTime.Now.Year;
df["Age"] = birthdayColumn.Apply(d => currentYear - d.Year);
df.PrettyPrint();
```

Thank you!

pgovind:
Hey @zHaytam, have you looked at https://github.com/dotnet/corefxlab/blob/master/src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs#L488? The `ApplyElementwise` API does something very similar to what you want. The only difference is that it works in place. To do what you want, you can do:

``` csharp
df["Age"] = df["Birthday"].Clone().ApplyElementwise((d, index) => return currentYear - d.Year));
```

zHaytam:
Hello @pgovind, the problem with `ApplyElementWise` is that the output type is the same as the input. My birthday column is of type `DateTime` while the age column, I want it to be `int` :/

pgovind:
Ah, my bad. I didn't see the return type. Yup, you're right. This is on my list of new APIs to add to `PrimitiveDataFrameColumn`. Would you be interested in putting up a PR perhaps? We could get this API in for the next preview :)

zHaytam:
I would gladly try. I'm pretty sure the example I gave is far from being the best way to do it, I'm assuming it needs to be implemented directly in the Container?

pgovind:
Yup. I would define the API in `PrimitiveDataFrameColumn`, maybe create the `returnColumn` in the API and then implement the loop inside `PrimitiveDataFrameColumnContainer`.

zHaytam:
[...]

PR #2807, opened 2019-12-19
Add Apply<TResult>method to PrimitiveDataFrameColumn

zHaytam:
For issue #2805.

This PR adds an `Apply<TResult>` method to `PrimitiveDataFrameColumn` that takes a `Func<T?, TResult?>` and returns a new column with the new type.

Example usage (taken from the written unit test):
```csharp
int[] values = { 1, 2, 3, 4, 5 };
var col = new PrimitiveDataFrameColumn<int>("Ints", values);
PrimitiveDataFrameColumn<double> newCol = col.Apply(i => i + 0.5d);
```

### DIFF
--- a/src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
+++ b/src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
@@ -487,6 +487,13 @@ namespace Microsoft.Data.Analysis
 
         public void ApplyElementwise(Func<T?, long, T?> func) => _columnContainer.ApplyElementwise(func);
 
+        public PrimitiveDataFrameColumn<TResult> Apply<TResult>(Func<T?, TResult?> func) where TResult : unmanaged
+        {
+            var resultColumn = new PrimitiveDataFrameColumn<TResult>("Result", Length);
+            _columnContainer.Apply(func, resultColumn._columnContainer);
+            return resultColumn;
+        }
+
         /// <summary>
         /// Clips values beyond the specified thresholds
         /// </summary>
--- a/src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
+++ b/src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
@@ -487,6 +487,12 @@ namespace Microsoft.Data.Analysis
 
         public void ApplyElementwise(Func<T?, long, T?> func) => _columnContainer.ApplyElementwise(func);
 
+        /// <summary>
+        /// Applies a function to all the values
+        /// </summary>
+        /// <typeparam name="TResult">The new column's type</typeparam>
+        /// <param name="func">The function to apply</param>
+        /// <returns>A new PrimitiveDataFrameColumn containing the new values</returns>
         public PrimitiveDataFrameColumn<TResult> Apply<TResult>(Func<T?, TResult?> func) where TResult : unmanaged
         {
             var resultColumn = new PrimitiveDataFrameColumn<TResult>("Result", Length);
--- a/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
+++ b/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
@@ -1946,6 +1946,7 @@ namespace Microsoft.Data.Analysis.Tests
                 }
             }
         }
+
         [Fact]
         public void TestAppendRow()
         {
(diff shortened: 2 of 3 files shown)


src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
```diff
@@ -487,6 +487,13 @@ public override GroupBy GroupBy(int columnIndex, DataFrame parent)
 
         public void ApplyElementwise(Func<T?, long, T?> func) => _columnContainer.ApplyElementwise(func);
 
+        public PrimitiveDataFrameColumn<TResult> Apply<TResult>(Func<T?, TResult?> func) where TResult : unmanaged
```
eerhardt: Can you add xml doc comments to the new public API?

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -2027,5 +2028,22 @@ public void TestAppendEmptyValue()
             Assert.Equal(13, df.Rows.Count);
             Assert.Equal(1, df.Columns[2].NullCount);
         }
+
+        [Fact]
+        public void TestApply()
+        {
+            int[] values = { 1, 2, 3, 4, 5 };
+            var col = new PrimitiveDataFrameColumn<int>("Ints", values);
+            var df = new DataFrame(col);
```
eerhardt: I don't think `df` is used. Can it be removed?

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
eerhardt: The existing method is named `ApplyElementwise`, but this is just `Apply`. Is there a reason for the discrepancy?

/cc @pgovind - thoughts?

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
zHaytam: Yes, definitely.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
zHaytam: Personally, I find `ApplyElementwise` wouldve been better as `Apply` too since the `func` parameter explains that it applies the function to each element.
Also, what `Apply` does is also similar to what pandas' `apply` method does, so that should feel good for developers with a python background.

pgovind: @zHaytam : I agree with @eerhardt that this should be named `ApplyElementwise`. We wouldn't need to go through API review then. 

Also, how do you(and @eerhardt) feel about renaming to `ApplyElementwise` and merging the 2 overloads so we end up with 1 `ApplyElementwise` method that looks like this:

``` csharp
public PrimitiveDataFrameColumn<TResult> ApplyElementwise<TResult>(Func<T?, long, TResult?> func, bool inPlace = false) where TResult : unmanaged
``` 
2 reasons:
1. Adding the `inPlace` parameter makes it more consistent with APIs like `Clip`, `FillNulls`, `Multiply` etc
2. Having just 1 API is cleaner IMO. 
The minor problem I see is a statement such as `intColumn.ApplyElementwise((value, index) => return 0.0f), inPlace: true)`. We would have to throw here. Arguably it is ugly that we allow users to call this API and throw immediately? Thoughts? 

This will only need minor changes I think. You'd just have to check if `inPlace` is set and set `resultColumn` to `this` or a new `PrimitiveDataFrameColumn` and everything else should just work.

eerhardt: I think there are separate discussions here, so I'll try to separate them:

## The name `Apply` or `ApplyElementwise`.

We don't have `Elementwise` in every method name that does something element-wise. For example, `Clip` and `FillNulls` don't have `Elementwise` in their name. So does `Apply` need to? My personal opinion is that it is obvious the method acts on each element, so we should be able to just name it `Apply`. The reason we added `Elementwise` to some method names was so `Equals` and `GreaterThan` could return a collection of bools instead of a single bool.

## Merging this into a single method

I think it makes sense to keep them separate overloads:

```
public PrimitiveDataFrameColumn<T> Apply(Func<T?, long, T?> func, bool inPlace = false)
public PrimitiveDataFrameColumn<TResult> Apply<TResult>(Func<T?, TResult?> func) where TResult : unmanaged
```

This solves the issue that if the `T`'s don't match up, you can't do an `inPlace` operation.

pgovind: Ok, that sounds fine to me. I'll rename `ApplyElementwise` to `Apply` the next time I go through this file or @zHaytam, you can do it in this PR too if you want :). I'm approving the PR. Thanks for the work!!

zHaytam: Sounds good to me too! One question though, the current `ApplyElementwise` doesn't have an `inPlace` argument, should that be added?

pgovind: @zHaytam, since we're not touching the `ApplyElementwise` method, maybe not in this PR. I think this is good to go in as it is. We should add the `inPlace` option later.

pgovind:
Squash and merging this. Thank you @zHaytam  for the patch!

PR #2808, opened 2019-12-21
Added static factory methods to DataFrameColumn 

MgSam:
I added static factory methods only for the constructor overloads where type inference is possible. 

I experimented with adding overloads for all constructors, but for the overloads where `values` are not provided it didn't feel helpful. It seemed weird to do `DataFrameColumn.Create("foo")` and get back `StringDataFrameColumn`. (Also seemed like a bad idea because it would preclude adding any other "special" DataFrameColumn types in the future).

I also thought it would be inconsistent to allow `DataFrameColumn.Create<int>("Foo")` but not have any counterpart for `String`. Please let me know if you disagree, if so I can add the other overloads back in.

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrameColumn.cs
+++ b/src/Microsoft.Data.Analysis/DataFrameColumn.cs
@@ -24,8 +24,6 @@ namespace Microsoft.Data.Analysis
             DataType = type;
         }
 
-        #region Static Factory Methods
-
         /// <summary>
         /// A static factory method to create a <see cref="PrimitiveDataFrameColumn{T}"/>. 
         /// It allows you to take advantage of type inference based on the type of the values supplied.
@@ -64,8 +62,6 @@ namespace Microsoft.Data.Analysis
         {
             return new StringDataFrameColumn(name, values);
         }
-
-        #endregion
 
         private long _length;
         public long Length
--- a/src/Microsoft.Data.Analysis/DataFrameColumn.cs
+++ b/src/Microsoft.Data.Analysis/DataFrameColumn.cs
@@ -51,7 +51,7 @@ namespace Microsoft.Data.Analysis
         }
 
         /// <summary>
-        /// A static factory method to create a <see cref="StringDataFrameColumn{T}"/>. 
+        /// A static factory method to create a <see cref="StringDataFrameColumn"/>. 
         /// It allows you to take advantage of type inference based on the type of the values supplied.
         /// </summary>
         /// <typeparam name="T">The type of the column to create.</typeparam>
--- a/src/Microsoft.Data.Analysis/DataFrameColumn.cs
+++ b/src/Microsoft.Data.Analysis/DataFrameColumn.cs
@@ -24,6 +24,49 @@ namespace Microsoft.Data.Analysis
             DataType = type;
         }
 
+        #region Static Factory Methods
+
+        /// <summary>
+        /// A static factory method to create a <see cref="PrimitiveDataFrameColumn{T}"/>. 
+        /// It allows you to take advantage of type inference based on the type of the values supplied.
+        /// </summary>
+        /// <typeparam name="T">The type of the column to create.</typeparam>
+        /// <param name="name">The name of the column.</param>
+        /// <param name="values">The initial values to populate in the column.</param>
+        /// <returns>A <see cref="PrimitiveDataFrameColumn{T}"/> populated with the provided data.</returns>
+        public static PrimitiveDataFrameColumn<T> Create<T>(string name, IEnumerable<T?> values) where T : unmanaged
+        {
+            return new PrimitiveDataFrameColumn<T>(name, values);
+        }
+
+        /// <summary>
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -24,6 +24,49 @@ public DataFrameColumn(string name, long length, Type type)
             DataType = type;
         }
 
+        #region Static Factory Methods
```
gfoidl: Nit remove the `#region`, it not necessary and not used anywhere else.

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -24,6 +24,49 @@ public DataFrameColumn(string name, long length, Type type)
...
+
+        /// <summary>
+        /// A static factory method to create a <see cref="PrimitiveDataFrameColumn{T}"/>. 
+        /// It allows you to take advantage of type inference based on the type of the values supplied.
+        /// </summary>
+        /// <typeparam name="T">The type of the column to create.</typeparam>
+        /// <param name="name">The name of the column.</param>
+        /// <param name="values">The initial values to populate in the column.</param>
+        /// <returns>A <see cref="PrimitiveDataFrameColumn{T}"/> populated with the provided data.</returns>
+        public static PrimitiveDataFrameColumn<T> Create<T>(string name, IEnumerable<T> values) where T : unmanaged
```
gfoidl: The only difference to the above overload is `T?` vs `T`. 

So is
```suggestion
        public static PrimitiveDataFrameColumn<T> Create<T>(string name, IEnumerable<T> values) where T : unmanaged, notnull
```
useful?

MgSam: Microsoft.Data.Analysis is targeting .NET Standard 2.0 using C# 7.3, so the `notnull` constraint is not available.

src/Microsoft.Data.Analysis/DataFrameColumn.cs
MgSam: Changed.

src/Microsoft.Data.Analysis/DataFrameColumn.cs
gfoidl: Oh, yes thanks for the hint!

MgSam:
@pgovind 
Sorry, forgot to tag you sooner.

eerhardt:
I think the code looks good. Thanks for the contribution @MgSam.

Do you think you can add some tests, or convert a few of the existing tests, to get coverage on these new APIs methods?

pgovind:
I think this change looks good too. To get test coverage, I think if you change the `MakeDataFrameWithAllMutableColumnTypes` method in DataFrameTests to use the new APIs, that should automatically give us coverage.

MgSam:
@eerhardt @pgovind I've updated some parts of the unit tests to use the static factory methods.

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -24,6 +24,45 @@ public DataFrameColumn(string name, long length, Type type)
...
+        {
+            return new PrimitiveDataFrameColumn<T>(name, values);
+        }
+
+        /// <summary>
+        /// A static factory method to create a <see cref="StringDataFrameColumn{T}"/>.
```
eerhardt: (nit) there is no such thing has `StringDataFrameColumn{T}`. It is just `StringDataFrameColumn`. Repeated below on line 60.

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -24,6 +24,45 @@ public DataFrameColumn(string name, long length, Type type)
...
+        public static PrimitiveDataFrameColumn<T> Create<T>(string name, IEnumerable<T> values) where T : unmanaged
+        {
+            return new PrimitiveDataFrameColumn<T>(name, values);
+        }
+
+        /// <summary>
+        /// A static factory method to create a <see cref="StringDataFrameColumn{T}"/>. 
+        /// It allows you to take advantage of type inference based on the type of the values supplied.
+        /// </summary>
+        /// <typeparam name="T">The type of the column to create.</typeparam>
```
eerhardt: There is no `T` type param.

These XML errors should have given you at least a warning. Are you seeing any?

MgSam: Sorry I missed this message- no I didn't see any warnings. I've removed the errant `{T}`.

pgovind: I think you missed the `T` in the documentation comments on lines 57 and 60?

eerhardt:
@MgSam - if you merge https://github.com/MgSam/corefxlab/pull/1 into your fork, I think this PR will be ready for merge.

MgSam:
@eerhardt Done. Sorry for the back and forth!

Issue #2713, opened 2019-08-16
Add more comprehensive tests for DataFrame.ReadCsv


pgovind:
The sole test now just reads in standard csv data in a stream. We'd need tests that cover nulls, missing data, conversions etc.

eerhardt:
Also add:

* invalid files
* strings with commas in them
* different separators (`|`, `;`, etc)
* headers vs. no headers

jwood803:
@eerhardt I'll take a shot at these, if that's cool. It looks like the header/no header tests are already there, though.

By invalid files, though, do you mean non-CSV files or was it something specific to test for?

eerhardt:
> I'll take a shot at these, if that's cool.

That's awesome!

> By invalid files, though, do you mean non-CSV files or was it something specific to test for?

Just generally invalid files. Could be
* malformed CSV files 
    * rows having too many/few columns
    * mixing delimiters - commas and vertical bars, etc
* binary files
* plain old text files

Things like that. To ensure we gracefully handle "invalid" files.

Issue #2787, opened 2019-11-21
[DataFrame] Allow quoting when loading CSV


lqdev:
Given the following code:

```fsharp
let df = DataFrame.LoadCsv("Data/test.csv")
```

And data that looks like the snippet below:

```text
Project Name,Neighborhood,Zip Code,TtlProjUnits,RentUnits,OwnUnits,TtlMarket,MarketRent,MarketOwn,Total Income- Restricted,Income- Restricted Rental,Income- Restricted Ownership,Tenure,Public/ Private,Includes Elderly Units?,Section_8
AB&W Cooperative Building,Roxbury,02121,24,24,0,0,0,0,24,24,0,Rental,Private,,
```

If a row contains the separator character, although surrounded in quotes:

```text
"Berkeley Investments, Inc.",South Boston,02127,97,0,97,89,0,89,8,0,8,Ownership,Private,,
```

The following error is thrown:

```text
Unhandled exception. System.ArgumentException: Expected value to be of type System.Single (Parameter 'val')
   at Microsoft.Data.Analysis.DataFrame.AppendRow(List`1 columns, Int64 rowIndex, String[] values)
   at Microsoft.Data.Analysis.DataFrame.LoadCsv(Stream csvStream, Char separator, Boolean header, String[] columnNames, Type[] dataTypes, Int64 numberOfRowsToRead, Int32 guessRows, Boolean addIndexColumn)
   at Microsoft.Data.Analysis.DataFrame.LoadCsv(String filename, Char separator, Boolean header, String[] columnNames, Type[] dataTypes, Int32 numRows, Int32 guessRows, Boolean addIndexColumn)
   at Program.main(String[] argv) in C:\Users\lqdev\Development\dotnet\FsDataFrameSample\FsDataFrameSample\Program.fs:line 9   
```

This is the source dataset for reference: https://data.boston.gov/dataset/income-restricted-housing

pgovind:
This is a consequence of how `LoadCsv` is implemented. At the moment, we're blindly splitting on "," on each line we read from a file. The ideal solution is to parse each line and build the column values as we go. That's essentially what `CsvReader`(an external lib) does. For a quick-fix, I can run a RegEx on each line being read in and split the values that way. It would be less efficient though. 

[...]

PR #2811, opened 2020-01-10
Add additional tests for ReadCsv

jwood803:
Additional tests for #2713.

### DIFF
--- a/tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
+++ b/tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
@@ -104,5 +104,156 @@ CMT,1,1,181,0.6,CSH,4.5";
             Assert.Equal("", nullRow[5]);
             Assert.Null(nullRow[6]);
         }
+
+        [Fact]
+        public void TestReadCsvWithPipeSeparator()
+        {
+            string data = @"vendor_id|rate_code|passenger_count|trip_time_in_secs|trip_distance|payment_type|fare_amount
+CMT|1|1|1271|3.8|CRD|17.5
+CMT|1|1|474|1.5|CRD|8
+CMT|1|1|637|1.4|CRD|8.5
+CMT|1|1|181|0.6|CSH|4.5";
+
+            Stream GetStream(string streamData)
+            {
+                return new MemoryStream(Encoding.Default.GetBytes(streamData));
+            }
+            DataFrame df = DataFrame.LoadCsv(GetStream(data), separator: '|');
+
+            Assert.Equal(4, df.Rows.Count);
+            Assert.Equal(7, df.Columns.Count);
+            Assert.Equal("CMT", df["vendor_id"][3]);
+
+            DataFrame reducedRows = DataFrame.LoadCsv(GetStream(data), separator: '|', numberOfRowsToRead: 3);
+            Assert.Equal(3, reducedRows.Rows.Count);
+            Assert.Equal(7, reducedRows.Columns.Count);
+            Assert.Equal("CMT", reducedRows["vendor_id"][2]);
+        }
+
+        [Fact]
+        public void TestReadCsvWithSemicolonSeparator()
+        {
+            string data = @"vendor_id;rate_code;passenger_count;trip_time_in_secs;trip_distance;payment_type;fare_amount
+CMT;1;1;1271;3.8;CRD;17.5
+CMT;1;1;474;1.5;CRD;8
+CMT;1;1;637;1.4;CRD;8.5
+CMT;1;1;181;0.6;CSH;4.5";
+
+            Stream GetStream(string streamData)
+            {
+                return new MemoryStream(Encoding.Default.GetBytes(streamData));
+            }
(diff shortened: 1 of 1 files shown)


tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
```diff
@@ -104,5 +104,156 @@ Stream GetStream(string streamData)
...
+                return new MemoryStream(Encoding.Default.GetBytes(streamData));
+            }
+
+            try
+            {
+                DataFrame df = DataFrame.LoadCsv(GetStream(data));
+            }
+            catch (Exception e)
+            {
+                Assert.Equal(typeof(IndexOutOfRangeException), e.GetType());
```
jwood803: @eerhardt I'm not sure if this is expected or even the best way to test this.

pgovind: I think the standard way of doing this is to use the `Assert.Throws` methods? `DataFrameTests.cs` has some examples :).

pgovind:
I just saw this PR. Tagging @eerhardt and myself as reviewers.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
jwood803: Thanks @pgovind! I've updated to use `Assert.Throws` instead.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
```diff
@@ -104,5 +104,140 @@ Stream GetStream(string streamData)
             Assert.Equal("", nullRow[5]);
             Assert.Null(nullRow[6]);
         }
+
+        [Fact]
+        public void TestReadCsvWithPipeSeparator()
+        {
+            string data = @"vendor_id|rate_code|passenger_count|trip_time_in_secs|trip_distance|payment_type|fare_amount
+CMT|1|1|1271|3.8|CRD|17.5
```
pgovind: So, I looked at the input in the existing tests, is there a reason you didn't add the `||||||` line here?

pgovind: Similar question with the `;` separator

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
```diff
@@ -104,5 +104,140 @@ Stream GetStream(string streamData)
...
+
+            DataFrame reducedRows = DataFrame.LoadCsv(GetStream(data), numberOfRowsToRead: 3);
+            Assert.Equal(3, reducedRows.Rows.Count);
+            Assert.Equal(6, reducedRows.Columns.Count);
+            Assert.Equal("CMT", reducedRows["vendor_id"][2]);
+
+        }
+
+        [Fact]
+        public void TestReadCsvWithCommaInString()
```
pgovind: Maybe remove this test? We actually have an issue filed against this: https://github.com/dotnet/corefxlab/issues/2787. We want to be able to handles values with `,` inside quotes.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
jwood803: I used the same input from the first test. Is that row needed to just make sure it doesn't cause an exception?

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
jwood803: Sure. Or would it be better to ignore this one for when the other issue gets fixed it can then be used for it? Just a thought :)

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
pgovind: That's one the reasons yes. The other being that for a `StringDataFrameColumn`, empty values are valid. For other column types, empty values are considered as `null` values and increase the `NullCount`.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
pgovind: Alright sure, we can keep it. I don't really have a strong preference here.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
jwood803: Great points! Updated to add the empty values column.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
jwood803: Skipped the test and removed the `Assert.Throws`. Not quite sure what the expected output should be in this case so I left the rest of the test as is, so it would still need to be updated.

Issue #2788, opened 2019-11-22
Rename DataFrame Sort methods to OrderBy


eerhardt:
This matches more closely with Linq methods that .NET developers are used to.

https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.orderby
https://docs.microsoft.com/en-us/dotnet/api/system.linq.enumerable.orderbydescending

jwood803:
@eerhardt I can work on this. Is it simply just renaming the `Sort` methods to `OrderBy`?

eerhardt:
> Is it simply just renaming the Sort methods to OrderBy?

I think we should also remove the `ascending` parameter and add an `OrderByDescending` method, to match Linq.

jwood803:
@eerhardt Just curious, is anything else needed for this? I can work on any updates if they are needed.

eerhardt:
I don't think anything else is needed for this. #2814 resolved this. Closing.

PR #2814, opened 2020-01-18
Rename Sort to OrderBy

jwood803:
Updates for #2788.

Also added the `OrderByDescending` function.

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -255,7 +255,7 @@ namespace Microsoft.Data.Analysis
         }
 
         /// <summary>
-        /// Orders the data frame by a specified column.
+        /// Sorts the rows of the DataFrame in ascending order according to the values in <param name="columnName">
         /// </summary>
         /// <param name="columnName">The column name to order by</param>
         public DataFrame OrderBy(string columnName)
@@ -264,7 +264,7 @@ namespace Microsoft.Data.Analysis
         }
 
         /// <summary>
-        /// Orders the data frame by a specified column in descending order.
+        /// Sorts the rows of the DataFrame in descending order according to the values in <param name="columnName">
         /// </summary>
         /// <param name="columnName">The column name to order by</param>
         public DataFrame OrderByDescending(string columnName)
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -255,7 +255,7 @@ namespace Microsoft.Data.Analysis
         }
 
         /// <summary>
-        /// Sorts the rows of the DataFrame in ascending order according to the values in <param name="columnName">
+        /// Orders the data frame by a specified column.
         /// </summary>
         /// <param name="columnName">The column name to order by</param>
         public DataFrame OrderBy(string columnName)
@@ -264,7 +264,7 @@ namespace Microsoft.Data.Analysis
         }
 
         /// <summary>
-        /// Sorts the rows of the DataFrame in descending order according to the values in <param name="columnName">
+        /// Orders the data frame by a specified column in descending order.
         /// </summary>
         /// <param name="columnName">The column name to order by</param>
         public DataFrame OrderByDescending(string columnName)
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -254,19 +254,14 @@ namespace Microsoft.Data.Analysis
             return ret;
         }
 
-        public DataFrame Sort(string columnName, bool ascending = true)
+        public DataFrame OrderBy(string columnName)
         {
-            DataFrameColumn column = this[columnName];
-            DataFrameColumn sortIndices = column.GetAscendingSortIndices();
-            List<DataFrameColumn> newColumns = new List<DataFrameColumn>(Columns.Count);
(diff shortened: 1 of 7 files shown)


jwood803:
@eerhardt Was just wondering, should we do the same for the `DataFrameColumn` to keep these two consistent?

eerhardt:
> Was just wondering, should we do the same for the DataFrameColumn to keep these two consistent?

Yes, I think it would make sense to keep the naming consistent across the whole library.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -254,19 +254,22 @@ public DataFrame Description()
             return ret;
         }
 
-        public DataFrame Sort(string columnName, bool ascending = true)
+        /// <summary>
+        /// Orders the data frame by a specified column.
```
pgovind: Does this sound better for the documentation here?:

`Sorts the rows of the dataframe in ascending order according to the values in <param name="columnName">`

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
```diff
@@ -169,10 +169,10 @@ public IEnumerator<string> GetEnumerator()
 
         public override DataFrameColumn Filter<U>(U lower, U upper) => throw new NotSupportedException();
 
-        public new StringDataFrameColumn Sort(bool ascending = true)
+        public new StringDataFrameColumn OrderBy()
         {
             PrimitiveDataFrameColumn<long> columnSortIndices = GetAscendingSortIndices();
-            return Clone(columnSortIndices, !ascending, NullCount);
+            return Clone(columnSortIndices, false, NullCount);
```
eerhardt: Can you use a named parameter for the hard-coded boolean? It makes the code a lot easier to read/understand:

```C#
return Clone(columnSortIndices, invertMapIndices: false, NullCount);
```

vs.

```C#
return Clone(columnSortIndices, false, NullCount);
```

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.Sort.cs
```diff
@@ -13,10 +13,10 @@ namespace Microsoft.Data.Analysis
     public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn
         where T : unmanaged
     {
-        public new PrimitiveDataFrameColumn<T> Sort(bool ascending = true)
+        public new PrimitiveDataFrameColumn<T> OrderBy()
```
eerhardt: We also should add `OrderByDescending()` here and in `StringDataFrameColumn`. We are losing the capability to sort descending and keep the same type.

eerhardt:
>> Was just wondering, should we do the same for the DataFrameColumn to keep these two consistent?

> Yes, I think it would make sense to keep the naming consistent across the whole library.

@jwood803 - sorry, I didn't 100% think this through when I answered the other day. But now that I see the code, it strikes me that an individual column having the method `OrderBy` and `OrderByDescending` doesn't make that much sense. Typically, when you have a collection of something, the .NET method is named "Sort" when you want to sort that collection without selecting something to use as a comparison.

https://docs.microsoft.com/en-us/dotnet/api/system.array.sort
https://docs.microsoft.com/en-us/dotnet/api/system.collections.generic.list-1.sort

but for Linq, it uses an `OrderBy` method when you want to order the collection based on some selector - for example "order a list of customers based on the Name property".

So now I'm wondering if it would be OK if `DataFrameColumn`'s method is named `Sort` (and possibly takes a `Comparison` and `IComparer`, like Array and List APIs do. And if `DataFrame`'s API is named `OrderBy` since you provide it a selector - which column to order the DataFrame by.

Thoughts?

/cc @pgovind - thoughts?

pgovind:
> So now I'm wondering if it would be OK if DataFrameColumn's method is named Sort (and possibly takes a Comparison and IComparer, like Array and List APIs do. And if DataFrame's API is named OrderBy since you provide it a selector - which column to order the DataFrame by.

I agree. I like retaining `DataFrameColumn.Sort(ascending = false)` and renaming `DataFrame.Sort` alone to `DataFrame.OrderBy`. 
Also, I think we shouldn't add an `IComparer` to `Sort` in this PR. Adding `IComparer` to `Sort` might require prototyping to minimize boxing (`DataFrameColumn` is weakly typed while `PrimitiveDataFrameColumn` has a type `T` in it). Do feel free to open another issue and put up a prototype though. At the moment, we just use `Comparer<T>.Default` internally in `PrimitiveDataFrameColumn<T>.Sort`

jwood803:
I'm cool with not changing the `DataFrameColumn` back to `Sort`. Would matching pandas to call it `SortValues` be any better, though? Just a thought when I looked at its API. No strong feelings towards it :)

pgovind:
> I'm cool with not changing the DataFrameColumn back to Sort. Would matching pandas to call it SortValues be any better, though? Just a thought when I looked at its API. No strong feelings towards it :)

Wait, maybe you have a typo? I think we want to keep `DataFrameColumn.Sort` and change `DataFrame.Sort`->`DataFrame.OrderBy` and DataFrame.OrderByDescending`. I wouldn't change to `SortValues` yet. I don't think we should give the impression that there's more than 1 way of sorting(indexes being the only other way I can think of, and we don't support indexes yet).

jwood803:
@eerhardt @pgovind Apologies for the delay. I believe I got things updated back to what they should. If not, just let me know and I"ll get it fixed. :)

pgovind:
LGTM. Thanks @jwood803 !

PR #2818, opened 2020-01-22
Bump the package version so the builds are sorted right

bartonjs:
A change about a year ago (acc1d8fcc07e48e971170897e683ab202e491afe) changed the preview version identifier from `-preview2-<date>-#` to `-e<date>-#`, which means packages that haven't bumped their version have been producing newer packages with "lower" versions than the build on 2019-01-17.

This change bumps the default package version from 0.1.1 to 0.1.2, which will put all newer builds on top.  Packages that don't use the default version are unaffected.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,7 +4,7 @@
     <RepoRoot>$(MSBuildThisFileDirectory)</RepoRoot>
     <VersionSuffix Condition="$(VersionSuffix) == ''">$([System.DateTime]::Now.ToString(preview2-yyMMdd-1))</VersionSuffix>
     <Copyright>&#169; Microsoft Corporation. All rights reserved.</Copyright>
-    <VersionPrefix>0.1.1</VersionPrefix>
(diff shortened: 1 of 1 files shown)


bartonjs:
https://dotnet.myget.org/feed/dotnet-corefxlab/package/nuget/System.Azure.Experimental should show the ordering inconsistency.

Santi, it looks like you tried to bump the version and scheme together; but somehow this package had two builds published with only half the change.  Scanning the gallery shows that it happened to other packages, too.

safern:
cc: @pgovind who is working on using arcade for corefxlab so we will not hit this issue anymore (and we will no longer publish to myget). 

https://github.com/dotnet/corefxlab/pull/2795

pgovind:
Yup, was just about to comment. With #2795, this shouldn't be an issue anymore. I'll clean up #2795 today and try to get it pushed this week.

bartonjs:
Well, at least bumping the version now will mean that the myget feed ends with the last build on top.

Issue #2821, opened 2020-01-24
Append 1 DataFrame to another


pgovind:
DataFrame.Append should allow for appending 2 DataFrames. At the moment, it only support appending 1 row

pgovind:
Fixed by #2823

PR #2823, opened 2020-01-24
Append rows to a DataFrame

pgovind:
Should be an easy review.

### DIFF
--- a/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
+++ b/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
@@ -1955,6 +1955,7 @@ namespace Microsoft.Data.Analysis.Tests
             Assert.Equal(10, df.Rows.Count);
             Assert.Equal(1, df.Columns[0].NullCount);
             Assert.Equal(1, df.Columns[1].NullCount);
+
             DataFrame ret = df.Append(df2.Rows, inPlace: false);
             Assert.Equal(10, df.Rows.Count);
             Assert.Equal(1, df.Columns[0].NullCount);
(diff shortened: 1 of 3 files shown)


tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1947,84 +1947,170 @@ public void TestMutationOnRows()
...
 
-            df.Append();
+            df.Append(inPlace: true);
             Assert.Equal(18, df.Rows.Count);
             Assert.Equal(18, df["Column1"].Length);
             Assert.Equal(18, df["Column2"].Length);
             Assert.Equal(5, df.Columns[0].NullCount);
             Assert.Equal(6, df.Columns[1].NullCount);
+            ret = df.Append(inPlace: false);
+            Assert.Equal(19, ret.Rows.Count);
```
eerhardt: Why does `inPlace: true` return 18 rows, but `inPlace: false` return 19 rows?

pgovind: Ah, you're looking at the wrong dataframe. Maybe I set it up badly. After line 2081, `df` has 18 rows. After 2087, `df` still has 18 rows, but `ret` has a new row appended to it. Hence the `19`

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1947,84 +1947,170 @@ public void TestMutationOnRows()
...
+                    DataFrameRow row = ret.Rows[i];
+                    for (int j = 0; j < check1.Columns.Count; j++)
+                    {
+                        if (i < check1.Rows.Count)
+                        {
+                            Assert.Equal(row[j], check1.Rows[i][j]);
+                        }
+                        else
+                        {
+                            Assert.Equal(row[j]?.ToString(), (check2.Rows[i - check1.Rows.Count][j])?.ToString());
```
eerhardt: Why are you `ToString()`'ing here, but not in the `Assert.Equal` above that is checking `check1`?

pgovind: Because the types don't match(but the values do) as part of testing `Append`. `df.Columns[0]` is `float` while `df2.Columns[1]` is `int`. So the assert would complain that I'm comparing a float(from `row`) and an int(from `check2`). `Check1.Columns[0]` holds `float`s, so the types match and the Assert will succeed. Hence the `ToString`.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
eerhardt: OK, maybe add an assert here that `df` isn't changed by calling `df.Append(inPlace:false)`.

---
In reply to: [371531652](https://github.com/dotnet/corefxlab/pull/2823#discussion_r371531652) [](ancestors = 371531652)

pgovind:
Fixes https://github.com/dotnet/corefxlab/issues/2821

Issue #2815, opened 2020-01-21
DataFrame should support windowing/rolling functions


pgovind:
At the moment, no support for rolling/windowing exists. This is useful when one wants to calculate a windowed sum/average of 3 values for example.

wantsoft:
How can we use it? any instruction?

PR #2827, opened 2020-02-04
Add support for window operations on columns

pgovind:
This patch also adds support for `Rolling` operations on columns.

Fixes 1 part of https://github.com/dotnet/corefxlab/issues/2815. We should add `Expanding` APIs too, but that can come in another PR.


src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
```diff
@@ -141,5 +141,26 @@ protected override void ClearItems()
...
+        /// <param name="name">The name of the column</param>
+        /// <returns><see cref="DataFrameColumn"/> as <typeparamref name="T"/>. Null if types don't match</returns>
+        public T GetColumn<T>(string name)
+            where T : class
+        {
+            int columnIndex = IndexOf(name);
+            if (columnIndex == -1)
+                throw new ArgumentException(Strings.InvalidColumnName, nameof(name));
+            DataFrameColumn column = this[columnIndex];
+            T ret = column as T;
```
gfoidl: Maybe
```c#
DataFrameColumn column = this[columnIndex];

if (column is T ret)
{
    return ret;
}

throw new ArgumentException(...);
```

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
```diff
@@ -0,0 +1,213 @@
...
+            LinkedList<KeyValuePair<double, long>> sortedList = new LinkedList<KeyValuePair<double, long>>();
+            LinkedList<T?> windowValues = new LinkedList<T?>();
+            IDoubleConverter<T> doubleConverter = DoubleConverter<T>.Instance;
+            return _currentColumn.Apply<T>((T? value) =>
+            {
+                double doubleValue = value.HasValue ? doubleConverter.GetDouble(value.Value) : double.MaxValue;
+                // This algorithm builds a sorted list. Time complexity is O(N) here since each value is added/removed to sortedList only once.
+                // If the values at the end of the sortedList are greater than value, delete them since they can never be the min in this window
+                // If value is null, treat it as double.MaxValue
+                if (value.HasValue)
```
gfoidl: Is this needed as `doubleValue` already does this distinction? I.e. it is either the value or `double.MaxValue`, so this `if` can be omitted and both `while`-loops can be "merged" to one (use the first one and it's good).

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
```diff
@@ -0,0 +1,213 @@
...
+                else
+                {
+                    while (sortedList.Count != 0 && sortedList.Last.Value.Key >= double.MaxValue)
+                    {
+                        sortedList.RemoveLast();
+                        windowValues.RemoveLast();
+                    }
+                }
+
+                if (value.HasValue)
```
gfoidl: Same here. Just use `doubleValue`.

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
```diff
@@ -0,0 +1,213 @@
...
+
+        /// <summary>
+        /// Finds the min of the values in this window
+        /// </summary>
+        /// <returns>Returns a DataFrame containing the min in each window</returns>
+        public new PrimitiveDataFrameColumn<T> Min()
+        {
+            _currentIndex = 0;
+
+            LinkedList<KeyValuePair<double, long>> sortedList = new LinkedList<KeyValuePair<double, long>>();
```
gfoidl: Can you init these locals insided the lambda to avoid the closure (allocation)?

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
```diff
@@ -0,0 +1,213 @@
...
+        /// Finds the max of the values in this window
+        /// </summary>
+        /// <returns>Returns a DataFrame containing the max in each window</returns>
+        public new PrimitiveDataFrameColumn<T> Max()
+        {
+            _currentIndex = 0;
+
+            LinkedList<KeyValuePair<double, long>> sortedList = new LinkedList<KeyValuePair<double, long>>();
+            LinkedList<T?> windowValues = new LinkedList<T?>();
+            IDoubleConverter<T> doubleConverter = DoubleConverter<T>.Instance;
```
gfoidl: Same comments as for `Min` apply here.

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
```diff
@@ -0,0 +1,213 @@
...
+
+                return windowValues.Count > 0 ? windowValues.First.Value : null;
+            });
+
+        }
+
+        public override PrimitiveDataFrameColumn<int> Count()
+        {
+            _currentIndex = 0;
+            int count = 0;
```
gfoidl: Init inside lambda to avoid closure.

src/Microsoft.Data.Analysis/PrimitiveColumnContainer.cs
```diff
@@ -251,6 +251,57 @@ public void ApplyElementwise(Func<T?, long, T?> func)
...
+                    long curIndex = i + prevLength;
+                    bool isValid = IsValid(nullBitMapSpan, i);
+                    if (list.Count == windowSize)
+                    {
+                        list.RemoveFirst();
+                        list.AddLast(isValid ? span[i] : default(T?));
+                    }
+                    else
+                    {
+                        list.AddLast(isValid ? span[i] : default(T?));
```
gfoidl: Can be outside the `if`? It's the same as in the true-branch.

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
```diff
@@ -169,12 +170,40 @@ public IEnumerator<string> GetEnumerator()
...
+            {
+                string value = this[i];
+                if (values.Count == windowSize)
+                {
+                    values.RemoveFirst();
+                    values.AddLast(value);
+                }
+                else
+                {
+                    values.AddLast(value);
```
gfoidl: Can be outside the `if`? It's the same as in the true-branch.

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
pgovind: Yup, my bad. I added the `doubleValue` at the end and didn't realize this could be cleaned up.

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
pgovind: But these are not local for each iteration in `Apply` though? `sortedList` needs to be global for the algorithm to work. If I moved `sortedList` inside the lambda, it would be allocated for each iteration in `Apply`?

src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -238,6 +238,11 @@ public virtual StringDataFrameColumn Info()
         /// </summary>
         public virtual DataFrameColumn Description() => throw new NotImplementedException();
 
+        public virtual DataFrameColumnWindow Rolling(int windowsSize)
```
eerhardt: This needs XML doc comments

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -203,6 +203,23 @@ public void TestIndexer()
             Assert.Throws<ArgumentException>(() => dataFrame["Int5"]);
         }
 
+        [Fact]
+        public void GetColumnTests()
+        {
+            DataFrame dataFrame = MakeDataFrameWithAllColumnTypes(10);
+            PrimitiveDataFrameColumn<int> ints = dataFrame.Columns.GetColumn<PrimitiveDataFrameColumn<int>>("Int");
```
eerhardt: I wonder if this API wouldn't be better:

```C#
PrimitiveDataFrameColumn<int> ints = dataFrame.Columns.GetPrimitiveColumn<int>("Int");
```

src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
```diff
@@ -141,5 +141,26 @@ protected override void ClearItems()
             _columnNames.Clear();
             _columnNameToIndexDictionary.Clear();
         }
+
+        /// <summary>
+        /// Searches for a <see cref="DataFrameColumn"/> with the specified <paramref name="name"/> and attempts to return it as <typeparamref name="T"/>. If <typeparamref name="T"/> does not match <see cref="DataFrameColumn.DataType"/>, returns null.
+        /// </summary>
+        /// <typeparam name="T">The type to cast the column to</typeparam>
+        /// <param name="name">The name of the column</param>
+        /// <returns><see cref="DataFrameColumn"/> as <typeparamref name="T"/>. Null if types don't match</returns>
```
eerhardt: Your doc and your behavior doesn't seem to match. The doc says it returns null if the types don't match. But the code throws an ArgumentException.

src/Microsoft.Data.Analysis/DoubleConverter.cs
```diff
@@ -30,7 +30,7 @@ public static IDoubleConverter<T> GetDoubleConverter<T>()
             }
             else if (typeof(T) == typeof(decimal))
             {
-                throw new NotImplementedException();
+                return (IDoubleConverter<T>)new DecimalDoubleConverter();
```
eerhardt: Is there a test for this change?

eerhardt:
> using System;

copyright on all code files

---
Refers to: src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs:1 in 83b9e30. [](commit_id = 83b9e3063b11e04827101ef71aa8f1a01e8225d3, deletion_comment = False)

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
```diff
@@ -0,0 +1,174 @@
...
+using System.Collections.Generic;
+using System.Diagnostics;
+
+namespace Microsoft.Data.Analysis
+{
+    /// <summary>
+    /// A window for <see cref="PrimitiveDataFrameColumn{T}"/> to support rolling window operations
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class PrimitiveDataFrameColumnRollingWindow<T> : DataFrameColumnWindow
```
eerhardt: The file name and the class name should match. That way someone can fine the file a class is in

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
```diff
@@ -169,12 +170,36 @@ public IEnumerator<string> GetEnumerator()
 
         public override DataFrameColumn Filter<U>(U lower, U upper) => throw new NotSupportedException();
 
+        public new StringDataFrameColumnRollingWindow Rolling(int windowSize)
+        {
+            return new StringDataFrameColumnRollingWindow(windowSize, this);
```
eerhardt: What about ArrowStringDataFrameColumn? Should it implement a `Rolling` method?

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
```diff
@@ -169,12 +170,36 @@ public IEnumerator<string> GetEnumerator()
 
         public override DataFrameColumn Filter<U>(U lower, U upper) => throw new NotSupportedException();
 
+        public new StringDataFrameColumnRollingWindow Rolling(int windowSize)
```
eerhardt: You should also be overriding the base `Rolling` method.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
```diff
@@ -328,6 +328,11 @@ public override string ToString()
             return $"{Name}: {_columnContainer.ToString()}";
         }
 
+        public new PrimitiveDataFrameColumnRollingWindow<T> Rolling(int windowSize)
```
eerhardt: override the base `Rolling` method as well.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: That was my first idea. I wasn't sure if we were ok with adding new APIs for every new column type we come up with? I would be ok with introducing `GetPrimitiveColumn`, `GetStringColumn` and `GetArrowStringColumn`.

src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
pgovind: Word wrap issues :/. Fixed

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -2136,5 +2153,72 @@ public void TestApply()
...
+                if (windowValues.Count < 2)
+                {
+                    return null;
+                }
+                return windowValues.Sum();
+            });
+
+            int?[] verifySum = { null, 3, 5, 7, 9, 9, 4, null, -1 };
+            PrimitiveDataFrameColumn<int> verifySumColumn = new PrimitiveDataFrameColumn<int>("Verify", verifySum);
+            Assert.True(verifySumColumn.ElementwiseEquals(sum).All());
```
eerhardt: I wonder if we shouldn't have a helper method that does this comparison, and that helper method would be able to spit out detailed failure information about which values didn't match, at what index, etc.

The way it is written now, if a failure happens, all the diagnostic information you get is "Expected True, got False".

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
eerhardt: For new column types that don't ship inside the same assembly, they can add extension methods.

From a callers stand-point, I would rather write `dataFrame.Columns.GetPrimitiveColumn<int>` vs `dataFrame.Columns.GetColumn<PrimitiveDataFrameColumn<int>>`. For one - there isn't a need for a nested generic type and two it is shorter.

src/Microsoft.Data.Analysis/DataFrameColumnWindow.cs
```diff
@@ -0,0 +1,31 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data.Analysis
+{
+    /// <summary>
+    /// A window object that can be used for expanding and rolling computations
```
eerhardt: What does `expanding` computation mean? Do we have that ability?

src/Microsoft.Data.Analysis/DataFrameWindow.Rolling.PrimitiveDataFrameColumn.cs
gfoidl: Ah, I didn't realize that `_currentColumn.Apply<T>` is an iteration. The you are right.

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
pgovind: Added a `ArrowStringDataFrameColumnRollingWindow` now but it's super limited considering `ArrowStringDataFrameColumn` is immutable

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
pgovind: Done.

src/Microsoft.Data.Analysis/DataFrameColumnWindow.cs
pgovind: Expanding is another type of window computation. A rolling window slides over the values with the window size constant. An expanding window expands as it slides over the values. So at an index `i`, it would include everything from 0..i. We don't support it yet. You're right, I should change the doc to be more clear.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: I'll set this up tomorrow. I think the rest of the comments are addressed.

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumnRollingWindow.cs
```diff
@@ -0,0 +1,28 @@
...
+            _currentColumn = currentColumn;
+        }
+
+        public override PrimitiveDataFrameColumn<int> Count() => StringDataFrameColumnRollingWindow.CountImplementation(_currentIndex, _currentColumn, _windowSize);
+
+        public override DataFrameColumn Max() => throw new NotSupportedException();
+
+        public override DataFrameColumn Min() => throw new NotSupportedException();
+
+        // ArrowStringDataFrameColumnRollingWindow does not get an Apply function because ArrowStringDataFrameColumn is immutable
```
eerhardt: I'm not following this. The `Apply` function doesn't mutate the current column. But instead it returns a new instance of the column, right? We should be able to create a new instance given the values, or even create a `StringDataFrameColumn`. Maybe this could be logged as a separate issue, but this should be possible.

eerhardt:
> using System;

copyright on all code files

---
Refers to: src/Microsoft.Data.Analysis/ArrowStringDataFrameColumnRollingWindow.cs:1 in f3ab53b. [](commit_id = f3ab53bc36a4be8830e2c72bbbd923b1d219ae8b, deletion_comment = False)

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumnRollingWindow.cs
```diff
@@ -0,0 +1,28 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using System.Xml.Schema;
+
+namespace Microsoft.Data.Analysis
+{
+    public class ArrowStringDataFrameColumnRollingWindow : DataFrameColumnWindow
```
eerhardt: xml doc comments.

src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
```diff
@@ -141,5 +141,70 @@ protected override void ClearItems()
...
+            }
+
+            throw new ArgumentException(string.Format(Strings.BadColumnCast, column.DataType, typeof(string)));
+        }
+
+        /// <summary>
+        /// Searches for an <see cref="ArrowStringDataFrameColumn"/> with the specified <paramref name="name"/> and attempts to return it as an <see cref="ArrowStringDataFrameColumn"/>. If <see cref="DataFrameColumn.DataType"/> is not of type string, an exception is thrown.
+        /// </summary>
+        /// <param name="name">The name of the column</param>
+        /// <returns><see cref="ArrowStringDataFrameColumn"/>. Throws an exception if types don't match</returns>
```
eerhardt: FYI - the typical way of documenting an exception is:

```
/// <exception cref="ArgumentException">When it throws</exception>
```

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumnRollingWindow.cs
```diff
@@ -0,0 +1,178 @@
...
+    /// <summary>
+    /// A window for <see cref="PrimitiveDataFrameColumn{T}"/> to support rolling window operations
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class PrimitiveDataFrameColumnRollingWindow<T> : DataFrameColumnWindow
+        where T : unmanaged
+    {
+        private int _windowSize;
+        private PrimitiveDataFrameColumn<T> _currentColumn;
+        private long _currentIndex;
```
eerhardt: Why is `_currentIndex` a field of the class? It feels like it should be a local variable in the methods that use it.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -203,6 +203,23 @@ public void TestIndexer()
...
+            Assert.NotNull(ints);
+            Assert.Throws<ArgumentException>(() => dataFrame.Columns.GetPrimitiveColumn<float>("Int"));
+
+            StringDataFrameColumn strings = dataFrame.Columns.GetStringColumn("String");
+            Assert.NotNull(strings);
+            Assert.Throws<ArgumentException>(() => dataFrame.Columns.GetStringColumn("String"));
+
+            ArrowStringDataFrameColumn arrowStrings = dataFrame.Columns.GetArrowStringColumn("ArrowString");
+            Assert.NotNull(arrowStrings);
+            Assert.Throws<ArgumentException>(() => dataFrame.Columns.GetArrowStringColumn("ArrowString"));
```
eerhardt: Why are you expecting this to be NotNull above, but expecting it to throw here?

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumnRollingWindow.cs
pgovind: Initially I was going to have just 1 `RollingWindow` object and reset the `_currentColumn` and `_currentIndex` as needed. But, since I'm not adding `Rolling` on `DataFrame` yet, we could change `_currentIndex` to a local.

src/Microsoft.Data.Analysis/DoubleConverter.cs
pgovind: Added one now. We didn't have 1 before because there was no way to reach this code.

src/Microsoft.Data.Analysis/PrimitiveColumnContainer.cs
```diff
@@ -280,7 +325,7 @@ public void Apply<TResult>(Func<T?, TResult?> func, PrimitiveColumnContainer<TRe
                     bool isValid = IsValid(nullBitMapSpan, i);
                     TResult? value = func(isValid ? span[i] : default(T?));
                     resultSpan[i] = value.GetValueOrDefault();
-                    SetValidityBit(resultNullBitMapSpan, i, value != null);
+                    resultContainer.SetValidityBit(resultNullBitMapSpan, i, value != null);
```
pgovind: This was a bug. Fixed it and added a unit test for it now

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
```diff
@@ -328,6 +328,16 @@ public override string ToString()
...
```
pgovind: I wonder if inherit doc works here

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
```diff
@@ -32,6 +32,7 @@ public StringDataFrameColumn(string name, long length) : base(name, length, type
                     buffer.Add(default);
                 }
             }
+            _nullCount = length;
```
pgovind: Another subtle bug that I found and added unit tests for

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: That was a copy-paste error. Fixed

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumnRollingWindow.cs
pgovind: Fixed now, this was my bad. It does get an `Apply` function. I pushed a half complete PR so I could pick it up on another machine. Should have marked it as WIP. This PR is ready for review now

src/Microsoft.Data.Analysis/ArrowStringDataFrameColumnRollingWindow.cs
```diff
@@ -0,0 +1,37 @@
...
+
+namespace Microsoft.Data.Analysis
+{
+    /// <summary>
+    /// A window for <see cref="ArrowStringDataFrameColumn"/> to support rolling window operations
+    /// </summary>
+    public class ArrowStringDataFrameColumnRollingWindow : DataFrameColumnWindow
+    {
+        private int _windowSize;
+        private ArrowStringDataFrameColumn _currentColumn;
```
eerhardt: (nit) `private readonly ArrowStringDataFrameColumn _currentColumn;`

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -2136,5 +2159,133 @@ public void TestApply()
                 Assert.Equal(newCol[i], values[i] + 0.5d);
             }
         }
+
+        private void RollingColumnWindowVerifyElementwiseEquals<T>(PrimitiveDataFrameColumn<T> verify, PrimitiveDataFrameColumn<T> values)
```
eerhardt: This isn't really "RollingColumnWindow" specific, is it? Can't this just be "VerifyElementwiseEquals(PrimitiveDataFrameColumn<T> expected, PrimitiveDataFrameColumn<T> actual)`?

src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
```diff
@@ -141,5 +141,73 @@ protected override void ClearItems()
...
+        /// Searches for a <see cref="PrimitiveDataFrameColumn{T}"/> with the specified <paramref name="name"/> and attempts to return it as <see cref="PrimitiveDataFrameColumn{T}"/>. If <typeparamref name="T"/> does not match <see cref="DataFrameColumn.DataType"/>, an exception is thrown.
+        /// </summary>
+        /// <typeparam name="T">The type to cast the column to</typeparam>
+        /// <param name="name">The name of the column</param>
+        /// <returns><see cref="PrimitiveDataFrameColumn{T}"/> with <paramref name="name"/></returns>
+        /// <exception cref="ArgumentException">Throws this exception if types don't match</exception>
+        public PrimitiveDataFrameColumn<T> GetPrimitiveColumn<T>(string name)
+            where T : unmanaged
+        {
+            int columnIndex = IndexOf(name);
```
eerhardt: We should have an indexer by column name, so we don't need to duplicate this logic every time:

```C#
int columnIndex = IndexOf(name);
            if (columnIndex == -1)
            {
                throw new ArgumentException(Strings.InvalidColumnName, nameof(name));
            }
            DataFrameColumn column = this[columnIndex];
```

src/Microsoft.Data.Analysis/PrimitiveColumnContainer.cs
```diff
@@ -251,6 +251,52 @@ public void ApplyElementwise(Func<T?, long, T?> func)
...
+        public void Apply<TResult>(Func<LinkedList<T?>, long, TResult?> func, PrimitiveColumnContainer<TResult> resultContainer, int windowSize)
+            where TResult : unmanaged
+        {
+            LinkedList<T?> list = new LinkedList<T?>();
+            for (int b = 0; b < Buffers.Count; b++)
+            {
+                ReadOnlyDataFrameBuffer<T> buffer = Buffers[b];
+                long prevLength = checked(Buffers[0].Length * b);
+                DataFrameBuffer<T> mutableBuffer = DataFrameBuffer<T>.GetMutableBuffer(buffer);
+                Buffers[b] = mutableBuffer;
```
eerhardt: Why are we setting anything here? Or making a mutableBuffer on the current column container? This shouldn't be modifying the current container at all - it should just be modifying the `resultContainer`.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
eerhardt: It won't work on `new` members. Only override and interface implementations.

---
In reply to: [379294117](https://github.com/dotnet/corefxlab/pull/2827#discussion_r379294117) [](ancestors = 379294117)

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumnRollingWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+{
+    /// <summary>
+    /// A window for <see cref="PrimitiveDataFrameColumn{T}"/> to support rolling window operations
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
+    public class PrimitiveDataFrameColumnRollingWindow<T> : DataFrameColumnWindow
+        where T : unmanaged
+    {
+        private int _windowSize;
+        private PrimitiveDataFrameColumn<T> _currentColumn;
```
eerhardt: (nit) `readonly` here too.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumnRollingWindow.cs
```diff
@@ -0,0 +1,183 @@
...
+                currentIndex++;
+                return count;
+            });
+        }
+    }
+
+    public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn
+        where T : unmanaged
+    {
+        internal PrimitiveDataFrameColumn<U> ApplyRollingFunc<U>(Func<LinkedList<T?>, long, U?> func, int windowSize)
```
eerhardt: Why not put this in the file that contains the rest of the class? It is a bit confusing when a class is spread all over the place - and with other classes in the same file.

src/Microsoft.Data.Analysis/PrimitiveColumnContainer.cs
```diff
@@ -251,6 +251,52 @@ public void ApplyElementwise(Func<T?, long, T?> func)
...
+
+                for (int i = 0; i < span.Length; i++)
+                {
+                    long curIndex = i + prevLength;
+                    bool isValid = IsValid(nullBitMapSpan, i);
+                    if (list.Count == windowSize)
+                    {
+                        list.RemoveFirst();
+                    }
+                    list.AddLast(isValid ? span[i] : default(T?));
```
eerhardt: I'm not sure a LinkedList is the best option here. Every time you call `AddLast` it will allocate a new node:

[https://source.dot.net/#System.Collections/System/Collections/Generic/LinkedList.cs,146](https://source.dot.net/#System.Collections/System/Collections/Generic/LinkedList.cs,146)

Instead, maybe we could have a class that holds a `ColumnContainer`, a `long index` and an `int length` - and it implements `IReadOnlyList<T>`. We give it to the `func` as an `IReadOnlyList`. This would mean we only need to allocate 1 object for the whole `Apply`, and we just update the index and length as needed. When the user tries to get the value - we look it up on the column container.

pgovind:
Not sure if this will make it into 0.3.0. Closing this for now. We can re-open if we have time

PR #2857, opened 2020-03-03
Explode column types and generate converters

pgovind:
First commit of two. Pretty straightforward. This PR explodes the column types and adds converters. The converters are used in the next PR which adds binary operation APIs on the exploded columns.


src/Microsoft.Data.Analysis/Converters.tt
```diff
@@ -0,0 +1,85 @@
...
+<#
+                string capitalizedLoopPrimitive = GetCapitalizedPrimitiveTypes(typeLoop.Key);
+#>
+                return (I<#=typeFirstCharUpper#>Converter<T>)new <#=capitalizedLoopPrimitive#><#=typeFirstCharUpper#>Converter();
+            }
+<#
+    }
+#>
+            throw new NotSupportedException();
+
```
gfoidl: Nit: remove empty line

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.tt
```diff
@@ -28,10 +28,11 @@ namespace Microsoft.Data.Analysis
         public override PrimitiveDataFrameColumn<bool> <#=method.MethodName#>(DataFrameColumn column)
 <# } #>
         {
-            switch (column)
+<#          // The column here could be a PrimitiveColumn or an exploded column like IntDataFrameColumn#>
+            switch (column.DataType)
             {
 <# foreach (TypeConfiguration type in typeConfiguration) { #>
-                case PrimitiveDataFrameColumn<<#=type.TypeName#>> <#=type.TypeName#>Column:
+                case Type <#=type.TypeName#>Type when <#=type.TypeName#>Type == typeof(<#=type.TypeName#>):
```
gfoidl: That's nice 👍

src/Microsoft.Data.Analysis/IntDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data.Analysis
+{
+    public partial class IntDataFrameColumn : PrimitiveDataFrameColumn<int>
```
eerhardt: Int32. We should be using the `System.XXX` type names in APIs. Using the C# names (int, long, short) is confusing to other languages - i.e. F#.

src/Microsoft.Data.Analysis/IntDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+    {
+        public IntDataFrameColumn(string name, IEnumerable<int?> values) : base(name, values) { }
+
+        public IntDataFrameColumn(string name, IEnumerable<int> values) : base(name, values) { }
+
+        public IntDataFrameColumn(string name, long length = 0) : base(name, length) { }
+
+        public IntDataFrameColumn(string name, ReadOnlyMemory<byte> buffer, ReadOnlyMemory<byte> nullBitMap, int length = 0, int nullCount = 0) : base(name, buffer, nullBitMap, length, nullCount) { }
+
+        internal IntDataFrameColumn(PrimitiveDataFrameColumn<int> intColumn) : base(intColumn.Name, intColumn._columnContainer) { }
```
eerhardt: I don't see this ctor being used. Can we remove it until it is necessary?

eerhardt: We would also not need to make `_columnContainer` internal on the PrimitiveDataFrameColumn class.

---
In reply to: [387168118](https://github.com/dotnet/corefxlab/pull/2857#discussion_r387168118) [](ancestors = 387168118)

src/Microsoft.Data.Analysis/FloatDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data.Analysis
+{
+    public partial class FloatDataFrameColumn : PrimitiveDataFrameColumn<float>
```
eerhardt: `SingleDataFrameColumn`.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.cs
```diff
@@ -16,33 +16,33 @@ public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn
     {
         public override DataFrameColumn Add(DataFrameColumn column, bool inPlace = false)
         {
-            switch (column)
+            switch (column.DataType)
             {
-                case PrimitiveDataFrameColumn<bool> boolColumn:
+                case Type boolType when boolType == typeof(bool):
```
eerhardt: I think your existing pattern matching will work correctly here. Why this change?

Also, for the existing code, you don't need the additional cast below. You can just say:

```C#
            switch (column)
            {
                case PrimitiveDataFrameColumn<bool> boolColumn:
                    return AddImplementation(boolColumn, inPlace);
```

pgovind: Right, I thought the same thing. But, we have new column types now, so we need to test against the datatype it holds instead of the column type itself. So `IntDataFrameColumn` and `PrimitiveDataFrameColumn<int>` would both work now

src/Microsoft.Data.Analysis/IntDataFrameColumn.cs
pgovind: Yup. Removed in all the new columns

src/Microsoft.Data.Analysis/IntDataFrameColumn.cs
pgovind: Done.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.cs
eerhardt: I tested this in an isolated environment, and it works:

```C#
    class Primitive<T> where T : struct
    {
    }
    class Int32Column : Primitive<int>
    {
    }
    class Program
    {
        {
            Print(new Int32Column());
            Print(new Primitive<int>());
            Print(new Primitive<bool>());
        }
        private static void Print(object o)
        {
            switch (o)
            {
                case Primitive<int> intColumn:
                    Console.WriteLine("Got int");
                    break;
                case Primitive<long> intColumn:
                    Console.WriteLine("Got long");
                    break;
                default:
                    Console.WriteLine("got nothing");
                    break;
            }
        }
    }
```

prints:
```
Got int
Got int
got nothing
```

src/Microsoft.Data.Analysis/UIntDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data.Analysis
+{
+    public partial class UIntDataFrameColumn : PrimitiveDataFrameColumn<uint>
```
eerhardt: `UInt32`. Same for ULong and UShort.

src/Microsoft.Data.Analysis/ShortDataFrameColumn.cs
```diff
@@ -0,0 +1,25 @@
...
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Microsoft.Data.Analysis
+{
+    public partial class ShortDataFrameColumn : PrimitiveDataFrameColumn<short>
```
eerhardt: `Int16DataFrameColumn`

src/Microsoft.Data.Analysis/Converters.cs
```diff
@@ -0,0 +1,948 @@
...
+    {
+        public static IDoubleConverter<T> Instance { get; } = DoubleConverter.GetDoubleConverter<T>();
+    }
+    internal static class DoubleConverter
+    {
+        public static IDoubleConverter<T> GetDoubleConverter<T>()
+        {
+            if (typeof(T) == typeof(byte))
+            {
+                return (IDoubleConverter<T>)new ByteDoubleConverter();
```
eerhardt: (nit) is the cast necessary here?

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
```diff
@@ -38,6 +38,56 @@
         return $"{keyword} (typeof(T) == typeof({type.TypeName}))";
     }
 
+        // A way to discern implicit conversions. For ex: short has primitivitty 2. int has primitivitty 3. primitivitty(short) + primitivitty(int) > 2 * primitivitty(short) implying that a conversion has to take place
```
eerhardt: Are these changes necessary/intended?

src/Microsoft.Data.Analysis/Converters.cs
pgovind: Yup, an implicit conversion does not exist between IDoubleConverter<T> and ByteDoubleConverter (even though it implements IDoubleConverter<byte>)

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
pgovind: Yup. `Converters.tt` uses this map to decide which APIs to generate.

PR #2861, opened 2020-03-05
Address remaining concerns from the 2nd DataFrame API Review 

pgovind:
All of these changes are from the API review. Pretty straightforward

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.BinaryOperators.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.BinaryOperators.cs
@@ -12,32 +12,12 @@ namespace Microsoft.Data.Analysis
 {
     public partial class DataFrame
     {
-        public static DataFrame operator +(DataFrame df, bool value)
-        {
-            return df.Add(value);
-        }
-        
-        public static DataFrame operator +(bool value, DataFrame df)
-        {
-            return df.ReverseAdd(value);
-        }
-
         public static DataFrame operator +(DataFrame df, byte value)
         {
             return df.Add(value);
         }
         
         public static DataFrame operator +(byte value, DataFrame df)
-        {
-            return df.ReverseAdd(value);
-        }
-
-        public static DataFrame operator +(DataFrame df, char value)
-        {
-            return df.Add(value);
-        }
-        
-        public static DataFrame operator +(char value, DataFrame df)
         {
             return df.ReverseAdd(value);
         }
@@ -142,32 +122,12 @@ namespace Microsoft.Data.Analysis
             return df.ReverseAdd(value);
         }
 
-        public static DataFrame operator -(DataFrame df, bool value)
-        {
-            return df.Subtract(value);
-        }
-        
-        public static DataFrame operator -(bool value, DataFrame df)
-        {
-            return df.ReverseSubtract(value);
-        }
-
         public static DataFrame operator -(DataFrame df, byte value)
         {
             return df.Subtract(value);
         }
         
         public static DataFrame operator -(byte value, DataFrame df)
-        {
-            return df.ReverseSubtract(value);
-        }
-
-        public static DataFrame operator -(DataFrame df, char value)
-        {
-            return df.Subtract(value);
-        }
-        
-        public static DataFrame operator -(char value, DataFrame df)
         {
             return df.ReverseSubtract(value);
         }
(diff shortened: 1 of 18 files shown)


src/Microsoft.Data.Analysis/DataFrame.BinaryOperators.tt
```diff
@@ -21,18 +21,12 @@ namespace Microsoft.Data.Analysis
...
-        {
-            return df.<#=method.MethodName#>(value);
-        }
-
-        public static DataFrame operator <#=method.Operator#>(bool value, DataFrame df)
-        {
-            return df.Reverse<#=method.MethodName#>(value, inPlace: false);
-        }
-
+<#      // There seem to be very few use cases for an API such as df.And/Or/Xor(true). #>
```
pgovind: Maybe replace very few with no use cases

src/Microsoft.Data.Analysis/DataFrame.BinaryOperators.cs
```diff
@@ -32,16 +22,6 @@ public partial class DataFrame
             return df.ReverseAdd(value);
         }
 
-        public static DataFrame operator +(DataFrame df, char value)
```
eerhardt: For conversational sake - could this be useful for appending a character to string columns? Or would they just use a full string in that case?

src/Microsoft.Data.Analysis/DataFrame.BinaryOperators.cs
```diff
@@ -662,36 +562,6 @@ public partial class DataFrame
             return df.ReverseModulo(value);
         }
 
-        public static DataFrame operator &(DataFrame df, bool value)
```
eerhardt: Why are these being removed?

eerhardt:
>         /// Clips values beyond the specified thresholds on numeric columns

Should the summary be updated to not say "clip" anymore?

[https://docs.microsoft.com/en-us/dotnet/api/system.math.clamp?view=netcore-3.1](https://docs.microsoft.com/en-us/dotnet/api/system.math.clamp?view=netcore-3.1)

Uses the summary:

> Returns value clamped to the inclusive range of min and max.

---
Refers to: src/Microsoft.Data.Analysis/DataFrame.cs:251 in 16dc6ea. [](commit_id = 16dc6ea2108e1d64eed3426a430f973cd08a7a26, deletion_comment = False)

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -278,15 +253,15 @@ public DataFrame OrderByDescending(string columnName)
         /// <typeparam name="U"></typeparam>
         /// <param name="lower">Minimum value. All values below this threshold will be set to it</param>
         /// <param name="upper">Maximum value. All values above this threshold will be set to it</param>
-        public DataFrame Clip<U>(U lower, U upper, bool inPlace = false)
+        public DataFrame Clamp<U>(U lower, U upper, bool inPlace = false)
```
eerhardt: Maybe `lower` => `min` and `upper` => `max`?

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1654,30 +1606,30 @@ public void TestMerge()
...
+            Assert.Null( merge.Columns["Int_left"][12]);
+            Assert.Null( merge.Columns["Int_left"][5]);
             VerifyMerge(merge, left, right, JoinAlgorithm.FullOuter);
 
             // Inner merge 
             merge = left.Merge<int>(right, "Int", "Int", joinAlgorithm: JoinAlgorithm.Inner);
             Assert.Equal(9, merge.Rows.Count);
             Assert.Equal(merge.Columns.Count, left.Columns.Count + right.Columns.Count);
-            Assert.Equal(merge["Int_right"][2], right["Int"][2]);
+            Assert.Equal( merge.Columns["Int_right"][2], right.Columns["Int"][2]);
```
eerhardt: Are there unnecessary whitespace changes here?

src/Microsoft.Data.Analysis/DataFrame.BinaryOperators.cs
pgovind: 2 things:
1. This operator is on the DataFrame itself. So it'd be useful only if we had a DataFrame composed of only char columns. Even then, the current behavior is not great: 'a' + 'a' results in whatever character `194` is. 
2. Assuming you mean the same operator but defined on the `DataFrameColumn` type, then no, it wouldn't be useful. We'd just use the full `StringDataFrameColumn` in that case. We don't have the ability to go from a `PrimitiveDataFrameColumn<char>` -> `StringDataFrameColumn` yet.

src/Microsoft.Data.Analysis/DataFrame.BinaryOperators.cs
pgovind: Because it'd only be useful when we have a `DataFrame` composed of only Boolean columns(because `&/|/^ ABoolValue` don't make sense on any other column). I think the feeling in the API review was that it didn't make sense to keep this API for a corner case. The operators are still defined on the base `DataFrameColumn` type though, so a statement like `boolColumn & true` is valid.

src/Microsoft.Data.Analysis/DataFrame.cs
pgovind: Yup. Changed the parameter names in the `Filter` too

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: Yup. Search and replace artifact. Good catch!

PR #2867, opened 2020-03-09
Add binary operations and operators on the exploded columns

pgovind:
There's still room for improvement in the binary operation implementations, but this PR is a step in the right direction. To reduce the number of copies I'm making, I'd have to write another text template in `Primit..ColumnContainer`, but that is a PR for another time. I'd be ok shipping this for now.

The easiest way to review this PR is to not look at the *.cs files, but instead to look at the *.tt files. All the *.tt files (and therefore, the *.cs files) follow the same pattern.


tests/Microsoft.Data.Analysis.Tests/Microsoft.Data.Analysis.Tests.csproj
```diff
@@ -13,4 +13,27 @@
     <ProjectReference Include="..\..\src\Microsoft.Data.Analysis\Microsoft.Data.Analysis.csproj" />
   </ItemGroup>
 
+  <ItemGroup>
+    <None Update="DataFrameColumn.BinaryOperationTests.Bad.tt">
```
pgovind: Remove this item

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -2176,5 +2176,6 @@ void AssertLengthTypeAndValues(DataFrameColumn column, Type type)
             DataFrameColumn ushortColumn = DataFrameColumn.Create("Ushort", Enumerable.Range(0, length).Select(x => (ushort)x));
             AssertLengthTypeAndValues(ushortColumn, typeof(ushort));
         }
+
```
pgovind: Revert this change

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.Combinations.ttinclude
```diff
@@ -0,0 +1,258 @@
...
+        new List<string> {"ushort", "ulong", "ulong"},
+        new List<string> {"ushort", "ushort", "ushort"},
+    };
+}
+
+public class ComparisonOperationCombinations
+{
+    public static List<List<string>> comparisonOperationCombinations = new List<List<string>>
+    {
+        new List<string> {"bool", "bool", "bool"},
```
eerhardt: Do you really need the `bool` at the end? Won't all comparison operations result in a `bool`?

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.Combinations.ttinclude
```diff
@@ -0,0 +1,258 @@
...
+// See the LICENSE file in the project root for more information.
+
+// Generated from PrimitiveDataFrameColumn.BinaryOperations.Combinations.tt. Do not modify directly
+
+public class BinaryOperationCombinations
+{
+
+    public static List<List<string>> binaryOperationCombinations = new List<List<string>>
+    {
+        new List<string> {"byte", "byte", "byte"},
```
eerhardt: Instead of `List<string>`, would it be more maintainable/readable to use a custom type where you can name each field? Or even at a minimum use a `Tuple<string, string, string>`, so you know there are always 3 strings (no more, no less).

eerhardt:
> <#@ template debug="false" hostspecific="false" language="C#" #>

Is this file still used? It is a `tt` file, but it doesn't have a backing `.cs` file.

---
Refers to: src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.Combinations.tt:1 in 1828b50. [](commit_id = 1828b50732ee12ffe3c9b7c1b36d4eeb6f4b3bf6, deletion_comment = False)

eerhardt:
>     public partial class ByteDataFrameColumn

Can we collapse these multiple `partial class XXX` declarations into a single class declaration in this file?

---
Refers to: src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperators.cs:20 in 1828b50. [](commit_id = 1828b50732ee12ffe3c9b7c1b36d4eeb6f4b3bf6, deletion_comment = False)

eerhardt:
>         public static UShortDataFrameColumn operator +(ByteDataFrameColumn left, UShortDataFrameColumn right)

We are still using the C# names here: `UShort`, `ULong` above. These should be the `System.XXX` names.

---
Refers to: src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperators.cs:78 in 1828b50. [](commit_id = 1828b50732ee12ffe3c9b7c1b36d4eeb6f4b3bf6, deletion_comment = False)

eerhardt:
>         public BoolDataFrameColumn ElementwiseEqualsImplementation(BoolDataFrameColumn column)

I don't think these `XXXImplementation` methods should be public.

---
Refers to: src/Microsoft.Data.Analysis/DataFrameColumn.BinaryOperationImplementations.Exploded.cs:1721 in 1828b50. [](commit_id = 1828b50732ee12ffe3c9b7c1b36d4eeb6f4b3bf6, deletion_comment = False)

eerhardt:
>     public partial class ByteDataFrameColumn

Same question as the other file, can we collapse these `partial class` declarations so the class isn't spread out 10 times in the same file?

---
Refers to: src/Microsoft.Data.Analysis/DataFrameColumn.BinaryOperationImplementations.Exploded.cs:157 in 1828b50. [](commit_id = 1828b50732ee12ffe3c9b7c1b36d4eeb6f4b3bf6, deletion_comment = False)

eerhardt:
>             DoubleDataFrameColumn otherdoubleColumn = column.CloneAsDoubleColumn();

Do we have to use this intermediate column? It seems wasteful to clone the ushort column into a double column, just to do the addition, and then we throw it away.

---
Refers to: src/Microsoft.Data.Analysis/DataFrameColumn.BinaryOperationAPIs.ExplodedColumns.cs:263 in 1828b50. [](commit_id = 1828b50732ee12ffe3c9b7c1b36d4eeb6f4b3bf6, deletion_comment = False)

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.Combinations.ttinclude
pgovind: Good idea. Made a `TypeCombination` class.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.Combinations.ttinclude
pgovind: No we don't, but I found it helpful when reasoning about the text templates. Also, I'm just re-using the `TypeCombination` class from the binary ops now. I could change the code here to remove the 3rd "bool", but that would be a bunch of accompanying changes and result in the same code?

pgovind:
> Is this file still used? It is a tt file, but it doesn't have a backing .cs file.

Yup. That's because it results in a .ttinclude file that it included in other .tt files.

pgovind:
Filed https://github.com/dotnet/corefxlab/issues/2870 to reduce the number of copies we're making

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
```diff
@@ -20,7 +20,7 @@ namespace Microsoft.Data.Analysis
     public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn, IEnumerable<T?>
         where T : unmanaged
     {
-        private PrimitiveColumnContainer<T> _columnContainer;
+        internal PrimitiveColumnContainer<T> _columnContainer;
```
eerhardt: I'm not a huge fan of making fields exposed outside of a class. If we really need to get at this from outside the class, can we make an `internal` property to expose it instead?

eerhardt:
>         public DecimalDataFrameColumn Add(DecimalDataFrameColumn column)

The following doesn't compile:

```
        float a = 1.1f;
        decimal de = 2.2m;
        var c = a - de;
```

Since there is no implicit conversion between Single and Decimal. I think we should probably exclude these and make the developer pick if they want to go to float addition, or decimal addition.

---
Refers to: src/Microsoft.Data.Analysis/DataFrameColumn.BinaryOperationAPIs.ExplodedColumns.cs:187 in 995e99b. [](commit_id = 995e99bf1ce56cd684b67f6fda38cccacb87d3bd, deletion_comment = False)

eerhardt:
>         public DecimalDataFrameColumn Add(DecimalDataFrameColumn column)

See [https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#arithmetic-operators](https://github.com/dotnet/csharplang/blob/master/spec/expressions.md#arithmetic-operators)

My understanding of how this works in C# is that there is an implicit conversion from one type to the other, and then the operator is performed on 2 variables of the same type.

See [https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#implicit-numeric-conversions](https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#implicit-numeric-conversions) for the implicit conversions in C#. I think we should have as similar behavior as possible here.

---
In reply to: [597883511](https://github.com/dotnet/corefxlab/pull/2867#issuecomment-597883511) [](ancestors = 597883511)

---
Refers to: src/Microsoft.Data.Analysis/DataFrameColumn.BinaryOperationAPIs.ExplodedColumns.cs:187 in 995e99b. [](commit_id = 995e99bf1ce56cd684b67f6fda38cccacb87d3bd, deletion_comment = False)

eerhardt:
> <#@ template debug="false" hostspecific="false" language="C#" #>

Can we also get some more "end user" type of functional tests? That way we can ensure the API feels natural, and we can look at some non-generated code for how the operators can be used.

---
Refers to: tests/Microsoft.Data.Analysis.Tests/DataFrameColumn.BinaryOperationTests.tt:1 in 995e99b. [](commit_id = 995e99bf1ce56cd684b67f6fda38cccacb87d3bd, deletion_comment = False)

eerhardt:
>         public BooleanDataFrameColumn ElementwiseEqualsImplementation(BooleanDataFrameColumn column)

These XXXImplementation methods are still public.  (At least that's how it is showing up for me)

---
Refers to: src/Microsoft.Data.Analysis/DataFrameColumn.BinaryOperationImplementations.Exploded.cs:1721 in 995e99b. [](commit_id = 995e99bf1ce56cd684b67f6fda38cccacb87d3bd, deletion_comment = False)

pgovind:
> Since there is no implicit conversion between Single and Decimal.

Fixed. Also fixed all the related combinations: long + ulong(not allowed) etc. We mimic the C# grammar now with 1 exception: `And, Or and Xor` are defined only on bool columns. I figured this is ok since its only a restriction.

> These XXXImplementation methods are still public.

Fixed. Missed 1 spot last time in the template. 

> Can we also get some more "end user" type of functional tests? That way we can ensure the API feels natural, and we can look at some non-generated code for how the operators can be used.

Done.

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
```diff
@@ -88,6 +120,21 @@
         return false;
     }
 
+    // These are a subset of the implicit conversions allowed in C#. They are used to generate the return type for binary ops
```
eerhardt: It might be good to add a reference to the "spec" here: [https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#implicit-numeric-conversions](https://github.com/dotnet/csharplang/blob/master/spec/conversions.md#implicit-numeric-conversions)

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
```diff
@@ -88,6 +120,21 @@
...
+    public Dictionary<string, LinkedList<string>> primitiveTypeToImplicitConversions = new Dictionary<string, LinkedList<string>> {
+        {"sbyte", new LinkedList<string>(new List<string> {"int", "long", "float", "double", "decimal"})},
```
eerhardt: Why `LinkedList` here? why not just `List<string>`? The order in `List<T>` is the priority order.

You may also consider exposing this as `IReadOnlyDictionary<string, IReadOnlyList<string>`. There's no reason this should ever need to change.

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
```diff
@@ -105,28 +152,40 @@
...
-        }
-        int t2Level;
-        if (!primitiveTypeToPrimitivityLevelMap.TryGetValue(t2.TypeName, out t2Level))
+        primitiveTypeToImplicitConversions.TryGetValue(t1.TypeName, out LinkedList<string> t1ImplicitConversions);
+        return t1ImplicitConversions.First() ?? string.Empty;
+    }
+
+    public string GetBinaryOperationReturnType(string t1, string t2)
+    {
+        if (t1 == "long" && t2 == "ulong" || t1== "ulong" && t2== "long")
```
eerhardt: (super nit) whitespacing around `==`

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
```diff
@@ -88,6 +120,21 @@
         return false;
     }
 
+    // These are a subset of the implicit conversions allowed in C#. They are used to generate the return type for binary ops
+    public Dictionary<string, LinkedList<string>> primitiveTypeToImplicitConversions = new Dictionary<string, LinkedList<string>> {
+        {"sbyte", new LinkedList<string>(new List<string> {"int", "long", "float", "double", "decimal"})},
+        {"byte", new LinkedList<string>(new List<string> {"int", "uint", "long", "ulong", "float", "double", "decimal"})},
```
eerhardt: 1. you are missing `short` in these.
2. The way you have `GetBinaryOperationReturnType` implemented, you will need the actual type at the front of this list too. The way you have it today, when you add 2 `byte` columns, you are returning an `int` column. It should be returning a `byte`.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -18,6 +18,16 @@
       <AutoGen>True</AutoGen>
       <DependentUpon>Converters.tt</DependentUpon>
     </None>
+    <None Include="DataFrameColumn.BinaryOperationImplementations.Exploded.cs">
```
eerhardt: Is there a reason these files are called `DataFrameColumn.XXX` but really only apply to `PrimitiveDataFrameColumn`s?

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.Combinations.tt
```diff
@@ -0,0 +1,116 @@
...
+} 
+
+// If the C# spec doesn't allow some implicit conversions, don't define that API
+string returnType = GetBinaryOperationReturnType(type, type2);
+if (returnType == string.Empty)
+{
+    continue;
+}
+#>
+        new TypeCombination("<#=type.TypeName#>", "<#=type2.TypeName#>", "<#=GetBinaryOperationReturnType(type, type2)#>"),
```
eerhardt: (nit) No need to call `GetBinaryOperationReturnType` again here. Can just use `returnType` variable.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.BinaryOperations.Combinations.ttinclude
```diff
@@ -0,0 +1,271 @@
...
+           _thisColumnType = thisColumnType;
+           _otherColumnType = otherColumnType;
+           _returnColumnType = returnColumnType;
+       }
+       public string ThisColumnType => _thisColumnType;
+       public string OtherColumnType => _otherColumnType;
+       public string ReturnColumnType => _returnColumnType;
+}
+
+public class BinaryOperationCombinations
```
eerhardt: (nit) `public static class`. we only have static members on this class.

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
eerhardt: I just looked at the compiler's behavior, and adding 2 bytes returns an `Int32` in C#. So I think I'm wrong here.

---
In reply to: [395296638](https://github.com/dotnet/corefxlab/pull/2867#discussion_r395296638) [](ancestors = 395296638)

src/Microsoft.Data.Analysis/ColumnArithmeticTemplate.ttinclude
pgovind: It was `List<string>` before. I considered `LinkedList<string>` for a while when I wanted to change the order. Back to `List<string>` now.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
pgovind: I think my reasoning was that these are now exploded column types and not `PrimitiveDataFrameColumn`s. I don't have a strong preference here, I can rename the files.

eerhardt: We should be consistent. The operators: `+`, `-`, etc. are in `PrimitiveDataFrameColumn.BinaryOperators`. But these aren't.

IMO, it would make the most sense for them all to be in files named `PrimitiveDataFrameColumn.XXX`, since they only operate on primitive columns.

pgovind: Ok, renamed everything to PrimitiveDataFrameColumn.XXX

PR #2874, opened 2020-03-15
The ToSpan<T> dependency on ResizableArray has been removed.  Resizab…

AlexanderSWilliams:
…leArray is not necessary and in most cases slower than precalculating the exact capacity of an array when it is possible.

### DIFF
--- a/src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
@@ -16,6 +16,7 @@ namespace System.Buffers
             while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
                 capacity += buffer.Length;
 
+            position = sequence.Start;
             var count = 0;
             byte[] array = new byte[capacity];
             while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
@@ -33,6 +34,7 @@ namespace System.Buffers
             while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
                 capacity += buffer.Length;
 
+            position = sequence.Start;
             var count = 0;
             byte[] array = new byte[capacity];
             while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
--- a/src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
+++ b/src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
@@ -12,24 +12,36 @@ namespace System.Buffers
         public static ReadOnlySpan<byte> ToSpan(this ReadOnlySequence<byte> sequence)
         {
             SequencePosition position = sequence.Start;
-            ResizableArray<byte> array = new ResizableArray<byte>(1024);
+            int capacity = 0;
+            while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
+                capacity += buffer.Length;
+
+            var count = 0;
+            byte[] array = new byte[capacity];
             while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
             {
-                array.AddAll(buffer.Span);
+                buffer.Span.CopyTo(array.AsSpan(count));
+                count += buffer.Length;
             }
-            return array.Span;
+            return array.AsSpan();
         }
 
         public static ReadOnlySpan<byte> ToSpan<T>(this T sequence) where T : ISequence<ReadOnlyMemory<byte>>
         {
             SequencePosition position = sequence.Start;
-            ResizableArray<byte> array = new ResizableArray<byte>(1024);
+            int capacity = 0;
+            while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
+                capacity += buffer.Length;
+
+            var count = 0;
+            byte[] array = new byte[capacity];
             while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
             {
-                array.AddAll(buffer.Span);
+                buffer.Span.CopyTo(array.AsSpan(count));
+                count += buffer.Length;
             }
-            array.Resize(array.Count);
-            return array.Span.Slice(0, array.Count);
+
+            return array.AsSpan();
         }
 
         // TODO: this cannot be an extension method (as I would like it to be).
--- a/tests/Benchmarks/System.Buffers/ToSpan.cs
+++ b/tests/Benchmarks/System.Buffers/ToSpan.cs
@@ -1,0 +1,35 @@
+﻿using System;
+using System.Buffers;
+using System.Buffers.Tests;
+using System.Collections.Generic;
+using System.Collections.Sequences;
+using System.Linq;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+
+namespace Benchmarks.System.Buffers
+{
+    public class ToSpan
+    {
+        public static ReadOnlySequence<byte> Sequence = BufferFactory.Create(Enumerable.Repeat(0, 10000000)
+            .Select(x => new Memory<byte>(new byte[] { 0 })));
+
+        [Benchmark]
+        public void NewToSpan()
+        {
+            var Result = Sequence.ToSpan();
+        }
+
+        [Benchmark]
+        public void OldToSpan()
+        {
+            SequencePosition position = Sequence.Start;
+            ResizableArray<byte> array = new ResizableArray<byte>(1024);
+            while (Sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
+            {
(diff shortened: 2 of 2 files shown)


ahsonkhan:
Seems reasonable.

> in most cases slower than precalculating the exact capacity of an array when it is possible.

Can you share some benchmark numbers for comparison?

AlexanderSWilliams:
I just added a worst-case benchmark to my pull request.  The benchmark might have adversely affected the continuous integration build time.  If that's the case, I can revert it.

The main point is that the Resizable array approach has quadratic complexity.  Hopefully, the benchmark I added makes that more clear.

ahsonkhan:
Can you share the result of the benchmark as a comment here?

AlexanderSWilliams:
I changed Enumerable.Repeat(0, 10000000) to Enumerable.Repeat(0, 100000) so that I was testing a ReadOnlySequence<byte> that had 100000 individual byte arrays of length 1.  Below was the output of the benchmark statistics:

|    Method |       Mean |      Error |     StdDev |
|---------- |-----------:|-----------:|-----------:|
| NewToSpan |   2.308 ms |  0.0756 ms |  0.2193 ms |
| OldToSpan | 970.400 ms | 28.5945 ms | 83.4116 ms |

I noticed that by default a ResizableArray starts at length 1024 in the OldToSpan method.  So, testing with Enumerable.Repeat(0, 1024) should give OldToSpan the greatest edge over NewToSpan because in NewToSpan needs to do 1024 additions.   The benchmark confirms that OldToSpan is slightly faster in this case:

|    Method |     Mean |    Error |   StdDev |
|---------- |---------:|---------:|---------:|
| NewToSpan | 17.34 μs | 0.471 μs | 1.336 μs |
| OldToSpan | 16.89 μs | 0.384 μs | 1.090 μs |

However, even at Enumerable.Repeat(0, 2048) NewToSpan is more performant.

|    Method |      Mean |    Error |    StdDev |
|---------- |----------:|---------:|----------:|
| NewToSpan |  35.08 μs | 0.967 μs |  2.837 μs |
| OldToSpan | 236.42 μs | 5.603 μs | 16.076 μs |

tests/Benchmarks/System.Buffers/ToSpan.cs
```diff
@@ -0,0 +1,35 @@
+﻿using System;
```
ahsonkhan, line 1: Add the license header

tests/Benchmarks/System.Buffers/ToSpan.cs
```diff
@@ -0,0 +1,35 @@
...
+using System.Buffers;
+using System.Buffers.Tests;
+using System.Collections.Generic;
+using System.Collections.Sequences;
+using System.Linq;
+using System.Text;
+using BenchmarkDotNet.Attributes;
+
+namespace Benchmarks.System.Buffers
```
ahsonkhan, line 10: Can you please share the comparison of how much memory is allocated/used up for each of those benchmarks, and not just runtime.

You can get that by using the `MemoryDiagnoser` attribute. For example:
https://github.com/dotnet/corefxlab/blob/00da38200bf18b875b013343dadf6b8566385b96/tests/Benchmarks/Microsoft.Experimental.Collections/DictionarySlimRemove.cs#L12-L13

src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
```diff
@@ -12,24 +12,38 @@ public static class Sequence
...
         public static ReadOnlySpan<byte> ToSpan<T>(this T sequence) where T : ISequence<ReadOnlyMemory<byte>>
         {
             SequencePosition position = sequence.Start;
-            ResizableArray<byte> array = new ResizableArray<byte>(1024);
+            int capacity = 0;
+            while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
+                capacity += buffer.Length;
+
+            position = sequence.Start;
+            var count = 0;
```
ahsonkhan, line 38: nit: don't use var here
```suggestion
            int count = 0;
```

src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
```diff
@@ -12,24 +12,38 @@ public static class Sequence
...
+            byte[] array = new byte[capacity];
```
ahsonkhan, line 21: Instead of looping through all the ROSequence segments, could this just be `sequence.Length` which gives you the length of the sequence already?

https://docs.microsoft.com/en-us/dotnet/api/system.buffers.readonlysequence-1.length?view=netcore-3.1

We could even add a check up front that if `sequence.Length > int.Max`, throw invalid operation exception to avoid accidental integer overflows.

```suggestion
            int count = 0;
            byte[] array = new byte[(int)sequence.Length];
```

src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
```diff
@@ -12,24 +12,38 @@ public static class Sequence
         public static ReadOnlySpan<byte> ToSpan(this ReadOnlySequence<byte> sequence)
         {
             SequencePosition position = sequence.Start;
-            ResizableArray<byte> array = new ResizableArray<byte>(1024);
+            int capacity = 0;
+            while (sequence.TryGet(ref position, out ReadOnlyMemory<byte> buffer))
```
ahsonkhan, line 16: Is there any point (based on segment size and number of segments within the ROSequence) where iterating through it twice (in two loops) in your new solution is slower than the extra array allocations and data copies of using ResizableArray?

Maybe if we have 1_024 empty segments, the cost of the iterating twice would be higher than the allocation/data copy (since there would be nothing to copy). Might be worth measuring just to see the performance characteristics of edge cases.

tests/Benchmarks/System.Buffers/ToSpan.cs
```diff
@@ -0,0 +1,35 @@
...
+{
+    public class ToSpan
+    {
+        public static ReadOnlySequence<byte> Sequence = BufferFactory.Create(Enumerable.Repeat(0, 10000000)
+            .Select(x => new Memory<byte>(new byte[] { 0 })));
```
ahsonkhan, line 15: You measured `ROS<byte>` with 1_024 segments each of size 1. That isn't going to be common in practice. I would recommend improving the test cases in this benchmark by including more realistic permutations of `ROS<byte>` with 16, 64, 512 segments of size 1_024, 4096 bytes (leaving the ResizableArray starting size fixed at 1_024).

I am confident the new approach in your PR will be faster regardless (even more so if you use the constant time `Length` property rather than calculating it yourself), but having such details within the benchmark will be helpful.

tests/Benchmarks/System.Buffers/ToSpan.cs
```diff
@@ -0,0 +1,35 @@
...
+
+        [Benchmark]
+        public void NewToSpan()
+        {
+            var Result = Sequence.ToSpan();
```
ahsonkhan, line 20: nit: instead of assigning it to a local, return it, if possible.

Also, don't use var since it isn't super clear from the right hand side of the expression, what the type is.
```suggestion
            ReadOnlySpan<byte> result = Sequence.ToSpan();
```
OR
```suggestion
            return Sequence.ToSpan();
```

ahsonkhan (approved):
Other than using the `Length` property and adding a license header, looks good.

tests/Benchmarks/System.Buffers/ToSpan.cs
AlexanderSWilliams: Which license file should be added?

src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs
AlexanderSWilliams: 1024 segments gives the greatest disparity between the two methods.  Using ResizableArray is only slightly faster in that case.  For the smaller values, there is no discernible difference

ahsonkhan: What happens if the segments are empty?

There are two factors, **segment size** and **number of segments**. I understand that 1024 number of segments give the greatest disparity and how lowering that doesn't impact the result, but what about changing the **segment size** all the way down to 0?

tests/Benchmarks/System.Buffers/ToSpan.cs
ahsonkhan: See other files in the repo. Each file should have the following text at the top of the file:
```csharp
// Licensed to the .NET Foundation under one or more agreements.
```

For example:
https://github.com/dotnet/corefxlab/blob/ce5c992ecfc1b0dc823a2fd20e4865d2e455b693/src/System.Buffers.Experimental/System/Buffers/Sequences/SequenceExtensions.cs#L1-L4

PR #2878, opened 2020-03-20
Add APIs to get the strongly typed columns from a DataFrame

pgovind:
Add APIs to DataFrameColumnCollection to get the strongly typed columns

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
+++ b/src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
@@ -176,11 +176,11 @@ namespace Microsoft.Data.Analysis
         }
 
         /// <summary>
-        /// Searches for an <see cref="PrimitiveDataFrameColumn{T}"/> with the specified <paramref name="name"/> and attempts to return it as an <see cref="PrimitiveDataFrameColumn{T}"/>. If <typeparamref name="T"/> does not match <see cref="PrimitiveDataFrameColumn.DataType"/>, an exception is thrown.
+        /// Gets the <see cref="PrimitiveDataFrameColumn{T}"/> with the specified <paramref name="name"/>.
         /// </summary>
         /// <param name="name">The name of the column</param>
         /// <returns><see cref="PrimitiveDataFrameColumn{T}"/>.</returns>
-        /// <exception cref="ArgumentException">Throws this exception if types don't match</exception>
+        /// <exception cref="ArgumentException">A column named <paramref name="name"/> cannot be found, or if the column's type doesn't match.</exception>
         public PrimitiveDataFrameColumn<T> GetPrimitiveColumn<T>(string name)
             where T : unmanaged
         {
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
```diff
@@ -175,5 +175,278 @@ protected override void ClearItems()
...
+        public PrimitiveDataFrameColumn<T> GetPrimitiveColumn<T>(string name)
+            where T : unmanaged
+        {
+            DataFrameColumn column = this[name];
+            if (column is PrimitiveDataFrameColumn<T> ret)
+            {
+                return ret;
+            }
+
+            throw new ArgumentException(string.Format(Strings.BadColumnCast, column.DataType, typeof(T)), nameof(T));
```
gfoidl: I don't know if it's worth it...

Perf-wise this will prevent this method from being inlined. Not only the `throw`, but also the construction of the exception-message.
So if relevant, move the "throwing" to a ThrowHelper or a local function. Then this (tiny) method will be inlined into the caller.

src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
```diff
@@ -175,5 +175,278 @@ protected override void ClearItems()
             }
         }
 
+        /// <summary>
+        /// Searches for an <see cref="PrimitiveDataFrameColumn{T}"/> with the specified <paramref name="name"/> and attempts to return it as an <see cref="PrimitiveDataFrameColumn{T}"/>. If <typeparamref name="T"/> does not match <see cref="PrimitiveDataFrameColumn.DataType"/>, an exception is thrown.
```
eerhardt: 1. I don't know if I'd use the term `Searches` here. That kind of makes it sound like this is expensive, when it isn't.
2. I don't think we need to put `If <typeparamref name="T"/> does not match <see cref="PrimitiveDataFrameColumn.DataType"/>, an exception is thrown.` in the summary. Which exceptions are thrown and why should be in the `exception` section.

How about 

```
        /// <summary>
        /// Gets the <see cref="PrimitiveDataFrameColumn{T}"/> with the specified name.
        /// </summary>
        /// <param name="name">The name of the column.</param>
        /// <returns>The <see cref="PrimitiveDataFrameColumn{T}"/> with the specified name.</returns>
        /// <exception cref="ArgumentException">A column named <paramref name="name"/> cannot be found, or if the column's type doesn't match.</exception>
```

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -2179,5 +2179,75 @@ public void TestBinaryOperationsOnExplodedNumericColumns()
             Assert.True(reverseInPlace.ElementwiseEquals(ints).All());
             Assert.False(reverseInPlace.ElementwiseEquals(reverse).All());
         }
+
+        [Fact]
+        public void GetColumnTests()
+        {
+            DataFrame dataFrame = MakeDataFrameWithAllColumnTypes(10);
+            PrimitiveDataFrameColumn<int> primitiveInts = dataFrame.Columns.GetPrimitiveColumn<int>("Int");
```
eerhardt: I wonder if we need to put `Column` in the method name.... We are already on a `Column` collection.

```
dataFrame.Columns.GetPrimitive<int>("Age");
dataFrame.Columns.GetInt32("Age");
dataFrame.Columns.GetSingle("Wage");

dataFrame.Columns.GetString("City");
```

src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
pgovind: 1. Hmmm, I guess you are right. "Search" might lead to an impression that this is an expensive method. Maybe "Look up" would have sounded better and implied a O(1) op.

2. Cool. I'll change it.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
pgovind: I'm not opposed to the idea. I wonder if we have naming guidelines here, and what the API review folks would accept

eerhardt: Maybe let's leave it as-is for now, and make a proposal to drop `Column`, and see how folks feel about it.

src/Microsoft.Data.Analysis/DataFrameColumnCollection.cs
pgovind: Filed https://github.com/dotnet/corefxlab/issues/2881 to address this.

Issue #2860, opened 2020-03-04
Update Spark after 0.3.0 is released


pgovind:
Update .NET for Spark with:
1. Strong binary operations
2. Use ArrowStringColumn.`Apply` instead of `ToArrowArray`

eerhardt:
I think we did this, right? Can this be closed?

pgovind:
Indeed. I haven't done a sweep of the issues yet, but closing this.

PR #2889, opened 2020-03-23
Add Apply to ArrowStringDataFrameColumn

pgovind:
Fixes 2 in https://github.com/dotnet/corefxlab/issues/2860

### DIFF
--- a/src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
+++ b/src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
@@ -601,5 +601,24 @@ namespace Microsoft.Data.Analysis
         {
             return StringDataFrameColumn.ElementwiseNotEqualsImplementation(this, column);
         }
+
+        /// <summary>
+        /// Applies a function to all the values
+        /// </summary>
+        /// <param name="func">The function to apply</param>
+        /// <returns>A <see cref="ArrowStringDataFrameColumn"/> containing the new string values</returns>
+        /// <remarks>This function converts from UTF-8 to UTF-16 strings</remarks>
+        public ArrowStringDataFrameColumn Apply(Func<string, string> func)
+        {
+            ArrowStringDataFrameColumn ret = new ArrowStringDataFrameColumn(Name);
+            Encoding encoding = Encoding.UTF8;
+            for (long i = 0; i < Length; i++)
+            {
+                string cur = this[i];
+                string funcResult = func(cur);
+                ret.Append(funcResult != null ? encoding.GetBytes(funcResult) : default(ReadOnlySpan<byte>));
+            }
+            return ret;
+        }
     }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Analysis/ArrowStringDataFrameColumn.cs
```diff
@@ -601,5 +601,24 @@ public override PrimitiveDataFrameColumn<bool> ElementwiseNotEquals(DataFrameCol
...
+        /// <remarks>This function converts from UTF-8 to UTF-16 strings</remarks>
+        public ArrowStringDataFrameColumn Apply(Func<string, string> func)
+        {
+            ArrowStringDataFrameColumn ret = new ArrowStringDataFrameColumn(Name);
+            Encoding encoding = Encoding.UTF8;
+            for (long i = 0; i < Length; i++)
+            {
+                string cur = this[i];
+                string funcResult = func(cur);
+                ret.Append(funcResult != null ? encoding.GetBytes(funcResult) : default(ReadOnlySpan<byte>));
```
pgovind: This does go from UTF8 to UTF16 and back. Not sure if there's any way around this. 1 thing I can think of is to have another API like so: `Apply(Func<ReadOnlySpan<byte>, ReadOnlySpan<byte>) func)`, but that would just mean that a user has to do the UTF8->UTF16 conversion to make sense of the bytes? IMO, since this is a convenience function, I'd be ok with documenting it and accepting the penalty.

gfoidl:
```suggestion
                ret.Append(funcResult != null ? encoding.GetBytes(funcResult) : ReadOnlySpan<byte>.Empty);
```
is IMHO more readable.

gfoidl: > Apply(Func<ReadOnlySpan<byte>, ReadOnlySpan<byte>) func)

This won't work anyway, as {RO}Span is not allowed as type parameter (stack only).

I don't know any simple workaround for UTF8 -> UTF16 either 😞 

> `encoding.GetBytes(funcResult)`

Allocates `byte[]`, so I would be worried about that more. You could use the overload `GetBytes(ROS<char>, Span<byte>)` to avoid this allocation.

eerhardt: We have been prototyping with a Utf8String class - 

https://github.com/dotnet/runtime/blob/f79afdbc3145285bb38113913bf9ca185cb8c699/src/libraries/System.Utf8String.Experimental/ref/System.Utf8String.Experimental.cs#L67

It is still experimental, for now. But at some point it might be good to try it here and see how well it works.

You'd still have to pay the cost of copying the data around, but at least you wouldn't need to transcode it.

PR #2891, opened 2020-03-24
Fix versioning to allow for individual stable packages

pgovind:
This should let us mark individual packages are stable in the future. 

Thanks @safern! 

cc @joperezr @eerhardt @ericstj

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -29,4 +29,7 @@
     <Content Include="$(RepoRoot)THIRD-PARTY-NOTICES.TXT" Pack="true" PackagePath="\" />
     <Content Include="$(RepoRoot)LICENSE" Pack="true" PackagePath="\" />
   </ItemGroup>
+  <PropertyGroup>
+    <SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>
+  </PropertyGroup>
 </Project>
--- a/Directory.Build.targets
+++ b/Directory.Build.targets
@@ -14,8 +14,4 @@
     <_TranslateUrlPattern>(https://dnceng%40dev\.azure\.com/dnceng/internal/_git|https://dev\.azure\.com/dnceng/internal/_git|https://dnceng\.visualstudio\.com/internal/_git|dnceng%40vs-ssh\.visualstudio\.com:v3/dnceng/internal|git%40ssh\.dev\.azure\.com:v3/dnceng/internal)/([^/-]+)-(.+)</_TranslateUrlPattern>
     <_TranslateUrlReplacement>https://github.com/$2/$3</_TranslateUrlReplacement>
   </PropertyGroup>
-
-  <PropertyGroup>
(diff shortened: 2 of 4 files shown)


safern:
> Changes themselves look good, but do keep in mind that this means that if we want to ship some other package as stable it will require a code change every time. 

It would just be once. We should leave the property set to false in each projects' `Directory.Build.props` as I suggested for `Microsoft.Data.Analysis` now. I think shipping stable should be intentional and also, bumping package versions require a code change as well.

pgovind:
> You missed to set it to false for Microsoft.Data.Analysis -- we should have it there always, as only stable versions will be produced if DotNetFinalVersionKind == release

I removed it on purpose because we already produced a 0.3.0 for Microsoft.Data.Analysis, so I didn't want any stable artifacts anymore from this change.... Maybe I'm understanding it wrong: We set `<SuppressFinalPackageVersion>true</SuppressFinalPackageVersion>` in Directory.Build.Props. Setting `DotNetFinalVersionKind == release` shouldn't make any difference now unless each csproj also overrides the `<SuppressFinalPackageVersion>` property to `false` right?

pgovind:
> Changes themselves look good, but do keep in mind that this means that if we want to ship some other package as stable it will require a code change every time. Check my internal reply saying that it might not be that bad to keep the workflow we have where we build everything as stable but only publish the ones we actually care about stabilizing.

I'm leaning towards keeping our existing workflow and not having this change too. From what I can tell, if we accept this PR to have the ability to produce individual stable packages, we'd need 2 PRs each time: 1 to bump up the version and set the `<SuppressFinalPackageVersion>` flag and a follow up to remove the `<SuppressFinalPackageVersion>` flag once the stable artifacts are produced. I'd much rather just set the `DotNetFinalVersionKind == release` property once when we want stable packages and forget about it other times.

safern:
Right, but we should leave `<SuppressFinalPackageVersion>false</SuppressFinalPackageVersion>` in that project so that you don't require code changes every time you want to produce a stable build. 

The way to produce a stable build is by setting: `DotNetFinalVersionKind` to release. That's how all the repos are setup. Having it checked in, won't make it produce stable versions of that package, unless we set DotNetFinalVersionKind to release, if that is not set, non-stable versions are produced for all projects. 

That's why I think this model is better than producing stable versions for all packages always, as we can leave this properties checked in, and it won't require code changes to produce stable version for those packages as we control the build by setting a variable on the build before queuing.

pgovind:
Oh, I see. Yea you are right. That makes my reply to Jose also irrelevant. Cool, I'll make the changes.

eerhardt:
Now that we've shipped `0.3.0` of Microsoft.Data.Analysis, we should bump its VersionPrefix to `0.4.0` right?

joperezr:
Good point @eerhardt, @pgovind we will need one more PR that will do the versioning changes for Microsoft.Data.Analysis

eerhardt:
Created https://github.com/dotnet/corefxlab/pull/2892

PR #2892, opened 2020-03-25
Bump Microsoft.Data.Analysis version to 0.4.0

eerhardt:


src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -5,7 +5,7 @@
     <LangVersion>7.3</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <SuppressFinalPackageVersion>false</SuppressFinalPackageVersion>
-    <VersionPrefix>0.3.0</VersionPrefix>
+    <VersionPrefix>0.4.0</VersionPrefix>
     <IsShipping>true</IsShipping>
```
safern: Nit: Now that you're on this we can remove the `IsShipping` property. That is set to true by default in arcade, and we were using this for  the old mechanism to set stable packages before we adopted arcade in this repo.

eerhardt:
> I'll take a look to the produced package once CI is done 

I haven't looked at the CI since we've updated to arcade, but it looks like the packages aren't being built and the tests don't run. Am I missing something?

safern:
Hmm I'm seeing the same thing. Maybe we need to pass down -package -test in CI to the build script.

Issue #2708, opened 2019-08-05
CI should emit failing test information to the Tests tab 


pgovind:
We have to open the logs at the moment to see the test info.

safern:
Closing as corefxlab will be archived soon.

PR #2893, opened 2020-03-25
Pass in -ci and -test to CI

pgovind:
I think this is all we needed. Putting this up to test what happens in a public build

### DIFF
--- a/corefxlab-base.yml
+++ b/corefxlab-base.yml
@@ -46,7 +46,7 @@ jobs:
                 /p:DotNetArtifactsCategory=$(_DotNetArtifactsCategory)
                 /p:OfficialBuildId=$(BUILD.BUILDNUMBER)          
             - name: _buildArguments
-              value: -Configuration $(_BuildConfig) -ci -sign /p:BuildPackages=true -pack -publish -test $(_InternalBuildArgs)
+              value: -Configuration $(_BuildConfig) -ci -sign -pack -publish $(_InternalBuildArgs)
             - name: _SignType
               value: real
             - group: DotNet-MyGet-Publish
@@ -60,7 +60,7 @@ jobs:
 
           - ${{ if eq(parameters.isOfficialBuild, 'false') }}:
             - name: _buildArguments
-              value: -Configuration $(_BuildConfig) -test -ci
+              value: -Configuration $(_BuildConfig) -test -ci -pack
 
         pool: ${{ parameters.pool }}
 
--- a/corefxlab-base.yml
+++ b/corefxlab-base.yml
@@ -46,7 +46,7 @@ jobs:
                 /p:DotNetArtifactsCategory=$(_DotNetArtifactsCategory)
                 /p:OfficialBuildId=$(BUILD.BUILDNUMBER)          
             - name: _buildArguments
-              value: -Configuration $(_BuildConfig) -ci -sign /p:BuildPackages=true -pack -publish $(_InternalBuildArgs)
+              value: -Configuration $(_BuildConfig) -ci -sign /p:BuildPackages=true -pack -publish -test $(_InternalBuildArgs)
             - name: _SignType
               value: real
             - group: DotNet-MyGet-Publish
@@ -60,7 +60,7 @@ jobs:
 
           - ${{ if eq(parameters.isOfficialBuild, 'false') }}:
             - name: _buildArguments
-              value: -Configuration $(_BuildConfig)
+              value: -Configuration $(_BuildConfig) -test
 
         pool: ${{ parameters.pool }}
 
--- a/corefxlab-base.yml
+++ b/corefxlab-base.yml
@@ -60,7 +60,7 @@ jobs:
 
           - ${{ if eq(parameters.isOfficialBuild, 'false') }}:
             - name: _buildArguments
-              value: -Configuration $(_BuildConfig) -test
+              value: -Configuration $(_BuildConfig) -test -ci
 
         pool: ${{ parameters.pool }}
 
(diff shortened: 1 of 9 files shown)


corefxlab-base.yml
```diff
@@ -46,7 +46,7 @@ jobs:
                 /p:DotNetArtifactsCategory=$(_DotNetArtifactsCategory)
                 /p:OfficialBuildId=$(BUILD.BUILDNUMBER)          
             - name: _buildArguments
-              value: -Configuration $(_BuildConfig) -ci -sign /p:BuildPackages=true -pack -publish $(_InternalBuildArgs)
+              value: -Configuration $(_BuildConfig) -ci -sign /p:BuildPackages=true -pack -publish -test $(_InternalBuildArgs)
```
safern: is the `BuildPackages` property necessary?

safern: This is going to add testing to the official build, do we want that? I would remove it from here as in official builds we usually don't run tests.

pgovind: Oh, I didn't know that. Ok

corefxlab-base.yml
```diff
@@ -60,7 +60,7 @@ jobs:
 
           - ${{ if eq(parameters.isOfficialBuild, 'false') }}:
             - name: _buildArguments
-              value: -Configuration $(_BuildConfig)
+              value: -Configuration $(_BuildConfig) -test -ci
```
safern: add `-pack` here as well so that we exercise building the packages on PR builds as well.

pgovind:
It looks like the CI systems are not able to locate xunit in their nuget cache. Not sure how to fix this?

eerhardt:
How are you seeing the logs? I don't see the test logs being captured anywhere. what is the error?

safern:
He reproed locally. We're looking at it offline. I need to disconnect now for a bit but I'll repro locally and figure it out later.

eerhardt:
> He reproed locally.

We should be capturing the logs as well, so that you don't need to rely on reproing it locally to figure out the issue.

pgovind:
> We should be capturing the logs as well, so that you don't need to rely on reproing it locally to figure out the issue.

We do actually. It's here: https://dev.azure.com/dnceng/public/_build/results?buildId=573155&view=artifacts&type=publishedArtifacts

safern:
We also have this issue: https://github.com/dotnet/corefxlab/issues/2708

I'll take care of it once we figure out this issue.

eng/Versions.props
```diff
@@ -14,7 +14,6 @@
     <SystemThreadingTasksExtensionsVersion>4.5.1</SystemThreadingTasksExtensionsVersion>
     <LibuvVersion>1.9.1</LibuvVersion>
     <TestSdkVersion>15.0.0</TestSdkVersion>
-    <XunitVersion>2.2.0</XunitVersion>
```
pgovind: That's it? :O

eng/Tests.ruleset
```diff
@@ -0,0 +1,22 @@
+<RuleSet Name="Tests ruleset" Description="All Rules are disabled." ToolsVersion="15.0">
```
pgovind: This is something that xunit knows to interpret/use?

safern:
Ok so I found the reason why tests were failing. The reason is because we now use arcade `-test` target to run our test projects, which invokes xunit.console.dll directly. However, we were overriding arcade version to 2.2.0, and from version 2.4.0 I believe the structure of the package changed, adding the TFM to the paths. So arcade was searching for the .dll with `XunitPackageDir\netcoreapp2.0\xunit.console.dll`. 

https://github.com/dotnet/arcade/blob/d799f35d3ae963f3dc935a7dac1d4c791e85d8e6/src/Microsoft.DotNet.Arcade.Sdk/tools/XUnit/XUnit.targets#L55

So what I did is remove the version we were using and use arcade's version. However, that pulls in the `xunit.analyzers` which is goodness to write good tests. There were some build errors, so I added  a `Tests.ruleset` to disable the errors, and we can then progressively fix them.

eng/Tests.ruleset
safern: No, this is something Roslyn understands. 

The newest version of xunit brings in analyzers as well. This is the way to configure roslyn analyzers. Then you can set the severity to error, warning or none, on a per rule, per analyzer basis. So whenever we want to enable these rules, we will have to fix the errors we get by the analyzer.

safern: It was just a smaller change to do this and  then we can enable the rules by chunks.

pgovind: I see. So when we want to turn on one of these rules, we change `Action="None"` -> say `Action="Error"(or Warning)`?

safern: Yeah, or if you don't list them in this file they will take whatever default Action the analyzer defines: http://xunit.net/xunit.analyzers/rules/

tests/System.Buffers.Primitives.Tests/UsageScenarioTests.cs
```diff
@@ -41,8 +41,6 @@ public void CtorSpanOverByteArrayValidCasesWithPropertiesAndBasicOperationsCheck
             Span<byte> span = new Span<byte>(array);
             Assert.Equal(array.Length, span.Length);
 
-            Assert.NotSame(array, span.ToArray());
```
safern: It seems like in 3.0 this behavior changed and span.ToArray returns the same, or the test was wrong.

tests/System.Memory.Polyfill.Tests/System.Memory.Polyfill.Tests.csproj
```diff
@@ -1,6 +1,6 @@
 <Project Sdk="Microsoft.NET.Sdk" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <PropertyGroup>
-    <TargetFrameworks>netcoreapp2.0;netcoreapp3.0</TargetFrameworks>
```
safern: We don't install any runtimes when bootstrapping, which we could by adding an entry to `global.json`, however, I don't think it makes sense to test against an older framework.

Issue #2906, opened 2020-04-28
Passing in DataTypes to DataFrame.LoadCsv loses column name information


pgovind:
This was reported offline to me by Olav Tellefsen.

PR #2907, opened 2020-04-28
Fix https://github.com/dotnet/corefxlab/issues/2906

pgovind:
We were dropping the column names when `DataTypes` was being passed in to `LoadCsv`. 

I suggest looking at the diff with whitespace turned off

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -227,7 +227,7 @@ namespace Microsoft.Data.Analysis
                 {
                     if ((numberOfRowsToRead == -1) || rowline < numberOfRowsToRead)
                     {
-                        if (linesForGuessType.Count < guessRows || guessRows == 0)
+                        if (linesForGuessType.Count < guessRows || (header && rowline == 0))
                         {
                             var spl = line.Split(separator);
                             if (header && rowline == 0)
@@ -235,11 +235,6 @@ namespace Microsoft.Data.Analysis
                                 if (columnNames == null)
                                 {
                                     columnNames = spl;
-                                    if (guessRows == 0)
-                                    {
-                                        ++rowline;
-                                        break;
-                                    }
                                 }
                             }
                             else
@@ -250,7 +245,7 @@ namespace Microsoft.Data.Analysis
                         }
                     }
                     ++rowline;
-                    if (rowline == guessRows)
+                    if (rowline == guessRows || guessRows == 0)
                     {
                         break;
                     }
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -198,55 +198,58 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
                 numberOfRowsToRead++;
+            }
 
             List<DataFrameColumn> columns;
             long streamStart = csvStream.Position;
             // First pass: schema and number of rows.
             using (var streamReader = new StreamReader(csvStream, encoding ?? Encoding.UTF8, detectEncodingFromByteOrderMarks: true, DefaultStreamReaderBufferSize, leaveOpen: true))
             {
                 string line = null;
-                if (dataTypes == null)
```
eerhardt, line 216: Do you remember why we were doing this in the first place? Or was this just a mistake?

eerhardt: In the case of `dataTypes` being passed in, should we stop processing after the header row?

---
In reply to: [416981821](https://github.com/dotnet/corefxlab/pull/2907#discussion_r416981821) [](ancestors = 416981821)

pgovind: Yea I was wondering about it too. I think I was using dataTypes inside as part of "guessing" the types at one point. Either that or it was a mistake made from trying to optimize the csv read.

pgovind: Good idea. I fixed that, and found some other corner cases and added unit tests for them.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -198,55 +198,68 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                line = streamReader.ReadLine();
+                while (line != null)
                 {
-                    line = streamReader.ReadLine();
-                    while (line != null)
+                    if ((numberOfRowsToRead == -1) || rowline < numberOfRowsToRead)
                     {
-                        if ((numberOfRowsToRead == -1) || rowline < numberOfRowsToRead)
+                        if (linesForGuessType.Count < guessRows || guessRows == 0)
```
eerhardt, line 230: Isn't the right check here:

`if (linesForGuessType.Count < guessRows || (header && rowline == 0))`

?

If `guessRows == 0` and `!header`, then there is no reason to get here, right?

pgovind: Yup, your idea is better. The only other thing I had to do was to change 
``` csharp
 if (rowline == guessRows)
                    {
                        break;
                    }
```
to 
``` csharp
 if (rowline == guessRows || guessRows == 0)
                    {
                        break;
                    }
```
Otherwise we'd end up reading the whole stream when guessRows = 0 unnecessarily.

PR #2916, opened 2020-05-15
Improve LoadCsv to handle null values when deducing the column types

### LINKED ISSUE #2915: DataFrame.LoadCsv doesn't handle `null` values when it attempts to guess the column types
As soon as we encounter a `null` value, we deduce the whole value to be a `StringDataFrameColumn` instead of adding a `null` value to the real column type.

pgovind:
Fixes #2915 

Instead of falling back to a `StringDataFrameColumn`, we now deduce the column type correctly and append a `null` value to whatever the column type becomes.

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Data.Analysis
 
                 string val = line[col];
 
-                if (val == "null" || val == "Null")
+                if (val.Equals("null", StringComparison.OrdinalIgnoreCase))
                 {
                     continue;
                 }
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -24,7 +24,7 @@ namespace Microsoft.Data.Analysis
 
                 string val = line[col];
 
-                if (val.Equals("null", StringComparison.OrdinalIgnoreCase))
+                if (string.Equals(val, "null", StringComparison.OrdinalIgnoreCase))
                 {
                     continue;
                 }
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -23,6 +23,12 @@ namespace Microsoft.Data.Analysis
                     throw new FormatException(string.Format(Strings.LessColumnsThatExpected, nbline + 1));
 
                 string val = line[col];
+
+                if (val == "null" || val == "Null")
+                {
+                    continue;
+                }
+
                 bool boolParse = bool.TryParse(val, out bool boolResult);
                 if (boolParse)
                 {
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -463,35 +463,42 @@ namespace Microsoft.Data.Analysis
             bool columnMoveNext = columnEnumerator.MoveNext();
             if (row != null)
             {
-                    // Go through row first to make sure there are no data type incompatibilities
-                    IEnumerator<object> rowEnumerator = row.GetEnumerator();
-                    bool rowMoveNext = rowEnumerator.MoveNext();
-                    List<object> cachedObjectConversions = new List<object>();
-                    while (columnMoveNext && rowMoveNext)
+                // Go through row first to make sure there are no data type incompatibilities
+                IEnumerator<object> rowEnumerator = row.GetEnumerator();
+                bool rowMoveNext = rowEnumerator.MoveNext();
+                List<object> cachedObjectConversions = new List<object>();
+                while (columnMoveNext && rowMoveNext)
+                {
+                    DataFrameColumn column = columnEnumerator.Current;
+                    object value = rowEnumerator.Current;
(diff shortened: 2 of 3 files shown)


src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -463,35 +463,42 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
             bool columnMoveNext = columnEnumerator.MoveNext();
             if (row != null)
             {
-                    // Go through row first to make sure there are no data type incompatibilities
-                    IEnumerator<object> rowEnumerator = row.GetEnumerator();
-                    bool rowMoveNext = rowEnumerator.MoveNext();
-                    List<object> cachedObjectConversions = new List<object>();
-                    while (columnMoveNext && rowMoveNext)
+                // Go through row first to make sure there are no data type incompatibilities
```
pgovind, line 466: Might be easier to view this diff with whitespace turned off

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -23,6 +23,12 @@ private static Type GuessKind(int col, List<string[]> read)
                     throw new FormatException(string.Format(Strings.LessColumnsThatExpected, nbline + 1));
 
                 string val = line[col];
+
+                if (val == "null" || val == "Null")
```
gfoidl, line 27:
```suggestion
                if (val.Equals("null", StringComparison.OrdinalIgnoreCase))
```
Though I don't have a strong opinion on this...
It would also handle `NULL`, so maybe it's a bit safer.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -463,35 +463,42 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
-                        DataFrameColumn column = columnEnumerator.Current;
-                        object value = rowEnumerator.Current;
-                        // StringDataFrameColumn can accept empty strings. The other columns interpret empty values as nulls
-                        if (value is string stringValue && string.IsNullOrEmpty(stringValue) && column.DataType != typeof(string))
+                        if (string.IsNullOrEmpty(stringValue) && column.DataType != typeof(string))
                         {
                             value = null;
                         }
-                        if (value != null)
+                        if (stringValue == "null" || stringValue == "Null")
```
gfoidl, line 481:
```suggestion
                        if (stringValue.Equals("null", StringComparison.OrdinalIgnoreCase))
```
Same here?

src/Microsoft.Data.Analysis/DataFrame.IO.cs
pgovind: Yup. Good idea.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -23,6 +23,12 @@ private static Type GuessKind(int col, List<string[]> read)
                     throw new FormatException(string.Format(Strings.LessColumnsThatExpected, nbline + 1));
 
                 string val = line[col];
+
+                if (val.Equals("null", StringComparison.OrdinalIgnoreCase))
```
eerhardt, line 27: Are you guaranteed `val` is never actually `null` here? If it is, you will get a `NullReferenceException`. Instead, it is usually safer to say `string.Equals(val, "null", StringComparison.OrdinalIgnoreCase)`

gfoidl: Ahh, good catch! This is what I actually wanted to suggest 😉 
My bad...Thanks.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -463,35 +463,42 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
-                        DataFrameColumn column = columnEnumerator.Current;
-                        object value = rowEnumerator.Current;
-                        // StringDataFrameColumn can accept empty strings. The other columns interpret empty values as nulls
-                        if (value is string stringValue && string.IsNullOrEmpty(stringValue) && column.DataType != typeof(string))
+                        if (string.IsNullOrEmpty(stringValue) && column.DataType != typeof(string))
                         {
                             value = null;
                         }
-                        if (value != null)
+                        if (stringValue.Equals("null", StringComparison.OrdinalIgnoreCase))
```
eerhardt, line 481: `else if` here? Or is there a reason to check this even if the above is true?

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -463,35 +463,42 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
+                    DataFrameColumn column = columnEnumerator.Current;
+                    object value = rowEnumerator.Current;
+                    // StringDataFrameColumn can accept empty strings. The other columns interpret empty values as nulls
+                    if (value is string stringValue)
                     {
-                        DataFrameColumn column = columnEnumerator.Current;
-                        object value = rowEnumerator.Current;
-                        // StringDataFrameColumn can accept empty strings. The other columns interpret empty values as nulls
-                        if (value is string stringValue && string.IsNullOrEmpty(stringValue) && column.DataType != typeof(string))
+                        if (string.IsNullOrEmpty(stringValue) && column.DataType != typeof(string))
```
eerhardt, line 477: You know `stringValue` is not null here, so you can change `string.IsNullOrEmpty(stringValue)` => `stringValue.Length == 0`.

PR #2922, opened 2020-05-20
Increment version and stop producing stable packages

pgovind:
Just cleaning up the M.D.A version and a minor update to the README

### DIFF
--- a/README.md
+++ b/README.md
@@ -102,9 +102,9 @@ https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-experimental-
 
 You can add this feed among your NuGet sources and install the packages (keep in mind that packages are pre-release packages).
 
-To produce a stable build for a project, set the following property in the .csproj.
+To produce a stable build for a project, set the following property in the .csproj. After the build artifacts have been produce, consider incrementing the patch version number in the csproj to prevent conflicts.
 ```
-<SuppressFinalPackageVersion>false</SuppressFinalPackageVersion>
+<DotNetFinalVersionKind>release</DotNetFinalVersionKind>
 ```
 ## License
 
--- a/README.md
+++ b/README.md
@@ -102,7 +102,7 @@ https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-experimental-
 
 You can add this feed among your NuGet sources and install the packages (keep in mind that packages are pre-release packages).
 
-To produce a stable build for a project, set the following property in the .csproj. After the build artifacts have been produce, consider incrementing the patch version number in the csproj to prevent conflicts.
+To produce a stable build for a project, set the following property in the .csproj. After the build artifacts have been produced, consider incrementing the minor version number in the csproj to preserve artifact version ordering.
 ```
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -5,8 +5,7 @@
     <LangVersion>7.3</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <SuppressFinalPackageVersion>false</SuppressFinalPackageVersion>
-    <DotNetFinalVersionKind>release</DotNetFinalVersionKind>
-    <VersionPrefix>0.4.0</VersionPrefix>
+    <VersionPrefix>0.4.1</VersionPrefix>
```
eerhardt, line 8: ?? Why not `0.5.0`? We should reserve `0.4.1` in case we need to patch a bad bug in `0.4.0`.

pgovind: because this produces 0.4.1.alpha-something. When we want a stable 0.4.1, we just set `<DotNetFinalVersionKind>release</DotNetFinalVersionKind>` again. The problem with making it `0.5.0` is that we'd start producing `0.5.0.alpha-something` and if we ever make a `0.4.1`, we'd run into the minor issue that `0.4.1` will list below the older `0.5.0.alpha` packages.

README.md
```diff
@@ -102,9 +102,9 @@ https://dev.azure.com/dnceng/public/_packaging?_a=feed&feed=dotnet-experimental-
 
 You can add this feed among your NuGet sources and install the packages (keep in mind that packages are pre-release packages).
 
-To produce a stable build for a project, set the following property in the .csproj.
+To produce a stable build for a project, set the following property in the .csproj. After the build artifacts have been produce, consider incrementing the patch version number in the csproj to prevent conflicts.
```
pgovind, line 105: I should change this from "prevent conflicts" -> "preserve package version ordering"

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
eerhardt: > we'd run into the minor issue that 0.4.1 will list below the older 0.5.0.alpha packages.

This is expected. If we ever made `0.4.1` it would be branched off of `0.4.0`, not `master`.

Read https://semver.org/ to understand when to bump a minor version and when to bump the patch., As we want to add more APIs/features, we want to bump the minor version.

pgovind: Oh, you're thinking about `0.4.1.alpha-` as an official built artifact too. I was thinking `0.4.0` is the current official build, and we'd still have an option to make a `0.4.1` or a `0.5.0` and have them list in the correct order. This is all just to get around the fact that we don't have official release branches in corefxlab. I changed the version to `0.5.0`. If you change your mind and want it back to `0.4.1`, let me know

PR #2929, opened 2020-06-02
Table based Half implementation

pgovind:
An implementation of a IEEE 754 compliant float16  type. We already have a working Half type implemented in corefxlab. This is a slightly different approach in that it relies on pre-computed tables to perform the bit shifts to convert to/from float/Half. I'm calling the new type `TableBasedHalf` and put it in it's own cs files. I have 2 reasons for this:
1) I'm curious about the performance of the 2 approaches (on demand bit shifts vs table based). Having both types here will let us benchmark them easily. But, before I run the benchmarks, I'd like to know the answer to **2**
2) I'm not very familiar with writing intrinsics code, so I don't know if table based lookups can even be implemented efficiently using intrinsics.


tannergooding:
This will take some time to review as the tables are massive, the tables will add some binary bloat and some startup overhead for initializing the tables.

My gut reaction is that, even if this saves some cycles, it likely isn't going to be worthwhile in real-world code.
For scenarios where this code is going to be "hot", there will likely be hardware support for `Half` and we can just use the intrinsic for converting between `half` and `float`/`double`.
For the cases where it isn't hot, the lookup table won't be hot either and will incur a lookup into uncached memory (often multiple due to having so many separate arrays and with them being so large).

src/System.Numerics.Experimental/Half.v1.cs
```diff
@@ -241,39 +241,54 @@ public static bool IsSubnormal(Half value)
                 && ((absValue & ExponentMask) == 0);    // is subnormal (has a zero exponent)
         }
 
-        public static Half Parse(string s, NumberStyles style = DefaultParseStyle, IFormatProvider formatProvider = null)
+        public static Half Parse(string s)
+        {
+            return (Half)float.Parse(s, style: DefaultParseStyle, provider: null);
```
tannergooding, line 246: This isn't quite right and may result in some differences (iirc). Fixing the actual parsing code should be simple though as it "should" just be updating a few numbers representing the buffer size and max digits to parse.

PR #2931, opened 2020-06-10
Add DataFrame object formatter.

dcostea:
New Microsoft.Data.Analysis.Interactive project was created (as suggested).

**DataFrame object formatter** was added. By default:
- Paginated at 10 rows per page 
- Max limit is 10.000 rows

cc @eerhardt @pgovind  @colombod @jonsequitur 

Untitled

### DIFF
--- a/Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
+++ b/Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
@@ -1,4 +1,8 @@
-﻿using System;
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
 using System.Collections.Generic;
 using System.Linq;
 using System.Threading.Tasks;
@@ -20,7 +24,7 @@ namespace Microsoft.Data.Analysis.Interactive
 
                 var uniqueId = DateTime.Now.Ticks;
 
-                var maxMessage = df.Rows.Count > MAX ? $" (showing a max of {MAX} rows!)" : string.Empty;
+                var maxMessage = df.Rows.Count > MAX ? $" (showing a max of {MAX} rows)" : string.Empty;
                 var title = h3[style: "text-align: center;"]($"DataFrame - {df.Rows.Count} rows {maxMessage}");
 
                 var header = new List<IHtmlContent>
@@ -99,7 +103,7 @@ namespace Microsoft.Data.Analysis.Interactive
             static string BuildHideRowsScript(long uniqueId)
             {
                 var script = $"var allRows = document.querySelectorAll('#table_{uniqueId} tbody tr:nth-child(n)'); ";
-                script += "for (var i = 0; i < allRows.length; i++) { allRows[i].style.display='none'; } ";
+                script += "for (let i = 0; i < allRows.length; i++) { allRows[i].style.display='none'; } ";
                 return script;
             }
 
@@ -107,7 +111,7 @@ namespace Microsoft.Data.Analysis.Interactive
             {
                 var script = $"var page = parseInt(document.querySelector('#page_{uniqueId}').innerHTML) - 1; ";
                 script += $"var pageRows = document.querySelectorAll(`#table_{uniqueId} tbody tr:nth-child(n + ${{page * {size} + 1 }})`); ";
-                script += $"for (var j = 0; j < {size}; j++) {{ pageRows[j].style.display='table-row'; }} ";
+                script += $"for (let j = 0; j < {size}; j++) {{ pageRows[j].style.display='table-row'; }} ";
                 return script;
             }
 
--- a/Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
+++ b/Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
@@ -1,0 +1,131 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Linq;
+using System.Threading.Tasks;
+using Microsoft.AspNetCore.Html;
+using Microsoft.DotNet.Interactive;
+using Microsoft.DotNet.Interactive.Formatting;
+using static Microsoft.DotNet.Interactive.Formatting.PocketViewTags;
+
+namespace Microsoft.Data.Analysis.Interactive
+{
+    public class DataFrameKernelExtension : IKernelExtension
+    {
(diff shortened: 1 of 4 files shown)


Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
```diff
@@ -0,0 +1,131 @@
+﻿using System;
```
eerhardt, line 1: Can you add a copyright to this file? You can find one at the top of any existing file in the Microsoft.Data.Analysis library.

Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
```diff
@@ -0,0 +1,131 @@
...
+        public Task OnLoadAsync(IKernel kernel)
+        {
+            Formatter<DataFrame>.Register((df, writer) =>
+            {
+                const int MAX = 10000;
+                const int SIZE = 10;
+
+                var uniqueId = DateTime.Now.Ticks;
+
+                var maxMessage = df.Rows.Count > MAX ? $" (showing a max of {MAX} rows!)" : string.Empty;
```
eerhardt, line 23: I think we can remove the exclamation point here.

Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
```diff
@@ -0,0 +1,131 @@
...
+                writer.Write($"<script>{BuildPageScript(uniqueId, SIZE)}</script>");
+
+            }, "text/html");
+
+            return Task.CompletedTask;
+
+
+            static string BuildHideRowsScript(long uniqueId)
+            {
+                var script = $"var allRows = document.querySelectorAll('#table_{uniqueId} tbody tr:nth-child(n)'); ";
```
eerhardt, line 101: I wonder if there is a more maintainable way of writing this JS code, like a resource or something.
@jonsequitur @colombod ?

Microsoft.Data.Analysis.Interactive/Microsoft.Data.Analysis.Interactive.csproj
```diff
@@ -0,0 +1,18 @@
...
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <PackageReference Include="HtmlAgilityPack" Version="1.11.24" />
+    <PackageReference Include="Microsoft.CodeAnalysis.Analyzers" Version="3.0.0">
+      <PrivateAssets>all</PrivateAssets>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+    </PackageReference>
+    <PackageReference Include="Microsoft.Data.Analysis" Version="0.4.0" />
```
eerhardt, line 14: This should be a `ProjectReference`

Microsoft.Data.Analysis.Interactive/Microsoft.Data.Analysis.Interactive.csproj
```diff
@@ -0,0 +1,18 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <GeneratePackageOnBuild>true</GeneratePackageOnBuild>
```
eerhardt, line 5: This should be removed. In the end, we will only want a single nuget package that contains both `Microsoft.Data.Analysis` and `Microsoft.Data.Analysis.Interactive` libraries.

Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
dcostea: sure

Microsoft.Data.Analysis.Interactive/Microsoft.Data.Analysis.Interactive.csproj
dcostea: indeed

Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
colombod: Not that I know of .. :(

Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
```diff
@@ -0,0 +1,131 @@
...
+                script += "for (var i = 0; i < allRows.length; i++) { allRows[i].style.display='none'; } ";
```
colombod, line 102: avoid using var as they get hoisted

Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
```diff
@@ -0,0 +1,131 @@
...
+                return script;
+            }
+
+            static string BuildPageScript(long uniqueId, int size)
+            {
+                var script = $"var page = parseInt(document.querySelector('#page_{uniqueId}').innerHTML) - 1; ";
```
colombod, line 108: swap all `var` with `let`, this will make this code safer

Microsoft.Data.Analysis.Interactive/Microsoft.Data.Analysis.Interactive.csproj
colombod: if there are 2 dlls it makes it easy to load the Microsoft.Data.Analysis and not the extension one when loading the package. We have sample for that packaging

Microsoft.Data.Analysis.Interactive/DataFrameKernelExtension.cs
dcostea: here I need "page" variable to be reusable, but I see your point.

eerhardt:
> I'm afraid I need some assistance with the next step

Checkout how the .NET Interactive samples are doing it: https://github.com/dotnet/interactive/tree/master/samples/extensions. It may make sense to split out the `nuget` package into its own project that references the `Microsoft.Data.Analysis` library and the `Microsoft.Data.Analysis.Interactive` project, and packages them together into a single nuget.
Then each .csproj that contained code would have `<IsPackable>false</IsPackable>`, so it wouldn't make another nuget package.

jonsequitur:
And by all means, if there's a cleaner approach, let us know.

PR #2933, opened 2020-06-12
Add Microsoft.Data.Analysis.nuget project

dcostea:
Microsoft.Data.Analysis.nuget project was added for including Microsoft.Data.Analysis.Interactive project.

cc @eerhardt


dcostea:
I think someone with write access rights may help to fix the merge conflict resulted from moving the Microsoft.Data.Analysis.Interactive project at the right place inside _src_ folder.
Thank you.

<<<<<<< master
Project("{9A19103F-16F7-4668-BE54-9A1E7A4F7556}") = "Microsoft.Data.Analysis.Interactive", "src\Microsoft.Data.Analysis.Interactive\Microsoft.Data.Analysis.Interactive.csproj", "{0C11386C-E3CC-45EB-B932-9297A8DD6649}"
EndProject
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Data.Analysis.nuget", "src\Microsoft.Data.Analysis.nuget\Microsoft.Data.Analysis.nuget.csproj", "{D4B13DCE-913F-41BE-B132-ED2175848DA4}"
=======
Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.Data.Analysis.Interactive", "Microsoft.Data.Analysis.Interactive\Microsoft.Data.Analysis.Interactive.csproj", "{0C11386C-E3CC-45EB-B932-9297A8DD6649}"
>>>>>>> master

pgovind:
I'll do this later today @dcostea.

dcostea:
> Edit: nvm, it looks like the PR was already rebased.
@pgovind, Is it anything left to do, from my side?

eerhardt:
> <Project Sdk="Microsoft.NET.Sdk">

I should have caught this last time, but we put the original Interactive project in the wrong place. It is in the root, and not under `src`. Can you delete that project (or logically move it to this location)?

---
Refers to: src/Microsoft.Data.Analysis.Interactive/Microsoft.Data.Analysis.Interactive.csproj:1 in fbe6dba. [](commit_id = fbe6dba85b8d058357fa4df727b263f3992ae8b2, deletion_comment = False)

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
```diff
@@ -0,0 +1,17 @@
+<Project Sdk="Microsoft.NET.Sdk">
```
eerhardt, line 1: Can you upload the resulting `.nupkg` from calling `dotnet pack` on this project? I want to take a look at it to see what it all contains.

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
```diff
@@ -0,0 +1,17 @@
...
+
+  <PropertyGroup>
+    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <IsPackable>true</IsPackable>
```
eerhardt, line 5: I think you are going to need to set a property to tell `dotnet pack` not to include the binary from this `.csproj`. I think these 3 properties should do it:

[https://github.com/dotnet/machinelearning/blob/3d11ce42ab108d467ea6076dc14eaded57f46172/pkg/Directory.Build.props#L5-L7](https://github.com/dotnet/machinelearning/blob/3d11ce42ab108d467ea6076dc14eaded57f46172/pkg/Directory.Build.props#L5-L7)

```xml
    <NoBuild>true</NoBuild>
    <IncludeBuildOutput>false</IncludeBuildOutput>
    <IncludeSymbols>false</IncludeSymbols>
```

You will also need to move a bunch of the nuget properties from `Microsoft.Data.Analysis.csproj` to this project, since that is where nuget will read them from now.

```xml
    <VersionPrefix>0.5.0</VersionPrefix>
    <PackageRequireLicenseAcceptance>false</PackageRequireLicenseAcceptance>
    <Description>This package contains easy-to-use and high-performance libraries for data analysis and transformation.</Description>
    <PackageReleaseNotes>Initial preview of robust and extensible types and algorithms for manipulating structured data that supports aggregations, statistical funtions, sorting, grouping, joins, merges, handling missing values and more. </PackageReleaseNotes>
    <PackageTags>ML.NET ML Machine Learning Data Science DataFrame Preparation DataView Analytics Exploration</PackageTags>
```

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -5,14 +5,10 @@
     <LangVersion>7.3</LangVersion>
     <AllowUnsafeBlocks>true</AllowUnsafeBlocks>
     <SuppressFinalPackageVersion>false</SuppressFinalPackageVersion>
-    <VersionPrefix>0.5.0</VersionPrefix>
```
eerhardt, line 8: You still want the Version here as well, so we get the right AssemblyVersion.

eerhardt: Might want to move that to a shared .props file.

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
dcostea: > ```
> NoBuild>true</NoBuild>
>     <IncludeBuildOutput>false</IncludeBuildOutput>
>     <IncludeSymbols>false</IncludeSymbols>
> ```

These settings break the build, you can see here: https://dev.azure.com/dnceng/public/_build/results?buildId=694421&view=results
and here: https://dev.azure.com/dnceng/public/_build/results?buildId=694435&view=results

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
dcostea: I'm not sure where do you want it uploaded. Anyway, I will try to upload it after the new project is in the right place (after changes are accepted).

eerhardt: You can rename it as a `.zip` file and upload it to GitHub:

image

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
eerhardt: You'll need at least

```xml
<IncludeBuildOutput>false</IncludeBuildOutput>
```

So the .dll produced by this .csproj doesn't get included in the resulting .nupkg

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
dcostea: When trying to upload, I get this:

Untitled

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
dcostea: It makes sense. 
I have added and 5 of 6 jobs are successfully run. The OSX x64 Release is failing. Any idea?

src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj
```diff
@@ -0,0 +1,23 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
```
pgovind, line 1: The aim here is to start including MDA.Interactive.dll into MDA.Nuget right? I think the existing MDA.csproj is already creating MDA.Nuget, so we don't need this new csproj? Instead we may need to add the add the properties that @eerhardt suggests, but in such a way to result in MDA.Interactive.dll landing in a place that the notebook can pick up the formatter automatically. Give me a few minutes while I experiment locally.

pgovind: Alright, so I think the right thing to do here is wait for @jonsequitur to chime in. Specifically, does an extension's Nuget package need to have a specific structure for the notebook to pick it up? 

Adding a new csproj here to package both MDA and the formatter is a valid approach. It's just that we now have to point notebooks users to the new nuget package instead of MDA.nupkg right? Ideally, the formatter would ship with MDA.nupkg. That means the formatting code should live within MDA.csproj (not something I'm in favor of at the moment) or we have to use advanced dotnet pack rules: https://docs.microsoft.com/en-us/nuget/reference/msbuild-targets#advanced-extension-points-to-create-customized-package?

jonsequitur: Sorry I missed this comment. We have samples for building extensions here: https://github.com/dotnet/interactive/tree/master/samples/extensions. The approach we'll want to take here is to create a separate project that references both Microsoft.Data.Analysis and Microsoft.DotNet.Interactive. The two projects' outputs get packed into a single NuGet package using a third project like this: https://github.com/dotnet/interactive/blob/master/samples/extensions/Library.nuget/Library.nuget.csproj.

@dcostea, I'm happy to help out here. 

We'll also want to update to the latest Microsoft.DotNet.Interactive packages as we've made some minor breaking changes.

dcostea: Jon, thank you!

Instead of adding
```<ItemGroup>
   <None Include="../Library/bin/$(Configuration)/$(TargetFramework)/*.dll"  Pack="true" PackagePath="lib/$(TargetFramework)" />
  </ItemGroup>
```
I have used ```<IsPackable>false</IsPackable>``` in Microsoft.Data.Analysis.Interactive and Microsoft.Data.Analysis.Interactive csproj files, but I'm not sure this is acceptable.

The dependency on Microsoft.DotNet.Interactive is the latest.

@jonsequitur, could you review the latest commit?

eerhardt:
It looks like the builds are actually failing, but for some reason Azure DevOps is reporting they still pass:

https://dev.azure.com/dnceng/public/_build/results?buildId=721436&view=logs&j=98d43ca7-10c7-5044-d8dc-07206263d511&t=acc458f0-2e0e-5219-28cc-1f016a65ca57&l=250

```
/home/vsts/work/1/s/.dotnet/sdk/5.0.100-preview.6.20310.4/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(198,5): error NU5128: Some target frameworks declared in the dependencies group of the nuspec and the lib/ref folder do not have exact matches in the other location. Consult the list of actions below: [/home/vsts/work/1/s/src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj]
/home/vsts/work/1/s/.dotnet/sdk/5.0.100-preview.6.20310.4/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(198,5): error NU5128: - Add lib or ref assemblies for the netcoreapp3.1 target framework [/home/vsts/work/1/s/src/Microsoft.Data.Analysis.nuget/Microsoft.Data.Analysis.nuget.csproj]
##[error].dotnet/sdk/5.0.100-preview.6.20310.4/Sdks/NuGet.Build.Tasks.Pack/build/NuGet.Build.Tasks.Pack.targets(198,5): error NU5128: (NETCORE_ENGINEERING_TELEMETRY=Build) Some target frameworks declared in the dependencies group of the nuspec and the lib/ref folder do not have exact matches in the other location. Consult the list of actions below:
- Add lib or ref assemblies for the netcoreapp3.1 target framework
```

@pgovind - do you know why the CI infrastructure isn't failing here?

dcostea:
> you can repro this locally by executing `.\build.cmd -pack` from the root of the repo

I saw that too and I didn't know what to think.

See below the result on my machine:
```
Build succeeded.

C:\repos\corefxlab\.dotnet\sdk\5.0.100-preview.6.20310.4\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(198,5): error NU5128: Some target frameworks declared in the dependencies group of the nuspec and the lib/ref folder do not have exact matches in the other location. Consult the list of actions below: [C:\repos\corefxlab\src\Microsoft.Data.Analysis.nuget\Microsoft.Data.Analysis.nuget.csproj]
C:\repos\corefxlab\.dotnet\sdk\5.0.100-preview.6.20310.4\Sdks\NuGet.Build.Tasks.Pack\build\NuGet.Build.Tasks.Pack.targets(198,5): error NU5128: - Add lib or ref assemblies for the netcoreapp3.1 target framework [C:\repos\corefxlab\src\Microsoft.Data.Analysis.nuget\Microsoft.Data.Analysis.nuget.csproj]
    0 Warning(s)
    1 Error(s)
```

dcostea:
@eerhardt , I have installed .NET 5 preview 6 and ./build.cmd looks good, the error does not reproduce anymore.
I have added some unit tests but it seems the build machine doesn't have .NET Core 3.1. I cannot downgrade MDA.Interactive to .NET Core 3.0 to make it build successfully, because it needs .NET Core 3.1. Should I remove the freshly added unit test project?

eerhardt:
>I have added some unit tests but it seems the build machine doesn't have .NET Core 3.1. 

You can add the `3.1` runtime here:

dcostea:
> You can add the `3.1` runtime here:
> 
> https://github.com/dotnet/corefxlab/blob/fb1fe923b19ea052341cff707094b49214d458dd/global.json#L5-L7

@eerhardt , I have the unit tests passing now, but the issue related to .NET 5 is still there. Maybe @pgovind can help with that?

pgovind:
I'll take a look at the failing CI jobs and packing errors tomorrow :) Thanks for the patience on this!

pgovind:
Alright, I took a look at the build failures and @joperezr helped me figure out what was going on. Basically the wrapper MDA.nuget project wouldn't actually build the MDA and MDAI projects. Also, @joperezr pointed out that making a wrapper project to include 2 other projects isn't the preferred approach to do this. Instead, I decided to include MDAI within the MDA nuget package for netcoreapp3.1+ scenarios (but MDAI still lives as it's own project. During `-pack` we just include everything from MDAI). I pushed a commit to the current branch @dcostea (I hope you don't mind!).

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -13,8 +13,24 @@
...
+    <MSBuild Projects="./../Microsoft.Data.Analysis.Interactive/Microsoft.Data.Analysis.Interactive.csproj"
+             Targets="_GetBuildOutputFilesWithTfm"
+             Properties="TargetFramework=$(TargetFramework)">
+      <Output TaskParameter="TargetOutputs" ItemName="BuildOutputInPackage" />
+    </MSBuild>
+  </Target>
+
+  <ItemGroup Condition="'$(TargetFramework)' == 'netcoreapp3.1'">
+    <PackageReference Include="HtmlAgilityPack" Version="1.11.24" />
+    <PackageReference Include="Microsoft.DotNet.Interactive" Version="1.0.0-beta.20154.7" />
```
jonsequitur, line 30: You'll want to update to a newer version. There are some breaking changes, outlined here: https://github.com/dotnet/interactive/issues/512.

eerhardt, line 30: This isn't correct. The `Microsoft.Data.Analysis` package shouldn't have a dependency on `Microsoft.DotNet.Interactive`, on any TFM.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -13,8 +13,24 @@
     <GenerateDocumentationFile>true</GenerateDocumentationFile>
     <!-- Documentation warnings -->
     <NoWarn>$(NoWarn);1591</NoWarn>
+    <TargetsForTfmSpecificBuildOutput>GatherProjectReferenceAssets</TargetsForTfmSpecificBuildOutput>
```
eerhardt, line 16: `TargetsForTfmSpecificBuildOutput` will put the `Microsoft.Data.Analysis.Interactive` assembly in the `lib` folder. We explicitly don't want that. We need it to be in the `interactive-extensions/dotnet` folder in the nupkg.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
joperezr: We added it because this package would now also ship Microsoft.Data.Analysis.Interactive on the lib so we had to also bring in it's dependencies. From what I'm reading in your comments that is not what we wanted apparently and we wanted M.D.A.I to instead be placed in a different folder, in that case should it be packaged as a tool? (in other words, should it carry its dependencies like MDI with it so that it doesn't have a package dependency on that?)

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
joperezr: I see, I believe there was a misunderstanding here as @pgovind seemed to be under the impression that this is where we wanted it to live. In that case, same question as I asked above, how do interactive-extensions/dotnet usually ship? I assume carrying their own dependencies? 

We can still use TargetsForTfmSpecificBuildOutput and just set a different TargetPath or alternatively we could use TargetsForTfmSpecificContentInPackage instead.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
eerhardt: > From what I'm reading in your comments that is not what we wanted apparently and we wanted M.D.A.I to instead be placed in a different folder, in that case should it be packaged as a tool? (in other words, should it carry its dependencies like MDI with it so that it doesn't have a package dependency on that?)

The M.D.A.I. assembly will only be loaded by `dotnet-interactive`, so we don't need to carry those dependencies in our package. Those dependencies will be provided by `dotnet-interactive` itself.

@jonsequitur @colombod - do we have any documentation explaining the dotnet-interaction extension nupkg layout and expectations? I see https://github.com/dotnet/interactive/blob/master/docs/extending-dotnet-interactive.md, but that doesn't explain packaging expectations.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
pgovind: Yup, once @joperezr explained that having a wrapper nuget csproj wasn't a very robust way to packing 2 libs, I thought that left us with 2 options:
1. Ship MDAI and point notebook users to the MDAI package (while non-notebook users still use MDA). I don't think we should ever do this.
2. Ship MDAI with MDA, so the notebook users pick up the formatter automatically. Ideally we could've done this without modifying MDA.csproj, but this is where the robustness issue lies.

Just to be clear, I'm still open to having a wrapper nuget csproj if we decide that we absolutely don't want MDA to have a dependency on `Microsft.DotNet.Interactive`.

I'm just brain-storming here, but considering the nuget package issues(https://github.com/dotnet/interactive/tree/master/samples/extensions seems to be manually copying all the dlls during `pack`?), I wonder if there's another way for a library to say "If I'm loaded in an interactive environment, load these other nuget packages". Maybe some sort of asset(a text file?) in the library nuget package that lists other nuget packages to be downloaded by dotnet-interactive? (I don't know very much about package issues, so this could be a terrible idea!)

joperezr: What you are brainstorming seems to be similar to the convention that there seems to be in place already for stuff located under interactive-extensions/dotnet folder from @eerhardt's comment. So in this case, all we want is to keep stuff similar to what we have them now, except that instead of also adding MDAI into the lib/netcoreapp3.1 folder as we are doing now we should instead place it under the interactive-exteinsions folder and we should also remove the bit of adding the package references that are MDAI specific into the MDA package as from @eerhardt response this seems to be already present inbox on dotnet interactive, so no need to carry them with you nor have a package reference to them.

joperezr:
@dcostea I took the liberty to push directly to your PR too in order to address the comments by @eerhardt I hope that was ok. After my last change, you can now call dotnet pack to both the MDA.csproj or MDAI.csproj and they will both generate the same package, which layout is as follows:
```
Icon.png
LICENSE
Microsoft.Data.Analysis.nuspec
out.txt
THIRD-PARTY-NOTICES.TXT
[Content_Types].xml
interactive-extensions\dotnet\Microsoft.Data.Analysis.Interactive.dll
lib\netstandard2.0\Microsoft.Data.Analysis.dll
lib\netstandard2.0\Microsoft.Data.Analysis.xml
```

src/Microsoft.Data.Analysis.Interactive/Directory.Build.targets
```diff
@@ -0,0 +1,14 @@
+<Project>
+  <Import Project="./../../Directory.Build.targets" />
+
+  <!-- Overriding PackDependsOn so instead we just call Pack on the MDA project. -->
+  <PropertyGroup>
+    <PackDependsOn>PackMDAProject</PackDependsOn>
```
eerhardt, line 6: Does this conflict with having `<IsPackable>false</IsPackable>` in the .csproj?

I don't think we need this. `dotnet pack` on this project is fine to be a no-op.

joperezr: > Does this conflict with having <IsPackable>false</IsPackable> in the .csproj?

Nope, IsPackable only matters when you actually use the Pack targets in the SDK but the fact that we are overriding this here makes it that setting it to false is actually not required any longer.

> I don't think we need this. dotnet pack on this project is fine to be a no-op.

I'm fine with removing this file too, I only added it in case people wanted to be able to call dotnet pack on this project and actually generate the package where this assembly is packed, even if it is not part of this project. I'm fine with either keeping it or removing it in case we don't think a lot of people will use it.

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
dcostea: > You'll want to update to a newer version.

@jonsequitur, @eerhardt could you direct me to a newer version?
The latest nuget version is 1.0.0-beta.20154.7 (see  https://www.nuget.org/packages/Microsoft.DotNet.Interactive)
Thank you.

pgovind:
I saw a demo of this locally as well. Super cool! I'm going to go ahead and merge it. Thanks for all the work @dcostea @eerhardt @jonsequitur and @joperezr

dcostea:
Thank you, everyone, @eerhardt @pgovind @jonsequitur @joperezr @colombod for your support. I hope this is my first contribution from a long series.

Issue #2926, opened 2020-05-30
Data.Analysis : float/double values don't convert properly with . decimals in csv file


chriss2401:
If I have a csv dataframe that looks like this:

Date;Value
12-04-1989;20,7
03-11-1990;22,1

Then my two float values will properly be loaded when I call `DataFrame.LoadCsv` (with values of `20.7` and `22.1`

But if my separator is the default one (',') and my floats are with a dot instead, they will get ignored and I will get 207 and 221 as values.

The issue comes from here : 
https://github.com/dotnet/corefxlab/blob/master/src/Microsoft.Data.Analysis/DataFrame.cs#L488

Since if you write `object value = Convert.ChangeType("20.7", typeof(double));` you will get 207 as a result.

chriss2401:
Apparently you need to add `CultureInfo.InvariantCulture` to the function if your double/float has dots instead of commas:

https://stackoverflow.com/questions/7725020/how-to-use-convert-changetype-when-conversiontype-is-decimal-and-input-is-40

So a potential fix could be to check whether the type of the column is a double/float and call `ChangeType` in the right way depending on whether the value(s) is described with dots or commas. I can open a branch for this if the maintainers agree with this approach.

eerhardt:
> So a potential fix could be to check whether the type of the column is a double/float and call ChangeType in the right way depending on whether the value(s) is described with dots or commas. I can open a branch for this if the maintainers agree with this approach.

We may also think about accepting a `CultureInfo` in the public API. That way the caller can specify what culture to use to parse numbers. The caller is going to know best whether they want `CurrentCulture` or `InvariantCulture`.

@chriss2401 - if you have a fix and test cases, feel free to send a PR and @pgovind and I can take a look.

Note also #2927 is re-doing how LoadCsv is implemented. So we may want to consider that as well.

chriss2401:
@eerhardt Good idea. I can implement this in a couple of days unless you guys decide to add it to #2927 , which in that case I can just close the issue.

pgovind:
I think taking a `CultureInfo` in the public API is the right approach. @chriss2401 : If you have a fix and unit tests, I don't mind accepting it into either `DataFrame.LoadCsv` or #2927 (feel free to push a commit to that PR). If you make a new PR, I'll port your fix to #2927 when I go through there next :)

chriss2401:
[...]

Issue #2902, opened 2020-04-09
Improve error experience when LoadCsv throws


eerhardt:
If you have .csv file that has some error in the data, for example:

```
Date | Daily minimum temperatures
1/1/1981 | 20.7
1/2/1981 | 17.9
1/3/1981 | 18.8
1/4/1981 | 14.6
1/5/1981 | 15.8
...
7/20/1982 | ?0.2
7/21/1982 | ?0.8
```

Here you get an exception:

```
System.FormatException: Input string was not in a correct format.
   at System.Number.ThrowOverflowOrFormatException(ParsingStatus status, TypeCode type)
   at System.String.System.IConvertible.ToSingle(IFormatProvider provider)
   at System.Convert.ChangeType(Object value, Type conversionType, IFormatProvider provider)
   at Microsoft.Data.Analysis.DataFrame.Append(IEnumerable`1 row, Boolean inPlace)
   at Microsoft.Data.Analysis.DataFrame.LoadCsv(Stream csvStream, Char separator, Boolean header, String[] columnNames, Type[] dataTypes, Int64 numberOfRowsToRead, Int32 guessRows, Boolean addIndexColumn, Encoding encoding)
```

Which isn't super helpful to tell you where or what was causing the problem. We should log a more helpful error in this situation so users know what is wrong.

We should also consider having a mode where things like this turn into `N/A` values, like `Single.Nan`.

chriss2401:
I've wrapped a simple try catch around the function where it "blows up" :
https://github.com/chriss2401/corefxlab/commit/e1f74ec49ffaf3803052b28c4ff20c5f507ad63a

Regarding the NaN comment, would it make sense that it by default converts it to `Single.NaN` inside the catch area, doesn't throw an exception but informs the user that this value was converted to NaN ?

chriss2401:
@eerhardt @pgovind could this fix potentially also go into #2927 ?

pgovind:
Yup, @chriss2401, I think it makes sense for this fix to go into #2927. At some point, I expect to remove the `LoadCsv` method from `DataFrame` and move it to `DataFrameIO`. 

[...]

PR #2938, opened 2020-06-21
Robust dataframe io

chriss2401:
Hi @eerhardt and @pgovind ,

The following pull request adds the following features to `Microsoft.Data.Analysis` :

* Separate M.D.A and M.D.A.I.O projects (pulled from `pgovind:DataframeIO` )
* Removed M.D.A.I.O related duplicate code from M.D.A and moved needed code to IO project (i.e. main high level `LoadCsv` function).
*  Added a `CultureInfo` object as a input parameter in the public API with unit tests - related to #2926 
* Added better exception handling/message printing when type convertions fail with unit tests - related to #2902 
    * If the type that fails is single/double, a corresponding `NaN` is assigned.

Please review at your own convenience. 

Christos.


dnfadmin:
[CLA assistant check](https://cla.dotnetfoundation.org/dotnet/corefxlab?pullRequest=2938) <br/>Thank you for your submission, we really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla.dotnetfoundation.org/dotnet/corefxlab?pullRequest=2938) before we can accept your contribution.<br/><br/>:x: chriss2401 [sign now](https://cla.dotnetfoundation.org/dotnet/corefxlab?pullRequest=2938)<br/><sub>You have signed the CLA already but the status is still pending? Let us [recheck](https://cla.dotnetfoundation.org/check/dotnet/corefxlab?pullRequest=2938) it.</sub>

src/Microsoft.Data.Analysis.IO/DataFrameIO.cs
```diff
@@ -183,29 +89,125 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
-        /// <returns><see cref="DataFrame"/></returns>
-        public static DataFrame LoadCsv(Stream csvStream,
+        /// <param name="cultureInfo">culture info for processing floats/doubles</param>
+        /// <returns>DataFrame</returns>
+        public static DataFrame LoadCsv(string filename,
+                                char separator = ',', bool header = true,
+                                string[] columnNames = null, Type[] dataTypes = null,
+                                int numRows = -1, int guessRows = 10,
+                                bool addIndexColumn = false, Encoding encoding = null,
+                                CultureInfo cultureInfo = null)
```
chriss2401, line 183: Related to #2926 , here is the `CultureInfo` exposed to the public API.

src/Microsoft.Data.Analysis.IO/DataFrameIO.cs
```diff
@@ -183,29 +89,125 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
                                 long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false,
-                                Encoding encoding = null)
+                                Encoding encoding = null, CultureInfo cultureInfo = null)
         {
-            if (!csvStream.CanSeek)
+            // if we have a comma separator and the cultureInfo has not been specified by the user, 
+            // we set it to Invariant Culture (since logically floats/doubles wouldn't be represented with commas in this case)
+            if(separator.Equals(',') && cultureInfo is null)
```
chriss2401, line 202: In the case where the `CultureInfo` object is not null and the file has a comma based separator, we change the CultureInfo to Invariant in order to handle Double/Single values.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -485,7 +487,30 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
                     if (value != null)
                     {
-                        value = Convert.ChangeType(value, column.DataType);
+                        try
+                        {
+                            value = Convert.ChangeType(value, column.DataType, cultureInfo);
+                        }
+                        catch(FormatException)
+                        {
+                            Console.Write($"Value \"{value}\" cannot be converted to type {column.DataType} (Column name: {column.Name}). ");
```
chriss2401, line 496: Related to #2902 , give a message to the user what value can't be converted to which type. If Double/Single, assign a `NaN` value instead. Otherwise throw a `FormatException` as previously ( this also doesn't break unit test `TestAppendRow` which checks for `FormatException` ).

tests/Microsoft.Data.Analysis.IO.Tests/DataFrame.IOTests.CultureInfo.cs
```diff
@@ -0,0 +1,61 @@
...
+using System;
+using System.Collections.Generic;
+using System.Globalization;
+using System.IO;
+using System.Text;
+using Xunit;
+
+namespace Microsoft.Data.Analysis.IO.Tests
+{
+    public partial class DataFrameIOTests
```
chriss2401, line 14: Added two simple unit tests in relation to #2902

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -14,6 +14,94 @@ namespace Microsoft.Data.Analysis.Tests
 {
     public partial class DataFrameTests
     {
+        internal static void VerifyColumnTypes(DataFrame df, bool testArrowStringColumn = false)
```
chriss2401, line 17: This is duplicated both in M.D.A and M.D.A.IO, since both unit test projects use this function. I found duplicating the function was the easiest work around for now.

src/Microsoft.Data.Analysis.IO/DataFrameIO.cs
pgovind: So, for now I think this is ok since it's being more restrictive. I'm wondering if we can lose the `separator.Equals(',')` part of the if condition safely, but we can do that later too.

src/Microsoft.Data.Analysis.IO/Microsoft.Data.Analysis.IO.csproj
```diff
@@ -0,0 +1,37 @@
+﻿<Project Sdk="Microsoft.NET.Sdk">
+
+  <PropertyGroup>
+    <TargetFrameworks>netcoreapp3.0;net461</TargetFrameworks>
```
pgovind, line 4: Should be change net461 to net472? @eerhardt

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -485,7 +487,30 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
+
+                            if (column.DataType == typeof(double))
+                            {
+                                Console.WriteLine("Converting to Double.NaN instead.");
+                                value = Double.NaN;
```
pgovind, line 501: I like this change, but do you think you could separate it out into it's own PR? Reason is that I don't think we've tested any of our existing APIs with `NaN` values. So I'd like to get coverage by modifying `DataFrameTests.cs:MakeDataFrameWithNumericColumns` generate a row with `Double.NaN` and `Single.NaN`. That should automatically give us coverage over all the APIs

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -230,4 +230,5 @@
       <CustomToolNamespace>Microsoft.Data</CustomToolNamespace>
     </EmbeddedResource>
   </ItemGroup>
+
```
pgovind, line 233: Maybe revert this change as your modifying the PR?

eerhardt:
I have a couple concerns with the proposal here.

1. Loading and writing a .csv file to/from DataFrame seems like a pretty intrinsic operation. I don't think we want a separate assembly/NuGet package for this. Customers should be able to just get the `Microsoft.Data.Analysis` package and load a CSV.
2. I'm a bit concerned about taking a dependency on `Microsoft.VisualBasic.FileIO.TextFieldParser`.
    1. It isn't available in `netstandard2.0`.
    2. There are a few public blogs saying it isn't very performant, and it looks like it hasn't been updated or enhanced in a few years.

I think we should fold this functionality back M.D.A and investigate what we can do about removing the dependency on `TextFieldParser`. (One option could be to port the code to C# and modify it as appropriate.)

chriss2401:
Hi @pgovind and @eerhardt,

If there is consensus on keeping only one project (M.D.A), then I can roll back the changes and open a PR that only closes #2926, and then another one for #2902.

Another option could be that there is a high level public function in M.D.A which loads a csv and then all the helper functions are in the I.O. related one. That way there are two separations and the user can just call M.D.A.

pgovind:
@chriss2401 I think rolling back the changes and fixing only #2926 and #2902 is a good idea! I really like the fixes for those 2 issues. Once my .NET 5 work is done, I'll work on a full fledged `LoadCsv` method that should fix all our parsing concerns with `LoadCsv`

chriss2401:
> 
> 
> @chriss2401 I think rolling back the changes and fixing only #2926 and #2902 is a good idea! I really like the fixes for those 2 issues. Once my .NET 5 work is done, I'll work on a full fledged `LoadCsv` method that should fix all our parsing concerns with `LoadCsv`

Sounds good! I opened a PR for #2902 , will afterwards open one for #2926. Closing this PR :)

PR #2939, opened 2020-06-28
Prevent DataFrame.Sample() method from returning duplicated rows

### LINKED ISSUE #2806: DataFrame.Sample doesn't check for already generated indices
Hello,

The `Sample` method in `DataFrame` (code [here](https://github.com/dotnet/corefxlab/blob/master/src/Microsoft.Data.Analysis/DataFrame.cs#L325)) does not check if an index was already generated by `rand`. Most of the time I get duplicate rows because of it.

RamonWill:
**Issue**

> The Sample method in DataFrame (code here) does not check if an index was already generated by rand. Most of the time I get duplicate rows because of it.

**Solution**
I have amended the Sample method to implement the Fisher-Yates shuffle so that the sample returned is unique and still random. Additions also include a new string resource so that an exception is throw if the sample size requested is greater than the number of rows. Tests for row uniqueness and the exception being thrown have been included also.

Kind Regards,
Ramon
Fixes: #2806

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -316,14 +316,29 @@ namespace Microsoft.Data.Analysis
         /// <param name="numberOfRows">Number of rows in the returned DataFrame</param>
         public DataFrame Sample(int numberOfRows)
         {
-            Random rand = new Random();
-            PrimitiveDataFrameColumn<long> indices = new PrimitiveDataFrameColumn<long>("Indices", numberOfRows);
-            int randMaxValue = (int)Math.Min(Int32.MaxValue, Rows.Count);
-            for (long i = 0; i < numberOfRows; i++)
+            if (numberOfRows > Rows.Count)
             {
-                indices[i] = rand.Next(randMaxValue);
+                throw new ArgumentException(string.Format(Strings.ExceedsNumberOfRows, Rows.Count), nameof(numberOfRows));
+            }
+            
+            int shuffleSize = (int)Math.Min(Int32.MaxValue, Rows.Count);
+            int[] shuffleArray = Enumerable.Range(0, shuffleSize).ToArray();
+            Random rand = new Random();
+            while (shuffleSize > 1)
+            {
+                int randomIndex = rand.Next(shuffleSize--);
+                int temp = shuffleArray[shuffleSize];
+                shuffleArray[shuffleSize] = shuffleArray[randomIndex];
+                shuffleArray[randomIndex] = temp;
             }
 
+            PrimitiveDataFrameColumn<long> indices = new PrimitiveDataFrameColumn<long>("Indices", numberOfRows);
+            
+            for(long i = 0; i < numberOfRows; i++)
+            {
+                indices[i] = shuffleArray[i];
+            }
+            
             return Clone(indices);
         }
 
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -320,18 +320,21 @@ namespace Microsoft.Data.Analysis
             {
                 throw new ArgumentException(string.Format(Strings.ExceedsNumberOfRows, Rows.Count), nameof(numberOfRows));
             }
+
+            int shuffleLowerLimit = 0;
+            int shuffleUpperLimit = (int)Math.Min(Int32.MaxValue, Rows.Count);
             
-            int shuffleSize = (int)Math.Min(Int32.MaxValue, Rows.Count);
-            int[] shuffleArray = Enumerable.Range(0, shuffleSize).ToArray();
+            int[] shuffleArray = Enumerable.Range(0, shuffleUpperLimit).ToArray();
             Random rand = new Random();
-            while (shuffleSize > 1)
+            while (shuffleLowerLimit < numberOfRows)
             {
-                int randomIndex = rand.Next(shuffleSize--);
-                int temp = shuffleArray[shuffleSize];
-                shuffleArray[shuffleSize] = shuffleArray[randomIndex];
+                int randomIndex = rand.Next(shuffleLowerLimit, shuffleUpperLimit);
+                int temp = shuffleArray[shuffleLowerLimit];
+                shuffleArray[shuffleLowerLimit] = shuffleArray[randomIndex];
                 shuffleArray[randomIndex] = temp;
+                shuffleLowerLimit++;
             }
-            ArraySegment<int> segment = new ArraySegment<int>(shuffleArray, 0, numberOfRows);
+            ArraySegment<int> segment = new ArraySegment<int>(shuffleArray, 0, shuffleLowerLimit);
 
             PrimitiveDataFrameColumn<int> indices = new PrimitiveDataFrameColumn<int>("indices", segment);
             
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -316,14 +316,29 @@ public DataFrame AddSuffix(string suffix, bool inPlace = false)
         /// <param name="numberOfRows">Number of rows in the returned DataFrame</param>
         public DataFrame Sample(int numberOfRows)
         {
+            if (numberOfRows > Rows.Count)
+            {
+                throw new ArgumentException(string.Format(Strings.ExceedsNumberOfRows, Rows.Count), nameof(numberOfRows));
+            }
+            
+            int shuffleSize = (int)Math.Min(Int32.MaxValue, Rows.Count);
+            int[] shuffleArray = Enumerable.Range(0, shuffleSize).ToArray();
```
pgovind, line 325: So here we're allocating 2 times: once for `shuffleArray` and once for `indices`. I think a better way would be to get rid of `indices` and change `shuffleArray` to a `PrimitiveDataFrameColumn<int>`. Then after the shuffle, you should just be able to call `Clone(shuffleArray)`

RamonWill: Hi @pgovind ,

I believe with that approach I will end up cloning the whole column dataframe with the rows in a different order since to my knowledge i can't slice a column. 

Is there currently a way to slice a `PrimitiveDataframeColumn` without it being a `IReadOnlyList`? i think it would make this method more efficient. With the way this PR is currently set up I still need to use indices in order to return the `numberOfRows` that the user specifies.

I have also realised that I can use an `ArraySegment<T>` instead of the final for loop so i will update this PR again shortly.

pgovind: Hmm, I don't quite follow. I think the following will succeed right?

```
            int shuffleSize = (int)Math.Min(Int32.MaxValue, Rows.Count);
            PrimitiveDataFrameColumn<long> shuffleArray = new PrimitiveDataFrameColumn<long>("indices", Enumerable.Range(0, shuffleSize).Select(x => (long)x));
            Random rand = new Random();
            while (shuffleSize > 1)
            {
                int randomIndex = rand.Next(shuffleSize--);
                long? temp = shuffleArray[shuffleSize];
                shuffleArray[shuffleSize] = shuffleArray[randomIndex];
                shuffleArray[randomIndex] = temp;
            }
            //ArraySegment<int> segment = new ArraySegment<int>(shuffleArray, 0, numberOfRows);

            //PrimitiveDataFrameColumn<int> indices = new PrimitiveDataFrameColumn<int>("indices", segment);
            
            return Clone(shuffleArray);
        }
```

RamonWill: Hi @pgovind ,

That wouldn't work unfortunately because even though the indices are shuffled we would be returning a shuffleArray of size Rows.Count, when we should be returning an shuffleArray of size numberOfRows. 

So if the user has a dataframe of 10 rows and wanted a sample of 7 (where 7=numberOfRows)  your example would return all 10 rows (Rows.Count) but in a different order when we just need to return 7. However, if there was a way to slice the shuffleArray in your example to a size numberOfRows we could use your method which does look cleaner. 

The thing is though when i attempt to slice the column it returns a IReadOnlyList which i can't use in the clone method:

```diff
            int shuffleSize = (int)Math.Min(Int32.MaxValue, Rows.Count);
            PrimitiveDataFrameColumn<long> shuffleArray = new PrimitiveDataFrameColumn<long>("indices", Enumerable.Range(0, shuffleSize).Select(x => (long)x));
            Random rand = new Random();
            while (shuffleSize > 1)
            {
                int randomIndex = rand.Next(shuffleSize--);
                long? temp = shuffleArray[shuffleSize];
                shuffleArray[shuffleSize] = shuffleArray[randomIndex];
                shuffleArray[randomIndex] = temp;
            }
            //ArraySegment<int> segment = new ArraySegment<int>(shuffleArray, 0, numberOfRows);

            //PrimitiveDataFrameColumn<int> indices = new PrimitiveDataFrameColumn<int>("indices", segment);
            
+            return Clone(shuffleArray[0, numberOfRows]); // I was hoping something like this would work
        }
```
 
Kind Regards,

Ramon

pgovind: Ah right, my bad in the example I provided. Can you replace the `int shuffleSize` line with this instead?:
```
            int shuffleSize = (int)Math.Min(Int32.MaxValue, numberOfRows); // numberOfRows instead of Rows.Count
            // The rest of the method remains the same
```

RamonWill: Hi @pgovind, 

So the issue with making that change is that it will set the maxValue parameter for rand.Next to numberOfRows which means that if the user wanted a sample of 7 rows out of 10, then indexes 7, 8 and 9 will be omitted from the shuffle and the Clone() that we return. As a result we would end up shuffling just the indexes between 0 to numberOfRows exclusive.

 However, i did manage to find an optimisation taking this into account. Rather than `while (shuffleSize > 1)` which starts at the end of the array and shuffles backwards. I could do something like the below:

```diff
            int shuffleLowerLimit = 0;
            int shuffleUpperLimit = (int)Math.Min(Int32.MaxValue, Rows.Count); // this wil prevent us from ignoring any indexes
            
            int[] shuffleArray = Enumerable.Range(0, shuffleUpperLimit).ToArray();
            Random rand = new Random();
            
            while (shuffleLowerLimit < numberOfRows) // this will shuffle from the start up to numberOfRows
            {
                int randomIndex = rand.Next(shuffleLowerLimit, shuffleUpperLimit);
                int temp = shuffleArray[shuffleLowerLimit];
                shuffleArray[shuffleLowerLimit] = shuffleArray[randomIndex];
                shuffleArray[randomIndex] = temp;
                shuffleLowerLimit++;
            }
            ArraySegment<int> segment = new ArraySegment<int>(shuffleArray, 0, shuffleLowerLimit);
```

This will only shuffle the numbersOfRows that the user requested. So if a user has 1000 rows and wanted a sample of 7, My previous commit would shuffle all 1000 rows, but this commit will just shuffle 7 rows and take all indexes from 0-999 into consideration.

Apologies for the constant changes but this review and your time is seriously much appreciated, I would not have made these considerations myself. I will make another commit shortly with these changes.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -316,14 +316,28 @@ public DataFrame AddSuffix(string suffix, bool inPlace = false)
...
-                indices[i] = rand.Next(randMaxValue);
+                int randomIndex = rand.Next(shuffleLowerLimit, shuffleUpperLimit);
+                int temp = shuffleArray[shuffleLowerLimit];
+                shuffleArray[shuffleLowerLimit] = shuffleArray[randomIndex];
+                shuffleArray[randomIndex] = temp;
+                shuffleLowerLimit++;
             }
+            ArraySegment<int> segment = new ArraySegment<int>(shuffleArray, 0, shuffleLowerLimit);
 
+            PrimitiveDataFrameColumn<int> indices = new PrimitiveDataFrameColumn<int>("indices", segment);
```
pgovind, line 339: Ah, ideally, we'd have had a `DataFrame.Clone` overload that takes an `IEnumerable<int>` already so we didn't need to allocate another indices column here. If you're interested, we can create another issue for it?

pgovind (approved):
Just the 1 comment for future work if you're interested :) Otherwise this is good to go in. Nice work!

src/Microsoft.Data.Analysis/DataFrame.cs
RamonWill: Will raise the issue now. Thanks for the help and review!

PR #2940, opened 2020-07-01
Add System.Web compatibility assembly

ericstj:
This assembly implements most the surface area of System.Web by throwing
PlatformNotSupportedException.

Some API which can be thought of as "capability" checks return default
values to indicate the API is not supported.

Right now the process to define API is to add the DocID to NotSupported.Exclude.txt 
and add the definition to the manual file.  If we wanted to make this a single step
we could add a task that examined source files and emitted the DocIDs to exclude, 
but given the small number and compiler errors would force correctness I haven't
bothered.  However it might be an interesting improvement to support 
"Partial PNSE assemblies".  /cc @Anipik 

This enables libraries built against desktop using these parts of
System.Web to JIT successfully and execute as if they were running
outside the context of a web project.

This only exposes the System.Web API for runtime, it provides no compile 
time assembly. 

Fixes https://github.com/dotnet/runtime/issues/698

This PR also demonstrates how someone can build their own System.Web and reference it from their application.  See the tests for the workarounds necessary to get this binary to win over the shared framework version. The only change necessary to pull this out of CoreFxLab to any customer repo (without arcade) is to add [MSFT.snk](https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.Arcade.Sdk/tools/snk/MSFT.snk) to the project dir and replace `StrongNameKeyId` with the following:
```xml
    <SignAssembly>true</SignAssembly>
    <PublicSign>true</PublicSign>
    <AssemblyOriginatorKeyFile>MSFT.snk</AssemblyOriginatorKeyFile>
```
cc @ryandle


ericstj:
Private built package is here: https://www.myget.org/feed/ericstj-staging/package/nuget/System.Web.Compatibility

tests/System.Web.Compatibility.Tests/System.Web.Compatibility.Tests.cs
```diff
@@ -0,0 +1,43 @@
...
+        [Fact]
+        public void HostingEnvironmentIsHostedDoesNotThrow()
+        {
+            Assert.False(HostingEnvironment.IsHosted);
+        }
+
+        [Fact]
+        public void HostingEnvironmentSiteNameDoesNotThrow()
+        {
+            Assert.Null(HttpContext.Current);
```
nil4, line 28:
```suggestion
            Assert.Null(HostingEnvironment.SiteName);
```

ericstj:
FYI: this assembly is produced out of corexlab as it is just meant to help unblock folks building legacy codebases.  This is a 1MB assembly that only throws PNSE.  It's not something we'd want to commit to the shared framework or even the windows-compat pack.  If folks find this useful and would like us to support it more we could consider bringing it to dotnet/runtime.

PR #2945, opened 2020-07-28
Add WriteCsv method (plus unit tests). This solves #2917.

dcostea:
In my work with DataFrame I need to save the new data to csv file, so I think we can have WriteCsv. This pairs nicely with LoadCsv (I tried to stick to conventions used there) and it solves #2917, as well.

I have added a few unit tests.

Still, this is an idea and I'm open to suggestions.

@pgovind @eerhardt, what do you think?

Issue #2917, opened 2020-05-19
Writing dataframe to txt doesn't work properly


liugaocn:
I don't know if I am using the write method to export dataframe to txt. Anyways, I just use StreamWriter, but the question is that when My dataframe is really large, it won't write to txt properly. My code is: 
```c#
using(StreamWriter sw = new StreamWriter(SpecialDirectories.Desktop + "\\test.txt", false, encoding: System.Text.Encoding.UTF8))
{
    sw.Write(df);
    sw.Flush();
    sw.Write("End!");
}
```
My table has more then 30 thousand rows, but i only can get about 30 rows from my txt. However, it can write "End!" to the end of the txt properly, by the way.
When I use Console.WriteLine(df), it works properly, though. Then I wonder if it is a bug or just I am not using it right.

pgovind:
Hmmm, how did you create your DataFrame? If you read in data from a csv file, can you share it here and I can repro and debug locally?

liugaocn:
@pgovind 
[This is the link to the file. ](https://www.jianguoyun.com/p/Dc0ocqEQpIb6BhiRrKED )

liugaocn:
@pgovind Is there any solutions to this problem? 
Maybe the dataframe is garbage collected. I guess.

pgovind:
I haven't had a chance to look at it. I'm busy helping out with a .net preview release. I can get back to this in a couple of weeks though and will try to repro/diagnose this then.

dcostea:
@pgovind in my work with dataframe I need to write the changes to an external CSV file.
I think a method like ```public static void WriteCsv(DataFrame dataFrame, string filename,
                                char separator = ',', bool header = true,
                                bool addIndexColumn = false, Encoding encoding = null)``` would be helpful.
I have already the method body using StreamWriter.

May I open a PR?

pgovind:
@dcostea : Yes!! We'd love to see a PR. Tag me or @eerhardt  and we'll be happy to help review it! 

Note: I think this should eventually end up in the new MDAIO project (https://github.com/dotnet/corefxlab/pull/2938). But if you add it into MDA itself for now, that's ok too. I'll move it the next time I go through there :)

liugaocn:
> @pgovind in my work with dataframe I need to write the changes to an external CSV file.
[...]

PR #2947, opened 2020-07-29
Add WriteCsv plus unit tests.

dcostea:
I added WriteCsv method (plus some unit tests) to write a DataFrame to CSV file. This solves #2917.

@eerhardt @pgovind I have recreated the PR following Eric's recommendations from #2945. Please review.

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -397,7 +397,6 @@ namespace Microsoft.Data.Analysis
 
                             if (t == typeof(float))
                             {
-                                //record.AppendFormat(cultureInfo, "{0:G9}", Convert.ToSingle(cell));
                                 record.AppendFormat(cultureInfo, "{0:G9}", cell);
                                 continue;
                             }
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -5,6 +5,7 @@
 using System;
 using System.Collections.Generic;
 using System.IO;
+using System.Linq;
 using System.Text;
 
 namespace Microsoft.Data.Analysis
@@ -304,6 +305,76 @@ namespace Microsoft.Data.Analysis
                     columns.Insert(0, indexColumn);
                 }
                 return ret;
+            }
+        }
+
+        /// <summary>
+        /// Writes a DataFrame into a CSV.
+        /// Follows pandas API.
+        /// </summary>
+        /// <param name="dataFrame"><see cref="DataFrame"/></param>
+        /// <param name="path">CSV file path</param>
+        /// <param name="separator">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="addIndexColumn">add one column with the row index</param>
+        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        public static void WriteCsv(DataFrame dataFrame, string path,
+                                   char separator = ',', bool header = true,
+                                   bool addIndexColumn = false, Encoding encoding = null)
+        {
+            using (FileStream csvStream = new FileStream(path, FileMode.Create))
+            {
+                WriteCsv(dataFrame: dataFrame, csvStream: csvStream,
+                           separator: separator, header: header,
+                           addIndexColumn: addIndexColumn, encoding: encoding);
+            }
+        }
+
+        /// <summary>
+        /// Writes a DataFrame into a CSV.
+        /// Follows pandas API.
+        /// </summary>
+        /// <param name="dataFrame"><see cref="DataFrame"/></param>
+        /// <param name="csvStream">stream of CSV data to be write out</param>
+        /// <param name="separator">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="addIndexColumn">add one column with the row index</param>
+        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        public static void WriteCsv(DataFrame dataFrame, Stream csvStream,
+                           char separator = ',', bool header = true,
+                           bool addIndexColumn = false, Encoding encoding = null)
+        {
+            using (StreamWriter csvFile = new StreamWriter(csvStream, encoding ?? Encoding.UTF8))
+            {
+                if (dataFrame != null)
+                {
+                    List<string> selectedColumnNames = dataFrame.Columns.GetColumnNames().ToList();
+
+                    if (addIndexColumn)
+                    {
+                        selectedColumnNames.Insert(0, "index");
+                    }
+
+                    if (header)
+                    {
+                        var headerColumns = string.Join(separator.ToString(), selectedColumnNames);
+                        csvFile.WriteLine(headerColumns);
+                    }
+
+                    var index = 0;
+                    foreach (var row in dataFrame.Rows)
+                    {
+                        var rowCells = row.ToList();
+
+                        if (addIndexColumn)
+                        {
+                            rowCells.Insert(0, index++);
+                        }
+                        var rowValues = string.Join(separator.ToString(), rowCells);
+
+                        csvFile.WriteLine(rowValues);
+                    }
+                }
             }
         }
     }
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -409,7 +409,6 @@ namespace Microsoft.Data.Analysis
 
                             if (t == typeof(decimal))
                             {
-                                //record.AppendFormat(cultureInfo, "{0:G31}", Convert.ToDecimal(cell));
                                 record.AppendFormat(cultureInfo, "{0:G31}", cell);
                                 continue;
                             }
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -318,7 +318,7 @@ namespace Microsoft.Data.Analysis
         /// <param name="separator">column separator</param>
         /// <param name="header">has a header or not</param>
         /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
-        /// <param name="cultureInfo">culture info for handeling decimal values</param>
+        /// <param name="cultureInfo">culture info for formatting values</param>
         public static void WriteCsv(DataFrame dataFrame, string path,
                                    char separator = ',', bool header = true,
                                    Encoding encoding = null, CultureInfo cultureInfo = null)
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -4,9 +4,11 @@
 
 using System;
 using System.Collections.Generic;
+using System.Globalization;
 using System.IO;
 using System.Linq;
 using System.Text;
+using Microsoft.ML;
 
 namespace Microsoft.Data.Analysis
 {
@@ -316,17 +318,17 @@ namespace Microsoft.Data.Analysis
         /// <param name="path">CSV file path</param>
         /// <param name="separator">column separator</param>
         /// <param name="header">has a header or not</param>
-        /// <param name="addIndexColumn">add one column with the row index</param>
         /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        /// <param name="cultureInfo">culture info for handeling decimal values</param>
         public static void WriteCsv(DataFrame dataFrame, string path,
                                    char separator = ',', bool header = true,
-                                   bool addIndexColumn = false, Encoding encoding = null)
+                                   Encoding encoding = null, CultureInfo cultureInfo = null)
         {
             using (FileStream csvStream = new FileStream(path, FileMode.Create))
             {
                 WriteCsv(dataFrame: dataFrame, csvStream: csvStream,
                            separator: separator, header: header,
-                           addIndexColumn: addIndexColumn, encoding: encoding);
+                           encoding: encoding, cultureInfo: cultureInfo);
             }
         }
 
@@ -338,41 +340,78 @@ namespace Microsoft.Data.Analysis
         /// <param name="csvStream">stream of CSV data to be write out</param>
         /// <param name="separator">column separator</param>
         /// <param name="header">has a header or not</param>
-        /// <param name="addIndexColumn">add one column with the row index</param>
-        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        /// <param name="encoding">the character encoding. Defaults to UTF8 if not specified</param>
+        /// <param name="cultureInfo">culture info for handeling decimal values</param>
         public static void WriteCsv(DataFrame dataFrame, Stream csvStream,
                            char separator = ',', bool header = true,
-                           bool addIndexColumn = false, Encoding encoding = null)
+                           Encoding encoding = null, CultureInfo cultureInfo = null)
         {
+            if (cultureInfo is null)
+            {
+                cultureInfo = CultureInfo.InvariantCulture;
+            }
+            else
+            {
+                if (cultureInfo.NumberFormat.NumberDecimalSeparator.Equals(",") && separator.Equals(','))
+                {
+                    throw new ArgumentException("Decimal separator cannot match the column separator");
+                }
+            }
+
             using (StreamWriter csvFile = new StreamWriter(csvStream, encoding ?? Encoding.UTF8))
             {
                 if (dataFrame != null)
                 {
-                    List<string> selectedColumnNames = dataFrame.Columns.GetColumnNames().ToList();
-
-                    if (addIndexColumn)
-                    {
-                        selectedColumnNames.Insert(0, "index");
-                    }
+                    var columnNames = dataFrame.Columns.GetColumnNames();
 
                     if (header)
                     {
-                        var headerColumns = string.Join(separator.ToString(), selectedColumnNames);
+                        var headerColumns = string.Join(separator.ToString(), columnNames);
                         csvFile.WriteLine(headerColumns);
                     }
 
-                    var index = 0;
+                    var currPosition = csvStream.Position;
+
                     foreach (var row in dataFrame.Rows)
                     {
-                        var rowCells = row.ToList();
+                        var record = new StringBuilder();
 
-                        if (addIndexColumn)
+                        bool firstRow = true;
+                        foreach (var cell in row)
                         {
-                            rowCells.Insert(0, index++);
-                        }
-                        var rowValues = string.Join(separator.ToString(), rowCells);
+                            if (!firstRow)
+                            {
+                                record.Append(separator);
+                            }
+                            else
+                            {
+                                firstRow = false;
+                            }
 
-                        csvFile.WriteLine(rowValues);
+                            Type t = cell?.GetType();
+
+                            if (t == typeof(decimal))
+                            {
+                                record.Append(((decimal)cell).ToString(cultureInfo));
+                                continue;
+                            }
+
+                            if (t == typeof(double))
+                            {
+                                record.Append(((double)cell).ToString(cultureInfo));
+                                continue;
+                            }
+
+                            if (t == typeof(float))
+                            {
+                                record.Append(((float)cell).ToString(cultureInfo));
+                                continue;
+                            }
+
+                            record.Append(cell);
+                        }
+
+                        csvFile.WriteLine(record);
                     }
                 }
             }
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +307,75 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+        /// </summary>
+        /// <param name="dataFrame"><see cref="DataFrame"/></param>
+        /// <param name="csvStream">stream of CSV data to be write out</param>
+        /// <param name="separator">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="addIndexColumn">add one column with the row index</param>
+        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        public static void WriteCsv(DataFrame dataFrame, Stream csvStream,
+                           char separator = ',', bool header = true,
+                           bool addIndexColumn = false, Encoding encoding = null)
```
pgovind, line 345: I think we should take in a `CultureInfo` object in the API too. Similar to https://github.com/dotnet/corefxlab/pull/2938/

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +307,75 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+        {
+            using (StreamWriter csvFile = new StreamWriter(csvStream, encoding ?? Encoding.UTF8))
+            {
+                if (dataFrame != null)
+                {
+                    List<string> selectedColumnNames = dataFrame.Columns.GetColumnNames().ToList();
+
+                    if (addIndexColumn)
+                    {
+                        selectedColumnNames.Insert(0, "index");
```
pgovind, line 355: So normally I'd agree with this name, but if we already had a column called `index` in `dataFrame`, we'll write out the data fine, but if we try to read it back, I think we'll throw an error about duplicated column names. For now, I think we should  remove the `bool addIndexColumn` parameter. We can always call `LoadCsv` with `addIndexColumn = true` to add the `index` column back during read.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +307,75 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                    }
+
```
pgovind, line 357: I think a better approach is to use a `StringBuilder` object here and replace all calls to `string.Join` with `StringBuilder.Append`. If you then also added `CultureInfo`, you'll be able to pass it into the `StringBuilder` object and get the correct formatting for locales.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
```diff
@@ -603,5 +603,75 @@ Stream GetStream(string streamData)
...
+
+        [Fact]
+        public void TestWriteCsvWithHeader()
+        {
+            using MemoryStream csvStream = new MemoryStream();
+            DataFrame dataFrame = MakeDataFrameWithAllColumnTypes(10, true);
+
+            DataFrame.WriteCsv(dataFrame, csvStream);
+
+            var text = Encoding.UTF8.GetString(csvStream.ToArray());
```
pgovind, line 615: Can you try replacing the 4 lines here (and in the other test cases) with `DataFrame readIn = DataFrame.LoadCsv(csvStream)`? That would test that we are able to read in what we wrote out.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
dcostea: I have replaced Join with StringBuilder for rows. (I didn't do it for the header)
I have removed the index column parameter.
I'm not totally convinced this is what you intended.

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
dcostea: @pgovind I'm not sure how I can test WriteCsv if I replace it with LoadCsv. Could you, please, give me more details?

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                    {
+                        var headerColumns = string.Join(separator.ToString(), columnNames);
+                        csvFile.WriteLine(headerColumns);
+                    }
+
+                    var currPosition = csvStream.Position;
+
+                    foreach (var row in dataFrame.Rows)
+                    {
+                        var record = new StringBuilder();
```
pgovind, line 377: I think you can move this declaration outside the for loop so we reduce the number of allocations.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                            if (t == typeof(float))
+                            {
+                                record.Append(((float)cell).ToString(cultureInfo));
+                                continue;
+                            }
+
+                            record.Append(cell);
+                        }
+
+                        csvFile.WriteLine(record);
```
pgovind, line 414: Once `record` is moved outside the loop, we should add a `record.Clear()` here.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                                record.Append(separator);
+                            }
+                            else
+                            {
+                                firstRow = false;
+                            }
+
+                            Type t = cell?.GetType();
+
+                            if (t == typeof(decimal))
```
pgovind, line 393: Can we keep the `float` case alone here and replace the `decimal` and `double` cases with `bool` and `string`? That would mean we can write out numeric, boolean and string values and give writeCsv parity with `LoadCsv`.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                    foreach (var row in dataFrame.Rows)
+                    {
+                        var record = new StringBuilder();
+
+                        bool firstRow = true;
+                        foreach (var cell in row)
+                        {
+                            if (!firstRow)
+                            {
+                                record.Append(separator);
```
pgovind, line 384: Is this is trying to add a separator to the beginning of a row? Do we really need this?

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                            else
+                            {
+                                firstRow = false;
+                            }
+
+                            Type t = cell?.GetType();
+
+                            if (t == typeof(decimal))
+                            {
+                                record.Append(((decimal)cell).ToString(cultureInfo));
```
pgovind, line 395: I'm wondering if this overload is better here than calling `ToString(cultureInfo)`: https://docs.microsoft.com/en-us/dotnet/api/system.text.stringbuilder.appendformat?view=netcore-3.1#System_Text_StringBuilder_AppendFormat_System_IFormatProvider_System_String_System_Object_System_Object_System_Object_

tests/Microsoft.Data.Analysis.Tests/DataFrame.IOTests.cs
pgovind: Sure, what I meant was, instead of doing `csvStream.ToArray()` and reading the stream that way, can you try with `DataFrame.LoadCsv(csvStream)` and then test the resulting DataFrame's values? I see that you have already tried this but commented out that line in other test cases( like `readIn = DataFrame.LoadCsv...`). Any reason why?

pgovind (commented):
Apologies for the late review. I left some comments now :)

src/Microsoft.Data.Analysis/DataFrame.IO.cs
dcostea: It makes sense!

src/Microsoft.Data.Analysis/DataFrame.IO.cs
dcostea: The opposite. It's adding a separator if not the beginning of a row. I use this logic to prefix the subsequent rows with a separator.

dcostea:
> Apologies for the late review. I left some comments now :)

Do not worry. I'm sure you guys are very busy with .NET 5 :)

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
                 return ret;
             }
         }
+
+        /// <summary>
+        /// Writes a DataFrame into a CSV.
+        /// Follows pandas API.
```
eerhardt, line 315: What does `Follows pandas API` mean?

dcostea: I'm not sure, but I found it in the summary of LoadCSV too, so I borrowed it for consistency.

dcostea: My feeling was this is similar to Pandas, but I didn't check myself.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+        /// <param name="header">has a header or not</param>
+        /// <param name="encoding">the character encoding. Defaults to UTF8 if not specified</param>
+        /// <param name="cultureInfo">culture info for handeling decimal values</param>
+        public static void WriteCsv(DataFrame dataFrame, Stream csvStream,
+                           char separator = ',', bool header = true,
+                           Encoding encoding = null, CultureInfo cultureInfo = null)
+        {
+            if (cultureInfo is null)
+            {
+                cultureInfo = CultureInfo.InvariantCulture;
```
eerhardt, line 351: Typically, if someone doesn't specify a culture, .NET APIs use the CurrentCulture.

Examples:

[https://source.dot.net/#System.Private.CoreLib/StringWriter.cs,23](https://source.dot.net/#System.Private.CoreLib/StringWriter.cs,23)
[https://source.dot.net/#System.Private.CoreLib/DateTime.cs,1472](https://source.dot.net/#System.Private.CoreLib/DateTime.cs,1472)
[https://source.dot.net/#System.Drawing.Common/System/Drawing/FontConverter.cs,37](https://source.dot.net/#System.Drawing.Common/System/Drawing/FontConverter.cs,37)

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+            }
+            else
+            {
+                if (cultureInfo.NumberFormat.NumberDecimalSeparator.Equals(",") && separator.Equals(','))
```
eerhardt, line 355: Why are we only checking for `,` here? The exception says the decimal separator cannot match the column separator - so why isn't this line `if (cultureInfo.NumberFormat.NumberDecimalSeparator == separator)`?

Also, should this check happen regardless of whether the caller specified a culture or not?

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                {
+                    var columnNames = dataFrame.Columns.GetColumnNames();
+
+                    if (header)
+                    {
+                        var headerColumns = string.Join(separator.ToString(), columnNames);
+                        csvFile.WriteLine(headerColumns);
+                    }
+
+                    var currPosition = csvStream.Position;
```
eerhardt, line 373: This variable doesn't appear to be used.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,112 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+
+                    foreach (var row in dataFrame.Rows)
+                    {
+                        var record = new StringBuilder();
+
+                        bool firstRow = true;
+                        foreach (var cell in row)
```
eerhardt, line 380: (minor perf question/suggestion): I wonder if we used the `IDataView` API here we could avoid the boxing and unboxing of every primitive value in the DataFrame.

For example, for each column, we can get a `Getter` and use it to get the value generically and write it to the StringBuilder.

See [https://github.com/dotnet/machinelearning/blob/5370692f6d1e6f1f686596d530032b4226dc9e80/src/Microsoft.ML.Data/DataDebuggerPreview.cs#L44-L60](https://github.com/dotnet/machinelearning/blob/5370692f6d1e6f1f686596d530032b4226dc9e80/src/Microsoft.ML.Data/DataDebuggerPreview.cs#L44-L60) for a general structure on how to do this.

I think what we have here is good for a first implementation to get the functionality added. But we may want to benchmark alternatives to see how well this performs.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
eerhardt: I don't think this line has value to callers.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,108 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+
+                            if (t == typeof(bool))
+                            {
+                                record.AppendFormat(cultureInfo, "{0}", Convert.ToBoolean(cell));
+                                continue;
+                            }
+
+                            if (t == typeof(float) || t == typeof(double) || t == typeof(decimal))
+                            {
+                                record.AppendFormat(cultureInfo, "{0}", Convert.ToSingle(cell));
```
eerhardt, line 401: I have a few concerns with this:

1. Converting all values to "single" can change the value. Especially for `decimal` types.
2. Should we be specifying a precision or "round trip" format here? That way the same value will be parsed back.
    - https://docs.microsoft.com/en-us/dotnet/standard/base-types/standard-numeric-format-strings#the-round-trip-r-format-specifier

I'm wondering if we should probably split the 3 types out and write them each separately, as appropriate.

eerhardt: For (2) above, I just spoke with @tannergooding, and he mentioned that with .NET Core 3, you don't need to specifying "round trip" or a precision. Just calling `ToString()` on float, double, and decimal will produce the "shortest round-trippable string" for each type.

So I think we can get rid of the cast to `float` altogether and just call `ToString()` for these 3 types.

dcostea: Agreed.
It seems that "G17" for Double and "G9" for Single is even better. In addition, decimal support G but not R.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,123 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+
+        /// <summary>
+        /// Writes a DataFrame into a CSV.
+        /// </summary>
+        /// <param name="dataFrame"><see cref="DataFrame"/></param>
+        /// <param name="path">CSV file path</param>
+        /// <param name="separator">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        /// <param name="cultureInfo">culture info for handeling decimal values</param>
```
eerhardt, line 321:
```suggestion
        /// <param name="cultureInfo">culture info for formatting values</param>
```

`DateTime` values could be formatted with this culture info as well.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,123 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+
+        /// <summary>
+        /// Writes a DataFrame into a CSV.
+        /// </summary>
+        /// <param name="dataFrame"><see cref="DataFrame"/></param>
+        /// <param name="csvStream">stream of CSV data to be write out</param>
+        /// <param name="separator">column separator</param>
+        /// <param name="header">has a header or not</param>
+        /// <param name="encoding">the character encoding. Defaults to UTF8 if not specified</param>
+        /// <param name="cultureInfo">culture info for handeling decimal values</param>
```
eerhardt, line 342:
```suggestion
        /// <param name="cultureInfo">culture info for formatting values</param>
```

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,123 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                            else
+                            {
+                                firstRow = false;
+                            }
+
+                            Type t = cell?.GetType();
+
+                            if (t == typeof(bool))
+                            {
+                                //record.AppendFormat(cultureInfo, "{0}", Convert.ToBoolean(cell));
```
eerhardt, line 394:
```suggestion
```

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,123 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                                record.AppendFormat(cultureInfo, "{0}", cell);
+                                continue;
+                            }
+
+                            if (t == typeof(float))
+                            {
+                                //record.AppendFormat(cultureInfo, "{0:G9}", Convert.ToSingle(cell));
```
eerhardt, line 401:
```suggestion
```

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,123 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                                record.AppendFormat(cultureInfo, "{0:G9}", cell);
+                                continue;
+                            }
+
+                            if (t == typeof(double))
+                            {
+                                //record.AppendFormat(cultureInfo, "{0:G17}", Convert.ToDouble(cell));
```
eerhardt, line 408:
```suggestion
```

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,123 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                                record.AppendFormat(cultureInfo, "{0:G17}", cell);
+                                continue;
+                            }
+
+                            if (t == typeof(decimal))
+                            {
+                                //record.AppendFormat(cultureInfo, "{0:G31}", Convert.ToDecimal(cell));
```
eerhardt, line 415:
```suggestion
```

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -306,5 +309,123 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                    {
+                        var headerColumns = string.Join(separator.ToString(), columnNames);
+                        csvFile.WriteLine(headerColumns);
+                    }
+
+                    var record = new StringBuilder();
+
+                    foreach (var row in dataFrame.Rows)
+                    {
+                        bool firstRow = true;
```
eerhardt, line 378:
```suggestion
                        bool firstColumn = true;
```

or 

```suggestion
                        bool firstColumnInRow = true;
```

eerhardt (approved):
I think this is pretty close. After a couple minor nits I think we can merge this.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
pgovind: Agree with the rename. This is what was confusing me the last time I reviewed this :).

pgovind (approved):
Just the 1 comment to rename a var. Everything else looks really good! Great work!

dcostea:
> Just the 1 comment to rename a var. Everything else looks really good! Great work!

Thank you everyone for all the help! :)

src/Microsoft.Data.Analysis/DataFrame.IO.cs
dcostea: > Agree with the rename. This is what was confusing me the last time I reviewed this :).

It makes sense!

PR #2948, opened 2020-08-04
Additional System.Web compatibility

IanKemp:
* ~~Not sure how to test instance members of `HttpContext` as the types required by its constructor aren't implemented yet. Should I just go and implement all dependencies as no-ops?~~ Just dummied what I needed anyway.
* GenAPI treats `GeneratePlatformNotSupportedAssemblyMessage` as a literal string not a symbol (https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.GenAPI/build/Microsoft.DotNet.GenAPI.targets#L103), therefore the generated `System.Web.notsupported.cs` does `throw new PlatformNotSupportedException("Strings.PlatformNotSupportedSystemWeb")` instead of the intended `throw new PlatformNotSupportedException(Strings.PlatformNotSupportedSystemWeb)`

cc @ericstj


ericstj:
> GenAPI treats GeneratePlatformNotSupportedAssemblyMessage as a literal string not a symbol (https://github.com/dotnet/arcade/blob/master/src/Microsoft.DotNet.GenAPI/build/Microsoft.DotNet.GenAPI.targets#L103), therefore the generated System.Web.notsupported.cs does throw new PlatformNotSupportedException("Strings.PlatformNotSupportedSystemWeb") instead of the intended throw new PlatformNotSupportedException(Strings.PlatformNotSupportedSystemWeb)

I see, looks like I need to rename the resources type: https://github.com/dotnet/arcade/blob/d6dd7a000fe9908e4caae5eb9aa30ecc9cb917ff/src/Microsoft.Cci.Extensions/Writers/CSharp/CSDeclarationWriter.Methods.cs#L312-L319

src/System.Web.Compatibility/System.Web.manual.cs
```diff
@@ -13,16 +16,43 @@ public partial class HttpContext
...
+            get => null;
+        }
+
+        public IPrincipal User
+        {
+            get => null;
+            set => throw new PlatformNotSupportedException(Strings.PlatformNotSupportedSystemWeb);
+        }
+
+        public HttpContext(HttpRequest request, HttpResponse response) { }
```
ericstj, line 33: If you want to define the constructor you'll need to suppress it in the exclusions file.  The syntax to add for these 3 constructors is as follows:
```
M:System.Web.HttpContext.#ctor(System.Web.HttpRequest,System.Web.HttpResponse)
M:System.Web.HttpRequest.#ctor(System.String,System.String,System.String)
M:System.Web.HttpResponse.#ctor(System.IO.TextWriter)
```

ericstj: I see you had these, but had spaces.  The docID format is rather rigid here.  Let me fix this up for you.

ericstj:
Thanks for the contribution.  I went ahead and merged the change after fixing it up.  Let me know if the new library works for you.

PR #2952, opened 2020-08-11
Fix bug in DataFrame.OrderBy() when columns contain nulls on different rows

### LINKED ISSUE #2903: Sorting DataFrame that has null values in numerical columns
Hello,

I found that calling `DataFrame.OrderBy` on a data frame that has null values in a numeric column throws an ArgumentException with `MismatchedColumnLengths` error. This occurs regardless of the column to sort by (ie. the column to sort by does not have to be the numeric column).

I believe this is because when the DataFrame clones new sorted columns, numeric columns with null values produce copied clones with length shorter than the original column.

Minimal Example to produce this error (see attached `example.csv.txt`) txt file because Github does not support csv files -_-

```
var df = DataFrame.LoadCsv(@"./example.csv.txt");
df.OrderBy("a");
```

[example.csv.txt](https://github.com/dotnet/corefxlab/files/4517470/example.csv.txt)

Microsoft.Data.Analysis v0.3.0

Thanks

RamonWill:
**Issue**
A user found a bug in the OrderBy code but in fact the test case they raised also raises another bug. To summarise: 

1. If the column we want to sort contains nulls then the `.GetAscendingSortIndices()` method will ignore any index that has a null value and the code will break when this mismatched column is added to the final dataframe.
 
2. The code will also break if the column we want to sort doesn't contain nulls but the other columns do contain nulls. This is because the  nullCount parameter  in the following Clone method `DataFrameColumn newColumn = oldColumn.Clone(sortIndices, !isAscending, oldColumn.NullCount);` will append additional nulls to the end of the column which will cause the columns to be mismatched

This passed the previous test cases because the null values are all on the same row for the dataframe used in the test. So any nulls that are removed from point 1 will be offset by the nulls added in point 2( if ALL values in the row are null).

**Solution**
I have made adjustments to the original Sort column and updated/extended the tests to capture the issue raised. I have created a variable sortIndicesNull to capture all the indexes that get omitted by `GetAscendingSortIndices()` i then add these to the end or beginning of a new sorted column and the code continues as it did before.

Kind Regards,
Ramon
Fixes #2903

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -597,15 +597,45 @@ namespace Microsoft.Data.Analysis
         private DataFrame Sort(string columnName, bool isAscending)
         {
             DataFrameColumn column = Columns[columnName];
-            DataFrameColumn sortIndices = column.GetAscendingSortIndices();
+
+            List<long?> sortIndicesNonNull = column.GetAscendingSortIndices().ToList();
+            HashSet<long?> lookUp = new HashSet<long?>(sortIndicesNonNull);
+            List<long?> sortIndicesNull = new List<long?>();
+            
+            for (int i = 0; i < column.Length; i++)
+            {
+                if (!lookUp.Contains(i))
+                {
+                    sortIndicesNull.Add(i);
+                }
+            }
+            DataFrameColumn sortedCol;
+            if (isAscending)
+            {
+                // if the sort is ascending then add the null values to the end of the new column.
+                sortIndicesNonNull.AddRange(sortIndicesNull);
+                sortedCol = new PrimitiveDataFrameColumn<long>("temp", sortIndicesNonNull);
+            }
+            else
+            {
+                /* if the sort is descending then reverse the null values
+                 * and then add the nullvalues to the front the of the new column. 
+                 * This is because the final clone is inverted and will put all null values at the end.*/
+                sortIndicesNull.Reverse();
+                sortIndicesNull.AddRange(sortIndicesNonNull);
+                sortedCol = new PrimitiveDataFrameColumn<long>("temp", sortIndicesNull);
+            }
+
             List<DataFrameColumn> newColumns = new List<DataFrameColumn>(Columns.Count);
+
             for (int i = 0; i < Columns.Count; i++)
             {
                 DataFrameColumn oldColumn = Columns[i];
-                DataFrameColumn newColumn = oldColumn.Clone(sortIndices, !isAscending, oldColumn.NullCount);
+                DataFrameColumn newColumn = oldColumn.Clone(sortedCol, !isAscending);
                 Debug.Assert(newColumn.NullCount == oldColumn.NullCount);
                 newColumns.Add(newColumn);
             }
+
             return new DataFrame(newColumns);
         }
 
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -597,41 +597,14 @@ namespace Microsoft.Data.Analysis
         private DataFrame Sort(string columnName, bool isAscending)
         {
             DataFrameColumn column = Columns[columnName];
-
-            List<long?> sortIndicesNonNull = column.GetAscendingSortIndices().ToList();
-            HashSet<long?> lookUp = new HashSet<long?>(sortIndicesNonNull);
-            List<long?> sortIndicesNull = new List<long?>();
-            
-            for (int i = 0; i < column.Length; i++)
-            {
-                if (!lookUp.Contains(i))
-                {
-                    sortIndicesNull.Add(i);
-                }
-            }
-            DataFrameColumn sortedCol;
-            if (isAscending)
-            {
-                // if the sort is ascending then add the null values to the end of the new column.
-                sortIndicesNonNull.AddRange(sortIndicesNull);
-                sortedCol = new PrimitiveDataFrameColumn<long>("temp", sortIndicesNonNull);
-            }
-            else
-            {
-                /* if the sort is descending then reverse the null values
-                 * and then add the nullvalues to the front the of the new column. 
-                 * This is because the final clone is inverted and will put all null values at the end.*/
-                sortIndicesNull.Reverse();
-                sortIndicesNull.AddRange(sortIndicesNonNull);
-                sortedCol = new PrimitiveDataFrameColumn<long>("temp", sortIndicesNull);
-            }
-
+            DataFrameColumn sortIndices = column.GetAscendingSortIndices();
             List<DataFrameColumn> newColumns = new List<DataFrameColumn>(Columns.Count);
 
             for (int i = 0; i < Columns.Count; i++)
             {
                 DataFrameColumn oldColumn = Columns[i];
-                DataFrameColumn newColumn = oldColumn.Clone(sortedCol, !isAscending);
+                long nullCount = oldColumn.Length - sortIndices.Length;
+                DataFrameColumn newColumn = oldColumn.Clone(sortIndices, !isAscending, nullCount);
                 Debug.Assert(newColumn.NullCount == oldColumn.NullCount);
                 newColumns.Add(newColumn);
             }
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -596,15 +596,22 @@ namespace Microsoft.Data.Analysis
 
         private DataFrame Sort(string columnName, bool isAscending)
         {
+            DataFrameColumn sortIndices;
             DataFrameColumn column = Columns[columnName];
-            DataFrameColumn sortIndices = column.GetAscendingSortIndices();
+            if (isAscending)
+            {
+                sortIndices = column.GetAscendingSortIndices();
+            }
+            else
+            {
+                sortIndices = column.GetDescendingSortIndices();
+            }
             List<DataFrameColumn> newColumns = new List<DataFrameColumn>(Columns.Count);
 
             for (int i = 0; i < Columns.Count; i++)
             {
                 DataFrameColumn oldColumn = Columns[i];
-                long nullCount = oldColumn.Length - sortIndices.Length;
-                DataFrameColumn newColumn = oldColumn.Clone(sortIndices, !isAscending, nullCount);
+                DataFrameColumn newColumn = oldColumn.Clone(sortIndices);
                 Debug.Assert(newColumn.NullCount == oldColumn.NullCount);
                 newColumns.Add(newColumn);
             }
--- a/src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.Sort.cs
+++ b/src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.Sort.cs
@@ -14,20 +14,53 @@ namespace Microsoft.Data.Analysis
     {
         public new PrimitiveDataFrameColumn<T> Sort(bool ascending = true)
         {
-            PrimitiveDataFrameColumn<long> sortIndices = GetAscendingSortIndices();
-            return Clone(sortIndices, !ascending, NullCount);
+            PrimitiveDataFrameColumn<long> sortIndices;
+            if (ascending)
+            {
+                sortIndices = GetAscendingSortIndices();
+            }
+            else
+            {
+                sortIndices = GetDescendingSortIndices();
+            }
+            return Clone(sortIndices);
         }
 
         internal override PrimitiveDataFrameColumn<long> GetAscendingSortIndices()
         {
             // The return sortIndices contains only the non null indices. 
-            GetSortIndices(Comparer<T>.Default, out PrimitiveDataFrameColumn<long> sortIndices);
+            PrimitiveDataFrameColumn<long> sortIndices = GetSortIndices(Comparer<T>.Default, out PrimitiveDataFrameColumn<long> columnNullIndices);
+            if(columnNullIndices.Length > 0)
+            {
+                for(int i = 0; i < columnNullIndices.Length; i++)
+                {
+                    sortIndices.Append(columnNullIndices[i]);
+                }
+            }
             return sortIndices;
         }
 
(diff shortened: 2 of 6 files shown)


pgovind:
I missed this because there weren't any tags on the PR. @pgovind @eerhardt

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -802,32 +802,42 @@ public void TestOrderBy()
             df.Columns["Int"][0] = 100;
             df.Columns["Int"][19] = -1;
             df.Columns["Int"][5] = 2000;
+            df.Columns["Int"][6] = null;
```
eerhardt, line 805: Would it be possible to make a new test that has null values and leave this test untouched?

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -597,15 +597,45 @@ private void OnColumnsChanged()
         private DataFrame Sort(string columnName, bool isAscending)
         {
             DataFrameColumn column = Columns[columnName];
-            DataFrameColumn sortIndices = column.GetAscendingSortIndices();
+
+            List<long?> sortIndicesNonNull = column.GetAscendingSortIndices().ToList();
```
eerhardt, line 601: Why `ToList()` here?

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -597,15 +597,45 @@ private void OnColumnsChanged()
         private DataFrame Sort(string columnName, bool isAscending)
         {
             DataFrameColumn column = Columns[columnName];
-            DataFrameColumn sortIndices = column.GetAscendingSortIndices();
```
eerhardt, line 600: Would it be better if we fixed the `GetAscendingSortIndices()` method itself? I see there is a comment in it saying "The return sortIndices contains only the non-null indices". Maybe we could add an option to that method telling it to return `null` indices or not.

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
RamonWill: Hi eerhardt, sure thing

src/Microsoft.Data.Analysis/DataFrame.cs
RamonWill: The reasoning behind it was so that i could use the AddRange method further down. Which would allow me to append the omitted values (Nulls) to the front or end of the list.

src/Microsoft.Data.Analysis/DataFrame.cs
RamonWill: sure thing @eerhardt , 

I think my initial reason for not changing it was because the .Median method used it and i think i came across another problem (That doesn't spring to mind at the moment). But what i will do is look over this issue again and see improve the commit based on your review.

Thanks,
Ramon

pgovind: Actually, I don't think we need to change `GetAscendingSortIndices`, but rather `Sort` itself. The bug is this call `oldColumn.Clone(sortIndices, !isAscending, oldColumn.NullCount);`. `sortIndices` here is guaranteed to be either less than or equal to `oldColumn.Length` in size. So, if we replace `oldColumn.NullCount` with `oldColumn.Length - sortIndices.Length`, that should fix the bug :)

RamonWill: @pgovind that works! I will update my commit and just add a new test case :).

RamonWill: Hi @pgovind ,
On closer inspection the issue with the solution you mentioned is that Nulls are appended to the end of columns where it isn't needed. 

I created a testcase called TestOrderByWithNulls  in my new commit which fails because the null counts will no longer match between the old and new columns with your proposed solution. (Since the nullCount parameter will append null values inplace of actual values). 

Apologies for the lack of oversight on my side for pushing this new commit in the first place, I really shouldn't have made the assumption there that it worked. 

Do let me know if you have any other solutions? I can look at ways to improve my first commit which worked but needs improvement.

Kind Regards,
Ramon

pgovind: Ah indeed, I didn't consider that case fully in my suggestion. I spent some time looking at this today and this is my suggestion. We need to change the `GetAscendingSortIndices` method like @eerhardt suggests to return the null indices in a column in addition to the sort indices. We also need to create a similar `GetDescendingSortIndices` method. Here is what I would do:
1) Change the current `GetSortIndices` method to look like:
```
private PrimitiveDataFrameColumn<long> GetSortIndices(IComparer<T> comparer, out PrimitiveDataFrameColumn<long> columnNullIndices)
```
If we look at the implementation of `GetSortIndices` in PrimitiveDataFrameColumn.Sort.cs,  after the `IntrospectiveSort` call, we have a for loop to only populate `nonNullSortIndices`. I would write an else condition here and populate the `columnNullIndices` (after `new`-ing it) here like : `columnNullIndices.Append(sortIndices[i] + b * _columnContainer.Buffers[0].Length);`. The `GetSortIndices` method must also now return `columnSortIndices`.
2) Now that `GetSortIndices` is changed, we need to modify its callers, i.e. `GetAscendingSortIndices`. The return value of the `GetSortIndices` call are the `sortIndices` (which still only contain non-null values). The out parameter in `GetSortIndices` contains the `nullIndices`. A simple for loop that appends `nullIndices[i]` to `sortIndices` would suffice. At this point, for an ascending sort call, we have the full `sortIndices` column and don't have to deal with adding `nulls`.

Now we deal with callers of `GetAscendingSortIndices`: 
3) In `DataFrame.Sort` in `DataFrame.cs`, we should simply change the `old.Column.Clone(sortIndices, ...)` to just `oldColumn.Clone(sortIndices)`.
4) Similar change in `DataFrameColumn.Sort` in `DataFrameColumn.cs`.
5) Similar change in `PrimitiveDataFrameColumn.Sort` in `PrimitiveDataFrameColumn.cs`.
6) For `PrimitiveDataFrameColumn.Median`, we probably need to calculate `middle` now as :
```
// The last NullCount indices in sortIndices need to be ignored
            long middle = (sortIndices.Length - NullCount) / 2;
```
At this point, all the asserts associated with `var sortedDf = df.OrderBy("Int");` in the new unit test should pass. 
Then, we need to repeat all the steps above for `StringDataFrameColumn.cs` (except `median` of course) so that we can sort in ascending order via a `StringDataFrameColumn`. This should let all the asserts under `sortedDf = df.OrderBy("String");` in the unit test pass. 

Finally, we can leverage these changes to make a new `GetDescendingSortIndices` API. 
1) Add a new internal `GetDescendingSortIndices` API on `DataFrameColumn` (similar to the existing `GetAscendingSortIndices` virtual method.
2) Override the `GetDescendingSortIndices` method in `PrimitiveDataFrameColumn.Sort.cs`. Something like the following would work:
``` csharp
        internal override PrimitiveDataFrameColumn<long> GetDescendingSortIndices()
        {
            Comparison<T> descendingComparison = new Comparison<T>((T a, T b) =>
            {
                return Comparer<T>.Default.Compare(b, a);
            });
            Comparer<T> descending = Comparer<T>.Create(descendingComparison);
            // The return sortIndices contains only the non null indices. 
            PrimitiveDataFrameColumn<long> sortIndices = GetSortIndices(descending, out PrimitiveDataFrameColumn<long> nullIndices);
            for (long i = 0; i < nullIndices.Length; i++)
            {
                sortIndices.Append(nullIndices[i]);
            }
            return sortIndices;
        }
```
This has similar behavior to ascending sort in that it sorts the valid values in descending order and adds the `null` values to the end of the column. 
Then, we'd need to change
3) Both `PrimitiveDataFrameColumn.Sort` and `StringDataFrameColumn`: A simple `if (ascending)` condition will let us call `GetAscendingSortIndices` or `GetDescendingSortIndices` appropriately. Nothing else needs to be changed here I think.
4) Similar change in `DataFrameColumn.Sort` in `DataFrameColumn.cs`. 
5) Similar change in `DataFrame.Sort` in `DataFrame.cs`. 

This should enable all the descending sort tests in the new unit test pass.

I think I covered all the cases to consider, but let me know if I missed something! I would recommend getting the ascending sort changes working first for `PrimitiveDataFrameColumn` -> Ascending sort changes for `StringDataFrameColumn` -> Descending sort changes in the same order. Let me know if you need help!! I can answer questions or review changes :)

RamonWill: Hi @pgovind ,
Much appreciated for the detailed response/solution. I have implemented your changes and it works fine! 

> 3) In DataFrame.Sort in DataFrame.cs, we should simply change the old.Column.Clone(sortIndices, ...) to just oldColumn.Clone(sortIndices).

Just an FYI I followed the above for all Sort methods.

Kind Regards,
Ramon

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.Sort.cs
```diff
@@ -14,20 +14,53 @@ public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn
...
+                sortIndices = GetDescendingSortIndices();
+            }
+            return Clone(sortIndices);
         }
 
         internal override PrimitiveDataFrameColumn<long> GetAscendingSortIndices()
         {
             // The return sortIndices contains only the non null indices. 
-            GetSortIndices(Comparer<T>.Default, out PrimitiveDataFrameColumn<long> sortIndices);
+            PrimitiveDataFrameColumn<long> sortIndices = GetSortIndices(Comparer<T>.Default, out PrimitiveDataFrameColumn<long> columnNullIndices);
```
pgovind, line 32: Can you delete the 2 `// The return sortIndices contains only the non null indices.` comments here?

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
```diff
@@ -171,29 +171,73 @@ public IEnumerator<string> GetEnumerator()
...
 
-        private void GetSortIndices(Comparer<string> comparer, out PrimitiveDataFrameColumn<long> columnSortIndices)
+        internal override PrimitiveDataFrameColumn<long> GetDescendingSortIndices()
+        {
+            Comparison<string> descendingComparison = new Comparison<string>((string a, string b) =>
+            {
+                return Comparer<string>.Default.Compare(b, a);
+            });
+            Comparer<string> descending = Comparer<string>.Create(descendingComparison);
+            // The return sortIndices contains only the non null indices.
```
pgovind, line 206: Remove the comment here too?

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
```diff
@@ -171,29 +171,73 @@ public IEnumerator<string> GetEnumerator()
...
                 {
                     sortIndices[i] = i;
                 }
                 // TODO: Refactor the sort routine to also work with IList?
                 string[] array = buffer.ToArray();
                 IntrospectiveSort(array, array.Length, sortIndices, comparer);
+                // do a for loop here and get the nonnulls together
+                for(int i=0; i<sortIndices.Length; i++)
+                {
+                    if (buffer[i] is null)
```
pgovind, line 235: Shouldn't this be `buffer[sortIndices[i]]` (similar change in the `Append` below)?

pgovind (commented):
Just 1 comment and 2 nits about removing in-source comments. This is almost good to go!!

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
```diff
@@ -226,7 +226,7 @@ public override double Median()
             if (Length == 0)
                 return 0;
             PrimitiveDataFrameColumn<long> sortIndices = GetAscendingSortIndices();
-            long middle = sortIndices.Length / 2;
+            long middle = (sortIndices.Length - NullCount) / 2;
```
eerhardt, line 229: Do we have a test for this?

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.Sort.cs
```diff
@@ -14,20 +14,53 @@ public partial class PrimitiveDataFrameColumn<T> : DataFrameColumn
...
+            if(columnNullIndices.Length > 0)
```
eerhardt, line 33: (nit) You don't need a Length check before the `for` loop. The `for` loop will no-op if `Length == 0`.

src/Microsoft.Data.Analysis/StringDataFrameColumn.cs
eerhardt: I don't think that will work either. The issue is there are multiple `_stringBuffers`, and `i` is relative to the current buffer. So we need to account for how many buffers was before this buffer. Just like we do in PrimitiveDataFrameColumn.GetSortIndices.

---
In reply to: [501318342](https://github.com/dotnet/corefxlab/pull/2952#discussion_r501318342) [](ancestors = 501318342)

pgovind: Fixed this locally.

src/Microsoft.Data.Analysis/PrimitiveDataFrameColumn.cs
pgovind: Shockingly, no. Adding 1 now

pgovind:
Hey @RamonWill, I pushed a commit to address the comments on this PR. I'm hoping to release the next preview of DataFrame soon, and this PR was almost ready. I hope you don't mind!

tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs
```diff
@@ -1418,6 +1464,16 @@ public void TestColumnFilter()
             Assert.Equal(7, filtered[3]);
         }
 
+        [Fact]
+        public void TestMeanAndMedian()
+        {
+            Int32DataFrameColumn ints = new Int32DataFrameColumn("Ints", new int?[] { 1, 3, 3, 2, null, 9, 10 });
```
eerhardt, line 1470: The "median" test here isn't very great because if you don't ignore the `null`, the median is still `3`. Can we get a different set that would actually change the median if the `null` wasn't considered?

PR #2953, opened 2020-08-13
Resolves #2902 (Improve loadcsv error experience)

### LINKED ISSUE #2902: Improve error experience when LoadCsv throws

chriss2401:
Hi @pgovind and @eerhardt ,

As discussed here is the first out of the two small PR for MDA which resolves #2902. The workaround here is that when `Convert.ChangeType` does not succeed, we print a descriptive message of which value in which column causes this issue and to which type it cannot be converted. If the type that it was attempted to be converted to was `Double` or `Single`, we then assign their corresponding `NaN` values instead.

I've also added a simple unit test, which tries to trigger this behavior. Please review and comment at your convenience.

Christos.


src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -485,10 +485,29 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
-                        value = Convert.ChangeType(value, column.DataType);
-                        if (value is null)
+                        try
                         {
-                            throw new ArgumentException(string.Format(Strings.MismatchedValueType, column.DataType), value.GetType().ToString());
+                            value = Convert.ChangeType(value, column.DataType);
+                        }
+                        catch (FormatException)
+                        {
+                            Console.Write($"Value \"{value}\" cannot be converted to type {column.DataType} (Column name: {column.Name}). ");
```
eerhardt, line 494: We shouldn't be writing to the Console inside a library.

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -485,10 +485,29 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
+
+                            if (column.DataType == typeof(double))
+                            {
+                                Console.WriteLine("Converting to Double.NaN instead.");
+                                value = Double.NaN;
```
pgovind, line 499: I don't think we've tested any of our existing APIs with NaN values. Can you get test coverage by modifying DataFrameTests.cs:MakeDataFrameWithNumericColumns to generate a row with Double.NaN and Single.NaN values in the appropriate columns. That should automatically give us coverage over all the APIs

src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -485,10 +485,29 @@ public DataFrame Append(IEnumerable<object> row = null, bool inPlace = false)
...
+                            }
+                            else if (column.DataType == typeof(float))
+                            {
+                                Console.WriteLine("Converting to Single.NaN instead.");
+                                value = Single.NaN;
+                            }
+                            //throw FormatException with above message
```
pgovind, line 506: Can you also delete this comment?

src/Microsoft.Data.Analysis/DataFrame.cs
pgovind: >TestBinaryOperationsWithColumns 
Yup, that's the test I was worried about. 

I think the best approach here would be to use `null` instead of `NaN`. All the column types handle `null`  properly and I think it is reasonable to use `null` for now. When we eventually add support for `NaN`, we can re-visit this and change `null` to `NaN`!

chriss2401: Hey @pgovind ,

Changing `NaN` to `null` fixes most failing tests, but `TestAppendRow` fails [here](https://github.com/dotnet/corefxlab/blob/master/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs#L2070) , since these 3 lines expect a `FormatException`, but everything is returned to `null` now. Solution : Remove those 3 lines ? 

Otherwise it looks good, I will just wrap a try-catch around `Convert.ChangeType` and change everything to null if an exception is caught atm. I will also remove my `NaN` related unit test.

pgovind: Ah, that's a good test case. I'm surprised that it is not throwing a `FormatException`. The columns in that test case are `int` and `bool` right? So we should be hitting the last else condition here and throwing a `FormatException`? However, there is a bigger problem. Currently, if we had a `DoubleDataFrameColumn` and tried to append a string `"foo"` to it, we throw a `FormatException`. With this change, we would instead be appending a `null` value silently. I think the current behavior is more desirable here by default. 
1) We should definitely improve the exception message with the row number (`RowCount`), column name, the column's DataType and the value throwing the error. 
2) Keep the `if` blocks here that convert invalid values to `null` for `double` and `float` data types, but run them only when a new optional boolean flag to the `Append` method has been set. 

Thoughts @eerhardt ?

eerhardt: > Currently, if we had a DoubleDataFrameColumn and tried to append a string "foo" to it, we throw a FormatException. With this change, we would instead be appending a null value silently. I think the current behavior is more desirable here by default.

If you try appending using the API the wrong type to a column, it should throw.

When reading from a CSV file, it may make more sense to continue loading and put `null` or `NaN` in the cell instead. Maybe we could have a `strict` option when loading, that says whether to throw or ignore invalid data. Another thought would be to store the parse errors in a collection that can be inspected after loading.

Issue #2934, opened 2020-06-15
Column name indexing removed in .4?


Hamaze:
Why was indexing into a DataFrame removed in latest? Looking at the commit history too I can see the block was deleted. This makes very an awkward use experience now.

pgovind:
Hello! Which API are you talking about? I still see this: https://github.com/dotnet/corefxlab/blob/f44a099c2a3bb8b0feedc92cdf9f66aba793a82c/src/Microsoft.Data.Analysis/DataFrame.cs#L72

Hamaze:
I referred above the change I'm referring to

pgovind:
Ah I see. It still exists! We just moved that to the `DataFrameColumnCollection` class. See this: https://github.com/dotnet/corefxlab/blob/f44a099c2a3bb8b0feedc92cdf9f66aba793a82c/tests/Microsoft.Data.Analysis.Tests/DataFrameTests.cs#L648

rhysparry:
I think @Hamaze was specifically asking why the API was changed that way. Maybe you can point to the API review?

I know that I'd rather write:
```csharp
df["Int3"] = df["Int1"] * 2 + df["Int2"];
```
As opposed to:
```csharp
df.Columns["Int3"] = df.Columns["Int1"] * 2 + df.Columns["Int2"]; 
```

Hamaze:
Yeah, the former is more pandas-esque and comfortable IMO.

MikaelUmaN:
Agreed.

I suspect the reason is for the row filter to work.

But you are more often interested in column selection than row selection so it's better to have the penalty there instead.

```
df.Rows[2 .. 12]

df["Int3"] = df["Int1"] * 2 + df["Int2"];

```

pgovind:
Just out of curiosity, are you using DataFrame in a notebook? Reason I ask is that we've worked on a cool extension for DataFrame in notebooks that'll let you write `df.Int1 * 2 + df.Int2`. To be specific, with the new extension you can now refer to a column as a field of a `DataFrame` object. With intellisense enabled in notebooks, this will be very discoverable.

Hamaze:
I'm not using it in a notebook context for this exercise. Using it in an asp.net app.

MikaelUmaN:
Haven't tested the new notebook support yet but will do.

I would say that even though notebooks are very useful, I much prefer the experience to be the same when doing normal software and when doing notebooks.

[...]

PR #2959, opened 2020-08-28
ColumnName indexer on DataFrame

pgovind:
Fixes https://github.com/dotnet/corefxlab/issues/2934

This change brings back the column name indexer on `DataFrame`. I modified a couple test cases to test the new getter and setter.

cc @Hamaze

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.cs
@@ -148,6 +148,18 @@ namespace Microsoft.Data.Analysis
         }
 
         /// <summary>
+        /// An indexer based on <see cref="DataFrameColumn.Name"/>
+        /// </summary>
+        /// <param name="columnName">The name of a <see cref="DataFrameColumn"/></param>
+        /// <returns>A <see cref="DataFrameColumn"/> if it exists.</returns>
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.Analysis/DataFrame.cs
```diff
@@ -147,6 +147,18 @@ public DataFrame(params DataFrameColumn[] columns)
...
 
+        /// <summary>
+        /// An indexer based on <see cref="DataFrameColumn.Name"/>
+        /// </summary>
+        /// <param name="columnName">The name of a <see cref="DataFrameColumn"/></param>
+        /// <returns>A <see cref="DataFrameColumn"/> if it exists.</returns>
+        /// <exception cref="ArgumentException">Throws if <paramref name="columnName"/> is not present in this <see cref="DataFrame"/></exception>
+        public DataFrameColumn this[string columnName]
+        {
+            get => Columns[columnName];
```
gfoidl, line 158: The case that the column doesn't exist is handled and tested.
What's about the case of `columnName = null`?

pgovind: As it turns out, we have too many overloads of this indexer, so the only way to call this is to do `df[(string)null]`. I fixed the null case just in case though :)

PR #2978, opened 2020-10-13
Update FromArrowRecordBatches for dotnet-spark

pgovind:
2 things going on in this PR:

1. Update `FromArrowRecordBatch` just in case we have a `RecordBatch` with a `StructArray` in it. We'll flatten out the `StructArray` into a regular `DataFrame`. Once this goes in, I'll open another PR to update the version number for MDA.
2. Update the Arrow dependency to the latest version. This will prevent accidental "API not found" errors at runtime in the dotnet-spark repo. 

**OLD**
The following methods on `DataFrameColumn` are being made public:
1. GetArrowField
2. GetMaxRecordBatchLength
3. ToArrowArray

These 3 methods are the ones we need to support Spark 3.0.

There is an argument to be made here that these APIs should remain protected. The alternative we have here is to update just the existing `DataFrame.ToArrowRecordBatches()` method to return a Spark 3.0 compatible `RecordBatch`. Because dotnet-spark's dependencies on MDA are specified as exact versions, this should work and no backend changes would be needed on the dotnet-spark side! I'm inclined to update `DataFrame.ToArrowRecordBatches()` personally, but I don't mind making these 3  methods public either.

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
@@ -94,7 +94,7 @@ namespace Microsoft.Data.Analysis
                     StructArray structArray = (StructArray)arrowArray;
                     StructType structType = (StructType)field.DataType;
                     IEnumerator<Field> fieldsEnumerator = structType.Fields.GetEnumerator();
-                    IEnumerator<Apache.Arrow.Array> structArrayEnumerator = structArray.Fields.GetEnumerator();
+                    IEnumerator<IArrowArray> structArrayEnumerator = structArray.Fields.GetEnumerator();
                     while (fieldsEnumerator.MoveNext() && structArrayEnumerator.MoveNext())
                     {
                         AppendDataFrameColumnFromArrowArray(fieldsEnumerator.Current, structArrayEnumerator.Current, ret);
--- a/src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
@@ -11,11 +11,11 @@ namespace Microsoft.Data.Analysis
 {
     public partial class DataFrame
     {
-        private static void AppendDataFrameColumnFromArrowArray(Field field, IArrowArray arrowArray, DataFrame ret, string structName = "")
+        private static void AppendDataFrameColumnFromArrowArray(Field field, IArrowArray arrowArray, DataFrame ret, string fieldNamePrefix = "")
         {
             IArrowType fieldType = field.DataType;
             DataFrameColumn dataFrameColumn = null;
-            string fieldName = structName + field.Name;
+            string fieldName = fieldNamePrefix + field.Name;
             switch (fieldType.TypeId)
             {
                 case ArrowTypeId.Boolean:
@@ -98,7 +98,7 @@ namespace Microsoft.Data.Analysis
                     IEnumerator<IArrowArray> structArrayEnumerator = structArray.Fields.GetEnumerator();
                     while (fieldsEnumerator.MoveNext() && structArrayEnumerator.MoveNext())
                     {
-                        AppendDataFrameColumnFromArrowArray(fieldsEnumerator.Current, structArrayEnumerator.Current, ret, field.Name + ".");
+                        AppendDataFrameColumnFromArrowArray(fieldsEnumerator.Current, structArrayEnumerator.Current, ret, field.Name + "_");
                     }
                     break;
                 case ArrowTypeId.Decimal:
--- a/src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
@@ -3,11 +3,7 @@
 // See the LICENSE file in the project root for more information.
 
 using System;
-using System.Collections;
 using System.Collections.Generic;
-using System.Diagnostics;
-using System.Text;
-using System.Threading.Tasks;
 using Apache.Arrow;
 using Apache.Arrow.Types;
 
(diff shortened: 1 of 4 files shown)


src/Microsoft.Data.Analysis/DataFrameColumn.cs
```diff
@@ -223,12 +223,12 @@ public virtual DataFrameColumn Sort(bool ascending = true)
         protected virtual DataFrameColumn FillNullsImplementation(object value, bool inPlace) => throw new NotImplementedException();
 
         // Arrow related APIs
-        protected internal virtual Field GetArrowField() => throw new NotImplementedException();
+        public virtual Field GetArrowField() => throw new NotImplementedException();
```
eerhardt, line 226: Since these are now public APIs, we should have unit tests for them.

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
```diff
@@ -137,14 +151,18 @@ public IEnumerable<RecordBatch> ToArrowRecordBatches()
...
             for (int i = 0; i < columnCount; i++)
             {
                 DataFrameColumn column = Columns[i];
                 Field field = column.GetArrowField();
-                schemaBuilder.Field(field);
+                fields.Add(field);
             }
 
-            Schema schema = schemaBuilder.Build();
+            StructType structType = new StructType(fields);
```
eerhardt, line 162: Wrapping the arrays in a `StructType` feels too much like Spark-specific functionality here to me. I think we should move that logic to dotnet/spark instead of having it in here.

My thinking is that if I had a `DataFrame` with 5 columns, I should get a `RecordBatch` with 5 columns. Not 1 `StructType` column with 5 fields.

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
```diff
@@ -29,101 +137,7 @@ public static DataFrame FromArrowRecordBatch(RecordBatch recordBatch)
             foreach (IArrowArray arrowArray in arrowArrays)
             {
                 Field field = arrowSchema.GetFieldByIndex(fieldIndex);
-                IArrowType fieldType = field.DataType;
```
pgovind, line 32: All of this is just refactored into the `AppendDataFrameColumnFromArrowArray` method

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
pgovind: Removed this now.

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
```diff
@@ -3,18 +3,126 @@
...
+                    dataFrameColumn = new UInt32DataFrameColumn(field.Name, uintValueBuffer, uintNullBitMapBuffer, arrowArray.Length, arrowArray.NullCount);
+                    break;
+                case ArrowTypeId.UInt64:
+                    PrimitiveArray<ulong> arrowUlongArray = (PrimitiveArray<ulong>)arrowArray;
+                    ReadOnlyMemory<byte> ulongValueBuffer = arrowUlongArray.ValueBuffer.Memory;
+                    ReadOnlyMemory<byte> ulongNullBitMapBuffer = arrowUlongArray.NullBitmapBuffer.Memory;
+                    dataFrameColumn = new UInt64DataFrameColumn(field.Name, ulongValueBuffer, ulongNullBitMapBuffer, arrowArray.Length, arrowArray.NullCount);
+                    break;
+                case ArrowTypeId.Struct:
+                    StructArray structArray = (StructArray)arrowArray;
```
pgovind, line 94: This case is the only new addition.

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
```diff
@@ -3,18 +3,126 @@
...
+                    StructType structType = (StructType)field.DataType;
+                    IEnumerator<Field> fieldsEnumerator = structType.Fields.GetEnumerator();
+                    IEnumerator<IArrowArray> structArrayEnumerator = structArray.Fields.GetEnumerator();
+                    while (fieldsEnumerator.MoveNext() && structArrayEnumerator.MoveNext())
+                    {
+                        AppendDataFrameColumnFromArrowArray(fieldsEnumerator.Current, structArrayEnumerator.Current, ret);
```
eerhardt, line 100: Should we append the `Struct`'s field name to the beginning of each sub-column's name? That way we won't get column name collisions.

tests/Microsoft.Data.Analysis.Tests/ArrayComparer.cs
```diff
@@ -54,6 +55,24 @@ public ArrayComparer(IArrowArray expectedArray)
...
+        {
+            Assert.IsAssignableFrom<StructArray>(_expectedArray);
+            StructArray expectedArray = (StructArray)_expectedArray;
+
+            Assert.Equal(expectedArray.Length, array.Length);
+            Assert.Equal(expectedArray.NullCount, array.NullCount);
+            Assert.Equal(expectedArray.Offset, array.Offset);
+            Assert.Equal(expectedArray.Data.Children.Length, array.Data.Children.Length);
+
+            var arrayFieldsEnumerator = array.Fields.GetEnumerator();
```
eerhardt, line 68: We should be comparing the 2 `Fields.Count` are equal. If `array.Fields` is less, the while loop below will just skip the `expectedArray.Fields`.

Check out how we do it in the Arrow tests:

[https://github.com/apache/arrow/blob/04660f8afbf5bee052f8881b36b5c59a8c8ca7b5/csharp/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs#L120-L125](https://github.com/apache/arrow/blob/04660f8afbf5bee052f8881b36b5c59a8c8ca7b5/csharp/test/Apache.Arrow.Tests/ArrowReaderVerifier.cs#L120-L125)

tests/Microsoft.Data.Analysis.Tests/ArrowIntegrationTests.cs
```diff
@@ -49,14 +50,26 @@ public void TestArrowIntegration()
                 .Append("UByteColumn", false, new UInt8Array.Builder().AppendRange(Enumerable.Repeat((byte)1, 10)).Build())
                 .Build();
 
-            DataFrame df = DataFrame.FromArrowRecordBatch(originalBatch);
+            ArrowBuffer.BitmapBuilder nullBitmapBufferBuilder = new ArrowBuffer.BitmapBuilder();
```
eerhardt, line 53: Can you make a new test for this scenario instead? That way we ensure the original test works as it previously did.

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
pgovind: Done.

tests/Microsoft.Data.Analysis.Tests/ArrayComparer.cs
pgovind: Done

tests/Microsoft.Data.Analysis.Tests/ArrowIntegrationTests.cs
pgovind: Done.

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
```diff
@@ -3,18 +3,127 @@
...
-using System.Text;
-using System.Threading.Tasks;
 using Apache.Arrow;
 using Apache.Arrow.Types;
 
 namespace Microsoft.Data.Analysis
 {
     public partial class DataFrame
     {
+        private static void AppendDataFrameColumnFromArrowArray(Field field, IArrowArray arrowArray, DataFrame ret, string structName = "")
```
eerhardt, line 14: (nit) I wouldn't call this `structName`, but instead `namePrefix` or similar.

src/Microsoft.Data.Analysis/DataFrame.Arrow.cs
```diff
@@ -3,18 +3,127 @@
...
+                    dataFrameColumn = new UInt64DataFrameColumn(fieldName, ulongValueBuffer, ulongNullBitMapBuffer, arrowArray.Length, arrowArray.NullCount);
+                    break;
+                case ArrowTypeId.Struct:
+                    StructArray structArray = (StructArray)arrowArray;
+                    StructType structType = (StructType)field.DataType;
+                    IEnumerator<Field> fieldsEnumerator = structType.Fields.GetEnumerator();
+                    IEnumerator<IArrowArray> structArrayEnumerator = structArray.Fields.GetEnumerator();
+                    while (fieldsEnumerator.MoveNext() && structArrayEnumerator.MoveNext())
+                    {
+                        AppendDataFrameColumnFromArrowArray(fieldsEnumerator.Current, structArrayEnumerator.Current, ret, field.Name + ".");
```
eerhardt, line 101: Should we use `_` instead of `.`? That way it can be a valid identifier.

pgovind: Ah good idea. Yup, I'll do that

tests/Microsoft.Data.Analysis.Tests/ArrowIntegrationTests.cs
```diff
@@ -60,6 +61,71 @@ public void TestArrowIntegration()
...
+                    .Append(prependColumnNamesWith + "ULongColumn", false, new UInt64Array.Builder().AppendRange(Enumerable.Repeat((ulong)1, 10)).Build())
+                    .Append(prependColumnNamesWith + "ByteColumn", false, new Int8Array.Builder().AppendRange(Enumerable.Repeat((sbyte)1, 10)).Build())
+                    .Append(prependColumnNamesWith + "UByteColumn", false, new UInt8Array.Builder().AppendRange(Enumerable.Repeat((byte)1, 10)).Build())
+                    .Build();
+
+                return ret;
+            }
+
+            RecordBatch originalBatch = CreateRecordBatch();
+            ArrowBuffer.BitmapBuilder nullBitmapBufferBuilder = new ArrowBuffer.BitmapBuilder();
```
eerhardt, line 105: (nit) (and we probably need to fix it everywhere in Arrow), we should move away from the term "null bitmap", because the values in it are `true` when the value is valid and `false` when the value is `null`. Thus the more correct term is `validity bitmap`.

See https://arrow.apache.org/docs/format/Columnar.html#validity-bitmaps

pgovind:
Now that we've determined that Spark is unlikely to need this new API, we can keep the `Struct_childColumnName` I think. Other than that, this PR should be good to go in

eerhardt (approved):
:shipit:

PR #2980, opened 2020-10-16
Fix build break with an arcade property

pgovind:


eng/Configurations.props
```diff
@@ -0,0 +1,4 @@
+<Project>
+    <!-- Turn off end of life target framework checks as we intentionally build older .NETCoreApp configurations. -->
+    <CheckEolTargetFramework>false</CheckEolTargetFramework>
```
safern, line 3: Let's not add a new file for this. We can just set this in `Directory.Build.props` in the dotnet/runtime repo it is in  `Configurations.props` as that is the shared file across all pieces for this kind of settings.

pgovind:
Closing in favor of https://github.com/dotnet/corefxlab/pull/2974

Issue #2984, opened 2020-10-22
[DataFrame] Add a LoadCsv overload or other method that takes CSV content directly  


jonsequitur:
Getting from a CSV string to a `DataFrame` requires some ceremony involving creating a stream first. While this might not be the ideal approach from a performance perspective for larger volumes of data, it's very useful for prototyping with sample subsets of data. An API like this will also play nicely with the .NET Interactive `#!value` magic command.

```csharp
DataFrame.LoadCsvFromString(@"col1,col2,col3
1,2,3
4,5,6
");
```

pgovind:
I think this is a good idea. I'll put a PR up for this soon and we can include this in the next preview

PR #2988, opened 2020-10-28
Implement DataFrame.LoadCsvFromString

pgovind:
Fixes https://github.com/dotnet/corefxlab/issues/2984

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -175,8 +175,8 @@ namespace Microsoft.Data.Analysis
         private static DataFrame ReadCsvLinesIntoDataFrame(IEnumerable<string> lines,
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
-                                long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false,
-                                Encoding encoding = null)
+                                long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false
+                                )
         {
             if (dataTypes == null && guessRows <= 0)
             {
@@ -272,7 +272,7 @@ namespace Microsoft.Data.Analysis
             return ret;
         }
 
-        private struct CsvLines : IEnumerable<string>
+        private class CsvLines : IEnumerable<string>
         {
             private CsvLineEnumerator enumerator;
             public CsvLines(CsvLineEnumerator csvLineEnumerator)
@@ -285,7 +285,7 @@ namespace Microsoft.Data.Analysis
             IEnumerator IEnumerable.GetEnumerator() => enumerator;
         }
 
-        private struct CsvLineEnumerator : IEnumerator<string>
+        private class CsvLineEnumerator : IEnumerator<string>
         {
             private StreamReader streamReader;
             private string currentLine;
@@ -330,16 +330,14 @@ namespace Microsoft.Data.Analysis
         /// <param name="numberOfRowsToRead">number of rows to read not including the header(if present)</param>
         /// <param name="guessRows">number of rows used to guess types</param>
         /// <param name="addIndexColumn">add one column with the row index</param>
-        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
         /// <returns><see cref="DataFrame"/></returns>
         public static DataFrame LoadCsvFromString(string csvString,
                                 char separator = ',', bool header = true,
                                 string[] columnNames = null, Type[] dataTypes = null,
-                                long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false,
-                                Encoding encoding = null)
+                                long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false)
         {
             string[] lines = csvString.Split(new[] { Environment.NewLine }, StringSplitOptions.None);
-            return ReadCsvLinesIntoDataFrame(lines, separator, header, columnNames, dataTypes, numberOfRowsToRead, guessRows, addIndexColumn, encoding);
+            return ReadCsvLinesIntoDataFrame(lines, separator, header, columnNames, dataTypes, numberOfRowsToRead, guessRows, addIndexColumn);
         }
 
         /// <summary>
@@ -375,7 +373,7 @@ namespace Microsoft.Data.Analysis
             {
                 CsvLineEnumerator linesEnumerator = new CsvLineEnumerator(streamReader);
                 IEnumerable<string> lines = new CsvLines(linesEnumerator);
-                return ReadCsvLinesIntoDataFrame(lines, separator, header, columnNames, dataTypes, numberOfRowsToRead, guessRows, addIndexColumn, encoding);
+                return ReadCsvLinesIntoDataFrame(lines, separator, header, columnNames, dataTypes, numberOfRowsToRead, guessRows, addIndexColumn);
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -174,31 +172,12 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
-        /// <param name="header">has a header or not</param>
-        /// <param name="columnNames">column names (can be empty)</param>
-        /// <param name="dataTypes">column types (can be empty)</param>
-        /// <param name="numberOfRowsToRead">number of rows to read not including the header(if present)</param>
-        /// <param name="guessRows">number of rows used to guess types</param>
-        /// <param name="addIndexColumn">add one column with the row index</param>
-        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
-        /// <returns><see cref="DataFrame"/></returns>
-        public static DataFrame LoadCsv(Stream csvStream,
+        private static DataFrame ReadCsvLinesIntoDataFrame(IEnumerable<string> lines,
```
pgovind, line 175: This method is just a refactor of the logic in `LoadCsv` with 1 simple change: The first parameter here is an `IEnumerable<string>`.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -214,87 +193,189 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+            }
+
+            if (dataTypes == null && guessRows <= 0)
+            {
+                throw new ArgumentException(string.Format(Strings.ExpectedEitherGuessRowsOrDataTypes, nameof(guessRows), nameof(dataTypes)));
+            }
+
+            using (var streamReader = new StreamReader(csvStream, encoding ?? Encoding.UTF8, detectEncodingFromByteOrderMarks: true, DefaultStreamReaderBufferSize, leaveOpen: true))
+            {
+                CsvLineEnumerator linesEnumerator = new CsvLineEnumerator(streamReader);
```
pgovind, line 376: Just a couple structs to wrap a `StreamReader` within an `IEnumerable<string>`.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -214,87 +193,189 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+        /// <param name="addIndexColumn">add one column with the row index</param>
+        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        /// <returns><see cref="DataFrame"/></returns>
+        public static DataFrame LoadCsvFromString(string csvString,
+                                char separator = ',', bool header = true,
+                                string[] columnNames = null, Type[] dataTypes = null,
+                                long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false,
+                                Encoding encoding = null)
+        {
+            string[] lines = csvString.Split(new[] { Environment.NewLine }, StringSplitOptions.None);
```
eerhardt, line 341: Are we just completely giving up on quoting? Or is this just a "first effort" to get the API in?

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -214,87 +193,189 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+        /// <param name="numberOfRowsToRead">number of rows to read not including the header(if present)</param>
+        /// <param name="guessRows">number of rows used to guess types</param>
+        /// <param name="addIndexColumn">add one column with the row index</param>
+        /// <param name="encoding">The character encoding. Defaults to UTF8 if not specified</param>
+        /// <returns><see cref="DataFrame"/></returns>
+        public static DataFrame LoadCsvFromString(string csvString,
+                                char separator = ',', bool header = true,
+                                string[] columnNames = null, Type[] dataTypes = null,
+                                long numberOfRowsToRead = -1, int guessRows = 10, bool addIndexColumn = false,
+                                Encoding encoding = null)
```
eerhardt, line 339: Encoding doesn't make any sense for a string. strings are always UTF-16.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -214,87 +193,189 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
-                    columns.Insert(0, indexColumn);
+                    indexColumn[i] = i;
                 }
-                return ret;
+                columns.Insert(0, indexColumn);
+            }
+            return ret;
+        }
+
+        private struct CsvLines : IEnumerable<string>
```
eerhardt, line 275: Does this need to be a `struct`? You are going to box it as soon as you pass it into `ReadCsvLinesIntoDataFrame(IEnumerable<string> lines` anyway.

If you really wanted to save the single alloc, you could make `ReadCsvLinesIntoDataFrame(IEnumerable<string> lines` be generic `where T : IEnumerable<string>`.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -214,87 +193,189 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+                currentLine = null;
+            }
+
+            public string Current => currentLine;
+
+            object IEnumerator.Current => currentLine;
+
+            public void Dispose()
+            {
+                throw new NotImplementedException();
```
eerhardt, line 306: Why throw on Dispose?

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -214,87 +193,189 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
...
+            }
+
+            public bool MoveNext()
+            {
+                currentLine = streamReader.ReadLine();
+                return currentLine != null;
+            }
+
+            public void Reset()
```
eerhardt, line 315: Do we need to support Reset()? Who calls it? You can save a `long` field if it wasn't implemented.

eerhardt: Ah, nevermind. I see where we call it now.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
pgovind: This is just a "first effort". Or more accurately the easiest way to get this API in with just some minor refactoring. I'll work on a proper csv parser shortly and re-visit this implementation at that point. I still think splitting on a new line and sending each line to the csv parser will work.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
pgovind: Ah yup you're right. Fixed

src/Microsoft.Data.Analysis/DataFrame.IO.cs
pgovind: Oh, I didn't realize that value type -> interfaces also does boxing. Just looked it up. Nice! Fixed it now. Just made it a class.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
pgovind: I put that in so we catch it if someone calls this method. We're not dealing with any unmanaged resources here, so no one should be calling `Dispose`

eerhardt (approved):
LGTM

PR #2997, opened 2020-11-13
Support for delimited/fixed-width fields in a text file

pgovind:
I have the whole thing working locally, but I split it up for easier review. This PR is part 1 of 2. This part sets up the TextFieldParser class with configurations for delimited/fixed-width files and implements reading the lines into memory. 

Part 2 will implement `public ReadFields()` that will parse the lines that are read in and return individual fields from each line.

### DIFF
--- a/src/Microsoft.Data.Analysis/TextFieldParser.cs
+++ b/src/Microsoft.Data.Analysis/TextFieldParser.cs
@@ -1,0 +1,614 @@
+﻿// Licensed to the .NET Foundation under one or more agreements.
+// The .NET Foundation licenses this file to you under the MIT license.
+// See the LICENSE file in the project root for more information.
+
+using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Diagnostics;
+using System.IO;
+using System.Runtime.CompilerServices;
+using System.Text;
+using System.Text.RegularExpressions;
+
+[assembly: InternalsVisibleTo("Microsoft.Data.Analysis.Tests, PublicKey= 0024000004800000940000000602000000240000525341310004000001000100b5fc90e7027f67871e773a8fde8938c81dd402ba65b9201d60593e96c492651e889cc13f1415ebb53fac1131ae0bd333c5ee6021672d9718ea31a8aebd0da0072f25d87dba6fc90ffd598ed4da35e44c398c454307e8e33b8426143daec9f596836f97c8f74750e5975c64e2189f45def46b2a2b1247adc3652bf5c308055da9")]
+
+namespace Microsoft.Data.Analysis
+{
+    internal enum FieldType
+    {
+        Delimited,
+        FixedWidth
+    }
+
+    internal class TextFieldParser : IDisposable
+    {
+        private delegate int ChangeBufferFunction();
+
+        private bool _disposed;
+
+        private TextReader _reader;
+
+        private string[] _commentTokens = null;
+
+        private long _lineNumber = 1;
+
+        private bool _endOfData;
+
+        private string _errorLine = "";
+
+        private long _errorLineNumber = -1;
+
+        private FieldType _textFieldType = FieldType.Delimited;
+
+        private int[] _fieldWidths;
+
+        private int[] _fieldWidthsCopy;
+
+        private string[] _delimiters;
+
+        private string[] _delimitersCopy;
+
+        private Regex _whiteSpaceRegEx = new Regex("\\s", RegexOptions.CultureInvariant);
+
+        private bool _trimWhiteSpace = true;
+
+        private int _position;
+
+        private int _peekPosition;
+
+        private int _charsRead;
+
+        private const int DEFAULT_BUFFER_LENGTH = 4096;
+
+        private char[] _buffer = new char[DEFAULT_BUFFER_LENGTH];
+
+        private bool _hasFieldsEnclosedInQuotes = true;
+
+        private int _maxBufferSize = 10000000;
+
+        private bool _leaveOpen;
+
+        private char[] newLineChars = Environment.NewLine.ToCharArray();
+
+        public string[] CommentTokens
+        {
+            get => _commentTokens;
+            set
+            {
+                CheckCommentTokensForWhitespace(value);
+                _commentTokens = value;
+            }
+        }
+
+        public bool EndOfData
+        {
+            get
+            {
+                if (_endOfData)
+                {
+                    return _endOfData;
+                }
+                if ((_reader == null) | (_buffer == null))
+                {
+                    _endOfData = true;
+                    return true;
+                }
+                if (PeekNextDataLine() != null)
+                {
+                    return false;
+                }
+                _endOfData = true;
+                return true;
+            }
+        }
+
+        public long LineNumber
+        {
+            get
+            {
+                if (_lineNumber != -1 && ((_reader.Peek() == -1) & (_position == _charsRead)))
+                {
+                    // Side effect of a property. Not great. Just leaving it in for now.
+                    CloseReader();
+                }
+                return _lineNumber;
+            }
+        }
+
+        public string ErrorLine => _errorLine;
+
+        public long ErrorLineNumber => _errorLineNumber;
+
+        public FieldType TextFieldType
+        {
+            get =>_textFieldType;
+            set
+            {
+                ValidateFieldTypeEnumValue(value, "value");
+                _textFieldType = value;
+            }
+        }
+
+        public int[] FieldWidths
+        {
+            get =>_fieldWidths;
+            private set
+            {
+                if (value != null)
+                {
+                    ValidateFieldWidthsOnInput(value);
+                    _fieldWidthsCopy = (int[])value.Clone();
+                }
+                else
+                {
+                    _fieldWidthsCopy = null;
+                }
+                _fieldWidths = value;
+            }
+        }
+
+        public string[] Delimiters
+        {
+            get => _delimiters;
+            private set
+            {
+                if (value != null)
+                {
+                    ValidateDelimiters(value);
+                    _delimitersCopy = (string[])value.Clone();
+                }
+                else
+                {
+                    _delimitersCopy = null;
+                }
+                _delimiters = value;
+            }
+        }
+
+        public bool TrimWhiteSpace
+        {
+            get =>_trimWhiteSpace;
+            set
+            {
+                _trimWhiteSpace = value;
+            }
+        }
+
+        public bool HasFieldsEnclosedInQuotes
+        {
+            get =>_hasFieldsEnclosedInQuotes;
+            set
+            {
+                _hasFieldsEnclosedInQuotes = value;
+            }
+        }
+
+        public TextFieldParser(string path)
+        {
+            InitializeFromPath(path, Encoding.ASCII, detectEncoding: true);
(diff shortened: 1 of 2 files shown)


src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+
+namespace Microsoft.Data.Analysis
+{
+    public enum FieldType
+    {
+        Delimited,
+        FixedWidth
+    }
+
+    public class TextFieldParser : IDisposable
```
pgovind, line 22: We could in theory split this off as a separate csproj to make its own Nuget package. Something to consider ...

pgovind, line 22: Also, don't bother about docs just yet. I'll add the docs in Part 2

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+    {
+        private delegate int ChangeBufferFunction();
+
+        private bool _disposed;
+
+        private TextReader _reader;
+
+        private string[] _commentTokens = new string[0];
```
gfoidl, line 30:
```suggestion
        private string[] _commentTokens = Array.Empty<string>();
```

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+        private bool _leaveOpen;
+
+        private char[] newLineChars = Environment.NewLine.ToCharArray();
+
+        public string[] CommentTokens
+        {
+            get
+            {
+                return _commentTokens;
+            }
```
gfoidl, line 79:
```suggestion
            get => _commentTokens
```
? Makes the code shorter.
On other places as well.

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+            InitializeFromStream(stream, defaultEncoding, detectEncoding);
+        }
+
+        public TextFieldParser(TextReader reader)
+        {
+            if (reader == null)
+            {
+                throw new ArgumentNullException(nameof(reader));
+            }
+            _reader = reader;
```
gfoidl, line 248:
```suggestion
            _reader = reader ?? throw new ArgumentNullException(nameof(reader));
```

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+
+        private void ValidateFieldTypeEnumValue(FieldType value, string paramName)
+        {
+            if (value < FieldType.Delimited || value > FieldType.FixedWidth)
+            {
+                throw new InvalidEnumArgumentException(paramName, (int)value, typeof(FieldType));
+            }
+        }
+
+        ~TextFieldParser()
```
gfoidl, line 365: There are no unmanaged ressources, so the finalizer isn't needed.

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+            {
+                throw new ArgumentNullException(nameof(defaultEncoding));
+            }
+            _reader = new StreamReader(stream, defaultEncoding, detectEncoding);
+            ReadToBuffer();
+        }
+
+        private string ValidatePath(string path)
+        {
+            if (!File.Exists(path))
```
gfoidl, line 426: Just to point out: this doesn't guarantee that at the time of use of the file it actually exists.
The file could be deleted between this check and the actual use. So there's a race condition.

But I would leave it as is, because it's a best effort solution which is perfectly fine.

gfoidl (commented):
There are quite a lot of allocations. 
If these methods are hot, we should consider spanifying some things to avoid those allocations and have more efficient parsing (of lines).

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -326,7 +326,7 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
         /// <param name="csvStream">stream of CSV data to be write out</param>
         /// <param name="separator">column separator</param>
         /// <param name="header">has a header or not</param>
-        /// <param name="encoding">the character encoding. Defaults to UTF8 if not specified</param>
+        /// <param name="encoding">the character encoding. Defaults to ASCII if not specified</param>
```
eerhardt, line 329: We should keep the same docs, and change the code. It is wrong for ReadCsv to default to Utf8 and WriteCsv to default to Ascii. Let's use Utf8 for both defaults.

src/Microsoft.Data.Analysis/TextFieldParser.cs
eerhardt, line 22: Does this need to be public?

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+            string line;
+            do
+            {
+                line = ReadNextLine(ref _peekPosition, BufferFunction);
+            }
+            while (IgnoreLine(line));
+            return line;
+        }
+
+        private string ReadNextLine(ref int Cursor, ChangeBufferFunction ChangeBuffer)
```
eerhardt, line 534: (nit) parameter names should be camel cased.

src/Microsoft.Data.Analysis/TextFieldParser.cs
eerhardt: Or just leave it `null` to start.

---
In reply to: [523403969](https://github.com/dotnet/corefxlab/pull/2997#discussion_r523403969) [](ancestors = 523403969)

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+
+        private int[] _fieldWidthsCopy;
+
+        private string[] _delimiters;
+
+        private string[] _delimitersCopy;
+
+        private int[] _whitespaceCodes = new int[] { '\u0020' };
+
+        private Regex _whiteSpaceRegEx = new Regex("\\s", RegexOptions.CultureInvariant);
```
eerhardt, line 52: It would probably be faster just to use `char.IsWhiteSpace` than this Regex.

src/Microsoft.Data.Analysis/TextFieldParser.cs
```diff
@@ -0,0 +1,642 @@
...
+
+        private int[] _fieldWidths;
+
+        private int[] _fieldWidthsCopy;
+
+        private string[] _delimiters;
+
+        private string[] _delimitersCopy;
+
+        private int[] _whitespaceCodes = new int[] { '\u0020' };
```
eerhardt, line 50: This doesn't appear used.

src/Microsoft.Data.Analysis/TextFieldParser.cs
pgovind: Ah, I thought I removed this. Artifact of using ILSpy to convert this code :). Fixed now

src/Microsoft.Data.Analysis/TextFieldParser.cs
pgovind: Not necessarily. 
1) I wasn't sure if we wanted to make `TextFieldParser` it's own Nuget package. 
2) It was easier to test with it being public. 

Made it internal now. Also using internalsvisibleto for the tests

src/Microsoft.Data.Analysis/TextFieldParser.cs
pgovind: Nice catch. Removed!

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -17,13 +17,12 @@
        -->
     <NoWarn>$(NoWarn);1591;NU5100</NoWarn>
     <TargetsForTfmSpecificContentInPackage>$(TargetsForTfmSpecificContentInPackage);AddMDAIToInteractiveExtensionsFolder</TargetsForTfmSpecificContentInPackage>
+    <AssemblyOriginatorKeyFile>C:\Users\prgovi\.nuget\packages\microsoft.dotnet.arcade.sdk\6.0.0-beta.20525.1\tools\snk/35MSSharedLib1024.snk</AssemblyOriginatorKeyFile>
```
eerhardt, line 20: This can't be checked in

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -17,13 +17,12 @@
...
   </PropertyGroup>
 
   <!-- The following properties are set to package M.D.A.Interactive with the M.D.A nuget package. If M.D.A.I undergoes TFM or dependency changes, we need to update the TargetFramework passed in below-->
   <Target Name="AddMDAIToInteractiveExtensionsFolder">
-    <MSBuild Projects="./../Microsoft.Data.Analysis.Interactive/Microsoft.Data.Analysis.Interactive.csproj"
```
eerhardt, line 24: I know VS fights us here, but it is easier to read the original way.

src/Microsoft.Data.Analysis/DataFrame.IO.cs
```diff
@@ -344,7 +344,7 @@ private static DataFrameColumn CreateColumn(Type kind, string[] columnNames, int
 
             if (encoding is null)
             {
-                encoding = Encoding.ASCII;
+                encoding = Encoding.UTF8;
```
pgovind, line 347: A unit test is failing due to this change. I wonder if we're now seeing BOM in the beginning of the stream and consequently failing the parse

pgovind: What do you think about keeping this as ASCII for now so I can get this PR(and it's follow up) in? I'll log a bug to change this to UTF8. Tracking down what is happening here might take some time. @eerhardt

eerhardt: That sounds fine to me. Log a bug and fix it in a separate PR.

pgovind: Filed https://github.com/dotnet/corefxlab/issues/3003

pgovind: Alright, moving back to ASCII fixed the CI failures. Should be ready to review now

src/Microsoft.Data.Analysis/Microsoft.Data.Analysis.csproj
```diff
@@ -1,5 +1,5 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
-  
+
```
pgovind, line 2: Delete the changes to this file

eerhardt (approved):
Looks good

pgovind:
Thanks for the reviews @gfoidl and @eerhardt

PR #3012, opened 2020-12-16
DataFrame.Load(IDataReader)

MarkPflug:
This was spurred by a discussion here: 
https://github.com/dotnet/runtime/discussions/37711

Basically, this allows loading from any data source that exposes data as IDataReader, which would include pretty much all relational databases, as well as a variety of csv data sources (most CSV apis don't support strongly typed schemas, so somewhat less valuable than LoadCsv). 

This is a draft work in progress. I'm curious if this is functionality you all would be interested in merging before I spend any more time on it. 

Still needs:
1) fix up todos and not implemented
2) docs
3) tests

### DIFF
--- a/src/Microsoft.Data.Analysis/DataFrame.IO.cs
+++ b/src/Microsoft.Data.Analysis/DataFrame.IO.cs
@@ -236,7 +236,7 @@ namespace Microsoft.Data.Analysis
             for (int i = 0; i < numberOfColumns; ++i)
             {
                 Type kind = dataTypes == null ? GuessKind(i, linesForGuessType) : dataTypes[i];
-                columns.Add(CreateColumn(kind, columnNames[i], i));
+                columns.Add(CreateColumn(kind, columnNames?[i], i));
             }
 
             DataFrame ret = new DataFrame(columns);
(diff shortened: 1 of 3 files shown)


MarkPflug:
Likewise, it might be useful for DataFrame to expose the data it contains as an IDataReader for interop in the other direction.

// I think that would be exposed by adding a single public method:
DbDataReader GetDataReader();

Everything else could be completely opaque behind the DbDataReader. This could be done as a separate PR.

MarkPflug:
Lesson learned: look at open PRs before creating a new PR. I guess this duplicates the effort of https://github.com/dotnet/corefxlab/pull/2924

Was that PR going to move forward?

pgovind:
Hey @MarkPflug ! I did follow the recent discussion in the csv thread over at runtime. Thanks for opening this PR. As you found out though, #2924 already implements reading from an IDataReader. I am interested in merging that PR. What do you think about adding tests to #2924 instead of working on this PR? Reason I prefer #2924 is that I think it's a little bit cleaner than this PR (since it doesn't touch `GetColumnName`).

MarkPflug:
I came to the realization that I don't really have the time to do complete this work. Sorry.

PR #3025, opened 2021-02-11
remove fake keys

danmoseley:


danmoseley:
This repo appears to have a dependency on myget: `D:\a\1\s\artifacts\toolset\restore.proj : error : Unable to load the service index for source https://dotnet.myget.org/F/dotnet-try/api/v3/index.json.`

@pgovind I suggest we ask @terrajobst to flip the archived flag, since the build doesn't even work. I do not want it accidentally using myget. Sound OK?

Meanwhile merging to avoid tools flagging this.

pgovind:
> Sound OK?

Yup. I could've sworn I removed all the myget stuff from this repo...

danmoseley:
I pushed a change to remove it from nuget.config just now.

